PR #198, opened 2014-11-16
Fneves/autocomplete tags on deploy

fneves:
## Description

This pull request adds the ability of having autocomplete with branches and tags on the autocomplete screen.
It also bumps up the version of typeahead javascript plugin supporting the new suggestion engine Bloodhound.

This works in two different ways:
1. If there is no locally cached repository, it gets the tags and branches remotely using git ls-remote command.
2. If a locally cached repository exists, it pulls the last changes and queries it for its branches and tags. 

All the changes are cached for so that multiple consecutive requests don't trigger multiple pulls or git ls-remotes in a short period of time. This cache uses a default TTL of 30 minutes and has an extra parameter (hit_count) that guarantees that if we hit the endpoint at least 6 times, it will refresh the cache even before the TTL.

The option of using git ls-remote when the repository is not locally cached is due to the fact that a big repository when cloned might take some time, and the repository would be locked for all that time.

This is an autocomplete feature meaning that the user does not really want to wait a long time to get the autocomplete suggestions and it would be strange that his deployment would then be waiting for a lock on the git repo. 
This way we use the git ls-remote to return the tags long way before the repository would be cloned.

## Screenshots: 

versions

autocomplete

### References
 - Jira link: [SAMSON-109] (https://zendesk.atlassian.net/browse/SAMSON-109)

### Risks
 - There was some refactoring on the JobExecution so that the logic of repository setup could be shared between two controllers. Although all the tests should cover this refactoring we should keep an eye for possible problems that might show up from cloning repositories.


fneves:
@steved555 I don't think anything bad would happen as the ```clean``` is basically a ```rm_rf``` operation that would occur before the repo being cloned. If the directory wasn't there it would just ignore it. I've modified the code however, and am now checking if the record is persisted or not. I've also detected that after I deleted a project on the admin page, the repo wasn't being deleted as I expected. This is due to the soft deletion. Updated the clean to the correct callback and fixed the spec.

fneves:
It seems that code climate hanged while analysing the code. I don't seem to have permissions to go into the link and check what is going on. Can anyone give me permissions or cancel and retry the code climate execution? Not even sure if that it possible, just want to unlock it.

steved555:
Sorry, I should have been clearer. Because you've changed the repository_url already when you call `clean_repository`, `repository_directory` is technically the one you'll create in `clone_repository`. Essentially, you'll be orphaning the directories you're changing from. You'll have to use `repository_url_was` for the one you want to clean. Unless I'm misinterpreting the meaning of `clean_repository`.

fneves:
```clean_repository``` uses the stored version version of the GitRepository before the update, still containing the old repository directory. The clean_repository acts on the GitRepository object that has not been updated yet. 
There was a bug with clean_repository that was not resetting the memoized variables of ```@repository``` and ```@repository_directory``` that would cause problems the next time that the current object was used, since the repository would not be updated.
Fixed [that](https://github.com/zendesk/samson/blob/fneves/autocomplete_tags_on_deploy/app/models/project.rb#L134), and fixed the [spec](https://github.com/zendesk/samson/blob/fneves/autocomplete_tags_on_deploy/test/models/project_test.rb#L160) to assert that the repository that is being removed is the old one.

**Comment by steved555** [comment]
Theoretically, yes, but in practice it hasn't been set yet on an update. This is an project update I made with the patch below:

```
Started PATCH "/projects/sso" for 127.0.0.1 at 2015-01-06 09:33:26 -0800
Processing by ProjectsController#update as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"...", "project"=>{"name"=>"SSO Authentication", "repository_url"=>"git@github.com:steved555/charcoal.git", "permalink"=>"sso", "release_branch"=>"master"}, "id"=>"sso"}
  User Load (0.3ms)  SELECT  `users`.* FROM `users` WHERE `users`.`deleted_at` IS NULL AND `users`.`id` = 1  ORDER BY `users`.`id` ASC LIMIT 1
  Project Load (0.3ms)  SELECT  `projects`.* FROM `projects` WHERE `projects`.`deleted_at` IS NULL AND `projects`.`permalink` = 'sso' LIMIT 1
   (0.2ms)  BEGIN
  Project Exists (0.3ms)  SELECT  1 AS one FROM `projects` WHERE `projects`.`permalink` = 'sso' AND (id <> 1) LIMIT 1
cleaning /Users/sdavidovitz/src/samson/cached_repos/a48666a4c3d3523257f5b5183587c3f4
  SQL (8.0ms)  UPDATE `projects` SET `repository_url` = 'git@github.com:steved555/charcoal.git', `updated_at` = '2015-01-06 17:33:26.814247' WHERE `projects`.`id` = 1
cloning to /Users/sdavidovitz/src/samson/cached_repos/a48666a4c3d3523257f5b5183587c3f4
   (11.5ms)  COMMIT
Redirected to http://localhost:9080/projects/sso
Completed 302 Found in 34ms (ActiveRecord: 20.5ms)
```

```diff
diff --git a/app/models/git_repository.rb b/app/models/git_repository.rb
index 2ebb1f5..9a1a0df 100644
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -21,6 +21,7 @@ class GitRepository
   end

   def clone!(executor: TerminalExecutor.new(StringIO.new), from: repository_url, to: repo_cache_dir, mirror: false)
+    Rails.logger.info("cloning to #{to}")
     return executor.execute!("git -c core.askpass=true clone --mirror #{from} #{to}") if mirror
     executor.execute!("git clone #{from} #{to}")
   end
@@ -62,6 +63,7 @@ class GitRepository
   end

   def clean!
+    Rails.logger.info("cleaning #{repo_cache_dir}")
     FileUtils.rm_rf(repo_cache_dir)
   end
```

fneves:
That problem is occurring due to the lazy initialisation of the ```@repository``` variable. On an update, the repository would be initialised with the already new ```repository_url```. 
Created a new method ```clean_old_repository``` that is now being used on the after_update. This will use the ```repository_url_was``` attribute, allowing the correct removal of the old repository directory.

app/models/project.rb
```diff
@@ -5,6 +5,9 @@ class Project < ActiveRecord::Base
 
   validates :name, :repository_url, presence: true
   before_create :generate_token
+  after_save :clone_repository, if: :repository_url_changed?
+  before_update :clean_old_repository, if: proc { persisted? && repository_url_changed? }
```
steved555: I think `before_update` does not execute on new records.

app/models/project.rb
```diff
@@ -86,16 +89,73 @@ def webhook_stages_for_branch(branch)
...
+  end
+
+  private
+
+  def log
+    Rails.logger
+  end
+
+  def clean_old_repository
+    old_repo_dir = Digest::MD5.hexdigest([repository_url_was, id].join)
```
steved555: this should probably be a method

steved555:
:+1: thanks for putting up with me

fneves:
No worries. Refactored that method, removed the ```persisted?``` check and added another spec to test that ```clean_old_repository``` never gets called upon creation.

sandlerr:
:fireworks:

Issue #177, opened 2014-10-29
AngularJS all the way?


zendesk-shender:
Now that there are more requests for dynamic updates for pages (current deploy page, current deploy badge in header, user search, job execution, ...), and since some of the pages already use AngularJS to some extent, can we start switching over to it fully?

Advantages:
- Making the samson rails app into an API focused backend will facilitate 3rd party scripting (QA in zendesk currently use some curl requests against samson APIs in their scripts).
- A more consistent UI which currently consists of Rails views, AngularJS, and direct jQuery HTML changes.
- More structured mechanism for updating content via data binding.

We don't have to do it all at once, but rather page by page, and a good starting point is the Current-Deploys page which shares 90% of the functionality of the Recent-Deploys page but one is written using Angular and the other is pure Rails.

grosser:
I think most pages do not need to be live / fancy and our feature velocity will be higher when doing simpl-stupid html instead of live angular/ember/xyz. Most the js I saw was just plain old js and the only really fancy thing we do is the progress display.

zendesk-shender:
@grosser 
Aye, but more and more feature requests are requiring JS:
- Updating current-deploys badge on menu bar
- Allowing user to hide/show deploy-group selection on stage-edit page to override.
- Environment Dashboard auto-refresh
- Pipeline progression UI updates
- ....

I think we've done a lot with the current backend HTML rendering, but to make this app shiney and responsive, I think it'll require a clientside JS framework and Radar integration.

pswadi-zendesk:
@zendesk-shender Radar should be available now.

vanchi-zendesk:
We are waiting behind a carson deploy to get all internal references out of
radar_client_rb. Should be available very soon.. :)

zendesk-shender:
Thanks @vanchi-zendesk

grosser:
For shiny feature we should use it, but what I want to avoid is making it
harder to add simple features.
Maybe some middle ground could be making static/simple things work
transparently,
but not sure if we realyy have to convert every last page or can just make
the fancy pages dynamic and the static pages static.

zendesk-shender:
In terms of code maintenance, it might start getting ugly if some of the
pages are stored as templates for clientside JS, while the rest is stored
as rails templates.

And the 2 main reasons I brought this up originally were:
- Of the clientside JS, it's diverging already into
jquery/non-jquery/angular section, so settling on a single JS framework
would help regardless.
[...]

PR #201, opened 2014-11-17
Shender/samson 107 autorefresh current deploys

zendesk-shender:
Describe your PR

This PR is primarily for discussion purposes and is in no way ready for merging. It is related to issue #177 where we voted to move more fully to angular on the frontend, and leave the rails app as more of an API.

The file structure for the javascript is the traditional AngularJS and not the newer (e.g. egghead.io) one which incorporates the templates and groups files in terms of area rather than functionality, so:
- controllers
- filters
- directives
- services

I moved the relevant views (active.html.erb, recent.html.erb, ..) over to /assets/templates where they can be retrieved on-demand by the angular engine via the angular-rails-templates gem.

To get push notifications I setup a basic Radar (https://github.com/zendesk/radar) installation, and used the Radar Client Gem (currently private in Zendesk). 
Configuration settings are stored in the .env file and accessed by both the NodeJS Radar app, and the Rails app.

I basically reorganized both the 'Recent Deploys' and 'Current Deploys' pages because they share most of their functionality.

Anyway, I will appreciate the feedback. Obviously I'll be adding tests as soon as it's clear I don't have to completely re-implement things.

/cc @zendesk/samson, @pswadi-zendesk 

### References
 - Jira link:

### Risks
 - None


zendesk-shender:
Closing, for now, angularjs is already slowly taking over, so this will be broken up into smaller pieces!

PR #210, opened 2014-12-04
use a karma helper for angular testing

steved555:
@princemaple what do you think about this? I was playing around in a personal project and got this library to help load assets from gem paths.


jwswj:
@steved555 ?

steved555:
Closing, if this is useful for any angular work feel free to pull it in.

PR #223, opened 2015-01-09
Send an email when a new project is created

dragonfax:
Needed for security compliance.
https://zendesk.atlassian.net/browse/INFR-823

The email sent is rather simple. It should come from the same address as all other deploy emails.
```
Subject: Samson Project Created: Lotus

Jason Stillwell (jstillwell@zendesk.com) just created a new project Lotus
```

@zendesk/samson 
cc @pswadi-zendesk @lcore

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -33,6 +33,9 @@ class ProjectsController < ApplicationController
     @project = Project.new(project_params)
 
     if @project.save
+      if ENV['PROJECT_CREATED_NOTIFY_ADDRESS']
+        ProjectMailer.created_email(@current_user,@project).deliver_later
+      end
       redirect_to project_path(@project)
       Rails.logger.info("#{@current_user.name_and_email} created a new project #{@project.to_param}")
     else
--- a/app/mailers/project_mailer.rb
+++ b/app/mailers/project_mailer.rb
@@ -1,6 +1,4 @@
 class ProjectMailer < ApplicationMailer
-  layout 'mailer'
-
   def created_email(user, project)
     address = ENV['PROJECT_CREATED_NOTIFY_ADDRESS']
     subject = "Samson Project Created: #{project.name}"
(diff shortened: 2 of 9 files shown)


app/controllers/projects_controller.rb
```diff
@@ -33,6 +33,9 @@ def create
     @project = Project.new(project_params)
 
     if @project.save
+      if ENV['PROJECT_CREATED_NOTIFY_ADDRESS']
+        ProjectMailer.created_email(@current_user,@project).deliver_later
```
steved555: We don't actually have a queue so this essentially runs inline anyway

dragonfax: Yeah, I was just thinking about the future. Even the rails docs say use
_later unless you NEED to have it always delivered immediately.

jwswj:
Can you add an example in https://github.com/zendesk/samson/blob/master/.env.example?

app/mailers/project_mailer.rb
```diff
@@ -0,0 +1,7 @@
+class ProjectMailer < ApplicationMailer
+  layout 'mailer'
+
+  def created_email(user, project)
+    mail(to: ENV['PROJECT_CREATED_NOTIFY_ADDRESS'], subject: "Samson Project Created: #{project.name}", body: "#{user.name_and_email} just created a new project #{project.name}")
```
jwswj: long line. individual variables?

dragonfax: Yes, much more readable now. Thanks.

app/mailers/project_mailer.rb
```diff
@@ -0,0 +1,10 @@
+class ProjectMailer < ApplicationMailer
+  layout 'mailer'
```
steved555: Where does this layout come from? Is this a default?

dragonfax: Ah, I thought I had removed everything from `rails generate` but I had missed that line.

test/test_helper.rb
```diff
@@ -1,5 +1,7 @@
 ENV["RAILS_ENV"] ||= "test"
 
+ENV['PROJECT_CREATED_NOTIFY_ADDRESS'] = 'blah@example.com'
```
steved555: I'd rather see this set in .travis.yml / the tests themselves

steved555:
+1

PR #227, opened 2015-01-15
nicer dashes in stage name badges

sandlerr:
Stages with dashes in them don't look great. The badge service renders `--` as a dash but `&mdash;` literally.

cc @princemaple @pswadi-zendesk @grosser

### DIFF
--- a/app/controllers/stages_controller.rb
+++ b/app/controllers/stages_controller.rb
@@ -108,9 +108,7 @@ class StagesController < ApplicationController
   def badge_safe(string)
     CGI.escape(string.gsub('&', '&amp;'))
       .gsub('+','%20')
-      .gsub /([^-])(--)*-([^-])/ do |r|
-        "#{$1}--#{$3}"
-      end
+      .gsub(/-+/,'--')
   end
 
   def check_token
--- a/app/controllers/stages_controller.rb
+++ b/app/controllers/stages_controller.rb
@@ -106,7 +106,7 @@ class StagesController < ApplicationController
   private
 
   def badge_safe(string)
-    CGI.escape(string.gsub('&', '&amp;'))
+    CGI.escape(string.gsub('&', '%26'))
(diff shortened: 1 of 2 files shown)


app/controllers/stages_controller.rb
```diff
@@ -106,7 +106,11 @@ def clone
   private
 
   def badge_safe(string)
-    CGI.escape(string.gsub('&', '&amp;')).gsub('-', '&mdash;').gsub('+','%20')
+    CGI.escape(string.gsub('&', '&amp;'))
+      .gsub('+','%20')
+      .gsub /([^-])(--)*-([^-])/ do |r|
```
princemaple: /(?:--)*-?/ -> '--'
Does this do what you want?

grosser: how about `gsub(/\-{2,}/, '--') ?

grosser:
:+1: a before/after pic would be nice :)

app/controllers/stages_controller.rb
sandlerr: http://rubular.com/r/mQqDnMGSk4

Not sure - Rubular seems to think it will insert `--` between every char?

grosser: http://rubular.com/r/wqPccvUfM2

grosser: it also replaces -- with itself so that should be good ...

princemaple:
:+1: when the regex is settled.

sandlerr:
Screenshots ripped directly from the test:
Before: before
After: after

princemaple:
lol &amp; 

:+1:

grosser:
:+1: 
maybe gsub & 'and'

sandlerr:
@grosser &26 works for &

grosser:
:heart:

PR #228, opened 2015-01-16
stop samson trying to open a URI as a path

sandlerr:
@princemaple @grosser these are broken at the moment


grosser:
maybe into the top somewhere ?
how did that work before oO
:+1: anyway ...

Issue #4, opened 2014-05-29
Better approach to managing integrations


jwswj:
The number of integrations with Samson could grow quite quickly, however not everyone wants them all to be visible when configuring stages.

It would be nice if super admins could have a place where they could toggle the integrations they would like for their version of Samson. This probably belongs as a new page in the admin drop down "Configuration"...

Thoughts @dasch @steved555 @tonymilne?

sandlerr:
https://github.com/zendesk/samson/pull/259

PR #229, opened 2015-01-22
Enable Configurable Flowdock Buddy Requests 

fneves:
# Description

This PR will allow you to send a buddy request to the configured flows that allow you to have buddy request notifications.

You will then be able to send a custom message to previously configured flows.

# Screenshots:

## Configuring the stage flows to enable flow notifications:

flowdock_configurable

## Sending a buddy request to flowdock:

waiting_for_buddy

## Successful notification:

successful_buddy_request

### References
 - Jira link: [SAMSON-106](https://zendesk.atlassian.net/browse/SAMSON-106)

### Risks
 - None


app/assets/javascripts/directives/buddy_request_box.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
```
sandlerr: This won't do anything, right? `"use strict";` would need to be right at the top of `application.js` and would then affect the entire application.

fneves: I think you are right. I was using it for my IDE only. I should probably remove it.

app/assets/javascripts/directives/buddy_request_box.js
```diff
@@ -0,0 +1,101 @@
...
+  self.buddyRequest = function (deploy, message) {
+    var promise = $http.post('/integrations/flowdock/notify', { deploy_id: deploy, message: message });
+    promise.success(function (data) {
+        $('#buddyRequestInfoBox').html('<div class="alert alert-success"><a class="close" data-dismiss="alert">×</a><span>'+ data.message + '</span></div>');
+        console.log(data.message);
+    });
+    promise.error(function () {
+      $('#buddyRequestInfoBox').html('<div class="alert alert-danger"><a class="close" data-dismiss="alert">×</a>' +
+      '<span>Error! Could not send buddy request!</span></div>');
+      console.log('Could not notify flows!');
```
sandlerr: can probably remove this and the log statement above

app/assets/javascripts/directives/buddy_request_box.js
```diff
@@ -0,0 +1,101 @@
...
+      console.log('Could not fetch the flowdock users!');
+    });
+
+    return users;
+  };
+
+  self.buddyRequest = function (deploy, message) {
+    var promise = $http.post('/integrations/flowdock/notify', { deploy_id: deploy, message: message });
+    promise.success(function (data) {
+        $('#buddyRequestInfoBox').html('<div class="alert alert-success"><a class="close" data-dismiss="alert">×</a><span>'+ data.message + '</span></div>');
```
princemaple: doing jQuery dom manipulation in angular already suggests that something is wrong...

app/assets/javascripts/directives/buddy_request_box.js
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  return {
+    users: self.users,
+    buddyRequest: self.buddyRequest
+  }
+}]);
+
+samson.controller('BuddyNotificationsCtrl', ['$scope','$rootScope', '$injector', 'Flowdock',
+  function($scope, $rootScope, $injector, flowdock) {
```
princemaple: you are not using the `$injector`

app/assets/templates/directives/buddy_request_box.html
```diff
@@ -0,0 +1,17 @@
+<div id="buddyRequestInfoBox"></div>
+<div class="panel panel-info" ng-controller="BuddyNotificationsCtrl as buddyCtrl">
```
princemaple: apply the controller directly to the directive itself, then you don't have to do this, and you don't have to do dom manipulation with jQuery.

app/assets/javascripts/directives/buddy_request_box.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+var MentionsBox = function (users) {
```
princemaple: this should be a factory (or service, if you like), and be injected.

app/assets/templates/directives/buddy_request_box.html
princemaple: I meant you could define the controller in the directive definition, not moving the `ng-controller` to the directive

princemaple:
Actually, we don't think it's a good idea to send buddy approval request to anyone. The idea is, when you deploy stuff, you should already have a buddy. Is that right, @jwswj ?

I'll cc for you, @zendesk/samson

jwswj:
:-1: I just don't like this feature. It's nothing to do with the code. Just that I believe you should have a buddy before you get to this point in the deploy process. They should know what you are deploying and should be able to fix it if anything goes wrong in the future. Doing this is like Github putting a button on each pull request which would annoy people via Flowdock when you wanted someone to review your PR. The better solution here would be to just provide the URL and a button you could click to copy the URL for the deploy, then you can get that to you buddy using whatever chat mechanism you are currently using. Also, this feature is very Flowdock specific. What about all the companies who are using Slack, GChat or the thousands of other messaging applications?

rdhanoa:
@fneves @pswadi-zendesk  About flowdock notifications feature, I learnt that this is available only for production stages. We in QA can really leverage this feature to monitor deploys if
- this feature is added to staging deploys.
- with an ability to send notification only for failed deploys.

princemaple:
@rdhanoa do you mean deploy notifications?
They are configured per stage.

pswadi-zendesk:
@princemaple : About local deploy buddy: deploy buddy could be remote, as in WFH or in another office.  Idea is, as long as someone is available to keep an eye on deploy along with the deployer, it is not required to be local.
@jwswj : About why flowdock notificaiton and not others, like Slack or GChat, because we already have an existing integration with flowdock. This is just an addition to that. If we add ability to integrate with other chat messengers, the this code should be improved to lookup identities across these messengers.

Part of the "buddy approval" is for audit purpose. Kind of like trust but verify. You are right, there needs to be a buddy when a deploy happens, irrespective of buddy approval system. Right now folks pick this URL and paste it in flowdock asking for users to click on it to approve.  We are just making this process easy for them.

app/assets/javascripts/directives/buddy_request_box.js
jwswj: +1

jwswj:
My point around flowdock was we should be making it easier for other integrations to use this type of functionality, rather than just making it generic for flowdock. For a long time (https://github.com/zendesk/samson/issues/4) we've wanted an easier way to integrate with other platforms, coding like this just makes that further from ever being possible which decreases the chances of us building a really strong community around the product, which decreases the amount of effort we've put into open sourcing it.

My suggestion:
- Make the sharing of the URL a lot easier by building great "Copy" functionality at this point in the deploy process. Let people manually open their chat tool of choice and paste the URL. This keeps the implementation really simple, without adding all this custom flowdock bloat and doesn't really cause the developer much more pain at all.

fneves:
@jwswj I agree with your point that we should make it easier for integrations with 3rd party software such as Flowdock, Hipchat, Mailchimp...
We would benefit greatly if we could develop a plugin architecture that would easily allow us to add/remove plugins on the fly. 
As a user I would really like the ability to browse a set of plugins and with a single click fetch and install the integration with my desired application, and use it straight away.

The integration with Flowdock and Github are, at the moment, the best candidates to be extracted into a plugin. 

However, at the moment we are supporting Flowdock integration, and this feature seems like it could help in some workflows of some teams, as @pswadi-zendesk referred. And, for that reason, I believe it adds value to the product. Later, it can be extracted when we have a plugin infrastructure in place.
If we can deliver this plugin infrastructure I strongly believe that the community around Samson is going to grow a lot faster.

jwswj:
> Later, it can be extracted when we have a plugin infrastructure in place.
This basically never happens.

So you don't like my suggestion of making it a little more generic for now?
Is there a plan for building a better plugin infrastructure?

jwswj:
Implementation would look something like:
samson-copy-func

fneves:
@jwswj The reason I don't like it is because the browser already gives us the URL that we can copy. Isn't that the current workflow that everybody is already doing? It would make sense if the URL was something different. 
Github has that feature on the clone URL. But the clone URL that it allows you to select can be different than the one you have on the browser, such as a SSH specific URL. 
On this case, as it is the same, I am not sure if gives us any improvement on the current workflow.

I don't mind starting to design a plugin architecture and compromise myself to extract this to a plugin when that architecture is in place. But designing this kind of architecture/framework/SDK is going to be a lot of work and it won't happen anytime soon, specially if I'll be the only one involved. 
I will open an issue and start the discussion and hopefully we can have more people involved in this.

fneves:
@pswadi-zendesk @jwswj I have started a discussion about a possible plugin architecture here [#234](https://github.com/zendesk/samson/issues/234)

Please let me know your thoughts.

fneves:
@pswadi-zendesk @princemaple @jwswj @sandlerr I've fixed those angular issues. Can you have another look?

Regarding the extraction to a plugin + a plugin architecture, [here](https://github.com/zendesk/samson_sdk/pull/2) and [here](https://github.com/zendesk/samson/compare/fneves/plugins_architecture) you can see a prototype for what I am doing to enable plugin development. I am using this feature as the example plugin for the sdk.

PR #230, opened 2015-01-22
Removing chdir that can cause problems with Puma multithreaded model

fneves:
This PR removes the previously introduced chdir and replaces them with a cd #{pwd} that avoids changing the process working directory.

A chdir changes the process directory not the thread working directory. This causes problems whenever this is running on a multithreaded environment.

### References
 - Jira link: 

### Risks
 - None
Describe your PR

/cc @zendesk/other

### References
 - Jira link:

### Risks
 - None


pswadi-zendesk:
:+1:

PR #231, opened 2015-01-23
Fixing the tag list command

fneves:
# Description
The previous command would list all shas associated with a tag. This was not efficient as it was iterating through all the list of shas to get its description. 
In some cases describe was failing because it could not find the commit. I am assuming this might happen because the history might have been rewritten.

This was having an impact in the autocomplete feature for some projects, causing the tags not to be displayed.

This command was fixed so that it now works for all projects even if history was modified.

The --sort command is only available in git 2.0+ so Travis git version had to be updated, and container based builds to be disabled on the travis file.

/cc @zendesk/samson @pswadi-zendesk @steved555 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -54,7 +54,7 @@ class GitRepository
   end
 
   def tags
-    cmd = 'git describe --tags --abbrev=0 `git rev-list --tags --max-count=600`'
+    cmd = "git for-each-ref refs/tags --sort=-authordate --format='%(refname)' --count=600 | sed 's/refs\\/tags\\///g'"
     success, output = run_single_command(cmd) { |line| line.strip }
     success ? output : []
   end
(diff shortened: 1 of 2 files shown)


steved555:
+1

grosser:
@fneves want to rebase and merge that ?

fneves:
This will require git 2.x and just noticed that Ubuntu 12, 14 do not support git 2.x on their official package repos.

Maybe its better fallback to a simple  ```git tag``` and do the version sorting and filtering in ruby? 
It won't be the exact same thing as I won't be able to order by the time when  tag was created.

@grosser @steved555 What do you think?

pswadi-zendesk:
Yeah, unfortunately Ubuntu has not updated to git 2.x officially and hence Ops is wary of trying out PPP packages. Lets go back to filtering in ruby.  :cry:

fneves:
@pswadi-zendesk @steved Updated the command with one that does the same but compatible with git v1.7.9.5

jonmoter:
:+1:

pswadi-zendesk:
:+1:  dizzy with back and forward slashes.. ;-)

PR #232, opened 2015-01-27
Add validation for the repository URL during project creation/editing.

zendesk-shender:
Describe your PR
Seemed there was an increase of people putting invalid github URLs in their project settings. So I added a validation function for it to give more immediate feedback about their mistake.

samson

/cc @zendesk/samson @pswadi-zendesk  @steved555 @princemaple 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -70,7 +70,7 @@ class GitRepository
   end
 
   def valid_url?
-    cmd = "GIT_ASKPASS='true' git ls-remote -h #{repository_url}"
+    cmd = "git -c core.askpass=true ls-remote -h #{repository_url}"
     output = StringIO.new
     executor = TerminalExecutor.new(output)
     return true if executor.execute!(cmd)
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -69,6 +69,15 @@ class GitRepository
     Dir.exist?("#{repo_dir}/.git") || File.exist?("#{repo_dir}/HEAD")
   end
 
+  def valid_url?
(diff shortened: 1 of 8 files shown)


sandlerr:
You can shell execute a command more simply like this, right? https://github.com/zendesk/samson/blob/3a3a45199a5b7ec0fa2c126498da904be0a4195b/config/initializers/revision.rb#L6

Was there a reason you didn't implement this using octokit, since we are already quite tied to Github?

zendesk-shender:
@sandlerr , I didn't want to completely tie us to github as we theoretically support any git repo, whether served locally, or remotely besides on github. I thought we specialized the code only for PRs and Changeset support? Besides authentication obviously. :)

I also didn't use the backticks, as within that same class someone encapsulated the PTY.spawn method and I figured I'd keep it consistent.

zendesk-shender:
Added a change to address @fneves offline comment about re-using the existing run_single_command function.

sandlerr:
Isn't `core.askpass` is supposed to be set to the path of an executable? Are you setting it just to disable the interactive prompt?

zendesk-shender:
@sandlerr , yes core.askpass is supposed to be a command, and 'true' is a builtin shell function that just returns an error code of 0.

I asked this question before when someone else used it this way as well! :)

davecocoa:
:clap: Thank you! This much effort to accommodate my mistakes is greatly appreciated!

zendesk-shender:
Don't worry @davecocoa , you weren't the only one! :)

sandlerr:
+1 looks good! cc @zendesk/samson

zendesk-shender:
merged!

PR #235, opened 2015-02-05
This might have better wording

apanzerj:
I'm not happy with the wording yet but this is the best I could come up with right now. 

Thoughts?


grosser:
:+1:

PR #237, opened 2015-02-10
Fix job log style

princemaple:
Style fix.

**Before**
p2_deploy__failed_

**after**
p2_deploy__failed_

also, the real diff
1__zsh

/cc @zendesk/samson


adammw:
:+1: when :green_heart:

PR #238, opened 2015-02-11
Style change, fix #153

### LINKED ISSUE #153: Buttons get messed up in a narrower browser window.
image

from: @halcyonCorsair

princemaple:
@pswadi-zendesk

### DIFF
--- a/app/views/deploys/_output.html.erb
+++ b/app/views/deploys/_output.html.erb
@@ -1,6 +1,6 @@
 <div class="output <%= 'hidden' if @deploy.waiting_for_buddy? || @deploy.pending_non_production? %>">
-  <div class="row">
-    <div class="btn-toolbar col-md-4 col-md-offset-7 pull-right" id="output-toolbar">
+  <div class="clearfix">
+    <div class="btn-toolbar pull-right" id="output-toolbar">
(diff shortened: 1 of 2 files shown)


princemaple:
cc @zendesk/samson

princemaple:
@sandlerr

sandlerr:
Does this move the buttons slightly to the left?

princemaple:
Nope, it has `pull-right`, not "moved". It will no longer limit the width of the container of the buttons, so the buttons can be on the same line. Effectively yes, though.

princemaple:
#153 imagine this works as expected.

sandlerr:
Just tried this in dev tools. It does move the buttons to the right which is what I originally meant. They don't line up with the edge of the terminal output.

princemaple:
OK will check again later.

princemaple:
@sandlerr Fixed

first_deploy__running__-_p1

sandlerr:
so the overhang was caused by the negative margins on `row`? They're often quite inconvenient. :+1:

princemaple:
Yeah you have to go full bootstrap to avoid any possible trouble.

PR #239, opened 2015-02-13
fix deploy link

grosser:
atm deploys point to old project urls since they are cached since 2+months

@pswadi-zendesk @steved555 

### Risks
 - Low: more resource usage due to cache expiration


steved555:
+1

PR #240, opened 2015-02-17
Filter deploy list on user and project names

stubotnik:
Fixed search filter on /deploys/recent. 

Was only filtering on top-level attributes of deploy object, which does not include the project or user name as displayed on the page. 

/cc @pswadi-zendesk, @steved555, @princemaple, @zendesk-shender   

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-137

### Risks
 - None

### DIFF
--- a/app/assets/javascripts/timeline.js
+++ b/app/assets/javascripts/timeline.js
@@ -68,10 +68,12 @@ samson.constant("StatusFilterMapping",
 samson.filter("projectUserFilter",
   function() {
     return function(deploys, search) {
-      if (search !== undefined && search !== null && search.length) {
-        var searchRegExp = new RegExp(search, 'i');
+      if (typeof search == 'string' && search.length) {
+        var lowerCaseSearch =  search.toLowerCase();
+
         return deploys.filter(function(deploy) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/timeline.js
```diff
@@ -65,6 +65,20 @@ samson.constant("StatusFilterMapping",
   }
 );
 
+samson.filter("projectUserFilter",
+  function() {
+    return function(deploys, search) {
+      if (search !== undefined && search !== null && search.length) {
```
sandlerr: I guess you could do `if (typeof search == 'string' && search.length) {`

sandlerr:
Would it be worth making it obvious this is actually a regex or escaping regex special characters?

princemaple:
:+1:

Feels like it's about time to extract all the filtering logic out of the view. Wish I have more time.

stubotnik:
@sandlerr Now that you mention it I don't think there is any need for this to be a regex. Have switched to lower-cased string compares

stubotnik:
requested edits pushed. +1?

cc @sandlerr

jwswj:
:+1:

sandlerr:
:+1: looks good!

PR #242, opened 2015-02-21
fix unable to run by line number when first test is generated by a macro

grosser:
https://github.com/judofyr/minitest-line/pull/11

now this works: `mtest test/controllers/stages_controller_test.rb:19`

@jonmoter


princemaple:
:+1:

PR #243, opened 2015-02-22
fix mtest running multiple tests with the exact same name

grosser:
before `mtest test/controllers/stages_controller_test.rb:19` would run all tests that are names `it "renders"` but we only want one of them :)

https://github.com/judofyr/minitest-line/pull/12

@princemaple 

### Risks
 - None


princemaple:
:+1: 
Do you think it's a good idea to make Samson builds only run for the PR build too?

PR #245, opened 2015-02-22
fix incorrect method call

sandlerr:
cc @princemaple @grosser @steved555 

airbrake: https://zendesk.airbrake.io/projects/95346/groups/1335122691522767311

actual method: https://github.com/zendesk/samson/blob/master/app/models/zendesk_notification.rb#L12

:+1:

### DIFF
--- a/test/models/deploy_service_test.rb
+++ b/test/models/deploy_service_test.rb
@@ -160,6 +160,15 @@ class DeployServiceTest < ActiveSupport::TestCase
       job_execution.send(:run!)
     end
 
+    it "sends Zendesk notifications if the stage has them enabled" do
+      stage.stubs(:comment_on_zendesk_tickets?).returns(true)
+
+      ZendeskNotification.any_instance.expects(:deliver)
+
+      service.deploy!(stage, reference)
+      job_execution.send(:run!)
+    end
+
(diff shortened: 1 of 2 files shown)


princemaple:
:+1:

grosser:
test ?

sandlerr:
@grosser done

test/models/deploy_service_test.rb
```diff
@@ -160,6 +160,15 @@ class DeployServiceTest < ActiveSupport::TestCase
       job_execution.send(:run!)
     end
 
+    it "sends Zendesk notifications if the stage has them enabled" do
+      stage.stubs(:comment_on_zendesk_tickets?).returns(true)
+
+      ZendeskNotification.any_instance.expects(:deliver)
```
grosser: hmm that test would have passed before with `:deliver_now`, any other sideeffect we could spec on ?

grosser:
:+1:

test/models/deploy_service_test.rb
sandlerr: we could assert that it responds_to deliver, but if someone renames the method again it will break [this](https://github.com/zendesk/samson/blob/master/test/models/zendesk_notification_test.rb#L20) test

PR #248, opened 2015-02-24
dry up user

grosser:
@pswadi-zendesk @steved555 @jonmoter 

### Risks
 - None


jonmoter:
I notice this adds `integration: true` as a parameter to `User.create_with`, which wasn't there before. What effect does that have?

grosser:
hehe, yes, copy paste error :)

grosser:
fixed, +1 ?

jonmoter:
:+1:

PR #250, opened 2015-02-25
send emails on failed deploys

grosser:
@pswadi-zendesk @jonmoter @steved555 @princemaple @sandlerr 

screen shot 2015-02-24 at 5 53 59 pm

### DIFF
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -70,15 +70,23 @@ class DeployService
   end
 
   def send_after_notifications(stage, deploy)
-    if stage.send_email_notifications?
-      DeployMailer.deploy_email(stage, deploy).deliver_now
-    end
-
+    send_deploy_email(stage, deploy)
+    send_failed_deploy_email(stage, deploy)
     send_flowdock_notification(stage, deploy)
     send_datadog_notification(stage, deploy)
     send_github_notification(stage, deploy)
     send_zendesk_notification(stage, deploy)
     update_github_deployment_status(stage, deploy)
+  end
+
+  def send_deploy_email(stage, deploy)
+    return unless stage.send_email_notifications?
+    DeployMailer.deploy_email(stage, deploy).deliver_now
+  end
+
+  def send_failed_deploy_email(stage, deploy)
+    return unless emails = stage.automated_failure_emails(deploy)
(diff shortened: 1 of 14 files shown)


grosser:
bump good to go, had to do some nasty hacks in the schema + added tests for them since I was running into mysql errors https://travis-ci.org/zendesk/samson/jobs/52194828

app/models/deploy_service.rb
```diff
@@ -78,9 +78,15 @@ def send_after_notifications(stage, deploy)
     send_datadog_notification(stage, deploy)
     send_github_notification(stage, deploy)
     send_zendesk_notification(stage, deploy)
+    send_failed_deploy_notification(stage, deploy)
```
sandlerr: `send_failed_deploy_email` ?

app/models/stage.rb
```diff
@@ -154,6 +154,27 @@ def send_github_notifications?
...
+    return if !email_committers_on_automated_deploy_failure? && static_emails_on_automated_deploy_failure.blank?
+    return unless deploy.failed?
+    return unless deploy.user.integration?
+    last_deploy = deploys.finished_naturally.prior_to(deploy).first
+    return if last_deploy.try(:failed?)
+
+    emails = []
+
+    # static notification
+    emails.concat static_emails_on_automated_deploy_failure.to_s.split(/, ?/)
```
sandlerr: wouldn't the regex be `/[, ?/` ?

app/models/stage.rb
```diff
@@ -154,6 +154,27 @@ def send_github_notifications?
...
+
+    # authors of commits after last successful deploy
+    if email_committers_on_automated_deploy_failure?
+      changeset = deploy.changeset_to(last_deploy)
+      emails.concat changeset.commits.map(&:author_email).compact
+    end
+
+    emails.presence
```
sandlerr: is it worth putting `emails.uniq.presence`?

sandlerr:
:+1: assuming the non-nullable columns will work with existing records

app/models/stage.rb
grosser: I don't think emails are getting sent in duplicates, but also no harm in doing it :)

PR #251, opened 2015-02-26
Shender/environment stage grouping

zendesk-shender:
This PR is for the Runway Project that aims to have Samson become a bit more host-aware, so any stages that target the same hosts can be identified and dashboards can be created to show what's installed on each target system.
We have chosen to use 'Deploy Group' to denote the target deployment hosts, as we in Zendesk use whole datacenters as our target rather than deploying to individual machines.

Environments - equivalent to 'production', 'staging', 'testing' labels for deployment flows
DeployGroups - assigned an environment and correspond to a group of hosts that a deploy will target.
Stage - A custom deploy config for a project that can target multiple DeployGroups.

This PR lays the groundwork for for the dashboard views, a later PR will implement them.

To enable the below UIs you have to set the 'DEPLOY_GROUP_FEATURE' flag in the .env file.

envs
deploygroups
dgedit
envedit
stageedit
stageeditmulti

/cc @samson @pswadi-zendesk @sandlerr @grosser @gabetax @jonmoter

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-5 https://zendesk.atlassian.net/browse/RUN-26

### Risks
 - None


db/schema.rb
```diff
@@ -11,16 +11,33 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150109005441) do
+ActiveRecord::Schema.define(version: 20150223135916) do
 
-  create_table "commands", force: true do |t|
-    t.text     "command",    limit: 16777215
+  create_table "commands", force: :cascade do |t|
+    t.text     "command",    limit: 10485760
```
zendesk-shender: I'm guessing the recent upgrade of rails caused the activerecord schema generation to change the :force flag?
I don't know why the limit was the old value because it's very clear that it's supposed to be 10MB from the migration files.

db/schema.rb
```diff
@@ -31,11 +48,18 @@
     t.datetime "deleted_at"
   end
 
-  add_index "deploys", ["deleted_at"], name: "index_deploys_on_deleted_at", using: :btree
-  add_index "deploys", ["job_id", "deleted_at"], name: "index_deploys_on_job_id_and_deleted_at", using: :btree
-  add_index "deploys", ["stage_id", "deleted_at"], name: "index_deploys_on_stage_id_and_deleted_at", using: :btree
```
zendesk-shender: If we want the btree index back, I'll have to add another migration to recreate the indexes directive specifying the btree option:
add_index :deploys, :stage_id, using: 'btree')

grosser:
no too worried about the schema changes, maybe best not to commit them and make a clean "dump new schema" PR

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,53 @@
+class Admin::DeployGroupsController < ApplicationController
+  before_action :authorize_admin!
+  before_action :authorize_super_admin!, only: [ :create, :new, :update, :destroy ]
+
+  rescue_from ActiveRecord::RecordNotFound do |error|
```
grosser: do we really need this ? <-> just 404 ?

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def new
+    @deploy_group = DeployGroup.new
+  end
+
+  def create
+    deploy_group = DeployGroup.create!(deploy_group_params)
+    Rails.logger.info("#{current_user.name_and_email} created the DeployGroup #{deploy_group.name}")
+    redirect_to action: 'index'
+  rescue => ex
```
grosser: that is normally

```Ruby
@deploy_group = DeployGroup.create!(deploy_group_params)
if @deploy_group
  redirect
else
  flash + render 'new'
end
```

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    flash[:error] = "Failed to create DeployGroup: #{ex.message}"
+    redirect_to :back
+  end
+
+  def update
+    deploy_group = DeployGroup.find(params[:id])
+    deploy_group.update_attributes!(deploy_group_params)
+    Rails.logger.info("#{current_user.name_and_email} changed the DeployGroup #{deploy_group.name}")
+    redirect_to action: 'index'
+  rescue => ex
```
grosser: same here, `if deploy_group.update_attributes(deploy_group_params)` -> redirect else render

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    flash[:error] = "Failed to update DeployGroup: #{ex.message}"
+    redirect_to :back
+  end
+
+  def destroy
+    deploy_group = DeployGroup.find(params[:id])
```
grosser:
```Ruby
def deploy_group
  @deploy_group ||= DeployGroup.find(params[:id])
end
```

and maybe `before_action :deploy_group, only: [:show, :edit, :update, :destroy]`

app/controllers/admin/environments_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    redirect_to action: 'index'
+  rescue => ex
+    flash[:error] = "Failed to create environment: #{ex.message}"
+    redirect_to :back
+  end
+
+  def update
+    env = Environment.find(params[:id])
+    env.update_attributes!(env_params)
+    Rails.logger.info("#{current_user.name_and_email} changed the environment #{env.name}")
```
grosser: do we want/need all this logging, user and request is already in the logs, so we could search for `EnvironmentsController#update` etc to get what was done

app/controllers/stages_controller.rb
```diff
@@ -145,4 +145,11 @@ def find_project
   def find_stage
     @stage = @project.stages.find_by_param!(params[:id])
   end
+
+  def update_deploy_groups
+    # http://stackoverflow.com/questions/8929230/why-is-the-first-element-always-blank-in-my-rails-multi-select-using-an-embedde
+    deploy_group_ids = params[:stage].fetch(:deploy_groups, []).select { |id| !id.empty? }
```
grosser: `.reject(&:empty?)`

grosser: can this whole action should just be a call to update ?

app/models/stage.rb
```diff
@@ -154,6 +156,15 @@ def send_github_notifications?
     update_github_pull_requests
   end
 
+  def production?
+    if ENV['DEPLOY_GROUP_FEATURE']
```
grosser:
```Ruby
if ENV['DEPLOY_GROUP_FEATURE']
  deploy_groups.detect { |deploy_group| deploy_group.environment.is_production? }
else
  super
end
```

app/views/stages/_fields.html.erb
```diff
@@ -23,23 +23,42 @@
...
-        <%= form.label :confirm do %>
-          <%= form.check_box :confirm %>
-          Confirm before deployment
-        <% end %>
+  <% if ENV['DEPLOY_GROUP_FEATURE'] %>
+    <div class="form-group">
+      <%= form.label :name, 'Deploy Groups', class: 'col-lg-2 control-label' %>
+      <div class="col-lg-4">
+        <%= form.collection_select(:deploy_groups, DeployGroup.all, :id, :name,
+                                   {
```
grosser: please use 2-space indent

db/migrate/20150223135916_create_deploy_groups.rb
```diff
@@ -0,0 +1,23 @@
+class CreateDeployGroups < ActiveRecord::Migration
+  def change
+    create_table :environments do |t|
+      t.string :name, null: false
+      t.boolean :is_production, default: false
```
grosser: `null: false` or it can be set to nil

db/migrate/20150223135916_create_deploy_groups.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      t.timestamps null: false
```
grosser: I though null false was the default for timestamps anyway

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,48 @@
+require_relative '../../test_helper'
+
+describe Admin::DeployGroupsController do
+  as_a_admin do
+    it 'get index as admin succeeds' do
+      get :index
+      response.success?.must_equal true
```
grosser: `assert_response :success`

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,48 @@
...
+      assigns(:deploy_groups).count.must_equal 3
```
grosser: this will break if someone adds new fixtures -> `DeployGroup.count`

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    unauthorized :post, :create
+    unauthorized :delete, :destroy, id: 1
+    unauthorized :post, :update, id: 1
+  end
+
+  as_a_deployer do
```
grosser: deployer should come above admin

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,48 @@
...
+    unauthorized :get, :index
+    unauthorized :post, :create
+    unauthorized :delete, :destroy, id: 1
+    unauthorized :post, :update, id: 1
+  end
+
+  as_a_super_admin do
+    before { request.env["HTTP_REFERER"] = admin_deploy_groups_url }
+
+    it 'should create an environment' do
```
grosser: avoiding `should` makes for easier to read its -> `it 'creates an environment'`

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,48 @@
...
+      post :create, deploy_group: {name: 'pod666', environment_id: environments(:staging_env).id}
+      assert_redirected_to admin_deploy_groups_path
+      deploy_groups = DeployGroup.where(name: 'pod666')
+      deploy_groups.count.must_equal 1
+      deploy_groups.first.environment.name.must_equal 'Staging'
+    end
+
+    it 'should delete an environment' do
+      id = deploy_groups(:deploy_group_pod100).id
```
grosser: `let(:deploy_group) { deploy_groups(:deploy_group_pod100) }` and reuse in edit + destroy

test/controllers/admin/deploy_groups_controller_test.rb
grosser: also this is deploy groups and not environment

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,48 @@
+require_relative '../../test_helper'
+
+describe Admin::DeployGroupsController do
```
grosser: tests for new/update ?

grosser: tests for create fails/update fails

test/controllers/admin/environments_controller_test.rb
```diff
@@ -0,0 +1,44 @@
...
+      response.success?.must_equal true
+      assigns(:environments).count.must_equal 2
+    end
+
+    unauthorized :post, :create
+    unauthorized :delete, :destroy, id: 1
+    unauthorized :post, :update, id: 1
+  end
+
+  as_a_deployer do
```
grosser: deployer above admin (least amount of rights to highest amount of rights)

test/controllers/admin/environments_controller_test.rb
```diff
@@ -0,0 +1,44 @@
+require_relative '../../test_helper'
+
+describe Admin::EnvironmentsController do
+  as_a_admin do
+    it 'get index as admin succeeds' do
+      get :index
+      response.success?.must_equal true
```
grosser: `assert_response :success`

test/controllers/admin/environments_controller_test.rb
```diff
@@ -0,0 +1,44 @@
+require_relative '../../test_helper'
+
+describe Admin::EnvironmentsController do
+  as_a_admin do
```
grosser: tests for update + new + update-fails / create fails

test/fixtures/deploy_groups.yml
```diff
@@ -0,0 +1,13 @@
+# Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
```
grosser: If I want to learn about fixtures I can use google -> remove it ;)

test/models/deploy_group_test.rb
```diff
@@ -0,0 +1,32 @@
+require_relative '../test_helper'
+
+describe DeployGroup do
+  let(:prod_env) { environments(:production_env) }
+
+  it 'should be saved with valid name/env' do
```
grosser: you can make it simpler to understand by grouping them to their method
`describe "#new" do` / `describe "#create" do`

test/models/deploy_group_test.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  it 'should require a name' do
+    deploy_group = DeployGroup.new(name: nil, environment: prod_env)
+    deploy_group.valid?.must_equal false
+  end
+
+  it 'should require an environment' do
+    deploy_group = DeployGroup.new(name: 'foo')
+    deploy_group.valid?.must_equal false
```
grosser: for easy debugging:

```Ruby
def assert_valid(record)
  assert record.valid?, record.errors.full_messages
end

def refute_valid(record)
  ...
end
```

test/models/deploy_group_test.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  it 'should require a unique name' do
+    deploy_group = DeployGroup.new(name: 'Pod1', environment: prod_env)
+    deploy_group.valid?.must_equal false
+  end
+
+  it 'should be queried by environment' do
+    DeployGroup.create!(name: 'Pod666', environment: prod_env)
+    DeployGroup.create!(name: 'Pod667', environment: prod_env)
```
grosser: create on that is not in the environment to make the test more meaningful

test/models/environment_test.rb
```diff
@@ -0,0 +1,28 @@
+require_relative '../test_helper'
+
+describe Environment do
+  it 'should be saved with valid name' do
```
grosser: `describe '.new'`

test/models/environment_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  it 'should default false for is_production' do
+    env = Environment.new(name: 'foo')
+    env.save.must_equal true
+    env.reload.name.must_equal 'foo'
+    env.is_production.must_equal false
+  end
+
+  it 'should require a unique name' do
```
grosser: `describe 'validations'`

test/models/stage_test.rb
```diff
@@ -289,4 +289,22 @@
       assert_equal @clone.new_relic_applications.map(&:attributes), subject.new_relic_applications.map(&:attributes)
     end
   end
+
+  describe 'production flag' do
+    before { ENV['DEPLOY_GROUP_FEATURE'] = '1' }
+    after { ENV['DEPLOY_GROUP_FEATURE'] = nil }
+
+    it 'should be true for stage with production deploy_group' do
+      stage = stages(:test_production)
```
grosser: `let(:stage) { stages(:test_production) }`

test/models/stage_test.rb
```diff
@@ -289,4 +289,22 @@
...
+      stage.update!(production: false)
+      stage.production?.must_equal true
+    end
+
+    it 'should be false for stage with no deploy_group' do
```
grosser: what if production is true ?

test/models/stage_test.rb
```diff
@@ -289,4 +289,22 @@
...
+      stage = stages(:test_production)
+      stage.update!(production: false)
```
grosser: this can go in the before

app/controllers/admin/deploy_groups_controller.rb
sandlerr: if you go with that it doesn't have to be an instance variable and probably use a version of `create` that doesn't raise

app/views/admin/deploy_groups/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Edit DeployGroup</h1>
```
sandlerr: maybe a whitespace here

app/views/admin/deploy_groups/index.html.erb
```diff
@@ -0,0 +1,36 @@
+<h1>Deploy Groups</h1>
+
+<section class="clearfix tabs">
+  <div class="table table-hover table-condensed">
+    <table class="table">
+      <thead>
+        <tr>
+          <th>ID</th>
```
sandlerr: do we normally put this in views? not sure

app/views/admin/environments/index.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        <th>Name</th>
+        <th>Affects Production?</th>
+        <th></th>
+      </tr>
+      </thead>
+
+      <tbody>
+        <% @environments.each do |env| %>
+          <tr>
+            <td><%= env.id%></td>
```
sandlerr: likewise with including the ID

sandlerr:
can we `soft_delete` DeployGroups and Environments?

db/migrate/20150223135916_create_deploy_groups.rb
zendesk-shender: It is, that was what the rails model generator spat out.

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: Oops, thanks. Copy/paste fail!

app/views/admin/deploy_groups/index.html.erb
zendesk-shender: Yup, all the admin screens show the ID, so I just kept it consistent.

app/controllers/admin/environments_controller.rb
zendesk-shender: removed!

app/controllers/stages_controller.rb
zendesk-shender: if you mean why couldn't the '@stage.update_attributes(stage_params)' handle the array of IDs, it couldn't handle the array of IDs with or without the blank element that had to be stripped.
Do you know of another way of doing this?

app/models/stage.rb
zendesk-shender: thanks, fixed!

app/views/stages/_fields.html.erb
zendesk-shender: @grosser Not sure what you mean? This is 2-space indentation, and I'm aligning the parameters to the method at the same level according to: https://github.com/bbatsov/ruby-style-guide#no-double-indent.

grosser: the `# good (normal indent)` example is what I meant:
```
foo(
  bar: 1
)
```

this way there is no wasted space, everything aligns naturally (2,4,6,8 spaces) and changing the initial line does not mean you have to re-indent all following lines

app/controllers/admin/deploy_groups_controller.rb
grosser: ah yes that's what I mean :)

app/controllers/stages_controller.rb
grosser: make a stage_params method that takes care of blank removal

db/migrate/20150223135916_create_deploy_groups.rb
grosser: never mind then :)

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: done

app/views/admin/environments/index.html.erb
zendesk-shender: Yup, showing ID is consistent.

zendesk-shender:
@grosser @sandlerr 
Please review changes. Cheers!

zendesk-shender:
@sandlerr I also changed to using soft_deletes for environments and deploy_groups

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    @deploy_group = DeployGroup.new
+  end
+
+  def create
+    new_deploy_group = DeployGroup.create(deploy_group_params)
+    if new_deploy_group.persisted?
+      redirect_to action: 'index'
+    else
+      flash[:error] = "Failed to create deploy group: #{new_deploy_group.errors.full_messages}"
+      redirect_to action: :new
```
grosser: we should render so the form is still filled

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+  end
+
+  def update
+    if deploy_group.update_attributes(deploy_group_params)
+      redirect_to action: 'index'
+    else
+      flash[:error] = "Failed to update deploy group: #{deploy_group.errors.full_messages}"
+      redirect_to :back
```
grosser: same here

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      redirect_to action: 'index'
+    else
+      flash[:error] = "Failed to create deploy group: #{new_deploy_group.errors.full_messages}"
+      redirect_to action: :new
+    end
+  end
+
+  def update
+    if deploy_group.update_attributes(deploy_group_params)
+      redirect_to action: 'index'
```
grosser: maybe add in `notice: "Saved"`

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  private
+
+  def deploy_group_params
+    params.require(:deploy_group).permit(:name, :environment_id)
+  end
+
+  def deploy_group
+    @deploy_group ||= DeployGroup.find(params[:id])
+  rescue
```
grosser: just 404 / do we need to be this fancy ?

grosser: rescuing  everything can hide ugly bugs -> `rescue ActiveRecord::RecordNotFound`

app/controllers/stages_controller.rb
```diff
@@ -145,4 +145,10 @@ def find_project
   def find_stage
     @stage = @project.stages.find_by_param!(params[:id])
   end
+
+  def update_deploy_groups
```
grosser: maybe do this in stage_params / before the bad data even reaches the model ?

app/views/admin/deploy_groups/index.html.erb
```diff
@@ -0,0 +1,36 @@
...
+              <%= link_to "Delete", admin_deploy_group_path(deploy_group), method: :delete, data: { confirm: "Are you sure?" } %>
+            </td>
+          </tr>
+        <% end %>
+      </tbody>
+    </table>
+  </div>
+  <div class="admin-actions">
+    <div class="pull-right">
+      <%= link_to "Add a new deploy group", new_admin_deploy_group_path, class: "btn btn-default" %>
```
grosser: maybe just `New` or `Add` :)

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,94 @@
...
+  as_a_deployer do
+    unauthorized :get, :index
+    unauthorized :post, :create
+    unauthorized :delete, :destroy, id: 1
+    unauthorized :post, :update, id: 1
```
grosser: should be `:put, :update` ?

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    unauthorized :post, :create
+    unauthorized :delete, :destroy, id: 1
+    unauthorized :post, :update, id: 1
+    unauthorized :get, :new
+  end
+
+  as_a_super_admin do
+    it 'get :index succeeds' do
```
grosser: you can make this a macro like `it_renders_index` and use it in both places

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,94 @@
...
+      get :index
+      assert_response :success
+      assigns(:deploy_groups).count.must_equal 3
+      deploy_groups = assigns(:deploy_groups)
+      deploy_groups.include?(deploy_groups(:deploy_group_pod1)).must_equal true
```
grosser: not sure if this includes test is productive, there is no logic that is being tested, it just assignes all

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,94 @@
...
+      deploy_groups.include?(deploy_groups(:deploy_group_pod2)).must_equal true
+      deploy_groups.include?(deploy_groups(:deploy_group_pod100)).must_equal true
+    end
+
+    it 'get :new succeeds' do
+      get :new
+      assert_response :success
+      assigns(:deploy_group).wont_be_nil
```
grosser: `assert assigns(:deploy_group)` might read a bit easier

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    describe '#create' do
+      it 'valid deploy_group' do
```
grosser: `it 'creates a deploy group'`

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,94 @@
...
+        deploy_group_count = DeployGroup.count
+        post :create, deploy_group: {name: nil}
+        assert_redirected_to new_admin_deploy_group_path
+        flash[:error].must_equal 'Failed to create deploy group: ["Name can\'t be blank", "Environment can\'t be blank"]'
+        DeployGroup.count.must_equal deploy_group_count
+      end
+    end
+
+    describe '#delete' do
+      it 'success' do
```
grosser: `it 'succeeds'` or `it 'deletes'` :)

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+
+      it 'fail for non-existent deploy_group' do
+        delete :destroy, id: -1
+        assert_redirected_to admin_deploy_groups_path
+        flash[:error].must_equal 'Failed to find the deploy group: -1'
+      end
+    end
+
+    describe '#edit' do
```
grosser: `describe '#update'`

test/controllers/admin/environments_controller_test.rb
```diff
@@ -0,0 +1,88 @@
+require_relative '../../test_helper'
+
+describe Admin::EnvironmentsController do
+  as_a_deployer do
+    unauthorized :get, :index
+    unauthorized :post, :create
+    unauthorized :delete, :destroy, id: 1
+    unauthorized :post, :update, id: 1
```
grosser: test for `:get, :new`

test/controllers/admin/environments_controller_test.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  as_a_admin do
+    it 'get index as admin succeeds' do
+      get :index
+      assert_response :success
+      envs = assigns(:environments)
```
grosser: how about `assigns(:environments).must_equal Environment.all` ?

test/controllers/admin/environments_controller_test.rb
```diff
@@ -0,0 +1,88 @@
...
+        assert_redirected_to admin_environments_path
+        flash[:error].must_equal 'Failed to find the environment: -1'
+      end
+    end
+
+    describe '#edit' do
+      before { request.env["HTTP_REFERER"] = admin_environments_url }
+
+      it 'save' do
+        id = environments(:production_env).id
```
grosser: can be reused as `let(:environment){ environments(:production_env) }`

test/models/deploy_group_test.rb
```diff
@@ -0,0 +1,37 @@
+require_relative '../test_helper'
+
+describe DeployGroup do
+  let(:prod_env) { environments(:production_env) }
+
+  describe '.new' do
+    it 'saved with valid name/env' do
```
grosser: `it saves` :)

test/models/stage_test.rb
```diff
@@ -289,4 +289,27 @@
       assert_equal @clone.new_relic_applications.map(&:attributes), subject.new_relic_applications.map(&:attributes)
     end
   end
+
+  describe 'production flag' do
+    let(:stage) { stages(:test_production) }
+    before { ENV['DEPLOY_GROUP_FEATURE'] = '1' }
+    after { ENV['DEPLOY_GROUP_FEATURE'] = nil }
+
+    it 'should be true for stage with production deploy_group' do
```
grosser: `it is`

grosser:
:+1: looks good :)

app/controllers/admin/deploy_groups_controller.rb
zendesk-shender: done.

app/controllers/admin/deploy_groups_controller.rb
zendesk-shender: done.

app/controllers/admin/deploy_groups_controller.rb
zendesk-shender: For JSON responses, yea,  I'd return a 404 happily, but for the UI, might as well try to be a bit friendlier.

app/controllers/admin/deploy_groups_controller.rb
zendesk-shender: done.

app/views/admin/deploy_groups/index.html.erb
zendesk-shender: done! :)

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: seems to work either way.

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: Changed to ensure it actually displays all deploy_groups using assert_select.

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: thanks.

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: done!

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: grammar nazi! :)

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: yup!

test/controllers/admin/environments_controller_test.rb
zendesk-shender: done.

test/controllers/admin/environments_controller_test.rb
zendesk-shender: Changed to check it rendered all the environments.

test/controllers/admin/environments_controller_test.rb
zendesk-shender: done.

test/models/deploy_group_test.rb
zendesk-shender: yup.

zendesk-shender:
@grosser , final change to review: 
I changed the production check to allow for a slow rollout, so that if there is a stage with no deploy_groups assigned, it falls back to the state of the old production field.
```
  def production?
    if ENV['DEPLOY_GROUP_FEATURE']
      deploy_groups.empty? ? super : deploy_groups.any? { |deploy_group| deploy_group.environment.is_production? }
    else
      super
    end
  end
```

grosser:
that sounds good

grosser:
we should also hide the checkbox in the ui then since it won't have any
effect, right ?

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      render 'new'
+    end
+  end
+
+  def update
+    if deploy_group.update_attributes(deploy_group_params)
+      flash[:notice] = "Successfully saved deploy group: #{deploy_group.name}"
+      redirect_to action: 'index'
+    else
+      flash[:error] = "Failed to update deploy group: #{deploy_group.errors.full_messages}"
```
grosser: the errors are normally displayed in the form by rails with `f.error_messages`, but I guess this works too ..

app/controllers/stages_controller.rb
grosser: this ?

app/views/admin/deploy_groups/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Edit Deploy Group</h1>
+
+<%= form_for [:admin, @deploy_group], html: { class: "form-horizontal" } do |form| %>
+  <%= render 'admin/deploy_groups/fields', form: form %>
```
grosser: this is all identical to new except for the button label, so maybe put all into fields and either use "Save" for both or do `label = (@deploy_group.new_record? ? "Create" : "Save")`

app/views/admin/environments/_fields.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <%= form.label :name, "Name", class: "col-lg-2 control-label" %>
+    <div class="col-lg-4">
+      <%= form.text_field :name, class: "form-control" %>
+    </div>
+  </div>
+
+  <div class="form-group">
+    <div class="col-lg-offset-2 col-lg-2">
+      <div class="checkbox">
+        <%= form.label :is_production do %>
```
grosser: this guy should be a static display once you added a deploy group, so you can still see it but no longer modify flip it

app/views/admin/environments/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Edit Environment</h1>
+
+<%= form_for [:admin, @environment], html: { class: "form-horizontal" } do |form| %>
```
grosser: same edit+new partial reuse should be possible here

app/views/stages/_fields.html.erb
```diff
@@ -23,23 +23,44 @@
...
-          Confirm before deployment
-        <% end %>
+  <% if ENV['DEPLOY_GROUP_FEATURE'] %>
+    <div class="form-group">
+      <%= form.label :name, 'Deploy Groups', class: 'col-lg-2 control-label' %>
+      <div class="col-lg-4">
+        <%= form.collection_select(
+          :deploy_groups, DeployGroup.all, :id, :name,
+          {
+            selected: (@stage.deploy_groups.ids if @stage && @stage.deploy_groups),
```
grosser: can there be no `@stage` ?
would `selected: @stage.deploy_groups.ids` also work since it's an empty array ?

app/views/admin/environments/_fields.html.erb
grosser: never mind, this is environment not stage ...

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,91 @@
...
+    unauthorized :post, :update, id: 1
+    unauthorized :get, :new
+  end
+
+  as_a_super_admin do
+    it 'get :index succeeds' do
+      get :index
+      assert_response :success
+      assert_template :index
+      assert_select('tbody tr').count.must_equal DeployGroup.count
```
grosser: that is could be selecting anything / start failing when someone adds a new table ;)

test/controllers/admin/environments_controller_test.rb
grosser: reuse the index test via a it_renders_index class method ?

grosser:
good catch on the fallback behavior, that would have made everything be non-production :D 

:+1: looks safe!

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def new
+    @deploy_group = DeployGroup.new
+  end
+
+  def create
+    @deploy_group = DeployGroup.create(deploy_group_params)
+    if @deploy_group.persisted?
+      redirect_to action: 'index'
```
sandlerr: maybe flash a success message?

app/controllers/admin/deploy_groups_controller.rb
sandlerr: since this is an array, won't it still have the square brackets?

```
irb(main):001:0> arr= ['a', 'b', 'c']
=> ["a", "b", "c"]
irb(main):002:0> "#{arr}"
=> "[\"a\", \"b\", \"c\"]"
```

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,52 @@
+class Admin::DeployGroupsController < ApplicationController
+  before_action :authorize_admin!
+  before_action :authorize_super_admin!, only: [ :create, :new, :update, :destroy ]
+  before_action :deploy_group, only: [:edit, :update, :destroy]
```
sandlerr: am I crazy or is there no `edit` method here?

app/controllers/admin/environments_controller.rb
```diff
@@ -0,0 +1,52 @@
+class Admin::EnvironmentsController < ApplicationController
+  before_action :authorize_admin!
+  before_action :authorize_super_admin!, only: [ :create, :new, :edit, :update, :destroy ]
+  before_action :environment, only: [:edit, :update, :destroy]
```
sandlerr: is there some Rails magic where you don't explicitly have to define the default CRUD actions?

app/controllers/stages_controller.rb
```diff
@@ -145,4 +145,10 @@ def find_project
...
+    # http://stackoverflow.com/questions/8929230/why-is-the-first-element-always-blank-in-my-rails-multi-select-using-an-embedde
+    deploy_group_ids = params[:stage].fetch(:deploy_groups, []).reject(&:empty?)
+    @stage.deploy_groups = DeployGroup.find(deploy_group_ids)
```
sandlerr: how is this persisted?

app/controllers/admin/deploy_groups_controller.rb
grosser: both! :D

app/models/deploy_group.rb
```diff
@@ -0,0 +1,9 @@
+class DeployGroup < ActiveRecord::Base
+  has_soft_deletion default_scope: true
+
+  belongs_to :environment
+  has_and_belongs_to_many :stages
+
+  validates_presence_of :name, :environment_id
```
sandlerr: could validate the presence of the `environment` rather than `environment_id`: http://guides.rubyonrails.org/active_record_validations.html#presence

app/controllers/stages_controller.rb
grosser: assigning an association auto-persists it

app/models/deploy_group.rb
grosser: it's a speed (no extra load) vs correctness tradeoff, either way should be
fine though ...

app/models/stage.rb
```diff
@@ -154,6 +156,14 @@ def send_github_notifications?
...
+      deploy_groups.empty? ? super : deploy_groups.any? { |deploy_group| deploy_group.environment.is_production? }
```
sandlerr: I wonder if we could use `has_many :through` here so that you could do `stage.environments` directly

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,91 @@
...
+      assert_response :success
+      assert assigns(:deploy_group)
+    end
+
+    describe '#create' do
+      it 'creates a deploy group' do
+        post :create, deploy_group: {name: 'pod666', environment_id: environments(:staging_env).id}
+        assert_redirected_to admin_deploy_groups_path
+        deploy_groups = DeployGroup.where(name: 'pod666')
+        deploy_groups.count.must_equal 1
```
sandlerr: Just out of interest, the first hit on Google for 'minitest assert changed' is this: http://grosser.it/2012/05/10/assert_change-and-assert_no_change-for-testunit-minitest-inspired-by-rspec/

sandlerr:
Since we and anyone else using samson won't switch immediately to deploy groups, can we have a 'legacy' fixture or two that are still tested?

Looking awesome! :+1:

test/controllers/admin/deploy_groups_controller_test.rb
grosser: I think we can use `assert_difference "deploy_groups.count", +1 do`

app/controllers/admin/deploy_groups_controller.rb
sandlerr: touche

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  private
+
+  def deploy_group_params
+    params.require(:deploy_group).permit(:name, :environment_id)
+  end
+
+  def deploy_group
+    @deploy_group ||= DeployGroup.find(params[:id])
+  rescue ActiveRecord::RecordNotFound
```
grosser: this can also be counter productive, for example when I type something by hand (the only real usecase except for bookmarks) I get redirected to a page I did not want to go to and have to re-type the whole url ... 
basically it's breaking the established 404 behavior that every other page uses for minimal gain and added code/test cost ... so imo not worth it

app/views/stages/_fields.html.erb
zendesk-shender: Yea, there can be no @stage if it's a new Project. I got caught out on that one.

app/controllers/stages_controller.rb
zendesk-shender: Thanks for making me keep at this, I got the IDs working, and activerecord seems to handle the blank ID when using the deploy_group_ids assignment.

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: looks a bit ugly, but I created a class method and had to put it at top of class.

test/controllers/admin/environments_controller_test.rb
zendesk-shender: done like above.

app/controllers/admin/deploy_groups_controller.rb
zendesk-shender: I was mainly copying the edit stage. And @sandlerr I fixed the display of square brackets. Seems if you don't interpolate the string it'll create separate flashes for each msg.

app/views/admin/deploy_groups/edit.html.erb
zendesk-shender: Good call, thanks!

test/controllers/admin/deploy_groups_controller_test.rb
zendesk-shender: Hah, I'm willing to take that risk, if they add a new table in the simple admin index, then they can fix this test! :)

app/controllers/admin/deploy_groups_controller.rb
zendesk-shender: done

app/controllers/admin/environments_controller.rb
zendesk-shender: Yup, I was in 2 minds as to whether to leave it there. But decided it would remove a few lines that @grosser can comment on! :smile:

app/models/deploy_group.rb
zendesk-shender: grand, I'll leave it as-is then.

app/models/stage.rb
zendesk-shender: True, though I think I'll leave that for another PR for the moment. I really want to get this merged, and for now we don't need the query, as the next step is for environment dashboards which use the association from the other direction.

app/controllers/admin/deploy_groups_controller.rb
zendesk-shender: This is the way the Projects and Stages pages work:
```
  rescue_from ActiveRecord::RecordNotFound do
    flash[:error] = "Project not found."
    redirect_to root_path
  end
```
So I should probably put my original code back for consistency.

zendesk-shender:
We already hide the checkbox in the UI if the DEPLOY_GROUPS_FEATURE flag is
set.

app/controllers/admin/deploy_groups_controller.rb
zendesk-shender: Aye, there is magic with rails afoot with not needing to declare this method explicitly.

zendesk-shender:
@grosser 
Not sure what the workflow, but do previous +1's count if you subsequently comment on things? :)

grosser:
my rule of thumb is keep +1 unless your doing major changes / no comment is left with "you should change blah"

but :+1: anyway ;)

zendesk-shender:
Merged, thanks folks for reviewing! :)

app/controllers/admin/deploy_groups_controller.rb
jonmoter: @sandlerr yup.  If you do a `resources :deploy_groups` in your routes.rb, it creates the routes described [here](http://guides.rubyonrails.org/routing.html#crud-verbs-and-actions). Each of those methods has default behavior inherited from `ActionController::Base`.  In the case of `edit`, it looks for an `edit.html.erb` file in the appropriate views subdirectory, and renders that.

app/models/stage.rb
sandlerr: absolutely, I think it wouldn't be trivial to do because its already `has_and_belongs_to_many`

PR #252, opened 2015-02-26
unify lock controller / logic / rendering

grosser:
@pswadi-zendesk @sandlerr @jonmoter @jwswj 

 - allow description for global locks
 - unify lock forms
 - unify lock display


app/controllers/locks_controller.rb
```diff
@@ -1,31 +1,32 @@
...
+    redirect_to :back, notice: "Unlocked"
   end
 
   protected
 
-  def project
-    @project ||= Project.find_by_param!(params[:project_id])
+  def for_global_lock?
+    case action_name
+    when "create" then !params[:lock].try(:[], :stage_id)
```
jwswj: What does `:[]` do?

jwswj:
Like it @grosser :+1:

app/controllers/locks_controller.rb
grosser: x[1] -> x.send(:[], 1)

grosser:
@zendesk/runway test

PR #254, opened 2015-03-02
hotfix for lock without user

sandlerr:
:koala:
No references to `stage.user` are present in the code - this should allow us to load the homepage while we figure out what went wrong.

/cc @zendesk/quokka

### DIFF
--- a/app/models/lock.rb
+++ b/app/models/lock.rb
@@ -18,7 +18,7 @@ class Lock < ActiveRecord::Base
   end
 
   def summary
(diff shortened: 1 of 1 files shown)


epssy:
:+1:

atroche:
:+1:

princemaple:
@grosser did a refactor around locks recently?

grosser:
https://github.com/zendesk/samson/pull/255

grosser:
not sure how this happens before and after we always set the user and required either deployer or admin to be logged in

PR #257, opened 2015-03-03
Delete the locks of a user when that one is deleted

fneves:
This PR makes Samson soft_delete the locks that a user had put in a stage when that user is soft_deleted.
The current behaviour (the locks being left there) was also causing the main page to return a 500 when a user that had a lock in a stage was deleted, see exception below:

```
ActionView::Template::Error (undefined method `name' for nil:NilClass):
    10:           <li class="<%= 'locked' if stage.locked? %>">
    11:             <%= link_to stage.name, project_stage_path(project, stage), class: 'stage-link' %>
    12:             <%= content_tag :span, 'Deploying', class: 'label label-primary' if stage.currently_deploying? %>
    13:             <%= content_tag :i, '', class: 'fa fa-lock', title: stage.lock.summary if stage.locked? %>
    14:
    15:             <div class="pull-right">
    16:               <% if stage.last_deploy.try(:failed?) %>
```

In line 13, ```stage.lock.summary``` would internally call ```user.name```, but if the user is soft deleted, we would be calling the ```name``` method on ```nil```.

/cc @zendesk/samson @steved555 @grosser @jonmoter

### References
 - Jira link: [INFR-1018](https://zendesk.atlassian.net/browse/INFR-1018)

### Risks
 - None


app/controllers/admin/users_controller.rb
```diff
@@ -1,6 +1,7 @@
 class Admin::UsersController < ApplicationController
   before_action :authorize_admin!
   before_action :authorize_super_admin!, only: [ :update, :destroy ]
+  after_action :remove_user_locks, only: [:destroy]
```
grosser: that sounds like a model concern / if it is done in a different controller we would still want it to be done

test/controllers/admin/users_controller_test.rb
```diff
@@ -87,6 +87,7 @@
 
   describe 'a DELETE to #destroy' do
     let(:user) { users(:viewer) }
+    let!(:locks) { Array.new(2, Lock.create!(user: users(:viewer), stage: stages(:test_staging))) }
```
grosser: this does not do what you want `Array.new(2, :x) -> [:x, :x]`
also you cannot have 2 locks on the same stage

app/controllers/admin/users_controller.rb
grosser: -> `has_many :locks, dependent: :destroy` should already take care of it

jonmoter: In that case, curious that we have some soft-deleted users that own locks that haven't been deleted.

grosser: user does not have has_many :locks -> when user is soft_deleted the lock
stays behind

pdeuter:
is this PR to remedy this ticket: https://zendesk.atlassian.net/browse/INFR-1018

if yes, please reference this ticket.

app/controllers/admin/users_controller.rb
jonmoter: In that case, agreed a better fix is to add that association to the User object, and have it handle the cascading soft delete.

grosser:
yes it is

on top of this change we should probably also do a user.try(:name) ||
"UNKNOWN USER" as before, the fix I did was to fix nils, but does not take
care of not founds

app/controllers/admin/users_controller.rb
fneves: Locks are using ```soft_delete!``` as well. 
And dependent can't point to :soft_delete! ```[:destroy, :delete_all, :nullify, :restrict_with_error, :restrict_with_exception]```
I don't think it is reasonable to override any of these methods, as it would be strange to modify the expected behaviour of ActiveRecord. It might be better to stay with the callback, moved into a concern.

grosser: not sure what you are trying to do, just add `has_many :lock, dependent: :destroy` and everything should work fine

fneves:
@grosser 
Updated the User to contain the  ```has_many:  :locks, dependent: :destroy```. I had no idea that the soft_delete gem would trigger the cascade soft_deletion if ```dependent: :destroy``` was set. It is nice that it does.

test/controllers/admin/users_controller_test.rb
```diff
@@ -97,15 +97,10 @@
     end
 
     as_a_super_admin do
-      setup do
-        delete :destroy, id: user.id
-      end
 
-      it 'soft delete the user' do
+      it 'soft delete the user and all the locks he was associated with' do
```
grosser: it does not test that the locks are gone as the `it` suggests

grosser:
:+1:

test/controllers/admin/users_controller_test.rb
fneves: I had forgot to push

Issue #256, opened 2015-03-03
Local GIT repository


defender:
Hi

We are not using Github but local git repository based on attlasian Stash. 
What should be done from my side to make it work

Thanks,

grosser:
should be possible, just have to change a few places and add some ENV flag that disables the github specific features, might even be possible with some added logic like "when path is local don't do github foobar"

defender:
Thanks grosser

Can you please write down step by step  instructions 

Thanks again

grosser:
I'd start with just using a local repo and seeing what breaks and then disabling that feature if it's not critical or fixing it with a `if url.include?('github.com') ... original code else local code`

geshan:
what about private gitlab?

jwswj:
Samson could be made to work with any Git based service with a little work. It would be great to see someone contribute a pull request for Stash, Bitbucket or Gitlab.

geshan:
thanks.

Issue #52, opened 2014-07-24
Slack notification support


ryandjurovich:
Would be great to have a message sent to a [Slack](https://slack.com/) channel when deployments start/finish.

Advice from the Slack team is the API [chat.postMessage](https://api.slack.com/methods/chat.postMessage) endpoint would be best.

For this we'll need to store config for the token and channel to post to.

grosser:
do you want to write a plugin for that ?

PR #259, opened 2015-03-04
plugins

grosser:
@zendesk/runway 

plugin architecture based on rails engines in gems + callback hooks

hooks get called by loading every `*/samson_plugin.rb` in the load-path

### Risks
 - None


app/controllers/projects_controller.rb
```diff
@@ -23,9 +23,8 @@ def index
 
   def new
     @project = Project.new
-
+    @project.stages.build(name: "Production")
```
gabetax: Accidental duplication of the following line?

grosser: ah yes, too much copy-paste :D

gabetax:
:+1:, but I'd like to hear @fneves's input too.

Any thoughts on extracting tests to live with the plugin? It's certainly easier for the tests to exist with the primary code base for now, but private plugins could be tricky. Maybe a minimal test harness?

grosser:
I think tests belong into the plugin, maybe they get their own test-samson
to be able to run them ... otherwise it could be tricky ... but for now
keeing them here until we know what we want :)

plugins/flowdock/lib/samson_flowdock/samson_plugin.rb
```diff
@@ -0,0 +1,43 @@
+module SamsonFlowDock
+  class Engine < Rails::Engine
+  end
+end
+
+Samson::Hooks.callback :stage_defined do
+  Stage.class_eval do
+    has_many :flowdock_flows
```
sandlerr: since this will be the canonical example, can we work out where the migration will live and how it will run?

grosser: added an example migration

fneves:
I really like this. It follows what I was doing, but in a better, simpler and clearer way.
However, I am not sure that we want to allow a plugin developer to do everything that you can do with a Rails engine. 

I have thought about migrations, and my opinion is that we should not allow migrations on a plugin. We should allow some simple storage (K, V) and that would be it. If anything more complex is required that should rely on an external 3rd party service. This would work for most plugins, and we would avoid problems like a bad plugin that deleting one column that it shouldn't in one bad migration.

I also think that plugins should also have some kind of common template that they share with each other. I have done an example of a WebPlugin (also for a Flowdock integration, but with a little more complex web interface) where a user can define endpoints (we can change the name if you want) that he can interact from the interface (Ajax, Angular, Ember, etc...).

```ruby
module Plugins
  class FlowdockPlugin
    include SamsonSdk::Plugins::WebPlugin
    hook :show_deploy

    stage_config_ui_endpoint do
      render text: 'This is what is going to show in the stage config iframe placeholder'
    end

    ui_endpoint do
      @deploy = Deploy.find(params[:deploy_id])
      render :ui
    end

    get_endpoint :users do
      users = Rails.cache.fetch(:flowdock_users, expires_in: 5.minutes, race_condition_ttl: 5) do
        FlowdockService.new.users
      end
      render json: { users: users }
    end

    post_endpoint :notify do
      @deploy = Deploy.find(params[:deploy_id])
      FlowdockService.new.notify(@deploy, params[:message])
      render json: { message: 'Successfully sent a buddy request to the flows!' }
    end
  end
end
```
Behind the scenes this is generating a Rails controller and several routes, but as a plugin it seems a simple class with some structure. I was not using Rails engines when I developed the prototype, but that is the right direction as it avoids having to do a lot the code that I had to do.

I would prefer to have IFrames with the ui of each plugin, so that they could be sandboxed and not interfere with core Samson UI. This would also give us the ability to load different Javascripts and Stylesheets depending on the plugin, i.e could load the core, and whatever dependencies the particular plugin had.

Another idea I would like to explore are refinements. I believe that one plugin changes made to one core model should not be seen by another plugin or core code. We could use ruby refinements so that those changes would be constrained to the plugin and nothing else.

app/controllers/stages_controller.rb
```diff
@@ -133,9 +129,8 @@ def stage_params
       :update_github_pull_requests,
       :comment_on_zendesk_tickets,
       command_ids: [],
-      flowdock_flows_attributes: [:id, :name, :token, :_destroy],
-      new_relic_applications_attributes: [:id, :name, :_destroy]
-    )
+      new_relic_applications_attributes: [:id, :name, :_destroy],
```
zendesk-shender: I saw your justification for this on your earlier PR, but it does break with the ruby style guide: https://github.com/bbatsov/ruby-style-guide#no-trailing-params-comma

zendesk-shender:
I like how elegant this is. Nice hacks on the gemfile and treating plugins like gems! :)
I'm on the fence with allowing plugins to change the DB:
- Pro: It allows plugin creators to create their activerecord models with all the nice helpers it provides, and if they want to add columns to, for example, a stage to allow extra info saved, then they can simplify even more of their code.
- Con: Plugin creators might try changing core tables and conflict with one another, or just plain break code functionality.

zendesk-shender:
And I agree with keeping tests with the plugin to make the creator responsible for them just like any other gem. We could get crazy and make a plugin generator that includes a test harness to call the various hooks for them.

app/controllers/stages_controller.rb
grosser: meh ... ok :)

grosser: I don't think iframes will work since we want to submit a form with
additional fields and not have multiple indepdent forms + multiple
independent controllers.

Users have to trust the plugins code they install, so they might as well
trust their migrations. The changes it will do are visible in development
since it modified the schema.rb, so there should not be a "oops my column
is gone" moment.

A separate plugin language/framework also means more work for us to write
it + more work to use it (I assume most plugins will be written by us) +
reinventing rails things / forcing people to learn new tools for
questionable gain down the line.

It might be simplest to initially go with the engine solution and then see
if we can refactor the extracted plugins to a nicer framework where it
makes sense/we see a common pattern. This way we will see what things are
needed/reusable before building a framework and always have the escape
hatch of just using rails.

fneves: For IFrames we could use (postMessages)[https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage] to transfer information to the parent window every time an input would change. The parent form would be responsible to then send all the data to the controller. The controller can be the same, but that controller could send the plugin data to a plugin service object that knows how to store the configuration for all the plugins.

I was actually thinking that the plugins would be an easy way to get other people involved and contribute more. After all it seems that there is some curiosity on how to integrate this with other applications  (#256), (#52). This was part of my reasoning for having some kind of plugin templating DSL, to try to make it easier for other people to start and do something. Also, if we wanted to have a plugin that would support Mercurial instead of Git, we would not need a Rails engine. This is a simple example that we will have several classes of plugins with different requirements.

I am not against starting with simple rails engines. We should however look beyond that and try to prepare some kind of plugin framework, with some rules and conventions that can simplify plugin development and avoid boilerplate code. 
I agree that we should always be able to fallback to rails whenever it is really necessary.

grosser:
migrations: I think 99% of the time this should be fine, for the remaining 1% we can add an exception if we run into it

grosser:
Made plugins opt-in via `ENV['PLUGINS']` so they are not all loaded by default.
I'm also thinking the plugin columns should not be part of our 'clean' schema.rb but that would make testing super complicated :/

fneves:
I would say that this is good to start. I have thought about my own opinions and I think we should put this in despite what I wanted. If anyone wants to do the framework that I was talking about, that can be done on a separate gem that we can bring in later. I still think Sandboxing the UI with IFrames would be nicer, but this works as well and the implementation is simpler.

If we want to have assets like extra javascript, and css, how will they be included on the assets pipeline? As an example, the plugin I was doing had an extra js plugin that I had to include in. As I was using Iframes I had a separate layout where I could dynamically link the javascript/stylesheet that I wanted. How can we make it work here?

grosser:
Great, we can always extract more sophisticated methods on top of this
structure :)
Another hook into the layout should do it + add into app/assets of the
plugin

PR #261, opened 2015-03-04
make projects controller tests test for specifica roles and not be admin...

grosser:
... by default

cleanup I did not want to mess up another PR with

 - controller tests should have no user by default
 - should test users in order of least permissions

also simplifying the `unauthorized` helper + giving it  a nicer failure message

@zendesk/runway 

### Risks
 - None


jonmoter:
:+1:

PR #262, opened 2015-03-04
remove unnecessary Project#changeset_for_release

jonmoter:
the logic for that already exists in Release#changeset

/cc @zendesk/runway 

### Risks
 - None

### DIFF
--- a/test/models/release_test.rb
+++ b/test/models/release_test.rb
@@ -51,4 +51,26 @@ describe Release do
       stage.deploys.create!(reference: options.fetch(:reference), job: job)
     end
   end
+
+  describe "#changeset" do
+    let(:project) { projects(:test) }
+    let(:author) { users(:deployer) }
+
+    it "returns changeset" do
+      changeset = Changeset.new("url", "foo/bar", "a", "b")
+      project.releases.create!(author: author, commit: "bar", number: 50)
+      release = project.create_release(commit: "foo", author: author)
+
+      Changeset.stubs(:find).with("bar/foo", "bar", "foo").returns(changeset)
+      assert_equal changeset, release.changeset
+    end
+
+    it "returns empty changeset" do
+      changeset = Changeset.new("url", "foo/bar", "a", "a")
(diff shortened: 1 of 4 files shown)


test/models/release_test.rb
```diff
@@ -51,4 +51,26 @@ def create_deploy!(options)
       stage.deploys.create!(reference: options.fetch(:reference), job: job)
     end
   end
+
+  describe "#changeset" do
+    let(:project) { projects(:test) }
+    let(:author) { users(:deployer) }
+
+    it "returns changeset" do
+      changeset = Changeset.new("url", "foo/bar", "a", "b")
```
grosser: `let(:changeset)` to remove duplication

test/models/release_test.rb
```diff
@@ -51,4 +51,26 @@ def create_deploy!(options)
...
+      project.releases.create!(author: author, commit: "bar", number: 50)
+      release = project.create_release(commit: "foo", author: author)
+
+      Changeset.stubs(:find).with("bar/foo", "bar", "foo").returns(changeset)
+      assert_equal changeset, release.changeset
+    end
+
+    it "returns empty changeset" do
```
grosser: maybe `it "returns empty changeset when there is no prior release"` ?

test/models/release_test.rb
```diff
@@ -51,4 +51,26 @@ def create_deploy!(options)
...
+      changeset = Changeset.new("url", "foo/bar", "a", "a")
+      release = project.releases.create!(author: author, commit: "bar", number: 50)
+
+      Changeset.stubs(:find).with("bar/foo", nil, "bar").returns(changeset)
```
grosser: these are both expects / you'd want this test to fail if they are not called

grosser:
:+1:

jonmoter:
Okay, good refactorings.  I just cut & pasted the test from one file to another, with minimal refactoring, but might as well clean it up while I'm at it

PR #263, opened 2015-03-04
remove nonstandard 404 behavior

grosser:
@zendesk/runway 

now we get a nice:

screen shot 2015-03-04 at 10 53 08 am

and a backtrace in development:

screen shot 2015-03-04 at 12 14 28 pm

### Risks
 - None

### DIFF
--- a/app/controllers/admin/commands_controller.rb
+++ b/app/controllers/admin/commands_controller.rb
@@ -1,10 +1,6 @@
 class Admin::CommandsController < ApplicationController
   before_action :authorize_admin!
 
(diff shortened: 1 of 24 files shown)


pswadi-zendesk:
:heart: Could we also send a message to a configurable flow and / or update datadog metrics when such an error happens. Would be good for us to be aware of.

grosser:
datadog already counts all status codes, so we can make a dashboard with
samson.app.web.request.status.404

sandlerr:
:+1:

PR #264, opened 2015-03-05
speed up lookup of whether deploy is a hotfix [RUN-64]

jonmoter:
DRY up similar code between Release and Deploy.
add a 1-year cached value whether a Deploy or Release is a hotfix.

if we want to pre-warm the cache, post-deploy we can run a command like this in console:

```ruby
Deploy.where(created_at > 6.months.ago).each { |deploy| deploy.hotfix? }
```

/cc @zendesk/samson 

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-64

### Risks
 - None

### DIFF
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -3,7 +3,7 @@ module ReleasesHelper
     path = project_release_path(project, release)
     classes = %w(release-label label)
 
-    if release.changeset.hotfix?
+    if release.hotfix?
       classes << "label-warning"
     else
       classes << "label-success"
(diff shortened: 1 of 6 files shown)


sandlerr:
awesome :+1:

dasch:
I don't see a need to use a module here – can't you push the `hotfix?` method changes into Changeset and just delegate from the other classes? I don't mind a bit of code duplication: the semantics are different.

All you need for caching are the starting and ending commit ids, right?

jonmoter:
The module isn't purely necessary, it was to avoid duplicating code.

The issue with delegating to `Changeset` is that I'm trying to avoid calling `Changset.find`, which requires making an API call out to Github. All those API calls are what is slowing down the rendering of the Stages page so much.

Another way I could refactor this would be to change `Changeset` so that it doesn't call out to GitHub in the `::find` method. Instead it would lazily load and memoize the data when `Changeset#comparison` gets called.

jonmoter:
After talking with @grosser, I'll whip up a PR with the alternate implementation I mention above.

jonmoter:
Check out at alternate implementation at #269.

jonmoter:
Closing in favor of #269.

PR #265, opened 2015-03-05
store exceptions + show exceptions + fix cache not being reset in all tests

grosser:
@zendesk/runway 

screen shot 2015-03-04 at 10 01 44 pm

### Risks
 - None

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -16,8 +16,14 @@ class Changeset
     end
 
     new(comparison, repo, previous_commit, commit)
-  rescue Octokit::NotFound, Octokit::InternalServerError
-    new(NullComparison, repo, previous_commit, commit)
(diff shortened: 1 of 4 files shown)


grosser:
@zendesk/runway

app/models/changeset.rb
```diff
@@ -16,8 +16,14 @@ def self.find(repo, previous_commit, commit)
     end
 
     new(comparison, repo, previous_commit, commit)
-  rescue Octokit::NotFound, Octokit::InternalServerError
-    new(NullComparison, repo, previous_commit, commit)
+  rescue Octokit::NotFound, Octokit::InternalServerError => e
+    error = case e
```
jonmoter: I try to avoid creating local variables with the same name as methods. Maybe `errmsg`?

jonmoter:
Other than the suggestion of a different local variable name, looks good.  :+1:

PR #266, opened 2015-03-05
do not checkout a repo when creating/updating the project + do not fail ...

grosser:
...when cache is deleted

@zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -1,5 +1,4 @@
 class GitRepository
-
   attr_reader :repository_url, :repository_directory
 
   # The directory in which repositories should be cached.
@@ -14,6 +13,7 @@ class GitRepository
 
   def setup!(executor, temp_dir, git_reference)
     executor.output.write("Beginning git repo setup\n")
+    return false unless clone!(executor: executor, from: repository_url, to: repo_cache_dir, mirror: true) unless locally_cached?
     return false unless update!(executor: executor)
     return false unless clone!(executor: executor, from: repo_cache_dir, to: temp_dir)
     return false unless checkout!(executor: executor, pwd: temp_dir, git_reference: git_reference.shellescape)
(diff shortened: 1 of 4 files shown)


test/models/git_repository_test.rb
```diff
@@ -1,91 +1,95 @@
 require_relative '../test_helper'
 
-describe GitRepository, :model do
-
+describe GitRepository do
```
grosser: everything is `ActiveSupport::TestCase` now, no longer need the `:model` hack

app/models/git_repository.rb
```diff
@@ -14,25 +13,31 @@ def initialize(repository_url:, repository_dir:)
 
   def setup!(executor, temp_dir, git_reference)
     executor.output.write("Beginning git repo setup\n")
+    return false unless clone!(executor: executor, from: repository_url, to: repo_cache_dir, mirror: true) unless locally_cached?
     return false unless update!(executor: executor)
     return false unless clone!(executor: executor, from: repo_cache_dir, to: temp_dir)
     return false unless checkout!(executor: executor, pwd: temp_dir, git_reference: git_reference.shellescape)
-    true
+    commit_from_ref(git_reference)
```
grosser: before `commit_from_ref` needed to be called after setup -> now 1 less thing you have to know, setup just returns the ref or false

fneves:
The git repository was actually being cloned when the project was created so that we could assume that we always had the cached repository. The autocomplete when deploying relies on the local cached repository being there. If this is merged, that feature, and possibly other features that are now relying on the cached repo being there, will break.

grosser:
That's what I get for just grepping in app, I'll rework :)

grosser:
revert the project change, now just a simple cleanup + ensure directory is there, looking good ?

fneves:
This would work if repositories were relatively small. If you do a test with a very big repository (600, 700MB) you will break the feature. Its just going to take too long, a [timeout](https://github.com/zendesk/samson/blob/master/lib/references_service.rb#L28) is going to be triggered,and the autocomplete is not going to work.

Conceptually, when I think of a model called GitRepository I would imagine that it would respect the same concepts that I would have when working with Git. So if I have a method called update (maybe should be called pull) I would expect it to trigger an update operation, i.e. a pull. Not a clone and a pull.

Also, I would expect that checkout would be a public interface to the programmer because Git offers us that feature.

What is the problem we are trying to solve with this PR?

grosser:
The problem I'm trying to solve is that when the cache is removed you are unable to deploy.

Making the update also clone would be nice, but you are right, it could mean that if a big repo is missing it will fail after 2 minutes and then try again and so forth, so that could do a lot of duplicate work ... removed it.

I try to make every method that is not used from the outside private, so it's easier to see what can be refactored without a global search + so it does not need unit tests.

Good to go ?

grosser:
@fneves

fneves:
:+1:

PR #267, opened 2015-03-05
fix https urls not being supported and new deploys not showing a diff

grosser:
@zendesk/runway 

### Risks
 - None


jonmoter:
I think this should be a fix for [SAMSON-139](https://zendesk.atlassian.net/browse/SAMSON-139)

jonmoter:
:+1:

PR #268, opened 2015-03-05
hardcode test env values so tests do not fail locally

grosser:
@zendesk/runway 

tests should not fail locally and have consistent env variables -> hardcode them

### Risks
 - None

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -47,3 +47,8 @@ Samson::Application.configure do
 
   config.active_support.test_order = :random
 end
+
+ENV['ZENDESK_URL'] = 'https://test.support.zendesk'
+ENV['SECRET_TOKEN'] = 'd6054cf90db212c8fbc070c896c30398e3275532c5602bdf00cb153b806c000e4e46fac2f3acc0783822b8f6d30b5913b6fbcfdd24914553e745b8aa8ddfa5a4'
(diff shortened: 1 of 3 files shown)


config/environments/test.rb
```diff
@@ -47,3 +47,8 @@
 
   config.active_support.test_order = :random
 end
+
+ENV['ZENDESK_URL'] = 'https://test.support.zendesk'
+ENV['SECRET_TOKEN'] = 'd6054cf90db212c8fbc070c896c30398e3275532c5602bdf00cb153b806c000e4e46fac2f3acc0783822b8f6d30b5913b6fbcfdd24914553e745b8aa8ddfa5a4'
+ENV['DEFAULT_URL'] = 'http://www.test-url.com'
+ENV['ZENDESK_TOKEN'] = 'c30398e3275532c5602bdf00cb153b806c000e4e46fac2f3acc0783822b8f6d3'
```
jonmoter: How about using `||=` for these values in case you want to override them when running a test?

jonmoter:
Other than suggestion, looks good. :+1:

config/environments/test.rb
grosser: That's what I'm trying to avoid, either the test works everywhere with
exactly these settings or we have do change something.
Using ||= would also mean I can pollute my tests by accident when I
have ZENDESK_URL
etc already set.

PR #269, opened 2015-03-06
refactor Changeset for better caching [RUN-64]

jonmoter:
lazy load GitHub comparison information only when needed.
add a 1-year cached value whether a Changeset is a hotfix.

this is an alternative implementation to PR #264. I like this one better.

if we want to pre-warm the cache, we could run a command like this post-deploy:
```ruby
Deploy.where(created_at > 6.months.ago).each { |deploy| deploy.changeset.hotfix? }
```

/cc @zendesk/samson, @grosser. @dasch 

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-64

### Risks
 - Refactors some logic around Changesets, so could affect displays of commits, authors, files, etc. Basically, any information we get from Github.

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -72,7 +72,7 @@ class Changeset
 
   def find_comparison
     if empty?
-      NullComparison.new('Start and end commits are the same')
+      NullComparison.new(nil)
     else
       Rails.cache.fetch(cache_key) do
         GITHUB.compare(repo, previous_commit, commit)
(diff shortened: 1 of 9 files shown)


app/models/changeset.rb
```diff
@@ -84,11 +70,33 @@ def error
 
   private
 
+  def find_comparison
+    if empty?
+      NullComparison.new('Start and end commits are the same')
```
grosser: I would not consider this an error / would not want to see the warning UI, maybe pass nil

app/models/changeset.rb
```diff
@@ -84,11 +70,33 @@ def error
...
+      Rails.cache.fetch(cache_key) do
+        GITHUB.compare(repo, previous_commit, commit)
+      end
+    end
+  rescue Octokit::NotFound, Octokit::InternalServerError => e
+    error_msg = case e
+      when Octokit::NotFound then "Commit not found"
+      when Octokit::InternalServerError then "Internal error #{e.message}"
+      else
+        "Unknown error"
```
grosser: if you indent `case` then you get this weirdness at the end that always looks like something is out of sync

```
  else
    "xxx"
end
```

vs 

```
else
  "xxx"
end
```

grosser:
:+1:

dasch:
:heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: 

:+1:

PR #272, opened 2015-03-06
also cache nil deploys to generate less queries if there is no current/l...

grosser:
...ast deploy

@zendesk/runway 

### Risks
 - None


jonmoter:
:+1:

grosser:
dammit ... this did not build on travis but just code climate :/

grosser:
reverted on master

PR #274, opened 2015-03-06
extract plugin tests

grosser:
@zendesk/runway how does this look ?

 - run normal + plugin tests on rake
 - plugin tests use the same setup (for now ... later we could simplify them / make them independent)
 - samson tests test that hooks are called correctly

(this is not all the flowdock related test code, just a start)

### Risks
 - None


grosser:
@zendesk/runway any feedback ? :)

test/test_helper.rb
```diff
@@ -62,6 +62,12 @@ def unstub_project_callbacks
     Project.any_instance.unstub(:clone_repository)
     Project.any_instance.unstub(:clean_repository)
   end
+
+  def record_hooks(callback, &block)
+    called = []
+    Samson::Hooks.with_callback(callback, lambda{ |*args| called << args }, &block)
+    called
+  end
```
jonmoter: How about adding a short comment above this method explaining its usage, or an example? It took me reading over it a few times and looking at an example to figure out its purpose.

lib/samson/hooks.rb
```diff
@@ -48,6 +48,13 @@ def view(name, partial)
         hooks(name) << partial
       end
 
+      def with_callback(name, block)
+        hooks(name) << block
+        yield
+      ensure
+        hooks(name).pop
+      end
```
jonmoter: Something about this method rubs me the wrong way. Maybe it's having a single method that takes 2 different blocks, and it's not clear which one is the callback. Maybe just having a comment for the method explaining its purpose would clear it up.

grosser:
added comments for both, better ?

jonmoter:
Sure, thanks.  Let's give it a while.  :+1:

PR #275, opened 2015-03-07
Make deploy confirm workflow more intuitive

dadah89:
- stage is static
 - button initially reads "Review" if confirm is necessary and then changes to "Deploy!" (so users know that Deploy! triggers a deploy)
 - changing the reference while reviewing revert to initial state
 - removed unnecessary logic from cancel button
 - shows stage in route and breadcrumb

screen shot 2015-03-06 at 4 29 00 pm

screen shot 2015-03-06 at 4 31 27 pm

@grosser


grosser:
:+1:  @zendesk/runway

pswadi-zendesk:
:+1: after fixing the tests

grosser:
@dadah89 I'll fix them :)

PR #278, opened 2015-03-10
seed release and fix helper

grosser:
that's what I get for copying the test output :/

@princemaple 

### Risks
 - None


princemaple:
:+1:

PR #279, opened 2015-03-10
Fix redeploy

princemaple:
@grosser 

another one

/cc @zendesk/samson 

### Related
- https://github.com/zendesk/samson/pull/278
- https://github.com/zendesk/samson/pull/277


sandlerr:
:+1:

PR #280, opened 2015-03-10
Change from data-confirm to data-confirmation to avoid HTML reserved att...

zendesk-shender:
There was a conflict with the use of the 'data-confirm' element attribute (http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html) with the rails helpers. So I changed it back to the 'data-confirmation' attribute.

samson

/cc @zendesk/samson @grosser @princemaple @sandlerr 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -94,7 +94,7 @@ $(function () {
 
   function toggleConfirmed() {
     confirmed = !confirmed;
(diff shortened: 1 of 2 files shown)


princemaple:
:+1:

zendesk-shender:
merged!

grosser:
:heart:  good find :D

PR #281, opened 2015-03-10
convert service objects into callbacks

grosser:
This service objects looks like unnecessary indirection/abstraction to me,
now we had to remember to never create a release, but let the release service do it for us.
The service also hides the normal AR callbacks like .create / .create! / .create(validate: false) etc that can be useful.
Also now the release knows how to create itself and does not spread that logic between itself+project+service, resulting in tests no longer running with half-set-up releases -> a few more stubs necessary.

@jwswj @zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -2,6 +2,8 @@ class Release < ActiveRecord::Base
   belongs_to :project, touch: true
   belongs_to :author, polymorphic: true
 
+  before_create :assign_release_number
+
   def self.sort_by_version
     order(number: :desc)
   end
@@ -36,5 +38,12 @@ class Release < ActiveRecord::Base
 
   def self.find_by_version!(version)
     find_by_number!(version[/\Av(\d+)\Z/, 1].to_i)
+  end
+
+  private
+
+  def assign_release_number
+    latest_release_number = project.releases.last.try(:number) || 0
+    self.number = latest_release_number + 1
   end
 end
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -20,7 +20,7 @@ class Integrations::BaseController < ApplicationController
     if project.create_releases_for_branch?(branch)
       unless project.last_released_with_commit?(commit)
         release_service = ReleaseService.new(project)
-        release_service.create_release(commit: commit, author: user)
+        release_service.create_release!(commit: commit, author: user)
       end
     end
   end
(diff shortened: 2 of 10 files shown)


jwswj:
I think @ilkkao or @dasch wrote this initially. Not sure why it was built as a "service".

gabetax:
Personally, I don't like hooks that perform actions outside of the object itself. `assign_number` is OK to me, but triggering github interaction and deploys just by calling `.save` on a model feels like an over-reach of concerns.

grosser:
Yeah I feel a bit weird about the github callback too, but my line of thought was that normally we also send emails during object creation and that is also an outside interaction.

Not doing the callback would mean leaving an object created on the commandline in an invalid/unexpected state, which is also not nice, not sure if there is a nice middle ground ...

app/models/release.rb
```diff
@@ -2,6 +2,10 @@ class Release < ActiveRecord::Base
   belongs_to :project, touch: true
   belongs_to :author, polymorphic: true
 
+  before_create :assing_release_number
```
sandlerr: `assign_release_number` (spelling correction)

app/models/release.rb
```diff
@@ -35,9 +39,28 @@ def author
...
+    if version =~ /\Av(\d+)\Z/
+      number = $1.to_i
+      find_by_number(number)
+    end
+  end
+
+  private
+
+  def assing_release_number
+    latest_release_number = project.releases.last.try(:number) || 0
```
sandlerr: would it be better to an ActiveRecord query that's equivalent to `ORDER BY number DESC LIMIT 1` ?
Not sure if its possible for IDs to be out of sync with release numbers - if not, you could do `project.releases.count + 1`

grosser: they should be always in sync, but number might not be the same as count when you delete 1 or somehow else start at a higher number

grosser:
To me the only reason to keep the service thing would be "it's so super complicated, we have to extract this" or "we need normal creation and creation without all the callbacks", but none of these seems to apply here ...

dasch:
We _never_ do outside interaction from within a model in HC, and it's served us well. I see no reason to start doing that in Samson.

grosser:
Classic is using all kind of callbacks and it also "served us well" ... so yeah either one works ... leaving the out-of-model callback in ReleaseService ... good to go ?

dasch:
Those callbacks were a constant source of pain for me, but your mileage may vary :-)

:+1:

PR #283, opened 2015-03-10
restructure releases test

grosser:
@zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -34,10 +34,7 @@ class Release < ActiveRecord::Base
     super || NullUser.new(author_id)
   end
 
-  def self.find_by_version(version)
-   if version =~ /\Av(\d+)\Z/
-     number = $1.to_i
-     find_by_number(number)
-   end
- end
+  def self.find_by_version!(version)
+    find_by_number!(version[/\Av(\d+)\Z/, 1].to_i)
+  end
 end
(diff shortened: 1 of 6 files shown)


app/models/release.rb
```diff
@@ -34,10 +34,7 @@ def author
...
   end
 
-  def self.find_by_version(version)
-   if version =~ /\Av(\d+)\Z/
-     number = $1.to_i
-     find_by_number(number)
-   end
- end
+  def self.find_by_version!(version)
+    find_by_number!(version[/\Av(\d+)\Z/, 1].to_i)
```
grosser: basically making sure it's always a NotFound / 404 instead of having to deal with custom logic, nobody should find with an unknown version so no need to add extra fancy logic

test/controllers/releases_controller_test.rb
```diff
@@ -2,28 +2,50 @@
 
 describe ReleasesController do
   let(:project) { projects(:test) }
+  let(:release) { releases(:test) }
 
-  describe "#create" do
-    let(:release_params) { { commit: "abcd" } }
+  as_a_viewer do
```
grosser: structuring by viewer/deployer/admin give the test a nice flow/structure, since an action is only tested once and does not need to repeat the scopes multiple times, the alternative would be to do viewer/deployer etc inside a `describe action` block ...

test/models/project_test.rb
```diff
@@ -55,7 +55,7 @@ def clone_repository(project)
     it "defaults to release number 1" do
```
zendesk-shender: I don't think the description matches the test anymore... maybe "increments release number"?

zendesk-shender:
:+1: after change test name.
Thanks for educating me on the string[regex, capture_group] api! :)

PR #284, opened 2015-03-10
do not compile assets during test runs for faster tests

grosser:
not sure how this worked on travis before, but if I don't have public/assets compiled tests refuse to run or time out while trying to compile, so just disable assets in test ... nice side effect is that tests get faster

@zendesk/runway 

### Risks
 - None


sandlerr:
:+1:

PR #285, opened 2015-03-10
enable code coverage reporting via Code Climate

jonmoter:
I set the CODECLIMATE_REPO_TOKEN env variable in the Travis admin UI

/cc @zendesk/samson 

### Risks
 - None


Gemfile
```diff
@@ -84,4 +84,5 @@ group :test do
   gem 'webmock', require: false
   gem 'simplecov', require: false
   gem 'query_diet'
+  gem 'codeclimate-test-reporter', require: false
```
grosser: we should get rid of auto-require ... but that's a different PR :)

grosser:
:+1: 

/fyi @apanzerj

apanzerj:
:-D

PR #286, opened 2015-03-10
catch all kinds of Oktokit errors

grosser:
@zendesk/runway 

previously not catching Octokit::Unauthorized

### Risks
 - None

### DIFF
--- a/test/models/changeset_test.rb
+++ b/test/models/changeset_test.rb
@@ -14,10 +14,16 @@ describe Changeset do
       Changeset.new("foo/bar", "a", "b").comparison.to_h.must_equal :x => "y"
     end
 
-    it "catches exceptions" do
-      GITHUB.expects(:compare).raises(Octokit::NotFound)
-      comparison = Changeset.new("foo/bar", "a", "b").comparison
-      comparison.error.must_equal "Commit not found"
+    {
+      Octokit::NotFound => "Github: Not found",
+      Octokit::Unauthorized => "Github: Unauthorized",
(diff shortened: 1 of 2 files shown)


jonmoter:
:+1:

test/models/changeset_test.rb
```diff
@@ -14,10 +14,16 @@
...
     end
 
-    it "catches exceptions" do
-      GITHUB.expects(:compare).raises(Octokit::NotFound)
-      comparison = Changeset.new("foo/bar", "a", "b").comparison
-      comparison.error.must_equal "Commit not found"
+    {
+      Octokit::NotFound => "Github: Not found",
+      Octokit::Unauthorized => "Github: Unauthorized",
+      Octokit::InternalServerError => "Github: Internal server error",
```
zendesk-shender: Use ruby2.x hash syntax?
```
{
  Octokit::NotFound: "Github: Not found",
  Octokit::Unauthorized: "Github: Unauthorized"
...
```
And no trailing comma! ;)

grosser: has syntax only works with symbols (and strings in 2.2)

```
 {Fixnum: 123}
=> {:Fixnum=>123}

class A; class B;end;end
=> nil
>> {A::B: 1}
SyntaxError: (irb):2: syntax error, unexpected ':', expecting =>
{A::B: 1}
```

zendesk-shender: Ah, thanks for info.

PR #287, opened 2015-03-11
select date for deploy display

grosser:
@zendesk/runway 

https://zendesk.atlassian.net/browse/RUN-12

screen shot 2015-03-10 at 9 00 32 pm

input is just a text field ... do we want some fancy date-time picker ... idk

### Risks
 - None

### DIFF
--- a/app/views/dashboards/show.html.erb
+++ b/app/views/dashboards/show.html.erb
@@ -1,6 +1,13 @@
 <h1><%= @environment.name %></h1>
 
 <section>
+  <p>
+    <%= form_tag "", method: :get do %>
+      Status at <%= datetime_local_field_tag :before, @before.strftime("%Y-%m-%dT%H:%M:%S") %>
+      <%= submit_tag "Timetravel" %>
+    <% end %>
+  </p>
+
   <table class="table">
     <thead>
     <tr>
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -72,18 +72,20 @@ class ActiveSupport::TestCase
     Project.any_instance.unstub(:clean_repository)
   end
 
-  def map_ar_queries
+  def ar_queries
     QueryDiet::Logger.queries.map(&:first) - ["select 1"]
   end
 
   def assert_sql_queries(count, &block)
-    old = map_ar_queries.size
+    old = ar_queries
     yield
-    new = map_ar_queries.size - old
+    new = ar_queries
+    new_count = new.size - old.size
+    message = new[old.size..-1].join("\n")
     if count.is_a?(Range)
-      assert_includes count, new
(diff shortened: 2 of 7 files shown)


test/test_helper.rb
```diff
@@ -72,18 +72,20 @@ def unstub_project_callbacks
...
-    old = map_ar_queries.size
+    old = ar_queries
     yield
-    new = map_ar_queries.size - old
+    new = ar_queries
+    new_count = new.size - old.size
+    message = new[old.size..-1]
     if count.is_a?(Range)
-      assert_includes count, new
+      assert_includes count, new_count, message
```
grosser: now shows all offending queries when it fails

zendesk-shender:
I thought people were wanting a simple x-axis time-line where deploys appear as data-points on the graph?
The use-case that came to mind was someone seeing what deploys were taking place recently to cause a possible breakage. It's a bit more cumbersome for the user to select the time of the breakage and step back in time bit by bit to see the various deploys.

It also seems the 'Recent Deploys' page would solve the above use-case, so not sure if we need the timeline after all?

app/views/dashboards/show.html.erb
```diff
@@ -1,6 +1,13 @@
 <h1><%= @environment.name %></h1>
 
 <section>
+  <p>
+    <%= form_tag "", method: :get do %>
+      Status at <%= text_field_tag :before, @before.to_s(:db) %>
```
zendesk-shender: If you use 'datetime_local_field_tag' you'll get a datepicker for free on most browsers.
samson

grosser: fancy :D

grosser:
hmm let's try that ... could get a bit messy when we have many deploys ...

jonmoter:
@zendesk-shender, there are two JIRA stories related to time.  [RUN-12](https://zendesk.atlassian.net/browse/RUN-12) asks for a grid of all the versions live on all the different projects at a specific time.  I think that's what @grosser implemented here.  My thinking is that if we know a bug or outage happened on Thursday 2pm, you could look at this graph to see an overview of what was live at that moment.

There's also [RUN-13](https://zendesk.atlassian.net/browse/RUN-13), which is an x-axis timeline of when different versions of projects went live in a particular DeployGroup. This would show in a given pod, which were the new deploys that happened near a particular time.

I think of them as two different visualizations of the same info, that might be useful depending on context.

grosser:
now with fancy select, ok to ship this, then I'll go work on RUN13 next

grosser:
@jonmoter @zendesk-shender

zendesk-shender:
fancy datetime picker FTW! :)  :+1:

PR #289, opened 2015-03-11
fix bug in calculating last deploy for comparisons [BILL-1019]

jonmoter:
Fixes a problem in comparing a new deploy with the "previous" deploy for that stage.

### References
 - Jira link: https://zendesk.atlassian.net/browse/BILL-1019

### Risks
 - None

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -110,7 +110,7 @@ class Deploy < ActiveRecord::Base
   end
 
   def self.prior_to(deploy)
-    deploy.persisted? ? where("#{table_name}.id < ?", deploy.id) : self
+    deploy.persisted? ? where("#{table_name}.id < ?", deploy.id) : all
   end
 
   def self.expired
--- a/test/models/deploy_test.rb
+++ b/test/models/deploy_test.rb
@@ -85,9 +85,13 @@ describe Deploy do
 
   describe ".prior_to" do
     let(:deploys) { Array.new(3).map { create_deploy! } }
+    let(:prod_stage) { stages(:test_production) }
+    let(:prod_deploy) { create_deploy!(stage: prod_stage) }
+
     before do
       Deploy.delete_all
       deploys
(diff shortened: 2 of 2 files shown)


app/models/deploy.rb
```diff
@@ -110,7 +110,7 @@ def self.finished_naturally
   end
 
   def self.prior_to(deploy)
-    deploy.persisted? ? where("#{table_name}.id < ?", deploy.id) : self
+    deploy.persisted? ? where("#{table_name}.id < ?", deploy.id) : all
```
grosser: all is deprecated ... maybe `.where()` or `.where('1=1')` ?

grosser:
:+1: except for the .all

app/models/deploy.rb
princemaple: You sure all is deprecated? I think it was a bug in 4.0 saying it's
deprecated

jonmoter: To be fair, calling `#all` turns it into an array instead of an arel, which would interfere with chaining. So it's probably better to use `where('1=1')`, unless you know of another method that returns an arel but preserves chaining.

princemaple: Not sure if this is true either.

grosser: `.where(nil)` seems to be best / does not add an extra condition

princemaple: Where(nil) is a workaround in 4.0
In 4.1+ `all` should work as expected.

jonmoter: Ah yes, good point. It appears that `all` is not deprecated, and now returns an arel:

```
irb(main):002:0> Deploy.where('id > 5').all.class
=> Deploy::ActiveRecord_Relation
```

and it's still listed in the API docs: http://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-all

PR #290, opened 2015-03-11
mini-profile to make perfomance more visible

grosser:
@zendesk/runway 

tried bullet too, but the first thing it found as a false positive and there was no way to ignore it, also ui was basically non-existent ...

screen shot 2015-03-11 at 3 56 28 pm
screen shot 2015-03-11 at 3 56 21 pm
screen shot 2015-03-11 at 3 56 02 pm

### Risks
 - None


zendesk-shender:
:heart: it,  :+1:

gabetax:
:+1: I'm a fan.

PR #291, opened 2015-03-11
remove includes, auto-add them

grosser:
knowing in the controller that models your views will needs usually ends up in messy code + missing includes / too many includes ... let's have rails just auto-detect includes that it needs

```
>> Deploy.all.map(&:stage)
  Deploy Load (0.4ms)  SELECT `deploys`.* FROM `deploys` WHERE `deploys`.`deleted_at` IS NULL  ORDER BY `deploys`.`created_at` DESC, `deploys`.`id` DESC
  Stage Load (0.5ms)  SELECT `stages`.* FROM `stages` WHERE `stages`.`deleted_at` IS NULL AND `stages`.`id` IN (554917358, 685639643)  ORDER BY `stages`.`order` ASC
```

@zendesk/runway 

### Risks
 - Missing includes / page slowness


zendesk-shender:
It's a bit concerning that 1/2 the unit tests seem to have failed just by including this gem.

If the gem helps most cases, and where it didn't detect the eager loads properly still allowed AR to fallback on loading associations as needed then grand.
But if it doesn't then debugging problems becomes a bit more complicated with the extra layer of magic.

grosser:
all tests fixed by adding `auto_include: false` to to has_many :through that have ordering on the through association as per https://github.com/salsify/goldiloader/issues/18

so should be good/safe :)

princemaple:
Just curious, this gem loads all associations no matter what, do we not care?

grosser:
I only loads associations that get used in an N+1 way.
For example a false-positive would be `Stage.all.first.project` -> would trigger an unnecessary Project include, but that usecase is very rare -> better to fix the 95% usecase.

grosser:
good to go ?

we can compare newrelic times after the deploy to see if something went wrong

princemaple:
:+1:

grosser:
Looks like response times were not affected either way by replacing includes by goldiloader

screen shot 2015-03-19 at 10 01 22 pm

PR #292, opened 2015-03-11
simple fix for RUN-28

grosser:
@zendesk/runway 

basically RUN-28 asks for considering all deploys to belongs to all stages that include all the deploy groups, for example:

stage A: 1,2
stage B: 1
stage C: 2,3

so deploy to A would show up in B, but not in C

adding this logic is pretty difficult sql wise and maybe no longer necessary since we already have the dashboard + adding a small fix here to show the deploy date, so in the overview of A and B it is obvious what was deployed last / is correct

screen shot 2015-03-11 at 4 53 25 pm

### Risks
 - None

### DIFF
--- a/app/views/projects/_stage.html.erb
+++ b/app/views/projects/_stage.html.erb
@@ -4,12 +4,12 @@
       <%= link_to stage.name, project_stage_path(project, stage) %>
       <%= stage_lock_icon stage %>
     </td>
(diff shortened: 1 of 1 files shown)


zendesk-shender:
:+1: to change, and agree with this info already being displayed in environment dashboards.

PR #293, opened 2015-03-12
do not preload byebug, it is not necessary

grosser:
shaves of ~0.1s of test startup time

@zendesk/runway 

### Risks
 - None


sandlerr:
:+1:

PR #294, opened 2015-03-12
Add a Dockerfile

steved:
@zendesk/runway

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM ruby:2.2.1
+
+RUN mkdir /app
+WORKDIR /app
+
+# Mostly static
+ADD config.ru /app/
+ADD Rakefile /app/
+ADD bin /app/bin
+ADD public /app/public
+ADD db /app/db
+ADD vendor/assets /app/vendor/assets
+
+# Gems
+ADD Gemfile /app/
+ADD Gemfile.lock /app/
+ADD vendor/cache /app/vendor/cache
+
+# Plugins need to be added before bundling
+# because they're loaded as gems
+ADD plugins /app/plugins
+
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ADD plugins /app/plugins
+
+RUN bundle install --without test sqlite postgres --quiet --local --jobs 4 || bundle check
+
+# Code
+ADD config /app/config
+ADD app /app/app
+ADD lib /app/lib
+
+RUN DATABASE_URL=mysql2://user:pass@127.0.0.1/null RAILS_ENV=development PRECOMPILE=1 bundle exec rake --trace assets:precompile
```
sandlerr: why this db config? Rails 4 and up always tries to connect even if you're precompiling?

steved555: Yeah, it looks for a configuration, even if it doesn't connect.

Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM jacobat/ruby:2.2.1
```
grosser: hmm this is unusable for anyone outside of zendesk ... some open-source image would be nice ...

config/puma.rb
```diff
@@ -1,4 +0,0 @@
-threads 8,250
```
grosser: where did that 8,250 config go ?

grosser:
:+1:

Dockerfile
steved555: Right, thanks.

config/puma.rb
steved555: This file wasn't loaded by `rails server`, so it seems kind of pointless. Unless people actually run `puma` directly.

sandlerr:
Already merged

PR #296, opened 2015-03-13
Fix problem with displaying dashboard waaaay back in time.

zendesk-shender:
There is a bug where if you timetravel before time began, i.e. where there were no deploys, the environment which is lazy-loaded wouldn't get loaded.

/cc @zendesk/runway 

### References
 - Jira link:

### Risks
 - None


jonmoter:
:+1:

grosser:
test!

otherwise :+1:

PR #297, opened 2015-03-13
fixtures do not get their model name in their name

grosser:
avoid redundancy and make things easier to read `users(:admin)` not `users(:admin_user)`

@zendesk/runway 

### Risks
 - None


gabetax:
Sure :+1:

PR #300, opened 2015-03-16
stop caching fast things

grosser:
@zendesk/runway the only thing slow in the deploy is the hotfix check, which is already cached ... and the only n+1 query is the previous_deploy check which is only needed for the hotfix -> stop caching html and cache the data ?

### Risks
 - Deploy display being slow


grosser:
hmmm release also uses hotfix? ... I'll leave it like it is for now ...

PR #302, opened 2015-03-17
on Stage Edit, show environment for each deploy group and sort them

jonmoter:
/cc @zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/views/stages/_fields.html.erb
+++ b/app/views/stages/_fields.html.erb
@@ -28,7 +28,7 @@
       <%= form.label :name, 'Deploy Groups', class: 'col-lg-2 control-label' %>
       <div class="col-lg-4">
         <%= form.collection_select(
-          :deploy_group_ids, DeployGroup.all, :id, :name,
+          :deploy_group_ids, @deploy_groups, :id, :long_name,
           {
             selected: (@stage.deploy_group_ids if @stage && @stage.deploy_groups),
             include_blank: false,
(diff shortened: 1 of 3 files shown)


app/views/stages/_fields.html.erb
```diff
@@ -28,7 +28,7 @@
       <%= form.label :name, 'Deploy Groups', class: 'col-lg-2 control-label' %>
       <div class="col-lg-4">
         <%= form.collection_select(
```
grosser: might be simpler to move the lookup here so we don't have to repeat logic in 2 placed + break when that logic gets out of sync

jonmoter: As a rule of thumb, I dislike seeing DB queries like `DeployGroup.includes(:environment).order(:environment_id, :name).all` in the view code.

grosser: no need to add includes anymore ;)

grosser: then maybe a helper like `has_deploy_group?`

grosser: we are triggering all kinds of queries in the view layer, so a strict
separation does not exist anyway. This is just filling a select, maybe
should even be cached, so I would be easier to handle in the view.

jonmoter: Why wouldn't I want to add the `includes(:environment)`?  Is that something that's now handled by that gem you added?

grosser: exactly :)

jonmoter:
There we go, @grosser, better?

grosser:
:+1:

PR #305, opened 2015-03-17
RFC: Guides

mwerner:
I was deploying Lotus today, jumping between the Github page I thought was the official process to deploy Lotus and Samson when I figured it would be much more straightforward if there was one markdown document that was on the project I was deploying.

So I added that.

Trust me, I'm torn on having yet another place where documentation exists, but I feel like this could be useful.

@steved @jonmoter @fneves @zendesk-shender @jish @shajith @chrsjxn @ducarroz 

screen shot 2015-03-17 at 3 48 27 pm

### DIFF
--- a/app/assets/stylesheets/guides.scss
+++ b/app/assets/stylesheets/guides.scss
@@ -1,0 +1,4 @@
+#guide_body {
+  font-family: Courier;
+  height: 390px;
+}
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -1,0 +1,42 @@
+class GuidesController < ApplicationController
+  before_action :authorize_admin!, except: [:show]
+
+  helper_method :project
+  helper_method :guide
+
+  def create
+    @guide = Guide.new(guide_params.merge({
+      project_id: project.id
+    }))
+
+    if guide.save
+      redirect_to project_guide_path(project)
+    else
+      flash[:error] = guide.errors.full_messages
+      render :new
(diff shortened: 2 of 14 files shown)


mwerner:
I don't know what the deal is with the `schema.rb` file, but we can deal with that if we think this is a good idea.

jish:
I believe the dashboards in Samson are an iframed in page? I know `iframe` is a bit of a bad word too, but we could eliminate the "documentation in multiple places" problem by using an iframe.

mwerner:
Yeah, although i think we get benefits by having it being data in samson. There's no doubt which the official guide is and you can more easily link to and access information in samson.

jish:
Oh, so like Samson becomes the source of truth for deployment documentation?

mwerner:
bump

The thinking was it would be good to have a authoritative deploy doc close to the tool doing the job. If no one sees value in the PR, I'll close.

pswadi-zendesk:
I like the idea of having master doc. However adding it to "samson" seems unnatural. As @jish was mentioning, rather than having the doc in multiple places, perhaps the doc is in the same github repo and we could add config in project for samson that can refer to it. That way, the doc needs to be with the project and if standardized, Samson can use it for each of the project from the repository itself. What do you think?

mwerner:
I don't feel strongly one way or the other. I guess I thought we would just not have any deployment documentation elsewhere since samson is our deploy tool. I can close it, sounds fine

PR #306, opened 2015-03-17
on Environment dashboards, hide projects with no deploys

jonmoter:
by default hides rows for Projects that don't have any deploys in that environment.  Adds a UI element to hide/show them.

/cc @zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/assets/javascripts/dashboards.js
+++ b/app/assets/javascripts/dashboards.js
@@ -1,0 +1,5 @@
+$(function () {
+  $('#no-deploy-toggler').click(function(e) {
+    $('tr.no-elements').toggleClass('hidden');
+  });
+});
--- a/app/assets/javascripts/dashboards.js
+++ b/app/assets/javascripts/dashboards.js
@@ -1,5 +1,0 @@
-$(function () {
-  $('#no-deploy-toggler').click(function(e) {
-    $('tr.no-elements').toggleClass('hidden');
-  });
-});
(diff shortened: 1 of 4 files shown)


jonmoter:
samson

grosser:
No need for a toggle ... projects without deploys are useless ...

jonmoter:
Actually, seeing all the projects is useful for me to see which ones haven't been configured with their DeployGroup mapping.  So I added that for my own benefit. :-)

app/assets/javascripts/dashboards.js
```diff
@@ -0,0 +1,5 @@
+$(function () {
+  $('#no-deploy-toggler').click(function(e) {
+    $('tr.no-elements').toggleClass('hidden');
```
grosser: just `.toggle()` ?

grosser: could also do `$("a.toggle").click(function(e){ $(this).data('target').toggle() })` to get something reusable

grosser:
:+1: kind of useful ... kind of strange ... but why not ...

app/assets/javascripts/dashboards.js
jonmoter: I tried it with `.toggle()`, but it didn't work for some reason.  Good call on making it more reusable though.  I'll do that.

jonmoter:
@grosser I refactored the code a little per your suggestion to make it more reusable

grosser:
:+1:

PR #307, opened 2015-03-18
fail fast on unallowed parameters

grosser:
@zendesk/runway 

now we just fail when an unallowed parameter is sent, found missing permit for use_github_deployment_api ...

screen shot 2015-03-18 at 1 25 35 am

 - [x] fix up user_params in users_controller

### Risks
 - None

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -16,8 +16,7 @@ class Admin::UsersController < ApplicationController
 
   def update
     if user.update_attributes(user_params)
-      role = Role.find(user_params[:role_id]).name
-      Rails.logger.info("#{current_user.name_and_email} changed the role of #{user.name_and_email} to #{role}")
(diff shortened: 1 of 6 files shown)


grosser:
@zendesk/runway

app/controllers/admin/users_controller.rb
```diff
@@ -15,9 +15,8 @@ def index
   end
 
   def update
-    if user.update_attributes(user_params)
-      role = Role.find(user_params[:role_id]).name
-      Rails.logger.info("#{current_user.name_and_email} changed the role of #{user.name_and_email} to #{role}")
+    if user.update_attributes(params.require(:user).permit(:role_id))
```
zendesk-shender: I liked having user_params as separate function to keep noise down in this line, but not too bothered by it.

zendesk-shender:
:+1:

PR #308, opened 2015-03-18
Resolve and track tag for deploy and export to Job as TAG environment variable

gabetax:
For canary analysis, our applications must include the application version in metrics sent to statsd. The version string being sent should originate in Samson, so the string is uniform end-to-end. The tag is generated by `git describe`, so it will fall back to a unique human readable reference format (e.g. `v1.2.3-13-g1234abc`) if the deployed commit is not explicitly tagged.

The tag is exported to the execution so that capistrano can write the tag out on deploys. This also means we can (and need to) remove the `export TAG=$REVISION` global command in all of our projects.

### References

- Jira: [RUN-67](https://zendesk.atlassian.net/browse/RUN-67)

### Risks

Medium, due to our own capistrano scripts already using `ENV['TAG']`. We can toggle the `TAG=$REVISION` command per-stage until we're confident it's safe, and then delete the command entirely.

/cc @zendesk/runway @steved

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -39,6 +39,16 @@ class GitRepository
     description.split('-').last.sub(/^g/, '')
   end
 
+  def tag_from_ref(git_reference)
+    Dir.chdir(repo_cache_dir) do
+      tag = IO.popen(['git', 'describe', '--tags', git_reference], err: [:child, :out]) do |io|
+        io.read.strip
+      end
+
+      tag if $?.success?
+    end
+  end
+
   def repo_cache_dir
     File.join(cached_repos_dir, @repository_directory)
   end
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -105,6 +105,7 @@ class JobExecution
       "export DEPLOYER_EMAIL=#{@job.user.email.shellescape}",
       "export DEPLOYER_NAME=#{@job.user.name.shellescape}",
       "export REVISION=#{@reference.shellescape}",
+      "export TAG=#{@job.tag.to_s.shellescape}",
       "export CACHE_DIR=#{artifact_cache_dir}",
       "cd #{dir}",
       *@job.commands
@@ -127,7 +128,8 @@ class JobExecution
     locked = lock_project do
       return false unless @repository.setup!(@executor, dir, @reference)
       commit = @repository.commit_from_ref(@reference)
-      @job.update_commit!(commit)
+      tag = @repository.tag_from_ref(@reference)
+      @job.update_git_references!(commit: commit, tag: tag)
     end
 
     if locked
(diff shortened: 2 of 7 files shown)


grosser:
can capistrano send the deploy event with the tag <-> app does not need to know about anything ?

grosser:
isn't the user already giving us the tag that should be deployed, maybe we can just use that ?

gabetax:
If the deploy statsd event that needed version tagging, then cap could handle this and the application could stay ignorant. However, we need to include the the tag on literally every metric sent by the application itself. That needs to live inside the application.

I prefer using the `git describe --tags` output over the user-provided reference because users could provide a branch name (changes over time), a specific commit id (not readable), or any of the other `git rev-parse` gibberish. This shouldn't happen in production, but we should track and report it properly.

gabetax:
Oh, I also mixed up my JIRA links, which may have thrown you. RUN-67.

grosser:
Ah yes :D

gabetax:
ping @zendesk/runway

app/models/git_repository.rb
```diff
@@ -39,6 +39,20 @@ def commit_from_ref(git_reference)
     description.split('-').last.sub(/^g/, '')
   end
 
+  def tag_from_ref(git_reference)
+    Dir.chdir(repo_cache_dir) do
+      tag = IO.popen(['git', 'describe', '--tags', git_reference], err: [:child, :out]) do |io|
```
grosser: how about just this ?
```
IO.popen(['git', 'describe', '--tags', git_reference]).read
```

app/models/git_repository.rb
```diff
@@ -39,6 +39,20 @@ def commit_from_ref(git_reference)
...
+        io.read.strip
+      end
+
+      if $?.success?
```
grosser: `tag if $?.success?`

grosser:
:+1:

app/models/git_repository.rb
gabetax: The test output covering the nil case gets an error dumped to stderr that I was trying to avoid.

grosser: you can always do
silence_stderr { test }

but either way is fine I guess

app/models/job_execution.rb
```diff
@@ -126,8 +127,9 @@ def execute!(dir)
   def setup!(dir)
     locked = lock_project do
       return false unless @repository.setup!(@executor, dir, @reference)
-      commit = @repository.commit_from_ref(@reference)
-      @job.update_commit!(commit)
+      commit        = @repository.commit_from_ref(@reference)
```
sandlerr: bit of extra whitespace

sandlerr:
:+1:

PR #309, opened 2015-03-19
prevent slow preloading patterns

grosser:
copying the preload check from classic,
nothing in here atm, just making sure we stay fast

@zendesk/runway 

### Risks
 - None

### DIFF
--- a/config/initializers/zz.rb
+++ b/config/initializers/zz.rb
@@ -1,0 +1,11 @@
+if Rails.env.development?
+  # make sure we do not regress into slow startup time by preloading too much
+  Rails.configuration.after_initialize do
+    [
+      ActiveRecord::Base.send(:descendants).map(&:name),
+      ActionController::Base.descendants,
+      (File.basename($0) != "rake" && defined?(Rake) && "rake"),
+      (defined?(Mocha) && "mocha"),
+    ].compact.flatten.each { |c| raise "#{c} should not be loaded" }
+  end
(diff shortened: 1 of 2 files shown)


config/initializers/zz.rb
```diff
@@ -0,0 +1,11 @@
+if Rails.env.development?
+  # make sure we do not regress into slow startup time by preloading to much
+  Rails.configuration.after_initialize do
+    [
+      ActiveRecord::Base.send(:descendants).map(&:name),
+      ActionController::Base.descendants,
+      (File.basename($0) != "rake" && defined?(Rake) && "rake"),
```
zendesk-shender: Isn't the 3rd "rake" conditional obsolete since it'll always be 'true'?

config/initializers/zz.rb
```diff
@@ -0,0 +1,11 @@
+if Rails.env.development?
+  # make sure we do not regress into slow startup time by preloading to much
```
zendesk-shender: mispelled 'too' :)

config/initializers/zz.rb
grosser: it's so that the raise says "rake should not be preloaded" :)

config/initializers/zz.rb
grosser: fixed

zendesk-shender:
:+1:

PR #313, opened 2015-03-23
Make env dashboard dynamically loaded via angularjs for performance reasons

zendesk-shender:
For Performance reasons offload the loading of the Environment dashboards dynamically via javascript.

/cc @zendesk/runway @princemaple 

### References
 - Jira link:

### Risks
 - None


grosser:
is this somehow doing less queries ? / how is it faster ?

zendesk-shender:
It's not 'faster', it's more responsive in that it loads as much as it can, and then lets JS load the rest piece-meal. So instead of taking 4-5secs to load the whole page, it takes .5 secs to load page and then loads the details in the rest of the 4-5 secs.

This is a good first step if we ever incorporate Radar into Samson, as then we can just reload specific projects that have new deploys triggered.

zendesk-shender:
I gave up on trying to create a crazy join to do the work, or create a DB view that might do the join for you.

app/controllers/projects_controller.rb
```diff
@@ -61,6 +61,14 @@ def destroy
     redirect_to admin_projects_path
   end
 
+  def deploy_group_versions
+    before = params[:before] ? Time.parse(params[:before]).to_s(:db) : Time.now.to_s(:db)
```
jonmoter: I would assume that `Project#last_deploy_by_group` would take a `Time` as a parameter, rather than a string. Is there a reason you're transforming them into strings?

zendesk-shender: I left it like that as @grosser wrote the function signature, but yea, looking at it again, I don't see any reason not to change it.

zendesk-shender:
@jonmoter , comment addressed!

jonmoter:
Let's give it a try!  :+1:

PR #315, opened 2015-03-25
Only show successful deploy for timeline.

zendesk-shender:
Ensure only showing successful deploys on the timeline. Also use the existing controller action rather than creating new one.
Increase end-time on timeline to avoid cut-off nodes.
Use created_at timestamp as backup for deploys as older deploys didn't populate started_at.

/cc @zendesk/runway 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/assets/javascripts/controllers/deploy_groups_ctrl.js
+++ b/app/assets/javascripts/controllers/deploy_groups_ctrl.js
@@ -3,11 +3,7 @@ samson.controller('DeployGroupsCtrl', function($scope, $http, $location, $window
   function init() {
     $http.get($location.path() + '.json').success(function(result) {
       result.deploys.forEach(function(deploy) {
-        if (deploy.started_at != undefined) {
-          deploy.start = new Date(deploy.started_at);
-        } else {
-          deploy.start = new Date(deploy.created_at);
-        }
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/controllers/deploy_groups_ctrl.js
```diff
@@ -1,9 +1,13 @@
...
   function init() {
-    $http.get($location.path() + '/deploys.json').success(function(result) {
+    $http.get($location.path() + '.json').success(function(result) {
       result.deploys.forEach(function(deploy) {
-        deploy.start = new Date(deploy.started_at);
+        if (deploy.started_at != undefined) {
+          deploy.start = new Date(deploy.started_at);
+        } else {
+          deploy.start = new Date(deploy.created_at);
+        }
```
jonmoter: I think you could condense that to `deploy.start = new Date(deploy.started_at || deploy.created_at)`.

jonmoter:
Looks good.  :+1:

app/assets/javascripts/controllers/deploy_groups_ctrl.js
zendesk-shender: Ah, good point, thanks @jonmoter

PR #316, opened 2015-03-26
Tidy up deploy group controller to use as_json better.

zendesk-shender:
Tidying up ugly code in controller.

/cc @zendesk/runway 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -118,6 +118,10 @@ class Deploy < ActiveRecord::Base
     joins(:job).where(jobs: { status: 'pending'} ).where("jobs.created_at < ?", threshold)
   end
 
+  def url
+    Rails.application.routes.url_helpers.project_deploy_path(project, self)
+  end
+
   private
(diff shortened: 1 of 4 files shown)


sandlerr:
:+1: although it might be slightly confusing because Rails has both `xyz_path` and `xyz_url` which are slightly different (the path is relative, the url is absolute)

zendesk-shender:
K, might leave for now since I'd have to change all the JS that access that field if I renamed the function.

app/models/deploy.rb
```diff
@@ -118,6 +118,10 @@ def self.expired
     joins(:job).where(jobs: { status: 'pending'} ).where("jobs.created_at < ?", threshold)
   end
 
+  def url
+    Rails.application.routes.url_helpers.project_deploy_path(project, self)
```
steved: Definitely consider presenters/serializers if you need to do this for more things

PR #317, opened 2015-03-26
Regenerate schema.rb file and create missing migrations and fix existing...

zendesk-shender:
Regenerate schema.rb file to allow for less diffs when adding future migrations. Rails 4 changed some of the formatting for it.
I also ran into missing migrations, so I fixed/added those.

/cc @zendesk/runway @samson

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/db/migrate/20150224031239_add_email_settings_to_stages.rb
+++ b/db/migrate/20150224031239_add_email_settings_to_stages.rb
@@ -1,7 +1,7 @@
 class AddEmailSettingsToStages < ActiveRecord::Migration
   def change
     add_column :stages, :email_committers_on_automated_deploy_failure, :boolean, default: false, null: false
-    add_column :stages, :static_emails_on_automated_deploy_failure, :string
+    add_column :stages, :static_emails_on_automated_deploy_failure, :string, limit: 255
     add_column :users, :integration, :boolean, default: false, null: false
     User.reset_column_information
     User.where(name: ["Jenkins", "Semaphore", "Tddium", "Github", "Travis"]).update_all(integration: true)
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,8 +13,8 @@
 
 ActiveRecord::Schema.define(version: 20150312110723) do
 
-  create_table "commands", force: true do |t|
-    t.text     "command",    limit: 16777215
+  create_table "commands", force: :cascade do |t|
+    t.text     "command",    limit: 10485760
     t.datetime "created_at"
     t.datetime "updated_at"
     t.integer  "project_id"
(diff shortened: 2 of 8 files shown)


db/migrate/20150326121012_add_missing_migrations.rb
```diff
@@ -0,0 +1,15 @@
+class AddMissingMigrations < ActiveRecord::Migration
+  def change
+    create_table :flowdock_flows do |t|
```
sandlerr: this should be extracted to a plugin, right?

zendesk-shender: Yes, good point, I found the DB migrations in the plugin directories. I'll have to figure out why it didn't pick those up when regenerating the schema.

zendesk-shender:
Added rule to pickup plugin migrations as well. Thanks @sandlerr

db/schema.rb
```diff
@@ -162,41 +162,41 @@
...
     t.datetime "updated_at"
   end
 
-  add_index "stars", ["user_id", "project_id"], name: "index_stars_on_user_id_and_project_id", unique: true, using: :btree
+  add_index "stars", ["user_id", "project_id"], name: "index_stars_on_user_id_and_project_id", unique: true
 
-  create_table "users", force: true do |t|
-    t.string   "name"
+  create_table "users", force: :cascade do |t|
+    t.string   "name",                           null: false
```
jonmoter: Did you mean to add a non-null constraint to this column? The Travis tests are failing as a result of this.

zendesk-shender: Yea, looks like that was added expicitly in the EnforceLockUserId migration. I'll update the unit tests to reflect this as it looks like this should be a required field anyway.

zendesk-shender: Another good reason for this PR! :)
Getting all the migrations in sync and stop manually editing the schema.rb.

zendesk-shender:
ping @zendesk/runway @zendesk/samson

jonmoter:
:+1:

db/migrate/20150224031239_add_email_settings_to_stages.rb
```diff
@@ -1,7 +1,7 @@
 class AddEmailSettingsToStages < ActiveRecord::Migration
   def change
     add_column :stages, :email_committers_on_automated_deploy_failure, :boolean, default: false, null: false
-    add_column :stages, :static_emails_on_automated_deploy_failure, :string
+    add_column :stages, :static_emails_on_automated_deploy_failure, :string, limit: 255
```
sandlerr: I guess its conceivable for this not to be enough? Although maybe by that point its better to setup a mailing list.

sandlerr:
:+1:

PR #318, opened 2015-03-26
Fix preload test when executing rake tasks.

zendesk-shender:
When executing certain rake tasks that initialize the rails environment, it was throwing an error due to the preload check in /initializers/zz.rb

```
$  bundle exec rake db:setup
rake aborted!
false should not be loaded
/Users/shender/Code/zendesk/samson/config/initializers/zz.rb:14:in `block (2 levels) in <top (required)>'
/Users/shender/Code/zendesk/samson/config/initializers/zz.rb:14:in `each'
/Users/shender/Code/zendesk/samson/config/initializers/zz.rb:14:in `block in <top (required)>'
/Users/shender/Code/zendesk/samson/config/environment.rb:5:in `<top (required)>'
Tasks: TOP => db:setup => db:schema:load_if_ruby => environment
(See full trace by running task with --trace)
```

/cc @zendesk/runway @samson

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/config/initializers/zz.rb
+++ b/config/initializers/zz.rb
@@ -4,7 +4,7 @@ if Rails.env.development? && !ENV['SERVER_MODE']
     [
       ActiveRecord::Base.send(:descendants).map(&:name),
       ActionController::Base.descendants,
-      (File.basename($0) != "rake" && defined?(Rake) && "rake"),
+      (File.basename($0) != "rake" ? (defined?(Rake) && "rake") : nil),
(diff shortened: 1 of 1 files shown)


sandlerr:
so the logic is: if you're not running a rake task, make sure rake hasn't been loaded? :+1:

zendesk-shender:
@sandlerr, yup, I think that was the original intent.

PR #321, opened 2015-03-27
better error handling for GitRepository#commit_from_ref

jonmoter:
/cc @zendesk/samson 

### References
 - Airbrake error: https://zendesk.airbrake.io/projects/95346/groups/1383597137614811072

### Risks
 - Minimal risk, just improved error handling.

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -30,13 +30,10 @@ class GitRepository
   end
 
   def commit_from_ref(git_reference)
-    description = Dir.chdir(repo_cache_dir) do
-      IO.popen(['git', 'describe', '--long', '--tags', '--all', git_reference]) do |io|
-        io.read.strip
-      end
-    end
+    cmd = "git describe --long --tags --all #{git_reference}"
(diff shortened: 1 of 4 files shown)


app/models/git_repository.rb
```diff
@@ -30,13 +30,10 @@ def update!(executor: TerminalExecutor.new(StringIO.new))
   end
 
   def commit_from_ref(git_reference)
-    description = Dir.chdir(repo_cache_dir) do
-      IO.popen(['git', 'describe', '--long', '--tags', '--all', git_reference]) do |io|
-        io.read.strip
-      end
-    end
+    cmd = "git describe --long --tags --all #{git_reference}"
```
gabetax: Doesn't git_reference come from user input such that we want to shell escape it still?

jonmoter: Hm, fair enough. It didn't occur to me it was using `popen` to handle the shell escaping. I was just trying to simplify the code.  Will revert.

jonmoter:
Old PR.  Closing.

PR #323, opened 2015-03-30
Add Buildkite integration

mariovisic:
Allow samson to receive webhooks from the [Buildkite CI service](https://buildkite.com/) to notify of successful builds that can then kick off a deploy.

Looks like this: 
buildkite_hook

FYI: Some of the tests seem to be failing, this is due to the github API not returning an 'up' status due to the ongoing DDOS attacks.


jwswj:
:+1: 
/cc @zendesk/samson

jonmoter:
At some point we should probably re-implement some of these integrations using the plugin architecture. But this looks good to me.  :+1:

jwswj:
Thanks @mariovisic

PR #324, opened 2015-03-30
Buddy request in flowdock

fneves:
This PR adds a plugin that enables a user to send a buddy request to Flowdock from the UI.
This was previously submitted in a [PR#229](https://github.com/zendesk/samson/pull/229) and the behaviour is the same but it is now extracted as a plugin. I've bundled it with the current Flowdock plugin.

This PR adds the support for plugins with custom scripts and css to be included. It also adds support for angular tests in plugins.

output

/cc @zendesk/samson @grosser @pswadi-zendesk 

### References
 - Jira link: [SAMSON-106](https://zendesk.atlassian.net/browse/SAMSON-106)

### Risks
 - None


plugins/flowdock/lib/samson_flowdock/samson_plugin.rb
```diff
@@ -20,22 +21,47 @@ def flowdock_tokens
...
     end
+
+    def enabled_flows_names
+      flowdock_flows.enabled.map(&:name)
+    end
+  end
+end
+
+Samson::Hooks.callback :deploy_defined do
+  Deploy.class_eval do
```
sandlerr: stupid question: couldn't you do this without the hook and just define methods on the class like normal? (`class Deploy ...`)

plugins/flowdock/lib/samson_flowdock/flowdock_service.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    def chat_flow
+      @flow ||= Flowdock::Flow.new(api_token: tokens, external_user_name: 'Samson')
+    end
+
+    def tokens
+      @deploy.stage.flowdock_tokens
+    end
+
+    def url_helpers
```
sandlerr: having this method on a model seems a little unusual, although I can see why its needed. Would you mind replacing it by the more specific route you need? So instead of `def url_helpers` it would be `def project_deploy_url`?

app/views/layouts/application.html.erb
```diff
@@ -3,7 +3,15 @@
   <head>
     <title><%= content_for(:page_title) || "Samson" %></title>
     <%= stylesheet_link_tag    "application", media: "all" %>
+    <% Samson::Hooks.plugins.map(&:stylesheets).flatten.each do |css| %>
+      <%= stylesheet_link_tag asset_path(css) %>
+    <% end %>
+
     <%= javascript_include_tag "application" %>
+    <% Samson::Hooks.plugins.map(&:javascripts).flatten.each do |javascript| %>
```
sandlerr: should we add these to sprockets? Not really a big deal, might be nice to have a single `application.js` and `application.css`

plugins/flowdock/lib/samson_flowdock/samson_plugin.rb
fneves: When using ```class Deploy; ....; end``` syntax it always creates a class, or redefines it if it exists, failing silently if you did a typo.

When we use ```class_eval``` it assumes a class exists, and if it doesn't this will cause ```const_missing``` to be called on the parent module which usually throws an error.
Rails overrides ```const_missing``` and autoloads the class, throwing an error only if it cannot find it in the load path.

app/views/layouts/application.html.erb
fneves: I can't really do that since the javascripts I would have to include are now dynamic. ```assets:precompile``` happens before erb translates to clean js.

To be consistent I did the css in the same way.

plugins/flowdock/lib/samson_flowdock/samson_plugin.rb
sandlerr: @fneves, I'm pretty sure this extends classes - copy and paste this into `irb`:

```ruby
class Oracle
  def are_you_there?
    p 'yes'
  end
end

class Oracle
  def are_you_here?
    p 'yes'
  end
end

oracle = Oracle.new
oracle.are_you_there?
oracle.are_you_here?
```

Not 100% sure how it will work with inheritance but I've seen this pattern used

fneves: It does.
But if the class does not exist it will create it. And I don't want that. I want to be sure I am redefining it. ```class_eval``` guarantees me that, since it will load the class first.

app/views/layouts/application.html.erb
sandlerr: I can't see any `.erb.js` in this PR, shouldn't they be [here](https://github.com/zendesk/samson/tree/fneves/buddy_request_plugin/plugins/flowdock/app/assets/javascripts) or [here](https://github.com/zendesk/samson/tree/fneves/buddy_request_plugin/plugins/flowdock/app/views/samson_flowdock)?

plugins/flowdock/lib/samson_flowdock/samson_plugin.rb
sandlerr: OK. I'm not sure if it is worth the complexity since `Deploy` will always be defined when the Rails application loads. If we are going to have this hook we should probably have it when all models have been required rather than special casing `Deploy`.

sandlerr: If someone renames the `Deploy` class without changing it here then the plugin should have a failing test.

plugins/flowdock/lib/samson_flowdock/samson_plugin.rb
```diff
@@ -20,22 +21,47 @@ def flowdock_tokens
...
+      ":pray: #{user_tag(user)} is requesting approval for deploy #{deploy_url}"
+    end
+
+    private
+
+    def user_tag(user)
+      "@#{user.email.match(/(.*)@/)[1]}"
+    end
+
+    def url_helpers
```
sandlerr: this one is probably the same as the other one

app/views/layouts/application.html.erb
fneves: I haven't added any. I tried to ```erb``` [application.js](https://github.com/zendesk/samson/blob/fneves/buddy_request_plugin/app/assets/javascripts/application.js), since I believe that is should be Samson that will knowledge of which plugins to include. It didn't work since the assets will be precompiled before the erb translation takes effect, which would not include the javascripts from the plugin.
Is there a different way to add them to sprockets that would work?

zendesk-shender:
This plugin brings up the question of isolating plugins within IFrames, as this one included underscore.js which is already included in Samson.
So while the developer can look out for this, it would be a maintenance hassle to have Samson add a JS library and then all the plugin maintainers would have to remove their versions of that JS library.

fneves:
@zendesk-shender Removed the underscore as it was already being included. That was discussed during [here](https://github.com/zendesk/samson/pull/259). Maybe its time to revisit it again and rethink our solution before we start having too many plugins. @grosser what do you think?

@sandlerr I've refactored things a bit. Now we don't need the ```stage_defined``` and ```deploy_defined``` callbacks. Existent models are going to be redefined using the decorators which will be loaded only when the app is ready and the models are there. I've used class_eval, but I can change them into concerns if you think its best. 
I also changed the way the plugin lib directory is loaded so that we don't need to explicitly load the lib directory at the plugin itself. ```Samson::Hooks``` take care of that now.

app/views/layouts/application.html.erb
sandlerr: Something like this may be helpful: http://edgeguides.rubyonrails.org/engines.html#separate-assets-precompiling

There must be a standard way to adds assets from a rails engine - they may even be precompiled in by default?

sandlerr: there is already an envar afaik to load specific plugins only, this should also be applied when compiling assets.

sandlerr:
@fneves thanks for removing the callbacks. Would you mind if I make similar changes to the Slack integration introduced in #328 on your branch? I'll need to rebase.

fneves:
@sandlerr Not sure I understood, you want to put the Slack changes in my branch? or you want to copy the changes/rebase from my branch?

fneves:
@sandlerr assets are now being precompiled.
Also added two helpers that will help cleaning the application.rb
Merge master into my branch due to conflicts and refactored the slack integration to stop using the callbacks.

All good?

plugins/slack/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,42 @@
+Stage.class_eval do
+
+  cattr_reader(:slack_channels_cache_key) { 'slack-channels-list' }
```
sandlerr: This could also be a constant

lib/samson/hooks.rb
```diff
@@ -63,6 +110,32 @@ def render_views(name, view, *args)
...
+          each(&:require).
+          each(&:precompile_assets).
+          each(&:add_lib_path).
+          each(&:add_decorators)
+      end
+
+      def render_javascripts(view)
+        Samson::Hooks.plugins.map(&:javascripts).flatten.each do |script|
+          view.instance_exec do
+            concat view.javascript_include_tag(view.asset_path(script))
```
sandlerr: I'm still a little confused here. This will pull out all the files under `PLUGIN/assets/javascripts/*.js` and create URLs like `samson.com/flowdock_plugin.js` but are they namespaced under the plugin name in case of a filename conflict?
Also, I don't think the filenames in `app/assets` necessarily map to URLs - many assets can be compiled into one via application.js / css and also Rails fingerprinting.

That said, I am probably just being too pedantic and if you are confident that this is the way to go then let's do it :+1:

sandlerr:
Thanks for sticking with this and refactoring the Slack plugin!

lib/samson/hooks.rb
fneves: Actually they aren't name spaced. This will generate the URLs like ```samson.com/assets/flowdock_plugin.js``` which are of course non name spaced. I can solve this by putting the javascript inside a ```plugins``` directory. 
I don't know (apart from performance issues) what would be the advantage of compiling everything into a single application.js, the conflicts would still occur, since variables would redefine themselves (and we are not even handling version conflicts). 

I think a definite solution for this problem (asset conflicts (with names and versioning) will involve a lot more work than it is done here. I believe that sandboxing each plugin view in an IFrame with just the core scripts and their specific ones is the way to go.

This way each plugin would live on its own world, and we would handle the different javascript versions without clashes and name conflicts. 

But I really would like to have @grosser input here.

plugins/flowdock/app/views/samson_flowdock/_notify_buddy_box.html.erb
```diff
@@ -0,0 +1,29 @@
+<% if @deploy.waiting_for_buddy? && @deploy.stage.send_flowdock_notifications? %>
+  <buddy-request-box default-buddy-request-message = "<%= @deploy.default_flowdock_message(current_user) %>"
```
zendesk-shender: There are generally no spaces around '=' signs in HTML

plugins/flowdock/app/decorators/deploy_decorator.rb
```diff
@@ -0,0 +1,15 @@
+Deploy.class_eval do
+  def default_flowdock_message(user)
+    ":pray: #{user_tag(user)} is requesting approval for deploy #{deploy_url}"
+  end
+
+  def deploy_url
+    Rails.application.routes.url_helpers.project_deploy_url(self.stage.project, self)
+  end
```
zendesk-shender: There is already a deploy 'url' method inside that class.

plugins/flowdock/app/models/flowdock_notification.rb
```diff
@@ -1,51 +1,27 @@
 require 'flowdock'
 
 class FlowdockNotification
-  def initialize(stage, deploy)
-    @stage, @deploy = stage, deploy
+  def initialize(deploy)
+    @deploy = deploy
+    @stage = @deploy.stage
```
zendesk-shender: You aren't using the @stage variable outside this method.

plugins/flowdock/app/models/flowdock_notification.rb
```diff
@@ -1,51 +1,27 @@
...
-    end
-
-    buddy_request_content = text + " deploy " + url_helpers.project_deploy_url(@project, @deploy)
-    chat_flow.push_to_chat(:content => buddy_request_content, :tags => ["buddy-request", "completed"])
+    buddy_request_content = buddy_request_completed_message(approved, buddy)
+    flowdock_service.notify_chat(buddy_request_content, %w(buddy-request completed))
   end
 
   def deliver
     subject = "[#{@project.name}] #{@deploy.summary}"
```
zendesk-shender: Only using the @project variable once, so maybe just replace with ```@deploy.stage.project```

plugins/flowdock/lib/samson_flowdock/flowdock_service.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    def inbox_flow
+      @flow ||= Flowdock::Flow.new(
+        api_token: tokens,
+        source: "samson",
+        from: { name: user.name, address: user.email }
+      )
+    end
+
+    def project
```
zendesk-shender: Ehh... not used?

plugins/flowdock/lib/samson_flowdock/flowdock_service.rb
```diff
@@ -0,0 +1,64 @@
...
+      @project ||= stage.project
+    end
+
+    def stage
+      @stage ||= @deploy.stage
+    end
+
+    def user
+      @user ||= @deploy.user
+    end
```
zendesk-shender: This seems needless to create these functions to save you typing ```deploy```, but I'm not willing to argue over it! :)

plugins/flowdock/samson_flowdock.gemspec
```diff
@@ -1,6 +1,8 @@
...
-  s.authors = ["Michael Grosser"]
-  s.email = "michael@grosser.it"
-  s.add_runtime_dependency "flowdock", "~> 0.3.1"
+Gem::Specification.new 'samson_flowdock', '0.0.0' do |s|
+  s.summary = 'Samson flowdock integration'
+  s.authors = ['Michael Grosser']
+  s.email = 'michael@grosser.it'
+  s.add_runtime_dependency 'flowdock', '~> 0.5.0'
+  s.files = Dir['{app,config,db,lib}/**/*']
+  s.test_files = Dir['test/**/*.rb']
```
zendesk-shender: Looks like this will be deprecated soon, and it looks like we aren't using it directly:
https://github.com/rubygems/rubygems/issues/735

plugins/flowdock/app/assets/javascripts/flowdock_plugin.js
```diff
@@ -0,0 +1,112 @@
...
+    return $http.post('/flowdock/notify', { deploy_id: deploy, message: message });
+  };
+
+  return {
+    users: self.users,
+    buddyRequest: self.buddyRequest
+  }
+}]);
+
+samson.controller('BuddyNotificationsCtrl', ['$scope','$rootScope', 'Flowdock', 'Mentionbox',
```
zendesk-shender: We don't need the array of dependencies syntax since we use ngannotate-rails.

plugins/flowdock/app/assets/javascripts/flowdock_plugin.js
```diff
@@ -0,0 +1,112 @@
...
+  function($scope, $rootScope, flowdock, mentionsBox) {
```
zendesk-shender: Where is mentionsBox declared? I only see a mentionBox factory.

fneves: Its not a typo. With angular (without ngannotate-rails) using dependency injection I can name whatever I want to that variable, as long as I reference it correctly from the dependency array. From what I am understanding we started using  ngannotate-rails and now I just need to do:
``` 
samson.controller('BuddyNotificationsCtrl', function($scope, $rootScope, flowdock, mentionsBox) {
```
Is it?

Not sure I really like this magic.

fneves: Isn't that starting to couple all our Angular stuff too much to Rails?

fneves:
@zendesk-shender another look?

plugins/flowdock/app/assets/javascripts/flowdock_plugin.js
```diff
@@ -0,0 +1,111 @@
...
+
+  return self;
+});
+
+samson.factory('Flowdock', function ($rootScope, $http) {
+  var self = this;
+
+  self.users = function () {
+    var users = [];
+    var promise = $http.get('/flowdock/users');
```
zendesk-shender: You can just chain these functions rather then saving to 'promise'

zendesk-shender:
I'm happy enough to :+1: 
Don't know if you want to wait for @grosser

sandlerr:
resolved conflicts in #343

fneves:
@grosser What do you think about the js/css includes the way they are done here?

app/views/deploys/show.html.erb
```diff
@@ -33,6 +33,7 @@
 <section class="clearfix tabs">
   <div class="tab-content deploy-details">
     <div class="tab-pane active" id="output" data-stream-url="<%= stream_path(@deploy.job) %>" data-desktop-notify="<%= current_user.desktop_notify? %>">
+      <%= Samson::Hooks.render_views(:deploy_view, self, deploy: @deploy, project: @project) %>
```
grosser: since they are @ variables we don't need to pass them ... might still be nice to do just as an api ...

lib/samson/hooks.rb
```diff
@@ -25,6 +25,53 @@ def active?
...
+        find_assets(basedir, exts).map { |script| script.relative_path_from(Pathname.new(basedir)) }.map(&:to_s)
+      end
+
+      def find_assets(basedir, exts)
+        exts.map do |ext|
+          Dir.glob("#{basedir}/**/*.#{ext}").map { |asset| Pathname.new(asset) }
+        end.flatten
+      end
+
+      def assets_root(type)
```
grosser: can this just do `engine.paths['app/assets/javascripts']` and return if it does not exist ? (avoid include logic / checking all paths for existence)

lib/samson/hooks.rb
```diff
@@ -25,6 +25,53 @@ def active?
...
+
+      def add_decorators
+        engine.config.after_initialize do
+          Dir.glob(engine.config.root.join('app/decorators/**/*_decorator.rb')).each do |c|
+            require_dependency(c)
+          end
+        end
+      end
+
+      def precompile_assets
```
grosser: the method  is not really 'precompiling' the assets, so maybe something like `add_assets_to_precompile` would be better

lib/samson/hooks.rb
grosser: the more plugins we get the more sandboxing will be nice ... but not sure how much work that will be, also means we lose a lot of flexability

lib/samson/hooks.rb
```diff
@@ -63,6 +110,32 @@ def render_views(name, view, *args)
...
+        Samson::Hooks.plugins.
+          each(&:require).
+          each(&:precompile_assets).
+          each(&:add_lib_path).
+          each(&:add_decorators)
+      end
+
+      def render_javascripts(view)
+        Samson::Hooks.plugins.map(&:javascripts).flatten.each do |script|
+          view.instance_exec do
```
grosser: can we just call view.concat instead of doing instance_exec ?

lib/samson/hooks.rb
```diff
@@ -63,6 +110,32 @@ def render_views(name, view, *args)
...
+            concat view.javascript_include_tag(view.asset_path(script))
+          end
+        end
+        nil
+      end
+
+      def render_stylesheets(view)
+        Samson::Hooks.plugins.map(&:stylesheets).flatten.each do |stylesheet|
```
grosser: can be each since we don't use the return value

lib/samson/hooks.rb
```diff
@@ -75,5 +148,8 @@ def hooks(name)
 
 Dir["plugins/*/lib"].each { |f| $LOAD_PATH << f } # treat included plugins like gems
 
-Samson::Hooks.plugins.
-  each(&:require)
+Samson::Hooks.plugin_setup
+
+# This next line is just to make rails reload the decorators whenever we change code inside a plugin.
+# Without it we would have to restart the server every code change.
+ActionDispatch::Reloader.to_prepare { Samson::Hooks.plugins.map(&:add_decorators) }
```
grosser: funky :D

plugins/flowdock/app/assets/javascripts/flowdock_plugin.js
```diff
@@ -0,0 +1,108 @@
+samson.factory('Mentionbox', function ($rootScope, Flowdock) {
```
grosser: should this be namespaced to FlowDock::Mentionbox or something like this ?

plugins/flowdock/app/assets/javascripts/flowdock_plugin.js
```diff
@@ -0,0 +1,108 @@
...
+  };
+
+  return {
+    users: self.users,
+    buddyRequest: self.buddyRequest
+  }
+});
+
+samson.controller('BuddyNotificationsCtrl', function($scope, $rootScope, Flowdock, Mentionbox) {
+    $scope.title = 'Request a buddy!';
```
grosser: 3-space indent ?

plugins/flowdock/app/controllers/flowdock_controller.rb
```diff
@@ -0,0 +1,29 @@
+class FlowdockController < ApplicationController
+  before_action :find_deploy, only: [:notify]
+
+  rescue_from ActiveRecord::RecordNotFound do
```
grosser: this looks like it should go into application controller

plugins/flowdock/app/controllers/flowdock_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def notify
+    FlowdockNotification.new(@deploy).buddy_request(params[:message])
+    render json: { message: 'Successfully sent a buddy request to the flows!' }
+  end
+
+  private
+
+  def find_deploy
+    @deploy = Deploy.find(params[:deploy_id])
```
grosser: only used once, might as well inline this into notify and remove the before_filter

lib/samson/hooks.rb
```diff
@@ -25,6 +25,53 @@ def active?
...
         super @path
       end
+
+      def add_lib_path
+        engine.config.autoload_paths += Dir["#{engine.config.root}/lib/**/"]
+      end
+
+      def add_decorators
+        engine.config.after_initialize do
+          Dir.glob(engine.config.root.join('app/decorators/**/*_decorator.rb')).each do |c|
```
grosser: this will load all models and slow down the boot, if we keep it lazy deploy.rb -> calls `add_decorators 'deploy'` -> 'require_dependency' so we can avoid that (should also not need any reload hacks since require_dependency should be reloaded)

plugins/flowdock/db/migrate/20150327144239_add_flowdock_checkbox_to_flowdock_flows.rb
```diff
@@ -0,0 +1,5 @@
+class AddFlowdockCheckboxToFlowdockFlows < ActiveRecord::Migration
+  def change
+    add_column :flowdock_flows, :enabled, :boolean, default: false
```
grosser: this will disable all existing flows ? -> add with default true and then change to default false

plugins/flowdock/samson_flowdock.gemspec
```diff
@@ -1,6 +1,7 @@
-Gem::Specification.new "samson_flowdock", "0.0.0" do |s|
-  s.summary = "Samson flowdock integration"
-  s.authors = ["Michael Grosser"]
-  s.email = "michael@grosser.it"
-  s.add_runtime_dependency "flowdock", "~> 0.3.1"
+Gem::Specification.new 'samson_flowdock', '0.0.0' do |s|
+  s.summary = 'Samson flowdock integration'
+  s.authors = ['Michael Grosser']
+  s.email = 'michael@grosser.it'
```
grosser: I guess we can also remove personal info from here ... this was just me copy-pasting a local gemspec I had ...

plugins/flowdock/test/controllers/flowdock_controller_test.rb
```diff
@@ -0,0 +1,58 @@
...
+        fetched_users = JSON.parse(response.body)['users']
+        fetched_users.size.must_equal(3)
+        fetched_users.each do |user|
+          %w(id name avatar type).must_equal(user.keys)
+        end
+      end
+    end
+  end
+
+  as_a_viewer do
```
grosser: 2 as_a_viewer blocks ?

plugins/flowdock/test/controllers/flowdock_controller_test.rb
```diff
@@ -0,0 +1,58 @@
...
+describe FlowdockController do
+  let(:token) { 'asdkjh21s' }
+
+  before do
+    ENV['FLOWDOCK_API_TOKEN'] = token
+  end
+
+  as_a_viewer do
+    describe 'users' do
+
```
grosser: unnecessary newline

plugins/flowdock/test/models/stage_test.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  subject { stages(:test_staging) }
+
+  before do
+    subject.notify_email_address = "test@test.ttt"
+    subject.flowdock_flows = [FlowdockFlow.new(name: "test", token: "abcxyz", stage_id: subject.id, enabled: false)]
+    subject.save
+  end
+
+  describe 'flowdock notifications' do
```
grosser: I usually do `describe "#send_flowdock_notifications?"` so it's easy to see which block tests what

lib/samson/hooks.rb
fneves: If I use each, what I pass to ```flatten``` are the plugins objects and what I really what are the ```stylesheets``` (unless I am not interpreting correctly your comment).

app/views/deploys/show.html.erb
fneves: I like to have them there to make it clear what is sent to the views on this hook.

app/models/deploy.rb
```diff
@@ -119,7 +119,7 @@ def self.expired
   end
 
   def url
-    Rails.application.routes.url_helpers.project_deploy_path(project, self)
+    Rails.application.routes.url_helpers.project_deploy_url(project, self)
```
grosser: an object should not know it's url, that's the job of the helpers/views

config/environments/test.rb
```diff
@@ -54,3 +54,4 @@
 ENV['DEFAULT_URL'] = 'http://www.test-url.com'
 ENV['DATADOG_API_KEY'] = 'dapikey'
 ENV['DATADOG_APPLICATION_KEY'] = 'dappkey'
+Rails.application.routes.default_url_options[:host]= 'test.host'
```
grosser: this will also not be necessary when deploy no longer knows it's url

db/schema.rb
```diff
@@ -70,6 +70,7 @@
     t.integer  "stage_id",   null: false
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.boolean  "enabled",    default: false
```
grosser: no migration ?

fneves: not finished yet :)

app/models/deploy.rb
fneves: Yeah... I also find it non natural. But I needed to display the url on the Flowdock notification and there was this code here displaying the path. I can just move my code out of the model and into the FlowdockNotification as it was. This will allow me to remove the line in (test.rb#L57)[https://github.com/zendesk/samson/blob/fneves/buddy_request_plugin/config/environments/test.rb#L57]

Was just trying to reuse some code that already existed.

grosser: hmmm yeah that url helper thing might not be avoidable for
mails/notifications ... maybe they can do `.url_for(deploy)`

fneves:
@zendesk-shender @grosser simplified the assets inclusion in the layout. From now on we just include ```/app/assets/javascripts/#{plugin_name}/application.js``` and ```/app/assets/javascripts/#{plugin_name}/application.css```. The plugin developer can include any other js or css by using the ```require``` directive available in Rails (or any other tool that bundles it into an application.js).

Decorators are being lazy loaded. There is a new event ```:model_defined``` that will be fired when an ActiveRecord model is loaded. There will be a callback for this event that will be responsible to load the appropriate model decorators (for all active plugins).

I don't like the idea of firing the ```:model_defined``` at the end of each class definition because it feels like I'm repeating myself over and over again. Also I think that if someone added a new model to Samson there would be a chance of forgetting to fire that event. I just fire the event every time I detect that some class extends ```ActiveRecord::Base```.

db/schema.rb
```diff
@@ -70,6 +70,7 @@
     t.integer  "stage_id",   null: false
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.boolean  "enabled",    default: true
```
grosser: migration ?

lib/samson/hooks.rb
```diff
@@ -30,6 +32,24 @@ def add_migrations
         migrations = File.join(@folder, "db/migrate")
         Rails.application.config.paths["db/migrate"] << migrations if Dir.exist?(migrations)
       end
+
+      def add_lib_path
+        engine.config.autoload_paths += Dir["#{engine.config.root}/lib/**/"]
+      end
+
+      def add_decorator(klass_name)
+        return if klass_name.nil?
```
grosser: did that actually happen ?

lib/samson/hooks.rb
```diff
@@ -30,6 +32,24 @@ def add_migrations
...
+        decorator_file = engine.config.root.join("app/decorators/#{klass_name.downcase}_decorator.rb")
```
grosser: I'm thinking how about we just let people put duplicate models into app/models and load them ... a little more magical but also kind of conveniant (root app/models is higher up in the load_path anyway)

lib/samson/hooks.rb
```diff
@@ -80,6 +124,16 @@ def hooks(name)
 
 Dir["plugins/*/lib"].each { |f| $LOAD_PATH << f } # treat included plugins like gems
 
-Samson::Hooks.plugins.
-  each(&:require).
-  each(&:add_migrations)
+module LoadDecorators
```
grosser: maybe `Samson::LoadDecorators`

lib/samson/hooks.rb
```diff
@@ -80,6 +124,16 @@ def hooks(name)
...
+  def inherited(subclass)
+    Samson::Hooks.fire(:model_defined, subclass)
+    super
+  end
+end
+
+Samson::Hooks.callback :model_defined do |model_class|
```
grosser: instead of iterating over all plugins every time and doing a bunch of file.exist checks, maybe only pick the plugins that opted in via

```
Samson::Hooks.callback :model_defines, "Stage" -> # load plugin/app/models/stage.rb
```

plugins/flowdock/app/assets/javascripts/flowdock/jquery_mentions_input/jquery.mentionsInput.js
```diff
@@ -0,0 +1,564 @@
+/*
```
grosser: `gem 'rails-assets-jquery-mentions-input'`

plugins/flowdock/app/controllers/flowdock_controller.rb
```diff
@@ -0,0 +1,24 @@
+class FlowdockController < ApplicationController
+  rescue_from ActiveRecord::RecordNotFound do |error|
```
grosser: would be nice to have a global json 404, but that's a todo for later :)

plugins/flowdock/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,21 @@
+Stage.class_eval do
```
grosser: you can also just to `class Stage` -> constants etc scoping works as expected

db/schema.rb
grosser: found it :)

plugins/flowdock/app/assets/javascripts/flowdock/jquery_mentions_input/jquery.mentionsInput.js
grosser: I guess we have to declare the source global for that to work ...

grosser: should be fine ... there can't be conflicts ...

plugins/flowdock/test/angular/buddy_request_box_spec.js
```diff
@@ -0,0 +1,132 @@
+describe('Buddy requests notifications', function () {
```
grosser: :heart: js tests :)

plugins/flowdock/test/controllers/flowdock_controller_test.rb
```diff
@@ -0,0 +1,54 @@
+require_relative '../test_helper'
```
grosser: don't need the extra jquery in test/angular ?

plugins/flowdock/test/controllers/flowdock_controller_test.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    describe 'buddy request notifications' do
+      before do
+        FlowdockNotification.any_instance.expects(:buddy_request).once
+        deploy_id = deploys(:succeeded_test).id
+        post :notify,  deploy_id: deploy_id, message: 'Test'
+      end
+
+      it 'should send a buddy request' do
```
grosser: can leave out the should: `it 'sends a buddy request' do`

plugins/flowdock/test/models/stage_test.rb
```diff
@@ -0,0 +1,24 @@
...
+      subject.send_flowdock_notifications?.must_equal(false)
+    end
+
+    it 'returns that there are flows with enabled notifications' do
+      subject.flowdock_flows.first.update!(enabled: true)
+      subject.send_flowdock_notifications?.must_equal(true)
+    end
+  end
+end
+
```
grosser: extra newlines

plugins/flowdock/app/controllers/flowdock_controller.rb
fneves: Later in a another PR :) It would also be nice to have someone design a fancy 404 page for the html 404s.

plugins/flowdock/app/assets/javascripts/flowdock/jquery_mentions_input/jquery.mentionsInput.js
fneves: Can't do it at the moment. I had to fork the jquery-mentions-input to add support for a default value on the textarea. Submitted the [PR#131](https://github.com/podio/jquery-mentions-input/pull/131). When its accepted then we can switch.

plugins/flowdock/test/controllers/flowdock_controller_test.rb
fneves: Now that we have the gem we don't :)

lib/samson/hooks.rb
fneves: Strangely yes. When Kaminari is loading, there is a class with an empty name.

lib/samson/hooks.rb
fneves: So use the same mechanism (the ```model_defined``` callback) apart from the fact that is going to be defined in models we are going to re-open a class?

plugins/flowdock/app/decorators/stage_decorator.rb
fneves: I know, but I like to see a class eval on this cases due to 2 reasons.
First, when I am re-opening a class that is not defined by me (on this case the plugin) I like to use ```class_eval``` as this will clearly state *this model is defined elsewhere*.
Second, if anything happens to the model (deleted/removed) ```class_eval``` will raise an error, which will allow me to detect that I need to either rename the variable or the code no longer makes so exist.

lib/samson/hooks.rb
fneves: Preloading a decorators index (indexed by classname) when we run the plugin setup. Decorators will still be loaded on the ```model_defined``` callback.

fneves:
@grosser I'm now pre loading the decorators (only the paths) into an indexed based data structure. They will be required from there now, avoiding the file.exists and the iteration over the plugins (over and over again). I prefer it this way against changing the ```callback``` api. This way is also consistent with the rest of what we are doing (like loading migrations, etc...)

Not moving moving the decorators into models since I prefer to separate the two concerns, defining a plugin model and re-defining some bits from core models.

Let me know what you think.

plugins/flowdock/app/decorators/stage_decorator.rb
grosser: yeah true, will make the plugin fail nicely :)

FYI alternative approach we used before to get constants:

```
Stage
class Stage
   ... use constants ...
end
```

plugins/flowdock/app/helpers/flowdock_helper.rb
```diff
@@ -0,0 +1,9 @@
+module FlowdockHelper
+  def default_flowdock_message(deploy)
+    ":pray: #{user_tag(deploy.user)} is requesting approval for deploy #{project_deploy_url(deploy.project, deploy)}"
+  end
+
+  def user_tag(user)
+    "@#{user.email.match(/(.*)@/)[1]}"
```
grosser: possibly easier to read:
`"@#{user.email.split("@").first}"`

plugins/flowdock/app/models/flowdock_notification.rb
```diff
@@ -1,68 +1,48 @@
...
-      content: content,
-      tags: ["deploy", @stage.name.downcase],
-      link: url
-    )
-  rescue Flowdock::Flow::ApiError
-    # Invalid token or something.
+    subject = "[#{project.name}] #{@deploy.summary}"
+    flowdock_service.notify_inbox(subject, content, deploy_url)
+  rescue Flowdock::ApiError => e
+    Rails.logger.error("Could not deliver flowdock message: #{e.message}")
```
grosser: also send to airbrake ?

fneves:
@grosser All good now? :)

fneves:
By the way, my jquery-mentions-plugin PR got merged. Did we generate the assets gem or was it the author?

grosser:
I think it is auto-generated by pushing to bower

lib/samson/hooks.rb
```diff
@@ -68,8 +95,45 @@ def render_views(name, view, *args)
         nil
       end
 
+      def decorator(klass_name, path)
```
grosser: you can reuse hooks for that, same as `def view` does

lib/samson/hooks.rb
```diff
@@ -80,6 +144,16 @@ def hooks(name)
 
 Dir["plugins/*/lib"].each { |f| $LOAD_PATH << f } # treat included plugins like gems
 
-Samson::Hooks.plugins.
-  each(&:require).
-  each(&:add_migrations)
+module Samson::LoadDecorators
+  def inherited(subclass)
+    Samson::Hooks.fire(:model_defined, subclass)
```
grosser: it's simpler if the interface is symmetric, setup takes a name, so fire should too

grosser:
:+1:

fneves:
@grosser Symmetric api now in place and reusing hooks. :) 
Will merge this now :)

PR #326, opened 2015-03-31
fallback to commit ref if repo has no tags

gabetax:
If a git repository has no tags, previously we would `export TAGS=`, which was causing issues for capistrano scripts that acted on ENV['TAG'] without an empty string check. This falls back to the commit reference for these uncommon repositories.

/cc @zendesk/runway @steved

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -105,7 +105,7 @@ class JobExecution
       "export DEPLOYER_EMAIL=#{@job.user.email.shellescape}",
       "export DEPLOYER_NAME=#{@job.user.name.shellescape}",
       "export REVISION=#{@reference.shellescape}",
-      "export TAG=#{@job.tag.to_s.shellescape}",
+      "export TAG=#{(@job.tag || @job.commit).to_s.shellescape}",
       "export CACHE_DIR=#{artifact_cache_dir}",
       "cd #{dir}",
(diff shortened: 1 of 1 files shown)


app/models/git_repository.rb
```diff
@@ -45,7 +45,12 @@ def tag_from_ref(git_reference)
         io.read.strip
       end
 
-      tag if $?.success?
+      if $?.success?
+        tag
+      else
+        # no tags present
+        commit_from_ref(git_reference)
+      end
```
jonmoter: If there's no tag associated with a commit, I would assume `tag_from_ref` would return nil instead of having a fallback.

I think I would leave this as-is, and update `JobExecution#setup` to look like:

```ruby
tag = @repository.tag_from_ref(@reference) || @repository.commit_from_ref(@reference)
```

steved:
Thanks. Though will this break your statsd integration?

app/models/git_repository.rb
gabetax: agreed. reset back and changed.

gabetax:
@steved statsd analysis should be able to fallback in the same way. If not, people can start tagging releases if they want to opt in.

gabetax:
ping @zendesk/runway

jonmoter:
:+1:

PR #327, opened 2015-03-31
Add a retroactive approver address for deploy bypass

apanzerj:
Reviewers: @pswadi-zendesk @gabetax @jonmoter 

Description:

Adding one extra CC, the person who will do the retroactive approval when the bypass deploy is utilized.

cc: @pdeuter

### DIFF
--- a/app/mailers/deploy_mailer.rb
+++ b/app/mailers/deploy_mailer.rb
@@ -17,7 +17,8 @@ class DeployMailer < ApplicationMailer
     to_email = [BuddyCheck.bypass_email_address]
     to_email << BuddyCheck.bypass_jira_email_address if BuddyCheck.bypass_jira_email_address
 
-    mail(to: to_email, cc: user.email, subject: subject)
+    cc_email = [user.email, BuddyCheck.bypass_retroactive_approval_email]
+    mail(to: to_email, cc: cc_email, subject: subject)
   end
 
   def deploy_failed_email(stage, deploy, emails)
--- a/lib/policies/buddy_check.rb
+++ b/lib/policies/buddy_check.rb
@@ -9,6 +9,10 @@ module BuddyCheck
     ENV["BYPASS_EMAIL"]
   end
 
+  def bypass_retroactive_approval_email
+    ENV["BYPASS_RETROACTIVE_APPROVAL_EMAIL"]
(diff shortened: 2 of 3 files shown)


pswadi-zendesk:
Since this if or audit purpose, could we reuse audit email we already have? 
https://github.com/zendesk/samson/blob/master/.env.example#L28

apanzerj:
Is that being used inside Zendesk at the moment?

pswadi-zendesk:
yes, when the new projects are created, an email is sent out to this audit email address.

pdeuter:
No, we should not use that address.  We need two distinct email addresses, otherwise all those emails will come to me and that is not what we want.

app/mailers/deploy_mailer.rb
```diff
@@ -17,7 +17,8 @@ def bypass_email(stage, deploy, user)
     to_email = [BuddyCheck.bypass_email_address]
     to_email << BuddyCheck.bypass_jira_email_address if BuddyCheck.bypass_jira_email_address
 
-    mail(to: to_email, cc: user.email, subject: subject)
+    cc_email = [user.email, BuddyCheck.bypass_retroactive_approval_email]
```
jonmoter: Maybe add a `compact` to the end of this, for non-zendesk Samson users that don't have that environment variable defined.

jonmoter:
Are you going to open up a PR to chef_zendesk_samson to set the environment variable?  You'd probably want to set it around here:

https://github.com/zendesk/chef_zendesk_samson/blob/master/attributes/default.rb#L35

app/mailers/deploy_mailer.rb
jwswj:
```ruby
cc_email = [user.email]
cc_email << BuddyCheck.bypass_retroactive_approval_email if BuddyCheck.enabled?
```
or 
```ruby
cc_email = [user.email]
cc_email << BuddyCheck.bypass_retroactive_approval_email if BuddyCheck.bypass_retroactive_approval_email
```

jwswj: We are the only people I know who use the Buddy functionality, so it's nice to keep it separate.

apanzerj:
left this dangling...

PR #329, opened 2015-04-01
Use structs over stubs when testing serialized data

mariovisic:
Currently samson is setup to use memcached via dalli in all
environemnts, although for our testing we don't really need a persisted
cache store, and it saves us having to install memcached on our CI
server, so we've set the test environment to use a in memory cache.

The problem with this is that you can't actually serialize a stub
object. If you run the following code:
```ruby
Marshal.dump(stub("data", body: stub("body", {})))
```

You'll run into this error:
  > TypeError: no _dump_data is defined for class UnboundMethod

Dalli must either rescue from serlialization errors or uses another
method of serialization avoiding the issue.

Anyways, if you just change the stubs to open structs the tests pass
with other caches apart from dalli.

### DIFF
--- a/test/models/changeset/pull_request_test.rb
+++ b/test/models/changeset/pull_request_test.rb
@@ -1,16 +1,19 @@
 require_relative '../../test_helper'
 
 describe Changeset::PullRequest do
-  let(:data) { stub("data", user: user, merged_by: merged_by, body: body) }
+  DataStruct = Struct.new(:user, :merged_by, :body, :title)
+  UserStruct = Struct.new(:login)
+
+  let(:data) { DataStruct.new(user, merged_by, body) }
   let(:pr) { Changeset::PullRequest.new("xxx", data) }
-  let(:user) { stub(login: "foo") }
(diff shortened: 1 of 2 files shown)


jwswj:
@steved ?

steved:
This doesn't really fulfill the purpose of stubs since OpenStruct will just return `nil` for uninitialized attributes.

Dalli just eats the errors (you can see it in the log), so that probably isn't great either. Maybe we should just switch to a `null_store` in test environments, I don't really know how much we benefit from testing the caching.

mariovisic:
@steved would using a regular struct work OK? Calling methods that do not exist raise a NoMethodError

steved:
That'll probably work, though I think they'll need to be named classes.

mariovisic:
@steved modified to use structs, no need for named classes though in most classes, can be anonymous.

mariovisic:
Actually I can see that it fails to dump anonymous structs, have fixed!

steved:
+1

PR #330, opened 2015-04-01
Add webhook integration for deploying latest release

mariovisic:
Add a webhook which allows deploying a release for a given project. The
use case for this is having external services that aren't CI servers
being able to trigger deploys. An example is that we have a button in
the office which can be physically pushed that will ping the endpoint to
perform a deploy.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,3 +30,6 @@ GITHUB_TOKEN={fill me in}
 # DEPLOY_GROUP_FEATURE={optional, set to 1 to enable Environments and DeployGroups}
 
 # SLACK_TOKEN={ required for the slack integration }
+
+# WEBHOOK_SECRET={secret to authorize deploying latest release via webhook}
+
--- a/app/assets/javascripts/timeline.js
+++ b/app/assets/javascripts/timeline.js
@@ -83,7 +83,7 @@ samson.filter("projectUserFilter",
 
 samson.filter("userFilter",
   function() {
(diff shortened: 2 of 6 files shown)


mariovisic:
Let me know if this is something that you want in the core or not, I'm OK with keeping it just in our branch if that's the case.

This is the physical button we're going to be using to deploy, when you press it, it sends a http request to the endpoint: 
push button

jwswj:
We've got the ability to kick off a deployment by making a POST request to `/projects/#{project.id}/deploys` so Hubot can kick off deploys. Samson has basic auth https://github.com/zendesk/samson/blob/master/lib/warden/strategies/basic_strategy.rb so can I suggest that:
1. You pull the latest releases for a project (might need to add a JSON response to the releases controller).
2. Issue a deploy with the correct tag rather (than bake this logic into the integrations).

jwswj:
BTW - I want that button. Where can I buy that button?

mariovisic:
It's a [bt.tn button](http://bt.tn/)  The problem is that I can only really program it to ping a single HTTP endpoint; I don't think it's possible to have it get the latest release and then put that info into a new http request.

I'm OK with just keeping this change in our copy though in that case. The interface for setting up the button looks like this:

bt tn

jwswj:
Ah - that sucks. Would prefer to keep it in your fork for now.

PR #331, opened 2015-04-01
Only show buddy confirmationsettings when the feature is enabled

mariovisic:
The 'Does this stage impact production servers?' and 'Confirm before
deployment' options for a stage are only ever used to confirm
deployments when the buddy feature is activated. Therefore we do not
need to display them when the feature is disabled. Simiarly to how
fields are hidden when the 'deploy group feature' is not enabled.

### DIFF
--- a/app/views/stages/_fields.html.erb
+++ b/app/views/stages/_fields.html.erb
@@ -41,7 +41,7 @@
         ) %>
       </div>
     </div>
-  <% else %>
+  <% elsif BuddyCheck.enabled? %>
     <div class="form-group">
       <div class="col-sm-offset-2 col-sm-10">
         <div class="checkbox">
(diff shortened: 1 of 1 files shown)


jwswj:
@princemaple is this used in the timeline to group production deploys or are we still just using name?

princemaple:
I used the confirmation flag and have not ever changed it. Can't recall if someone changed it to use another flag. Definitely not stage name though. @jwswj

pswadi-zendesk:
Although similar, the two features are different in following respect : 
- 'Confirm before deployment' gives an opportunity for the deployer to look at the changes before doing the actual push. Typically this is useful for 'staging' environment. This is more of 'self check' and used even by non production environment.
- 'Buddy check' applies only for production and needs to be approved and someone else besides the deploy.
Since these two are used for different reasons, it is better to keep them separate, rather than forcing it to be used only for production environment.

mariovisic:
OH, That's what the confirm checkbox does! Would it be good to update the wording to say that it's to confirm deployment of a release? You always get a confirmation when pressing the deploy button on the  overview page.

mariovisic:
Also then wouldn't this setting here make the confirm checkbox not actually do anything? As it'll automatically release anyways:

stage_release

jwswj:
Yep - confirm before deployment only works when a user is manually deploying.

jwswj:
Closing this as the solution is to be clearer about what these checkboxes do.

PR #332, opened 2015-04-01
Remove plugin migrations as they are included in the main migration path...

zendesk-shender:
Don't need the plugin specific migration cmd, as the path glob is included in the application.rb.
Also, the slack db migration looked like it was hand-copied into the schema.rb and don't know why they limited the stage_id since those limits aren't in the migration file.

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -24,7 +24,7 @@ module Samson
     config.autoload_paths += Dir["#{config.root}/lib/**/"]
 
     # Add DB migrations from plugins
-    config.paths["db/migrate"] << File.join(config.root, 'plugins/**/db/migrate')
+    config.paths["db/migrate"] << config.root.join('plugins/**/db/migrate')
 
     config.cache_store = :dalli_store, { value_max_bytes: 3000000, compress: true, expires_in: 1.day }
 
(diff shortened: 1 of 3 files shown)


fneves:
:+1:

config/application.rb
```diff
@@ -24,7 +24,7 @@ class Application < Rails::Application
     config.autoload_paths += Dir["#{config.root}/lib/**/"]
 
     # Add DB migrations from plugins
-    config.paths["db/migrate"] << File.join(config.root, 'plugins/**/db/migrate')
+    config.paths["db/migrate"] << config.root.join('plugins/**/db/migrate')
```
grosser: is consolidating plugin logic into hooks.rb better ? -> deleted this instead

grosser: also this won't work if a plugin is a real gem

grosser: https://github.com/zendesk/samson/pull/353

zendesk-shender:
The initial problem was that regenerating the schema file didn't pick up our internal plugins' migrations automatically which I guess isn't a big deal if the person knows to migrate each of the plugins after that.

And wouldn't a real gem provide it's own migration mechanism, like gem:install?

grosser:
Most gems that come with migrations have a rails g xxx_migrations,
but since we all work on the same codebase I think having them
automatically be there is very convenient and does not have any downsides.

PR #333, opened 2015-04-01
Change how stage edit page allows user to select deploy_groups

zendesk-shender:
Hopefully making things a bit easier to select deploy groups for a stage.
apr 02 2015 14 01

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-66

### Risks
 - None

### DIFF
--- a/app/views/stages/_fields.html.erb
+++ b/app/views/stages/_fields.html.erb
@@ -26,19 +26,50 @@
   <% if ENV['DEPLOY_GROUP_FEATURE'] %>
     <div class="form-group">
       <%= form.label :name, 'Deploy Groups', class: 'col-lg-2 control-label' %>
-      <div class="col-lg-4">
-        <%= form.collection_select(
-          :deploy_group_ids, DeployGroup.order(:environment_id, :name).all, :id, :long_name,
-          {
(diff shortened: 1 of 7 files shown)


gabetax:
[indeterminate checkboxes](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate) could be a nice polish on the top levels, but not worth blocking on.

app/views/stages/_fields.html.erb
```diff
@@ -26,19 +26,54 @@
...
+              </tr>
+            </thead>
+            <tbody>
+              <% @environments.map {|e| e.deploy_groups.count }.max.to_i.times do |index| %>
+                <tr>
+                  <% @environments.each do |environment| %>
+                    <% if environment.deploy_groups[index] %>
+                      <td>
+                        <label>
+                          <input
```
gabetax: I'd be inclined to use the FormTagHelper methods if straight up FormHelper is too complicated in this scenario, but the only real win is consistency I suppose.

gabetax:
:+1: big UI improvement

zendesk-shender:
Thanks @gabetax for the pointer to the indeterminate checkbox state. That should work now as seen in the gif above.
I also changed the HTML to use the form tag helpers as much as possible as well.

PR #334, opened 2015-04-01
Check commit status when creating a manual build

gstamp:
A lot of CI tools use the github status API to indicate whether a given build is okay.

When doing a manual release this change queries the status and displays a warning if the build failed. It does not prevent the release, it just gives a warning.

It will look like this:

image

This could be extended to provide more information if desired but at the moment we're just keeping it reasonably simple.

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -79,12 +79,6 @@ class Changeset
       last_commit_status.error
   end
 
-  class StatusResult < Struct.new(:state, :error)
-    def self.null_result
-      new(nil, nil)
-    end
-  end
-
   private
 
   def find_comparison
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -28,6 +28,17 @@ class Changeset
     @commits ||= comparison.commits.map {|data| Commit.new(repo, data) }
   end
 
+  def last_commit_status
+    if comparison.commits.nil? || comparison.commits.empty?
+      nil
+    else
+      ref = comparison.commits.last["sha"]
+      Rails.cache.fetch(status_cache_key) do
+        GITHUB.combined_status(repo, ref)[:state]
+      end
+    end
+  end
+
   def files
     comparison.files
   end
@@ -85,6 +96,10 @@ class Changeset
 
   def cache_key
     [self.class, repo, previous_commit, commit].join('-')
+  end
+
+  def status_cache_key
+    [self.class, "status", repo, commit].join('-')
   end
 
   class NullComparison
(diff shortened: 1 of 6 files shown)


gstamp:
Will look into the build failure.

app/models/changeset.rb
```diff
@@ -28,6 +28,20 @@ def commits
     @commits ||= comparison.commits.map {|data| Commit.new(repo, data) }
   end
 
+  def last_commit_status
+    if comparison.commits.nil? || comparison.commits.empty?
```
sandlerr: you could use `.blank?` here http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F

app/models/changeset.rb
```diff
@@ -28,6 +28,20 @@ def commits
...
+      StatusResult.null_result
+    else
+      ref = comparison.commits.last["sha"]
+      state = Rails.cache.fetch(status_cache_key) do
```
sandlerr: let's give this an expiration time for the case when a flaky build is re-run and passes

gstamp:
@sandlerr Done.

sandlerr:
Great feature. Do you mind making the status check class its own file? 

Also I think the cache key doesn't need to care about the previous commit.

gstamp:
@sandlerr Not at all. I was considering making it it's own class, it just seemed a bit small. Small things tend to grow larger though (and probably will if we add extra information from the status API).

sandlerr:
It makes sense to me to have the Github call go inside the new class, maybe rename it to `StatusCheck` or something

steved:
There's a CommitStatus model / controller, maybe we can combine?
https://github.com/zendesk/samson/blob/master/app/models/commit_status.rb
https://github.com/zendesk/samson/blob/master/app/controllers/commit_statuses_controller.rb

You can see it in action in your screenshot, the SHA field is outlined red.

gstamp:
@steved Right!  Totally missed that.

I'll take a look. One difference that sticks out is that picks the status of the first status (`statuses.first.state`) rather than calling `GITHUB.combined_status()`.

gstamp:
I'm going to close this and open a new PR.

dasch:
Don't we already do this? The input field changes color depending on the build status.

gstamp:
@dasch That's why I closed the PR.

dasch:
Ok, sorry.

On Tue, Apr 7, 2015 at 10:41 AM, Glen Stampoultzis <notifications@github.com

PR #335, opened 2015-04-02
Status display box

gstamp:
Replaces my previous PR to show the current github status.

Includes two changes:

- A nice obvious warning box when there's a problem (pictured below)
- Change API to use combined status API rather than picking the first status seen.

image


sandlerr:
I like [combined_status](https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref) over [stasues](https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref). :+1:

jwswj:
:+1:

sandlerr:
thanks @gstamp!

PR #339, opened 2015-04-08
[WIP] Use websocket-rails for event-based page updates

zendesk-shender:
Very simple use case for testing websockets in Samson for updating pages dynamically without polling.

/cc @zendesk/samson

### References
 - Jira link:

### Risks
 - None


zendesk-shender:
Ditching this as we are going to use SSEs for the moment till we need bi-directional comms.

PR #341, opened 2015-04-09
Fix a bug where you could not selectively load plugins

mariovisic:
If you set the PLUGINS environment variable to only load some plugins,
it would not work properly as the name for each plugin was being
calculated incorrectly. It still worked fine if you had it set to ALL.

This patch fixes the name for plugins and so the loading then works
correctly. You can verify this by running this before/after the patch.

```ruby
Samson::Hooks.plugins.map { |plugin| plugin.instance_variable_get('@name') }
```

### DIFF
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -15,11 +15,11 @@ module Samson
       def initialize(path)
         @path = path
         @folder = File.expand_path("../../../", @path)
-        @name = File.dirname(@folder)
+        @name = File.basename(@folder)
       end
 
       def active?
-        Rails.env.test? || ENV["PLUGINS"] == "all" || ENV["PLUGINS"].to_s.split(",").include?(@name)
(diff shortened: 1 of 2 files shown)


gstamp:
Super handy! :-)

zendesk-shender:
Nice catch! :+1: 

Bonus points if you put in a test for the ENV['PLUGINS'] functionality! :)

mariovisic:
@zendesk-shender I was in the process, but the plugins detection method returns true in the test environment, I suppose I could stub that out though.

zendesk-shender:
Ah, yup: ``` Rails.env.stubs(:test?).returns(false)``` should work I think.

mariovisic:
Added tests, also now allows for spaces between commas, so `slack, zendesk` will now work, previously you'd have to do `slack,zendesk`

sandlerr:
:+1: thanks again :-) rock on!

PR #342, opened 2015-04-09
Fix buddy check tests when deploy_groups functionality is enabled.

zendesk-shender:
Just some simple fixes for if deploy group functionality is enabled by default for tests.

/cc @zendesk/samson @zendesk/runway 

### References
 - Jira link:

### Risks
 - None


jonmoter:
:+1:

sandlerr:
:+1:

PR #344, opened 2015-04-10
Tweak deploy_groups timeline and upgrade the JS test framework libraries...

zendesk-shender:
Set the timescale appropriately depending on too few or too many deploys being shown initially.
Also, lazy load the timeline to get better responsiveness.

Screenshot with lots of deploys which would have made the default view too big:
screen shot 2015-04-10 at 13 28 46

Screenshot with only a few deploys spread out over time, so we compress the timescale instead:
screen shot 2015-04-10 at 13 35 34

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/assets/javascripts/controllers/deploy_groups_ctrl.js
+++ b/app/assets/javascripts/controllers/deploy_groups_ctrl.js
@@ -1,15 +1,46 @@
 samson.controller('DeployGroupsCtrl', function($scope, $http, $location, $window) {
+  var MAX_PAGES = 20;
 
   function init() {
-    $http.get($location.path() + '.json').success(function(result) {
-      result.deploys.forEach(function(deploy) {
-        deploy.start = new Date(deploy.started_at || deploy.created_at);
+    getMoreDeploys(1);
+  }
+
+  function getMoreDeploys(page) {
+    if (page <= MAX_PAGES) {
+      $http.get($location.path() + '.json?page=' + page).success(function(result) {
+        if (result.deploys.length > 0) {
+          result.deploys.forEach(function(deploy) {
+            deploy.start = new Date(deploy.started_at || deploy.created_at);
+          });
+
+          $scope.items.add(result.deploys);
+          $scope.deployIds = $scope.deployIds.concat(_.pluck(result.deploys, 'id'));
+
+          adjustVisibleItems(result.deploys);
+          getMoreDeploys(page + 1);
+        }
       });
-      $scope.items.add(result.deploys);
-    });
+    }
+  }
+
+  function adjustVisibleItems() {
+    var visibleItems = $scope.timeline.getVisibleItems().length,
+        MIN_VIS_ITEMS = 10,
+        MAX_VIS_ITEMS = 20;
+
+    if (visibleItems > MAX_VIS_ITEMS) {
+      $scope.timeline.setOptions({
+        start: $scope.items.get($scope.deployIds[MAX_VIS_ITEMS-1]).start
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/controllers/deploy_groups_ctrl.js
```diff
@@ -1,15 +1,46 @@
...
 
   function init() {
-    $http.get($location.path() + '.json').success(function(result) {
-      result.deploys.forEach(function(deploy) {
-        deploy.start = new Date(deploy.started_at || deploy.created_at);
+    getMoreDeploys(1);
+  }
+
+  function getMoreDeploys(page) {
+    if (page < MAX_PAGES) {
```
sandlerr: nitpick: since this isn't zero-indexed it should probably be `<=`

app/assets/javascripts/controllers/deploy_groups_ctrl.js
```diff
@@ -1,15 +1,46 @@
...
+        if (result.deploys.length > 0) {
+          result.deploys.forEach(function(deploy) {
+            deploy.start = new Date(deploy.started_at || deploy.created_at);
+          });
+
+          $scope.items.add(result.deploys);
+          $scope.deployIds = $scope.deployIds.concat(_.pluck(result.deploys, 'id'));
+
+          adjustVisibleItems(result.deploys);
+          getMoreDeploys(page + 1);
```
sandlerr: will this know to stop somehow when you've already downloaded all the deploys?

app/assets/javascripts/controllers/deploy_groups_ctrl.js
```diff
@@ -1,15 +1,46 @@
...
+        MIN_VIS_ITEMS = 10,
+        MAX_VIS_ITEMS = 20;
+
+    if (visibleItems > MAX_VIS_ITEMS) {
+      $scope.timeline.setOptions({
+        start: $scope.items.get($scope.deployIds[MAX_VIS_ITEMS-1]).start
+      });
+    } else if (visibleItems <= MIN_VIS_ITEMS && $scope.items.length > MIN_VIS_ITEMS) {
+      $scope.timeline.setOptions({
+        start: $scope.items.get($scope.deployIds[MIN_VIS_ITEMS-1]).start
```
sandlerr: this appears to be the default behaviour unless I'm misreading `start` [here](http://visjs.org/docs/timeline.html#Configuration_Options)

sandlerr:
Looks good to me :+1:

app/assets/javascripts/controllers/deploy_groups_ctrl.js
zendesk-shender: Yea, this will stop when no more deploys are returned by the server or the MAX_PAGES is reached.

sandlerr: of course, it checks the length of the result.

app/assets/javascripts/controllers/deploy_groups_ctrl.js
zendesk-shender: I think since I'm dynamically loading in the items, the 'start' isn't updated by visjs, instead defaulting to 5 days always.

app/assets/javascripts/controllers/deploy_groups_ctrl.js
zendesk-shender: Good point, I'll fix that and then merge.

Issue #345, opened 2015-04-13
No way to delete macro?


gstamp:
Apologies if I've missed it but as far as I can tell there's no way to delete a macro through the GUI.

sandlerr:
Delete button has been added in #346

PR #346, opened 2015-04-13
delete macros

sandlerr:
#345 @zendesk/samson @zendesk/runway

### DIFF
--- a/app/helpers/deploys_helper.rb
+++ b/app/helpers/deploys_helper.rb
@@ -86,6 +86,10 @@ module DeploysHelper
     end
   end
 
+  def stop_button(classes = 'btn btn-danger btn-xl')
+    link_to "Stop", [@project, @deploy], method: :delete, class: classes
+  end
+
   private
 
     def deploy_status_panel_common(deploy, enabled, hash = { "cancelled" => "danger" } )
--- a/app/helpers/deploys_helper.rb
+++ b/app/helpers/deploys_helper.rb
@@ -86,8 +86,8 @@ module DeploysHelper
     end
   end
 
-  def stop_button(classes = 'btn btn-danger btn-xl')
-    link_to "Stop", [@project, @deploy], method: :delete, class: classes
+  def stop_button(options = {})
+    link_to "Stop", [@project, @deploy], options.merge({ method: :delete, class: options.fetch(:class, 'btn btn-danger btn-xl') })
   end
 
   private
--- a/app/views/admin/environments/index.html.erb
+++ b/app/views/admin/environments/index.html.erb
@@ -21,7 +21,7 @@
             <td>
               <%= link_to "Edit", edit_admin_environment_path(env) %>
               |
-              <%= link_to_delete(admin_environment_path(env)) %>
+              <%= link_to_delete([:admin, env]) %>
             </td>
           </tr>
         <% end %>
--- a/app/views/admin/environments/index.html.erb
+++ b/app/views/admin/environments/index.html.erb
@@ -21,7 +21,7 @@
             <td>
               <%= link_to "Edit", edit_admin_environment_path(env) %>
               |
-              <%= link_to "Delete", admin_environment_path(env), method: :delete, data: { confirm: "Are you sure?" } %>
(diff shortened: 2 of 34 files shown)


app/views/macros/edit.html.erb
```diff
@@ -5,5 +5,16 @@
 <h1>Edit macro</h1>
 
 <section>
+  <div class="row">
+    <div class="col-lg-2 button-label">Actions</div>
+    <div class="col-lg-6">
+      <%= link_to 'Execute', execute_project_macro_path(@project, @macro), method: :post, class: 'btn btn-primary', 'data-confirm' => "Are you sure you have no unsaved changes and would like to execute '#{@macro.name}'?" %>
```
grosser: you can do `data: {confirm: "xxxxx"}`

app/views/macros/edit.html.erb
```diff
@@ -5,5 +5,16 @@
...
+      <% if current_user.is_admin? %>
+        <%= link_to project_macro_path(@project, @macro), class: 'btn btn-danger', method: :delete, data: { confirm: "Are you sure?" } do %>
```
grosser: `link_to [@project, @macro]` will build the same path

grosser: maybe a `link_to_delete` helper would remove some duplication here like the classes/glyphicon etc

grosser:
:+1:

sandlerr:
@grosser refactored delete buttons & links

app/views/admin/environments/index.html.erb
```diff
@@ -21,7 +21,7 @@
             <td>
               <%= link_to "Edit", edit_admin_environment_path(env) %>
               |
-              <%= link_to "Delete", admin_environment_path(env), method: :delete, data: { confirm: "Are you sure?" } %>
+              <%= link_to_delete(admin_environment_path(env)) %>
```
grosser: `[:admin, env]` would also work :)

app/views/macros/edit.html.erb
```diff
@@ -5,5 +5,14 @@
 <h1>Edit macro</h1>
 
 <section>
+  <div class="row">
+    <div class="col-lg-2 button-label">Actions</div>
+    <div class="col-lg-6">
+      <%= link_to 'Execute', execute_project_macro_path(@project, @macro), method: :post, class: 'btn btn-primary', data: { confirm: "Are you sure you have no unsaved changes and would like to execute '#{@macro.name}'?" } %>
+      <% if current_user.is_admin? %>
+        <%= link_to_delete_button([@project, @macro]) %>
```
grosser: the authorization logic should ideally not be duplicated -> `current_user.can?(:destroy, @macro)` and use it in both places

app/views/admin/environments/index.html.erb
sandlerr: Couldn't find any docs so had to check the source, this magic comes from `url_for` which is also used in redirects. I'll go through and make these as consistent as I can.

sandlerr: avoiding adding any `options` [here](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/routing/polymorphic_routes.rb#L99) so just updating links to `#index`

grosser:
:+1:

sandlerr:
cc @zendesk/samson now I've done it I'm not sure about it. This PR will make it more dry and concise

app/views/macros/edit.html.erb
sandlerr: we have `.can#{do_stuff}` all over the place in our models and view helpers. I'll fix that in another PR

app/helpers/deploys_helper.rb
```diff
@@ -86,6 +86,10 @@ def stages_select_options
     end
   end
 
+  def stop_button(classes = 'btn btn-danger btn-xl')
```
grosser: I'd expect the first argument to be an options hash

grosser:
I like the dry-ness, also less things to type :)

PR #349, opened 2015-04-14
Fix bug where some buildkite integrations would not create a relase

mariovisic:
Buildkite sometimes sends `build.finished` events rather than `build`
and in those cases we weren't creating a release. I have removed that
check because we don't actually need to check the event type, because we
can get enough info from the fact that we have a successful build with a
valid commit.

### DIFF
--- a/test/controllers/integrations/buildkite_controller_test.rb
+++ b/test/controllers/integrations/buildkite_controller_test.rb
@@ -36,8 +36,6 @@ describe Integrations::BuildkiteController do
   before { Deploy.delete_all }
 
   context 'when buildkite passes a build event' do
-    before { request.headers['X-Buildkite-Event'] = 'build' }
-
     test_regular_commit 'Buildkite', failed: { build: { state: 'failed' }}, no_mapping: { build: { branch: 'non-existent-branch' } } do
       project.webhooks.create!(stage: stages(:test_staging), branch: 'master')
     end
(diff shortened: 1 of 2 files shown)


sandlerr:
:+1:

sandlerr:
@zendesk/samson

test/controllers/integrations/buildkite_controller_test.rb
```diff
@@ -36,7 +36,7 @@
   before { Deploy.delete_all }
 
   context 'when buildkite passes a build event' do
-    before { request.headers['X-Buildkite-Event'] = 'build' }
+    before { request.headers['X-Buildkite-Event'] = 'build.finished' }
```
zendesk-shender: We don't need this line now, right?

mariovisic: Correct, will fix that up before merging.

mariovisic:
Just hold off merging this, jus testing it out thoroughly today.

mariovisic:
Cool, this is ready to go, I thought it was not working due to this issue: https://github.com/zendesk/samson/issues/359

Basically you get webhooks out of order sometimes and it messes up the changeset displays.

sandlerr:
@zendesk/samson @zendesk/runway

zendesk-shender:
:+1:

grosser:
failing test would also be neat for any fix ;)

PR #350, opened 2015-04-14
include vis js as gem to mave raw source and no failing vis.map load

grosser:
screen shot 2015-04-14 at 9 38 52 am

@zendesk/runway 

### Risks
 - None


grosser:
we'll most likely want to move more of the vendor/assets things to rails-assets ... but 1 step at a time ...
what page can I see viz being used on to confirm it works ?

jonmoter:
@grosser you can see it in action on pages like this: https://samson.zende.sk/deploy_groups/2

grosser:
apparently there is a bug when using bundler with source block or `source: 'xxxx'` we can start using blocks once bundler is upgraded on samson ...

tested on staging and looks good https://samsontest.zende.sk/deploy_groups/2, not showing anything locally on master and this branch

good to go ?

jonmoter:
Is there anything we can do to fix the build on Travis?

grosser:
@zendesk-shender vis.js cannot be found by gulp anymore -> running karma through rake

zendesk-shender:
Aye, this is why I didn't use the gem for visjs originally, I couldn't find a good way to get the path from within the gem outside of the rails env.
I'll have a poke at the karma hacks you made to see if I can do anything, but I think you are more the expert here! :cry:

zendesk-shender:
Btw, if you are just trying to remove that .map 404, I think putting the vis.map file in the vendor/assets/javascript directory solves it, no? That file is part of the visjs download.

zendesk-shender:
@grosser I pushed up the fix for 'npm test', and we don't need it anymore in our travis.yml file as the extra 'rake test:js' you put in there does the same thing anyway.

I removed Gulp as we don't need that anymore either as we were only using it in its most basic form.

Rakefile
```diff
@@ -15,3 +16,49 @@ namespace :plugins do
...
+  def resolve_asset(file)
+    Rails.application.assets.find_asset(file).to_a.first.pathname.to_s
+  end
+
+  def build_karma_config(files)
+    config = File.read('test/karma.conf.js')
+    {
+      "INSERT_FILES_VIA_RAKE" => "\"#{files.join("\",\n\"")}\"",
+      "INSERT_BASE_PATH_VIA_RAKE" => Bundler.root.to_s,
+    }.each { |key, value| config.gsub!(key, value) || raise("#{key} not found") }
```
zendesk-shender: Could we not do all this via ERB and keep this task clean with a few ERB directives in the karma.conf.js file?

grosser:
at first I only wanted to fix the missing map, but then I also wanted to use versioned assets instead of copy-pasted ones :)

Rakefile
grosser: that sounds nice :)

grosser:
all erb now, much cleaner :)

good to go ?

grosser:
I'll convert other things to asset-gems after this is done

zendesk-shender:
Nice! :)
:+1:

PR #351, opened 2015-04-14
Gitlab support

deanperry:
Hey, I've just found Samson and it looks great!
I use a self hosted version of Gitlab for my projects so I've added support for Gitlab OAuth and showing Gitlab links instead of Github on the project pages.


app/models/project.rb
```diff
@@ -72,6 +72,18 @@ def repository_homepage
     "//#{Rails.application.config.samson.github.web_url}/#{github_repo}"
   end
 
+  def repository_homepage_gitlab
+    "//#{ENV['GITLAB_URL']}/#{github_repo}"
```
sandlerr: Would be better if `github_repo` was only used for Github or rename it to a more generic name. Also, mind putting the gitlab url as a config var like [this](https://github.com/zendesk/samson/blob/master/config/application.rb#L57)?

app/models/project.rb
```diff
@@ -72,6 +72,18 @@ def repository_homepage
...
+  end
+
+  def github?
+    repository_url.include?('github.com')
```
sandlerr: `repository_url.include? Rails.application.config.samson.github.web_url` - could make a shorter method to wrap that config variable

app/views/sessions/new.html.erb
```diff
@@ -12,5 +12,10 @@
         <%= image_tag image_url('google.png') %> Login with Google
       <% end %>
     <% end %>
+    <% if Rails.application.config.samson.auth.gitlab %>
+      <%= link_to omniauth_path('gitlab'), class: "action github" do %>
```
sandlerr: Nitpick: Gitlab's background seems to be `#548`?

deanperry:
Thanks for the comments, it was just a quick PR before I went to bed. I'll make those changes tomorrow. I'll also see if I can fix the build :)

sandlerr:
Thanks @deanperry looking good! FYI, many specs are failing because the envar isn't set so we are doing `string.include? nil` in Project

deanperry:
Ah I see. When logged in with GitHub is there a way to import all repos? If
so I should be able to use the gitlab API to do the same


On Wednesday, April 15, 2015, Roman Sandler <notifications@github.com>
wrote:

> Thanks @deanperry <https://github.com/deanperry> looking good! FYI, many
> specs are failing because the envar isn't set so we are doing string.include?
> nil in Project
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendesk/samson/pull/351#issuecomment-93117752>.
>


-- 
Sent from Gmail Mobile

sandlerr:
Github provides [this](https://developer.github.com/v3/repos/#list-organization-repositories) API but the feature doesn't exist in Samson yet. We find that many of our projects don't need a specific deployment step (for example [api clients](https://github.com/zendesk/zendesk_api_client_rb)).
We definitely welcome pull requests though!

app/views/shared/_gitlab_link.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= link_to project.repository_homepage_gitlab, class: "github", target: "_blank", title: "View repository on Gitlab" do %>
+  <%= image_tag "gitlab.png", width: 21, height: 21 %>
+<% end %>
```
zendesk-shender: Needs newline at EOF.

deanperry:
I've pushed some fixes

sandlerr:
@deanperry would you mind rebasing?

geshan:
is it ready to be merged?

deanperry:
I believe it is

sandlerr:
@deanperry this has merge conflicts due to changes that have been made in our master branch. We would very much like to merge your contribution but you'll need to get the latest master and do a `git rebase` of your branch onto it.

silverbp:
It would be nice to get this in there, we use gitlab as well!

beniwtv:
We also use Gitlab, so this would be definitively cool!

cmckni3:
We use GitLab as well. I have rebased @deanperry's work in my [fork](https://github.com/cmckni3/samson/tree/gitlab-support) and I am making sure everything works.

Should we add/modify some tests for this feature @zendesk/samson?

sandlerr:
@cmckni3 yes, we'll need to make some changes that aren't yet reflected in https://github.com/zendesk/samson/compare/master...cmckni3:gitlab-support - please put up a new pull request for this discussion.

cmckni3:
Sure @sandlerr. See #523.

deanperry:
Ok I'm closing this for #523

PR #353, opened 2015-04-15
plugins in gems do not work with just local paths

grosser:
partially reverting https://github.com/zendesk/samson/pull/332

### Risks
 - None


sandlerr:
:+1:

PR #354, opened 2015-04-16
Authorization through CanCanCan

sandlerr:
@zendesk/samson @zendesk/runway thoughts?

IMO https://github.com/zendesk/samson/commit/5f76d7b8f1dfbcb305959131f306167c4ac7e322 would be quite useful and save us from making a mess in future


grosser:
I don't like cancan because I think it's over-engineered / obscures the actual logic, afaik all we need is:

```Ruby
def can?(action, resource)
  case resource
  when User
    case action
    when :edit then resource == self
    else
      raise "Unsupported action #{action}"
    end
  else
      raise "Unsupported resource #{resource}"
  end
end
```

grosser:
the controller helpers can be
```Ruby
module ResourceAccess
  def ensure_can_edit_resource
    head :forbidden unless current_user.can?(:edit, resource)
  end
end

# and

before_filter :ensure_can_edit_resource, only: [:edit, :update, :destroy]

def item
  @item ||= Item.find(params[:id])
end
alias resource item

# in every controller that wants auth
```

PR #355, opened 2015-04-16
fix github webhook

sandlerr:
cc @zendesk/samson @zendesk/runway 

This was failing simply because `release_branch == 'master'`, but `params[:ref] == 'refs/heads/master'`

### DIFF
--- a/app/controllers/integrations/buildkite_controller.rb
+++ b/app/controllers/integrations/buildkite_controller.rb
@@ -24,10 +24,4 @@ class Integrations::BuildkiteController < Integrations::BaseController
   def build_param
     @build_param ||= params.fetch(:build, { })
   end
-
-  private
-
-  def service_type
-    'ci'
-  end
 end
--- a/app/controllers/integrations/buildkite_controller.rb
+++ b/app/controllers/integrations/buildkite_controller.rb
@@ -24,4 +24,10 @@ class Integrations::BuildkiteController < Integrations::BaseController
   def build_param
     @build_param ||= params.fetch(:build, { })
   end
+
+  private
+
+  def service_type
+    'ci'
+  end
 end
--- a/app/controllers/integrations/github_controller.rb
+++ b/app/controllers/integrations/github_controller.rb
@@ -16,7 +16,7 @@ class Integrations::GithubController < Integrations::BaseController
       request.body.tap(&:rewind).read
     )
 
-    request.headers['X-Hub-Signature'] == "sha1=#{hmac}"
+    Rack::Utils.secure_compare(request.headers['X-Hub-Signature'], "sha1=#{hmac}")
   end
 
   def valid_payload?
@@ -29,6 +29,6 @@ class Integrations::GithubController < Integrations::BaseController
 
   def branch
     # Github returns full ref e.g. refs/heads/...
-    params[:ref]
+    params[:ref][/refs\/heads\/(.+)/, 1]
   end
 end
--- a/app/controllers/integrations/github_controller.rb
+++ b/app/controllers/integrations/github_controller.rb
@@ -31,4 +31,10 @@ class Integrations::GithubController < Integrations::BaseController
     # Github returns full ref e.g. refs/heads/...
     params[:ref][/refs\/heads\/(.+)/, 1]
   end
+
+  private
+
+  def service_type
+    'code'
+  end
 end
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -1,11 +1,17 @@
 class Webhook < ActiveRecord::Base
   has_soft_deletion default_scope: true
-  validates :branch, uniqueness: { scope: [ :stage], conditions: -> { where("deleted_at IS NULL") }, message: "one webhook per (stage, branch) combination." }
+  validates :branch, uniqueness: { scope: [ :stage ], conditions: -> { where("deleted_at IS NULL") }, message: "one webhook per (stage, branch) combination." }
 
   belongs_to :project
   belongs_to :stage
 
+  SOURCES = %w[semaphore tddium travis jenkins buildkite github].freeze
+
   def self.for_branch(branch)
     where(branch: branch)
+  end
+
+  def self.for_source(service_type, service_name)
+    where(source: [ 'any', "any_#{service_type}", service_name ])
   end
 end
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -5,7 +5,9 @@ class Webhook < ActiveRecord::Base
   belongs_to :project
   belongs_to :stage
 
-  SOURCES = %w[semaphore tddium travis jenkins buildkite github].freeze
+  SOURCES = Rails.root.join('app','controllers', 'integrations').children(false).map do |controller_path|
+    controller_path.to_s[/\A(?!base)(\w+)_controller.rb\z/, 1]
+  end.compact.freeze
 
   def self.for_branch(branch)
     where(branch: branch)
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -5,10 +5,6 @@ class Webhook < ActiveRecord::Base
   belongs_to :project
   belongs_to :stage
 
-  SOURCES = Rails.root.join('app','controllers', 'integrations').children(false).map do |controller_path|
-    controller_path.to_s[/\A(?!base)(\w+)_controller.rb\z/, 1]
-  end.compact.freeze
-
   def self.for_branch(branch)
     where(branch: branch)
   end
--- a/db/migrate/20150416231106_add_source_to_webhook.rb
+++ b/db/migrate/20150416231106_add_source_to_webhook.rb
@@ -1,0 +1,6 @@
+class AddSourceToWebhook < ActiveRecord::Migration
+  def change
+    add_column :webhooks, :source, :string, default: 'any_ci'
+    Webhook.update_all source: 'any_ci'
+  end
+end
--- a/db/migrate/20150416231106_add_source_to_webhook.rb
+++ b/db/migrate/20150416231106_add_source_to_webhook.rb
@@ -1,6 +1,12 @@
 class AddSourceToWebhook < ActiveRecord::Migration
-  def change
-    add_column :webhooks, :source, :string, default: 'any_ci'
-    Webhook.update_all source: 'any_ci'
+  def up
+    add_column :webhooks, :source, :string, null: false, default: 'any_ci'
+    change_column_default :webhooks, :source, nil
+  end
+
+  def down
+    change_table :webhooks do |t|
+      t.remove :source
+    end
   end
 end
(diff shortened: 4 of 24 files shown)


grosser:
a test for both changes would be great

steved:
:-1: I needed this because if you have a CI hook and a Github hook you don't want them both triggering multiple deploys (i.e. production build and acceptance deploy).

steved:
It was a hack, the proper solution would probably differentiating between push and CI hooks.

app/controllers/integrations/github_controller.rb
```diff
@@ -16,7 +16,7 @@ def valid_signature?
       request.body.tap(&:rewind).read
     )
 
-    request.headers['X-Hub-Signature'] == "sha1=#{hmac}"
+    Rack::Utils.secure_compare(request.headers['X-Hub-Signature'], "sha1=#{hmac}")
```
steved: +1 on this though

sandlerr:
@zendesk/samson @zendesk/runway what do you think of this approach? Webhooks are now configured to actually listen to a branch as well as specific sources - options are github, travis, semaphore etc or any_ci, any_code, any.

app/models/webhook.rb
```diff
@@ -1,11 +1,17 @@
 class Webhook < ActiveRecord::Base
   has_soft_deletion default_scope: true
-  validates :branch, uniqueness: { scope: [ :stage], conditions: -> { where("deleted_at IS NULL") }, message: "one webhook per (stage, branch) combination." }
+  validates :branch, uniqueness: { scope: [ :stage ], conditions: -> { where("deleted_at IS NULL") }, message: "one webhook per (stage, branch) combination." }
 
   belongs_to :project
   belongs_to :stage
 
+  SOURCES = %w[semaphore tddium travis jenkins buildkite github].freeze
```
grosser: hmm this is repeating all controller names, would be neat to have a test that makes sure that list is complete

grosser:
sounds good to me

app/controllers/integrations/buildkite_controller.rb
```diff
@@ -24,4 +24,10 @@ def commit
   def branch
     params[:build][:branch]
   end
+
+  private
+
+  def service_type
```
steved: This should probably just be the default in the base controller

app/models/webhook.rb
steved: yeah, I would just use subclasses of integration::base and have the name be a class method

db/migrate/20150416231106_add_source_to_webhook.rb
```diff
@@ -0,0 +1,6 @@
+class AddSourceToWebhook < ActiveRecord::Migration
+  def change
+    add_column :webhooks, :source, :string, default: 'any_ci'
+    Webhook.update_all source: 'any_ci'
```
steved: Does the default not cover it?

sandlerr: My understanding is default is for new records, update_all is for existing ones in our db.

app/models/webhook.rb
sandlerr: Tried `SOURCES = Integrations::BaseController.subclasses.map(&:service_name).freeze` but had trouble because the subclasses hadn't been loaded when the constant was evaluated (base controller was probably autoloaded then).

app/models/webhook.rb
```diff
@@ -1,11 +1,19 @@
 class Webhook < ActiveRecord::Base
   has_soft_deletion default_scope: true
-  validates :branch, uniqueness: { scope: [ :stage], conditions: -> { where("deleted_at IS NULL") }, message: "one webhook per (stage, branch) combination." }
+  validates :branch, uniqueness: { scope: [ :stage ], conditions: -> { where("deleted_at IS NULL") }, message: "one webhook per (stage, branch) combination." }
 
   belongs_to :project
   belongs_to :stage
 
+  SOURCES = Rails.root.join('app','controllers', 'integrations').children(false).map do |controller_path|
```
grosser: this might be a future problem with plugin integrations not being found ... but they could register themselves somehow inside the plugin.

db/migrate/20150416231106_add_source_to_webhook.rb
grosser: the new column is added to all rows -> all get this default value -> no update necessary

would it make sense to remove the default after and make the column not-null so we are sure they get picked deliberately and minimize db/app coupling ?

test/controllers/integrations/github_controller_test.rb
```diff
@@ -8,7 +8,7 @@
 
   let(:payload) do
     {
-      ref: 'refs/heads/origin/dev',
```
grosser: are you sure they are not coming in with the origin ? <-> did you check a real request before changing the fixture ?

test/controllers/integrations/github_controller_test.rb
```diff
@@ -53,6 +53,31 @@
...
+      request.headers['X-Hub-Signature'] = "sha1=#{@hmac}"
+      time_success = Benchmark.measure do
+        1000.times { post :create, @payload }
+      end.total
+      request.headers['X-Hub-Signature'] = "sha1=#{@hmac.sub(/[0-8a-f]/, '9')}"
+      time_fail = Benchmark.measure do
+        1000.times { post :create, @payload }
+      end.total
+      time_difference = (time_success - time_fail).abs
+      pending { assert time_difference < 0.002 }
```
grosser: given ruby's slowness and built-in fastness I'd doubt that this test was ever red ... correct ?

grosser: ah it's also pending ... maybe kill the test ... and add a comment on why we are doing the secure-compare ... maybe using secure-compare is unnecessary/useless anyway since there is no way to detect a string compare diff in a 100ms response ...

db/migrate/20150416231106_add_source_to_webhook.rb
steved: I think if you add the non-null contraint, then the default will be added to all old rows. But I'm not 100% sure across the 3 DBs we support.

app/models/webhook.rb
steved: Yeah, I'd rather just rely on autoloading and controller registration. This is really only a problem in test and development, right?

test/controllers/integrations/github_controller_test.rb
sandlerr: The guy that wrote `bcrypt` thinks 20us is detectable http://codahale.com/a-lesson-in-timing-attacks/

We are nowhere near that timing safe in my testing, but then again other integrations don't even provide a HMAC. Successful comparisons take about 2-3 times longer even when `valid_payload?` returns `false`

grosser: hmm yeah true the mean might be different ... still a pending test does not help much / will be flaky at best

test/controllers/integrations/github_controller_test.rb
sandlerr: yep, you get `refs/heads/staging` or `refs/tags/v5`

test/controllers/integrations/github_controller_test.rb
sandlerr: so this has a crazy amount of variance in it unrelated to GC. I'm going to kill the test off.

app/models/webhook.rb
sandlerr: What if this class `require`s everything in `controlllers/integrations` and then uses the `IntegrationsController.descendants` method?

grosser: that sounds nice, loading more controllers is fine in dev/test as long as we don't load them at startup (triggers load of all kinds of other things)

the code can be test/dev only since production will eager load ... maybe be in an `if Rails.application.eager_load` block

grosser: if we want this to work with plugins and not preload any models/controllers (for fast startup) then we'd need some non-model constant that we fill manually like `Samson::Integration = ['Github', ...]` and then plugins would do `Samson::Integration << 'Foobar'`

grosser: or we could fill it from the file system lookup as you do here ... so maybe just move this outside of Webhook and keep it to be future proof ...

db/migrate/20150416231106_add_source_to_webhook.rb
```diff
@@ -0,0 +1,12 @@
+class AddSourceToWebhook < ActiveRecord::Migration
+  def up
+    add_column :webhooks, :source, :string, null: false, default: 'any_ci'
+    change_column_default :webhooks, :source, nil
```
sandlerr: @grosser @steved changed to make it `null false` and not only use the default value to update old records, verified that the default is applied to existing rows in all 3 dbs.
I used up and down because this isn't reversible otherwise.
The other option would be to go `Webhook.update_all source: 'any_ci'`

steved: I guess I was suggesting make it not-null completely. Any reason to have a null source?

sandlerr: This makes it not null, it only makes legacy webhooks into CI hooks. I definitely don't want null sources.

steved: Fair enough, I guess I thought it just doesn't seem to make much sense to require the app to validate the value as well.

lib/samson/integration.rb
```diff
@@ -0,0 +1,16 @@
+module Samson
+  class Integration
+
+    @@integrations = Rails.root.join('app','controllers', 'integrations').children(false).map do |controller_path|
```
grosser: just `SOURCES=` ? <-> no method missing ?

grosser:
:+1:

lib/samson/integration.rb
sandlerr: It doesn't seem idiomatic to have a constant that we expect to mutate if plugins want to register themselves here

grosser: then maybe this ?

```
cattr_accessor :sources
@@sources ||= xxxx

and use .sources
```

sandlerr: thanks!

lib/samson/integration.rb
```diff
@@ -0,0 +1,16 @@
+module Samson
+  class Integration
+
+    @@sources = Rails.root.join('app','controllers', 'integrations').children(false).map do |controller_path|
+      controller_path.to_s[/\A(?!base)(\w+)_controller.rb\z/, 1]
+    end.compact
+
+    cattr_reader :sources
+
+    def self.register_source(source)
```
grosser: not used <-> don't add ? ... maybe also just `Integration.sources << xxx` is a good enough api :)

sandlerr: ah of course, you can call methods on the array using its getter!

PR #356, opened 2015-04-16
install javascript via gems

grosser:
- documented versions
 - full source vs minified
 - full list of included js/css files for easy sanity check
 - easy to upgrade

### Risks
 - None


grosser:
@zendesk/runway 

screen shot 2015-04-16 at 10 04 22 am

config/environments/development.rb
```diff
@@ -27,7 +27,7 @@
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
-  config.assets.debug = false
+  config.assets.debug = true
```
grosser: need this on to see the list of files in the source view ... should not be much slower since local fetch is pretty fast

grosser:
list of files included when comparing master/this branch in source view is the same, so looking good afaik :)

zendesk-shender:
Looks like you can get rid of that hack if you replace the jquery rails gems by adding:
```
gem 'rails-assets-jqueryui'
gem 'rails-assets-jquery-ujs'
```
and removing:
```
gem 'jquery-rails'
gem 'jquery-ui-rails'
```
You'll also have to change ```jquery_ujs``` to ```jquery-ujs```, and ```jquery-ui``` to ```jquery-ujs```.

grosser:
I read a few threads and they were recommending against that,
only reason I could see to not do it is that rails produces the html and if
the html does not match the ujs then things won't work

how about we merge this one and then we convert the rest in another PR and
if something goes wrong just revert the jquery-rails conversion ?

zendesk-shender:
Fair enough! :)

zendesk-shender:
:+1:

sandlerr:
there are a few references to underscore in the codebase, is it gone?

grosser:
underscore is coming from a different gem, are you seeing any issues ?

sandlerr:
oh I see, the underscore asset you removed was only used to run the tests. Carry on, awesome cleanup :-)

PR #357, opened 2015-04-16
fix bug where buddy checks didn't work when deploy groups feature enabled

jonmoter:
since we redefined the Stage#production? method, that logic needs to be run in ruby instead of via a SQL query.
also refactored code to consolidate the places we check the DEPLOY_GROUP_FEATURE env var.

### Risks
 - Logic fix around buddy checks, when a past approval affects a new deploy

### DIFF
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -32,16 +32,13 @@ class DeployService
   end
 
   def latest_approved_deploy(reference, project)
-    deploy = nil
-    Deploy.where(reference: reference).where(' buddy_id is NOT NULL AND started_at > ?', BuddyCheck.period.hours.ago)
+    Deploy.where(reference: reference).where('buddy_id is NOT NULL AND started_at > ?', BuddyCheck.period.hours.ago)
       .includes(:stage)
-      .where(stages: {project_id: project, production: true})
+      .where(stages: {project_id: project})
       .each do |d|
-        next if bypassed?(d.stage, d, d.buddy)
-        deploy = d
-        break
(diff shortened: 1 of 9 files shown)


app/models/deploy_service.rb
```diff
@@ -32,16 +32,13 @@ def auto_confirm?(stage)
   end
 
   def latest_approved_deploy(reference, project)
-    deploy = nil
-    Deploy.where(reference: reference).where(' buddy_id is NOT NULL AND started_at > ?', BuddyCheck.period.hours.ago)
+    Deploy.where(reference: reference).where('buddy_id is NOT NULL AND started_at > ?', BuddyCheck.period.hours.ago)
       .includes(:stage)
-      .where(stages: {project_id: project, production: true})
+      .where(stages: {project_id: project})
       .each do |d|
```
grosser: `.detect` ?

app/views/projects/_stage.html.erb
```diff
@@ -4,7 +4,7 @@
       <%= link_to stage.name, [project, stage] %>
       <%= stage_lock_icon stage %>
     </td>
-    <% if ENV['DEPLOY_GROUP_FEATURE'] %>
+    <% if DeployGroup.enabled? %>
```
grosser: :heart: nice cleanup

grosser:
:+1:

grosser:
and the detect ?

app/models/deploy_service.rb
jonmoter: Ah, I missed that comment.  Yeah, `detect` would be better here.  I'll create another PR to clean up this class a little bit.

PR #358, opened 2015-04-16
covert everything to rails-assets

grosser:
@zendesk/runway 

seems to boot after a bit of hand-holding/renaming ... I'll wait for the first batch to reach production with out complaints and then try this ...

### Risks
 - all kinds of js broken


grosser:
currently fails to deploy because of an old bundler version ... I'll see if an bundler update fixes this

grosser:
turns out bootstrap-sass has more variables and stuff then the rails-assets variant, leaving it in for now, also `rake assets:precompile` shows less errors then `RAILS_ENV=staging rake assets:precompile`, so I'll follow up with another PR to unify that

grosser:
staging looks good, I'll just ship it, I was most worried about the bootstrap changes anyway :)

@zendesk/runway good to go ?

test/integration/cleanliness_test.rb
```diff
@@ -5,4 +5,10 @@
   it "does not have boolean limit 1 in schema since this breaks mysql" do
     File.read("db/schema.rb").wont_match /\st\.boolean.*limit: 1/
   end
+
+  it "does not include rails-assets-bootstrap" do
+    # make sure rails-assets-bootstrap did not get included by accident (dependency of some other bootstrap thing)
+    # if it is not avoidable see http://stackoverflow.com/questions/7163264
+    File.read('Gemfile.lock').wont_include 'rails-assets-bootstrap '
```
sandlerr: mind making this a regex like `rails-assets-bootstrap\w+(` ? It makes the intention a bit more clear imo. Otherwise people (actually me) will be tempted to remove the whitespace when they see it

sandlerr:
:+1:

zendesk-shender:
:+1:

PR #360, opened 2015-04-17
test compile

grosser:
@zendesk/runway finding compile errors on deploy is not fun :)

using RAILS_ENV=production since that showed more errors then in development (I'd guess because of the uglifier)

also making staging and production the same, so we don't have surprises ... diff was serve_static_files being on in staging + staging not using lograge ... hopefully this was just an oversight ... otherwise staging will be load production + change config

### Risks
 - None


jonmoter:
Travis build is failing

```
$ RAILS_ENV=production bundle exec rake assets:precompile
rake aborted!
No GitHub token available
```

grosser:
fixed, good to go ?

sandlerr:
:+1:

PR #361, opened 2015-04-20
Use SSEs to broadcast changes in deployments so clients don't have to poll

zendesk-shender:
Instead of polling every 5 seconds for number of active deploys, use SSEs to broadcast changes to connected users.

### References
 - Jira link:

### Risks
 - None


grosser:
Now we again do an extra request on every page load ? -> Timeout on load ?

grosser:
idk ... lot's of complexity and moving parts for very little gain ... is that worth it ?

grosser:
if it's the beginning of using sse for a bunch of stuff then that would make more sense :)

app/models/deploy_service.rb
```diff
@@ -54,6 +55,7 @@ def release_approved?(deploy)
 
   def send_before_notifications(stage, deploy, buddy)
     Samson::Hooks.fire(:before_deploy, stage, deploy, buddy)
+    SseRailsEngine.send_event('deploys', { type: 'start' })
```
grosser: test for these changes ?

app/controllers/deploys_controller.rb
```diff
@@ -98,6 +98,7 @@ def changeset
   def destroy
     if @deploy.can_be_stopped_by?(current_user)
       @deploy.stop!
+      SseRailsEngine.send_event('deploys', { type: 'finish' })
```
grosser: seems strange that these things are spread out into multiple files, maybe handle everything in the deploy ?

grosser: or the deploy service

app/models/deploy_service.rb
grosser: also don't need this / not used atm

app/assets/javascripts/services/sse_factory.js
```diff
@@ -0,0 +1,20 @@
+samson.factory('SseFactory', function($location) {
+  'use strict';
+
+  var sse = {
+    connection: null,
+
+    init: function() {
+      this.connection = new EventSource($location.protocol() + '://' + $location.host() + ':' + $location.port() + '/sse');
```
sandlerr: `window.location.origin` (if the port was the default it would be an empty string?)

zendesk-shender:
@grosser 
> idk ... lot's of complexity and moving parts for very little gain ... is that worth it ?
Yup, I was seeing this as the start for other things, like the active deploys page benefiting especially.

> Now we again do an extra request on every page load ?
We are currently doing polling requests every 5 seconds, so this is a big improvement IMO.

I'll add tests, and do some tidy up as suggested.

app/assets/javascripts/services/sse_factory.js
zendesk-shender: If the port was default, then it would still be '80', but I'll change it to the window.location.origin as that's cleaner.

app/controllers/deploys_controller.rb
zendesk-shender: Aye, the way the DeployService and Deploy classes are written/used makes it awkward to slot it in... I'll have a look.

zendesk-shender:
Ready for another review @zendesk/runway

app/models/deploy_service.rb
```diff
@@ -25,6 +26,11 @@ def confirm_deploy!(deploy, stage, reference, buddy = nil)
     end
   end
 
+  def stop_deploy!(deploy)
```
grosser: maybe just `stop!` since it's called DeployService :)

grosser:
:+1:

zendesk-shender:
Closing as I couldn't squash all my commits, so opening new PR #373

PR #363, opened 2015-04-21
Fixing staging deploy

fneves:
Fixing the path where the production.rb was being loaded. When assets where being pre-compiled the working dir would be ```tmp``` causing it to look on the wrong path.

/cc @zendesk/samson @grosser @zendesk-shender 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -1,1 +1,1 @@
-load(File.expand_path("../production.rb"))
+load(File.join(File.dirname(__FILE__), 'production.rb'))


zendesk-shender:
:+1: :+1: :+1:

config/environments/staging.rb
```diff
@@ -1 +1 @@
-load(File.expand_path("../production.rb"))
+load(File.join(File.dirname(__FILE__), 'production.rb'))
```
grosser: FYI I usually use this since I think it reads nicer / works nicely with deeply nested traversal:
`File.expand_path('../production.rb', __FILE__)`

fneves: Yeah that definitely reads better!

steved:
Can you symlink the two?

grosser:
https://github.com/zendesk/samson/pull/368

Issue #362, opened 2015-04-21
JavaScript compilation issues


mariovisic:
Getting this JavaScript error when compiling assets:

```
Uncaught Error: [$injector:modulerr] Failed to instantiate module samson due to:
Error: [$injector:unpr] Unknown provider: t
```

Seems to be due to the recent JavaScript changes, works fine running locally and not compiled.

sandlerr:
This works for me on master. What command did you run?

mariovisic:
`rake assets:precompile`

mariovisic:
Sorry, this happens once assets are precompiled, eg run `RAILS_ENV=production rake assets:precompile` then run the web server and it'll show that error in the JS

mariovisic:
Maybe try this: `rm -rf public/assets tmp/* && RAILS_ENV=production rake assets:precompile && RAILS_ENV=production rails server`

I can reproduce it currently on master. On 7a6ecde4b5790e516744eb56b2c30a3a1675c96d the issue is not present, so the issue would be somewhere here: https://github.com/zendesk/samson/compare/7a6ecde4b5790e516744eb56b2c30a3a1675c96d...master

mariovisic:
Have narrowed it down, the problem commit seems to be: 0ddbad3161a446167a8a55c3b17a468f0756c023

zendesk-shender:
Hmm... wonder if the ng-annotate gem is being circumvented with the assets changes,  it looks like the angularjs injector service is trying to inject 't' which probably indicates a minified dependency name.

zendesk-shender:
https://github.com/zendesk/samson/pull/364 should fix this.

zendesk-shender:
This looks resolved now! Enjoy! :)

mariovisic:
Awesome, works a treat, thanks @zendesk-shender :)

PR #364, opened 2015-04-21
Switch ng-annotate back to ruby version

zendesk-shender:
We can't use the JS version of this, as that was designed for nodejs projects. We need the rails specific version of this to run the ruby code in the compilation phase.

### References
 - https://github.com/zendesk/samson/issues/362

### Risks
 - None


fneves:
:+1: :+1:

PR #365, opened 2015-04-21
add hint for future cleanup

grosser:
@fneves just in case we forget and someone wants to clean that up

### Risks
 - None


fneves:
:+1:

PR #367, opened 2015-04-21
do less work when loading decorators

grosser:
@fneves 

before: each model load goes through all decorators and compares them to find the right one
after: each model load only goes through it's decorators

### Risks
 - None

### DIFF
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -39,14 +39,8 @@ module Samson
 
       def add_decorators
         Dir[decorators_root.join('**/*_decorator.rb')].each do |path|
-          Samson::Hooks.callback(:model_defined) do |model_klass_name|
-            require_dependency(path) if decorator_class(path) == model_klass_name
-          end
+          Samson::Hooks.decorator(decorator_class(path), path)
         end
-      end
-
-      def decorators_root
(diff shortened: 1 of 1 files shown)


grosser:
build is green, just travis bug not showing it ...

sandlerr:
:+1:

fneves:
I understand this simplifies things (in terms of optimisation it won't be such a big deal since we should not expect to have thousands of decorators).
I feel somehow strange to use two different models with hooks. One that uses events and callbacks another one that basically uses hooks to store and on demand do something (views and decorators). When I use hooks I am used to expect an event model, where I register a callback that will be triggered when the specific event is fired, and we are not doing it here, which can be confusing to someone looking at the code for the first time.

grosser:
yeah it's a little strange to use it for both. Ideally I'd like to only use it for data storage since it makes it easier to inspect, but that won't work with the dynamic nature of plugins.
I like that there is only 1 list of possible 'hooks' / we whitelist what can be added.

Exploring using callbacks for views here, do you think it's better ?
https://github.com/zendesk/samson/pull/374

Same could be done for decorators ... idk it feels like moving real methods into uninspectable lambdas is not a good move.

PR #368, opened 2015-04-21
use symlinks

grosser:
@steved @fneves 

was using load to easily add diffs / additions but either way is fine

### Risks
 - None


steved:
+1

PR #369, opened 2015-04-21
use profiler gems on staging

grosser:
this will also install them in production but not load them

@sandlerr 

### Risks
 - profiling being active in production


grosser:
works :)

screen shot 2015-04-21 at 1 05 57 pm

sandlerr:
:+1:

PR #370, opened 2015-04-21
refactor places where both deploy and stage passed as params

jonmoter:
`deploy belongs_to stage`, so there's no point in passing both of those,
and it's unclear what the behavior should be if you pass in a deploy
that doesn't belong to the stage.

### Risks
 - Code refactoring. Should not change any behavior, but touches a lot of files.


grosser:
:+1: :heart: sweet cleanup :)

PR #371, opened 2015-04-22
test new ruby

sandlerr:
let's start moving our deployment to 2.2.x


sandlerr:
@zendesk/samson

.travis.yml
```diff
@@ -3,7 +3,9 @@ cache:
   bundler: true
   directories:
     - node_modules
-rvm: 2.1.5
+rvm:
+  - 2.1.6
```
zendesk-shender: Why not leave this at 2.1.5 till we move to 2.2.2? The .ruby-version is still at 2.1.5, better to test with what we're running with.

sandlerr: @zendesk-shender I have a chef PR to move to 2.1.6 today, do you think we should skip that and go straight to 2.2.2?

zendesk-shender: I'm a sucker for the latest and greatest! :)
We should be able to test it out with 2.2.2 fairly easily. Samson is a relatively small tool and we should be able to roll it back if we run into trouble. We looking to update zendesk_voyager repo to 2.2.2 soon as well.

sandlerr:
@henders how is this?

henders:
:+1: from me, though I'm sweating a little! :)

PR #372, opened 2015-04-22
Error handling in the case the flowdock token is not set on the env

fneves:
Just better error handling when the FLOWDOCK_API_TOKEN is not set on the environment. Avoiding a 500 server error.

/cc @zendesk/samson @zendesk-shender 

### References
 - Jira link: 

### Risks
 - None


zendesk-shender:
:+1:

PR #374, opened 2015-04-22
WIP convert views to callbacks

grosser:


fneves:
:+1: I like this a lot more =)

grosser:
@zendesk/runway can we get 1 more opinion on data vs lambdas ?

grosser:
@zendesk/runway can I get 1 more opinion on data vs lambdas here ?

lib/samson/hooks.rb
```diff
@@ -78,7 +78,9 @@ def callback(name, &block)
       end
 
       def view(name, partial)
-        hooks(name) << partial
+        hooks(name) do |view, *args|
```
sandlerr: not sure I understand. Does `hooks(name)` return a method that will `yield`? So we are now storing the method itself rather than a string pointing to a view?

grosser: before we stored the name of the partial and did the rendering logic in render_views,
this would change it to store the rendering logic in the hook (to be consistent with other hooks that also only store lambdas)

consistency is nice, but simple inspectable data is too ... so choose wisely :)

sandlerr: I prefer simple inspectable data in this case

Issue #375, opened 2015-04-22
Questions around setting this up


lanceblais:
Hello, this is not so much an issue as an inquiry:

We're trying to get this running on Ubuntu 12.04 on AWS.

Once I got passed the required OS dependencies for the gems, I'm still having issues getting the configuration set up. 
I largely think the issues I'm having relate to not know the Rails config setup structure.
Could you help me understand which config files are needed to set samson up for production use?

Similarly, if there's any support for deploying this via chef / puppet / ansible / docker or even a detailed tutorial, that'd be great as well. 

We're really interested in getting it running!

sandlerr:
Once you've configured `database.yml` to suit the database you have installed you need to setup some environment variables. You can either do that by exporting them before starting the application or by copying [.env.example](https://github.com/zendesk/samson/blob/master/.env.example) to `SAMSON_DIR/.env` and setting values as you need. The only mandatory values are `SECRET_TOKEN` and either `GITHUB_TOKEN` or `AUTH_GITHUB=0`.

It should be possible in almost all cases to run Samson in production after modifying three files:

- database.yml for database settings
- .env for environment variables
- config/puma.rb for puma configuration. [Puma](https://github.com/puma/puma/) is the webserver we're using - to boot it using that file, run `puma -C config/puma.rb`

If you have a more specific issue that you are stuck on we would love to know where our [documentation](https://github.com/zendesk/samson#requirements) falls short.

lanceblais:
@sandlerr  Thanks for the info. Initially, I don't see the information about using Puma at all in the docs. The stuff around `.env` is there, but nothing about `database.yml` (I had to figure that one out myself).

Additionally, setting up the DB by running `bundle exec rake db:setup` (I think I need to use `RAILS_ENV=production` as well?

This is the stuff I'm not too sure about. 

It seems straightforward to setup on a local / development environment. Is Samson used per-developer environment, or shared in a hosted setting?

sandlerr:
Our intention is definitely for Samson to be run as a service for multiple users.

Yes, you should run `RAILS_ENV=production bundle exec rake db:setup` with a production block in `database.yml`.

lanceblais:
@sandlerr: I see the updates in the README which are great!

Was wondering: Is samson able to pull from private github repos?

sandlerr:
Yes, once a token is setup.

sandlerr:
[...]

PR #376, opened 2015-04-23
update readme

sandlerr:
To resolve missing info for #375 
@zendesk/samson

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,19 @@ open http://localhost:3000
  - Create a Stage
  - Deploy!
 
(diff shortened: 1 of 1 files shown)


fneves:
:+1:

Maybe we can do some kind of interactive script to make life easier for someone that has never setup a Rails app.

sandlerr:
We've already got a bootstrap script for the dev and test environments. Not sure about running an automated script for a production env, although we can definitely make things easier than they are currently.

PR #378, opened 2015-04-24
Change sse uri to match the current nginx rules.

henders:
A quick change to allow our nginx rules for both our SSE URIs.

### References
 - Jira link:

### Risks
 - None


fneves:
:+1:

PR #379, opened 2015-04-25
idea for log fetcher

grosser:
@zendesk/runway @steved 

after the main cap tasks have run (build container or deploy container)
listen to kafka and append whatever it has to say to the logs

I'll try to work with @bsandvik to get this running in staging / see if that works at all


plugins/server_logs/samson_slack/samson_plugin.rb
```diff
@@ -0,0 +1,18 @@
+require 'slack'
```
sandlerr: probably can remove this line and put this in `server_logs/lib/server_logs` because the constructor for a `Plugin` goes up 3 directories

grosser: fixed

steved:
I'd much rather this be a first-class citizen than a hack. I think there's a definite use-case (in general) for log streaming, especially if we someone ever wanted to split out command execution i.e. resque jobs. The deploy script is still going to have to have some notion of a job ending. For example, a health check passing or something like that.

grosser:
how about making the 'listen' part a command -> manually add it to command list `bin/listed_to <service-name>`, if we later end up with a model where there can be multiple jobs per deploy it would be easy to turn this into a different kind of job

steved:
That could work, though it would have to know to stop on a certain line or timeout. Might as well just add it to our cap deployment.

grosser:
lot more progress here, the place were I hook in is still a bit shady, but the kafka reading code is taking shape and is finally also tested :)

workflow is:
 - start listening for kafka messages on our topic
 - deploy
 - read the servers we deployed to from the deploy output
 - servers start sending their deploy messages + exit code as json, they all show up in the log
 - keep track of which servers are done
 - if all servers are done mark deploy as success
 - if not all servers respond in time mark deploy as failure

steved:
I'll review the code, but I think this plugin should stay private and separate until it's ready to be merged mainline (i.e. zendesk_samson_kafka or whatever).

plugins/server_logs/app/decorators/job_execution_decorator.rb
```diff
@@ -0,0 +1,13 @@
+# wait for reports from hosts to come back
+JobExecution.prepend(Module.new do
+  private
+
+  def execute_commands!(commands, &block)
```
steved: where does the block come from?

plugins/server_logs/app/decorators/job_execution_decorator.rb
```diff
@@ -0,0 +1,13 @@
...
+    if @stage.wait_for_server_logs
+      KafkaListener.new(@job.deploy_id, @output).listen(&block)
```
steved: shouldn't this be `listen { super }` or something?

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
...
+        print: false
+      }
+
+      kafka = listen_to_kafka(config)
+      server_listener = listen_for_server_list(config)
+
+      deploy_success = yield
+
+      # no servers found during deploy ?
+      unless config.fetch(:servers)
```
steved: should probably timeout+join the server_listener instead of killing it

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
...
+        server_listener.kill
+        write "did not find any servers in deploy log"
+        return false
+      end
+
+      # wait for this time after the deploy is done
+      config[:timeout] = AFTER_DEPLOY_TIMEOUT.from_now
+      config[:print] = :all
+
+      return false unless kafka.value
```
steved: Does value block?

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
...
+      when :new
+        received
+      else
+        []
+      end
+    end
+
+    # listen for all messages for the current deploy
+    def listen_for_messages(messages)
+      received = @consumer.fetch(max_wait_ms: 100).map!(&:value).map { |m| JSON.load(m) }
```
steved: map!.map is odd

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
...
+          print.each { |m| write m }
+
+          done_servers.concat(print.map { |m| m["host"] if m["exit"] == 0 }) # TODO handle other exit codes ...
+
+          # stop when we timeout or all servers are done
+          if Time.now > config.fetch(:timeout)
+            write "listener timed out"
+            break
+          elsif waiting_for_servers = config.fetch(:servers)
+            done_servers.each { |s| waiting_for_servers.delete(s) }
```
steved: `waiting - done`

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
...
+            done_servers.clear
+
+            if waiting_for_servers.none?
+              write "received messages from all hosts"
+              break
+            end
+          end
+        end
+
+        (config.fetch(:servers) || []).empty? # did we complete ?
```
steved: Won't this always be false if there are servers to listen to?

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
...
+          if Time.now > config.fetch(:timeout)
+            write "listener timed out"
+            break
+          elsif waiting_for_servers = config.fetch(:servers)
+            done_servers.each { |s| waiting_for_servers.delete(s) }
+            done_servers.clear
+
+            if waiting_for_servers.none?
+              write "received messages from all hosts"
+              break
```
steved: This should return maybe?

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
+require 'poseidon'
+
+module SamsonServerLogs
+  class KafkaListener
+    TOTAL_TIMEOUT = (Rails.env.test? ? 1.second : 30.minutes)
```
steved: cattr_accessors work nicely, because you can just set it in test

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
...
+    AFTER_DEPLOY_TIMEOUT = (Rails.env.test? ? 1.second : 10.minutes)
+    KAFKA_HOST = "kafka1.pod100" # TODO dynamic
```
steved: Might as well use ENV for now

plugins/server_logs/app/decorators/job_execution_decorator.rb
grosser: yep, fixed

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
steved: Ah I see, it's basically deleting them from the array

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
grosser: yes

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
```diff
@@ -0,0 +1,120 @@
...
+        received
+      else
+        []
+      end
+    end
+
+    # listen for all messages for the current deploy
+    def listen_for_messages(messages)
+      received = @consumer.fetch(max_wait_ms: 100).map!(&:value).map { |m| JSON.load(m) }
+      received.select! { |m| m["deploy"] == @deploy_id }
```
steved: I'm not 100% sure where this id will come from

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
grosser: damn .. too obvious :)

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
grosser: it will be false if there are servers left that did not report done, changed that to `@servers && @servers.empty?` so that it's false when no servers where found too

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
grosser: yep, much easier control flow now ...

plugins/server_logs/lib/samson_server_logs/kafka_listener.rb
grosser: We'll have to send it to the deploy script somehow ... but we need
something unique

PR #380, opened 2015-04-27
Fix issue where integrations created releases with old commits. Closes #359

### LINKED ISSUE #359: Handle webhooks in different orders
Currently when CI sends a webhook for a successful build on the release branch it creates a release for the passed in commit. In some circumstances it's possible for hooks to come in in reverse order. So for example I have two builds running very close together, build A and B. If build B finishes first for some reason, it's possible that A failed and was restarted, we'll get a webhook for B and then a webhook for A.

When this happens we get a release that displays no commits are present or pull requests, because the changeset will be B..A where A is an older commit than B.

I think in these circumstances one option would be to not create a release, when we know the commit on the webhook already features in the latest release.

Thoughts?

mariovisic:
If an integration is sent for a commit which is already included in a release,
it should not create a new release, this patch ensures that no new releases are
created when the latest release already contains the given commit.

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -76,6 +76,12 @@ class GitRepository
     valid
   end
 
+  def downstream_commit?(old_commit, new_commit)
+    cmd = "git rev-list --boundary #{old_commit}..#{new_commit}"
+    success, output = run_single_command(cmd) { |line| line.strip }
+    success ? output.blank? : false
+  end
+
   private
 
   def checkout!(executor: TerminalExecutor.new(StringIO.new), pwd: repo_cache_dir, git_reference:)
--- a/test/models/project_test.rb
+++ b/test/models/project_test.rb
@@ -16,19 +16,22 @@ describe Project do
     Project.create!(name: "hello", repository_url: url).token.wont_be_nil
   end
 
-  describe "#last_released_with_commit?" do
-    it "returns true if the last release had that commit" do
-      project.releases.create!(commit: "XYZ", author: author)
-      assert project.last_released_with_commit?("XYZ")
+  describe "#last_release_contains_commit?" do
+    let(:repository) { mock() }
+
+    before do
+      project.stubs(:repository).returns(repository)
+      repository.stubs(:downstream_commit?).with('LAST', 'NEW').returns(true)
     end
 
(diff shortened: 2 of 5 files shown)


test/models/project_test.rb
```diff
@@ -16,19 +16,17 @@ def clone_repository(project)
...
 
-    it "returns false if the last release had a different commit" do
-      project.releases.create!(commit: "123", author: author)
-      assert !project.last_released_with_commit?("XYZ")
+    before do
+      project.stubs(:repository).returns(repository)
+      repository.stubs(:downstream_commit?).with('LAST', 'NEW').returns(true)
     end
 
-    it "returns false if there have been no releases" do
```
sandlerr: can we keep this test? I'm pretty sure that's why there needs to be a guard around `last_release` in project.rb - in the case of the very first release `last_release` is nil

mariovisic: Ah yes, was an oversight, added back in.

mariovisic:
@sandlerr have added that test back in and added a new one for checking `downstream_commit?` when the commits given are the same.

app/models/git_repository.rb
```diff
@@ -76,6 +76,12 @@ def valid_url?
     valid
   end
 
+  def downstream_commit?(old_commit, new_commit)
+    cmd = "git rev-list --boundary #{old_commit}..#{new_commit}"
```
sandlerr: I think this could be a bit more descriptive?

we should be able to do something like:

`git merge-base --is-ancestor #{old_commit} #{new_commit}`

in which case we just use the value of `success` although it will be the inverse of what `git merge-base` actually returns because of how the TerminalExecutor works. When the old commit is an ancestor the git command returns 1, TerminalExecutor interprets it as an error (because it isn't 0) and sets success to false.

Are you ok with that?

mariovisic: Yes, I like the change, so long as the tests are green it's fine by me :)

sandlerr: If you don't mind making this change on your branch we can merge :+1:

mariovisic: Go for it :)

mariovisic:
Closing, see #382 382

PR #381, opened 2015-04-28
dockerise

sandlerr:
@jonmoter I was working on this last week, may be of some use

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
-FROM ruby:2.2.1
+FROM ruby:2.2.2
+
+RUN apt-get update
+RUN apt-get install -y nodejs
 
 RUN mkdir /app
 WORKDIR /app
@@ -9,7 +12,7 @@ ADD Rakefile /app/
 ADD bin /app/bin
 ADD public /app/public
 ADD db /app/db
-ADD vendor/assets /app/vendor/assets
+ADD .env.bootstrap /app/.env
 
 # Gems
 ADD Gemfile /app/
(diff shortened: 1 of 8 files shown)


sandlerr:
Something is up with mysql on travis, they're rolling back some changes

config/initializers/database.rb
```diff
@@ -3,7 +3,10 @@
 
   ActiveSupport.on_load(:active_record) do
     config = Rails.application.config.database_configuration[Rails.env]
-    config['pool'] = (ENV['DB_POOL'] || 100).to_i
-    ActiveRecord::Base.establish_connection(config)
+
+    if config
+      config['pool'] = (ENV['DB_POOL'] || 100).to_i
+      ActiveRecord::Base.establish_connection(config)
+    end
```
jonmoter: What are the scenarios where we'd want to run the database initializer but not attempt to connect to the database?

sandlerr: to precompile assets afaik. This is Steve D's work.

sandlerr:
cc @zendesk/samson

Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM ruby:2.2.2
+
+RUN apt-get update
+RUN apt-get install -y wget apt-transport-https
```
grosser: these should be 1 line or you end up in a place where the update is cached and the install fails

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# Code
+ADD config /app/config
+ADD app /app/app
+ADD lib /app/lib
+
+RUN DATABASE_URL=mysql2://user:pass@127.0.0.1/null RAILS_ENV=development PRECOMPILE=1 bundle exec rake --trace assets:precompile
+
+EXPOSE 3000
+
+CMD DATABASE_URL=$MYSQL_URL bundle exec rails server --binding=0.0.0.0
```
grosser: should this be puma ?

grosser:
:+1:

Dockerfile
sandlerr: `rails s` magically boots puma. We could make it explicit (and use `config/puma.rb`) but I'm not sure if that matters for development

grosser: this is meant for production too
I don't think rails s actually loads puma.rb ... ran into this issue on
another project ...

sandlerr: sure, will fix

PR #382, opened 2015-04-28
Fix issue where integrations created releases with old commits

sandlerr:
cc @zendesk/samson @mariovisic 

refactor of #380


mariovisic:
Looks good to me.

henders:
Looks good to me as well, not sure what happened to travis as only the mysql build failed with an auth error.
:+1: when green! :)

PR #383, opened 2015-04-28
make ruby hash syntax consistent across project

jonmoter:
Cleanup so we don't use some of one hash syntax, and some of another.

/cc @zendesk/runway

### Risks
 - None


grosser:
:+1:

PR #385, opened 2015-04-28
have Jenkins client use Rails logger instead of logging to STDOUT

jonmoter:
The Jenkins API client defaults to a logger pointing to stdout

/cc @zendesk/runway


sandlerr:
:+1:

PR #386, opened 2015-04-28
allow configuring statsd via ENV vars [RUN-103]

jonmoter:
/cc @zendesk/runway 

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-103

### Risks
 - None

### DIFF
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -1,13 +1,31 @@
 require 'statsd'
 
-config_file = Rails.root.join('config/statsd.yml')
-raise "No such file: config/statsd.yml" unless File.exist?(config_file)
+module StatsdLoader
+  def self.create
+    config = config_from_env || config_from_file
+    raise('Could not initialize statsd by file or by ENV') if config.nil?
+    Statsd.new(config[:host], config[:port])
+  end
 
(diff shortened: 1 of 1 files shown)


config/initializers/statsd.rb
```diff
@@ -1,13 +1,31 @@
 require 'statsd'
 
-config_file = Rails.root.join('config/statsd.yml')
-raise "No such file: config/statsd.yml" unless File.exist?(config_file)
+module StatsdLoader
+  def self.create
+    config = config_from_env || config_from_file
+    raise('Could not initialize statsd by file or by ENV') if config.nil?
```
sandlerr: you could also do `config = config_from_env || config_from_file || raise( ... )` but probably shouldn't

sandlerr:
:+1: although I kinda prefer making everything inherit from zendesk_configuration if possible

jonmoter:
@sandlerr, zendesk_configuration isn't open-sourced, whereas Samson is. We don't want to add that dependency.

config/initializers/statsd.rb
jonmoter: Yeah, I thought of that, but it didn't look/feel quite right.

PR #387, opened 2015-04-29
refactor statsdloader

grosser:
having a bit of fun with this class :)

 - private methods
 - not reading things twice from ENV
 - using fetch to blow up when Rails.env is not configured or host/port is missing from config
 - using straight up read to get a simple error "Errno::ENOENT: No such file or directory config/statsd.yml"

@zendesk/runway looking good ?

### Risks
 - None

### DIFF
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -1,31 +1,29 @@
 require 'statsd'
 
-module StatsdLoader
-  def self.create
-    config = config_from_env || config_from_file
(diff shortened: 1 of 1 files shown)


jonmoter:
Since you renamed the module, you have to change the call to `StatsdLoader.create` right below the module.

You *did* try running your code before creating the PR, right?  ;-)

grosser:
I only work in the github editor ;)

jonmoter:
Okay, fine by me.  :+1:

Issue #388, opened 2015-04-29
Warn/Lock messages are incorrectly normalized


shadowhand:
It looks like samson does a lowercase transform on all messages, which makes messages like this end up weird:

> Wait until iOS testing is complete

Ends up as:

> Wait until ios testing is complete

Or another example:

> Check with Tim before this goes out, QA wanted to do another check

Ends up as:

> Check with tim before this goes out, qa wanted to do another check

This normalization makes the output worse, not better.

grosser:
fixed by #389 

![screen shot 2015-05-06 at 10 42 43 am](https://cloud.githubusercontent.com/assets/11367/7499463/acdf9752-f3dc-11e4-87cf-759bd4a93c77.png)

PR #389, opened 2015-04-30
capitalise first letter only

sandlerr:
@zendesk/samson #388

### DIFF
--- a/app/models/lock.rb
+++ b/app/models/lock.rb
@@ -24,7 +24,8 @@ class Lock < ActiveRecord::Base
 
   def reason
     return "Description not given." if description.blank?
-    description.capitalize
+    description[0] = description[0].capitalize
+    description
   end
 
   private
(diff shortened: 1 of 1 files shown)


princemaple:
1__ruby
Am I missing something?

sandlerr:
@princemaple see the linked issue

princemaple:
Probably just get rid of capitalization. If people do cases properly, we don't need it.
By doing what you do now, there are still cases like "iTunes" will end up badly.

sandlerr:
You can just put a whitespace in front of the word if you don't want it to be capiltalised. Not a big deal in this case but prefer not to make surprising changes to existing users.

steved:
Just remove it

jwswj:
Agree - just remove it

princemaple:
:+1:

jwswj:
I like the old style. Reads easier than the ternary if. #nit

PR #390, opened 2015-04-30
do not load from all paths in lib and support plugins with camelcase nam...

grosser:
...es

@zendesk/runway 

### Risks
 - None


jonmoter:
:+1:

PR #391, opened 2015-04-30
strange fix

grosser:
@zendesk/runway not sure why ... but this works :D 

### Risks
 - None


jonmoter:
:+1: I'll create another PR to clean up some of the hooks/plugin code.

PR #394, opened 2015-05-05
Make Current Deploys page angularized like Recent Deploys

henders:
Current Deploys page is almost identical to Recent Deploys page. So I made both use the same templates and made current deploys page use angular to get dynamic updates whenever deploys are created/ended.

screencast mov

### References
 - Jira link:

### Risks
 - None


app/assets/javascripts/services/deploy_helpers_factory.js
```diff
@@ -0,0 +1,26 @@
+samson.factory('DeployHelper', function($window, $log, Deploys) {
```
grosser: idk ... do we really need infinite scrolling ... maybe just pagination ... also makes it easier to look at super old stuff and to share urls

grosser: oh you just moved that around ... still not a fan but ... never mind if it's old ...

app/models/deploy_service.rb
```diff
@@ -108,4 +109,10 @@ def update_github_deployment_status(deploy)
       GithubDeployment.new(deploy).update_github_deployment_status(@deployment)
     end
   end
+
+  def send_sse_deploy_update(type, deploy)
+    return unless deploy.persisted?
+    rebased_deploy = JSON.parse(DeploySerializer.new(deploy).to_json)["deploy"]
```
grosser: maybe expose a method that shows the hash before it is jsonified like .present / .to_h instead of generating and parsing :)

test/karma.conf.js
```diff
@@ -15,15 +15,17 @@ module.exports = function(config) {
...
       '<%= resolve_asset("angular-mocks.js") %>',
       '<%= resolve_asset("underscore.js") %>',
+      '<%= resolve_asset("moment.js") %>',
       'test/angular/test_helper.js',
       'app/assets/javascripts/app.js',
-      'app/assets/javascripts/controllers/**/*.js',
-      'app/assets/javascripts/directives/**/*.js',
+      'app/assets/javascripts/filters/**/*.js',
       'app/assets/javascripts/services/**/*.js',
+      'app/assets/javascripts/directives/**/*.js',
```
grosser: cannot be alphabetical ?

grosser:
:heart: the reuse, a bit sad about infinite scrolling ... but overall :+1:

dasch:
:heart: i the releases page next? :-D

test/karma.conf.js
henders: Actually, this change was unnecessary, I was trying to debug some problem and thought the order mattered. Turned out it didn't, but I forgot about this change. I'll change it back for aesthetics! :)

app/models/deploy_service.rb
henders: Thanks, after digging into the serializer a bit I figured it out. I'm still not sure how it magically works on the controller side without instantiation though.

app/assets/javascripts/timeline.js
```diff
@@ -142,9 +142,9 @@ samson.filter("localize",
       var localDate = new Date(parseInt(ms));
 
       var day    = DAYS[localDate.getDay()],
-          year   = localDate.getFullYear(),
-          date   = localDate.getDate(),
-          month  = MONTHS[localDate.getMonth()];
+        year   = localDate.getFullYear(),
```
sandlerr: IMO either don't indent or align with the first variable name

app/assets/javascripts/controllers/current_deploys_ctrl.js
```diff
@@ -0,0 +1,52 @@
...
+
+  function init() {
+    $http.get('/deploys/active.json').success(function(result) {
+      $scope.deploys = result.deploys;
+      updateUpdatedAt();
+    });
+  }
+
+  $scope.addDeploy = function(deploy) {
+    $scope.deploys = [deploy].concat($scope.deploys);
```
sandlerr: does prepending rather than appending affect the displayed order?

henders: Yup its an array so the angular directive displays it in the same order.

app/assets/javascripts/timeline.js
henders: I've never seen the dont-indent style, but I'll change the indent back. My
IDE decided to do this anyway! :)

PR #373, opened 2015-04-22
Use SSEs to broadcast changes in Deployments so clients can get realtime updates via JS

zendesk-shender:
Instead of polling every 5 seconds for number of active deploys, use SSEs to broadcast changes to connected users.

/cc @zendesk/runway @zendesk/samson

### References
 - Jira link:

### Risks
 - None

PR #395, opened 2015-05-05
Fix Current Deploys badge

henders:
Active Deploys route is paginated, so current deploys badge would never go above 30.

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -1,7 +1,7 @@
 class DeploysController < ApplicationController
   before_action :authorize_deployer!, only: [:new, :create, :confirm, :update, :destroy, :buddy_check, :pending_start]
   before_action :find_project
-  before_action :find_deploy, except: [:index, :recent, :active, :new, :create, :confirm]
+  before_action :find_deploy, except: [:index, :recent, :active, :active_count, :new, :create, :confirm]
   before_action :stage, only: :new
 
   def index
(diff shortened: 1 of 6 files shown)


app/controllers/deploys_controller.rb
```diff
@@ -125,4 +128,8 @@ def find_project
   def find_deploy
     @deploy = Deploy.find(params[:id])
   end
+
+  def active_deploy_scope
```
grosser: might be more reusable with `def deploy_scope; @project.try(:deploys) || Deploy` and then `deploy_scope.active` but either way is fine

grosser:
if it's paginated then the pagination could just include the count like this: https://developer.zendesk.com/rest_api/docs/core/introduction#pagination

henders:
@grosser 
Sending all the deploy details down to the client seemed overkill and I don't know if I want to roll my own pagination implementation. It looks like Kaminari doesn't support shoving in the total count.

grosser:
:+1: separate endpoint is certainly good for performance reasons :)

if we run into other places then we should take a look at how to make our json nicely paginate

jwswj:
:+1:

app/views/shared/_header.html.erb
```diff
@@ -46,7 +46,7 @@
         <li class="current-deploys <%= 'active' if current_page?(controller: '/deploys', action: 'active') %>" ng-controller="currentDeployBadgeCtrl">
           <%= link_to active_deploys_path do %>
             Current Deploys
-            <%= content_tag(:span, '{{ currentActiveDeploys }}', class: 'badge badge-deploys', style: 'display: none', id: 'currentDeploysBadge', data: { url: active_deploys_url(format: 'json') }) %>
```
sandlerr: wasn't this used for polling?

sandlerr: #373 nope :+1:

app/controllers/deploys_controller.rb
henders: I'll leave it for now.

PR #397, opened 2015-05-06
samson env plugin

grosser:
@jonmoter simple env plugin to kick things of ... not sure if .env is the best file to use, but it kind of makes sense ...

@steved thoughts ?

### Risks
 - None

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150520210103) do
+ActiveRecord::Schema.define(version: 20150530010900) do
 
   create_table "build_statuses", force: :cascade do |t|
     t.integer  "build_id",   null: false
@@ -81,6 +81,22 @@ ActiveRecord::Schema.define(version: 20150520210103) do
   add_index "deploys", ["deleted_at"], name: "index_deploys_on_deleted_at"
   add_index "deploys", ["job_id", "deleted_at"], name: "index_deploys_on_job_id_and_deleted_at"
   add_index "deploys", ["stage_id", "deleted_at"], name: "index_deploys_on_stage_id_and_deleted_at"
+
+  create_table "environment_variable_groups", force: :cascade do |t|
+    t.string "name", limit: 255, null: false
+  end
+
+  add_index "environment_variable_groups", ["name"], name: "index_environment_variable_groups_on_name", unique: true, using: :btree
+
+  create_table "environment_variables", force: :cascade do |t|
+    t.string  "name",            limit: 255, null: false
+    t.string  "value",           limit: 255, null: false
+    t.integer "parent_id",       limit: 4,   null: false
+    t.string  "parent_type",     limit: 255, null: false
+    t.integer "deploy_group_id", limit: 4
+  end
+
+  add_index "environment_variables", ["parent_id", "parent_type", "name", "deploy_group_id"], name: "environment_variables_unique_deploy_group_id", unique: true, using: :btree
 
   create_table "environments", force: :cascade do |t|
     t.string   "name",                          null: false
@@ -218,6 +234,14 @@ ActiveRecord::Schema.define(version: 20150520210103) do
     t.datetime "created_at"
     t.datetime "updated_at"
   end
+
+  create_table "stage_environment_variable_groups", force: :cascade do |t|
+    t.integer "stage_id",                      limit: 4, null: false
+    t.integer "environment_variable_group_id", limit: 4, null: false
+  end
+
+  add_index "stage_environment_variable_groups", ["environment_variable_group_id"], name: "stage_environment_variable_groups_group_id", using: :btree
+  add_index "stage_environment_variable_groups", ["stage_id", "environment_variable_group_id"], name: "stage_environment_variable_groups_unique_group_id", unique: true, using: :btree
 
   create_table "stages", force: :cascade do |t|
     t.string   "name",                                                                       null: false
--- a/plugins/env/app/decorators/stage_decorator.rb
+++ b/plugins/env/app/decorators/stage_decorator.rb
@@ -1,0 +1,6 @@
+Stage.class_eval do
+  include AcceptsEnvironmentVariables
+
+  has_many :stage_environment_variable_groups
+  has_many :environment_variable_groups, through: :stage_environment_variable_groups
+end
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -1,0 +1,16 @@
+class EnvironmentVariable < ActiveRecord::Base
+  belongs_to :parent, polymorphic: true
+  belongs_to :deploy_group
+  validates :name, presence: true
+
+  def self.env(stage, deploy_group_id)
+    variables = stage.environment_variables + stage.environment_variable_groups.flat_map(&:environment_variables)
+    variables.each_with_object({}) do |ev, all|
+      if !ev.deploy_group_id
+        all[ev.name] ||= ev.value
+      elsif ev.deploy_group_id == deploy_group_id
+        all[ev.name] = ev.value
+      end
+    end
+  end
+end
--- a/plugins/env/app/models/environment_variable_group.rb
+++ b/plugins/env/app/models/environment_variable_group.rb
@@ -1,0 +1,10 @@
+class EnvironmentVariableGroup < ActiveRecord::Base
+  include AcceptsEnvironmentVariables
+
+  default_scope -> { order(:name) }
+
+  has_many :stage_environment_variable_groups
+  has_many :stages, through: :stage_environment_variable_groups
+
+  validates :name, presence: true
+end
--- a/plugins/env/app/models/stage_environment_variable_group.rb
+++ b/plugins/env/app/models/stage_environment_variable_group.rb
@@ -1,0 +1,4 @@
+class StageEnvironmentVariableGroup < ActiveRecord::Base
+  belongs_to :stage
+  belongs_to :environment_variable_group
+end
--- a/plugins/env/db/migrate/20150527210505_add_environment_variables.rb
+++ b/plugins/env/db/migrate/20150527210505_add_environment_variables.rb
@@ -1,0 +1,24 @@
+class AddEnvironmentVariables < ActiveRecord::Migration
+  def change
+    create_table :environment_variables do |t|
+      t.string :name, null: false
+      t.string :value, null: false
+      t.integer :parent_id, null: false
+      t.string :parent_type, null: false
+      t.integer :deploy_group_id
+    end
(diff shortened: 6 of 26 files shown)


grosser:
screen shot 2015-05-06 at 10 41 29 am

steved:
What are you going to put in here?
How do we handle sharing of configuration changes among different applications?

grosser:
For starters I think public configuration only, later maybe secret things but only via indirection (look it up in secret-store under key xyz) or via some acl system ...

To get shared configuration we'd need something like the stage/command building thing, but I'm trying to keep it simple for now since I'm not even sure if we end up using this.

steved:
So then what benefits does this have over keeping it in the repo? At least it's versioned there.

grosser:
Good point, for different stages and .env.staging etc would also work.

Benefits would be:
 - (later) reuse of configuration across projects (might also be possible on project level ...)
 - not adding staging/production config to projects
 - quick change without committing new code

Downsides:
 - env is unversioned
 - env might drift from what the project needs to be run
 - env might need to keep in sync with project version

so yeah not a clear winner ... for samson itself it would be nice since we deploy a project we cannot put secrets or specific configs into ... but that's also fixable via other ways ...

sandlerr:
I'm really hoping we will allow deployers to configure the environment, not just admins

grosser:
edit is restricted to admins, so that won't work :)
still better than going through ops ...

jonmoter:
Yeah, there are a few problems with this approach.

1. No way to share config across projects (e.g. Airbrake API key, redis hosts, etc.)
2. Shouldn't use for secrets
3. No way to have different config for each pod. Several projects have a single stage for deploying to multiple pods, and something like the DATABASE_URL will have a different value for each of the pods. So you can't have a single .env file that will work across multiple pods.

Especially due to point 3, I'm skeptical about getting this ENV management in Samson to play nicely with Capistrano deploys. It's necessary for docker deploys, but I haven't yet found an elegant way to do that with cap deploys.

grosser:
Sharing is easily added by copying the way we do deploy scripts.
Doing service discovery from a remote pod seems problematic to me  (samson
-> pod1/2/3/4) -> leave it to the runner ?
Secrets could be more like instructions "Lookup key xyz in secret store"
and be executed by the runner.
Since that moves a bunch of logic to the runner anyway we might as well
check in the list of required env vars and let the runner build it.

grosser:
also added ENV.json support so we can play with it / use it now.

screen shot 2015-05-08 at 11 51 45 am

grosser:
added support for groups

screen shot 2015-05-27 at 3 37 52 pm
screen shot 2015-05-27 at 3 35 20 pm
screen shot 2015-05-27 at 4 57 48 pm

produces ENV.json + ENV.podx.json for each deploy group of the stage

still need more tests, but looks good ?

grosser:
added tests

looking good ?

henders:
The Environment Group pages look cool. :)

For the stage edit the textarea env config now looks odd in comparison, and IMO the table based approach on the manage groups page looks better.

db/schema.rb
```diff
@@ -47,6 +47,21 @@
...
+  create_table "environment_variable_groups", force: :cascade do |t|
+    t.string "name", limit: 255, null: false
+  end
+
+  add_index "environment_variable_groups", ["name"], name: "index_environment_variable_groups_on_name", unique: true, using: :btree
+
+  create_table "environment_variables", force: :cascade do |t|
+    t.string  "name",                          limit: 255, null: false
+    t.string  "value",                         limit: 255, null: false
+    t.string  "environment_variable_group_id", limit: 255, null: false
```
henders: t.belongs_to instead of a string FK

plugins/env/db/migrate/20150527210505_add_environment_variables.rb
```diff
@@ -0,0 +1,22 @@
+class AddEnvironmentVariables < ActiveRecord::Migration
+  def change
+    create_table :environment_variables do |t|
+      t.string :name, null: false
+      t.string :value, null: false
+      t.string :environment_variable_group_id, null: false
```
henders: t.belongs_to instead of string FK.

plugins/env/db/migrate/20150527210505_add_environment_variables.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.integer :deploy_group_id
```
henders: same here

plugins/env/lib/samson_zendesk/samson_plugin.rb
```diff
@@ -0,0 +1,41 @@
...
+  {:environment_variable_group_ids => []}
+end
+
+Samson::Hooks.callback :after_deploy_setup do |dir, stage|
+  if stage.env.present? || stage.environment_variable_groups.any?
+    # TODO also include everything here / write multiple of these
+    File.write("#{dir}/.env", stage.env)
+
+    # propriatary manifest format
+    manifest = "#{dir}/ENV.json"
```
henders: Do we want to force people to declare a ENV.json manifest file? Maybe leave it up to our process to encourage it, and if it's not there just write the .env files anyway. I would like that as a user if I'm still in the process of creating a project/stage setup in Samson and haven't nailed down all the details yet. Otherwise I'd have to keep doing PRs in the github repo while twiddling settings.

plugins/env/app/models/environment_variable_group.rb
```diff
@@ -0,0 +1,18 @@
...
+  has_many :stage_environment_variable_groups
+  has_many :stages, through: :stage_environment_variable_groups
+
+  validates :name, presence: true
+  accepts_nested_attributes_for :environment_variables, allow_destroy: true
+
+  def self.to_hash(stage, deploy_group_id)
+    stage.environment_variable_groups.each_with_object({}) do |ev_group, all|
+      ev_group.environment_variables.select { |ev| ev.deploy_group_id == deploy_group_id }.each do |ev|
+        all[ev.name] = ev.value
```
henders: Are the 'all' scoped varables included previously to this hash?

plugins/env/db/migrate/20150527210505_add_environment_variables.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+    add_index :environment_variables, [:environment_variable_group_id, :name, :deploy_group_id], unique: true, name: "environment_variables_unique_deploy_group_id"
+
+    create_table :environment_variable_groups do |t|
+      t.string :name, null: false
+    end
+    add_index :environment_variable_groups, :name, unique: true
+
+    create_table :stage_environment_variable_groups do |t|
+      t.integer :stage_id, :environment_variable_group_id, null: false
```
henders: Rails HABTM table format without an ID field
```
create_table :stage_environment_variable_groups, id: false do |t|
      t.belongs_to :environment_variable_group, index: true
      t.belongs_to :stage, index: true
    end
```

plugins/env/app/models/environment_variable_group.rb
```diff
@@ -0,0 +1,18 @@
+class EnvironmentVariableGroup < ActiveRecord::Base
+  default_scope -> { order(:name) }
+
+  has_many :environment_variables
+  has_many :stage_environment_variable_groups
+  has_many :stages, through: :stage_environment_variable_groups
```
henders: You don't have to specify the through table if you use the HABTM relationship. See DeployGroups to Stages
```
  has_and_belongs_to_many :environment_variable_groups
```

plugins/env/app/models/stage_environment_variable_group.rb
```diff
@@ -0,0 +1,4 @@
+class StageEnvironmentVariableGroup < ActiveRecord::Base
+  belongs_to :stage
+  belongs_to :environment_variable_group
+end
```
henders: Don't need this class if using the HABTM relationship.

plugins/env/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,6 @@
+Stage.class_eval do
+  has_many :stage_environment_variable_groups
+  has_many :environment_variable_groups, through: :stage_environment_variable_groups
```
henders: You don't have to specify the through table if you use the HABTM relationship. See DeployGroups to Stages
```
  has_and_belongs_to_many :environment_variable_groups
```

henders:
Oh yea, another advantage of using the HABTM relationship between groups and stages is you can easily do the reverse lookup to see what stages are related to a particular group. (nice to show that if editing a group).

grosser:
Yeah at the end it felt weird to have 2 different interfaces, I'll turn
that into a table too

plugins/env/lib/samson_zendesk/samson_plugin.rb
grosser: Yeah I want to write out both, just needs a bit of refactoring love :)

plugins/env/app/models/environment_variable_group.rb
grosser: the deploy_group specific settings are merged into the 'all' settings to
not have to build them multiple times

plugins/env/app/models/environment_variable_group.rb
grosser: I prefer concrete models over habtm since most of the time some logic ends
up there anyway + it's 1 less abstraction / way of handling things and
allows me to do some efficient queries like stage_environment_variable_
groups.pluck(:stage_id)

plugins/env/db/migrate/20150527210505_add_environment_variables.rb
grosser: I'm using a unique index on both for the first one ... but yeah same reason: I want to see the actual structure and not the railsy abstraction

grosser:
Can also do that without, just needs another has_many :stages, through :)

grosser:
New index page with stages
screen shot 2015-05-28 at 1 48 34 pm

New stage ui with table:
screen shot 2015-05-28 at 1 50 09 pm

and generating .env files for all deploy groups

it no longer merges them with the existing settings but blows up if you did not specify all required keys + ignores keys that were not specified

any concerns still remaining ?

grosser:
Later this can set consul keys or whatever instead of writing these ENV.json files, but that should be easy ...

plugins/env/app/models/environment_variable.rb
```diff
@@ -0,0 +1,16 @@
+class EnvironmentVariable < ActiveRecord::Base
+  belongs_to :parent, polymorphic: true
+  belongs_to :deploy_group
```
jonmoter: Does this mean that if I want to have AIRBRAKE_API_KEY defined, I would need to manually add the same value to each deploy group?  (pod98-101, pod1-6, gamma1-6)

grosser: nope, you just add it to the "All" environment variable where deploy_group is nil

jonmoter:
This looks good to me. Let's merge it, play around with it some, and see what works about it, and what we want to tweak.

:+1:

grosser:
I'll get rosetta converted ... let's see if that works fine :)

PR #398, opened 2015-05-06
show commands that are being executed

grosser:
@zendesk/runway @steved 

Before:
screen shot 2015-05-06 at 11 21 06 am

After:
screen shot 2015-05-06 at 11 42 28 am

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -12,7 +12,7 @@ class GitRepository
   end
 
   def setup!(executor, temp_dir, git_reference)
-    executor.output.write("Beginning git repo setup\n")
+    executor.output.write("# Beginning git repo setup\n")
     return false unless clone!(executor: executor, from: repository_url, to: repo_cache_dir, mirror: true) unless locally_cached?
     return false unless update!(executor: executor)
     return false unless clone!(executor: executor, from: repo_cache_dir, to: temp_dir)
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -17,7 +17,7 @@ class JobExecution
 
   def initialize(reference, job)
     @output = OutputBuffer.new
-    @executor = TerminalExecutor.new(@output)
+    @executor = TerminalExecutor.new(@output, verbose: true)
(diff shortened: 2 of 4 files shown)


steved:
I still don't think it's necessary

grosser:
I think the advantages are:
 - can see where output came from with multiple debug statements
 - can see what failed when having the same command twice
 - easy to track down where extra warnings etc came from
 - less if/then/retval command injection
 - easier to correlate log output with deploy script you see in stage show/edit

and disadvantages:
 - more output / more noise

steved:
These are all debugging tools, on a daily basis no one cares about any of this. Maybe just add it to the log you can download.

grosser:
I look at the logs when something failed and getting instant feedback would
be great then. In the 99% case it is cap failing so the command won't be in
the way anyway since it's 100 lines above the failure cause, but for the 1%
case of just doing a new setup / setup failing having the commands is
helpful for debugging.


On Wed, May 6, 2015 at 11:44 AM, Steven Davidovitz <notifications@github.com

gabetax:
:+1: for me. I've frequently wanted this output immediately available when debugging failed builds or changes to the deployment scripts. It's not a significant amount of extra text in the output, and I don't think it would be obvious to check a separate downloadable log of it. Plus, we could build in output folding like travis has.

henders:
:+1:  for me too as I've often wondered what cmd generated which output during more complicated scripts, e.g. which conditional branch cmds were executed.

jonmoter:
Couldn't you just add a `set -x` command at the top if verbose == true?

henders:
+1 to the set -x idea, a bit more standard!

grosser:
I picked -x since it gives us '+ command' vs just 'command' from -v

grosser:
meh ... -x is just `+command` on unix vs `+ command` on mac ...

grosser:
-x and -v also make all subcommands log their commands which gets super messy ... reverted to the echo method

grosser:
confirmed on staging, looks good :)

jonmoter:
:+1:

PR #402, opened 2015-05-08
fix "JIRA Issues" tab on deploy details page

jonmoter:
clicking on the tab wasn't making an AJAX call to load the changeset

/cc @zendesk/runway


henders:
:+1:

PR #403, opened 2015-05-08
Pass Deploy_groups into Deploy script for use.

henders:
To further the integration of deploy groups, this passes down the list of affected deploy_groups in an environment variable $DEPLOY_GROUPS so the script can access it.

screen shot 2015-05-08 at 15 05 18

----------------------

pod1_deploy__succeeded__-_project0

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -101,28 +101,19 @@ class JobExecution
 
     FileUtils.mkdir_p(artifact_cache_dir)
     @output.write("\n# Executing deploy\n")
+    @output.write("# Deploy URL: #{@job.deploy.url}\n") if @job.deploy
 
-    commands = [
-      "export DEPLOYER=#{@job.user.email.shellescape}",
-      "export DEPLOYER_EMAIL=#{@job.user.email.shellescape}",
-      "export DEPLOYER_NAME=#{@job.user.name.shellescape}",
-      "export REVISION=#{@reference.shellescape}",
-      "export TAG=#{(@job.tag || @job.commit).to_s.shellescape}",
-      "export CACHE_DIR=#{artifact_cache_dir}",
-      "cd #{dir}",
-      *@job.commands
-    ]
-
+    cmds = commands(dir)
     payload = {
       stage: (stage.try(:name) || "none"),
       project: @job.project.name,
-      command: commands.join("\n")
+      command: cmds.join("\n")
     }
 
     ActiveRecord::Base.clear_active_connections!
 
     ActiveSupport::Notifications.instrument("execute_shell.samson", payload) do
-      payload[:success] = @executor.execute!(*commands)
+      payload[:success] = @executor.execute!(*cmds)
     end
   end
 
@@ -140,6 +131,23 @@ class JobExecution
       @output.write("Could not get exclusive lock on repo. Maybe another stage is being deployed.\n")
       false
     end
+  end
+
+  def commands(dir)
+    commands = [
+      "export DEPLOYER=#{@job.user.email.shellescape}",
+      "export DEPLOYER_EMAIL=#{@job.user.email.shellescape}",
+      "export DEPLOYER_NAME=#{@job.user.name.shellescape}",
+      "export REVISION=#{@reference.shellescape}",
+      "export TAG=#{(@job.tag || @job.commit).to_s.shellescape}",
+      "export CACHE_DIR=#{artifact_cache_dir}",
+      "cd #{dir}",
+      *@job.commands
+    ]
+    if @stage && group_names = @stage.deploy_groups.pluck(:name).sort.map!(&:shellescape).join(" ")
+      commands.unshift("export DEPLOY_GROUPS='#{group_names}'") if group_names.presence
+    end
+    commands
   end
 
   def artifact_cache_dir
--- a/test/models/job_execution_test.rb
+++ b/test/models/job_execution_test.rb
@@ -4,12 +4,15 @@ describe JobExecution do
 
   let(:repository_url) { Dir.mktmpdir }
   let(:repo_dir) { File.join(GitRepository.cached_repos_dir, project.repository_directory) }
+  let(:pod1) { deploy_groups(:pod1) }
+  let(:pod2) { deploy_groups(:pod2) }
+  let(:deploy_group_hack) { DeployGroup.create!(name: ';|sudo make-sandwich /', environment: pod1.environment) }
   let(:project) { Project.create!(name: 'duck', repository_url: repository_url) }
-  let(:stage) { Stage.create!(name: 'stage4', project: project) }
+  let(:stage) { Stage.create!(name: 'stage4', project: project, deploy_groups: [pod1, pod2, deploy_group_hack]) }
   let(:user) { User.create!(name: 'test') }
   let(:job) { project.jobs.create!(command: 'cat foo', user: user, project: project) }
   let(:execution) { JobExecution.new('master', job) }
-  let(:deploy) { Deploy.create!(stage: stage, job: job, reference: 'masterCADF') }
+  let(:deploy) { Deploy.create!(stage: stage, job: job, reference: 'master') }
 
   before do
     Project.any_instance.stubs(:valid_repository_url).returns(true)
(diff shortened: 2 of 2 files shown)


grosser:
don't they need to be lowercase to be used in cap ?

app/models/job_execution.rb
```diff
@@ -101,6 +101,7 @@ def execute!(dir)
 
     FileUtils.mkdir_p(artifact_cache_dir)
     @output.write("Executing deploy\n")
+    @output.write("Deploy URL: #{@job.deploy.url}\n") if @job.deploy
```
grosser: Isn't the log only shown on the deploy page anyway ?

app/models/job_execution.rb
```diff
@@ -109,6 +110,7 @@ def execute!(dir)
       "export REVISION=#{@reference.shellescape}",
       "export TAG=#{(@job.tag || @job.commit).to_s.shellescape}",
       "export CACHE_DIR=#{artifact_cache_dir}",
+      "export DEPLOY_GROUPS='#{@stage.deploy_groups.map(&:name).join(' ') if @stage}'",
```
grosser: setting an empty set would be a little bit of clutter for non-deploy-grouped deploys
->

```Ruby
xxx = []
if @stage && deploy_groups = @stage.deploy_groups.map(&:name).join(' ').presence
  xxx << "export DEPLOY_GROUPS='#{deploy_groups}'"
end
```

grosser:
:+1: except for the also exporting empty

henders:
@grosser , yes, but I figure the script can take care of lowercasing them. Didn't want to change the case if someone else besides Zendesk needs it to exact match the DeployGroup name.

So they would use:
```
PODS = `echo $DEPLOY_GROUPS | | awk '{print tolower($0)}'`
cap $PODS deploy
```

app/models/job_execution.rb
henders: Yea, @pswadi-zendesk requested this for auditing purposes in the runway flow!

grosser:
hmm technically more correct, but a bit messy ... we can just change our deploy groups to be lowercase I guess.

app/models/job_execution.rb
sandlerr: probably also want to shell escape these. Not sure what would happen if they had a space or single quote

henders: Good point! I'll do that.

jonmoter:
Hey @henders, can you remind me of the use case for this? I know we spoke about it at some point, but I've forgotten how this will be used.

henders:
@jonmoter 
It's integrating DeployGroup functionality further into Samson. At the moment it's just visual fluff, but this change allows the stage to just do 'cap $DEPLOY_GROUPS deploy' instead of enumerating the deploy groups manually when they've already ticked the appropriate deploy group boxes.

henders:
@zendesk/runway @grosser 
Looking for another +1 on this since I refactored some code which still looks a bit off.

test/models/job_execution_test.rb
```diff
@@ -125,6 +127,15 @@
     lines.must_include "TAG=v1"
   end
 
+  it 'exports deploy_group information if deploy groups present' do
+    job.update(command: 'env')
+    job.deploy = deploy
+    execute_job
+    lines = job.output.split "\n"
+    lines.must_include "# Deploy URL: #{deploy.url}"
+    lines.must_include "DEPLOY_GROUPS=Pod1 Pod2"
```
jonmoter: Shouldn't there be single-quotes surrounding "Pod1 Pod2"?

app/models/job_execution.rb
```diff
@@ -101,28 +101,18 @@ def execute!(dir)
...
-      "cd #{dir}",
-      *@job.commands
-    ]
+    @output.write("# Deploy URL: #{@job.deploy.url}\n") if @job.deploy
 
     payload = {
       stage: (stage.try(:name) || "none"),
       project: @job.project.name,
-      command: commands.join("\n")
+      command: commands(dir).join("\n")
```
grosser: hmm calling commands twice is a bit strange ... maybe store it in some local variable ?

grosser: yeah ... it's cached but has an argument ... that smells like trouble :)
-> no cache but store as local variable

app/models/job_execution.rb
```diff
@@ -142,6 +132,25 @@ def setup!(dir)
...
+        "export DEPLOYER=#{@job.user.email.shellescape}",
+        "export DEPLOYER_EMAIL=#{@job.user.email.shellescape}",
+        "export DEPLOYER_NAME=#{@job.user.name.shellescape}",
+        "export REVISION=#{@reference.shellescape}",
+        "export TAG=#{(@job.tag || @job.commit).to_s.shellescape}",
+        "export CACHE_DIR=#{artifact_cache_dir}",
+        "cd #{dir}",
+        *@job.commands
+      ]
+      if @stage && @stage.deploy_groups.count > 0
```
grosser: saving an extra query and map! for a bit of performance win:
`if @stage && group_names = @stage.deploy_groups.pluck(:name).map!(&:shellescape).join(" ")`

app/models/job_execution.rb
```diff
@@ -142,6 +132,25 @@ def setup!(dir)
...
+        commands.insert(0, "export DEPLOY_GROUPS='#{@stage.deploy_groups.map(&:name).map(&:shellescape).join(' ')}'")
```
grosser: FYI insert 0 == unshift

test/models/job_execution_test.rb
```diff
@@ -125,6 +127,15 @@
...
+  end
+
```
grosser: a test shellescape ?

grosser: test for not having deploy groups -> no DEPLOY_GROUPS set

grosser:
we'll have to change our names deploy group names to lowercase to be able to use them with cap then or lowercase them in cap ?

gabetax:
We could leave it to the command to do any translation necessary: http://stackoverflow.com/a/2264537/119479 - janky, but flexible.

Alternatively, we could introduce this concept within the `HostGroup` model as a value that can be explicitly defined.

grosser:
meh just renaming them seems easier :)

test/models/job_execution_test.rb
henders: This is matching against the output of 'env' which won't have the quotes in them.

henders:
@grosser - addressed comments!

app/models/job_execution.rb
henders: Ah, thanks! :)

app/models/job_execution.rb
henders: Yea, good call... wasn't liking this smell.

app/models/job_execution.rb
```diff
@@ -142,6 +133,23 @@ def setup!(dir)
...
+      "export DEPLOYER=#{@job.user.email.shellescape}",
+      "export DEPLOYER_EMAIL=#{@job.user.email.shellescape}",
+      "export DEPLOYER_NAME=#{@job.user.name.shellescape}",
+      "export REVISION=#{@reference.shellescape}",
+      "export TAG=#{(@job.tag || @job.commit).to_s.shellescape}",
+      "export CACHE_DIR=#{artifact_cache_dir}",
+      "cd #{dir}",
+      *@job.commands
+    ]
+    if @stage && group_names = @stage.deploy_groups.pluck(:name).map!(&:shellescape).join(" ")
```
grosser: `.presence` or it will assign an empty string when empty

test/models/job_execution_test.rb
```diff
@@ -123,6 +126,16 @@
...
+    job.output.wont_include "DEPLOY_GROUPS"
+  end
+
+  it 'exports deploy_group information if deploy groups present' do
+    job.update(command: 'env | sort')
+    job.deploy = deploy
+    execute_job
+    lines = job.output.split "\n"
+    lines.must_include "# Deploy URL: #{deploy.url}"
+    lines.must_include 'DEPLOY_GROUPS=\;\|sudo\ make-sandwich\ / Pod1 Pod2'
```
grosser: a `it "does not add deploy group if deploy groups are missing"` would be nice :)

grosser:
:+1:

PR #406, opened 2015-05-12
[WIP] add per-project user permissions [RUN-117]

jonmoter:
add ability to grant users admin-level permissions to a single project, rather than globally.
so far I've updated the database and models.

TODO:
- [ ] UI to show/add/remove user permissions on Project Settings page
- [ ] UI to show/add/remove project permissions on Users admin page

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-117

### Risks
 - Change in behavior in permissions code; confirm we haven't added any security holes.

### DIFF
--- a/app/models/concerns/permalinkable.rb
+++ b/app/models/concerns/permalinkable.rb
@@ -10,6 +10,12 @@ module Permalinkable
   module ClassMethods
     def find_by_param!(param)
       find_by_permalink!(param)
+    rescue ActiveRecord::RecordNotFound
+      if param =~ /^\d+$/
+        find_by_id!(param)
+      else
+        raise
+      end
     end
   end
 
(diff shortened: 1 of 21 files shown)


app/models/concerns/permalinkable.rb
```diff
@@ -10,6 +10,12 @@ module Permalinkable
   module ClassMethods
     def find_by_param!(param)
       find_by_permalink!(param)
+    rescue ActiveRecord::RecordNotFound
```
grosser: we can just call find_by_id for models that have a id-permalink otherwise you might up in weird places where a stage is named just 1 ... or I guess disallow all number permalinks ...

henders:
How were you envisioning the UI component of the permissions? Overhaul the current user-perms page?

jonmoter:
I haven't thought it through really fully.  We'd probably want to have an `admin/users#show` action where you could look at a single user, see a list of all the projects they have permissions to, and add/remove/change the permissions there.

We'd also want to have a tab/setting on the Project Settings page that shows users who have special permissions on that project, and the ability to add/remove/change users there.

jonmoter:
Closing this PR in favor of #560.

PR #407, opened 2015-05-12
fix sporadic runtime errors seen in production due to MRI bug

jonmoter:
see explanation at https://github.com/puma/puma/issues/647
example error: https://zendesk.airbrake.io/projects/95346/groups/1338411361327591219
```
NoMethodError: undefined method `url_options' for #<Module:0x007f029d1080c8>`
```

/cc @zendesk/runway

### Risks
 - None


grosser:
:+1: sad that we have to do something like that :(

PR #408, opened 2015-05-12
fix exception thrown when X-Hub-Signature header missing

jonmoter:
example error: https://zendesk.airbrake.io/projects/95346/groups/1415261459885533031

/cc @zendesk/runway

### Risks
 - None

### DIFF
--- a/app/controllers/integrations/github_controller.rb
+++ b/app/controllers/integrations/github_controller.rb
@@ -16,7 +16,7 @@ class Integrations::GithubController < Integrations::BaseController
       request.body.tap(&:rewind).read
     )
 
-    Rack::Utils.secure_compare(request.headers['X-Hub-Signature'], "sha1=#{hmac}")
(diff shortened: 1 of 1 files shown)


app/controllers/integrations/github_controller.rb
```diff
@@ -10,6 +10,8 @@ def deploy?
   end
 
   def valid_signature?
+    return false unless request.headers['X-Hub-Signature'].present?
```
grosser: extract a method and reuse below ?

grosser: or just do a `.to_s` below

jonmoter: that's true, a `to_s` would be simpler.  good call.

jonmoter:
@grosser updated to use `to_s`

grosser:
:+1:

PR #412, opened 2015-05-18
fix stage svgs

sandlerr:
@zendesk/samson the intention of #392 was only to allow dots in project id, not any parameter within that block that happens to be called id.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,4 @@
 Samson::Application.routes.draw do
-  resources :projects, constraints: { id: /[^\/]+/ }
   resources :projects do
     resources :jobs, only: [:index, :new, :create, :show, :destroy]
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,6 @@
 Samson::Application.routes.draw do
-  resources :projects, constraints: { id: /[^\/]+/ } do
+  resources :projects, constraints: { id: /[^\/]+/ }
+  resources :projects do
     resources :jobs, only: [:index, :new, :create, :show, :destroy]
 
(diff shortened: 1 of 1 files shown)


grosser:
I think that's the wrong direction ... how about just not allowing dots ? / use .parameterize to generate the to_param

sandlerr:
Let's go with parameterize, I'll update this branch

sandlerr:
@grosser we already have parameterize [here](https://github.com/zendesk/samson/blob/master/app/models/concerns/permalinkable.rb#L27), it seems @salamagd manually changed the permalink in the database to have a dot.

[parameterize](http://apidock.com/rails/ActiveSupport/Inflector/parameterize) allows dashes and underscores, should we only support that?

grosser:
that should be supported by default, maybe the problematic permalink was generated before we used parameterize ... so let's just change it or make a change migration and stop :)

sandlerr:
You already did that [here](https://github.com/zendesk/samson/blob/master/db/migrate/20141021202831_fix_invalid_permalinks.rb) :-)

grosser:
idk ... just fix it in the db and close this ... we can investigate if it happens again :)

sandlerr:
This PR now reverts #392

salamagd:
Sorry! I must've not been on the latest upstream with parameterize included (I didn't change it in the DB).

sandlerr:
Does it still include the dot on the latest version? If so then we can revert that change safely. We should have caught this earlier :crying_cat_face:

grosser:
:+1: no idea how that happened ... let's investigate if it happens again :)

sandlerr:
We don't have routing tests for anything, if we did they would have caught this.

PR #414, opened 2015-05-20
rails assets source block

sandlerr:
cc @zendesk/samson


grosser:
:+1: 

did you try deploying that ?

sandlerr:
Good point, I'll make sure our bundler is clever enough and then merge.

sandlerr:
hit this bug: https://github.com/rails-assets/rails-assets/issues/225 as we are 1.8.3

sandlerr:
deploys are failing, not sure I understand why.

sandlerr:
deploys are failing for an unrelated reason, merging

PR #416, opened 2015-05-20
Create models to represent Builds

jonmoter:
The Build models will be used to associate Docker builds with a given Github SHA.
The BuildStatus class will be used when running tests or analysis on that build, such as Code Climate, Travis, Jenkins, and possibly Canary runs.

/cc @zendesk/runway 

### Risks
 - new code, minimal risk

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -2,10 +2,15 @@ class Build < ActiveRecord::Base
   SHA1_REGEX = /\A[0-9a-f]{40}/i
 
   belongs_to :project
+  has_many :statuses, class: 'BuildStatus'
   has_many :deploys
   has_many :releases
 
   validates :project, presence: true
   validates :git_sha, format: SHA1_REGEX, allow_nil: true
   validates :container_sha, format: SHA1_REGEX, allow_nil: true
+
+  def successful?
+    statuses.successful.all?
+  end
 end
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -1,8 +1,8 @@
 class Build < ActiveRecord::Base
-  SHA1_REGEX = /\A[0-9a-f]{40}/i
+  SHA1_REGEX = /\A[0-9a-f]{40}\Z/i
 
   belongs_to :project
-  has_many :statuses, class: 'BuildStatus'
+  has_many :statuses, class_name: 'BuildStatus'
   has_many :deploys
   has_many :releases
 
@@ -11,6 +11,6 @@ class Build < ActiveRecord::Base
   validates :container_sha, format: SHA1_REGEX, allow_nil: true
 
   def successful?
-    statuses.successful.all?
+    statuses.all?(&:successful?)
   end
 end
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -1,0 +1,11 @@
+class Build < ActiveRecord::Base
+  SHA1_REGEX = /\A[0-9a-f]{40}/i
+
+  belongs_to :project
+  has_many :deploys
+  has_many :releases
+
+  validates :project, presence: true
+  validates :git_sha, format: SHA1_REGEX, allow_nil: true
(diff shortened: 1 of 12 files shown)


app/models/build.rb
```diff
@@ -0,0 +1,16 @@
+class Build < ActiveRecord::Base
+  SHA1_REGEX = /\A[0-9a-f]{40}/i
```
grosser: `\z` ?

app/models/build.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  belongs_to :project
+  has_many :statuses, class: 'BuildStatus'
```
grosser: afaik this should be class_name ... class might do something unexpected

app/models/build.rb
```diff
@@ -0,0 +1,16 @@
...
+  has_many :deploys
+  has_many :releases
+
+  validates :project, presence: true
+  validates :git_sha, format: SHA1_REGEX, allow_nil: true
+  validates :container_sha, format: SHA1_REGEX, allow_nil: true
+
+  def successful?
+    statuses.successful.all?
```
grosser: this should be `statuses.all?(&:successful?)` ? <-> add tests ?

app/models/build.rb
```diff
@@ -0,0 +1,16 @@
+class Build < ActiveRecord::Base
```
grosser: maybe be more specific like ContainerBuild

db/migrate/20150508140514_create_build_artifact.rb
```diff
@@ -0,0 +1,12 @@
+class CreateBuildArtifact < ActiveRecord::Migration
+  def change
+    create_table :builds do |t|
+      t.belongs_to :project,    null: false, index: true
+      t.string :git_sha,        index: true
+      t.string :git_ref
+      t.string :container_sha,  index: true
+      t.string :container_ref
+      t.timestamps
```
grosser: any uniqueness we should enforce ?

test/models/build_status_test.rb
```diff
@@ -0,0 +1,48 @@
...
+
+describe BuildStatus do
+  let(:build) { builds(:docker_build) }
+
+  def valid_build_status options = {}
+    BuildStatus.new(options.reverse_merge(build: build, status: 'pending'))
+  end
+
+  describe 'validations' do
+    it 'should validate presence of build' do
```
grosser: if you leave should out it's shorter :)
`it 'validates presence of build'`

grosser:
'Build' seems a bit ambiguous ... otherwise :+1:

app/models/build.rb
jonmoter: Whoops, good call.  Thanks.

app/models/build.rb
jonmoter: There are class methods on `BuildStatus` for each of the statuses, so either would work.

Good idea to add a test, will do.

gabetax:
If a "Build" is the primary association on for updating a "Status", I'd rather it not be exclusive to Containers. I'd like to be able to have a Status API for traditional `cap` deploys.

db/migrate/20150508140514_create_build_artifact.rb
jonmoter: Yeah, `[project, git_sha]` should probably be unique. Technically, `git_sha` could probably be unique, if we assume we'll never have 2 projects that point to the same repo.

app/models/build.rb
grosser: `statuses.successful` returns either an empty or filled array, so all will always be true

like asking `[<record1>].all?` -> always true

jonmoter: Hm.  <thinking>  Yeah, I guess that's right.  Will fix and add a test.

jonmoter:
Code updated to reflect suggestions and comments.

grosser:
:+1:

Issue #417, opened 2015-05-22
Install plugin as a Gem


kureikain:
Is there a way we can distribute plugin as Gem and include them in Gemfile? 

cc @grosser @steved

grosser:
Just make it  a gem and include it in the Gemfile or did you try that and it did not work ?

kureikain:
My bad, I tried it but forgot to set `files` in Gemspec and it didn't include file so Gem.find_files in lib/samson/hooks.rb fail to find it for including. Thanks.

grosser:
Sweet :D
What did you write ?
Can we link it in the readme ?

On Fri, May 22, 2015 at 11:50 AM, Vinh Quốc Nguyễn <notifications@github.com

kureikain:
Yes,

Here is what I do:


```
Gem::Specification.new "samson_hipchat", "0.0.0" do |s|
  s.summary = "Samson hipchat integration"
  s.authors = ["Vinh Nguyen"]
  s.add_runtime_dependency "hipchat", "~> 1.5"
  s.files = `git ls-files`.split("\n")
end
```

s.files is important because we are distributing file via Gem.

Then in Gemfile:

```
gem 'samson_hipchat', require: false
```

And I have to include this patch: https://github.com/zendesk/samson/pull/418

That's all I need to make my hipchat plugin works (slack clone btw).

I think Samson should go in this direction so we can distributed plugin as gem.

kureikain:
cc @grosser

grosser:
Glad to see someone use this, what did you do for testing ?

Can you make a PR to add the plugin to the readme ?

On Fri, May 22, 2015 at 12:52 PM, Vinh Quốc Nguyễn <notifications@github.com

kureikain:
> what did you do for testing ?

For testing, I'm kind of stupid.... I don't know how to test it outside of Samson environment(I'm a junor Ruby dev), so in my current dev I use sub module to pull it into `plugins` folder normally and run separate `rake test TEST=plugins/hipchat/tests/testfile.rb`

> Can you make a PR to add the plugin to the readme ?

Sure.
Will do.

grosser:
That sounds like a kind of ugly but reliable solution :)

PR #418, opened 2015-05-22
Fix to load plugin from Gem

kureikain:
For issue #417 .

The way we currently calculate plugin name is based on directory name.

This leads to two issues:

1. The path include version
    
    Such as ```/Users/kureikain/.rvm/gems/ruby-2.2.2/gems/samson_hipchat-0.0.2```
    So the line `@engine ||= Kernel.const_get("::Samson#{@name.camelize}::Engine")` will have @name like `samson_hipchat-0.0.2`.

2. The classname is correct because the germane may different from classname. 

    Example, we hava gem samson_hipchat in
    Such as /Users/kureikain/.rvm/gems/ruby-2.2.2/gems/samson_hipchat-0.0.2

    The real class is `SamsonHipchat` but it became `SamsonSamsonHipchat`. 

So, I think we should have a convention for Samson plugin Gem starting with `samson_`. This PR implements that and fixs version issue.

### DIFF
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -18,7 +18,7 @@ module Samson
       def initialize(path)
         @path = path
         @folder = File.expand_path('../../../', @path)
-        @name = File.basename(@folder)
+        @name = realname(File.basename(@folder))
       end
 
       def active?
@@ -52,6 +52,10 @@ module Samson
       end
 
       private
+
+      def realname(gemname)
+        gemname.sub(/-[^-]*\z/, '').sub(/\Asamson_/, "")
+      end
 
       def decorators_root
         @decorators_root ||= engine.config.root.join("app/decorators")
--- a/test/lib/samson/hooks_test.rb
+++ b/test/lib/samson/hooks_test.rb
@@ -46,6 +46,26 @@ describe Samson::Hooks do
         end
       end
     end
+
+    context 'when load via local plugin' do
+      let(:env) { 'other' }
+      let(:plugins) { 'all' }
+      let (:path) { '/path/to/samson/plugins/zendesk/lib/samson_zendesk/samson_plugin.rb' }
+
+      it 'return correct plugin name from folder' do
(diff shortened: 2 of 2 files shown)


kureikain:
cc @grosser

grosser:
that looks good, can you add a few test-cases around that ?

kureikain:
Sure, let me add some tests

kureikain:
I add 2 tests and rebase the PR. It tests 2 cases:

1. When load from local plugin, ensure we get correct plugin class name so it won't break current plugin architecture. 

2. When load from gem, ensure we get correct class name (remove version, replace `samson_` with empty string)

SQLite and MySQL passed the test (https://travis-ci.org/zendesk/samson/builds/63705267) but PostGreSQL failed, do you have any idea? I don't think my code has any relation with database?

lib/samson/hooks.rb
```diff
@@ -53,6 +56,14 @@ def engine
 
       private
 
+      def realname(gemname)
+        gem_name = gemname.split("-")
+        gem_name.pop if gem_name.length >=2
+        name = gem_name.join("-")
+        name = name.gsub(SAMSON_PLUGIN_PREFIX, "") if name.start_with? SAMSON_PLUGIN_PREFIX
```
grosser: you can simplify this line to `name.sub!(/\Asamson_/)`

lib/samson/hooks.rb
```diff
@@ -53,6 +56,14 @@ def engine
 
       private
 
+      def realname(gemname)
+        gem_name = gemname.split("-")
```
grosser: `gemname.sub!(/-[^-]*\z/, '')`

grosser:
logic / tests looks good, do you want to switch to regex or leave it ?

kureikain:
Switch to your shorter regex. much simpelr :). wait the new build.

grosser:
thx! looks good :)

PR #421, opened 2015-05-26
Plugin to allow writing stages with Ruby script instead of Bash

henders:
Added a dropdown for editing stage commands to denote whether it should be executed as a ruby or shell script.

This should lay some groundwork when we want to slot in a Docker executer for meatballs integration.

screen shot 2015-05-26 at 16 02 51

screen shot 2015-05-26 at 16 02 04

### References
 - Jira link:

### Risks
 - None


grosser:
Can we check in the code + have tests and then just call out to it ?

define foobarhelper executable and then use `foobarhelper args` as a build step

db/migrate/20150522121048_add_command_type_to_stages.rb
```diff
@@ -0,0 +1,7 @@
+class AddCommandTypeToStages < ActiveRecord::Migration
+  def change
+    change_table :stages do |t|
+      t.string :command_type
```
jonmoter: How about set a `null: false, default: 'shell_script'` or whatever the correct value would be?

jonmoter:
Question: why is it that when Samson currently runs commands, they're executed in a bash shell, rather than csh or tcsh or zsh or whatever?

In `TermnicalExecutor`, I see that it explicitly calls `/bin/sh -c #{command}` if you're running jruby, but there's no equivalent for MRI.  As far as I can tell, it just runs `PTY.spawn(commands)`, and I don't know how that determines what shell to use.

grosser:
that might be because jruby is just super slow to start anything so we optimized by concatting

grosser:
shell preference is just something that worked and is available ...

db/migrate/20150522121048_add_command_type_to_stages.rb
henders: I was a bit leery of doing that as at the time I was changing the class name a lot with indecisiveness, so not giving a default allows the default to be in code and thus refactorings become easier in the future and you then only have to worry about the non-default rows.

henders:
@grosser 
> Can we check in the code + have tests and then just call out to it ?
> define foobarhelper executable and then use foobarhelper args as a build step
Not sure what you mean. Do you mean have the user just write in the stage commands, and not bother with a plugin at all?
```
echo 'puts "this is where my ruby code would go"' > myscript.rb
ruby myscript.rb
```

Or were you saying to be able to swap out the JobExecution class completely with an arbitrary FooBarExecution, similar to how I implemented this but at a higher level?

grosser:
Yes, either that if it's not super complicated or have a class of 'build
commands' that are available via the ui just like commands  (all classes
inheriting from xyz or in folder xyz)

I just don't want to see / maintain a bunch of
copy-pasted/untested/unversioned ruby code :)

jonmoter:
Is that worse than a bunch of copy-pasted/untested/unversioned bash code?

grosser:
I'd say yes, since it invites people to do overly complex things ... all the bash foo we have is mostly very simple and hard to test

henders:
@grosser 
Sorry, but which of the options were you referring to when you said 'yes'?

Also at the end of the day, since Samson is still just a generic scheduler of scripts, the idea of this plugin was to allow them to write these deploy scripts in Ruby if they so chose if their bash-fu isn't up to snuff.

At the moment Voice(https://samson.zende.sk/projects/voice/stages/voice-staging) have some complicated bash scripts just because bash has ugly and limited conditional syntax.  And scripts like that don't really belong in the Git Repo as it's not related to the project itself.

grosser:
Hmmm yeah I guess for this kind of stuff it makes sense ... I was thinking
more in the direction of building and releasing images ... but I guess even
then it's nice to have a playground before turning it into a script that
needs to be deployed ... so :+1: even though I think it will end up in a
bunch of messy code :D

gabetax:
I still find myself skeptical of the value of this feature against the cost of carry. Maybe it's because I just read the [Martin Fowler YAGNI post](http://martinfowler.com/bliki/Yagni.html), or because I stared too long at the "fewer features" mantra on the Github carpet. If logic is too complicated to concisely express in a bash one-liner, I'd rather it live in the project repo (if it's project-specific) or a library like zendesk_deployment (if it's more generic), and then it can be in any language.

Are there any other existing scenarios where ruby execution is a win?

grosser:
Yeah that' kind of true ... complicated stuff should move to the project or
samson ...

henders:
@gabetax , the cost of carry doesn't seem to apply here as this was a plugin that I created to help pave the way for the docker functionality. So it's not a pure presumptive feature as it makes the definite docker functionality plugin easier (I hope), and it solves a current problem with the script in the Voice deploys.

And nope, I had no other existing scenarios besides personal preference for writing scripts in ruby. :)

I'll hold off on merging this anyway for more feedback, as @grosser made a good point about possibly up'ing the abstraction to the JobExecution level instead of the TerminalExecutor level where I was plugging in.

Also overall, I'm still treating Samson like Jenkins with it's rich plugin backing, and I've used the jenkins-ruby plugin plenty of times. So maybe this is my problem and not a Samson community's.

grosser:
maybe another simple abstraction is to make the docker foobar an executable
inside of samson and then just call it via bash, this way all the logging
and exit status is handled and it's easy to combine with other things

henders:
Not going to pursue this for now.

PR #422, opened 2015-05-27
Update SSE library for disconnect bug.

henders:
Update SSE library to pick up a fixed bug with client disconnects.

### References
 - Jira link:

### Risks
 - None


grosser:
:+1:

PR #423, opened 2015-05-29
consistent port

sandlerr:
We've always used port 9080, its mentioned in puma.rb also @zendesk/runway


grosser:
I'm kind of used to rails containers booting on 3000 so maybe change the internal port / docs to add -p 3000 and run with -p 3000 here ?

:+1: this is fine too since it's better than broken ;)

sandlerr:
merging for now, will put another PR for 3000

PR #425, opened 2015-05-29
A simplistic approach to pipelines.

henders:
Add functionality to allow user to specify other stages to automatically deploy when a deploy finishes.
These deploys are kicked off in parallel.

This is the stage-edit Screenshot:
screen shot 2015-06-02 at 12 58 31

And we append the deploy links to the Deploy output:
screen shot 2015-06-02 at 13 04 57

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -120,7 +120,8 @@ class JobExecution
     ActiveSupport::Notifications.instrument("execute_shell.samson", payload) do
       payload[:success] = @executor.execute!(*cmds)
     end
-    Samson::Hooks.fire(:before_execute_finish_msg, @job, @output)
+    Samson::Hooks.fire(:after_job_execution, @job, payload[:success], @output)
+    payload[:success]
   end
 
   def setup!(dir)
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -120,6 +120,7 @@ class JobExecution
     ActiveSupport::Notifications.instrument("execute_shell.samson", payload) do
       payload[:success] = @executor.execute!(*cmds)
     end
+    Samson::Hooks.fire(:before_execute_finish_msg, @job, @output)
   end
 
   def setup!(dir)
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -273,6 +273,7 @@ ActiveRecord::Schema.define(version: 20150619170905) do
     t.string   "static_emails_on_automated_deploy_failure",    limit: 255
     t.string   "datadog_monitor_ids",                          limit: 255
     t.string   "jenkins_job_names",                            limit: 255
+    t.string   "next_stage_ids"
   end
 
   add_index "stages", ["project_id", "permalink", "deleted_at"], name: "index_stages_on_project_id_and_permalink_and_deleted_at", length: {"project_id"=>nil, "permalink"=>191, "deleted_at"=>nil}, using: :btree
--- a/plugins/pipelines/app/decorators/stage_decorator.rb
+++ b/plugins/pipelines/app/decorators/stage_decorator.rb
@@ -1,21 +1,6 @@
-module Stage::PipelineProductionEnhancer
-  # Return true if any stages in the pipeline are marked production
-  def production?(check_next: true)
-    super || (
-      check_next &&
-        next_stage_ids.any? &&
-        Stage.find(job.deploy.stage.next_stage_ids).any? { |s| s.production?(check_next: false) }
-    )
-  end
-end
-
 Stage.class_eval do
-  prepend Stage::PipelineProductionEnhancer
+  prepend SamsonPipelines::StageConcern
   serialize :next_stage_ids, Array
 
-  def next_stage
-    return Stage.find(next_stage_ids.first) unless next_stage_ids.empty?
-    stages = project.stages.to_a
-    stages[stages.index(self) + 1]
-  end
+  validate :valid_pipeline?
 end
--- a/plugins/pipelines/app/decorators/stage_decorator.rb
+++ b/plugins/pipelines/app/decorators/stage_decorator.rb
@@ -1,0 +1,17 @@
+Stage.class_eval do
+  serialize :next_stage_ids, Array
+
+  puts "**** Stage Class Decorator is being eval'd"
+  def next_stage
+    return Stage.find(next_stage_ids.first) unless next_stage_ids.empty?
+    stages = project.stages.to_a
+    stages[stages.index(self) + 1]
+  end
+
+  # alias_method :old_production?, :production? if defined?(:production?)
+  # Return true if any stages in the pipeline are marked production
+  def production?
+    return old_production? if next_stage_ids.empty?
+    old_production? || Stage.find(job.deploy.stage.next_stage_ids).any?(:old_production?)
+  end
+end
--- a/plugins/pipelines/app/decorators/stage_decorator.rb
+++ b/plugins/pipelines/app/decorators/stage_decorator.rb
@@ -8,7 +8,7 @@ Stage.class_eval do
     stages[stages.index(self) + 1]
   end
 
-  # alias_method :old_production?, :production? if defined?(:production?)
+  alias_method :old_production?, :production?
   # Return true if any stages in the pipeline are marked production
   def production?
     return old_production? if next_stage_ids.empty?
--- a/plugins/pipelines/app/decorators/stage_decorator.rb
+++ b/plugins/pipelines/app/decorators/stage_decorator.rb
@@ -1,17 +1,21 @@
+module Stage::PipelineProductionEnhancer
+  # Return true if any stages in the pipeline are marked production
+  def production?(check_next: true)
+    super || (
+      check_next &&
+        next_stage_ids.any? &&
+        Stage.find(job.deploy.stage.next_stage_ids).any? { |s| s.production?(check_next: false) }
+    )
+  end
+end
+
 Stage.class_eval do
+  prepend Stage::PipelineProductionEnhancer
   serialize :next_stage_ids, Array
 
-  puts "**** Stage Class Decorator is being eval'd"
   def next_stage
     return Stage.find(next_stage_ids.first) unless next_stage_ids.empty?
     stages = project.stages.to_a
     stages[stages.index(self) + 1]
-  end
-
-  alias_method :old_production?, :production?
-  # Return true if any stages in the pipeline are marked production
-  def production?
-    return old_production? if next_stage_ids.empty?
-    old_production? || Stage.find(job.deploy.stage.next_stage_ids).any?(:old_production?)
   end
 end
--- a/plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
+++ b/plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
@@ -1,0 +1,36 @@
+module SamsonPipelines::StageConcern
+  # Return true if any stages in the pipeline are marked production
+  def production?
+    super || next_stages.any?(&:production?)
+  end
+
+  def next_stage
+    next_stage_ids.empty? ? super : Stage.find(next_stage_ids.first)
+  end
+
+  def next_stages
+    Stage.find(next_stage_ids)
+  end
+
+  protected
+
+  # Ensure we don't have a circular pipeline:
+  #
+  # potential race-condition if 2 stages are saved at same time:
+  #   stageA saved with pipelines to stageB and stageC
+  #   stageC saved with pipeline to stageA   => will validate if stageA above hasn't been written to DB yet
+  def valid_pipeline?(origin_id = id)
+    if next_stage_ids.any? { |next_id| next_id.to_i == origin_id.to_i }
+      errors[:base] << "Stage #{name} causes a circular pipeline with this stage"
+      return false
+    end
+
+    next_stages.each do |stage|
+      unless stage.valid_pipeline?(origin_id)
+        errors[:base] << "Stage #{stage.name} causes a circular pipeline with this stage"
+        return false
+      end
+    end
+    true
+  end
+end
--- a/plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
+++ b/plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
@@ -1,0 +1,23 @@
+<fieldset>
+  <legend>Pipelining</legend>
+  <p class="col-lg-offset-2">Set the stage that will get automatically deployed after this one.</p>
+  <div class="col-lg-4 col-lg-offset-2">
+    <table>
+      <% (@project.stages - [@stage]).each do |next_stage| %>
+        <tr>
+          <td>
+            <%= label_tag do %>
+              <%= check_box_tag(
+                    'stage[next_stage_ids][]',
+                    next_stage.id,
+                    (@stage && @stage.next_stage_ids.include?(next_stage.id.to_s))
+                  )
+              %>
+              <%= next_stage.name %>
+            <% end %>
+          </td>
+        </tr>
+      <% end %>
+    </table>
+  </div>
+</fieldset>
--- a/plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
+++ b/plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
@@ -1,6 +1,6 @@
 <fieldset>
   <legend>Pipelining</legend>
-  <p class="col-lg-offset-2">Set the stage that will get automatically deployed after this one.</p>
+  <p class="col-lg-offset-2">Set the stages that will get automatically deployed after this one.</p>
   <div class="col-lg-4 col-lg-offset-2">
     <table>
       <% (@project.stages - [@stage]).each do |next_stage| %>
--- a/plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
+++ b/plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
@@ -1,0 +1,27 @@
+module SamsonPipelines
+  class Engine < Rails::Engine
+  end
+
+  class << self
+    def start_pipelined_stages(job, output)
+      return if !job.deploy || job.deploy.stage.next_stage_ids.empty?
+
+      Stage.find(job.deploy.stage.next_stage_ids).each do |next_stage|
+        new_deploy = DeployService.new(job.deploy.user).deploy!(next_stage, reference: job.deploy.commit)
+        output.write("\n# Kicking off next stage: #{next_stage.name} - URL: #{new_deploy.url}\n")
+      end
+    rescue => ex
+      raise "Failed to start the next deploys in the pipeline: #{ex.message} - #{ex.backtrace}"
+    end
+  end
+end
+
+Samson::Hooks.view :stage_form, 'samson_pipelines/fields'
+
+Samson::Hooks.callback :stage_permitted_params do
+  { next_stage_ids: [] }
+end
+
+Samson::Hooks.callback :before_execute_finish_msg do |job, output|
+  SamsonPipelines.start_pipelined_stages(job, output)
+end
--- a/plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
+++ b/plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
@@ -3,15 +3,30 @@ module SamsonPipelines
   end
 
   class << self
-    def start_pipelined_stages(job, output)
-      return if !job.deploy || job.deploy.stage.next_stage_ids.empty?
+    def start_pipelined_stages(job, success, output)
+      return if !success || !job.deploy || job.deploy.stage.next_stage_ids.empty?
 
-      Stage.find(job.deploy.stage.next_stage_ids).each do |next_stage|
-        new_deploy = DeployService.new(job.deploy.user).deploy!(next_stage, reference: job.deploy.commit)
-        output.write("\n# Kicking off next stage: #{next_stage.name} - URL: #{new_deploy.url}\n")
+      deploy_service = DeployService.new(job.deploy.user)
+      job.deploy.stage.next_stages.each do |next_stage|
+        start_next_stage(next_stage, job, deploy_service, output)
       end
     rescue => ex
-      raise "Failed to start the next deploys in the pipeline: #{ex.message} - #{ex.backtrace}"
+      output.write("Failed to start the pipelined stages: #{ex.message}")
+    end
+
+    private
+
+    def start_next_stage(next_stage, current_job, deploy_service, output)
+      deploy = deploy_service.deploy!(next_stage, reference: current_job.deploy.commit)
+      if !deploy.persisted?
+        output.puts "# Pipeline: Failed to start the next stage '#{next_stage.name}': #{deploy.errors.full_messages}\n"
+      elsif next_stage.deploy_requires_approval?
+        deploy.update!(buddy: current_job.deploy.buddy)
+        deploy_service.confirm_deploy!(deploy)
+      end
+      output.puts "# Pipeline: Kicked off next stage: #{next_stage.name} - URL: #{deploy.url}\n"
+    rescue => ex
+      output.puts "# Pipeline: Failed to start the next stage '#{next_stage.name}': #{ex.message}\n"
     end
   end
 end
@@ -22,6 +37,6 @@ Samson::Hooks.callback :stage_permitted_params do
   { next_stage_ids: [] }
 end
 
-Samson::Hooks.callback :before_execute_finish_msg do |job, output|
-  SamsonPipelines.start_pipelined_stages(job, output)
+Samson::Hooks.callback :after_job_execution do |job, success, output|
+  SamsonPipelines.start_pipelined_stages(job, success, output)
 end
--- a/plugins/pipelines/test/lib/hooks_test.rb
+++ b/plugins/pipelines/test/lib/hooks_test.rb
@@ -1,0 +1,21 @@
+require_relative '../test_helper'
+
+describe 'zendesk hooks' do
+  let(:deploy) { deploys(:succeeded_test) }
+  let(:stage) { deploy.stage }
+  let(:next_stages) { [ stages(:test_production), stages(:test_production_pod) ] }
+  let(:fake_deploy) { Hashie::Mash.new(url: 'abc') }
+
+  describe :after_deploy do
+    it 'kicks off the next stages in the deploy' do
+      stage.update!(next_stage_ids: next_stages.map(&:id))
+      DeployService.any_instance.expects(:deploy!).returns(fake_deploy).twice
+      Samson::Hooks.fire(:after_deploy, deploy, nil)
+    end
+
+    it 'does not deploy another if the next_stage_id is nil' do
+      DeployService.any_instance.expects(:new).never
+      Samson::Hooks.fire(:after_deploy, deploy, nil)
+    end
+  end
+end
--- a/plugins/pipelines/test/lib/hooks_test.rb
+++ b/plugins/pipelines/test/lib/hooks_test.rb
@@ -2,20 +2,41 @@ require_relative '../test_helper'
 
 describe 'zendesk hooks' do
   let(:deploy) { deploys(:succeeded_test) }
+  let(:next_deploy) { deploys(:succeeded_production_test) }
   let(:stage) { deploy.stage }
   let(:next_stages) { [ stages(:test_production), stages(:test_production_pod) ] }
-  let(:fake_deploy) { Hashie::Mash.new(url: 'abc') }
+  let(:output) { StringIO.new }
+  let(:job) do
+    Job.create(
+      project: stage.project,
+      command: "echo hello world",
+      status: "running",
+      user: User.first,
+      deploy: deploy)
+  end
 
-  describe :after_deploy do
+  describe :after_job_execution do
     it 'kicks off the next stages in the deploy' do
       stage.update!(next_stage_ids: next_stages.map(&:id))
-      DeployService.any_instance.expects(:deploy!).returns(fake_deploy).twice
-      Samson::Hooks.fire(:after_deploy, deploy, nil)
+      DeployService.any_instance.expects(:deploy!).returns(next_deploy).twice
+      Samson::Hooks.fire(:after_job_execution, job, true, output)
+    end
+
+    it 'does not kick off the next stage in the pipeline if current stage failed' do
+      stage.update!(next_stage_ids: next_stages.map(&:id))
+      DeployService.any_instance.expects(:deploy!).never
+      Samson::Hooks.fire(:after_job_execution, job, false, output)
     end
 
     it 'does not deploy another if the next_stage_id is nil' do
-      DeployService.any_instance.expects(:new).never
-      Samson::Hooks.fire(:after_deploy, deploy, nil)
+      DeployService.any_instance.expects(:deploy!).never
+      Samson::Hooks.fire(:after_job_execution, job, true, output)
+    end
+
+    it 'does not deploy another if the deploy is nil' do
+      job = Job.create(project: stage.project, command: "echo", status: "running", user: User.first)
+      DeployService.any_instance.expects(:deploy!).never
+      Samson::Hooks.fire(:after_job_execution, job, true, output)
     end
   end
 end
--- a/plugins/pipelines/test/models/stage_test.rb
+++ b/plugins/pipelines/test/models/stage_test.rb
@@ -1,0 +1,23 @@
+require_relative '../test_helper'
+
+describe Stage do
+  describe '#next_stage' do
+    let(:project) { Project.new(name: 'foo') }
+    let(:stage1) { Stage.new(project: project, name: 'stage1') }
+    let(:stage2) { Stage.new(project: project, name: 'stage2') }
+    let(:stage3) { Stage.new(project: project, name: 'stage3') }
+
+    before do
+      project.stages = [stage1, stage2, stage3]
+    end
+
+    it 'should return next created stage if no pipeline set' do
+      stage2.next_stage.id.must_equal stage3.id
+    end
+
+    it 'should return next stage in pipeline if set' do
+      stage2.next_stage_ids = [ stage1.id, stage3.id ]
+      stage2.next_stage.id.must_equal stage1.id
+    end
+  end
+end
--- a/plugins/pipelines/test/models/stage_test.rb
+++ b/plugins/pipelines/test/models/stage_test.rb
@@ -1,23 +1,87 @@
 require_relative '../test_helper'
 
 describe Stage do
+  let(:project) { Project.create!(name: 'foo', repository_url: 'random') }
+  let(:stage1) { Stage.create!(project: project, name: 'stage1') }
+  let(:stage2) { Stage.create!(project: project, name: 'stage2') }
+  let(:stage3) { Stage.create!(project: project, name: 'stage3') }
+  let(:production) { deploy_groups(:pod1) }
+  let(:staging) { deploy_groups(:pod100) }
+
+  before do
+    Project.any_instance.stubs(:valid_repository_url).returns(true)
+    project.stages = [stage1, stage2, stage3]
+  end
+
   describe '#next_stage' do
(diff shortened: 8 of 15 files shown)


plugins/pipelines/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,7 @@
+Stage.class_eval do
+  def next_stage
+    return Stage.find(next_stage_id) if next_stage_id
```
grosser: smells like a `belongs_to :next_stage, class_name: "Stage"`

grosser:
hmm we already have the next-stage concept (via ordering in the projects stages tab)
maybe all we need is "Auto deploy next stage" checkbox ?

db/schema.rb
```diff
@@ -240,6 +240,7 @@
     t.string   "static_emails_on_automated_deploy_failure",    limit: 255
     t.string   "datadog_monitor_ids",                          limit: 255
     t.string   "jenkins_job_names",                            limit: 255
+    t.integer  "next_stage_id"
```
gabetax: How do you feel about reversing the direction of this association from a `belongs_to` to a `has_many`? That could allow for a tree of host groups, e.g. after pod 1, deploy to pods 2, 3, and 5 in parallel.

grosser: maybe just serialize :next_stage_ids, Array ;)

I like the idea since it would keep stages atomic ...

gabetax:
While we do have the ordering that's used already, I think this is something we want directly in our domain instead of inferred.

henders:
>hmm we already have the next-stage concept (via ordering in the projects stages tab)
maybe all we need is "Auto deploy next stage" checkbox ?

Like @gabetax said, having an inferred ordering would be a pain to work with if you didn't create your stages in the right order to start with.

sandlerr:
Kicking off several parallel deploys would be really cool :D

henders:
@gabetax @grosser @sandlerr 
Made the pipeline support multiple stages to allow fan-out.
Also just serialized the next_stage_ids as @grosser suggested as it kept things simple.

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -0,0 +1,27 @@
...
+
+Samson::Hooks.view :stage_form, 'samson_pipelines/fields'
+
+Samson::Hooks.callback :stage_permitted_params do
+  { next_stage_ids: [] }
+end
+
+Samson::Hooks.callback :after_deploy do |deploy|
+  begin
+    return unless deploy.succeeded? || deploy.stage.next_stage_id.empty?
```
grosser: I think ifs are easier to parse with multiple conditionals
`return if !deploy.succeeded? && !deploy.stage.next_stage_ids.empty?`
... see already found a bug ... should be:
`return if !deploy.succeeded? || deploy.stage.next_stage_ids.empty?`

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    log = ''
+    Stage.find(deploy.stage.next_stage_ids).each do |next_stage|
```
grosser:
```Ruby
log = each_with_object("") do |next_stage, log|
  log << something
end  
```

grosser: `state.next_stages` ?

grosser:
this won't work with buddy deploys, right ?

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -0,0 +1,27 @@
...
+      new_deploy = DeployService.new(deploy.user).deploy!(next_stage, deploy.commit)
+      log += "Kicking off next stage: #{next_stage.name} - URL: #{new_deploy.url}\n"
+    end
+    deploy.job.reload.update_output!(deploy.job.output + "\n#{log}") if log.presence
```
grosser: it should be done before the log is closed, otherwise this won't be streamed ?

plugins/pipelines/test/models/stage_test.rb
```diff
@@ -0,0 +1,23 @@
...
+    let(:project) { Project.new(name: 'foo') }
+    let(:stage1) { Stage.new(project: project, name: 'stage1') }
+    let(:stage2) { Stage.new(project: project, name: 'stage2') }
+    let(:stage3) { Stage.new(project: project, name: 'stage3') }
+
+    before do
+      project.stages = [stage1, stage2, stage3]
+    end
+
+    it 'should return next created stage if no pipeline set' do
```
grosser: no need for `shoulds`

grosser: why do we need a next stage if we are not doing pipelining ?

plugins/pipelines/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,9 @@
+Stage.class_eval do
+  serialize :next_stage_ids, Array
+
+  def next_stage
```
grosser: `next_stages` ?

grosser:
maybe filter everything that needs a buddy check :(

plugins/pipelines/app/decorators/stage_decorator.rb
henders: this function is used for that button on deploy page, and only expects one stage. So didn't want to change it for this PR.

plugins/pipelines/test/models/stage_test.rb
henders: There is the button on the deploy page that showed a naive 'next stage' before this feature.

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
henders: I didn't want to update the job DB record N times, but instead update the job after all pipeline stages are kicked off.

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
henders: ugh, nice find... damn last minute name changes. I should have had a test for that case as well.

plugins/pipelines/app/decorators/stage_decorator.rb
grosser: that button is then useless, right ?

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
grosser: You would just append to the log, no db updating necessary

plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
```diff
@@ -0,0 +1,23 @@
+<fieldset>
+  <legend>Pipelining</legend>
+  <p class="col-lg-offset-2">Set the stage that will get automatically deployed after this one.</p>
```
sandlerr: This doesn't quite line up because its missing the [grid-gutter](https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L327), can you make this also a `col-lg-10` or whatever?

plugins/pipelines/test/lib/hooks_test.rb
```diff
@@ -0,0 +1,21 @@
...
+
+describe 'zendesk hooks' do
+  let(:deploy) { deploys(:succeeded_test) }
+  let(:stage) { deploy.stage }
+  let(:next_stages) { [ stages(:test_production), stages(:test_production_pod) ] }
+  let(:fake_deploy) { Hashie::Mash.new(url: 'abc') }
+
+  describe :after_deploy do
+    it 'kicks off the next stages in the deploy' do
+      stage.update!(next_stage_ids: next_stages.map(&:id))
```
sandlerr: if you put this in your fixture it avoids making a query in the test run so the specs run a tiny bit faster

plugins/pipelines/app/decorators/stage_decorator.rb
henders: Not useless, if you aren't using this pipeline stuff, then the existing functionality is what you'll get. Which IMO was a pretty useless button in the first place. So yea, I'm probably agreeing with you! :)

grosser:
rebased it and ruby -Itest plugins/pipelines/test/models/stage_test.rb works ...

henders:
Argh, really? Alright, cheers for looking at it! :)  I'll give it a go!

henders:
@grosser ,

Still appears broken, so I made the test a bit clearer. It can't override the 'production?' instance method because the 'Stage' class hasn't finished eval'ing yet.

```
$ ruby -Itest plugins/pipelines/test/models/stage_test.rb
**** Stage Class Decorator is being eval'd
/Users/shender/Code/zendesk/samson/plugins/pipelines/app/decorators/stage_decorator.rb:11:in `alias_method': undefined method `production?' for class `Stage' (NameError)
	from /Users/shender/Code/zendesk/samson/plugins/pipelines/app/decorators/stage_decorator.rb:11:in `block in <top (required)>'
	from /Users/shender/Code/zendesk/samson/plugins/pipelines/app/decorators/stage_decorator.rb:1:in `class_eval'
	from /Users/shender/Code/zendesk/samson/plugins/pipelines/app/decorators/stage_decorator.rb:1:in `<top (required)>'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:274:in `require'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:274:in `block in require'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:240:in `load_dependency'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:274:in `require'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:360:in `require_or_load'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:317:in `depend_on'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:233:in `require_dependency'
	from /Users/shender/Code/zendesk/samson/lib/samson/hooks.rb:130:in `block in load_decorators'
	from /Users/shender/Code/zendesk/samson/lib/samson/hooks.rb:130:in `each'
	from /Users/shender/Code/zendesk/samson/lib/samson/hooks.rb:130:in `load_decorators'
	from /Users/shender/Code/zendesk/samson/lib/samson/hooks.rb:182:in `inherited'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/kaminari-0.16.3/lib/kaminari/models/active_record_extension.rb:10:in `inherited'
	from /Users/shender/Code/zendesk/samson/app/models/stage.rb:1:in `<top (required)>'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:274:in `require'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:274:in `block in require'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:240:in `load_dependency'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:274:in `require'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:360:in `require_or_load'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:494:in `load_missing_constant'
	from /opt/boxen/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:184:in `const_missing'
	from plugins/pipelines/test/models/stage_test.rb:3:in `<main>'
```

grosser:
ah now it's broken for me too :)

henders:
Damn @grosser , genius workaround, and it's not even a hack as that's what it was designed for.  Thanks for the education again! :)

plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
henders: Ah, I was cutting/pasting from other plugins here. It seems lined up on my screen with the other plugin settings.
Pic or it didn't happen! :)
@sandlerr

henders:
Ready for final review @zendesk/runway @grosser @sandlerr @zendesk/samson  @pschambacher

app/models/job_execution.rb
```diff
@@ -120,6 +120,8 @@ def execute!(dir)
     ActiveSupport::Notifications.instrument("execute_shell.samson", payload) do
       payload[:success] = @executor.execute!(*cmds)
     end
+    Samson::Hooks.fire(:before_execute_finish_msg, @job, payload[:success], @output)
```
grosser: msg -> message ?

grosser: or after_job_execution

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
```diff
@@ -0,0 +1,33 @@
+module PipelinePluginConcern
```
grosser: namespace it ?

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
```diff
@@ -0,0 +1,33 @@
...
+  # Return true if any stages in the pipeline are marked production
+  def production?
+    super || Stage.find(next_stage_ids).any? { |s| s.production? }
```
grosser: we should avoid the find if there are no ids

grosser: .any?(&:production?)

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def next_stage
+    return Stage.find(next_stage_ids.first) unless next_stage_ids.empty?
+    stages = project.stages.to_a
+    stages[stages.index(self) + 1]
```
grosser: isn't that super ?

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  # Ensure we don't have a circular pipeline:
```
grosser: private ?

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
```diff
@@ -0,0 +1,33 @@
...
+  #
+  # potential race-condition if 2 stages are saved at same time:
+  #   stageA saved with pipelines to stageB and stageC
+  #   stageC saved with pipeline to stageA   => will validate if stageA above hasn't been written to DB yet
+  def valid_pipeline?(origin_id = id)
+    Rails.logger.warn "valid_pipeline? #{origin_id} / #{id} - #{next_stage_ids}"
+    if next_stage_ids.any? { |next_id| next_id.to_i == origin_id.to_i }
```
grosser: `.include?(origin_id)` ?

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
```diff
@@ -0,0 +1,33 @@
...
+      errors[:base] << "Stage #{name} causes a circular pipeline with this stage"
+      return false
+    end
+
+    Stage.find(next_stage_ids).each do |stage|
```
grosser: maybe a next_stages method that does no queries when ids are empty

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
henders: Yup, to the .any?(&:production?), but it also seems activerecord is smart enough to skip the find if passed an empty array.

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
grosser: + caches the results

app/models/job_execution.rb
henders: Yup, I'll rename it.

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
grosser: hmm protected could still work

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
henders: This was the old function body. But yea, I can 'super' this like above now that the method overriding is working.

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
grosser: I think you can add all the errors and it will stop by itself -> less code + all bad stages show up

plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
grosser: stage -> stages

grosser: 'After deploying this, auto-deploy these stages'

plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <div class="col-lg-4 col-lg-offset-2">
+    <table>
+      <% (@project.stages - [@stage]).each do |next_stage| %>
+        <tr>
+          <td>
+            <%= label_tag do %>
+              <%= check_box_tag(
+                    'stage[next_stage_ids][]',
+                    next_stage.id,
+                    (@stage && @stage.next_stage_ids.include?(next_stage.id.to_s))
```
grosser: how can there be no @stage ?

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -0,0 +1,40 @@
+module SamsonPipelines
+  class Engine < Rails::Engine
+  end
+
+  class << self
+    def start_pipelined_stages(job, success, output)
+      return if !success || !job.deploy || job.deploy.stage.next_stage_ids.empty?
+
+      deploy_service = DeployService.new(job.deploy.user)
+      Stage.find(job.deploy.stage.next_stage_ids).each do |next_stage|
```
grosser: `.next_stages`

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -0,0 +1,40 @@
...
+        start_next_stage(next_stage, job, deploy_service, output)
+      end
+    rescue => ex
+      output.write("Failed to start the pipelined stages: #{ex.message}")
+    end
+
+    def start_next_stage(next_stage, current_job, deploy_service, output)
```
grosser: private ?

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -0,0 +1,40 @@
...
+      new_deploy = deploy_service.deploy!(next_stage, reference: current_job.deploy.commit)
```
grosser: new_deploy -> deploy ?

grosser: could we pass `buddy: current_job.deploy.buddy` and save the elsif below ?

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -0,0 +1,40 @@
...
+      if !new_deploy.persisted?
+        output.write("\n# Failed to start the next stage '#{next_stage.name}' in the pipeline: #{new_deploy.errors.full_messages}\n")
```
grosser: dealing with newlines can be avoided by using puts

plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
henders: When this is a new project and the stage fields are appended to project settings page.

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -0,0 +1,40 @@
...
+      elsif next_stage.deploy_requires_approval?
+        new_deploy.update!(buddy: current_job.deploy.buddy)
+        deploy_service.confirm_deploy!(new_deploy)
+      end
+      output.write("\n# Kicked off next stage: #{next_stage.name} - URL: #{new_deploy.url}\n")
```
grosser: prefixing all these with `Pipeline:` might make finding them easier / knowing where they came from

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
henders: That doesn't work for this field type - serialized.

plugins/pipelines/test/lib/hooks_test.rb
```diff
@@ -0,0 +1,42 @@
...
+  let(:job) do
+    Job.create(project: stage.project,
+               command: "echo hello world",
+               status: "running",
+               user: User.first,
+               deploy: deploy)
+  end
+
+
+  describe :before_execute_finish_msg do
```
grosser: we normally use `.xxx` for class and `#xxx` for instance methods

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
henders: sure.

plugins/pipelines/test/lib/hooks_test.rb
```diff
@@ -0,0 +1,42 @@
+require_relative '../test_helper'
+
+describe 'zendesk hooks' do
+  let(:deploy) { deploys(:succeeded_test) }
+  let(:next_deploy) { deploys(:succeeded_production_test) }
+  let(:stage) { deploy.stage }
+  let(:next_stages) { [ stages(:test_production), stages(:test_production_pod) ] }
+  let(:output) { StringIO.new }
+  let(:job) do
+    Job.create(project: stage.project,
```
grosser: this indentation styles tends to break down the longer the first line gets, 2-space does not :)

plugins/pipelines/test/lib/hooks_test.rb
grosser: ah never mind ... this is not  method but the hook :D

plugins/pipelines/test/lib/hooks_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    it 'does not deploy another if the next_stage_id is nil' do
+      DeployService.any_instance.expects(:new).never
+      Samson::Hooks.fire(:before_execute_finish_msg, job, true, output)
+    end
+
+    it 'does not deploy another if the deploy is nil' do
+      job = Job.create(project: stage.project, command: "echo", status: "running", user: User.first)
+      DeployService.any_instance.expects(:new).never
```
grosser: would be nice if these expectations were symmetric like `DeployService.any_instance.expects(:deploy!).never`

plugins/pipelines/test/lib/hooks_test.rb
henders: Changed! :)

plugins/pipelines/test/lib/hooks_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    Job.create(project: stage.project,
+               command: "echo hello world",
+               status: "running",
+               user: User.first,
+               deploy: deploy)
+  end
+
+
+  describe :before_execute_finish_msg do
+    it 'kicks off the next stages in the deploy' do
```
grosser: these 2 share the setup:
```Ruby
describe 'with next stages configured' do
  before { stage.update!(next_stage_ids: next_stages.map(&:id)) }
```

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
henders: Hmm... fair point.

plugins/pipelines/test/lib/hooks_test.rb
grosser: and the last one too ...

grosser: maybe best to have the default setup be the good case and then each test change 1 thing and then expect that it does not work

plugins/pipelines/test/models/stage_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  describe '#production?' do
+    it 'returns false if no pipeline set and not marked production' do
+      stage1.deploy_groups = [ deploy_groups(:pod100) ]
+      stage1.production?.must_equal false
+    end
+
+    it 'returns true if no pipeline set and marked production' do
+      stage1.deploy_groups = [ deploy_groups(:pod1) ]
```
grosser: let(:production) + let(:staging) might make this a bit more readable, not everyone might know what pod1 / pod100 means :)

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
henders: @grosser 
> could we pass buddy: current_job.deploy.buddy and save the elsif below ?

That would require changing the deploy_service code, so leaving it like this for the moment. Also the elsif would have to stay anyway unless we changed deploy_service further to auto-confirm deploy if buddy is set.
I'll leave that for another PR.

test/lib/samson/hooks_test.rb
```diff
@@ -1,7 +1,7 @@
 require_relative '../../test_helper'
 
 describe Samson::Hooks do
-  let(:number_of_plugins) { 4 }
+  let(:number_of_plugins) { 5 }
```
grosser: `Dir['plugins/*'].size` or something like that :D

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
grosser: `def next_stages` ;)

plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
grosser: can we just create an @stage in that case ... saw this weird check before and seems like a common stumbling block

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
henders: done

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
henders: Yup, protected worked!

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
henders: There was a mismatch when saving the IDs and reading them from the DB, they'd get converted to strings. So this just made sure production and testing code worked with minimum of fuss.

plugins/pipelines/app/models/concerns/pipeline_plugin_concern.rb
grosser: oh yes, must be some new 4.x thing :D

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
henders: done

plugins/pipelines/test/models/stage_test.rb
henders: Yea, 'pods' are now ingrained in my mind, so don't stop to think about others! :/

test/lib/samson/hooks_test.rb
henders: Yup, good point! :)

plugins/pipelines/test/lib/hooks_test.rb
henders: hah, alright done!

plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
henders: Yea, but I'll leave it out of this PR since it's getting a little hectic.

henders:
Alright @grosser, all good? :)

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -0,0 +1,37 @@
...
+    Rails.logger.warn "valid_pipeline? #{origin_id} / #{id} - #{next_stage_ids}"
+    if next_stage_ids.any? { |next_id| next_id.to_i == origin_id.to_i }
+      errors[:base] << "Stage #{name} causes a circular pipeline with this stage"
+      return false
+    end
+
+    next_stages.each do |stage|
+      unless stage.valid_pipeline?(origin_id)
+        errors[:base] << "Stage #{stage.name} causes a circular pipeline with this stage"
+        return false
```
grosser: did you try just setting the error and not returning false ?

plugins/pipelines/app/views/samson_pipelines/_fields.html.erb
grosser: ^

plugins/pipelines/test/lib/hooks_test.rb
```diff
@@ -0,0 +1,42 @@
...
+      project: stage.project,
+      command: "echo hello world",
+      status: "running",
+      user: User.first,
+      deploy: deploy)
+  end
+
+  describe :after_job_execution do
+    it 'kicks off the next stages in the deploy' do
+      stage.update!(next_stage_ids: next_stages.map(&:id))
```
grosser: can this go into the before block, so they all use the same base, for example test 4 has no ids -> tests the same as test 3

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
henders: I need it to return false as it's called recursively down the chain of linked stages.

grosser: ah, did not think of that :)

grosser:
:+1: so close to 100 comments ... it's a shame ...

plugins/pipelines/test/lib/hooks_test.rb
henders: Hmm... tests 1 & 2 need that line, but tests 3 & 4 can't have it.  So it won't be reducing the number of lines to move this into the before.

henders:
Hah, thanks @grosser, maybe next PR.. Now I just have to figure out why tests work locally but fail on Travis. Weird! Recreated local DB and everything.

plugins/pipelines/test/lib/hooks_test.rb
```diff
@@ -0,0 +1,42 @@
...
+      DeployService.any_instance.expects(:deploy!).never
+      Samson::Hooks.fire(:after_job_execution, job, false, output)
+    end
+
+    it 'does not deploy another if the next_stage_id is nil' do
+      DeployService.any_instance.expects(:deploy!).never
+      Samson::Hooks.fire(:after_job_execution, job, true, output)
+    end
+
+    it 'does not deploy another if the deploy is nil' do
```
grosser: ah ... this is for Macros ... `it does not deploy another when running a macro`

plugins/pipelines/test/lib/hooks_test.rb
```diff
@@ -0,0 +1,42 @@
...
+      job = Job.create(project: stage.project, command: "echo", status: "running", user: User.first)
```
grosser: not sure why this guy cannot have next_stage_ids ... it's a stage after all ...

grosser: ah derp ... ok ... yes the stage does not matter here :D 
could still be set and nothing would happen I guess

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -0,0 +1,37 @@
...
+  def next_stages
+    Stage.find(next_stage_ids)
+  end
+
+  protected
+
+  # Ensure we don't have a circular pipeline:
+  #
+  # potential race-condition if 2 stages are saved at same time:
+  #   stageA saved with pipelines to stageB and stageC
```
sandlerr: if you're really worried you can [lock](http://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html) all the stages

sandlerr: you'd have to build up recursive calls to `project.stages.each(&:lock)` and run your validation in the final block

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -0,0 +1,37 @@
...
+  #   stageC saved with pipeline to stageA   => will validate if stageA above hasn't been written to DB yet
+  def valid_pipeline?(origin_id = id)
+    Rails.logger.warn "valid_pipeline? #{origin_id} / #{id} - #{next_stage_ids}"
+    if next_stage_ids.any? { |next_id| next_id.to_i == origin_id.to_i }
+      errors[:base] << "Stage #{name} causes a circular pipeline with this stage"
```
sandlerr: aren't these both `errors[:next_stage_ids]`?

sandlerr:
Looks great! My only concern is that at this point only admins are able to find out the pipeline, and only if they go to the edit page. Would be great to surface this info so it is obvious to a deployer what will happen when they hit the deploy button.

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
henders: @sandlerr  When they show up in the UI, the formatting looks ugly because it prepends the field name if specified. This makes it look prettier.

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
henders: Yea, I wasn't too concerned with the timing here, so might leave that for a later PR.

henders:
@sandlerr Fair point about the visibility of the pipelines. I haven't seen the ideal spot to slot this into yet, so I'll create another PR for that.

henders:
Just looked at the comment count, @sandlerr were you trolling to do what @grosser couldn't do to get it up to 100+ comments? :)

sandlerr:
Not gonna lie Shane, I like a good milestone. Also want this in-demand
feature to be as good as we can get it!

PR #427, opened 2015-06-01
reduce size of indexed strings so indices created on MySQL [DBA-3014]

jonmoter:
see https://github.com/rails/rails/issues/9855

Yes, I know it's bad form to change migrations that have already been merged. But deploys to staging are failing on these migrations, so the best option is to fix it.

/cc @zendesk/runway 

### References
 - Jira link: https://zendesk.atlassian.net/browse/DBA-3014

### Risks
 - None

### DIFF
--- a/db/migrate/20150508140514_create_builds.rb
+++ b/db/migrate/20150508140514_create_builds.rb
@@ -1,10 +1,10 @@
-class CreateBuildArtifact < ActiveRecord::Migration
+class CreateBuilds < ActiveRecord::Migration
   def change
     create_table :builds do |t|
       t.belongs_to :project,    null: false, index: true
-      t.string :git_sha
+      t.string :git_sha,        limit: 128
       t.string :git_ref
-      t.string :container_sha,  index: true
+      t.string :container_sha,  limit: 128, index: true
       t.string :container_ref
       t.timestamps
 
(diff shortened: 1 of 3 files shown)


db/migrate/20150508140514_create_builds.rb
```diff
@@ -1,10 +1,10 @@
-class CreateBuildArtifact < ActiveRecord::Migration
+class CreateBuilds < ActiveRecord::Migration
   def change
     create_table :builds do |t|
       t.belongs_to :project,    null: false, index: true
-      t.string :git_sha
+      t.string :git_sha,        limit: 128
```
sandlerr: sha1 is always 40 characters - why pick 128?

sandlerr:
:+1: Hope these are the only ones causing problems.

db/migrate/20150508140514_create_builds.rb
jonmoter: sha2 can be up to 128 characters. Probably overkill, but figured it didn't hurt either.

grosser:
that sounds like the wrong solution ... indexes on 255 strings have to work too ...

jonmoter:
@grosser I believe merging this will have the staging builds pass, so that unblocks automated builds. I agree we should come up with a better solution, but I don't know what that is.  Check the comments on DBA-3014.

PR #428, opened 2015-06-01
Don't append to FETCH_HEAD when fetching

jacobat:
The -a flag will make the FETCH_HEAD file grow forever
for no reason.

/cc @steved @grosser


steved:
+1

grosser:
:+1:

PR #429, opened 2015-06-01
add row duplicator for environment variables

grosser:
staging looks silly with all these extra rows ... now uses less space and offers more functionality ...

@zendesk/runway 

screen shot 2015-06-01 at 11 28 12 am

### Risks
 - None

### DIFF
--- a/plugins/env/app/views/samson_env/_environment_variables.html.erb
+++ b/plugins/env/app/views/samson_env/_environment_variables.html.erb
@@ -1,5 +1,5 @@
 <% deploy_groups = [["All", nil]] + DeployGroup.all.map{ |dg| [dg.name, dg.id] } %>
-<% [deploy_groups.size, 2].max.times { form.object.environment_variables.build } %>
+<% form.object.environment_variables.build %>
 <%= form.fields_for :environment_variables do |fields| %>
   <div class="form-group">
     <div class="col-lg-2">
@@ -24,3 +24,4 @@
     <% end %>
   </div>
 <% end %>
+<%= link_to "Add row", "#", class: "duplicate_previous_row" %>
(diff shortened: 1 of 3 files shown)


plugins/env/app/views/samson_env/_environment_variables.html.erb
```diff
@@ -24,3 +24,16 @@
...
 <% end %>
+<%= link_to "Add row", "#", id: "add_environment_variable_row" %>
+
+<script>
+  $(function(){
+    $("#add_environment_variable_row").click(function(){
+      var $row = $(this).prev();
+      var $new_row = $row.clone();
+      $new_row.find('input').val('');
+      $row.after($new_row)
```
sandlerr: missing semicolon is a little distracting.
could also pass the event and do `e.preventDefault();` instead of `return false;`

sandlerr:
do we tend not to create asset files in plugins? Maybe overkill here anyway :+1:

grosser:
converted to assets, added preventDefault and also made groups duplicateable :)

plugins/env/app/views/samson_env/_fields.html.erb
```diff
@@ -15,5 +15,6 @@
       </div>
     <% end %>
   <% end %>
+  <%= link_to "Add row", "#", class: "duplicate_previous_row" %> |
```
sandlerr: sry for nitpicking.. is there a whitespace separating the pipe from the next link? Maybe use a html entity because lots of  people's text editor configs delete trailing whitespace?

grosser: there is no trailing at the right, the next line has whitespace, so it works out fine :)

screen shot 2015-06-01 at 9 06 18 pm

sandlerr: cool :+1: turns out you can explicity remove the whitespace by `<%-  -%>`

PR #430, opened 2015-06-01
fix Project#github_repo so it can handle repos that don't end with ".git"

jonmoter:
This logic error was causing 500 errors on the staging page if the project's repo didn't end with ".git"

/cc @zendesk/runway 

### References
 - Airbrake: https://zendesk.airbrake.io/projects/95346/groups/1433509502896252640

### Risks
 - None


grosser:
:+1:

grosser:
For extra niceness we could do git://xxx.git or http validation

PR #432, opened 2015-06-02
changes for plugin development

sandlerr:
@zendesk/runway

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,10 @@ RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add
 RUN echo 'deb https://deb.nodesource.com/node_0.12 jessie main' > /etc/apt/sources.list.d/nodesource.list
 RUN apt-get update && apt-get install -y nodejs
 
+ENV GEM_HOME=/bundle
+RUN gem update --system
+RUN gem install bundler
+
 RUN mkdir /app
 WORKDIR /app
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -5,6 +5,10 @@ RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add
 RUN echo 'deb https://deb.nodesource.com/node_0.12 jessie main' > /etc/apt/sources.list.d/nodesource.list
 RUN apt-get update && apt-get install -y nodejs
 
+ENV GEM_HOME=/bundle
+RUN gem update --system
+RUN gem install bundler
```
grosser: base image should be responsible for that ?

app/views/shared/_project_tabs.html.erb
```diff
@@ -29,5 +29,6 @@
   <li class="<%= 'active' if active == 'settings' %>">
     <%= link_to "Settings", edit_project_path(project) %>
   </li>
+  <%= Samson::Hooks.render_views(:project_tabs_view, self, { project: @project, active: active }) %>
```
grosser: Including the new hook with the PR that uses it makes it easier to review / change ...

Dockerfile
sandlerr: the base image has a different `GEM_HOME`, its not compatible with our tooling. Also I don't think its bundler is current.

app/views/shared/_project_tabs.html.erb
sandlerr: Agree, the PR that uses it isn't public though.

Dockerfile
sandlerr: zdi automatically sets the `GEM_HOME` to `/bundle` and also runs `bundle check` when running an app with the `-d` option - I was getting errors about bundler not being found because of this. This project is different to others because it uses the Docker hub's (Debian) base Ruby image rather than our internal Ubuntu one.

grosser:
:+1:

PR #433, opened 2015-06-02
Fixing typeahead display problems

fneves:
Samson is incorrectly displaying the auto complete suggestions in the deploys page.
bad_autocomplete

This is caused by the version of the typeahead that we are now using. Since we updated to the rails assets gem we started using the latest version of the typeahead plugin. This version removes the attribute ```value``` as the default key that will be used and changes the plugin generated html.

I've just set the correct key and fixed the styles.

fixed_autocomplete

### References
 - Jira link: [RUN-134](https://zendesk.atlassian.net/browse/RUN-134)

### Risks
 - None

### DIFF
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -66,6 +66,7 @@ $(function () {
     engine.initialize();
 
     $reference.typeahead(null, {
+      display: 'value',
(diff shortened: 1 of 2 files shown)


sandlerr:
:+1: awesome, thanks for debugging that!

PR #434, opened 2015-06-03
change travis mysql config

sandlerr:


sandlerr:
the bulk of the changes to `db/schema.rb` are explicit specifications of the default values

- integer limit: 4
- string limit: 255
- text limit: 65535

db/schema.rb
```diff
@@ -11,15 +11,15 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150530010900) do
+ActiveRecord::Schema.define(version: 20150529162928) do
```
sandlerr: the migration 20150530010900 doesn't exist, probably was created on a branch and flattened into a single migration

db/schema.rb
```diff
@@ -41,52 +41,52 @@
   add_index "builds", ["project_id"], name: "index_builds_on_project_id", using: :btree
 
   create_table "commands", force: :cascade do |t|
-    t.text     "command",    limit: 10485760
+    t.text     "command",    limit: 16777215
```
sandlerr: this is weird, the limit is set to `10.megabytes` which is 10485760 but that probably assuming non-unicode encoding anyway

db/schema.rb
```diff
@@ -133,104 +133,104 @@
...
-    t.text     "command",                                           null: false
-    t.integer  "user_id",                                           null: false
-    t.integer  "project_id",                                        null: false
-    t.string   "status",                        default: "pending"
-    t.text     "output",     limit: 1073741823
+    t.text     "command",    limit: 65535,                          null: false
+    t.integer  "user_id",    limit: 4,                              null: false
+    t.integer  "project_id", limit: 4,                              null: false
+    t.string   "status",     limit: 255,        default: "pending"
+    t.text     "output",     limit: 4294967295
```
sandlerr: this is meant to be `1.gigabyte - 1`, maybe this is too large now.

sandlerr:
cc @zendesk/runway this is green, just codeclimate not working

db/migrate/20140108010302_add_token_to_projects.rb
```diff
@@ -2,7 +2,7 @@ class AddTokenToProjects < ActiveRecord::Migration
   def change
     change_table :projects do |t|
       t.string :token
-      t.index :token
+      t.index :token, length: 191
```
grosser: that is a very odd length oO

db/migrate/20140109053431_job_text_upgrade.rb
```diff
@@ -1,5 +1,5 @@
 class JobTextUpgrade < ActiveRecord::Migration
   def change
-    change_column :jobs, :output, :text, :limit => 1.gigabyte - 1
+    change_column :jobs, :output, :text, :limit => 1.gigabyte / 4 - 1
```
grosser: that seems a bit crazy ... 10m should be enough ...

grosser:
needs some kind of enforcement test so new 255 things do not get added ... I still don't understand why production works and staging does not ... or why all our other projects work without this index patching foobar :(

db/migrate/20140109053431_job_text_upgrade.rb
sandlerr: 1Gb - 1byte is the max that postgresql allows.I'm a bit wary of reducing the limit in case there is old data (either ours or someone else's) that is too large

db/migrate/20140108010302_add_token_to_projects.rb
sandlerr: indices are limited to 767 bytes in mysql and each character is 4 bytes, 767/4 = 191.75. Unfortunately Rails just defaults to 255.

sandlerr:
I've now changed database.travis.yml to use utf8mb4 which means the build will be red for future changes. I'll add a patch to the example database configs but we don't even have database.yml in our docker image (just uses a URL where you can't specify the encoding)

jonmoter:
@grosser in our schema.rb, we currently don't have any indices that are just on a VARCHAR(255) column, except the ones you and I recently added. There are some compound keys that include a VARCHAR(255), like `index_projects_on_permalink_and_deleted_at`.  But I think MySQL handles compound keys differently.  So apparently we just haven't triggered this issue 'til now.

sandlerr:
other projects use `ROW_FORMAT=DYNAMIC`, that was an alternate solution here

sandlerr:
@jonmoter I had to add the length constraint to some old migrations to get db:migrate to work on an empty DB. I don't think all of those indices were later removed, seems like something has changed.

db/migrate/20140109053431_job_text_upgrade.rb
grosser: nah ... even if there is they and us don't want 1GB of foobar just sitting
there

db/migrate/20140108010302_add_token_to_projects.rb
grosser: we have done full indices before ... just need to set

mysql -u root -e "set GLOBAL innodb_large_prefix = true"
mysql -u root -e "set GLOBAL innodb_file_per_table = true"
mysql -u root -e "set GLOBAL innodb_file_format = 'barracuda'"

as we do in all other .travis.yml

jonmoter:
Interesting.  Maybe at some point the Samson DB was configured to utf8, but it was later changed to make the configuration consistent across all DBs. At this point I'm just guessing.

sandlerr:
@zendesk/runway have fixed so that new tables in mysql will allow long indices. Old tables have already been created with the default row_format so we'll have to put up with the 191 character limit. :+1: ?

sandlerr:
@grosser added a comment and a test

gabetax:
So can we skip all the 191 length indices with the ROW_FORMAT=DYNAMIC patch?

grosser:
:+1: at least going forward everything will be smooth ...

sandlerr:
@gabetax the ROW_FORMAT=DYNAMIC patch applies to newly created tables - most of the ones I'm patching already exist in our staging and prod environments as ROW_FORMAT=COMPACT.

db/schema.rb
```diff
@@ -96,23 +96,23 @@
     t.integer "deploy_group_id", limit: 4
   end
 
-  add_index "environment_variables", ["parent_id", "parent_type", "name", "deploy_group_id"], name: "environment_variables_unique_deploy_group_id", unique: true, using: :btree
+  add_index "environment_variables", ["parent_id", "parent_type", "name", "deploy_group_id"], name: "environment_variables_unique_deploy_group_id", unique: true, length: {"parent_id"=>nil, "parent_type"=>191, "name"=>191, "deploy_group_id"=>nil}, using: :btree
```
grosser: this looks broken oO

grosser: ah never mind ... that's what you added in the migration

sj26:
Why does dynamic row format fail the build? This is failing on our local fork (I'm working on removing our local fork) and I don't know why it matters.

steved:
We discussed it a little more here: https://github.com/zendesk/samson/pull/576#discussion_r41328550
It should probably be disabled on the ENV key as well.

sj26:
Thanks @steved!

PR #435, opened 2015-06-03
SAMSON-152 add DEPLOY_URL envvar

zehua:
Added DEPLOY_URL env var

/cc @zendesk/samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-152

### Risks
 - None


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'bundler'
+gem 'bundler', '>= 1.8.4'
```
zehua: As required by rails-assets.

test/models/job_execution_test.rb
```diff
@@ -139,8 +141,10 @@
 
   it 'does not export deploy_group information if no deploy groups present' do
     job.update(command: 'env | sort')
-    execute_job
-    job.output.wont_include "DEPLOY_GROUPS"
+    deploy.stub :stage, stage_no_groups do
```
zehua: Not sure if there's a better way to stub this. Tried stubbing just ```deploy.stage.deploy_groups``` without success.

jonmoter: Can probably just do a `deploy.stage = stage_no_groups`, without needing to stub. Did you try that?

jonmoter:
Looks good to me.  :+1:

test/models/job_execution_test.rb
zehua: I think that would work for this test. But will that affect other tests in the same file?

jonmoter: I don't think so. The next time a test is run, it will re-run all the `let` statements above, which would run the `Deploy.create!(...)` block and construct a new object.

sandlerr:
:+1: let's take it to staging, sometimes the javascript bumps break pages a little bit

PR #436, opened 2015-06-03
[WIP] native support for Samson building Docker Images [RUN-122]

jonmoter:
Not ready to merge yet. Still need to:
1. save build output to table
2. trigger builds automatically (possibly when a `Release` is currently made?)
3. figure out image naming

/cc @zendesk/runway, @zendesk/zendesk-meatballs 

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-122

### Risks
 - None

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,9 @@ gem 'sse-rails-engine'
 gem 'lograge'
 gem 'logstash-event'
 
+# Docker
+gem 'docker-api'
+
 group :production, :staging do
   gem 'rails_12factor'
   gem 'airbrake', '~> 4.1.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -105,6 +105,9 @@ GEM
     dalli (2.7.4)
     debug_inspector (0.0.2)
     docile (1.1.5)
+    docker-api (1.21.4)
+      excon (>= 0.38.0)
+      json
     dogapi (1.17.0)
       multi_json
     dogstatsd-ruby (1.4.1)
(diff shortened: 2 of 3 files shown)


grosser:
looks good

jacobat:
What is the benefit to this approach versus the current approach of using shell scripts?

grosser:
It will give us a bit finer controls and allow for easy output capturing also better for compliance since not everyone can do anything with cap ... but will be less flexible

jonmoter:
@jacobat the benefit is that we'll have a table of `builds` that represent a Git commit -> Docker Image mapping, and we'll have the SHAs of both. Otherwise we don't have anywhere to store that SHA, which is necessary for compliance.

That `Build` is a necessary component for creating pipelines. So the Build is run through integration tests, then deployed to master or staging, etc.  We can attach buddy approvals or other meta-information to a Build, rather than a deploy, or a git reference like "master", which can change.

@sandlerr is working on getting the `meatballs_client` gem integrated into Samson, so that we can create release docs and trigger deploys directly via Samson's business logic, and not require shell scripts to make it happen.

jacobat:
@jonmoter I don't see why you need to integrate Docker into Samson to achieve the goal of git commit -> docker image. You could define a protocol that "a build service" would respect.

I'm asking because I'm afraid we're losing flexibility and making things hard to change. For instance the docker-images/ruby builds uses a two step build and for now some projects will need to do "bundle package" before calling the Docker build command. Trying to cater to these different needs would seem difficult if you're integrating Docker directly in Samson.

jonmoter:
I want to integrate Docker into Samson because Docker is going to be the foundation of our deploy strategy in the future. The point of BigInfra is that we've committed to containers and 12-factor apps as basic patterns we'll be using. In meatballs, you're not making it optional to use Docker containers, right?

Linking together the commit and image is a feature, but it's not the only reason I want to integrate them. If we really don't want to sacrifice any flexibility, we should just stick with Capistrano scripts, and not bother with any of this container stuff.

jonmoter:
Closing this PR, 'cause I have stuff close to working in another branch.

PR #437, opened 2015-06-03
env vars improvements

grosser:
more screen space and reusing other env vars, so we can have NEWRELIC_APP_NAME="Zendesk $SERVICE_NAME pod $ZENDESK_POD_ID" instead of repeating it a bunch of times

screen shot 2015-06-03 at 12 04 17 pm

@zendesk/runway 

### Risks
 - None

### DIFF
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -5,11 +5,17 @@ class EnvironmentVariable < ActiveRecord::Base
 
   def self.env(stage, deploy_group_id)
     variables = stage.environment_variables + stage.environment_variable_groups.flat_map(&:environment_variables)
-    variables.each_with_object({}) do |ev, all|
+    env = variables.each_with_object({}) do |ev, all|
       if !ev.deploy_group_id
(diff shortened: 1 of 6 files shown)


sandlerr:
:+1: hopefully we don't find `\w` too permissive, could be `[A-Z_]` or something? Can these things ever have numbers?
Also might need to bump up the [limit](https://github.com/zendesk/samson/blob/master/db/schema.rb#L93) its 255 chars in the db.

grosser:
trying to stay unix compatible :)
-> [A-Za-z0-9_]

I'll wait for someone to hit the limit ... hopefully we don't ...

gabetax:
:+1:

gabetax:
We may benefit by providing some inline documentation on the existence and use of this feature.

Also, do we want some way to escape this (backslash before `$`?), in case you literally wanted the value to stay `$SERVICE_NAME`?

grosser:
screen shot 2015-06-03 at 6 01 48 pm


if we need escaping it it should be possible to add ... don't want to add a bunch of logic that nobody needs :)

PR #438, opened 2015-06-03
allow scoping environment variables to deploy group or environment fo…

grosser:
…r less repetetive configuration

had to copy-paste pod100 / pod101 configuration ... not fun ... so let's allow to scope by environment too

@zendesk/runway 

screen shot 2015-06-03 at 3 18 13 pm

### Risks
 - None

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -89,14 +89,15 @@ ActiveRecord::Schema.define(version: 20150529162928) do
   add_index "environment_variable_groups", ["name"], name: "index_environment_variable_groups_on_name", unique: true, length: {"name"=>191}, using: :btree
 
   create_table "environment_variables", force: :cascade do |t|
-    t.string  "name",            limit: 255, null: false
-    t.string  "value",           limit: 255, null: false
-    t.integer "parent_id",       limit: 4,   null: false
-    t.string  "parent_type",     limit: 255, null: false
-    t.integer "deploy_group_id", limit: 4
+    t.string  "name",        limit: 255, null: false
+    t.string  "value",       limit: 255, null: false
+    t.integer "parent_id",   limit: 4,   null: false
+    t.string  "parent_type", limit: 255, null: false
+    t.integer "scope_id",    limit: 4
+    t.string  "scope_type",  limit: 255
   end
 
-  add_index "environment_variables", ["parent_id", "parent_type", "name", "deploy_group_id"], name: "environment_variables_unique_deploy_group_id", unique: true, length: {"parent_id"=>nil, "parent_type"=>191, "name"=>191, "deploy_group_id"=>nil}, using: :btree
+  add_index "environment_variables", ["parent_id", "parent_type", "name", "scope_type", "scope_id"], name: "environment_variables_unique_scope", , length: {"parent_id"=>nil, "parent_type"=>191, "name"=>191, "scope_id"=>nil, "scope_type"=>191}, using: :btree
 
   create_table "environments", force: :cascade do |t|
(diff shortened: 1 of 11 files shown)


plugins/env/app/models/environment_variable.rb
```diff
@@ -1,16 +1,41 @@
...
+          (ev.scope_type == "DeployGroup" && ev.scope_id == deploy_group.id) || # matches deploy group
+          (ev.scope_type == "Environment" && ev.scope_id == deploy_group.environment_id) # matches deploy group's environment
+        ))
+      )
+      all[ev.name] = ev.value if match
+    end
+  end
+
+  # used to assign direct from form values
+  def scope_type_and_id=(value)
```
sandlerr: this kinda seems like controller logic since its deserializing POST / PUT requests

plugins/env/app/models/environment_variable.rb
```diff
@@ -1,16 +1,41 @@
...
+  end
+
+  private
+
+  def priority
+    case scope_type
+    when nil then 0
+    when "Environment" then 1
+    when "DeployGroup" then 2
+    else raise "Unsupported type: #{scope_type}"
```
sandlerr: would it be better to raise when assigning / reading the attribute?

plugins/env/db/migrate/20150603211116_change_env_deploy_group_to_scope.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeEnvDeployGroupToScope < ActiveRecord::Migration
+  def change
+    remove_index :environment_variables, name: "environment_variables_unique_deploy_group_id"
+    add_column :environment_variables, :scope_type, :string
+    rename_column :environment_variables, :deploy_group_id, :scope_id
+
+    EnvironmentVariable.update_all scope_type: "DeployGroup"
```
sandlerr: some of our older migrations have `Model.reset_column_information` here, whatever works

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150530010900) do
+ActiveRecord::Schema.define(version: 20150527210505) do
```
sandlerr: nitpick: timestamp of actual migration is 20150603211116 just for neatness

sandlerr:
:+1:

plugins/env/db/migrate/20150603211116_change_env_deploy_group_to_scope.rb
grosser: only need that if you are using update_attribute etc, but not for
update_column or update_all

plugins/env/app/models/environment_variable.rb
grosser: got a validation for that :)

plugins/env/app/models/environment_variable.rb
grosser: yep ... first I wanted to move it into the controller ... but I need it in
2 controllers ... so easier this way I guess ..

sandlerr: that's fine, I think this is better than making mixins for everything.

grosser:
added a comment field for good measure / to communicate things that might be confusing/missing

screen shot 2015-06-03 at 5 38 01 pm

grosser:
screen shot 2015-06-03 at 6 01 48 pm

PR #439, opened 2015-06-03
we do not need the vanilla env config once we generate for every depl…

grosser:
…oy group and keeping it makes required complicated

leaving the default file means:
 - all variables need a all default -> will not see if something is missing
 - adds the possibility of screwing up by somehow picking the default

@zendesk/runway 

all our projects now use ENV.xxx.json so ENV.json can go.
for everything else you now either pick deploy groups or not, but there is no more middle ground 

### Risks
 - None


sandlerr:
so the base file will continue to exist just to specify required keys?

grosser:
exactly, everything else is filled out via samson env variables

grosser:
+ it has service and other foobar

sandlerr:
:+1:

PR #440, opened 2015-06-04
turn off debug assets in development

sandlerr:
cc @zendesk/runway dev site is quite slow and needs a lot of requests to load each page, this helps a lot.


gabetax:
:+1: This has also been frustrating me.

PR #441, opened 2015-06-04
RUN-162 Fixed datadog tags striping semicolon bug

zehua:
Instead of overriding the ```datadog_tags``` field to return an array, we keep it and rename the method to ```datadog_tags_as_array```.

/cc @zendesk/samson 

I will be on a two-week leave since tomorrow. Feel free to update and merge when it looks good to you guys.

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-162

### Risks
 - None


grosser:
:+1:

gabetax:
:+1: Using the same field for serialized and non-serialized values is always trouble.

PR #442, opened 2015-06-04
add REVISION to Dockerfile

gabetax:
REVISION file is necessary for ZDI to push a release. Its creation is managed by the ZDI build process.

/cc @zendesk/runway


grosser:
:+1:

PR #443, opened 2015-06-05
better errors instead of web console

sandlerr:
@zendesk/runway @zendesk/samson 
the main advantage here is better_errors REPL is in the context of exceptions


grosser:
:+1: but maybe wait for more feedback ...

FYI afaik better errors won't work in boxen/docker-images since /__better_errors is routed to hc
https://github.com/zendesk/zendesk_boxen/blob/master/modules/projects/templates/help_center.nginx.conf.erb#L15-L20 not sure about docker-images though ...

sandlerr:
I'm running it successfully in docker-images because samson is its own `server` block. For multiple apps within the same server block, we may have to patch the gem to customise the route.

gabetax:
:+1: I'm a fan.

PR #445, opened 2015-06-05
we have a bunch of long env names ... need more space

grosser:
@zendesk/runway 

screen shot 2015-06-05 at 9 21 21 am
screen shot 2015-06-05 at 9 22 25 am

### Risks
 - None


jonmoter:
:+1:

PR #446, opened 2015-06-06
Refactor and clean up some GitRepository code

jonmoter:
DRY up some test support code related to GitRepository.
Allow GitRepository#commit_from_ref to return the full SHA, and handle errors.
More tests for GitRepository#commit_from_ref.

/cc @zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -29,14 +29,16 @@ class GitRepository
     executor.execute!("cd #{repo_cache_dir}", 'git fetch -p')
   end
 
-  def commit_from_ref(git_reference)
-    description = Dir.chdir(repo_cache_dir) do
-      IO.popen(['git', 'describe', '--long', '--tags', '--all', git_reference]) do |io|
+  def commit_from_ref(git_reference, length: 7)
+    Dir.chdir(repo_cache_dir) do
+      description = IO.popen(['git', 'describe', '--long', '--tags', '--all', "--abbrev=#{length || 40}", git_reference], err: [:child, :out]) do |io|
         io.read.strip
       end
-    end
 
-    description.split('-').last.sub(/^g/, '')
+      return nil unless $?.success?
+
+      description.split('-').last.sub(/^g/, '')
+    end
   end
 
   def tag_from_ref(git_reference)
--- a/test/models/git_repository_test.rb
+++ b/test/models/git_repository_test.rb
@@ -1,14 +1,15 @@
 require_relative '../test_helper'
 
 describe GitRepository do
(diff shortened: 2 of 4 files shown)


app/models/git_repository.rb
```diff
@@ -29,14 +29,16 @@ def update!(executor: TerminalExecutor.new(StringIO.new))
     executor.execute!("cd #{repo_cache_dir}", 'git fetch -p')
   end
 
-  def commit_from_ref(git_reference)
-    description = Dir.chdir(repo_cache_dir) do
-      IO.popen(['git', 'describe', '--long', '--tags', '--all', git_reference]) do |io|
+  def commit_from_ref(git_reference, length: 7)
```
grosser: with `length: 40` here the || 40 can go away ?

app/models/git_repository.rb
```diff
@@ -29,14 +29,16 @@ def update!(executor: TerminalExecutor.new(StringIO.new))
...
+    Dir.chdir(repo_cache_dir) do
```
grosser: maybe end the chdir block right here and then deal with the description outside of it, might make refactoring into a helper more obvious

test/models/git_repository_test.rb
```diff
@@ -65,6 +66,41 @@
...
       repository.commit_from_ref('master').must_match /^[0-9a-f]{7}$/
     end
+
+    it 'returns the full commit id with nil length' do
+      create_repo_with_tags
+      repository.clone!
+      repository.commit_from_ref('master', length: nil).must_match /^[0-9a-f]{40}$/
+    end
+
+    it 'returns nil if ref does not exist' do
```
grosser: a `does not allow script insertion` with `ref = 'head" ; do something evil'` could be nice

test/models/job_execution_test.rb
```diff
@@ -17,15 +18,7 @@
...
-    execute_on_remote_repo <<-SHELL
-      git init
-      git config user.email "test@example.com"
-      git config user.name "Test User"
-      echo monkey > foo
-      git add foo
-      git commit -m "initial commit"
-      git tag v1
-    SHELL
+    create_repo_with_tags('v1')
```
grosser: :heart:

grosser:
:+1:

app/models/git_repository.rb
jonmoter: The previous behavior of the method was that it returned a SHA with a length of 7. I wanted to keep the default behavior the same.

jonmoter:
Thanks.  I'm going to add that test, then squash all these into a single commit.  Once the tests pass, I'll merge.

PR #447, opened 2015-06-08
Docker database fixes

gabetax:
Previous Dockerfile shipped without any database configuration. I also had to change the seeds.rb file to use http based authentication. Github does not allow anonymous ssh communication, and requires a key to be present. Because of this, `rake db:reset` would hang during the `Project.create!` validation, which executes a `git ls-remote` on the repository. We could alternative create a dedicated dummy user, or we could mount in SSH keys (potentially risky?).

/cc @zendesk/runway


config/database.docker.yml
```diff
@@ -0,0 +1,23 @@
+<%
+  mysql    = URI(ENV['MYSQL_URL'] || 'mysql://root@127.0.0.1:3306/samson_development')
```
grosser: can this go into the normal database.yml ?

gabetax: Sure -- there is already a ton of database.yml files there. Is config/database.samson.yml doing anything, or can we eliminate that one too while we're at it?

grosser: idk ... kill whatever you can :D

sandlerr:
I should have mentioned this to you earlier - Rails allows to omit database.yml and start the application with `DATABASE_URL=$MYSQL_URL bundle exec rails s`. However that does not allow configs that can't be specified in a URI such as `collation: utf8mb4`.

This is probably a better fix than that :+1:

gabetax:
Just using "url" is easier, but it doesn't handle the test database, right? Maybe something like:

```yaml
<% uri = URI(ENV['MYSQL_URL'] || 'mysql://root@127.0.0.1:3306/samson_development') %>

development:
  url: <%= uri %>

test:
  <% uri.path = '/samson_test' %>
  url: <%= uri %>
```

sandlerr:
I'm honestly not sure how people are running tests using docker-images. I'm just shelling into the container, so happy to go with either of your solutions or even providing a different environment variable for the database when running specs.

gabetax:
@steved Any strong opinions on this setup?

steved:
I like the idea of using the Rails standard DATABASE_URL. As for the seed data, we'll probably have to fully document how to integrate ssh authentication, but https is fine for the demo. Essentially, no I have no strong opinions either way.

config/database.mysql.yml.example
```diff
@@ -1,18 +1,11 @@
...
-  host: 127.0.0.1
+<%
+  mysql    = URI(ENV['MYSQL_URL'] || 'mysql2://root@127.0.0.1:3306/samson_development')
+  database = mysql.path.gsub(%r(^/), "")
+%>
 
 development:
-  <<: *common
-  database: samson_development
+  url: <%= uri %>
```
sandlerr: the variable name should be `mysql` or `database` right?

PR #448, opened 2015-06-08
refactor more GitRepository code to simplify parameters passed in

jonmoter:
/cc @zendesk/runway 

### Risks
 - Confirm that logic still works for running git commands

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -7,10 +7,7 @@ class GitRepository
     Rails.application.config.samson.cached_repos_dir
   end
 
-  def initialize(repository_url: nil, repository_dir: nil, executor: nil)
-    raise ArgumentError.new("GitRepository.new: repository_url is required") if repository_url.blank?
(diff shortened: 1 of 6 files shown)


grosser:
:+1:

app/models/git_repository.rb
```diff
@@ -1,31 +1,39 @@
...
+  attr_accessor :executor
 
   # The directory in which repositories should be cached.
   cattr_accessor(:cached_repos_dir, instance_writer: false) do
     Rails.application.config.samson.cached_repos_dir
   end
 
-  def initialize(repository_url:, repository_dir:)
+  def initialize(repository_url: nil, repository_dir: nil, executor: nil)
+    raise ArgumentError.new("GitRepository.new: repository_url is required") if repository_url.blank?
```
sandlerr: these were already required. Just testing for empty string?

jonmoter: Ah, I didn't realize the `repository_url:,` syntax meant that it was required.  Will revert.

sandlerr:
:+1:

PR #449, opened 2015-06-08
fix jobs

sandlerr:
cc @zendesk/runway @zendesk/samson


grosser:
:+1:

test/models/job_execution_test.rb
```diff
@@ -124,6 +124,14 @@
     lines.must_include "TAG=v1"
   end
 
+  it 'works without a deploy' do
```
zehua: good one. wish this test was there before. :)

PR #450, opened 2015-06-08
scope environment variable groups by project and not by stage

grosser:
1 project has a set of settings, stages are just a selection of deploy groups

@zendesk/runway 

### Risks
 - None

### DIFF
--- a/plugins/env/app/controllers/admin/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/admin/environment_variable_groups_controller.rb
@@ -34,10 +34,10 @@ module Admin
     def preview
       @groups = if params[:group_id]
         @group = EnvironmentVariableGroup.find(params[:group_id])
-        SamsonEnv.env_groups(Stage.new(environment_variable_groups: [@group], deploy_groups: DeployGroup.all))
+        SamsonEnv.env_groups(Project.new(environment_variable_groups: [@group]), DeployGroup.all)
       else
-        @stage = Stage.find(params[:stage_id])
-        SamsonEnv.env_groups(@stage)
+        @project = Project.find(params[:project_id])
+        SamsonEnv.env_groups(@project, DeployGroup.all)
       end
     end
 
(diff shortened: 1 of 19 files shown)


plugins/env/test/controllers/environment_variable_groups_controller_test.rb
```diff
@@ -2,8 +2,9 @@
 
 describe Admin::EnvironmentVariableGroupsController do
   let(:stage) { stages(:test_staging) }
+  let(:project) { stage.project }
   let(:deploy_group) { stage.deploy_groups.first }
-  let(:env_group) do
+  let!(:env_group) do
```
sandlerr: let! is making a test fail, there is G1 on line 63 but uniqueness constraint

grosser: fixed, good to go ?

plugins/env/app/controllers/admin/environment_variable_groups_controller.rb
```diff
@@ -34,10 +34,10 @@ def destroy
     def preview
       @groups = if params[:group_id]
         @group = EnvironmentVariableGroup.find(params[:group_id])
-        SamsonEnv.env_groups(Stage.new(environment_variable_groups: [@group], deploy_groups: DeployGroup.all))
+        SamsonEnv.env_groups(Stage.new(project: Project.new(environment_variable_groups: [@group]), deploy_groups: DeployGroup.all))
```
sandlerr: This signature is becoming kinda complicated, If you were starting from scratch would it still take a stage? Maybe a project and an array or splat of deploy groups?

sandlerr: I guess this applies to other methods around this plugin as well so no prob if you want to leave it as is

sandlerr:
:+1:

grosser:
updated the interface to project + deploy_groups

PR #452, opened 2015-06-08
samson should default to ssl in production

kintner:
Samson should have safe defaults. 

/cc @zendesk/secdev @steved @jonmoter


sandlerr:
nginx takes care of it also but :+1:

kintner:
This setting sets the secure flag on cookies, enables HSTS, etc.

PR #453, opened 2015-06-09
build docker images from Build objects

jonmoter:
This PR adds functionality for building docker images using the docker-api gem.

Still a work in progress.  Things to do before merging:
- [x] add tests for DockerBuilderService
- [x] add tests for BuildsController
- [x] better error handling
- [x] add ENV vars or global preferences for docker registry
- [ ] authenticate with Docker registry
- [x] test pushing images
- [ ] specify whether a project uses Docker builds

/cc @zendesk/runway

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-122

### Risks
 - Changes to JobExecution class
 - A bunch of new Docker-related functionality


jonmoter:
FYI, here's a screenshot of the builds/show view after a docker build has been created:

build_3_-_account_service

app/controllers/builds_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+        render json: {}, status: 200
+      end
+    end
+  end
+
+  private
+
+  def find_project
+    @project = Project.find_by_param!(params[:project_id])
+  rescue ActiveRecord::RecordNotFound
```
sandlerr: what's the default exception behaviour here?

app/controllers/builds_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    redirect_to root_path
+  end
+
+  def find_build
+    @build = Build.find(params[:id])
+  end
+
+  def new_build_params
```
sandlerr: Not sure if we have a standard name for these things, there is `stage_permitted_params`, would be good to be consistent

app/helpers/builds_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def build_title build
+    "Build #{build.label || build.id}"
+  end
+
+  def short_sha value
+    "#{value.first(7)}..." if value
+  end
+
+  def git_ref_and_sha_for build, make_link: fase
```
sandlerr: `false`

app/controllers/builds_controller.rb
jonmoter: By default it shows a 404 page, like [this example](https://samsontest.zende.sk/projects/blah/deploys).  I suppose I could follow that behavior too.

app/models/build.rb
```diff
@@ -1,16 +1,61 @@
 class Build < ActiveRecord::Base
-  SHA1_REGEX = /\A[0-9a-f]{40}\Z/i
+  SHA1_REGEX = /\A[0-9a-f]{40,128}\Z/i
```
sandlerr: Does the name still match the functionality?

app/controllers/builds_controller.rb
jonmoter: In a lot of controllers, it's just "#{model}_params", like `deploy_params`, `job_params`, etc.

I'm going to create an `edit` endpoint that allows you to change some elements of a build, but not others.  So that will probably use a method like `edit_build_params`.

app/models/build.rb
```diff
@@ -1,16 +1,61 @@
...
+      'not built'
+    end
+  end
+
+  private
+
+  def validate_git_reference
+    return unless git_ref.present? || git_sha.present?
+
+    project.repository.setup_local_cache!
```
sandlerr: alternately we could use octokit here to speed up the response

app/models/docker_builder_service.rb
```diff
@@ -0,0 +1,93 @@
...
+    if push
+      # TODO: authenticate with the Docker registry
+      execution.output.write "Pushing Docker image to #{build.project.docker_repo_name}/#{build.docker_ref}"
+      @image.push do |output_chunk|
+        execution.output.write(parse_output_chunk(output_chunk))
+      end
+    end
+
+    @image
+  rescue Docker::Error::DockerError => e
```
sandlerr: do these set the job status?

app/models/docker_builder_service.rb
```diff
@@ -0,0 +1,93 @@
...
+      job.project = build.project
+      job.user_id = build.created_by || NullUser.new.id
+      job.command = ''    # doesn't seem right, but command can't be nil
+      job.commit = build.git_sha
+      job.tag = build.git_ref
+    end
+
+    build.save!
+
+    job_execution = JobExecution.start_job(build.git_sha, job) do |execution, tmp_dir|
```
sandlerr: so this doesn't actually return a JobExecution instance, not sure what a better variable name would be

app/views/builds/new.html.erb
```diff
@@ -0,0 +1,63 @@
...
+          <%= form.label :description, "Description", class: "col-lg-2 control-label" %>
+          <div class="col-lg-4">
+            <%= form.text_area :description, rows: 3, class: "form-control" %>
+          </div>
+        </div>
+
+        <div class="form-group">
+          <%= form.label :build_image, "Build Docker Image?", class: "col-lg-2 control-label" %>
+          <div class="col-lg-4">
+            <%= check_box_tag :build_image %>
```
sandlerr: could it be `form.check_box`?

app/helpers/builds_helper.rb
jonmoter: Oops. Thanks!

app/models/build.rb
jonmoter: Hmm, yeah. Probably a better idea to keep the SHA1 regex as is, and create a new regex to validate a Docker digest.

app/models/docker_builder_service.rb
jonmoter: It doesn't?  I'm pretty sure that's what `JobExecution.start_job` returns.

app/models/docker_builder_service.rb
jonmoter: If I catch the error and return `nil`, then the job is marked as a failed.  If I allow the exception to bubble up, then it runs `JobExecution#error!`, which sets the job to be in an "error" state.  Maybe it's better to let the exception bubble up.

app/views/builds/new.html.erb
jonmoter: If I use `form.check_box :build_image`, then it tries to call the `#build_image` method on the `@build` object. Since it doesn't exist, that throws an error.

app/models/docker_builder_service.rb
sandlerr: its [implicit return](https://github.com/zendesk/samson/pull/453/files#diff-199416dcde8818efe1b2180a2a2e9141L176) is used by the subscribe call a few lines down from here. The `tap` makes it look like it was originally supposed to return an instance.

sandlerr:
Looks good!

app/helpers/builds_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+  def build_page_title
+    "#{@build.nice_name} - #{@project.name}"
+  end
+
+  def build_title build
+    "Build #{build.label || build.id}"
+  end
+
+  def short_sha value
+    "#{value.first(7)}..." if value
```
sandlerr: could the ellipsis go in the views rather than the helper?

jonmoter: Sure, fair enough

app/views/builds/_builds_table.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    <%= link_to "Create Build", new_project_build_path(@project), class: "btn btn-default" %>
+  </p>
+
+  <table class="table">
+    <thead>
+    <tr>
+      <% unless @project %>
+          <th>Project</th>
+      <% end %>
+      <th>Id</th>
```
sandlerr: probably don't need this. It's a samson internal value and it's available in the URL to anyone that really needs it

sandlerr: maybe have the label instead?

app/views/builds/edit.html.erb
```diff
@@ -0,0 +1,69 @@
...
+                  <% @build.errors.full_messages.each do |message| %>
+                      <li><%= message %></li>
+                  <% end %>
+                </ul>
+              </div>
+            </div>
+        <% end %>
+
+        <div class="form-group">
+          <%= form.label :id, "Id", class: "col-lg-2 control-label" %>
```
sandlerr: also probably don't need the id

jonmoter:
This PR should be ready for review and merging now, @zendesk/runway

app/models/docker_builder_service.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    job_execution.on_complete do
+      send_after_notifications
+    end
+  end
+
+  def build_image(tmp_dir)
+    repository.setup!(tmp_dir, build.git_sha)
+
+    File.open("#{tmp_dir}/REVISION", 'w') { |f| f.write build.git_sha }
```
grosser: you can use `File.write file, content`

app/models/output_buffer.rb
```diff
@@ -27,6 +27,14 @@ def initialize
     @closed = false
   end
 
+  def puts(line)
```
grosser: this seems a bit counter-intuitive ... puts xxx.strip for usages with trailing whitespace instead ?

app/views/builds/edit.html.erb
```diff
@@ -0,0 +1,69 @@
...
+<section>
+  <%= form_for [@project, @build], html: { class: "form-horizontal" } do |form| %>
+      <fieldset>
+        <% if @build.errors.any? %>
+            <div class="row">
+              <div class="col-lg-5 col-lg-offset-1 alert alert-warning">
+                <p><strong>Ooops!</strong> There was an error:</p>
+                <ul>
+                  <% @build.errors.full_messages.each do |message| %>
+                      <li><%= message %></li>
```
grosser: indent is off

app/views/builds/edit.html.erb
```diff
@@ -0,0 +1,69 @@
...
+          </div>
+        </div>
+
+        <div class="form-group">
+          <div class="col-lg-offset-2 col-lg-10">
+            <%= form.submit "Save", class: "btn btn-primary" %>
+            <%= link_to "Cancel", :back, class: 'btn btn-default' %>
+          </div>
+        </div>
+      </fieldset>
```
grosser: also indent off here

app/views/builds/new.html.erb
```diff
@@ -0,0 +1,63 @@
+<%= breadcrumb @project, "Build" %>
+
+<section>
+  <%= form_for [@project, @build], html: { class: "form-horizontal" } do |form| %>
+      <fieldset>
```
grosser: same here

grosser:
code looks fine, I'm not really clear how this will be used
step 1: create new builds
step 2: deploy it ? how / where ?

jonmoter:
@grosser, this code handles the building of Docker images, and keeping track of the digest, which is required for security/compliance reasons.

@sandlerr has some code that leverages the `Build` object to write out a meatballs release document to one or more DeployGroups.

Once this gets merged in, I want to set up some hooks so that the docker images are automatically built and pushed on a merge to master, and/or a successful run of CI tests.

Thanks for the feedback.  I'll update the PR with your suggestions.

grosser:
:+1:

PR #455, opened 2015-06-09
each row needs a new unique name and id to make rails do the update l…

grosser:
…ogic correctly for environment_variables

atm you can only add 1 environment variable row at a time since the name is just copied (xxx[3][name])

now it will bump that number by 1 every time you hit 'new row'

@zendesk/runway 

### Risks
 - None


sandlerr:
this handles `input`, isn't there also `textarea` and `select`?

grosser:
fixed it to deal with all kinds of inputs, good to go ?

sandlerr:
:+1:

PR #459, opened 2015-06-10
this shouldn't be unique

sandlerr:
cc @jonmoter (this PR points to your branch)


sandlerr:
encountered an error when trying to build a particular commit a second time, this should definitely be possible.

PR #460, opened 2015-06-11
persist full output when an exception happens

grosser:
atm when for example an env variable is missing an exception is raised that is only visible to the deployer and is not stored ... lets fix that :)

@zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -22,7 +22,7 @@ class DeployService
     job_execution = JobExecution.start_job(deploy.reference, deploy.job)
     send_sse_deploy_update('start', deploy)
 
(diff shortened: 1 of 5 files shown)


jonmoter:
You could DRY it up by putting the `@job.update_output!(OutputAggregator.new(@output).to_s)` in the `ensure` block, rather than having it in both `run!` and `error!`

jonmoter:
And FYI, I was writing some code this afternoon to handle the exact same scenario.  :-)

grosser:
dried up and renamed subscribe to on_complete ... looks good ?

jonmoter:
:+1:

PR #461, opened 2015-06-11
Roman/more docker native

sandlerr:
@jonmoter this is your branch rebased with a couple of small changes I needed to get it to work


jonmoter:
FYI, I updated my branch to include Grosser's changes and your fix.  That way we can keep the comment history in that PR.

PR #462, opened 2015-06-12
kill -9 if command is not stopped after 10 seconds with a kill -2

grosser:
@zendesk/runway @steved @haabaato 

no more going to console (I hope ...)

### Risks
 - None


sandlerr:
:+1:

grosser:
seems to work on staging ... after hitting stop + waiting 10 seconds the build stopped :)

```
» trap "" 2; echo 111 && sleep 6 && echo 111 && sleep 60
111
111
sh: line 38: 25441 Killed                  sleep 60
```

PR #464, opened 2015-06-15
not having a revision is bad ...

grosser:
@gabetax 

### Risks
 - None

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,6 @@ RUN gem install bundler
 RUN mkdir /app
 WORKDIR /app
 
-ADD REVISION /REVISION
-
 # Mostly static
 ADD config.ru /app/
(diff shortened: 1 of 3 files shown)


gabetax:
:+1:

grosser:
added a bunch of other docker fixes, looks good ?

Dockerfile
```diff
@@ -41,6 +36,9 @@ ADD config /app/config
 ADD app /app/app
 ADD lib /app/lib
 
+ADD REVISION /
```
grosser: @jacobat do we need REVISION in the root for some reason / is just app good enough ?

gabetax: ZDI uses it in the release task.

gabetax:
:+1:

PR #467, opened 2015-06-17
Make 'Deploying' a link

adammw:
:koala:

samson

/cc @princemaple @sandlerr


princemaple:
cc @zendesk/samson
:+1: if deploy touches stage, on phone, too hard to check.

sandlerr:
:+1:

PR #468, opened 2015-06-17
Run on_completed hooks after updating the log

kureikain:
Hi,

Currently the `on_completed` hook is invoke before the output is stored. Therefore, it makes subscribers have no way to access the log, such as emailing it out(nicer than viewer it in the app). This change move it down after we store log output.

cc @zendesk/runway

### DIFF
--- a/test/models/job_execution_test.rb
+++ b/test/models/job_execution_test.rb
@@ -168,6 +168,18 @@ describe JobExecution do
     JobExecution.find_by_id(job.id).must_be_nil
   end
 
+  it 'calls subscribers after finishing' do
+    called_subscriber = false
+    execute_job { called_subscriber = true }
+    assert_equal true, called_subscriber
+  end
+
+  it 'saves job output before calling subscriber' do
+    output = nil
+    execute_job { output = job.output }
+    assert_equal 'monkey', output.split("\n").last.strip
+  end
+
   it 'cannot setup project if project is locked' do
     JobExecution.any_instance.stubs(lock_timeout: 0.5) # 2 runs in the loop
     project.repository.expects(:setup!).never
@@ -234,6 +246,7 @@ describe JobExecution do
 
   def execute_job(branch = 'master')
     execution = JobExecution.new(branch, job)
+    execution.on_complete { yield } if block_given?
     execution.send(:run!)
   end
 
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -86,12 +86,12 @@ class JobExecution
       @job.fail!
     end
 
-    @subscribers.each(&:call)
   rescue => e
     error!(e)
   ensure
     @output.close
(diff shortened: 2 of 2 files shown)


kureikain:
Hi @grosser 

This is a minor change and not sure if my approaching is good. Basically I want to have access to the output in `after_deploy` hook. Is this a good way?

grosser:
that looks good ... I think this was broken by a recent refactor ... can you add a test to document the way it is supposed to work ?

Btw are you working on a plugin / care to link it in the Readme /

test/models/job_execution_test.rb
```diff
@@ -234,6 +246,7 @@
 
   def execute_job(branch = 'master')
     execution = JobExecution.new(branch, job)
+    execution.on_complete { yield } if block_given?
```
kureikain: Make it easier to test with on_complete by passing a block as subscriber

test/models/job_execution_test.rb
```diff
@@ -168,6 +168,18 @@
     JobExecution.find_by_id(job.id).must_be_nil
   end
 
+  it 'calls subscribers after finishing' do
```
kureikain: ensure we called subsribers

test/models/job_execution_test.rb
```diff
@@ -168,6 +168,18 @@
...
+    called_subscriber = false
+    execute_job { called_subscriber = true }
+    assert_equal true, called_subscriber
+  end
+
+  it 'saves job output before calling subscriber' do
+    output = nil
+    execute_job do output = job.output end
```
kureikain: when subscribe is invoked, job's output should be stored.

grosser: please use curly lambdas for 1-liners :)

grosser:
looks good :)

kureikain:
@grosser I add 2 test cases:

- test call subscriber (I don't see us test it yet)
- test that once subscriber is invoked, the output should be stored in `job` object(in other word, job's record in database)

I refactor `execute_job` a bit so I can easily pass a block as a subscriber. 2/3 build is passed, one failed. I'm trying to rebuild.

And Yes, I'm working on simple plugin to email out the log of output. Nothing fancy but will add in readme once I'm done.

test/models/job_execution_test.rb
grosser: can you fix up this guy ?

kureikain:
@grosser Hey, the build is passed. Any chance this can be merged?

test/models/job_execution_test.rb
grosser: ^ can you change this to curl ?

kureikain:
@grosser I updated one line `do` block to curly lambda(is it correct, btw?) and rebase to current master

grosser:
:+1:

PR #469, opened 2015-06-17
user misconfiguration does not need to be reported to airbrake

grosser:
previous errors were mor of the mysql is broken / the app is failing kind,
but this is simply user error and does not need to be reported

@zendesk/runway 

### Risks
 - None

### DIFF
--- a/config/database.mysql.yml.example
+++ b/config/database.mysql.yml.example
@@ -1,4 +1,7 @@
-<% uri = URI(ENV['MYSQL_URL'] || 'mysql://root@127.0.0.1:3306/samson_development') %>
+<%
+  uri = URI(ENV['MYSQL_URL'] || 'mysql://root@127.0.0.1:3306/samson_development')
+  uri.scheme = 'mysql2'
+%>
 
 development:
   url: <%= uri %>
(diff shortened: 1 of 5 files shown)


config/database.mysql.yml.example
```diff
@@ -1,4 +1,7 @@
-<% uri = URI(ENV['MYSQL_URL'] || 'mysql://root@127.0.0.1:3306/samson_development') %>
+<%
+  uri = URI(ENV['MYSQL_URL'] || 'mysql://root@127.0.0.1:3306/samson_development')
```
grosser: @gabetax fyi this needed mysql2

grosser: leaving the mysql on purpose to only have 1 code-path

lib/samson/hooks.rb
```diff
@@ -1,5 +1,8 @@
 module Samson
   module Hooks
+    class UserError < StandardError
```
grosser: this is a weird place to add the error ... but idk where else to add ...

jonmoter:
Seems reasonable to me.  :+1:

config/database.mysql.yml.example
gabetax: I made the same change in my other repo. good call.

Issue #465, opened 2015-06-15
Samson won't start with current master without a docker daemon or a valid DOCKER_URL


kureikain:
Hi,

Samson seems introduce a bunch of Docker stuff and in a default environment, especially on Mac OSX, it won't start if we don't config DOCKER_URL to point to boot2docker such as tcp://192.168.59.103:2376. The issue is we don't have it in env file which generate when run bootstrap script. It seems docker is a requirement for Samson now which may make run samson inside Docker even harder?

I think we should change the env to DOCKER_HOST because that's what boot2docker use on Mac OSX. For Linux, they had socket.

So, if we want to continue use DOCKER_URL, we should put it in .env somewhere, or use DOCKER_HOST.

jonmoter:
Oh sorry, I didn't see this issue 'til now.  It wasn't intended for Docker to be required to run Samson. That was a snafu in the initializer code.  The intention is that Samson supports Docker, but doesn't require it.

kureikain:
Thanks @jonmoter Can you tell more about this new Docker feature? I made use of Docker heavily and not sure how I can utilize it.

jonmoter:
Recently I added a "builds" table. The idea is that a "build" is a particular version of the code (i.e. a GitHub commit sha) coupled with a Docker image.  I recently merged in some changes so that when you create a Build object, you can have that automatically run a `docker build` and a `docker push` command.

The next step is to (optionally) have that happen automatically when you merge to master (or whatever branch you choose.)

We want to get to the point where you can run a deploy script on a specified build. The deploy can be run on a build, and have the docker image passed in, so that the deploy script is basically just running "docker run $DOCKER_IMAGE_URL" on whichever nodes you want to deploy your code to.

Does that make sense?

PR #470, opened 2015-06-18
don't crash on startup if Docker isn't running (oops!)

jonmoter:
/cc @zendesk/runway 

### Risks
 - None


kureikain:
This is https://github.com/zendesk/samson/issues/465, right?

jonmoter:
Yup, this should fix #465.

grosser:
:+1:

PR #471, opened 2015-06-18
add command_name to deploy groups so they can be different from the u…

grosser:
…i name

atm we are not using $DEPLOY_GROUPS because they poorly map to some stages like master/staging/gamma servers, this should make this easier and we can finally start exclusively using $DEPLOY_GROUPS

@craig-day @jcheatham @zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/models/deploy_group.rb
+++ b/app/models/deploy_group.rb
@@ -5,7 +5,8 @@ class DeployGroup < ActiveRecord::Base
   has_and_belongs_to_many :stages
 
   validates_presence_of :name, :environment_id
-  validates_uniqueness_of :name
+  validates_uniqueness_of :name, :env_value
+  before_validation :initialize_env_value, on: :create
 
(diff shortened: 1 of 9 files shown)


app/models/deploy_group.rb
```diff
@@ -20,4 +21,10 @@ def deploys
   def long_name
     "#{name} (#{environment.name})"
   end
+
+  private
+
+  def initialize_command_name
+    self.command_name = name if command_name.blank?
```
gabetax: I wish there was an `||=` equivalent for .presence.

jcheatham:
:+1:

gabetax:
I'm not sure `command_name` is the most accurate name here. Maybe `env_value`?

:+1: otherwise.

grosser:
changed to env_value

PR #472, opened 2015-06-18
copy all things from manifest into ENV.json so we can get rid of it

grosser:
@zendesk/runway 

### Risks
 - None

### DIFF
--- a/plugins/env/lib/samson_env/samson_plugin.rb
+++ b/plugins/env/lib/samson_env/samson_plugin.rb
@@ -38,12 +38,17 @@ module SamsonEnv
 
     # writes a proprietary .json file with a env hash for each deploy group
     def write_env_json_file(env_json, manifest_json, groups)
-      return unless File.exist?(env_json)
-      json = JSON.load(File.read(env_json))
-      File.unlink(env_json)
+      return unless File.exist?(manifest_json)
+      json = if File.exist?(env_json)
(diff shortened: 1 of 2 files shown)


gabetax:
:+1:

plugins/env/lib/samson_env/samson_plugin.rb
```diff
@@ -38,12 +38,17 @@ def write_dotenv_file(base_file, groups)
 
     # writes a proprietary .json file with a env hash for each deploy group
     def write_env_json_file(env_json, manifest_json, groups)
-      return unless File.exist?(env_json)
-      json = JSON.load(File.read(env_json))
-      File.unlink(env_json)
+      return unless File.exist?(manifest_json)
```
gabetax: This method is starting to grow a little long. May be worth refactoring if it gets any larger. Were you planning on using the zendesk_service_manifest gem for this down the road?

grosser: not a big fan of this gem ... lots of code for very little functionality + it's private ...

grosser: also the env code will get deleted once rosetta+account are deployed with their update, so it should get less complex soon :)

grosser:
confirmed on staging, here we go :)
https://samsontest.zende.sk/projects/rosetta/deploys/2360

PR #473, opened 2015-06-18
Add extra hook to custom deployment

kureikain:
This PR add some hooks so user can add custom fields for adding deployment.

## View hooks

* before_deploy_form
* after_deploy_form

So user can hook into to display content around deploy form. In my case, I want to add some extra fields.

## Even hooks

* after_deploy_create
* deploy_permitted_params

So user can permit custom field and save those extra fields after `deploy` is created.

### DIFF
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -41,7 +41,7 @@ class DeploysController < ApplicationController
 
   def create
     deploy_service = DeployService.new(current_user)
-    @deploy = deploy_service.deploy!(stage, reference)
+    @deploy = deploy_service.deploy!(stage, deploy_params)
 
     respond_to do |format|
       format.html do
@@ -109,6 +109,10 @@ class DeploysController < ApplicationController
 
   protected
 
+  def deploy_permitted_params
+    [ :reference, :stage_id ] + Samson::Hooks.fire(:deploy_permitted_params)
+  end
+
   def reference
     deploy_params[:reference].strip
   end
@@ -118,7 +122,7 @@ class DeploysController < ApplicationController
   end
 
   def deploy_params
-    params.require(:deploy).permit(:reference, :stage_id)
+    params.require(:deploy).permit(deploy_permitted_params)
   end
 
   def find_project
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -5,8 +5,8 @@ class DeployService
     @user = user
   end
 
-  def deploy!(stage, reference)
-    deploy = stage.create_deploy(reference: reference, user: user)
+  def deploy!(stage, attributes)
+    deploy = stage.create_deploy(user, attributes)
     send_sse_deploy_update('new', deploy)
 
     if deploy.persisted? && (!stage.deploy_requires_approval? || release_approved?(deploy))
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -94,11 +94,8 @@ class Stage < ActiveRecord::Base
     confirm
   end
 
-  def create_deploy(options = {})
-    user = options.fetch(:user)
-    reference = options.fetch(:reference)
-
-    deploys.create(reference: reference) do |deploy|
+  def create_deploy(user, attributes = {})
+    deploys.create(attributes) do |deploy|
       deploy.build_job(project: project, user: user, command: command)
     end
   end
--- a/app/views/deploys/new.html.erb
+++ b/app/views/deploys/new.html.erb
@@ -31,6 +31,8 @@
         </div>
       </div>
 
+      <%= Samson::Hooks.render_views(:deploy_form, self, project: @project, form: form) %>
+
       <div class="form-group" id="new-deploy-buttons">
         <div class="col-lg-offset-2 col-lg-10">
           <%= form.submit "", class: "btn btn-primary", disabled: global_lock %>
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -8,6 +8,7 @@ module Samson
       :project_form,
       :deploys_header,
       :deploy_view,
+      :deploy_form, # for external plugin, so they can add extra form fields
       :admin_menu,
       :project_tabs_view
     ].freeze
@@ -15,6 +16,7 @@ module Samson
     EVENT_HOOKS = [
       :stage_clone,
       :stage_permitted_params,
+      :deploy_permitted_params, # for external plugin
       :project_permitted_params,
       :before_deploy,
       :after_deploy_setup,
--- a/test/models/deploy_service_test.rb
+++ b/test/models/deploy_service_test.rb
@@ -21,7 +21,7 @@ describe DeployService do
       SseRailsEngine.expects(:send_event).twice
       assert_difference "Job.count", +1 do
         assert_difference "Deploy.count", +1 do
-          service.deploy!(stage, reference)
+          service.deploy!(stage, reference: reference)
         end
       end
     end
@@ -37,7 +37,7 @@ describe DeployService do
 
       it "does not start the deploy" do
         service.expects(:confirm_deploy!).never
-        service.deploy!(stage, reference)
+        service.deploy!(stage, reference: reference)
       end
 
       describe "if release is approved" do
@@ -45,13 +45,13 @@ describe DeployService do
 
         it "starts the deploy, if in grace period" do
           service.expects(:confirm_deploy!).once
-          service.deploy!(stage_production_2, ref1)
+          service.deploy!(stage_production_2, reference: ref1)
         end
 
         it "does not start the deploy, if past grace period" do
           service.expects(:confirm_deploy!).never
           travel (BuddyCheck.period).hour + 1.minute do
-            service.deploy!(stage_production_2, ref1)
+            service.deploy!(stage_production_2, reference: ref1)
           end
         end
 
@@ -62,7 +62,7 @@ describe DeployService do
         it "does not start the deploy" do
           service.expects(:release_approved?).once
           service.expects(:confirm_deploy!).never
-          service.deploy!(stage, reference)
+          service.deploy!(stage, reference: reference)
         end
       end
 
@@ -73,7 +73,7 @@ describe DeployService do
         end
         it "it starts the deploy" do
           service.expects(:confirm_deploy!).once
-          service.deploy!(stage, reference)
+          service.deploy!(stage, reference: reference)
         end
       end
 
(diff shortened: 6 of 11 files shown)


kureikain:
@grosser 

I want to custom deploy form a bit and I tried to add some extra hooks. Not sure if this can be merged but I think it will make Samson more flexible when we add more useful hooks.

screen shot 2015-06-15 at 10 00 54 am

This PR allows me to do above form.

app/controllers/deploys_controller.rb
```diff
@@ -41,7 +43,11 @@ def new
 
   def create
     deploy_service = DeployService.new(current_user)
-    @deploy = deploy_service.deploy!(stage, reference)
+    if extra_attributes = deploy_params.except(:stage_id, :reference)
+      @deploy = deploy_service.deploy!(stage, reference, extra_attributes)
+    else
+      @deploy = deploy_service.deploy!(stage, reference)
+    end
```
grosser: this is always true

app/controllers/deploys_controller.rb
```diff
@@ -118,7 +124,8 @@ def stage
   end
 
   def deploy_params
-    params.require(:deploy).permit(:reference, :stage_id)
+    # params.require(:deploy).permit(:reference, :stage_id)
```
grosser: remove the comment ?

app/models/deploy_service.rb
```diff
@@ -5,11 +5,12 @@ def initialize(user)
     @user = user
   end
 
-  def deploy!(stage, reference)
+  def deploy!(stage, reference, attributes = nil)
```
grosser: empty hash would make more sense to me here as default

app/views/deploys/new.html.erb
```diff
@@ -18,6 +18,8 @@
         </div>
       <% end %>
 
+      <%= Samson::Hooks.render_views(:before_deploy_form, self, project: @project, form: form) %>
```
grosser: is 1 hook not good enough ?

test/models/deploy_service_test.rb
```diff
@@ -26,6 +26,22 @@
       end
     end
 
+    it "send after_deploy_create hook with parameter" do
```
grosser: `it sends`

app/controllers/deploys_controller.rb
```diff
@@ -1,4 +1,6 @@
 class DeploysController < ApplicationController
+  include DeployPermittedParams
```
grosser: only used in 1 place, why does it need to be a concern ?

grosser:
since this creates a deploy it might be simpler to just pass all extra attributes to the deploy and then make it a virtual attribute if you don't want to persist it

app/controllers/deploys_controller.rb
kureikain: Yes, my bad :(. I was just blindly follow StagePermittedParam :(

app/views/deploys/new.html.erb
kureikain: Actually one hook is good enough already but I think it will be nicer if we can make the field on top or below the revision field. It just looks nicer when the order is more natural imho. Should I remove `before` and keep `after` only?

app/models/deploy_service.rb
kureikain: totally agree

app/controllers/deploys_controller.rb
kureikain: my bad again.

app/views/deploys/new.html.erb
grosser: I'd think only after (what could be more important then the revision?) + rename to deploy_form

kureikain:
@grosser Thanks. I want to persist them so the team can look at old deploy later and see its description or any extra information attached to it. But I didn't know about that virtual attribute, I will do some homework so I can make use of them in future.

For now, do this have a change to get merge(of course after everything is fixed) because I'm maintaining my own fork now which is pain in the ass to keep up with samson upstream :(...

app/views/deploys/new.html.erb
grosser: Some comment in hook.rb like `# for external plugins` so nobody removes it in an effor to clean up unused hooks could be good too :)

grosser:
looks like reasonable additions to me + no crazy overhead, so yes :)

kureikain:
@grosser The build is passed :). I rebase to current master. You taught me some Ruby/Rails a long the way. Thanks.

app/controllers/deploys_controller.rb
```diff
@@ -109,6 +110,12 @@ def destroy
 
   protected
 
+  def deploy_permitted_params
+    @deploy_permitted_params ||=
+      ([ :reference , :stage_id
```
grosser: no need to cache since it is only used once ?
How about:

```Ruby
[ :reference , :stage_id] + Samson::Hooks.fire(:deploy_permitted_params)
```

grosser:
you can avoid this extra attributes passing around by adding a app/decorators/deploy_decorator.rb to do something like:
```
Deploy.class_eval do
  attr_accessor :my_extra_setting
end
```

and then use `deploy.my_extra_setting` in the callback

kureikain:
> you can avoid this extra attributes passing around by adding a app/decorators/deploy_decorator.rb to do something like:

oh, thanks for your teaching. I see that the way `deploy` object is crated right now is not directly creating a `deploy` model but creating via calling `deploy_service.deploy!` in https://github.com/zendesk/samson/blob/master/app/controllers/deploys_controller.rb#L44

```
    @deploy = deploy_service.deploy!(stage, reference)
```

So after creating it, I have to assign those attribute into `deploy` object so that I can use them in callback, right?

```
@deploy = deploy_service.deploy!(stage, reference, extra_attributes)
# becomes
@deploy = deploy_service.deploy!(stage, reference)
@deploy.assign_attributes extra_attributes
```

grosser:
Hmmm maybe instead of passing reference pass attributes ... this then goes into create_deploy

kureikain:
Ok, I pass all attributes to `deploy!` and using name parameter to make it  that we need at least stage, reference. Then I passed all of rest attributes into `create_deploy`. This way form fields can go into `deploy`

app/controllers/deploys_controller.rb
```diff
@@ -41,7 +41,8 @@ def new
 
   def create
     deploy_service = DeployService.new(current_user)
-    @deploy = deploy_service.deploy!(stage, reference)
+    attributes = deploy_params.except(:stage_id)
```
grosser: no harm in passing stage_id I guess ...

app/controllers/deploys_controller.rb
```diff
@@ -109,6 +110,11 @@ def destroy
 
   protected
 
+  def deploy_permitted_params
+    @deploy_permitted_params ||=
```
grosser: no need for the caching in `@deploy_permitted_params` ... it is not reused ... or is it ?

app/controllers/deploys_controller.rb
```diff
@@ -109,6 +110,11 @@ def destroy
...
+      [ :reference , :stage_id ] + Samson::Hooks.fire(:deploy_permitted_params)
```
grosser: please use `[:foo, :bar]` style for arrays :)

app/models/deploy_service.rb
```diff
@@ -5,11 +5,12 @@ def initialize(user)
     @user = user
   end
 
-  def deploy!(stage, reference)
-    deploy = stage.create_deploy(reference: reference, user: user)
+  def deploy!(stage, reference = nil, attributes = {})
```
grosser: that looks wrong ... I was expecting `deploy!(stage, attributes)` since both are mandatory

app/models/deploy_service.rb
```diff
@@ -5,11 +5,12 @@ def initialize(user)
...
+    deploy = stage.create_deploy(user: user, reference: reference, **attributes)
```
grosser: reference is already part of attributes

lib/samson/hooks.rb
```diff
@@ -18,6 +19,7 @@ class UserError < StandardError
       :project_permitted_params,
       :before_deploy,
       :after_deploy_setup,
+      :after_deploy_create, # for external plugin
```
grosser: can you use after_deploy_setup instead ?

kureikain: I cannot use `after_deploy_setup` because it seems has its own purpose and passing different data. It is fired after the job actually execute, after setting up necessary stuff.
```
Samson::Hooks.fire(:after_deploy_setup, dir, stage) 
```
So If I have to store extra fields somewhere around the time `deploy` is saved . maybe `after_deploy_save` a better name?

grosser: you could make after_deploy_setup also pass the deploy 

after_deploy_setup and after_deploy_save would be called at almost the same time, that's why I was thinking we don't need another hook ... store the attributes in the deploy and then use after_deploy_setup and read it back from the deploy

kureikain: My stupid. You are right. Much simpler this way. Let me redo this mess :)

kureikain:
@grosser  I fixed the code per review and rebase to current master :). See if we can do a Monday merge :cake:

app/models/deploy_service.rb
```diff
@@ -5,8 +5,8 @@ def initialize(user)
     @user = user
   end
 
-  def deploy!(stage, reference)
-    deploy = stage.create_deploy(reference: reference, user: user)
+  def deploy!(stage, attributes = {})
```
grosser: attributes does not need a default since references needs to be passed

app/models/stage.rb
```diff
@@ -96,9 +96,8 @@ def confirm_before_deploying?
 
   def create_deploy(options = {})
```
grosser: this should be attributes, not options :)

app/models/deploy_service.rb
```diff
@@ -5,8 +5,8 @@ def initialize(user)
...
+    deploy = stage.create_deploy(user: user, **attributes)
```
grosser: maybe make the interface be `user, attributes` ... then we don't need to merge or slice

kureikain: Agree, this makes the code much more simple to understand. Made the change.

kureikain:
@grosser Code updated :joy:. Made `create_deploy` signature to `(user, attributes)` and no longer need the ugly `**` or `fetch(:user)`

app/models/stage.rb
```diff
@@ -94,11 +94,8 @@ def confirm_before_deploying?
     confirm
   end
 
-  def create_deploy(options = {})
-    user = options.fetch(:user)
-    reference = options.fetch(:reference)
-
-    deploys.create(reference: reference) do |deploy|
+  def create_deploy(user, attributes = {})
+    deploys.create(attributes) do |deploy|
```
grosser: sweet, even simpler than the old code :D

grosser:
:+1: 

@zendesk/runway second sets of eyes plz in case I missed something ....

sandlerr:
:+1:

sandlerr:
Thanks @kureikain!

jwswj:
Really cool. Thanks @kureikain

PR #474, opened 2015-06-19
force_ssl in production

kintner:
@jonmoter can we try this again? It looks like the nginx changes have been cooked out.


jonmoter:
Okay, let's give it a shot in staging.  We now know what to revert if it doesn't work.

:+1:

jonmoter:
Actually, could you make this an ENV var instead of hard-coding it to true?

For people outside of Zendesk who use Samson, it could break their setup if suddenly it required SSL.

PR #478, opened 2015-06-19
add project 'deploy with docker' checkbox

gabetax:
Adds an explicit flag for whether we should apply any docker-specific behavior to this project, e.g. building images or presenting UI for host assignment.

We could alternately infer this from the presence of a `Dockerfile` in the repo, but until we're further down the road I'd rather not surprise anyone with behavior.

/cc @zendesk/runway


gabetax:
This is also featured flagged so we're not confusing the general public.

grosser:
Projects will still need old + new deployment while we figure out production issues, so if this is just for showing additional UI then :+1:

PR #479, opened 2015-06-19
when changing or deleting a deploy group projects#show stage stayed u…

grosser:
…nchanged

@zendesk/runway 

### Risks
 - None

### DIFF
--- a/test/models/deploy_group_test.rb
+++ b/test/models/deploy_group_test.rb
@@ -1,18 +1,30 @@
 require_relative '../test_helper'
 
 describe DeployGroup do
-  let(:prod_env) { environments(:production) }
+  let(:stage) { stages(:test_staging) }
+  let(:environment) { environments(:production) }
+  let(:deploy_group) { deploy_groups(:pod1) }
+
+  def self.it_expires_stage(method)
+    it "expires stages when #{method}" do
+      stage.deploy_groups << deploy_group
+      stage.update_column(:updated_at, 1.minute.ago)
+      old = stage.updated_at.to_s(:db)
+      deploy_group.send(method)
+      stage.reload.updated_at.to_s(:db).wont_equal old
+    end
+  end
 
   describe '.new' do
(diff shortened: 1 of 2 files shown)


test/models/deploy_group_test.rb
```diff
@@ -1,18 +1,30 @@
...
 describe DeployGroup do
-  let(:prod_env) { environments(:production) }
+  let(:stage) { stages(:test_staging) }
+  let(:environment) { environments(:production) }
+  let(:deploy_group) { deploy_groups(:pod1) }
+
+  def self.it_expires_stage(method)
+    it "expires stages when #{method}" do
+      stage.deploy_groups << deploy_group
+      stage.update_column(:updated_at, 1.minute.ago)
```
sandlerr: can this be set in the `let` block that creates the fixture?

sandlerr:
:+1:

test/models/deploy_group_test.rb
grosser: lets are cached per test case not test class, so there would be no benefit in using it.
A faster option would be to use timecop or make the fixture itself old, but that would add gem/mental overhead so ... meh ...

grosser:
Ill take that +1 and run with it, let me know if you have an idea to make it better :)

test/models/deploy_group_test.rb
sandlerr: My suggestion was to make an old fixture - not sure what we can do here but in factorygirl you could go `let(:stage) { create(:stage, updated_at: 1.minute.ago) }` or similar

grosser: create is even more expensive then an update_column :)

sandlerr: is there not an implicit create going on here? If this all in memory then great!

grosser: it's using a fixture -> no create in each test

PR #481, opened 2015-06-23
faster docker build

sandlerr:
@zendesk/runway @zendesk/samson

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ENV GEM_HOME=/bundle
 RUN gem update --system
 RUN gem install bundler
 
-RUN mkdir /app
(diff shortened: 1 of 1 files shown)


sandlerr:
We should be careful with `mkdir` in Dockerfiles, it seems to bust the cache unless we change the mtime.

grosser:
:+1: yeah ... that bug is annoying :/

grosser:
will be fixed in 1.8 https://github.com/docker/docker/pull/12031

sandlerr:
glad they changed that. Let's revert this when 1.8 is out.

princemaple:
OMG I'm so looking forward to 1.8 now.

PR #482, opened 2015-06-23
remove unneeded statement

sandlerr:
From the docs for [ADD](https://docs.docker.com/reference/builder/):

> If \<dest\> doesn’t exist, it is created along with all missing directories in its path.

This builds and caches fine.

@zendesk/runway @zendesk/samson @grosser @princemaple


princemaple:
:+1:

PR #484, opened 2015-06-24
Dockerfile remove GEM_HOME

gabetax:
By creating /bundle and then installing a gem into it, it prevents zdi from mounting a docker volume to /bundle in development mode, which causes a repeated bundling that gets thrown away between runs. zendesk/docker-images#543 ensures that we have bundler installed even if we change with GEM_HOME, so this line should no longer be necessary.

/cc @zendesk/runway

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@ RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add
 RUN echo 'deb https://deb.nodesource.com/node_0.12 jessie main' > /etc/apt/sources.list.d/nodesource.list
 RUN apt-get update && apt-get install -y nodejs
 
-ENV GEM_HOME=/bundle
 RUN gem update --system
 RUN gem install bundler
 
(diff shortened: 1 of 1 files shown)


grosser:
:+1:

Dockerfile
```diff
@@ -5,7 +5,6 @@ RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add
 RUN echo 'deb https://deb.nodesource.com/node_0.12 jessie main' > /etc/apt/sources.list.d/nodesource.list
 RUN apt-get update && apt-get install -y nodejs
 
-ENV GEM_HOME=/bundle
 RUN gem update --system
 RUN gem install bundler
```
grosser: that bundle + gem install should also no longer be necessary ?

gabetax: I was thinking it could be handy to get the latest gem and bundler for production images on build, but it looks like the 2.2.2 tag on Dockerhub is [regularly updated](https://github.com/docker-library/ruby/tree/master/2.2) with the latest.

grosser:
go forth and merge :)
also a new build / release would be great :)

gabetax:
built and released.

grosser:
:heart:

PR #486, opened 2015-06-25
Run different commands in a stage

kureikain:
Try to draft PR.

### DIFF
--- a/app/controllers/admin/commands_controller.rb
+++ b/app/controllers/admin/commands_controller.rb
@@ -51,7 +51,7 @@ class Admin::CommandsController < ApplicationController
   private
 
   def command_params
-    params.require(:command).permit(:command, :project_id)
+    params.require(:command).permit(:name, :command, :project_id)
   end
 
   def successful_response(notice)
--- a/app/models/command.rb
+++ b/app/models/command.rb
@@ -17,8 +17,12 @@ class Command < ActiveRecord::Base
     project_id.nil?
   end
 
-  def name_alias
-    name || command
+  def name
+    lines = command.split("\n")
+    if /^#.+/ =~ lines.first
+      lines.first.slice(1, lines.first.length)
+    else
+      command
+    end
   end
-
 end
--- a/app/models/command.rb
+++ b/app/models/command.rb
@@ -16,4 +16,9 @@ class Command < ActiveRecord::Base
   def global?
     project_id.nil?
   end
+
+  def name_alias
+    name || command
+  end
+
 end
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -166,4 +166,5 @@ class Deploy < ActiveRecord::Base
       "(with #{buddy.name})"
     end
   end
+
 end
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -166,5 +166,4 @@ class Deploy < ActiveRecord::Base
       "(with #{buddy.name})"
     end
   end
-
 end
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -95,8 +95,13 @@ class Stage < ActiveRecord::Base
   end
 
   def create_deploy(user, attributes = {})
-    deploys.create(attributes) do |deploy|
-      deploy.build_job(project: project, user: user, command: command)
+    attributes = attributes.symbolize_keys
+    deploy_command = command
+    if attributes[:command_id]
+      deploy_command = Command.find(attributes[:command_id]).try(:command) rescue command
+    end
+    deploys.create(attributes.except(:command_id)) do |deploy|
+      deploy.build_job(project: project, user: user, command: deploy_command)
     end
   end
 
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -96,12 +96,9 @@ class Stage < ActiveRecord::Base
 
   def create_deploy(user, attributes = {})
     attributes = attributes.symbolize_keys
-    deploy_command = command
-    if attributes[:command_id]
-      deploy_command = Command.find(attributes[:command_id]).try(:command) rescue command
-    end
+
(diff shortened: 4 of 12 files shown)


kureikain:
@grosser Very basic prototype of what we discussed about in other issue to run different tasks on same stage.

This GIF shows how it run.
untitled

- User click deploy and it runs normally, use commands which associated with the stage
- If user select the command to run, then that command will be run instead of the default commands. The list of commands to select include global command and project-specific command. 
- If command is named, the name appear in dropdown list, otheriwser, the command content .

The way it works it:

Currently, we create `deploy` by calling `deploy_service.create_deploy` which in turns call `stage.create_deploy`. 

Stage model create a `deploy` object, then build a associated `job` object and pass `command` to run when building it. That command is fetched inside stage. But now, we will manually pass a `command_id` (the command which user select when deploy) and build the command outselves instead of fetching default command of stages.

app/models/stage.rb
```diff
@@ -95,8 +95,13 @@ def confirm_before_deploying?
   end
 
   def create_deploy(user, attributes = {})
-    deploys.create(attributes) do |deploy|
-      deploy.build_job(project: project, user: user, command: command)
+    attributes = attributes.symbolize_keys
+    deploy_command = command
+    if attributes[:command_id]
+      deploy_command = Command.find(attributes[:command_id]).try(:command) rescue command
```
kureikain: If we passed in `command_id`, we will get that command to build job

app/views/admin/commands/_form.html.erb
```diff
@@ -5,6 +5,7 @@
     <div class="form-group">
       <%= form.label :command, class: "col-lg-2 control-label" %>
       <div class="col-lg-6">
+        <%= form.text_field :name, class: "form-control", rows: 10 %>
```
kureikain: add name for command

app/views/deploys/new.html.erb
```diff
@@ -35,6 +36,21 @@
 
       <div class="form-group" id="new-deploy-buttons">
         <div class="col-lg-offset-2 col-lg-10">
+
```
kureikain: Very ugly JS. Will need to remove and clean up/ inline JS later, re-style form

test/models/stage_test.rb
```diff
@@ -141,6 +141,12 @@
         subject.create_deploy(user, {reference: ""})
       end
     end
+
+    it "creates a new deploy with custom command" do
```
kureikain: Test that if we passed a custom command id, then the job has to contain that command.

app/controllers/admin/commands_controller.rb
```diff
@@ -51,7 +51,7 @@ def destroy
   private
 
   def command_params
-    params.require(:command).permit(:command, :project_id)
+    params.require(:command).permit(:name, :command, :project_id)
```
grosser: how about this: if the first line is a comment -> use that as name
so inline editing still works nicely + compiled command makes sense

app/models/command.rb
```diff
@@ -16,4 +16,9 @@ def self.for_project(project)
   def global?
     project_id.nil?
   end
+
+  def name_alias
+    name || command
+  end
+
```
grosser: extra newline at the end

app/models/deploy.rb
```diff
@@ -166,4 +166,5 @@ def deploy_buddy
       "(with #{buddy.name})"
     end
   end
+
```
grosser: plz kill that newline :)

grosser:
What I'd like to use is: all stage commands + replace the last one :)

grosser:
because usually it's cap xxx deploy and then you'd replace this with cap xxx migrate

app/controllers/admin/commands_controller.rb
kureikain: Good idea. Removing the overhead of new field too.

kureikain:
> What I'd like to use is: all stage commands + replace the last one :)
> because usually it's cap xxx deploy and then you'd replace this with cap xxx migrate

Yup, let me redo and fix code per review as well.

grosser:
idk .. the more I think about this the more it sounds like a bad idea... just make 3 stages and reuse the commands ... gives much more flexibility especially with upcoming deploy pipeline stuff and auditing is also easier ...

kureikain:
Haha, Ok.
Let me back to stage idea then. It's what I used to setup too. Could you let me know more about deploy pipeline or any upcoming idea? I have no idea what is next on Samson and at this point I alreayd married to it :) so...

grosser:
this is supposed to be coming up ... not sure why it has stalled ... maybe
poke him :D
https://github.com/zendesk/samson/pull/425

otherwise there is no real roadmap, we are working most on new docker
foobar ... and on whatever comes up / blocks a project ...

kureikain:
@grosser I looked that plugin up and I cannot figured out a way to setup my flow. It seems complicated to my tasks.

I really want to have a way to easy select and run separate commands in a stage. So I think I will try to turn this thing into a plugin, I may have to add extra hooks a long the way :) and hope you guys will accept PRs.

The reason we are doing this is because we came from webistrano, where we have an UI to select tasks. Eg: we have worker, search server, app, load balancer and by having many tasks in a stage we can do thing say: restart only worker, restart search daemon on search server. And most of the time those tasks are a single `cap` command. This is where I get confuse with pipeline stuff to set up my flow and I think pipeline is big for my purpose.

kureikain:
@grosser Good news. I managed to do it without extra hooks.

- First, I using `deploy_form` hook to render form
- Second, I set a virtual attribute `command_id` on deploy. 
- Next, I used `after_save` callback of `deploy` module via creating an decorator to attempt reading `command_id` and modify command of associated job.

The end result was much clear than putting them into samson core. 
Thanks to excellent hook system :)

grosser:
It's powerful ... not sure if it's excellent :D

PR #487, opened 2015-06-25
show shared commands, prevent edits of shared commands and show unuse…

grosser:
…d commands

@zendesk/runway 

 - lock -> shared -> think about editing + big warning in your face when you edit
 - no lock + unchecked -> unused -> can be deleted
 - no lock + checked -> 1-off command ... might be something fishy .. should use shared command

screen shot 2015-06-25 at 2 18 40 pm

screen shot 2015-06-25 at 2 18 46 pm

### Risks
 - None

### DIFF
--- a/app/views/stages/_fields.html.erb
+++ b/app/views/stages/_fields.html.erb
@@ -116,14 +116,19 @@
 
   <div id="commands">
     <%= form.collection_check_boxes(:command_ids, form.object.all_commands, :id, :command) do |b| %>
+      <% command = b.object %>
+      <% others = (command.stages - [@stage]).map(&:name) %>
+      <% others_warning = "#{others.size} other stages: #{others.join(", ")}" if others.any? %>
       <div class="row stage-bar bar">
         <div data-id="<%= b.value %>" class="col-lg-offset-2 col-lg-2 command-checkbox">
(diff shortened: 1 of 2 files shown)


app/views/stages/_fields.html.erb
```diff
@@ -116,14 +116,19 @@
 
   <div id="commands">
     <%= form.collection_check_boxes(:command_ids, form.object.all_commands, :id, :command) do |b| %>
+      <% command = b.object %>
+      <% others = (command.stages - [@stage]).map(&:name) %>
```
sandlerr: can `self.stages - [self]` be a method on the model?

grosser: I don't think this makes it any simpler ... it's and array an we do some simple array math :)

```Ruby
def others(stage)
  stages - [stage]
end
```

`command.others(stage)` vs `command.stages - [@stage]`

sandlerr: fair enough. I don't think I've ever seen so much logic in a view before but OTOH its just four local variable definitions.

sandlerr:
:+1:

grosser:
heheh ... I'm going to delete sooo many commands now muahahahah

PR #489, opened 2015-06-25
use form.object instead of @stage since project new renders stage form

grosser:
@zendesk/runway 

### Risks
 - None


gabetax:
:+1:

PR #490, opened 2015-06-26
longer docker timeout for slower ops like compiling native node modules

sandlerr:
@zendesk/runway @jonmoter 
I had this timeout several times on my laptop. The default is 60 sec iirc. Sometimes there is no output for quite a while. 10 minutes is probably a bit too long - what should we use?


sandlerr:
Maybe it should be an environment variable?

grosser:
:+1: maybe leave at 10 and see if anything hangs/complains ...

PR #491, opened 2015-06-26
allow plugins to define fixtures

sandlerr:
@zendesk/runway 
Had to do this a hacky symlink way - `ActiveSuport::TestCase` doesn't really make this easy. Overriding method missing and changing `ActiveSupport::TestCase.fixture_path` would be an option but I think it would get messy since fixtures in plugins will probably have a relation on core model fixtures like `Project`.

### DIFF
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -31,7 +31,7 @@ module Samson
     @@hooks = {}
 
     class Plugin
-      attr_reader :name
+      attr_reader :name, :folder
       def initialize(path)
         @path = path
         @folder = File.expand_path('../../../', @path)
@@ -135,6 +135,19 @@ module Samson
           each(&:add_migrations).
           each(&:add_assets_to_precompile).
           each(&:add_decorators)
+      end
+
+      def plugin_test_setup
+        fixture_path = ActiveSupport::TestCase.fixture_path
+        plugins.each do |plugin|
+          fixtures = Dir.glob(File.join(plugin.folder, 'test', 'fixtures', '*.yml'))
+          fixtures.each do |fixture|
+            yml_filename = fixture[/\w+\.yml\z/]
+            new_path = File.join(fixture_path, yml_filename)
+            File.symlink(fixture, new_path)
+            Minitest.after_run { File.delete(new_path) }
+          end
+        end
       end
 
       def render_javascripts(view)
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -56,6 +56,7 @@ class ActiveSupport::TestCase
   #
   # Note: You'll currently still have to declare fixtures explicitly in integration tests
   # -- they do not yet inherit this setting
+  Samson::Hooks.plugin_test_setup
   fixtures :all
 
   before do


grosser:
how about doing
```Ruby
fixtures Dir["{#{fixture_path}/,plugins/*/test/fixtures}{**,*}/*.{yml}"].map! { |f| f[(fixture_path.to_s.size + 1)..-5] }
```

grosser:
not sure if that full path thing is a thing ... but could do a `set_fixtures File.basename(f), file: f` or some such

sandlerr:
the public fixture methods are set up to take class / fixture names and then search the `fixture_path`. The `fixtures` method checks the existing filenames and define methods based on them but when you actually try to use those fixtures it needs the files to be in `fixture_path` again... so we need a `fixture_path` string that globs all the fixtures. It would be something like `**/fixtures` to include plugin and non-plugin.

grosser:
can we set it to "{,plugins/xyz/}fixtures" ?

sandlerr:
I tried

```
ActiveSupport::TestCase.fixture_path = '.'
fixtures Dir['{,plugins/*/}test/fixtures/*'].map { |f| f.sub(/\.yml\z/, '') }
```

unfortunately rails is inferring the table name (class name?) from the path...
`ActiveRecord::StatementInvalid: Mysql2::Error: Table 'samson_development.test_fixtures_builds' doesn't exist: DELETE FROM ``test_fixtures_builds`` `

sandlerr:
Also, nothing in this thread handles the case where a plugin is actually a gem...

sandlerr:
They weren't anticipating this being a glob directory :-) The code includes this line
```
(byebug) fixture_set_names.map! { |f| f[(fixture_path.to_s.size + 1)..-5] }
[nil, nil, "", nil, "", nil, "s", nil, "lications", nil, nil, "s", nil, nil, "ures/meatballs_releases"]
```

I'll submit a patch but in the meantime I think we need to go with symlinks.

grosser:
tried 
```
  fixtures '../../plugins/env/test/fixtures/stages'
  set_fixture_class '../../plugins/env/test/fixtures/stages' => Stage

```

and 

```
old = fixture_path
  [old, "plugins/env/test/fixtures"].each do |path|
    self.fixture_path = path
    fixtures :all
  end
  self.fixture_path = old
```

but no success :/

grosser:
how about .... plugins don't get to have fixtures ?
they should be small -> can just use create to get their models ?

sandlerr:
https://github.com/rails/rails/pull/20728

sandlerr:
@grosser don't want just the first commit from this branch?

grosser:
that would mean always having a dirty tree after running tests, right ?
.... is it good enough to just use create etc in plugin tests ?

sandlerr:
Not necessarily, the symlinks are deleted by `Minitest.after_run { File.delete(new_path) }`. We will probably have a dirty tree anyway because of migrations though.

grosser:
hmm yeah I it would be pretty clean ... still not sure if it's worth the complexity+overhead / if plugins need fixtures ... :+1: for the code though ... looks like there is no simpler way of doing it

PR #493, opened 2015-06-26
split .travis.yml for faster builds

grosser:
no longer doing the bundle-without dance since the whole bundle should be cached anyway ...
also the task is now easier to execute locally

`DB=mysql TASK=rake ./script/ci.rb`

### Risks
 - None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,5 @@ env:
     - DB=mysql TASK=js
     - DB=mysql TASK=precompile
     - DB=postgresql TASK=rake
-    - DB=postgresql TASK=js
-    - DB=postgresql TASK=precompile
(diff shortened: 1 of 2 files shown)


grosser:
this 'only' takes us down from ~5min to ~4min, but the big advantage is that rake (the most brittle part) is done instantly -> quick feedback / no need to wait for precompile to finish

screen shot 2015-06-26 at 9 44 49 am

grosser:
we could even stop doing the js/precompile builds on postgres/sqlite since they don't even use the database .... thoughts ?

sandlerr:
yes, let's only do one js and one precompile
:+1: faster tests! :heart:

PR #494, opened 2015-06-29
fix pushes

sandlerr:
@zendesk/runway this breaks in conjunction with https://github.com/zendesk/docker-images/pull/541 - a docker repo can't seem to contain more than one slash.
Companion fix: https://github.com/zendesk/docker-images/pull/550

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -37,7 +37,16 @@ class Project < ActiveRecord::Base
   end
 
   def docker_repo
-    "#{Rails.application.config.samson.docker.registry}/samson/#{repo_name}"
+    @docker_repo ||= begin
+      registry = Rails.application.config.samson.docker.registry
+      if Rails.env.production?
+        registry
+      else
+        host, namespace = registry.split '/'
+        namespace ||= 'samson'
+        "#{host}/#{namespace}_non_prod"
+      end
+    end
   end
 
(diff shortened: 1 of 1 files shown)


app/models/project.rb
```diff
@@ -37,7 +37,16 @@ def repo_name
...
 
   def docker_repo
-    "#{Rails.application.config.samson.docker.registry}/samson/#{repo_name}"
+    @docker_repo ||= begin
+      registry = Rails.application.config.samson.docker.registry
+      if Rails.env.production?
+        registry
+      else
+        host, namespace = registry.split '/'
+        namespace ||= 'samson'
```
grosser: maybe ignore the namespace and just use samson_dev ?

sandlerr: this is the least amount of work that allows to not overwrite each other's dev pushes. Do you think there is value in that? The proper solution would be to run a local registry.

grosser:
:+1: local registry would be a lot of overhead ...

jonmoter:
Grr. The [Docker Registry v2 spec](https://github.com/docker/distribution/blob/master/docs/spec/api.md#overview) says that a repository name is broken up into "path" components, separated by slashes. It says it requires at least 2 path components, but can have more.

But I guess something in the Docker engine code doesn't allow for that yet.  Sigh.

PR #495, opened 2015-06-29
Project#current_user accessor for permissions in plugin hooks

gabetax:
I have a plugin that needs to enforce permissions based on the user's role, e.g. only all super users to manage docker assignments in production. Normally I'd use a service object for this, but we don't have that luxury with plugins.

/cc @zendesk/runway

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -19,6 +19,9 @@ class Project < ActiveRecord::Base
   has_many :commands
   has_many :macros
 
+  # For permission checks on callbacks. Currently used in private plugins.
+  attr_accessor :current_user
+
   accepts_nested_attributes_for :stages
 
   scope :alphabetical, -> { order('name') }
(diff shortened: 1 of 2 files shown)


grosser:
global `User.current` maybe ?

app/models/project.rb
```diff
@@ -19,6 +19,9 @@ class Project < ActiveRecord::Base
   has_many :commands
   has_many :macros
 
+  # For permission checks on callbacks
```
grosser: + 'use in a private plugin' so nobody removes it

grosser:
:+1: a general solution might be nice, but this is pretty simple/clean ...

gabetax:
I'm hesitant about (thread-local) globals. We'd have to set it in a Rack middleware, right? Let's hold off until at least one other thing needs this functionality.

grosser:
yep, agreed

PR #497, opened 2015-06-29
fix working and logic of commands index

grosser:
@zendesk/runway 

there are no 'others' the count is the total usage count ...

screen shot 2015-06-29 at 4 34 18 pm

### Risks
 - None


sandlerr:
:+1:

PR #499, opened 2015-07-01
fix off by one error

sandlerr:
@grosser `b.value` is just the ID. I don't think it exposes a `checked?` method but this works
cc @zendesk/runway


grosser:
thx, I saw these weird errors and was wondering why they happened ...
the name `checked` does not make much sense then ... maybe just inline the 2 different usages

grosser:
:+1:

PR #500, opened 2015-07-01
env variable preview

grosser:
in our private plugin we fail when unable to resolve some values, but would be nice to be able to show a helpful message so users can see if it only affects a certain group they might not care for / gives them a complete picture of the issue instead of failing at the first one

@zendesk/runway 

### Risks
 - None

### DIFF
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -5,7 +5,9 @@ class EnvironmentVariable < ActiveRecord::Base
   validates :scope_type, inclusion: ["Environment", "DeployGroup", nil]
 
   class << self
-    def env(project, deploy_group)
+    # preview parameter can be used to not raise an error,
+    # but return a value with a helpful message
+    def env(project, deploy_group, preview: false)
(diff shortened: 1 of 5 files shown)


grosser:
@zendesk/runway

grosser:
@zendesk/runway

plugins/env/app/models/environment_variable.rb
```diff
@@ -5,7 +5,9 @@ class EnvironmentVariable < ActiveRecord::Base
   validates :scope_type, inclusion: ["Environment", "DeployGroup", nil]
 
   class << self
-    def env(project, deploy_group)
+    # preview parameter can be used to not raise an error,
+    # but return a value with a helpful message
+    def env(project, deploy_group, preview: false)
```
sandlerr: I'm confused, where does this method use `preview`?

grosser: it does not, I'm going to use it in the consul reader to either blow up or show a warning

sandlerr:
:+1:

PR #501, opened 2015-07-02
extract a helper

sandlerr:
@zendesk/runway @grosser for use in private plugin

### DIFF
--- a/plugins/env/app/decorators/projects_controller_decorator.rb
+++ b/plugins/env/app/decorators/projects_controller_decorator.rb
@@ -1,0 +1,3 @@
+ProjectsController.class_eval do
+  helper EnvironmentVariablesHelper
+end
--- a/plugins/env/app/decorators/projects_controller_decorator.rb
+++ b/plugins/env/app/decorators/projects_controller_decorator.rb
@@ -1,3 +1,0 @@
-ProjectsController.class_eval do
-  helper EnvironmentVariablesHelper
-end
--- a/plugins/env/app/helpers/environment_variables_helper.rb
+++ b/plugins/env/app/helpers/environment_variables_helper.rb
@@ -1,0 +1,9 @@
+module EnvironmentVariablesHelper
+  def env_deploygroup_array(include_all: true)
+    all = include_all ? [["All", nil]] : []
+    envs = Environment.all.map { |env| [env.name, "Environment-#{env.id}"] }
+    separator = [["----", nil]]
+    deploy_groups = DeployGroup.all.map { |dg| [dg.name, "DeployGroup-#{dg.id}"] }
+    all + envs + separator + deploy_groups
+  end
+end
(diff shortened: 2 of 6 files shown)


plugins/env/app/decorators/projects_controller_decorator.rb
```diff
@@ -0,0 +1,3 @@
+ProjectsController.class_eval do
+  helper EnvironmentVariablesHelper
```
grosser: don't we have `helpers :all` ?

plugins/env/app/helpers/environment_variables_helper.rb
```diff
@@ -0,0 +1,9 @@
+module EnvironmentVariablesHelper
```
grosser: maybe static method on SamsonEnv

plugins/env/app/decorators/projects_controller_decorator.rb
sandlerr: not that I could find. I think its better not to add that.

plugins/env/app/helpers/environment_variables_helper.rb
sandlerr: Will do. Wasn't sure if its more view or model logic... can we keep the functionality to decorate controllers still?

plugins/env/app/decorators/projects_controller_decorator.rb
grosser: Yeah ... I prefer explicit helpers too :)

plugins/env/app/helpers/environment_variables_helper.rb
grosser: Sounds like a useful improvement :)

sandlerr:
@grosser updated

plugins/env/app/views/samson_env/_environment_variables.html.erb
```diff
@@ -1,4 +1,4 @@
-<% scopes = [["All", nil]] + Environment.all.map{ |dg| [dg.name, "Environment-#{dg.id}"] } + [["----", nil]] + DeployGroup.all.map{ |dg| [dg.name, "DeployGroup-#{dg.id}"] } %>
+<% scopes = EnvironmentVariable.env_deploygroup_array %>
```
grosser: guess this could be `[["All", nil]]` + xxx to avoid the conditional ... but either way is fine

plugins/env/app/models/environment_variable.rb
```diff
@@ -15,6 +15,14 @@ def env(project, deploy_group)
       resolve_dollar_variables(env)
     end
 
+    def env_deploygroup_array(include_all: true)
```
grosser: maybe drop a `# used by private plugin` to be safe

grosser:
:+1:

PR #502, opened 2015-07-02
rename envar, set default value

sandlerr:
@zendesk/runway

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,7 +23,7 @@ GITHUB_TOKEN={fill me in}
 # BYPASS_DETAILS={optional 'Some text explaining bypass procedure'}
 # BUDDY_CHECK_FEATURE={optional, set to 1 to enable buddy_check feature; 0 to disable}
 
-# DEPLOY_MAX_MINUTES_PENDING={optional, set to 20 for max minutes a deploy is pending}
+# BUDDY_CHECK_TIME_LIMIT={optional, set to 20 for max minutes a deploy is pending}
 
 # PROJECT_CREATED_NOTIFY_ADDRESS=bobby-the-security-auditor@yourcompany.com
 
--- a/lib/policies/buddy_check.rb
+++ b/lib/policies/buddy_check.rb
@@ -18,7 +18,7 @@ module BuddyCheck
     ENV["BYPASS_JIRA_EMAIL"]
   end
 
-  def deploy_max_minutes_pending
+  def time_limit
     ENV.fetch("BUDDY_CHECK_TIME_LIMIT", ENV.fetch("DEPLOY_MAX_MINUTES_PENDING", 20)).to_i
   end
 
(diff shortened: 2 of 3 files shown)


sandlerr:
still reading the old envar - will deploy this, change the variable we pass in and then make another PR to remove the fallback.

.env.example
```diff
@@ -23,7 +23,7 @@ GITHUB_TOKEN={fill me in}
 # BYPASS_DETAILS={optional 'Some text explaining bypass procedure'}
 # BUDDY_CHECK_FEATURE={optional, set to 1 to enable buddy_check feature; 0 to disable}
 
-# DEPLOY_MAX_MINUTES_PENDING={optional, set to 20 for max minutes a deploy is pending}
+# BUDDY_CHECK_TIME_LIMIT={optional, set to 20 for max minutes a deploy is pending}
```
grosser: I was thinking it's a 'timeout', after a certain time it does not work anymore.
`BUDDY_CHECK_TIMEOUT={optional, default 20, how many minutes a commit can be redeployed without needing another buddy check}`

grosser: maybe change to 30 or 15 ... 20 seems arbitrary ...

lib/policies/buddy_check.rb
```diff
@@ -18,7 +18,7 @@ def bypass_jira_email_address
   end
 
   def deploy_max_minutes_pending
-    ENV["DEPLOY_MAX_MINUTES_PENDING"].to_i
+    ENV.fetch("BUDDY_CHECK_TIME_LIMIT", ENV.fetch("DEPLOY_MAX_MINUTES_PENDING", 20)).to_i
```
grosser: Like the fallback :)

grosser: maybe rename the method too

.env.example
sandlerr: so there are two different scenarios:
1. I deploy production, but within BUDDY_CHECK_TIMEOUT nobody approves my deploy. It gets cancelled.
2. I deploy production:datacenter1 with a buddy and some time later I try to deploy the same tag to production:datacenter2. If BUDDY_CHECK_TIMEOUT2 has elapsed then I need another approval.

Do we need separate variables here?

grosser: ahhh the other value is BUDDY_CHECK_GRACE_PERIOD ... so no need for any
changes :D

grosser: https://github.com/zendesk/samson/pull/503

ashmckenzie:
Heyas, am chasing up https://zendesk.atlassian.net/browse/OP-12629...  is this PR OK to go out?

sandlerr:
Hey Ash, I'd like to go ahead with this
cc @zendesk/runway

ashmckenzie:
LGTM :+1:

lib/policies/buddy_check.rb
sandlerr: done

PR #504, opened 2015-07-03
restore sortable behaviour to parent elements

sandlerr:
@zendesk/runway


grosser:
:+1: thx for fixing :)

PR #506, opened 2015-07-03
update gems

sandlerr:
@zendesk/runway


grosser:
:+1: here comes the bug train choo-choo :D

sandlerr:
First stop: staging

PR #509, opened 2015-07-07
Shender/fix project and stage caching

henders:
Cache wasn't taking into account changes to deploy group memberships for stages. So made it part of the cache_key.

project0

==================================================

editing_pods__project0_

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -16,10 +16,8 @@ class Stage < ActiveRecord::Base
     -> { order('stage_commands.position ASC') },
     through: :stage_commands, auto_include: false
 
-  # Force a dirty stage even if only the deploy_groups have changed.
-  has_and_belongs_to_many :deploy_groups,
-                          after_add: ->(stage, _) { stage.updated_at = Time.now },
-                          after_remove: ->(stage, _) { stage.updated_at = Time.now }
+  has_many :deploy_groups_stages
+  has_many :deploy_groups, through: :deploy_groups_stages
 
   default_scope { order(:order) }
 
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -16,7 +16,10 @@ class Stage < ActiveRecord::Base
     -> { order('stage_commands.position ASC') },
     through: :stage_commands, auto_include: false
 
-  has_and_belongs_to_many :deploy_groups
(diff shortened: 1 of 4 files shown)


app/models/stage.rb
```diff
@@ -191,6 +191,10 @@ def datadog_monitors
     datadog_monitor_ids.to_s.split(/, ?/).map { |id| DatadogMonitor.new(id) }
   end
 
+  def cache_key
+    "#{super}-#{self.deploy_groups.map(&:cache_key)}"
```
grosser: no need for self

grosser:
:+1: a little sad but I guess that's what it takes ...

sandlerr:
Alternatively could we make the update touch the stage, and the stage touch the project?

grosser:
that might be more efficient since we have very few updates

henders:
Phew, was hoping someone would have a better idea, thanks @sandlerr, I'll give that a go.

henders:
@grosser @sandlerr 
Ready for another review!

grosser:
How about having deploy_group_stage.rb to have `belongs_to :stage, touch: true` ?

grosser:
this won't work if you for example update a deploy-group, right ?

henders:
@grosser 
> this won't work if you for example update a deploy-group, right ?

Yea, I figured I'd leave that to another PR if you change the deploygroup name.

> How about having deploy_group_stage.rb to have belongs_to :stage, touch: true ?

Probably even cleaner, I'll give it a go.

henders:
Hmm... nvm, couldn't figure out how to get it working:
```
class DeployGroupsStage < ActiveRecord::Base
  belongs_to :stage, touch: true
  after_save :hello
  def hello
    Rails.logger.info "***************** HEELLLOOOOO "
  end
end
```

So leaving it the way it is.

grosser:
playing with this atm ... stay tuned

grosser:
pushed a fix, works when adding on both sides

henders:
Oh, nice! Thanks @grosser , do you want to +1 your code or do I give the +1 now? :)

grosser:
:+1: But you should +1 and merge too :)

henders:
Happy to do so! :)  :+1:

PR #510, opened 2015-07-08
force ssl except for ping

sandlerr:
@zendesk/runway @zendesk/samson @babbottscott 
https://zendesk.atlassian.net/browse/OP-12730

The problem is to use SSL means we need to access not just the Rails app but also nginx and friends. So the ping controller should be accessible without using ssl.

Alternative solution by @adammw would be for the health check to pass `X-Forwarded-Proto: https` header


grosser:
:+1: I kind of like ping being universally accessible ...

PR #511, opened 2015-07-09
Fix loading commit when comparing to master

kureikain:
@grosser 
The changeset is cached with cache key is based on previous commit
and the tag when deploying. If the tag is a branch name, such as
`master`, the cache key is same leads to us seeing old data. If we
push more commit to the branch, new changeset won't be updated.

I ran into this issue when demoing Samson today to the team :(.
Someone try to push more commit and changeset still shows nothing

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -1,5 +1,6 @@
 class Changeset
   attr_reader :repo, :previous_commit, :commit
+  BRANCH_TAGS = ["master", "develop"]
 
   def initialize(repo, previous_commit, commit)
     @repo, @commit = repo, commit
@@ -70,6 +71,11 @@ class Changeset
     if empty?
       NullComparison.new(nil)
     else
+      # for branches that need review we make sure to always get the correct cache, others might get an outdated changeset if they are reviewed with different shas
+      if BRANCH_TAGS.include?(commit)
+        @commit = GITHUB.branch(repo, commit).commit[:sha]
+      end
+
       Rails.cache.fetch(cache_key) do
         GITHUB.compare(repo, previous_commit, commit)
       end
--- a/test/models/changeset_test.rb
+++ b/test/models/changeset_test.rb
@@ -11,11 +11,25 @@ describe Changeset do
       Changeset.new("foo/bar", "a", "b").comparison.to_h.must_equal :x => "y"
     end
 
-    it "caches" do
-      stub_github_api("repos/foo/bar/compare/a...b", "x" => "y")
-      Changeset.new("foo/bar", "a", "b").comparison.to_h.must_equal :x => "y"
-      stub_github_api("repos/foo/bar/compare/a...b", "x" => "z")
-      Changeset.new("foo/bar", "a", "b").comparison.to_h.must_equal :x => "y"
+    describe "with a specificed SHA" do
+      it "caches" do
+        stub_github_api("repos/foo/bar/compare/a...b", "x" => "y")
+        Changeset.new("foo/bar", "a", "b").comparison.to_h.must_equal :x => "y"
+        stub_github_api("repos/foo/bar/compare/a...b", "x" => "z")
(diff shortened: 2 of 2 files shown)


grosser:
hmm smells like an incomplete solution ... can we resolve the commit first then cache it with that as key ?

kureikain:
@grosser Ok, so if we detect a branch tag: `master` or `develop` we will try to find the commit of branch and cache with that key.

grosser:
how about we always do it ? my branch might be named fafafafa  ...

grosser:
hmm ideally we'd want to avoid hitting the api altogether ... so maybe don't do lookups for stable things like tags `/^v[\d\.?]+$/`

kureikain:
@grosser Yes, I think of it as well but I don't know a good way to detect branch name or sha. We prolly can ping API to see if it's a branch name or not but the point of cache is to eliminate hitting API, besides making it fast.

> maybe don't do lookups for stable things like tags /^v[\d\.?]+$/

Currently we only look up if branch match `master`, or `developer`. So `/^v[\d\.?]+$/` will not be looked up anyway, right?

I think of an idea:

If the `reference` or `tag` is not 7 chars or 40 chars we always try to find if a branch match that name, and fetch commit sha. Otherwise, no need to looked up commit sha.

kureikain:
Prolly `master` or `develop` is common enough to use because if people are not using branch, they use either tag and commit sha which won't be effected by this issue. People who deploy from branch probably has some common name so I think we good to add some more term like: `deploy`?

grosser:
We also deploy branches a bunch of times, just most often do not need review.
I never saw anyone deploy a straight up sha ... just master/staging/production/branches/tags ...
excluding tags + `^v[\d\.?]+|:alpha:{7}|:alpha:{40}$` could work though ...

grosser:
maybe we can use the deploy itself as cache-key for the commit lookup ... or do the sha resolve earlier ...

kureikain:
> maybe we can use the deploy itself as cache-key for the commit lookup ... or do the sha resolve earlier ...

With this approach, we made an extra request to do sha resolve, for 99% times when we may not need to do it.

To my opinion, in practice, this issue may never happen. We commit code, we deploy, the current sha changed, the cache expired. 

This issue only happens when we *did* going to deploy ui, fetching the changeset, realizing that we forgot something, so pushed more commit into branch, go back to the UI, fetching the changeset -> no new data. 

So should we optimize for a thing rarely happens? I think no, right. Just excluding branch name+tag ^v[\d\.?]+|:alpha:{7}|:alpha:{40}$) and clearly document this behaviour in the code?

grosser:
ahhh that is true ... the current sha changes after each successful deploy ...
just doing the lookup for master is probably fine then ... it's the only diff that might change and someone would care about ...

kureikain:
Yes, that's why in original commit I only doing for `master` and `develop`. Let's keep this simple. Rebase and ready for the merge. Test is passed before but rebase force it rebuild

app/models/changeset.rb
```diff
@@ -70,6 +71,11 @@ def find_comparison
     if empty?
       NullComparison.new(nil)
     else
+      # We try to fetch the SHA of branch
```
grosser: something like this ?
`# for branches that need review we make sure to always get the correct cache, others might get a outdated changeset if they are reviewed with different shas`

grosser:
:+1: yeah ... not perfect ... but definitely better than before ;)

kureikain:
yup, updated the comment. Good enough for now :)

kureikain:
@grosser Get some time for review #friday :)

grosser:
:+1:

sandlerr:
What if we run a command in the local checkout of the repo to distinguish tags, commits and branches? Then we can cache everything but branches.

sandlerr:
Thanks @kureikain !

kureikain:
> What if we run a command in the local checkout of the repo to distinguish tags, commits and branches? Then we can cache everything but branches.

I think that this is a good idea. What do you think @grosser?

PR #515, opened 2015-07-17
Only display the build project tab if the deploy flag is set in the p…

henders:
Only display the "Builds" project tab if the docker
 flag is set in the project settings.

project0

samson

/cc @zendesk/runway 

### References
 - Jira link:

### Risks
 - None


jonmoter:
:+1:

PR #516, opened 2015-07-17
Trigger docker build creation when webhooks are triggered.

henders:
Workflow now is:
- Webhook gets fired
- Build + Release records are built if project.release_branch is set
- Docker image is created + pushed if project.deploy_with_docker + project.auto_release_docker_image are set
- Commit is deployed if the branch matches any webhook mappings to stages.

I'll leave the actual hooking up to appropriate docker image reference when kicking off the deploys to another PR when we get staging/production up to actually test docker image creation first.

Screenshot with docker feature enabled for project:
samson

Screenshot with docker feature disabled for project:
samson1

/cc @zendesk/runway 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -3,25 +3,20 @@ class Integrations::BaseController < ApplicationController
   skip_before_action :verify_authenticity_token
 
   def create
-    if deploy?
-      create_new_release
-      unless deploy_to_stages
-        head :unprocessable_entity
-        return
-      end
+    return head(:ok) if !deploy?
+    if project.create_releases_for_branch?(branch)
+      create_build_record
+      create_docker_image if project.deploy_with_docker? && project.auto_release_docker_image?
     end
-
-    head :ok
+    deploy_to_stages ? head(:ok) : head(:unprocessable_entity, message: "Failed to start all deploys")
   end
 
   protected
 
   def create_new_release
-    if project.create_releases_for_branch?(branch)
-      unless project.last_release_contains_commit?(commit)
-        release_service = ReleaseService.new(project)
-        release_service.create_release!(commit: commit, author: user)
-      end
+    unless project.last_release_contains_commit?(commit)
+      release_service = ReleaseService.new(project)
+      release_service.create_release!(commit: commit, author: user)
     end
   end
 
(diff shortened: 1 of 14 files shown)


app/controllers/integrations/base_controller.rb
```diff
@@ -3,22 +3,17 @@ class Integrations::BaseController < ApplicationController
...
-      create_new_release
-      unless deploy_to_stages
-        head :unprocessable_entity
-        return
-      end
-    end
-
-    head :ok
+    return head :ok unless deploy?
+    project.deploy_with_docker? ? create_docker_build : create_release_and_deploy
```
grosser: can we have both active at the same time ?

app/controllers/integrations/base_controller.rb
```diff
@@ -60,4 +55,25 @@ def service_type
   def service_name
     @service_name ||= self.class.name.demodulize.sub('Controller', '').downcase
   end
+
+  def create_docker_build
+    release = create_new_release
```
grosser: `create_new_release || latest_release` would avoid making create_release also do a find
or rename to `create_or_find_release`

db/migrate/20150717135909_remove_unique_index_from_builds.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveUniqueIndexFromBuilds < ActiveRecord::Migration
+  def change
+    # Remove unique index on git_sha, users should be able to rebuild docker images against same SHA
```
jonmoter: So that means you could have two different Builds that reference the same git commit, and potentially have different labels. What's the advantage of allowing 2 different Build instances that refer to the exact same version of the code?

And why do users need to be able to rebuild a docker image for that SHA, given that the code is exactly the same?  If that is necessary, couldn't we just create functionality that re-runs the `docker build` command?

henders: I guess I had thought images could be generated differently depending on environment variable settings without having to touch the github repo. Do the docker projects use the .env stuff? Or are they always going to read everything from Consul? @jonmoter

jonmoter: ENV vars are not baked into the docker image.  You build the docker image once and push it to the registry, and then use the same image for all deployments (master, staging, gamma, pod1-6, etc.) Environment variables are injected when the image is run in a container.

When Samson builds an image, it effectively does this:

```shell
git checkout $SHA
echo $SHA > REVISION
docker build -t $DOCKER_TAG .
docker push $DOCKER_TAG
```

So the contents of the image are based solely on the contents of the repo. Unless we changed the logic on how we generate the builds (like removing the REVISION file), building an image twice should always generate the exact same image.

That make sense?

henders: Ah, yea, I had thought that the .env file functionality could be part of the docker build stage. Where a .env file would be created and packaged as part of the image. 

Anyway, I'll re-add the unique index, and just re-push the docker image if the record already exists.

jonmoter: It could, but then we'd need to create different images for master, staging, each pod, and we wouldn't be able to reuse the same image for developers.  I think we're better off having a single image for each version of the code, and then inject configuration by ENV var or mounted files.

henders:
@zendesk/runway ready for review again!

jonmoter:
Looks good!  :+1: 

The one thing I notice is that both a Release and a Build gets created, which seems unnecessary.  At some point I'd like to combine the two of them together, but don't know how to do that cleanly.

PR #518, opened 2015-07-17
Fixes crash when a client disconnects

cmckni3:
Sorry, recreating with amended commit.

### DIFF
--- a/app/models/event_streamer.rb
+++ b/app/models/event_streamer.rb
@@ -42,8 +42,7 @@ class EventStreamer
     emit_event('started', @handler.call(:started, ''))
     @scanner = TerminalOutputScanner.new(output)
     @scanner.each {|event, data| emit_event(event, @handler.call(event, data)) }
-  rescue IOError
-  rescue ActionController::Live::ClientDisconnected
+  rescue IOError, ActionController::Live::ClientDisconnected
     # Raised on stream close
   ensure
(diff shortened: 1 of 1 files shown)


app/models/event_streamer.rb
```diff
@@ -43,6 +43,7 @@ def start(output)
     @scanner = TerminalOutputScanner.new(output)
     @scanner.each {|event, data| emit_event(event, @handler.call(event, data)) }
   rescue IOError
+  rescue ActionController::Live::ClientDisconnected
```
henders: Can just combine the exception catching, thus making it clear '# Raised on stream close' is referring to either exception being thrown rather than 2 empty blocks:
```
rescue IOError, ActionController::Live::ClientDisconnected
  # Raised on stream close
...
```

app/models/event_streamer.rb
```diff
@@ -81,6 +83,7 @@ def start_heartbeat!
           sleep(5) # Timeout of 5 seconds
         end
       rescue IOError
+      rescue ActionController::Live::ClientDisconnected
```
henders: Rescues don't fall-through like this, so combine the lines like in my comment above.

cmckni3: Good catch

henders:
Nice, :+1:

PR #519, opened 2015-07-21
Allow deploys to determine what stage they are running from.

henders:
Mainly an internal request so that stages can create a local asset cache based on the stage name they are running inside while still being able to use a command command that's shared across stages.

Using the permalink instead of the name so that it can be dual purposed to allow creating URL to the stage as well.

/cc @zendesk/runway 

### References
 - Jira link:

### Risks
 - None


lukkry:
+1

grosser:
:+1:

PR #521, opened 2015-07-21
clarify message

sandlerr:
@zendesk/samson @princemaple 
This is a constant source of confusion


princemaple:
Good time to extract a partial?
:+1:

sandlerr:
Hmm similar structure but a lot of the strings are different. I think it's symptomatic of classes that need to share a lot more code than they currently do.

PR #522, opened 2015-07-22
Add option to login using LDAP authentication

duduribeiro:
I don't know if it is a valid change for Samson. This change add a way to login using LDAP. It's a valid change for us since all users are stored on the LDAP server.
- The user email is the same registered on LDAP.
- The user is logged with the Viewer role (same as Goggle Auth). But if this PR is valid, we can improve it to retrieve a custom field from LDAP.

This PR is valid for us because we have 1 client that they servers can't reach the internet. So, the unique login option for it is the LDAP.

Is the plan for Samson to support new authentications? If yes, I can work on Gitlab login too (will help us to use the self hosted Gitlab from our client)

The new button on UI:
ui_change

After click on LDAP button, The user is redirected to a Omniauth default form page: (Maybe can change on a future)
omni_auth_login

After login:
after_login

Thanks so much =D

:beers:


beniwtv:
+1 I also use LDAP, this would be nice!

steved:
Thanks for the PR (and sorry it took so long)! I did make one small change to have it be disabled by default (so you need an explicit AUTH_LDAP=1 to enable). Feel free to add more functionality and authentication strategies.

Issue #127, opened 2014-10-06
Use permalinks for stages too


grosser:
`https://samson.foo/projects/something/stages/master` would be a nice url format

PR #524, opened 2015-07-23
First pass of deploying to Kubernetes through Samson

jonmoter:
This is a research project for deploying projects using Docker and [Kubernetes](http://kubernetes.io/). It leverages the [kubeclient](https://github.com/abonas/kubeclient) gem to connect to the kubernetes API.

I'm building this as a plugin, so this be selectively enabled if a user is interested in this functionality.

So far, here's the working functionality:
* reading in a Kubernetes config file to get client connection information
* configuring connections to one or more Kubernetes clusters
* assigning DeployGroups to Clusters
* setting up "Roles" for a Kubernetes deployment (app server, resque worker, etc.)
* reading in a ReplicationController definition file from the project repo
* deploying a project to Kubernetes based on its ReplicationController definition file(s)
* the beginnings of code to watch the status of the deploy

# Screenshots

Cluster management:

List of Clusters

Details of a cluster connection:

Cluster Attributes

Defining a "Role" in Kubernetes:

edit_role_ui

UI for deploying to one or more clusters:

k8s_deployment_ui


jonmoter:
@zendesk/samson, I have this plugin in a reasonably stable state, and this PR has gotten gigantic. I'd like to merge this into master, and then create smaller PRs for incremental code.

Most of the code changes are isolated in a plugin.  There are a handful of changes to controller and view code, but they're wrapped in some logic that checks whether the Kubernetes plugin has been enabled.

There's a lot here, so let me know if you have any questions about the functionality or design.

henders:
Can you put in an appropriate readme so that developers can figure out how to get this going against a Kubernetes server? I'm not sure where to start! :)

Also screenshots would be nice eye-candy to get the blood pumping before diving in!

jonmoter:
I've added a README file at plugins/kubernetes/README.md

jonmoter:
I've added some screenshots to the description above, and fixed the merge conflicts.

plugins/kubernetes/app/controllers/kubernetes_clusters_controller.rb
```diff
@@ -0,0 +1,79 @@
+class KubernetesClustersController < ApplicationController
+  before_action :authorize_admin!
+  before_action :authorize_super_admin!, only: [ :create, :new, :update ]
```
steved: edit too?

plugins/kubernetes/db/migrate/20150827181056_create_kubernetes_cluster.rb
```diff
@@ -0,0 +1,11 @@
+class CreateKubernetesCluster < ActiveRecord::Migration
+  def change
+    create_table :kubernetes_clusters do |t|
+      t.string :name
+      t.string :description
+      t.string :config_filepath
```
steved: This is a local filepath? Are you expecting this to be written out by a config management system? Can Samson be the source of truth here?

db/schema.rb
```diff
@@ -54,12 +54,14 @@
...
+    t.string   "name",                  limit: 255, null: false
+    t.integer  "environment_id",        limit: 4,   null: false
     t.datetime "deleted_at"
-    t.datetime "created_at",                 null: false
-    t.datetime "updated_at",                 null: false
-    t.string   "env_value",      limit: 255, null: false
+    t.datetime "created_at",                        null: false
+    t.datetime "updated_at",                        null: false
+    t.string   "env_value",             limit: 255, null: false
+    t.integer  "kubernetes_cluster_id", limit: 4
```
steved: I'd kind of rather see a join table with a unique deploy group id index to keep deploy groups from knowing anything about kubernetes. You should be able to keep everything in a plugin then as well.

app/controllers/builds_controller.rb
```diff
@@ -98,6 +106,12 @@ def start_docker_build
     DockerBuilderService.new(@build).run!(push: true)
   end
 
+  def find_kuber_releases
+    if @build.respond_to?(:kubernetes_releases)
+      @kuber_release_list = @build.kubernetes_releases.order('id desc')
+    end
+  end
```
steved: Are any of these still used? It looks like you have a KubernetesReleasesController

plugins/kubernetes/app/controllers/kubernetes_clusters_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  before_action :find_cluster, only: [:show, :edit, :update]
+  before_action :load_environments, only: [:new, :edit, :update]
+  before_action :load_default_config_file, only: [:new, :edit, :update]
+
+  def new
+    @cluster = Kubernetes::Cluster.new(config_filepath: @config_file.try(:filepath))
```
steved: @config_file should not be nil

plugins/kubernetes/app/controllers/kubernetes_releases_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def new_params
+    params.permit(:build_id)
+  end
+
+  def create_params
+    params.require(:kubernetes_release).permit(:build_id)
+  end
+
+  def selected_deploy_groups
```
steved: Used?

plugins/kubernetes/app/decorators/build_decorator.rb
```diff
@@ -0,0 +1,14 @@
+Build.class_eval do
+
+  has_many :kubernetes_release_groups, class_name: 'Kubernetes::ReleaseGroup'
+  attr_writer :project_name
```
steved: Why?

plugins/kubernetes/app/decorators/build_decorator.rb
```diff
@@ -0,0 +1,14 @@
...
+  attr_accessor :env
+
+  def project_name
+    @project_name ||= project.name.parameterize('-')
```
steved: I'd rather see this on project (`parameterized_name` or something), but it's fine

plugins/kubernetes/app/decorators/build_decorator.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  def env_for(deploy_group)
+    EnvironmentVariable.env(project, deploy_group)
```
steved: This seems odd to be on build

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -0,0 +1,58 @@
...
+    kuber_release.release_docs.each do |release_doc|
+      log 'creating ReplicationController', role: release_doc.kubernetes_role.name
+
+      rc = Kubeclient::ReplicationController.new(release_doc.rc_hash)
+      client.create_replication_controller(rc)
+    end
+
+    # Doing this for now to force the loading of these constants.
+    # I was getting a circular dependency error otherwise
+    Watchers::PodWatcher ; Watchers::ReplicationControllerWatcher ; Watchers::EventWatcher
```
steved: autoloading isn't thread safe, afaik. If you see this with eager loading on then it's a problem.

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -0,0 +1,126 @@
...
+
+        {
+          client_cert: user.client_cert,
+          client_key:  user.client_key,
+          ca_file:     cluster.ca_filepath,
+          verify_ssl:  OpenSSL::SSL::VERIFY_PEER
+        }
+      end
+    end
+  end
```
steved: Just an interesting note, you could use structs for these.
```
class Context < Struct.new(:name, :cluster, ...)
  def use_ssl?
    cluster...
  end
end
```

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    Thread.new { kuber_release.watch_pods }
+    Thread.new { kuber_release.watch_rcs }
+    Thread.new { sleep(5) ; kuber_release.watch_pod_events }
+  end
```
steved: You're kind of just throwing threads out into the world, which is a little scary

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -0,0 +1,126 @@
...
+      def url
+        server + '/api/'
+      end
+
+      def ca_filepath
+        @ca_filepath ||= begin
+          tmpfile = Tempfile.new(['ca', '.crt'])
+          tmpfile.write(Base64.decode64(ca_data))
+          tmpfile.close
+          tmpfile.path
```
steved: I think there's a possibility the tmpfile is gone by the time you use it, since they're removed on GC

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -0,0 +1,89 @@
...
+      {
+        project: release_group.build.project_name,
+        release_id: id.to_s
+      }
+    end
+
+    # TODO: define state machine for 'status' field
+
+    def nested_error_messages
+      errors.full_messages + release_docs.map(&:nested_error_messages).flatten
```
steved: There's a flat_map method

app/controllers/builds_controller.rb
jonmoter: No, I think that was functionality I had used at one point, but since removed. I'll delete these.

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -0,0 +1,23 @@
...
+  class Role < ActiveRecord::Base
+    self.table_name = 'kubernetes_roles'
+    belongs_to :project, inverse_of: :roles
+
+    DEPLOY_STRATEGIES = ['rolling_update', 'kill_and_restart']
+
+    validates :project, presence: true
+    validates :name, presence: true
+    validates :deploy_strategy, presence: true, inclusion: DEPLOY_STRATEGIES
+    validates :replicas, presence: true, numericality: { greater_than: 0 }
```
steved: Is it possible to scale down to 0?

db/schema.rb
jonmoter: Okay, that's a good call.  I'll refactor.

plugins/kubernetes/app/controllers/kubernetes_releases_controller.rb
jonmoter: Not anymore.  Will delete.

plugins/kubernetes/lib/samson_kubernetes/samson_plugin.rb
```diff
@@ -0,0 +1,9 @@
+require_relative 'hash_kuber_selector'
+
+module SamsonKubernetes
+  class Engine < Rails::Engine
```
steved: I'm a fan of the [isolate_namespace](http://guides.rubyonrails.org/engines.html#critical-files) option. Especially since you can pretty neatly isolate the `Kubernetes` namespace from everything else.

plugins/kubernetes/app/models/watchers/pod_watcher.rb
```diff
@@ -0,0 +1,84 @@
+module Watchers
+  class PodWatcher < BaseWatcher
+    POD_PHASES = %w[Pending Running Succeeded Failed Unknown]
+    
+    attr_reader :pod_info
+
+    def initialize(client, namespace, name: nil, label_selector: nil, log: true)
+      raise ArgumentError.new('Must specify either pod_name or label_selector') if name.blank? && label_selector.blank?
```
steved: Are you worried about empty names or nil ones? 
Keyword arguments like `initialize(x, y, name:, label_selector:)` will raise the appropriate error message

plugins/kubernetes/app/models/watchers/pod_watcher.rb
```diff
@@ -0,0 +1,84 @@
...
+        Time.parse(object.status.startTime)
+      end
+
+      def running?
+        phase == 'Running'
+      end
+
+      def ready?
+        object.status.conditions.present? &&
+          object.status.conditions.select { |c| c['type'] == 'Ready' }.all? { |c| c['status'] == 'True' }
```
steved: Just one loop? `all? {|c| c['status'] == 'True' && c['type'] == 'Ready'}`

plugins/kubernetes/app/models/watchers/pod_watcher.rb
```diff
@@ -0,0 +1,84 @@
...
+    # Keep track of the Pods that are created
+    # Watch for them to be all running and ready
+    # Once we have a set of all pod names, subscribe to events related to them
+    #   e.g. client.watch_events(fieldSelector: "involvedObject.name IN (pod1,pod2,...)")
+    #   look for any Events that look like a failure (e.g. failedScheduling)
+    # When all the pods are ready and running, mark the deploy as successful
+
+    def handle_notice(notice, &block)
+      return if handle_error(notice)
+      notice.extend(PWrapper)
```
steved: Runtime object extension used to be really expensive. I think it's been fixed in recent versions, but something to keep in mind if you're handling *a lot* of events.

http://dec0de.me/2013/02/dci-performance/
https://github.com/charliesome/charlie.bz/blob/master/posts/things-that-clear-rubys-method-cache.md

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -0,0 +1,146 @@
...
+    validate :validate_config_file, on: :create
+
+    after_create :save_rc_info    # do this after create, so id has been generated
+
+    # Definition of the ReplicationController, based on the rc_template, but
+    # updated with the image and metadata for this deploy.
+    #
+    # This hash can be passed to the Kubernetes API to create a ReplicationController.
+    def rc_hash
+      return @rc_hash if instance_variable_defined?(:@rc_hash)
```
steved: Can just be a `defined?(@rc_hash)` call

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -0,0 +1,146 @@
+module Kubernetes
+  class ReleaseDoc < ActiveRecord::Base
+    self.table_name = 'kubernetes_release_docs'
```
steved: isolate_namespace should actually help with all these table_names, or just

steved: Or defining table_name_prefix: http://apidock.com/rails/ActiveRecord/Base/table_name_prefix/class

plugins/kubernetes/app/models/kuber_deploy_service.rb
jonmoter: Yeah, good point.  This is actually temporary code while I figure out how to properly monitor the state of the pods on deployment.  I'll comment these out for the PR.

I'll also add a note to the Readme that this plugin is incomplete and under active development.

plugins/kubernetes/app/models/watchers/pod_watcher.rb
jonmoter: One of `name` or `label_selector` needs to be defined.  Using keyword arguments like that would throw an error if either of them were nil, which isn't what I want.

steved: Yeeep, totally missed that

plugins/kubernetes/app/controllers/kubernetes_clusters_controller.rb
jonmoter: Actually, I shouldn't create the `ClientConfigFile` if the file doesn't exist. Will fix that logic below.

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -0,0 +1,126 @@
+module Kubernetes
+  class ClientConfigFile
```
jonmoter: FYI, there's currently a PR open for kubeclient ([PR #127](https://github.com/abonas/kubeclient/pull/127/files)) to implement this as part of the client gem. Once that gets merged in, I'll just use that functionality.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
jonmoter: I'm looking at the documentation for [Rails Engines](http://edgeapi.rubyonrails.org/classes/Rails/Engine.html), and it appears that if I use `isolate_namespace`, my Kubernetes plugin code will be unable to access info from the main app.

> If the MyEngine engine is marked as isolated, FooController only has access to helpers from MyEngine, and url_helpers from MyEngine::Engine.routes.

So I wouldn't be able to generate links from the Kubernetes plugin to any other views (like DeployGroups), which seems like it would be limiting.

plugins/kubernetes/app/models/kubernetes/role.rb
jonmoter: I can't think of a scenario where you'd want to define a role that had zero replicas.  If we come up with one, I'll change this validation.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
steved: There's a `main_app` object that you call those helpers on. It's just that those helpers aren't included in all your controllers.

plugins/kubernetes/app/models/watchers/pod_watcher.rb
jonmoter: Not the same logic.  There could be other entries in the array where `c['type']` is something other than "Ready".

plugins/kubernetes/db/migrate/20150827181056_create_kubernetes_cluster.rb
jonmoter: Currently, the logic is expecting a filepath on the Samson server.  Yeah, I know this isn't ideal.

The config contains a bunch of information, including stuff like base-64 encoded certificates, CAs, etc. Creating a UI to manage all that would be a pain.  The config file is automatically generated by the Kubernetes installation scripts when you run the Vagrant installation, and could be generated by a config management system or scripts.

I'd like to start with this, and possibly add a future enhancement to allow inputting config information by a UI.

plugins/kubernetes/app/models/watchers/pod_watcher.rb
steved: Fair enough... `(c['status'] == 'True' && c['type'] == 'Ready') || c['type'] != 'Ready'`. Only really matters if there are a lot of conditions.

jonmoter:
@steved, updated PR based on feedback you gave me.

jonmoter:
Argh.  Test failing due to the nested attributes.  Fixing...

plugins/kubernetes/app/decorators/deploy_group_decorator.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  accepts_nested_attributes_for :cluster_deploy_group,
+                                allow_destroy: true,
+                                update_only: true,
+                                reject_if: lambda { |h| h[:kubernetes_cluster_id].blank? }
+
+  def kubernetes_namespace
+    cluster_deploy_group.try(:namespace)
+  end
+
```
steved: :scissors:

plugins/kubernetes/app/controllers/kubernetes_clusters_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+  def load_environments
+    @environments = Environment.includes(:deploy_groups).all
+  end
+
+  def new_cluster_params
+    params.require(:kubernetes_cluster).permit(:name, :config_filepath, :config_context, :description, { deploy_group_ids: [] })
+  end
+
+  def load_default_config_file
+    @config_file = File.exists?(ENV['KUBE_CONFIG_FILE']) ? Kubernetes::ClientConfigFile.new(ENV['KUBE_CONFIG_FILE']) : nil
```
steved: I'd rather see these two assignments wrapped in an if..else, it's a little clearer

steved:
+1, I just have some style issues. I hope we can come up with a better way than injecting into deploy groups.

plugins/kubernetes/app/decorators/deploy_group_decorator.rb
jonmoter: Remove the `kubernetes_namespace` method?  Why?

steved: just the newline, my ocd kicks in

jonmoter: Ah, the extra newline.  Got it.

PR #528, opened 2015-07-31
add New Relic instrumentation for repository actions [RUN-191]

jonmoter:
Add instrumentation for some known expensive operations, like git commands

/cc @zendesk/runway 

### References
 - Jira link: https://zendesk.atlassian.net/browse/RUN-191


grosser:
:+1:

PR #526, opened 2015-07-28
Revert "Merge pull request #462 from zendesk/hsiojo/kill9"

anamartinez:
This reverts commit 4cd66bd132aae02aca167e6c39a3154f38a7de93, reversing
changes made to 8472d4af9208a8f7eef236edc34bcf678fdf2b92.

@grosser

PR #529, opened 2015-08-03
Revert "Merge pull request #526 from zendesk/amartinez/revert_grosser"

grosser:
rerevert the kill commit since it was found to be not guilty ...
played with it a bit on staging, but I cannot find a way to make it not work or simulate the remaining stuck errors we see on prod ... should still be a step better then before

@anamartinez @zendesk/runway 

### Risks
 - None

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -9,6 +9,7 @@ class JobExecution
   end
 
   cattr_accessor(:lock_timeout, instance_writer: false) { 10.minutes }
+  cattr_accessor(:stop_timeout, instance_accessor: false) { 15.seconds }
 
   cattr_reader(:registry, instance_accessor: false) { {} }
   private_class_method :registry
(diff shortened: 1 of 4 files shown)


app/models/job_execution.rb
```diff
@@ -37,7 +38,12 @@ def wait!
   end
 
   def stop!
-    @executor.stop!
+    @executor.stop! 'INT'
+    self.class.stop_timeout.times do
```
jonmoter: if `stop_timeout` is a time (like `15.seconds`), you shouldn't be doing `stop_timeout.times`

app/models/terminal_executor.rb
```diff
@@ -33,10 +33,10 @@ def execute!(*commands)
     execute_command!(command)
   end
 
-  def stop!
+  def stop!(signal)
     # Need pkill because we want all
     # children of the parent process dead
-    `pkill -INT -P #{pid}` if pid
+    `pkill -#{signal} -P #{pid}` if pid
```
jonmoter: We probably should guard against command-line injection attacks here.

app/models/job_execution.rb
grosser: changed it to .to_i.times to read better ... but works either way

app/models/terminal_executor.rb
grosser: changed to system + array

jonmoter:
:+1:

PR #530, opened 2015-08-03
fix not being able to click into text area

grosser:
@sandlerr not sure what this cancel option was supposed to do, but it breaks editing ...
/cc @zendesk/runway 

### Risks
 - None


sandlerr:
:+1: you mean it breaks being able to move the cursor with the mouse? Not sure what that's there for either, my change to this code was a fix because the sorting functionality wasn't working at all

grosser:
sorting works and editing the code inline works too now :)

PR #533, opened 2015-08-17
Fix heroku incompatibility.

henders:
Heroku strips git metadata when creating its deploy slugs. So just removing the exception that prevents startup.


sandlerr:
:+1:

grosser:
So heroku runs without revisions now ... maybe make it 'NO_REVISION_FOUND' so it's more obvious ?

PR #534, opened 2015-08-19
SAMSON-170: add simple JIRA issue key matching

yihongang:
Requires an environment variable `JIRA_URL` to be set.
Format should be a URL including the trailing slash.
Example: https://jira.zendesk.com/browse/

If `JIRA_URL` is not set, behaviour reverts to a more general case of the current behaviour.
It will match generic URLs with a /browse/ISSUE-KEY format instead of just Atlassian ones.

If `JIRA_URL` is set, plain issue keys (e.g., ABC-123) will have full URLs generated.
e.g., https://jira.zendesk.com/browse/ABC-123

Works for issue keys in both title and body of the PR.

### DIFF
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -1,13 +1,19 @@
 class Changeset::PullRequest
+  # Common patterns
+  CODE_ONLY = "[A-Z][A-Z\\d]+-\\d+"  # e.g., S4MS0N-123, SAM-456
+  PUNCT = "\\s|\\p{Punct}|~|="
+
   # Matches a section heading named "Risks".
   RISKS_SECTION = /#+\s+Risks.*\n/i
 
   # Matches URLs to JIRA issues.
-  JIRA_ISSUE = %r[https://\w+\.atlassian\.net/browse/[\w-]+]
+  JIRA_ISSUE_URL = %r[https?:\/\/[\da-z\.\-]+\.[a-z\.]{2,6}\/browse\/#{CODE_ONLY}(?=#{PUNCT}|$)]
 
   # Matches "VOICE-1234" or "[VOICE-1234]"
-  JIRA_CODE = %r[(\[)*([a-zA-Z]+)-((\d)+)(\])*]
+  JIRA_CODE_TITLE = %r[(\[)*(#{CODE_ONLY})(\])*]
 
+  # Matches "VOICE-1234" only
+  JIRA_CODE = %r[(?<=#{PUNCT}|^)(#{CODE_ONLY})(?=#{PUNCT}|$)]
   # Finds the pull request with the given number.
   #
   # repo   - The String repository name, e.g. "zendesk/samson".
@@ -34,7 +40,7 @@ class Changeset::PullRequest
   delegate :number, :title, :additions, :deletions, to: :@data
 
   def title_without_jira
-    title.gsub(JIRA_CODE, "").strip
+    title.gsub(JIRA_CODE_TITLE, "").strip
   end
 
   def url
@@ -62,15 +68,25 @@ class Changeset::PullRequest
   end
 
   def jira_issues
-    @jira_issues ||= parse_jira_issues!
+    @jira_issues ||= parse_jira_issues
   end
 
   private
 
-  def parse_jira_issues!
-    body.scan(JIRA_ISSUE).map do |match|
-      Changeset::JiraIssue.new(match)
+  def parse_jira_issues
+    custom_jira_url = ENV['JIRA_BASE_URL']
+    title_and_body = "#{title} #{body}"
+    jira_issue_map = {}
+    if custom_jira_url
+      title_and_body.scan(JIRA_CODE).each do |match|
+        jira_issue_map[match[0]] = custom_jira_url + match[0]
+      end
     end
+    # explicit URLs should take precedence for issue links
+    title_and_body.scan(JIRA_ISSUE_URL).each do |match|
+      jira_issue_map[match.match(JIRA_CODE)[0]] = match
+    end
+    jira_issue_map.values.map { |x| Changeset::JiraIssue.new(x) }
   end
 
   def body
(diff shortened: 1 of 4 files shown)


app/models/changeset/pull_request.rb
```diff
@@ -3,11 +3,14 @@ class Changeset::PullRequest
   RISKS_SECTION = /#+\s+Risks.*\n/i
 
   # Matches URLs to JIRA issues.
-  JIRA_ISSUE = %r[https://\w+\.atlassian\.net/browse/[\w-]+]
+  JIRA_ISSUE_URL = %r[https?:\/\/[\da-z\.\-]+\.[a-z\.]{2,6}\/browse\/[A-Z][A-Z\d]+-\d+]
```
sandlerr: wouldn't it make sense to inject JIRA_URL into the regex?

yihongang: There are two intentions in the JIRA issue linked to this PR. My bad I should have mentioned both.

The other intention behind the change is to support both JIRA Cloud (which would be hosted on *.atlassian.net) and JIRA Server (which could be hosted on any hostname) by default. So I think that the generic (non-JIRA_URL-injected) regex should still be there by default.

I guess the case could be made for conditionally injecting the regex, but that would be a design decision.

The current implementation is in line with the relevant description in the linked JIRA issue which I would quote here:
 
> If a full JIRA url is mentioned, we would take that full url. Otherwise, the issue key found will be prefixed with the base url configured below

zehua:
We should get the new `JIRA_URL` (maybe rename it as `JIRA_BASE_URL`) into `.env.example`, probably commented by default but with some explanation, so that users would know how to use it and what to expect.

app/models/changeset/pull_request.rb
```diff
@@ -62,15 +65,25 @@ def risks
...
   end
 
   private
 
-  def parse_jira_issues!
-    body.scan(JIRA_ISSUE).map do |match|
-      Changeset::JiraIssue.new(match)
+  def parse_jira_issues
+    custom_jira_url = ENV['JIRA_URL']
+    title_and_body = "#{title} #{body}"
```
zehua: We are scanning both the title and body for JIRA issue keys.

app/models/changeset/pull_request.rb
sandlerr: > This would also retain the flexibility of referencing issues from multiple JIRA instances by their full URLs, if that is a valid use case.

I guess this is what I'm really questioning. The main application of this I that I can see is for organisations that switch JIRA instances or move between on-premise and cloud. IMO this is an edge case, but I suppose since you've done the work it would be better to keep it.

Can you add a specific test case that make use of this functionality (i.e., pulling out issues from multiple JIRA domains?

zehua: The original idea is the if an absolute url is entered, we should always respect it. It would be surprising to the users if they entered a full url and the final url presented on the UI is different from what they entered.

I agree that one org using multiple JIRA instances is not common, but it does happen in large organizations, where they do it because JIRA was not able to scale that well. Given that we are limiting the support to one BASE_URL in the configuration at the moment, I just hope that we could at least have some workaround (i.e., asking users to always enter full urls) to this problem.

yihongang:
added `JIRA_BASE_URL` to `.env.example`

app/models/changeset/pull_request.rb
```diff
@@ -62,15 +65,25 @@ def risks
...
   private
 
-  def parse_jira_issues!
-    body.scan(JIRA_ISSUE).map do |match|
-      Changeset::JiraIssue.new(match)
+  def parse_jira_issues
+    custom_jira_url = ENV['JIRA_BASE_URL']
+    title_and_body = "#{title} #{body}"
+    jira_issue_map = {}
+    if custom_jira_url then
```
henders: You can remove the 'then'

app/models/changeset/pull_request.rb
```diff
@@ -3,11 +3,14 @@ class Changeset::PullRequest
...
-  JIRA_ISSUE = %r[https://\w+\.atlassian\.net/browse/[\w-]+]
+  JIRA_ISSUE_URL = %r[https?:\/\/[\da-z\.\-]+\.[a-z\.]{2,6}\/browse\/[A-Z][A-Z\d]+-\d+]
 
   # Matches "VOICE-1234" or "[VOICE-1234]"
-  JIRA_CODE = %r[(\[)*([a-zA-Z]+)-((\d)+)(\])*]
+  JIRA_CODE_TITLE = %r[(\[)*([a-zA-Z]+)-((\d)+)(\])*]
 
+  # Matches "VOICE-1234" only
+  PUNCT = "\\s|\\p{Punct}|~|="
+  JIRA_CODE = %r[(?<=#{PUNCT}|^)([A-Z][A-Z\d]+-\d+)(?=#{PUNCT}|$)]
```
henders: Might be overkill on the amount of matching punctuation, but that's grand. :)  Should fix the comment as it matches all variants of VOICE-1234 using '"{}[]()=, so maybe point out a few examples.

The Project key matching part '[A-Z][A-Z\d]+' differs from the above JIRA_CODE_TITLE version '[a-zA-Z]+', so they should be made consistent.
Default Jira configuration has: '[A-Za-z][A-Za-z]+-\d+', though a jira admin can change it unfortunately. However, not worth the complexity of making it configurable IMO.

henders:
Also, document the new variable in the README.  You may need to rebase from master to get the latest doc changes to avoid conflict.

app/models/changeset/pull_request.rb
zehua: Yeah. If the JIRA admin decides to do utilize his/her super power, even DVCS Connector / Bitbucket / Stash will not work properly.

yihongang: I think the comment is still accurate because the punctuations are look behind and look ahead only. I have moved out the code regex to a common one and added some examples of valid codes.

sandlerr:
Looks good!

yihongang:
rebased from master and added documentation to docs/extra_features.md

henders:
:+1:

yihongang:
squashed :)

PR #535, opened 2015-08-19
Allow airbrakes to be output to log when in development mode.

henders:
There are a few places in the code (especially Plugins) that don't log the error, and only send an Airbrake if it's defined.

This will allow developer to see when an airbrake is going to be sent when in development mode.
It just prints the error and a section of the backtrace.

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -1,7 +1,17 @@
-if !defined?(Rails) || Rails.env.staging? || Rails.env.production?
-  require 'airbrake'
+require 'airbrake'
 
+if !defined?(Rails) || Rails.env.staging? || Rails.env.production?
   Airbrake.configure do |config|
     config.api_key = ENV['AIRBRAKE_API_KEY']
+  end
+else
+  module Airbrake
+    def self.notify(ex, *args)
+      logger.error "AIRBRAKE: #{ex.message} - #{ex.backtrace[0..5].join("\n")}"
+    end
+
+    def self.notify_or_ignore(ex, *args)
+      notify(ex)
+    end
   end
 end
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -1,16 +1,15 @@
-require 'airbrake'
-
 if !defined?(Rails) || Rails.env.staging? || Rails.env.production?
+  require 'airbrake'
   Airbrake.configure do |config|
     config.api_key = ENV['AIRBRAKE_API_KEY']
   end
 else
   module Airbrake
-    def self.notify(ex, *args)
-      logger.error "AIRBRAKE: #{ex.message} - #{ex.backtrace[0..5].join("\n")}"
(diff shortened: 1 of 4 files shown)


jonmoter:
The change looks good, but you broke a test.

config/initializers/airbrake.rb
```diff
@@ -1,7 +1,17 @@
-if !defined?(Rails) || Rails.env.staging? || Rails.env.production?
-  require 'airbrake'
+require 'airbrake'
 
+if !defined?(Rails) || Rails.env.staging? || Rails.env.production?
   Airbrake.configure do |config|
     config.api_key = ENV['AIRBRAKE_API_KEY']
   end
+else
+  module Airbrake
```
grosser: no need to require in this case ?

grosser:
:+1: nice to see the errors ... airbrakes default behavior is to not log anything :(

config/initializers/airbrake.rb
gabetax: The ideal situation here would be to use an exception notification service adapter. I'm surprised there isn't any public gem (that I found) that handles this.

grosser: how about a `Samson.exception_handler` callback that gets a message +
exception ?

henders: Since this is just a development mode fix, keeping it simple is probably best. When an airbrake is thrown to the logs during development, the first thing I do is go directly to the line of code that threw it.

henders: @grosser 

> no need to require in this case ?

The 'require' was moved up out of the conditional.

grosser: can the require only be in the `if` part ?

grosser:
Also remove all `defined?(Airbrake)` checks :)

grosser:
I was thinking `Samson.exception_handler` would make it easy to plug in different backends, but maybe best to keep this PR simple ;)

config/initializers/airbrake.rb
henders: I'm overriding the Airbrake module instead of creating a new one because there were some functions setting the logger and calling other public methods if it was defined. So I decided to just override the particular methods that should output to the rails logger, and leave the rest of the methods as-is.

henders:
Removed defined?(Airbrake) check, @grosser   :)

config/initializers/airbrake.rb
grosser: not a big fan of silent swallowing ... would an `attr_accessor :logger` fix it ?
a simple grep only shows the Airbrake.notify calls ... where are the others coming from ?

henders: Hmm... you're right, I could have sworn the test helpers were using it. But nope, my mistake!

Made it a standalone dummy module, rather than monkey-patching.

grosser:
:+1:

PR #538, opened 2015-08-21
Allow binaries to be compiled in separate docker image to reduce size…

henders:
This plugin basically allows apps that deal with compiled languages to not have bloated docker deployment images.

At moment projects deal with deploying apps with compiled binaries in 2 ways:
- Compile the binaries locally and checkin to the github repo.
- Have the docker image contain all the tools/compilers/libraries necessary to compile the binaries and build it all into the deployment image.

With a Scala app, this can lead to 1.5GB docker images.

To solve this we split building into the docker image into 2 phases:
- 1: Build and spin up a temporary build container that has all the tools/libs necessary to compile the binaries. A script inside this compiles the binaries, and Samson retrieves those binaries that are zipped into a tarball
- 2: The main Dockerfile is then used to build the docker image which now has all the binaries available but only needs the runtime libs in the actual image for deployment.

/cc @zendesk/runway @stubotnik @fneves @andreionut 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/plugins/docker_binary_builder/README.md
+++ b/plugins/docker_binary_builder/README.md
@@ -1,0 +1,63 @@
+# Docker Builder for Binaries
+
+This plugin is useful when your projects are based on compiled languages and you don't want to base
+your deployment docker images on images that have all the tools/libs/compilers installed.
+
+Advantages to separating your docker images include:
+- Smaller deployment docker images.
+- Better security in your docker images to not be able to recompile code.
+
+## Getting Started
+
+This plugin makes a few assumptions:
+- You have 2 dockerfiles:
+ - 'Dockerfile' used to generate your main deployment docker image
+ - 'Dockerfile.build' used to create a temporary image to build any binaries you want to package into the former.
+- You have a build script 'build.sh' that can be executed to compile the needed binaries
+  - This is packaged inside the Dockerfile.build image and executed
+  - After binaries are compiled the build.sh script finishes by tar'ing all necessary files into an 'artifacts.tar' file.
+  
+If you want to share caches between builds, e.g. all your projects are Java and use Maven to build, then sharing the .m2 
+directory will be a huge time saving tactic to share all downloaded dependencies.
+
+If the directory '/opt/samson_build_cache' exists on the Docker host, it will mount it to '/build/cache' inside the 
+docker build image. That way you could then instruct Maven to use '/build/cache/.m2' as the cache directory for all your 
+projects.
+
+## Example Setup
+
+'Dockerfile' contents
+```
+# We only need the JRE base image now
+FROM java:8u45-jre
+
+# These jars will be built in the other Dockerfile.build image and then copied into here.
+ADD target/scala-2.11/myproject-*.jar
+```
+
+
+'Dockerfile.build' contents
+```
+FROM scala_sbt:2.11.7
+
+# create cache directory in case it couldn't be mounted correctly
+RUN mkdir -p /build/cache
+# '/app' will be the main directory in this image
+RUN mkdir -p /app/src
+
+# Add the scala build config files
+ADD project /app/project
+# Add the source directory into this image 
+ADD src/main /app/src/main
+# Add our main build script into the image
+ADD build.sh /app/build.sh
+
+WORKDIR /app
+```
+
+'build.sh' contents:
+```
+#!/usr/bin/env bash
+sbt -ivy /build/cache/.ivy2 clean compile assembly
+cd /app && tar -cvf /app/artifacts.tar target/scala-2.11/*.jar
+```
--- a/plugins/docker_binary_builder/app/models/binary_builder.rb
+++ b/plugins/docker_binary_builder/app/models/binary_builder.rb
@@ -1,0 +1,132 @@
+class BinaryBuilder
+  DOCKER_FILE = 'Dockerfile.build'.freeze
+  BUILD_SCRIPT = '/app/build.sh'.freeze
+  ARTIFACTS_FILE = 'artifacts.tar'.freeze
+  ARTIFACTS_FILE_PATH = "/app/#{ARTIFACTS_FILE}".freeze
+  DOCKER_HOST_CACHE_DIR = '/opt/samson_build_cache'.freeze
+  CONTAINER_CACHE_DIR = '/build/cache'.freeze
+
+  def initialize(dir, project, reference, output)
+    @dir = dir
+    @project = project
+    @git_reference = reference
+    @output_stream = output
+  end
+
+  def build
+    return unless @project.try(:deploy_with_docker?) && File.exists?(File.join(@dir, DOCKER_FILE))
+    @output_stream.puts "Connecting to Docker host with Api version: #{docker_api_version} ..."
+
+    @image = create_image
+    @container = Docker::Container.create(create_container_options)
+
+    start_build_script
+    retrieve_binaries
+
+    @output_stream.puts 'Continuing docker build...'
+  ensure
+    @output_stream.puts 'Cleaning up docker build image and container...'
+    @container.delete(force: true) if @container
+    @image.remove(force: true) if @image
+  end
+
+  private
+
+  def start_build_script
+    @output_stream.puts 'Now starting Build container...'
+    @container.tap(&:start).attach { |_stream, chunk| @output_stream.write chunk }
+  rescue => ex
+    @output_stream.puts "Failed to run the build script '#{BUILD_SCRIPT}' inside container."
+    raise ex
+  end
+
+  def retrieve_binaries
+    @output_stream.puts "Grabbing '#{ARTIFACTS_FILE_PATH}' from build container..."
+    artifacts_tar = Tempfile.new(['artifacts', '.tar'], @dir)
+    artifacts_tar.binmode
+    @container.copy(ARTIFACTS_FILE_PATH) { |chunk| artifacts_tar.write chunk }
+    artifacts_tar.close
+
+    untar(artifacts_tar.path)
+    untar(File.join(@dir, ARTIFACTS_FILE))
+  end
+
+  def untar(file_path)
+    @output_stream.puts "About to untar: #{file_path}"
+
+    File.open(file_path, 'rb') do |io|
+      Gem::Package::TarReader.new io do |tar|
+        tar.each do |tarfile|
+          destination_file = File.join @dir, tarfile.full_name
+          @output_stream.puts "    > #{tarfile.full_name}"
+
+          if tarfile.directory?
+            FileUtils.mkdir_p destination_file
+          else
+            destination_directory = File.dirname(destination_file)
+            FileUtils.mkdir_p destination_directory unless File.directory?(destination_directory)
+            File.open destination_file, "wb" do |f|
+              f.print tarfile.read
+            end
+          end
+        end
+      end
+    end
+  end
+
+  def create_container_options
+    options = {
+      'Cmd' => [BUILD_SCRIPT],
+      'Image' => "#{@project.send(:permalink_base)}_build:#{@git_reference}"
+    }
+
+    # Mount a cache directory for sharing .m2, .ivy2, .bundler directories between build containers.
+    api_version_major, api_version_minor = docker_api_version.scan(/(\d+)\.(\d+)/).flatten.map(&:to_i)
+    if api_version_major == 0 || (api_version_major == 1 && api_version_minor <= 14)
+      fail "Unsupported Docker api version '#{docker_api_version}', use at least v1.15"
+    elsif api_version_major == 1 && api_version_minor <= 19
+      options.merge!(
+        {
+          'Volumes' => {
+            '/opt/samson_build_cache' => {}
+          },
+          'HostConfig' => {
+            'Binds' => ["#{DOCKER_HOST_CACHE_DIR}:#{CONTAINER_CACHE_DIR}"],
+            'NetworkMode' => 'host'
+          }
+        }
+      )
+    else
+      options.merge!(
+        {
+          'Mounts' => [
+            {
+              'Source' => DOCKER_HOST_CACHE_DIR,
+              'Destination' => CONTAINER_CACHE_DIR,
+              'Mode' => 'rw,Z',
+              'RW' => true
+            }
+          ],
+          'HostConfig' => {
+            'NetworkMode' => 'host'
+          }
+        }
+      )
+    end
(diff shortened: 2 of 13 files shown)


plugins/docker_binary_builder/lib/samson_docker_binary_builder/samson_plugin.rb
```diff
@@ -0,0 +1,66 @@
...
+  # end
+  container = Docker::Container.create(options)
+
+  # Execute the commands
+  output.puts 'Now starting Build container...'
+  container.tap(&:start).attach { |stream, chunk| output.write "#{stream}: #{chunk}" }
+
+  # Copy artifacts out and untar them to cwd
+  output.puts 'Grabbing binaries from build...'
+  artifacts_tar = File.open(File.join(dir, 'artifacts.tar.tar'), 'wb+')
```
sandlerr: .tar.tar?

plugins/docker_binary_builder/lib/samson_docker_binary_builder/samson_plugin.rb
```diff
@@ -0,0 +1,66 @@
+module SamsonDockerBinaryBuilder
+  class Engine < Rails::Engine
+  end
+end
+
+def build_docker_binaries(dir, project, reference, output)
+  output.puts 'Now building the build container...'
+  Docker::Image.build_from_dir(dir, { 'dockerfile' => 'Dockerfile.build', 't' => "#{project.send(:permalink_base)}_build:#{reference}" })
```
sandlerr: I guess everything should be conditional on this dockerfile existing?

henders: Yup, code is a mess, was going to tidy it up :)

And yea, .tar.tar is a result of container.copy() automatically tar'ing it
for you! :/

henders:
Ready for comments now @sandlerr  :smiley:

henders:
@zendesk/samson @grosser @jonmoter 
What do you think of this? :)

henders:
cc'ing @steved

fneves:
In the description we forgot to say that the container is booted up with a shared volume with the docker host, where we basically share a folder with that "build" container. This folder will work as cache for systems like maven and sbt (ivy is the real dependency manager) where dependencies will be stored (/<shared_folder>/.ivy2, /<shared_folder>/.my, etc...). This will speedup the build of the binaries a lot.

Local tests showed a huge improvement when using the cache, reducing the total time of building the binaries from 15/20 minutes to about 2 minutes.

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    # Mount a cache directory for sharing .m2, .ivy2, .bundler directories between build containers.
+    api_version_major, api_version_minor = docker_api_version.scan(/(\d+)\.(\d+)/).flatten.map(&:to_i)
+    if api_version_major == 0 || (api_version_major == 1 && api_version_minor <= 14)
+      fail "Unsupported Docker api version '#{docker_api_version}', use at least v1.15"
+    elsif api_version_major == 1 && api_version_minor <= 19
+      options.merge!(
+        {
+          'Volumes' => {
+            '/opt/samson_build_cache' => {}
```
jonmoter: Should you use the `DOCKER_HOST_CACHE_DIR` constant here?

plugins/docker_binary_builder/README.md
```diff
@@ -0,0 +1,63 @@
...
+- Better security in your docker images to not be able to recompile code.
+
+## Getting Started
+
+This plugin makes a few assumptions:
+- You have 2 dockerfiles:
+ - 'Dockerfile' used to generate your main deployment docker image
+ - 'Dockerfile.build' used to create a temporary image to build any binaries you want to package into the former.
+- You have a build script 'build.sh' that can be executed to compile the needed binaries
+  - This is packaged inside the Dockerfile.build image and executed
```
jonmoter: In particular, it looks like it's required that the `Dockerfile.build` creates an image with a file called `/app/build.sh`, because you have that hard-coded as the CMD when running the container.

Rather than that, how about you expect that the image created by `Dockerfile.build` specifies its own `CMD` instruction, and then when you run it as a container, you don't override the `CMD`?  That would make for one less file that needs to be named something special, and gives some more flexibility to the person writing the build image.

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -0,0 +1,132 @@
+class BinaryBuilder
+  DOCKER_FILE = 'Dockerfile.build'.freeze
```
jonmoter: Could you call this `DOCKER_BUILD_FILE` so it's more clear which Dockerfile it's referring to?  I was a little confused when I came across the constant on line 17.

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -0,0 +1,132 @@
...
+    @project = project
+    @git_reference = reference
+    @output_stream = output
+  end
+
+  def build
+    return unless @project.try(:deploy_with_docker?) && File.exists?(File.join(@dir, DOCKER_FILE))
+    @output_stream.puts "Connecting to Docker host with Api version: #{docker_api_version} ..."
+
+    @image = create_image
```
jonmoter: Rename `create_image` to `create_build_image`, to clarify which Docker image we're talking about here.

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -0,0 +1,132 @@
...
+          end
+        end
+      end
+    end
+  end
+
+  def create_container_options
+    options = {
+      'Cmd' => [BUILD_SCRIPT],
+      'Image' => "#{@project.send(:permalink_base)}_build:#{@git_reference}"
```
jonmoter: Extract the image name into a private helper method, since it's a little complicated, and you use it more than once in the file.

jonmoter:
This looks good.  It took me reading through it a few times to understand what's going on, but this seems like a pretty good solution.

The only thing I'm iffy about is whether it's a good idea to automatically execute this behavior if a file called `Dockerfile.build` is present.  I'm tempted to attach another flag to Project to explicitly enable/disable this behavior.  But it's a relatively minor quibble.

I made a few small suggestions in comments.  But overall, I like.

plugins/docker_binary_builder/README.md
henders: We originally designed it that way to have the CMD inside the dockerfile, but the docker-api wasn't telling us when the cmd finished or the exit code.
This might have been user-error, but we gave it a few tries. I can have another look now that I'm much more familiar with the library.

plugins/docker_binary_builder/app/models/binary_builder.rb
henders: Ooops! :)

plugins/docker_binary_builder/app/models/binary_builder.rb
henders: done

plugins/docker_binary_builder/app/models/binary_builder.rb
henders: done

plugins/docker_binary_builder/app/models/binary_builder.rb
henders: Yup, good point.

plugins/docker_binary_builder/README.md
jonmoter: I don't know about the API itself, but if you call `docker inspect` on a container, you can find out whether it's still running, and what its exit code is.  I *believe* that's the exit code of the CMD that is run.

```
$ docker inspect MYSQL_DATA_NEW | jq '.[0].State'
{
  "Running": false,
  "Paused": false,
  "Restarting": false,
  "OOMKilled": false,
  "Dead": false,
  "Pid": 0,
  "ExitCode": 0,
  "Error": "",
  "StartedAt": "2015-08-28T22:58:21.370347184Z",
  "FinishedAt": "2015-08-28T22:58:21.406439187Z"
}
```

I assume the same information can be retrieved via the API.

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -0,0 +1,136 @@
+class BinaryBuilder
+  DOCKER_BUILD_FILE = 'Dockerfile.build'.freeze
+  BUILD_SCRIPT = '/app/build.sh'.freeze
+  ARTIFACTS_FILE = 'artifacts.tar'.freeze
+  ARTIFACTS_FILE_PATH = "/app/#{ARTIFACTS_FILE}".freeze
+  DOCKER_HOST_CACHE_DIR = '/opt/samson_build_cache'.freeze
```
steved: Should this be project specific?

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def retrieve_binaries
+    @output_stream.puts "Grabbing '#{ARTIFACTS_FILE_PATH}' from build container..."
+    artifacts_tar = Tempfile.new(['artifacts', '.tar'], @dir)
+    artifacts_tar.binmode
+    @container.copy(ARTIFACTS_FILE_PATH) { |chunk| artifacts_tar.write chunk }
+    artifacts_tar.close
+
+    untar(artifacts_tar.path)
```
steved: Do we want to 'touch' these files? Usually these are the ones that change the most anyway, but it could theoretically be a problem since we're recreating them each time.

steved:
Definitely interested to test this out as I've been using separate stages.

plugins/docker_binary_builder/app/models/binary_builder.rb
henders: We intentionally made it shared so that all Maven/SBT/etc.. based projects would benefit from each other's cache.

plugins/docker_binary_builder/app/models/binary_builder.rb
henders: Not sure what you mean. The binaries have just been compiled and packaged into the 'artifacts.tar' inside the build image. We're just copying them out, and since they don't already exist each time we build this project there isn't any 'touch'ing as they are new files each time.

steved: Given a deterministic build process with multiple artifacts, since the mtimes are different after un-tar, docker wouldn't be able to cache steps. Really not a big deal, probably nothing in practice.

plugins/docker_binary_builder/app/models/binary_builder.rb
steved: I'm a little wary. Sure the first setup would be slow, but all subsequent ones would still be faster. But it's optional anyway, so totally fine.

fneves: @steved, this is was granted us huge improvements in the build phase for our Scala project.
Maven and SBT(ivy) take a huge amount of time to download all the dependencies. We reduced our build phase to at most 2 minutes, instead of the 15, 20 minutes.

I tried using the layering cache that docker offers by adding the build file (build.scala), first and running a fetch dependencies command. However the build file changes too often causing the cache not to work on most of the times (causing the build of an image to take 15, 20 minutes,...painful).

plugins/docker_binary_builder/app/models/binary_builder.rb
fneves: It seems that that behaviour changed on docker 1.8.0, which makes the cache work even if the mtime changed: [Docker PR#12031](https://github.com/docker/docker/pull/12031)

steved: nice!

plugins/docker_binary_builder/app/models/binary_builder.rb
steved: I think understand, but it would still be cached per-project. So the first dependency download of an entirely new project would be very slow, but every one after would be just as fast. I'm just worried about unrelated projects clobbering each other.

jonmoter: I'm with @steved on this one.  Having a single cache directory that all projects share sounds like it could cause cross-project contamination. Like if you ran build scripts that compiled a bunch of .java files into .class files, and then added `*.class` into your archive, but whoops, that just included a bunch of object files from some completely different project.

+1 on having a cache directory per project.  (Ideally, there would be some mechanism to clean up that directory, but we can make that an enhancement later.)

fneves: Ok, so the main idea of this was to allow projects to share the .m2, .ivy to the projects. If we would expose these as ENV variables like M2_HOME or IVY_HOME when we start the container, the tools will automatically pickup where the cache is, and developer won't even need to know that they are being cached. This will be completely transparent to him.

As a big advantage we can think that a lot of JVM apps share lots of common dependencies (logging, etc..) so less space used than one directory per project, less probability of running out of space in the DockerHost machine.

As a disadvantage we can't protect against what @jonmoter said, i.e, a bad intentioned developer. As with this cache approach everything should be completely transparent to the developer, only in an intentioned way a developer could do it. So the problem still exists but with a lot smaller probability.

Creating a project specific directory with cached dependencies will handle the "security" problem, sure, but the disk space might become an issue.

jonmoter: How much disk space are we talking about per project?

I'm not so concerned about a developer with bad intentions. More like a developer who writes build scripts that work fine on their local instance, but starts breaking when running on Samson, with a dozen other projects using the same cache directory.  Or their build scripts start causing some other project to accidentally have problems.

This functionality should work with any project that can be compiled, right?  So it might work totally fine for Java-based projects using .m2 and .ivy.  But maybe it will cause problems when compiling code for Go, or C++, or some other language.

Also, what happens when Samson is running two builds at the same time that are both reading and writing files into the same directory?  That seems like another recipe for problems.

The way I see it, disk space is cheap.  I don't want people on the Ops or Infra teams spending a bunch of time hunting down deploy problems caused by this.

That being said, it would be a pretty simple change if we did find this was causing problems.  So I'm willing to give it a try.  But the first time we have a build problem that is caused by the shared directory, I'm going to refactor it.  :-)

jonmoter:
:+1:

PR #540, opened 2015-08-22
send the job's commit to jenkins as a param

anthonywoo:
We will need to know the `commit` (sha) of the deploy to know what version of the application to build on jenkins. We can't rely on the deploy's `reference` as sometimes it just sends `master` but that can be changing.

@rdhanoa


sandlerr:
:+1:

rdhanoa:
:+1:

PR #541, opened 2015-08-26
Add Zendesk Samson webhook hidden inside travis ENV var.

henders:
This is to see if Webhooks can be hidden inside env vars so that external users don't see our Samson server URL.

/cc @zendesk/runway 

### References
 - Jira link:

### Risks
 - None


henders:
Closing this, as we have a github hook. Should figure out why that isn't working instead of doing this.

PR #542, opened 2015-08-27
Stop waiting 5 secs for dangling threads to finish

henders:
Rather than timing out after 5 seconds if there are dangling threads for a test, just fail out immediately and kill them.

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -60,17 +60,26 @@ class ActiveSupport::TestCase
   fixtures :all
 
   before do
+    @before_threads = Thread.list
     Rails.cache.clear
     create_default_stubs
   end
 
-  after { sleep 0.1 while extra_threads.present? }
+  after { fail_if_dangling_threads }
+
+  def fail_if_dangling_threads
+    max_threads = ENV['CI'] ? 0 : 1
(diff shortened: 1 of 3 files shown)


test/test_helper.rb
```diff
@@ -60,17 +60,26 @@ class ActiveSupport::TestCase
...
+    @before_threads = Thread.list
     Rails.cache.clear
     create_default_stubs
   end
 
-  after { sleep 0.1 while extra_threads.present? }
+  after { fail_if_dangling_threads }
+
+  def fail_if_dangling_threads
+    max_threads = ENV['CI'] ? 0 : 1
```
grosser: that means each test can add 1 more thread like `3,4,5,6,7` vs being capped at 4 before

grosser:
I like the kill vs wait, but let's not let them accumulate :)

henders:
@grosser 
Not sure how they can accumulate, that's the job of the ensure block to 'kill_extra_threads' before moving onto next thread!

grosser:
hmm true .. but just because you kill something does not means it's dead 
tumblr_mcrhl3c0dq1rf67slo1_r1_500

:+1:

henders:
True, but the .kill.join combination should mitigate those dreaded zombies somewhat. :)

PR #544, opened 2015-09-02
make sure to lock on git setup for Build

steved:
@zendesk/samson 

Why are we creating build records without Docker deploys?
https://github.com/zendesk/samson/blob/master/app/controllers/integrations/base_controller.rb#L8

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -68,7 +68,9 @@ class Build < ActiveRecord::Base
 
     return if errors.include?(:git_ref) || errors.include?(:git_sha)
 
-    project.repository.setup_local_cache!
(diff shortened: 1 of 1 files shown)


steved:
This is also just a patch, maybe all of this has to be inside GitRepository with a java-like synchronized flag.

jonmoter:
Ultimately, I'd like the `Build` model to replace the `Release` model, since it has some more information (like saving the both the SHA of the git commit, and the branch/tag used to refer to it.)  But I need to come up with a migration plan.  Also, since we don't yet have Samson connecting directly to a docker engine, we can't yet trigger automatic builds.

:+1: on the fix though.  Thanks!

PR #545, opened 2015-09-03
add an incrementing per-project number to Build

jonmoter:
akin to what we already have in the Release model.
Eventually I'd like to remove the "Release" model, migrate data over, and just use Builds.

/cc @zendesk/samson 

### Risks
 - Minimal risk.  New column added to DB. UPDATE query done on migration, but on a small table.

### DIFF
--- a/db/migrate/20150903201829_add_number_to_builds.rb
+++ b/db/migrate/20150903201829_add_number_to_builds.rb
@@ -1,0 +1,9 @@
+class AddNumberToBuilds < ActiveRecord::Migration
+  def change
+    change_table :builds do |t|
+      t.integer :number, after: :project_id
+    end
+
+    Release.joins(:build).update_all('builds.number = releases.number')
+  end
+end
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -15,6 +15,8 @@ class Build < ActiveRecord::Base
 
   validate :validate_git_reference, on: :create
 
+  before_create :assign_number
+
   def nice_name
     "Build #{label.presence || id}"
   end
@@ -84,5 +86,10 @@ class Build < ActiveRecord::Base
         errors.add(:git_sha, 'is not a valid SHA for this project')
       end
     end
+  end
+
+  def assign_number
+    biggest_number = project.builds.maximum(:number) || 0
+    self.number = biggest_number + 1
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150720121725) do
+ActiveRecord::Schema.define(version: 20150903201829) do
 
   create_table "build_statuses", force: :cascade do |t|
     t.integer  "build_id",                                     null: false
@@ -28,6 +28,7 @@ ActiveRecord::Schema.define(version: 20150720121725) do
 
   create_table "builds", force: :cascade do |t|
     t.integer  "project_id",                       null: false
+    t.integer  "number",              limit: 4
     t.string   "git_sha",             limit: 255
     t.string   "git_ref",             limit: 255
     t.string   "docker_image_id",     limit: 255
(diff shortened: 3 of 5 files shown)


sandlerr:
don't the databases themselves provide sequences? do we need a uniqueness constraint?

jonmoter:
Ideally we'd create an index on `[project_id, number]` like we do with the `releases` table.  But since we've already created a bunch of `Build` records in production, I don't want to add that index, and then have it fail because there are a bunch of records without valid values.

I was thinking of making this a two step process, where we merge and deploy this change, it runs the `UPDATE` command to set the `number` of a bunch of Builds, and then have a separate PR that cleans up the rest and creates the unique index.

How's that sound?

sandlerr:
sounds good. :+1:

dasch:
Can we wait with touching Release until you have 100% parity? A bunch of us are basing our deployments around Release at the moment, we'd rather not have it break :-)

db/migrate/20150903201829_add_number_to_builds.rb
```diff
@@ -0,0 +1,9 @@
+class AddNumberToBuilds < ActiveRecord::Migration
+  def change
+    change_table :builds do |t|
+      t.integer :number, after: :project_id
+    end
+
+    Release.joins(:build).update_all('builds.number = releases.number')
```
henders: This doesn't seem to work on Sqlite:

```
➜  samson git:(master) be rake db:migrate
== 20150903201829 AddNumberToBuilds: migrating ================================
-- change_table(:builds)
   -> 0.0015s
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

SQLite3::SQLException: near ".": syntax error: UPDATE "releases" SET builds.number = releases.number WHERE "releases"."id" IN (SELECT "releases"."id" FROM "releases" INNER JOIN "builds" ON "builds"."id" = "releases"."build_id")/Users/shender/Code/zendesk/samson/db/migrate/20150903201829_add_number_to_builds.rb:7:in `change'
ActiveRecord::StatementInvalid: SQLite3::SQLException: near ".": syntax error: UPDATE "releases" SET builds.number = releases.number WHERE "releases"."id" IN (SELECT "releases"."id" FROM "releases" INNER JOIN "builds" ON "builds"."id" = "releases"."build_id")
/Users/shender/Code/zendesk/samson/db/migrate/20150903201829_add_number_to_builds.rb:7:in `change'
SQLite3::SQLException: near ".": syntax error
/Users/shender/Code/zendesk/samson/db/migrate/20150903201829_add_number_to_builds.rb:7:in `change'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

jonmoter:
@dasch yes, that's the plan.  I won't touch the `Release` model until I:

1. Have feature parity in the `Build` model
2. Check in with the teams that are creating Releases, and make sure they're okay with the change.

dasch:
What's the different between Relase and Build, btw? Couldn't we just rename Release to Build and keep the existing data?

jonmoter:
I wanted to create a record that maps a git commit sha to a docker image digest, and saved both the commit and the reference (branch, tag, etc).  I thought the `Release` model didn't necessarily save the commit, but now that I'm looking again, it looks like it does.

Honestly, now that I review my thinking, it probably makes more sense to add the docker information to the `Release` object, and maybe rename it to `Build`, 'cause I think that's a better name.  Then I wouldn't have to bother transferring all that information over.

dasch:
That may be a better idea. One problem with Release though is that it's created in response to a CI webhook rather than a GitHub event – if I had the chance, I'd go back and change that, as well as use the GitHub Status API events to set the CI status of the Release. Maybe it's easier to start from scratch with Build. Either way, conceptually they're the same (but Build is a better name). Note that the current "releases" tab should only show builds for the master branch, though.

As for the GitHub stuff: you can programatically add GH webhooks, which is really nice. It doesn't seem like our CI providers allow that, which makes setting it up somewhat manual and error prone. In Starfish I just use the GitHub webhooks.

PR #546, opened 2015-09-04
allow selective disabling of plugins via ENV variable

jonmoter:
Adds the capability to say "all plugins except A, B, and C" via a syntax like:

`PLUGINS="all,-slack,-env"`

/cc @zendesk/samson

### Risks
 - Minimal. No change to current behavior. Only adds new functionality.


sandlerr:
:+1: on :green_apple:

PR #547, opened 2015-09-04
Revert "add an incrementing per-project number to Build"

henders:
Reverts zendesk/samson#545

Still confused about the intent with the 1-to-many relationship with the 'releases' table. So just reverting it since it breaks SQLite migrations.

Describe your PR

/cc @zendesk/samson @zendesk/runway @jonmoter 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -15,8 +15,6 @@ class Build < ActiveRecord::Base
 
   validate :validate_git_reference, on: :create
 
-  before_create :assign_number
-
   def nice_name
     "Build #{label.presence || id}"
   end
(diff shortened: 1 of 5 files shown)


grosser:
no idea about that ... who wrote it ?

grosser:
ah .. Jon :)

jonmoter:
How about I just remove the `UPDATE` statement from the migration?  That's the thing that's breaking it.

(and what's up with SQLite not being able to handle that statement?)

henders:
Yup, remove the update for now. We can figure it out later..... or wait till Sqlite can understand it! :)

jonmoter:
I opened PR #548 to remove the UPDATE line from the migration.

henders:
Cool, I'll close this.

PR #548, opened 2015-09-04
remove UPDATE from migration that was breaking on SQLite

jonmoter:
reverts problem introduced in PR #545.

/cc @henders, @grosser


henders:
:+1:

PR #551, opened 2015-09-21
Allow viewer to access project overview

msufa:
Currently the Viewer is redirected straight to the Deploys page when opening a project. This is done without any notification, while when clicking on other restricted tabs the user gets a popup informing that they're not authorised to view the page.

This PR lifts that restriction. The "Deploy" buttons will be hidden from Viewer in both the Overview and Releases pages.

/cc @henders @fneves @grosser 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/app/views/projects/_stage.html.erb
+++ b/app/views/projects/_stage.html.erb
@@ -16,6 +16,10 @@
     <% else %>
       <td>-</td>
     <% end %>
-    <td align="right"><%= deploy_link @project, stage %></td>
+    <td align="right">
+      <% if current_user.is_deployer? %>
+        <%= deploy_link @project, stage %>
+      <% end %>
+    </td>
   </tr>
 <% end %>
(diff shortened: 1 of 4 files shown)


app/views/projects/_stage.html.erb
```diff
@@ -16,6 +16,10 @@
     <% else %>
       <td>-</td>
     <% end %>
-    <td align="right"><%= deploy_link @project, stage %></td>
+    <td align="right">
+      <% if current_user.is_deployer? %>
```
grosser: is is_deployer also true for admins ?

test/controllers/projects_controller_test.rb
```diff
@@ -238,9 +238,9 @@
 
   describe "a GET to #show" do
     as_a_viewer do
-      it "redirects to the deploys page" do
+      it "does not redirect to the deploys page" do
```
grosser: `it 'renders'` ;)

grosser:
I can't see any secret information on the overview page, so I'd say :+1:

app/views/projects/_stage.html.erb
msufa: Yup, is_deployer is true for every user with role_id above viewer.

PR #553, opened 2015-09-22
Create new "project admin" role [SAMSON-169]

sbrnunes:
Continuing the work started by @jonmoter: adding ability to grant users admin-level permissions to a single project, rather than globally.

Reviewed the database model (renamed table ProjectRole to UserProjectRole), created ProjectRole "enum" as a first class citizen, containing all project-level roles.

Reviewed the authorisation "filters" when accessing project related controllers:
- System level admins / super-admins can still access the projects
- Non system level admins can now access a project if assigned the role of project admin

Added support for changing users project-level roles through the UI:
- By drilling down to a specific User details page from the Users page accessible through the Admin menu (accessible to system-level admins and super-admins)
- By clicking on the new Users tab on the Project settings page (accessible only to project admins)

/cc @zendesk/samson @jonmoter @henders @msufa 

### Steps to reproduce
- Login as a deployer / viewer
- Confirm that the User does not have admin rights for a specific project
- Login as an admin / super-admin, and make the previous User admin for the previous project
- Login again with the first User
- Confirm that the User has now admin rights for the same project, can see the Users tab and change the project-level roles for another User

### Tasks
 - [x] UI to show/add/remove user permissions on Project Settings page (added by Jon)
 - [x] UI to show/add/remove project permissions on Users admin page (added by Jon)
 - [ ] Review failing tests and create new tests for new classes

### References

### Risks

### Screenshots
- User details page in the Admin section
screenshot 2015-09-22 16 52 51

- User role just updated
screenshot 2015-09-22 16 53 26

- Users tab in the Project settings page
screenshot 2015-09-22 16 54 05


henders:
Want to add some pics as well so the others can see what I saw?

sbrnunes:
I just uploaded a few screenshots from the UI changes.

Thanks for the heads up.

jonmoter:
In the "Project Level Roles" screenshot, I see it lists the users that have Deployer or Admin permissions to that project, and you can change their level.  But I don't see any way on that page to add a new user, or remove an existing user with permissions.  Is there some other UI that allows that?

app/views/stages/index.html.erb
```diff
@@ -1,13 +1,13 @@
...
-  <ul id="stages" data-url="<%= reorder_project_stages_path(@project) %>" data-sortable="<%= current_user.is_admin? %>">
+  <ul id="stages" data-url="<%= reorder_project_stages_path(@project) %>" data-sortable="<%= is_user_admin_for_project? %>">
     <% @stages.each.with_index do |stage, index| %>
-      <li class="stage-bar bar list-unstyled clearfix<%= ' sortable' if current_user.is_admin? %>" data-id="stage_id_<%= stage.id %>">
+      <li class="stage-bar bar list-unstyled clearfix<%= ' sortable' if is_user_admin_for_project? %>" data-id="stage_id_<%= stage.id %>">
         <%= link_to stage.name, [@project, stage] %>
         <%= stage_lock_icon stage %>
         <div class="btn-group pull-right">
-          <% if current_user.is_admin? %>
+          <% if current_user.is_project_admin_for?(@project) %>
```
jonmoter: Should this be `is_user_admin_for_project?` ?

db/migrate/20150914100056_create_default_project_roles.rb
```diff
@@ -0,0 +1,19 @@
...
+
+        if !UserProjectRole.exists?(project: project, user: user)
+          if user.role_id == Role::ADMIN.id
+            UserProjectRole.create!(project: project, user: user, role_id: ProjectRole::ADMIN.id)
+          elsif user.role_id == Role::DEPLOYER.id
+            UserProjectRole.create!(project: project, user: user, role_id: ProjectRole::DEPLOYER.id)
+          end
+        end
+      }
+    }
```
jonmoter: So the idea behind this migration is that we're taking every user that is currently a Deployer or an Admin, and adding an entry in the user_project_roles table for every project? What's the advantage of that?

If we want to downgrade some users, we could selectively give them UserProjectRole entries for the couple of projects they need permissions for, and then change their "baseline" role down to viewer.  But I think that would be more of a manual process, that would require looking at which projects individuals need access to.  This script would add tens of thousands of rows to the table for the instance we're running at Zendesk.

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150416231106) do
+ActiveRecord::Schema.define(version: 20150511231229) do
```
jonmoter: This schema.rb doesn't include the migrations to rename the table, or create the default project roles.

db/migrate/20150914100056_create_default_project_roles.rb
sbrnunes: You're absolutely right!

This script was the first thing I wrote, following the database model changes. Initially, I was thinking that system-level admins wouldn't have admin rights over the projects, that only super-admins and project-level admins would have it. So, my intention was to give admin rights to all projects to all current system-level admins, to avoid stripping them out of admin rights at the project level. After understanding better the requirements and how Samson works, I went back on this assumption and kept the admin rights for system-level admins at the project level. I should have come back and review this script. My bad.

Regarding the Deployer role, you're right, it makes no sense. Again, initially I was assuming that we also wanted to restrict system level deployers from being able to deploy any project, and start managing this permissions on a more granular basis. But after discussing this across the team here, I realised that a system-level Deployer should be able to deploy any project, which makes this part of the script unnecessary.

This being said, I'll remove the script. Thanks for the heads up.

db/schema.rb
sbrnunes: I had a few questions about this schema.rb, wether it should be commited or not. After double checking that with Shane, I had to clean it up because of the way how Boxen bootstraps the projects. All Zendesk projects were using the same database, which screwed up this file. All tables, from all projects, were added to this file. :-P

I'll commit the file ASAP.

sbrnunes:
At the moment, on the "Project Level Roles", we're listing all the Users, not only the Users already assigned to that project. Users not assigned to the project have both radio buttons "blanked" (example: first User on the screenshot).

This list is paginated and can be filtered by using the search bar. It was the simplest thing to do, but I agree, this is not ideal, in terms of user experience. I was discussing this with @henders yesterday, and we can improve the usability here by, first, listing the Users already assigned to the project and, second, adding a form to add a new User to the project. So, instead of listing all Users, we could provide a typeahead search field for searching for  specific Users and, then, functionality to assign this specific User(s) to the project.

@stubotnik @pswadi-zendesk : should I continue with this improvements throughout the following sprint?

app/views/stages/index.html.erb
sbrnunes: Well... we're asking this question to the User itself, so, putting "user" on the method looks a little redundant, IMHO.

What do you think about: `current_user.is_admin_for(project)`.

stubotnik:
@sbrnunes let's discuss offline

app/views/stages/index.html.erb
```diff
@@ -1,13 +1,13 @@
...
-  <ul id="stages" data-url="<%= reorder_project_stages_path(@project) %>" data-sortable="<%= current_user.is_admin? %>">
+  <ul id="stages" data-url="<%= reorder_project_stages_path(@project) %>" data-sortable="<%= is_user_admin_for_project? %>">
     <% @stages.each.with_index do |stage, index| %>
-      <li class="stage-bar bar list-unstyled clearfix<%= ' sortable' if current_user.is_admin? %>" data-id="stage_id_<%= stage.id %>">
+      <li class="stage-bar bar list-unstyled clearfix<%= ' sortable' if is_user_admin_for_project? %>" data-id="stage_id_<%= stage.id %>">
         <%= link_to stage.name, [@project, stage] %>
         <%= stage_lock_icon stage %>
         <div class="btn-group pull-right">
-          <% if current_user.is_admin? %>
+          <% if current_user.is_admin_for?(@project) %>
```
jonmoter: On lines 4 and 6 of this file, you're using the helper method `is_user_admin_for_project?`, but on this line you're saying `current_user.is_admin_for?(@project)`.  Is there a reason you don't want to use the helper method?

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+                var message = "Failed to assign role '" + roleNameFor(project_role.role_id) + "' to User " + getUserName() + " on project " + getProjectName();
+                showErrorMessage(message);
+            }
+        );
+    }
+
+    function getUniqueId()
+    {
+        return getAttributeAsInt('data-id');
+    }
```
henders: I think your java roots are showing through :). I don't think there's a need for this level of indirection.

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+
+    function showSuccessMessage(message) {
+        messageCenterService.add('success', message);
+    }
+
+    function showErrorMessage(message) {
+        messageCenterService.add('danger', message);
+    }
+});
```
henders: missing newline at eof.  There is a rubymine setting to automatically ensure this is there.

app/assets/javascripts/services/project_roles_svc.js
```diff
@@ -0,0 +1,28 @@
+samson.service('projectRolesService', function ($http) {
```
henders: Rubymine's default JS style to add spaces before parenthesis doesn't match the rest of our codebase.

app/assets/javascripts/services/project_roles_svc.js
```diff
@@ -0,0 +1,28 @@
...
+    };
+
+    this.createProjectRole = function (project_role) {
+        var data = {
+            project_role: {
+                user_id: project_role.user_id,
+                project_id: project_role.project_id,
+                role_id: project_role.role_id
+            }
+        };
```
henders: why not just
```
var data = {
  project_role: project_role
};
```
Sending the local uniqueID doesn't seem too bad a thing for more readable code.

app/assets/stylesheets/admin/users.scss
```diff
@@ -26,3 +43,6 @@
 .save-failure {
   color: $danger;
 }
+
+
+
```
henders: 1 empty line is enough :)

app/assets/stylesheets/projects.scss
```diff
@@ -134,3 +134,44 @@ pre.pre-inline {
...
+
+.editable-role .editable-input {
+  display: table-cell;
+  vertical-align: middle;
+}
+
+.editable-role .editable-input label {
+  padding: 0;
+  margin: 0 15px 0 0;
+}
```
henders: missing EOL mark.

app/controllers/concerns/authorization.rb
```diff
@@ -2,26 +2,25 @@ module Authorization
...
 
-  def unauthorized!
-    # Eventually to UnauthorizedController
-    throw(:warden)
+  def authorize_super_admin!
+    unauthorized! unless current_user.is_super_admin?
+  end
+
+  def authorize_admin!
+    unauthorized! unless current_user.is_admin?
```
henders: Re-ordering lines of code makes the diff harder because Git can't figure out what actually got added and not just moved. 

In this case, I think you just removed the 2 helper_methods?

app/controllers/builds_controller.rb
```diff
@@ -1,6 +1,13 @@
 class BuildsController < ApplicationController
-  before_action :authorize_deployer!
-  before_action :find_project
+  include CurrentProject
```
henders: You're already including this module in the ProjectLevelAuthorization below.

app/controllers/builds_controller.rb
```diff
@@ -1,6 +1,13 @@
...
+  include ProjectLevelAuthorization
+
+  before_action do
+    find_project(params[:project_id])
+  end
```
henders: I would move this fully into the CurrentProject module so it's not cut-pasted in all the controllers. Ruby modules have access to the class instance variables, so you can access the 'params' inside it.

app/controllers/concerns/current_project.rb
```diff
@@ -0,0 +1,18 @@
+module CurrentProject
+  extend ActiveSupport::Concern
+
+  included do
+    helper_method :current_project
+  end
+
+  def current_project
+    @project
+  end
```
henders: Functionless methods like this seem wrong to me.
```
def current_project
  @project ||= Project.find_by_param!(params[:project_id])
end
```
And then delete the ```find_project``` below. And delete the before_action find_project everywhere, as it'll only try to find it when a controller tries to access it.

app/controllers/locks_controller.rb
```diff
@@ -1,32 +1,51 @@
...
+  end
+
   before_action :authorize_admin!, if: :for_global_lock?
 
   def create
     attributes = params.require(:lock).
-      permit(:description, :stage_id, :warning).
-      merge(user: current_user)
+        permit(:description, :stage_id, :warning).
+        merge(user: current_user)
```
henders: Oh yea, change rubymine to have only 2 space indentation for all types as well :)

app/controllers/macros_controller.rb
```diff
@@ -1,8 +1,12 @@
 class MacrosController < ApplicationController
-  before_action :authorize_deployer!
-  before_action :authorize_admin!, only: [:new, :create, :edit, :update, :destroy]
+  include CurrentProject
+  include ProjectLevelAuthorization
```
henders: CurrentProject already included in ProjectLevelAuthorization

app/controllers/projects_controller.rb
```diff
@@ -92,9 +97,7 @@ def project_params
   end
 
   def project
-    @project ||= Project.find_by_param!(params[:id]).tap do |project|
-      project.current_user = current_user
-    end
+    @project
```
henders: Instead of doing this, just make it use the ```current_project``` or ```@project``` elsewhere in the code.

app/controllers/locks_controller.rb
```diff
@@ -1,32 +1,51 @@
...
 
   protected
 
   def for_global_lock?
     case action_name
-    when "create" then !params[:lock].try(:[], :stage_id)
-    when "destroy" then !lock.stage_id
-    else
-      raise "Unsupported action"
+      when 'create' then
```
fneves: Indent `when` as deep as `case`.

app/controllers/locks_controller.rb
```diff
@@ -1,32 +1,51 @@
...
     end
   end
 
   def lock
     @lock ||= Lock.find(params[:id])
   end
+
+  def find_project
+    case action_name
+      when 'create' then
```
fneves: Indent `when` as deep as `case`.

app/controllers/macros_controller.rb
```diff
@@ -68,10 +72,6 @@ def command_params
     params.require(:commands).permit(ids: [])
   end
 
-  def find_project
-    @project = Project.find_by_param!(params[:project_id])
-  end
-
   def find_macro
     @macro = @project.macros.find(params[:id])
```
fneves: What about if (assuming there is a method that does the same for project)
```ruby
def macro
 @macro ||= project.macros.find(params[:id])
end
```
and remove from the before action?

app/controllers/macros_controller.rb
```diff
@@ -1,8 +1,12 @@
 class MacrosController < ApplicationController
-  before_action :authorize_deployer!
-  before_action :authorize_admin!, only: [:new, :create, :edit, :update, :destroy]
+  include CurrentProject
+  include ProjectLevelAuthorization
 
-  before_action :find_project
+  before_action do
+    find_project(params[:project_id])
```
fneves: Replace this method with a lazy method for project

```ruby
def project
  @project ||= Project.find(params[:project_id])
end
```

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: Yeah... I probably went too far with the getters. Was just trying to keep the actual code simpler and cleaner.

I removed all getters except for `getUserName()` and `getProjectName()`. Each one of this methods is used in several places, so, I think it's cleaner to keep the logic responsible for getting the corresponding attribute values in a separate utility method rather than having it scattered across the code.

Are you guys ok with that?

fneves: What if instead of having all these getters, you just have a method called `getProject` that returns a structure with nicely defined attributes that you can access with the `.attribute` notation?

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+    function loadProjectRoles() {
+        projectRolesService.loadProjectRoles().then(
+            function (response) {
+                $scope.roles = response.data;
+            }
+        );
+    }
+
+    function exists(project_role) {
+        return project_role.id;
```
fneves: Exists makes me think of a boolean attribute. It does not look like `project_role.id` will return a boolean

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+            user_id: getUserId(),
+            project_id: getProjectId(),
+            role_id: getRoleId()
+        };
+
+        loadProjectRoles();
+    };
+
+    $scope.$watch('project_role.role_id', function (new_role_value, old_role_value) {
+        if (new_role_value !== old_role_value) {
```
fneves: This code seems a bit strange. The function will only return a value on this branch? What will happen if the `new_role_value === old_role_value`? It should return something no?

app/assets/javascripts/services/project_roles_svc.js
sbrnunes: That was my initial attempt, but I was having problems because, in the ruby controller, I only permit the following parameters: `user_id`, `project_id` and `role_id`. That's why I needed to create a different payload for this operation.

But you're right, I also didn't like this. The problem is when I initialise the JS controller: I always create this JSON object with the `id` attribute, which will contain the `id` of the association or `undefined` if it hasn't been created yet. 

I'm going to create an Angular factory to tackle this by building a proper Javascript object representing this data (ProjectRole) which, in it's turn, will know how to build the payload for a create and an update operation. I think I should have done this from the beginning. A reference for what I mean with this: https://medium.com/opinionated-angularjs/angular-model-objects-with-javascript-classes-2e6a067c73bc

Are you guys ok with that?

app/controllers/builds_controller.rb
sbrnunes: The params vary depending on the controller. We'll need `params[:project_id]` or `params[:id]`, depending wether we're executing an action against a resource belonging to a project or against the project itself. 

I agree with your solution below, to move this logic into the `CurrenProject#current_project` method, and that's what I'd like to use if not for this particular problem around the different parameters and all the effort from having to change all the ERB files.

sbrnunes: Just for everyone's information, I went ahead with the changes suggested and created `CurrentProject#current_project` as follows:
```
def current_project
    @project ||= Project.find_by_param!(params[:project_id]) if params.try(:[], :project_id)
  end
```
And updated all the files where we needed to change from using `@project` to start using this helper_method.

app/controllers/concerns/authorization.rb
sbrnunes: Yeah... this was a result of some initial hesitation around adding here the project level authorisation helper methods or creating a new module.

Answering your question, I just removed the authorise_deployer helper method (no longer needed).

I'm going to revert the file form master and apply the same changes again. Easy enough.

app/controllers/concerns/current_project.rb
sbrnunes: As I replied above, I also like this approach. Went ahead with the refactoring that was needed.

app/controllers/locks_controller.rb
sbrnunes: Done! ;)

Had the Continuation Indent set to 4 for Ruby files. Is there a common, shared, settings file to apply to the IDE?

app/controllers/locks_controller.rb
sbrnunes: Done!

app/controllers/locks_controller.rb
sbrnunes: Done!

app/controllers/macros_controller.rb
sbrnunes: Done as part of the refactoring discussed with Shane.

app/controllers/macros_controller.rb
sbrnunes: I didn't touch the macro. It was already there. ;)

Still, this type of things, do you guys go ahead and refactor it when you guys see something that could be improved?

app/controllers/projects_controller.rb
sbrnunes: After the refactoring following what we've discussed, I changed this method to return `current_project`

app/assets/javascripts/controllers/project_roles_ctrl.js
henders: This is a strange function anyway, I don't think it makes the code more readable then
```
if (project_role.id) { /* do stuff */ }
```

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+        );
+    }
+
+    function exists(project_role) {
+        return project_role.id;
+    }
+
+    function createProjectRole(project_role) {
+        projectRolesService.createProjectRole(project_role).then(
+            function (response) {
```
henders: space before parens... also maybe name the anonymous functions to make it clear the first one is 'success' and 2nd one is 'error'?
Not pushed about it though as people who know the $http interface would know it anyway.

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+    function getAttributeAsString(attr_name) {
+        var attr = getAttribute(attr_name);
+        return attr.length ? attr : undefined;
+    }
+
+    function getAttribute(attr_name) {
+        return getElement().getAttribute(attr_name);
+    }
+
+    function getElement() {
```
henders: I'd dump this getter too. Not much value.

app/assets/javascripts/controllers/project_roles_ctrl.js
henders: Isn't he already doing that above during the init of the project_role?

app/assets/javascripts/services/project_roles_svc.js
henders: I'll be curious to see how it turns out! :)  As long as the code is readable, and doesn't needlessly abstract code away, then all is well.

app/controllers/macros_controller.rb
henders: Yea, stop creating extra work @fneves  :)

app/controllers/project_roles_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    render json: ProjectRole.all.map { |role| { id: role.id, display_name: role.display_name } }, root: false
+  end
+
+  def create
+    new_role = UserProjectRole.create(create_params)
+
+    if new_role.persisted?
+      user = User.find(create_params[:user_id])
+      Rails.logger.info("#{current_user.name_and_email} granted the role of #{project_role_name} to #{user.name} on project #{current_project.name}")
+      render status: :created, json: {project_role: new_role}
```
henders: inconsistent spacing with hash 2 lines down.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  before_action do
+    find_project(params[:project_id])
+  end
+
+  before_action :authorize_project_admin!
+
+  def index
+    scope = User
+    scope = scope.search(params[:search]) if params[:search]
+    @users = scope.order(sort_column + ' ' + sort_direction).page(params[:page])
```
henders: Use string interpolation instead of concatenating. ```"#{sort_column} #{sort_direction}"```

app/helpers/application_helper.rb
```diff
@@ -55,7 +55,7 @@ def datetime_to_js_ms(utc_string)
 
   def sortable(column, title = nil)
     title ||= column.titleize
-    direction = (column == sort_column && sort_direction == "asc") ? "desc" : "asc"
+    direction = (column == params[:sort] && params[:direction] == "asc") ? "desc" : "asc"
```
henders: Why did you stop using the helpers here?

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: This function is not meant to return anything. It's meant to trigger a create / update operation against the backend when the new_value doesn't match the old_value. If the values match, it will just be ignored (and they do match when you're initialising the controller, they're both 'undefined' or empty, cannot remember for sure).

Anyway, I think it's a common approach when watching for changes on the model.

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: @henders, got confused by your comment. Are you suggesting to use the $http success / error calbacks (which got deprecated) or something in the likes of:

```
projectRolesService.createProjectRole(project_role).then(
  function success(response) { ... },
  function error() { ... }
);
```

I just never saw it being used like this. I agree that it makes it more explicit, but... maybe just a comment in the first line of each function saying `//Success` or `//Error` ? :)

What you guys think?

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: Removed when I removed the other getters. Didn't notice this was being used in just one place.

app/controllers/users_controller.rb
sbrnunes: Ups, my bad. I've fixed this in some places but missed this one. Thanks for the heads up.

app/helpers/application_helper.rb
sbrnunes: This helper methods were specific to the User model, and were defined in the UsersController. So, here, we ended up with what was supposed to be a cross-application helper tight to this specific use case of sorting the Users:
```
def sort_column
    User.column_names.include?(params[:sort]) ? params[:sort] : "created_at"
  end

  def sort_direction
    %w[asc desc].include?(params[:direction]) ? params[:direction] : "asc"
  end
```
I needed this `sortable` helper method for the projects as well, so I needed to make this `sortable` method agnostic of any model. There's a tradeoff here, though: I'm assuming whoever uses this will supply the correct column names. Before, by using the helper methods, we were falling back to a default column provided by the helper method. However, for the existing scenarios, it was not a problem as we're checking the sort columns in the controller before executing the query against the database.

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: The username and project name attributes were not being stored in the project_role structure.

I'll create a proper javascript model object, `ProjectRole`, to encapsulate all this logic. I should have done it from the beginning...

sbrnunes:
Closing this pull request and creating two new smaller ones.

PR #554, opened 2015-09-28
[WIP] Request Deployer access by creating a JIRA ticket from the authorization error popup.

msufa:
Make it possible to create a JIRA ticket with a request to grant Deployer access straight from the authorization error popup in Samson.

This functionality is enabled using an environment flag (see the .env.example file). A customer key and a private RSA key is needed for the application to be able to authenticate the user in JIRA. There needs to be a corresponding Application Link configured in the JIRA instance (TODO: all necessary information and links will be added to the Extra features readme).

The flow:

The Viewer navigates to a restricted page. "Request access" link is appended to the error message.

screen shot 2015-09-28 at 16 57 18

The authentication in JIRA is handled via OAuth. Clicking the link starts the OAuth dance if the user had not previously authorized the application. If the user is not logged into JIRA a login page is displayed.

screen shot 2015-09-28 at 17 14 33

Next, the user is prompted to authorise Samson for write access.

screen shot 2015-09-28 at 17 15 10

A success popup is displayed if the ticket was created.

screen shot 2015-09-28 at 17 15 30

If the operation failed, an error popup is displayed.

screen shot 2015-09-28 at 17 37 43

TODO:
- [ ] Update Extra features readme
- [ ] Controller Unit Tests

/cc @zendesk/samson 

### References

### Risks
 - When the user is authenticated in JIRA, we can get a flood of tickets (one every time the "Request access" link is clicked).


jonmoter:
In the referenced story (RUN-99), there are some comments from Praveen and Pascal about creating JIRA issues from an email.  Which implies a design where Samson sends an email to request access. But this implementation uses the JIRA API to create a story directly.

I think the JIRA API integration has a few downsides:
* more complicated
* only useful if someone uses JIRA, doesn't work with any other system
* requires user to auth with JIRA in order to make a request

What are your thoughts on that?

msufa:
You're right @jonmoter, I might have been a little to hasty going for the API integration. It doesn't complicate things much but it does require the user to have a JIRA account. The user needs to log into JIRA as well and authorize Samson.

We've talked with the lads here and decided that having a mailer would be much simpler and more robust. We'll go in that direction.

I will close this PR but leave the branch here for reference if anyone needs some kind of JIRA integration in the future.

PR #556, opened 2015-10-01
Use heading instead of legend in edit command form

msufa:
Across the Samson UI we consistently use headings as form titles. The new/edit command form is the only one that does this differently. This is a trivial PR which aims to fix the layout.

Before:

screen shot 2015-10-01 at 13 21 14

After:

screen shot 2015-10-01 at 13 21 44

/cc @zendesk/samson

### References
 - Jira link: N/A

### Risks
 - None


henders:
Cool, for a second I thought this affected the stage edit screen, but it doesn't! :)

:+1:

PR #558, opened 2015-10-01
Don't show stop button on job output view if no @deploy there. Otherw…

henders:
Don't show stop button on job output view if no @deploy exists. Otherwise it'll delete the project! :)

project1_deploy__running_

/cc @zendesk/samson @zendesk/runway 

### References
 - Jira link:

### Risks
 - None


msufa:
:+1:

PR #559, opened 2015-10-01
Allow creation of build with an existing git_sha

henders:
At moment it throws a 500 error in Samson if you try to create a 'build' with the same git_sha as last time.
This will just overwrite the build with the updated values as if you edited it.

This allows the docker image to be rebuilt.

/cc @zendesk/samson @zendesk/runway @fneves 

### References
 - Jira link:

### Risks
 - None


pschambacher:
:+1: LGTM

PR #560, opened 2015-10-01
Introduce project level access control [SAMSON-169]

sbrnunes:
Continuing the work started by @jonmoter: adding ability to grant users admin-level permissions to a single project, rather than globally.

Reviewed the database model (renamed table ProjectRole to UserProjectRole), created ProjectRole "enum" as a first class citizen, containing all project-level roles.

Reviewed the authorisation "filters" when accessing project related controllers:
- System level admins / super-admins can still access the projects
- Non system level admins can now access a project if assigned the role of project admin

Added support for changing users project-level roles through the UI:
- By drilling down to a specific User details page from the Users page accessible through the Admin menu (accessible to system-level admins and super-admins)
- By clicking on the new Users tab on the Project settings page (accessible only to project admins)

/cc @zendesk/samson 

### Steps to reproduce
- Login as a deployer / viewer
- Confirm that the User does not have admin rights for a specific project
- Login as an admin / super-admin, and make the previous User admin for the previous project
- Login again with the first User
- Confirm that the User has now admin rights for the same project, can see the Users tab and change the project-level roles for another User

### Tasks
 - [x] UI to show/add/remove user permissions on Project Settings page
 - [x] UI to show/add/remove project permissions on Users admin page

### References

### Risks

### Screenshots
- User details page in the Admin section
screenshot 2015-09-22 16 52 51

- User role just updated
screenshot 2015-09-22 16 53 26

- Users tab in the Project settings page
screenshot 2015-09-22 16 54 05


app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,69 @@
+samson.controller('ProjectRolesCtrl', function($scope, $element, $filter, userProjectRoleFactory, projectRoleFactory, projectRolesService, messageCenterService) {
+  $scope.project_role = {};
+  $scope.roles = [];
+
+  $scope.initModel = function() {
+    $scope.project_role = userProjectRoleFactory.buildFromDom($element[0]);
```
henders: I'd probably leave this named as 'build' since you're passing in DOM Element anyway. No need to make it that explicit, and seeing DOM made me think you were going to walk a whole tree of elements rather then just reading the data fields of that element.

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,69 @@
...
+      }
+      else {
+        return createProjectRole($scope.project_role, new_role_value);
+      }
+    }
+  });
+
+  function loadProjectRoles() {
+    projectRolesService.loadProjectRoles().then(
+      function(response) {
```
henders: I'd keep it consistent when passing in functions to methods. 
```projectRolesService.loadProjectRoles().then(function(response) {```

File: app/assets/javascripts/controllers/project_roles_ctrl.js [hunk 3]
```diff
@@ -0,0 +1,69 @@
+samson.controller('ProjectRolesCtrl', function($scope, $element, $filter, userProjectRoleFactory, projectRoleFactory, projectRolesService, messageCenterService) {
+  $scope.project_role = {};
+  $scope.roles = [];
+
+  $scope.initModel = function() {
+    $scope.project_role = userProjectRoleFactory.buildFromDom($element[0]);
+    loadProjectRoles();
+  };
+
+  $scope.$watch('project_role.role_id', function(new_role_value, old_role_value) {
```
henders: Watches can be a bit hairy in angular and can add a bit too much overhead to a page. I'm possibly a bit out of date on angular, but would a ng-change event bind on the control not be better?
```
<input type="radio" ng-model="project_role.role_id" ng-value="role.id" ng-change="roleChanged()" />
```

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,69 @@
...
+        });
+      }
+    );
+  }
+
+  function createProjectRole(project_role) {
+    projectRolesService.createProjectRole(project_role).then(
+      function() {
+        //Success
+        showSuccessMessage('User ' + project_role.user_name + ' has been granted the role ' + roleNameFor(project_role.role_id) + ' for project ' + project_role.project_name);
```
henders: Since this message is the same in both places, just make showSuccessMessage accept the project_role parameter and create the message down there.

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,69 @@
...
+      },
+      function() {
+        //Failure
+        showErrorMessage("Failed to assign role '" + roleNameFor(project_role.role_id) + "' to User " + project_role.user_name + " on project " + project_role.project_name);
```
henders: Same as above.

app/assets/javascripts/controllers/project_roles_ctrl.js
```diff
@@ -0,0 +1,69 @@
...
+      }
+    );
+  }
+
+  function roleNameFor(role_id) {
+    var filtered = $filter('filter')($scope.roles, {id: role_id});
```
henders: $filter is generally used for filtering out data for the user. Seems a bit weird to use it for searching an array of roles.

app/assets/javascripts/services/project_roles_svc.js
```diff
@@ -0,0 +1,27 @@
...
+      function() {
+        deferred.reject();
+      }
+    );
+    return deferred.promise;
+  };
+
+  this.updateProjectRole = function(project_role) {
+    return $http.put('/projects/' + project_role.project_id + '/project_roles/' + project_role.id,
+      project_role.buildUpdatePayload()
```
henders: Seems a bit needless to factor this into a separate object.

app/assets/javascripts/services/user_project_role_factory.js
```diff
@@ -0,0 +1,55 @@
...
+    return _.isNumber(this.id);
+  };
+
+  UserProjectRole.prototype.buildCreatePayload = function() {
+    return JSON.stringify(this, ['user_id', 'project_id', 'role_id']);
+  };
+
+  UserProjectRole.prototype.buildUpdatePayload = function() {
+    return JSON.stringify(this, ['role_id']);
+  };
```
henders: I don't think these 2 functions belong here, as the projectRolesService above is the one who knows what parameters it needs.

app/assets/javascripts/services/user_project_role_factory.js
```diff
@@ -0,0 +1,55 @@
...
+    var id = toInteger(element.getAttribute('data-id'));
+    var user_id = toInteger(element.getAttribute('data-user-id'));
+    var project_id = toInteger(element.getAttribute('data-project-id'));
+    var role_id = toInteger(element.getAttribute('data-role-id'));
+    var user_name = element.getAttribute('data-user-name');
+    var project_name = element.getAttribute('data-project-name');
+
+    return UserProjectRole.build(id, user_id, project_id, role_id)
+      .withUserName(user_name)
+      .withProjectName(project_name);
```
henders: This seems weird. Is UserProjectRole.build used elsewhere? If not, just make the constructor take the names as well.

app/controllers/concerns/current_project.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def current_project
+    @project
+  end
+
+  protected
+
+  def find_project(param)
+    # Will this return not found if project does not exist ?
```
henders: It will raise an ActiveRecord::RecordNotFound exception.

app/controllers/jobs_controller.rb
```diff
@@ -1,7 +1,11 @@
 class JobsController < ApplicationController
-  before_filter :authorize_admin!, only: [:new, :create, :destroy]
+  include CurrentProject
```
henders: already included in subsequent module.

app/controllers/macros_controller.rb
```diff
@@ -1,8 +1,12 @@
 class MacrosController < ApplicationController
-  before_action :authorize_deployer!
-  before_action :authorize_admin!, only: [:new, :create, :edit, :update, :destroy]
+  include CurrentProject
```
henders: already included

app/controllers/macros_controller.rb
```diff
@@ -1,8 +1,12 @@
...
+  include ProjectLevelAuthorization
 
-  before_action :find_project
+  before_action :authorize_deployer!
+  before_action do
+    find_project(params[:project_id])
+  end
```
henders: Does this not have to occur before the above authorize_deployer!? So that it can verify it's a deployer for this project?

app/controllers/project_roles_controller.rb
```diff
@@ -0,0 +1,52 @@
+class ProjectRolesController < ApplicationController
+  include CurrentProject
```
henders: already included.

app/controllers/project_roles_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def index
+    render json: ProjectRole.all.map { |role| {id: role.id, display_name: role.display_name} }, root: false
+  end
+
+  def create
+    new_role = UserProjectRole.create(create_params)
+
+    if new_role.persisted?
+      user = User.find(create_params[:user_id])
```
henders: Just use ```new_role.user```

app/controllers/project_roles_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    else
+      render status: :bad_request, json: {errors: new_role.errors.full_messages}
+    end
+  end
+
+  def update
+    project_role = UserProjectRole.update_user_role(params[:id], update_params)
+
+    if project_role.errors.empty?
+      user = User.find(project_role.user_id)
```
henders: ```new_role.user```

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: To be honest, I didn't even think about the performance side of things, as this Angular app is so small and so simple, but you're totally right. I threw it into a watch expression in the controller to keep the HTML simpler, instead of having to put the ng-change expression in each radio button, but there's no point on overloading the digest cycle with this.

I'm going to change it. Thanks for the heads up.

app/controllers/project_roles_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      Rails.logger.info("#{current_user.name_and_email} granted the role of #{project_role_name} to #{user.name} on project #{current_project.name}")
+      render status: :ok, json: project_role
+    else
+      render status: :bad_request, json: {errors: project_role.errors.full_messages}
+    end
+  end
+
+  private
+
+  def project_role_name
```
henders: Can just do ```project_role. display_name``` above instead

app/controllers/projects_controller.rb
```diff
@@ -1,8 +1,13 @@
 class ProjectsController < ApplicationController
+  include CurrentProject
```
henders: Already included

app/controllers/projects_controller.rb
```diff
@@ -92,9 +97,7 @@ def project_params
   end
 
   def project
-    @project ||= Project.find_by_param!(params[:id]).tap do |project|
-      project.current_user = current_user
-    end
+    @project
```
henders: Can instead do ```attr_reader :project``` instead of this blank method.

app/controllers/references_controller.rb
```diff
@@ -1,16 +1,13 @@
 class ReferencesController < ApplicationController
-  before_filter :find_project
-  before_filter :authorize_deployer!
+  include CurrentProject
+
+  before_action :authorize_deployer!
```
henders: Does this not have to occur below the find_project?

app/controllers/stages_controller.rb
```diff
@@ -1,13 +1,18 @@
 require 'open-uri' # needed to fetch from img.shields.io using open()
 
 class StagesController < ApplicationController
+  include CurrentProject
```
henders: already included.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,32 @@
+class UsersController < ApplicationController
+  include CurrentProject
```
henders: already included

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    @users = scope.order("#{sort_column} #{sort_direction}").page(params[:page])
+
+    respond_to do |format|
+      format.html
+      format.json { render json: @users }
+    end
+  end
+
+  def sort_column
+    User.column_names.include?(params[:sort]) ? params[:sort] : "created_at"
```
henders: hiding errors seems wrong :8ball:

app/helpers/admin/users_helper.rb
```diff
@@ -1,7 +1,13 @@
 module Admin::UsersHelper
 
-  def can_modify_roles?
-    current_user.is_super_admin?
+  def project_role_unique_id(user, project)
+    project_role = user.project_role_for(project)
+    project_role.id unless project_role.nil?
```
henders:
```
def project_role_unique_id(user, project)
  user.project_role_for(project).try(:id)
end
```

app/helpers/admin/users_helper.rb
```diff
@@ -1,7 +1,13 @@
...
+  end
+
+  def role_id_for(user, project)
```
henders:
```
def role_id_for(user, project)
  user.project_role_for(project).try(:role_id)
end
```

app/helpers/admin/users_helper.rb
henders: Actually why the 'unique' prefix? If it's from the DB, just name if 'id'

app/models/project.rb
```diff
@@ -35,6 +36,15 @@ class Project < ActiveRecord::Base
...
   }
 
+  scope :where_user_is_admin, ->(user) {
+    joins(:user_project_roles).where(user_project_roles: {
+        user_id: sanitize(user.id),
+        role_id: ProjectRole::ADMIN.id
+    })
+  }
+
+  scope :search, ->(query) { where("name like ?", "%#{query}%") }
```
henders: rename the 'query' parameter to 'name'

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: Done.

app/models/role.rb
```diff
@@ -4,10 +4,10 @@ class Role < ActiveHash::Base
...
 
   self.data = [
-    { id: 0, name: "viewer" },
-    { id: 1, name: "deployer" },
-    { id: 2, name: "admin" },
-    { id: 3, name: "super_admin" }
+    { id: 0, name: "viewer", display_name: "Viewer" },
+    { id: 1, name: "deployer", display_name: "Deployer" },
+    { id: 2, name: "admin", display_name: "Admin" },
+    { id: 3, name: "super_admin", display_name: "Super Admin" }
```
henders: Could probably have left the display_name out, and used 'role.name.humanize' when displaying it. But not too concerned.

app/models/user.rb
```diff
@@ -55,14 +58,18 @@ def gravatar_url
...
   end
 
-  Role.all.each do |role|
-    define_method "is_#{role.name}?" do
-      role_id >= role.id
-    end
+  def is_admin_for?(project)
+    project_role = user_project_roles.find_by(project: project)
+    project_role.nil? ? false : project_role.try(:is_project_admin?)
+  end
```
henders:
```
def is_admin_for?(project)
  user_project_roles.find_by(project: project).try(:is_project_admin?)
end
```

app/models/user.rb
```diff
@@ -55,14 +58,18 @@ def gravatar_url
...
+
+  def is_deployer_for?(project)
```
henders:
```
def is_deployer_for?(project)
  user_project_roles.find_by(project: project).try(: is_project_deployer?)
end
```

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: Done.

app/models/user_project_role.rb
```diff
@@ -0,0 +1,15 @@
+class UserProjectRole < ActiveRecord::Base
```
henders: I'm not sure you need to use this Join model. 
I think the nicer approach is to use the has_and_belongs_many approach: http://guides.rubyonrails.org/association_basics.html#the-has-and-belongs-to-many-association

That way the User and Project models ```has_and_belongs_to_many :ProjectRole```. And interactions seem a bit more intuitive with user.project_roles and project.project_roles.

app/views/admin/projects/_project.html.erb
```diff
@@ -3,7 +3,6 @@
   <td><%= link_to project.name, project %></td>
   <td>
     <%= link_to "Edit", edit_project_path(project) %>
-    |
```
henders: Getting rid of Delimiter? Did it look better without?

app/views/shared/_search_bar.html.erb
```diff
@@ -1,5 +1,5 @@
-<div id="user-search-bar" class="col-md-5 row pull-right">
```
henders: shared/_search_bar is a bit too generic a name, maybe 'shared/_user_search_bar'?

Because there are project search bars as well. Also, leave in the 'user-' prefix for the ID as well, since you're sharing the DOM with other partials.

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: Done.

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: Done.

db/migrate/20150924082915_create_user_project_roles.rb
```diff
@@ -0,0 +1,8 @@
+class CreateUserProjectRoles < ActiveRecord::Migration
```
henders: If you go with the has_and_belongs_to_many assoc, you'll have to rename this join table to project_role_users I think, as Rails alphabetises the tables

henders:
Overall nice job! 
Just a few bigger issues with the User/Project/ProjectRoles DB relationship, and the ```CurrentProject::current_project``` vs ```@project``` usage still seems a bit forced. I realize it was to limit the size of the PR (phew).

app/assets/javascripts/controllers/project_roles_ctrl.js
sbrnunes: Done. Much better, indeed:
```
function roleNameFor(role_id) {
    var role = _.findWhere($scope.roles, {id: role_id});
    return _.isUndefined(role) ? '' :  role.display_name;
  }
```
Would you create an Underscore mixin for this last line, to make it reusable, something like: `_.propertyOfOrDefault(obj, property, default_value)` ?

app/assets/javascripts/services/project_roles_svc.js
sbrnunes: Well, regarding this one and the next comment bellow, I think it could go both ways:
- I agree that the service is the one who knows which parameters it needs to send, but...
- The UserProjectRole is the one that actually knows which fields it supports, reason why I was asking it in each scenario :)

But in the end, I guess it makes it more explicit if, in the service, you can see which fields are being sent to the backend, so, I'll move this to the service. ;)

app/assets/javascripts/services/user_project_role_factory.js
sbrnunes: Done (following above considerations).

app/assets/javascripts/services/user_project_role_factory.js
sbrnunes: Yeah... I was trying to distinguish the mandatory attributes that represent a UserProjectRole, as it exists in the database, from the "optional" fields that are just used to display the success / error messages to the User.

I was, perhaps, overcomplicating this. Sorry about that.

app/controllers/jobs_controller.rb
sbrnunes: All this situations have been addressed in the following PR: [#567](https://github.com/zendesk/samson/pull/567)

app/controllers/macros_controller.rb
sbrnunes: All this situations have been addressed in the following PR: [#567](https://github.com/zendesk/samson/pull/567)

app/controllers/macros_controller.rb
sbrnunes: In this PR, `authorize_deployer` is only checking if the User is a system level deployer, so, it doesn't require the project being loaded. That's only required by the `authorize_project_admin`.

app/controllers/project_roles_controller.rb
sbrnunes: All this situations have been addressed in the following PR: #567

app/controllers/project_roles_controller.rb
sbrnunes: Doh! :)

app/controllers/project_roles_controller.rb
sbrnunes: You meant `project_role.role.display_name`, right? :)

Either way, done and tested.

app/controllers/projects_controller.rb
sbrnunes: As mentioned, all this situations have been addressed in the following PR: #567

app/controllers/projects_controller.rb
sbrnunes: Well spotted. Thanks.

app/controllers/references_controller.rb
sbrnunes: Yep, as I mentioned above, this is not a project-specific role check. That has been added in the PR about project-specific deployer role.

app/controllers/stages_controller.rb
sbrnunes: As mentioned, all this situations have been addressed in the following PR: #567 :)

app/controllers/users_controller.rb
sbrnunes: As mentioned, all this situations have been addressed in the following PR: #567

app/controllers/users_controller.rb
sbrnunes: Fixed.

app/helpers/admin/users_helper.rb
sbrnunes: Fair enough. :)

Fixed.

app/helpers/admin/users_helper.rb
sbrnunes: Fixed.

app/models/role.rb
sbrnunes: Damn, didn't find that function. I did search for capitalize, but then noticed the underscores and gave up.

I'll remove the `display_name` property and create a method with the same name, returning `name.humanize`. I won't have to change it in all the places but it will make the hash cleaner.

Thanks.

app/models/user.rb
sbrnunes: Done.

app/models/user.rb
sbrnunes: Done.

app/views/admin/projects/_project.html.erb
sbrnunes: I have no idea how did I delete this line. And even less how did I miss the change. 
I reverted it.

app/models/user_project_role.rb
sbrnunes: @henders : I have a few questions in here:

First one: according to the documentation, we should use `has_many` when we need an extra attribute:
> You should use has_many :through if you need validations, callbacks, or extra attributes on the join model.

Beside project and user, this user_project_role table contains a role_id column. Doesn't it prevent us from using the `has_and_belongs_to_many` relationship?

Second: after going through the documentation, I didn't fully understand your example, the part where you added `:ProjectRole` on the `has_and_belongs_to_many`. I would have understood something like `has_and_belongs_to_many :projects`, for User, and `has_and_belongs_to_many :users`, for Project. I also didn't find anything that could explain why `:ProjectRole`. Was it a typo? Just asking before I change the whole thing. :)

Finally: `ProjectRole` is not an `ActiveRecord::Base` descendant, instead, it inherits from `ActiveHash::Base`. I know ActiveHash allows us to treat a hash as a model-like entity, but doesn't it stop there? 

So, in the end, let me know if I'm doing the right thing or not. :)

One thing I noticed is that I was missing the `has_many :projects, through: :user_project_roles`, and the corresponding one for the the Project model, to get the users. It may come in handy.

app/views/shared/_search_bar.html.erb
sbrnunes: @henders, I've refactored it because I noticed that this partial could be reused everywhere, and not only for searching for users. 

But you're right about the `id`, I didn't remember we could end up with more than one search bar in the page. I removed the id, as it was only being used for styling purposes, and added a `search-bar`class to the class attribute to be able to style it from CSS.

Sounds good?

app/models/user_project_role.rb
steved: has_and_belongs_many should be avoided if possible, it locks you into some of the worst rails magic.

app/models/project.rb
sbrnunes: Done.

app/models/user_project_role.rb
sbrnunes: To sum up, following all the above considerations, I'll keep it as it is, if you're ok with that @henders.

But I'm still open to discussion so let me know if, in the meantime, you looked into this and have any further thoughts.

sbrnunes:
@henders, I've just committed all the (agreed) changes following all the above considerations, including the ability to show only the last flash message when changing the User role in a project.

sbrnunes:
@steved, @jonmoter: @henders is going to be out for a week, so would you guys be able to take a look at this PR, @henders's suggestions and corresponding fixes already committed?

app/assets/javascripts/services/project_roles_svc.js
```diff
@@ -0,0 +1,28 @@
...
+    );
+    return deferred.promise;
+  };
+
+  this.updateProjectRole = function(project_role) {
+    var payload = JSON.stringify(project_role, ['role_id']);
+    return $http.put('/projects/' + project_role.project_id + '/project_roles/' + project_role.id, payload);
+  };
+});
+
```
steved: :scissors:

app/assets/javascripts/services/project_roles_svc.js
```diff
@@ -0,0 +1,28 @@
+samson.service('projectRolesService', function($http, $q) {
+
```
steved: :scissors:

steved: Same with other file, I don't understand the extra whitespace

app/controllers/project_roles_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  include ProjectLevelAuthorization
+
+  before_action only: [:create, :update] do
+    find_project(params[:project_id])
+  end
+
+  before_action :authorize_project_admin!, only: [:create, :update]
+
+  def index
+    render json: ProjectRole.all.map { |role| {id: role.id, display_name: role.display_name} }, root: false
```
steved: I would prefer serializers over custom json creation

app/models/concerns/has_project_role.rb
```diff
@@ -0,0 +1,13 @@
+module HasProjectRole
+  extend ActiveSupport::Concern
```
steved: no need to extend the module if you don't use it

app/models/concerns/permalinkable.rb
```diff
@@ -9,7 +9,15 @@ module Permalinkable
 
   module ClassMethods
     def find_by_param!(param)
-      find_by_permalink!(param)
+      begin
+        find_by_permalink!(param)
+      rescue ActiveRecord::RecordNotFound
+        if param =~ /^\d+$/
```
steved: Use this to branch at the top of the method, exceptions as flow control isn't good.

app/models/job.rb
```diff
@@ -38,7 +38,7 @@ def started_by?(user)
   end
 
   def can_be_stopped_by?(user)
-    started_by?(user) || user.is_admin?
+    started_by?(user) || user.is_admin? || user.is_admin_for?(project)
```
steved: Should the second check be super admin? Or removed?

app/models/user.rb
```diff
@@ -55,14 +59,16 @@ def gravatar_url
...
+    user_project_roles.find_by(project: project).try(:is_project_deployer?)
   end
 
-  def role
-    Role.find(role_id)
+  def project_role_for(project)
+    user_project_roles.find_by(project: project)
```
steved: You define this but don't use it

app/models/user_project_role.rb
```diff
@@ -0,0 +1,15 @@
...
+  include HasProjectRole
+  belongs_to :project
+  belongs_to :user
+
+  validates_presence_of :project, :user
+  validates :role_id, inclusion: { in: ProjectRole.all.map(&:id) }
```
steved: ProjectRole.pluck(:id)

app/views/admin/users/_project.html.erb
```diff
@@ -0,0 +1,11 @@
+<tr>
+  <td><%= project.id %></td>
+  <td><%= project.name %></td>
+  <td class="editable-role">
+    <%= render partial: 'shared/project_role', locals: {user: user, project: project} %>
+  </td>
+</tr>
+
```
steved: :scissors:

app/views/admin/users/show.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        </tr>
+    <% else %>
+        <%= render partial: 'project', collection: @projects, as: :project, locals: {user: @user} %>
+    <% end %>
+    </tbody>
+  </table>
+  <%= paginate @projects %>
+</section>
+
+
```
steved: :scissors:

app/views/shared/_alerts.html.erb
```diff
@@ -8,3 +8,6 @@
     <% end %>
   <% end %>
 <% end %>
+
+<!-- Flash messages for ajax requests -->
+<mc-messages></mc-messages>
```
steved: newline eof

app/views/shared/_project_role.html.erb
```diff
@@ -0,0 +1,20 @@
...
+
+    <span class="editable-input">
+        <label ng-repeat="role in roles">
+          <input type="radio" ng-model="project_role.role_id" ng-value="role.id" ng-change="roleChanged()"/>
+          <span ng-bind="role.display_name"></span>
+        </label>
+    </span>
+
+</form>
+
```
steved: :scissors:

db/migrate/20150924082915_create_user_project_roles.rb
```diff
@@ -0,0 +1,8 @@
...
+  create_table :user_project_roles do |t|
+    t.belongs_to :project, null: false, index: true
+    t.belongs_to :user,    null: false, index: true
+    t.integer :role_id,    null: false
+    t.timestamps
```
steved: These should also be `null:false`

test/models/user_test.rb
```diff
@@ -228,4 +240,40 @@
...
+
+    it "is true for a user that has been granted the role of project admin" do
+      users(:deployer_project_admin).is_deployer_for?(projects(:test)).must_equal(true)
+    end
+
+    it "is false for users that have not been granted the roles of project deployer or project admin" do
+      users(:viewer).is_deployer_for?(projects(:test)).wont_equal(true)
+      users(:deployer).is_deployer_for?(projects(:test)).wont_equal(true)
+      users(:admin).is_deployer_for?(projects(:test)).wont_equal(true)
+      users(:super_admin).is_deployer_for?(projects(:test)).wont_equal(true)
```
steved: Shouldn't super_admins have this ability?

test/test_helper.rb
```diff
@@ -148,6 +150,15 @@ def unauthorized(method, action, params = {})
         end
       end
     end
+
+    %w{deployer_project_admin}.each do |user|
```
steved: Can this be added to the block above?

test/support/timeout.rb
```diff
@@ -15,4 +15,5 @@ def capture_exceptions(*args, &block)
     end
   end
 end
-Minitest::Test.prepend TimeoutEveryTestCase
+
+Minitest::Test.prepend TimeoutEveryTestCase unless ENV['SKIP_TIMEOUT']
```
steved: Why this setting?

app/models/concerns/permalinkable.rb
sbrnunes: I only touched this method because it had a bug. The way it was implemented, we were never falling back to finding by id when the record was not found through the permalink.

But you're right. We shouldn't be using an exception in here. I'll try to review this in some other PR.

app/models/job.rb
sbrnunes: A Job can be stopped by the User that started it, a system-level admin or super-admin (covered by the is_admin? check), or by a project-level admin (covered by the is_admin_for? check).

I believe all the checks we're doing here are relevant.

app/models/user.rb
sbrnunes: You're talking about the method:  `project_role_for` ? Is being used in the `UsersHelper`.

app/models/user_project_role.rb
sbrnunes: `ProjectRole` is an `ActiveHash` (https://github.com/zilkey/active_hash), not an `ActiveRecord`. According to their documentation, `pluck` is not supported, but I'll take a look.

Anyway, good to know that method. Was not aware of what it did.

app/models/concerns/has_project_role.rb
sbrnunes: Good point. When I wrote this I was still not aware of the differences between a regular module and a module extending `ActiveSupport::Concern`, and ended up following the same example we already had for `Role`.

I'll take it out. Thanks for spotting it.

app/models/user_project_role.rb
steved: Riiight, thanks.

app/models/user.rb
steved: Ah, ok. You should use it for the `is_X_for?` methods in here.

app/models/concerns/permalinkable.rb
steved: Even an OR condition would be fine with me.

app/models/job.rb
steved: Ah, I guess I misunderstood the difference between system admin and super admin. Cool.

app/models/project.rb
```diff
@@ -35,6 +37,15 @@ class Project < ActiveRecord::Base
       alphabetical
   }
 
+  scope :where_user_is_admin, ->(user) {
+    joins(:user_project_roles).where(user_project_roles: {
+        user_id: sanitize(user.id),
```
steved: I don't think you need to sanitize this, rails should take care of it

app/models/user_project_role.rb
```diff
@@ -0,0 +1,15 @@
...
+  validates_uniqueness_of :project_id, scope: :user_id
+
+  def self.update_user_role(id, attributes)
```
steved: I kind of get the usefulness, but this is also okay in controllers:

```
role = ProjectRole.find(id)
if project_role.update(role_params)
  success
else
  fail
end
```

test/models/user_test.rb
sbrnunes: Hum, this test is a little confusing in that matter, now that I think about it.

System-level roles and Project-specific roles have been completely decoupled. The rights for deploying a project are checked in the ProjectLevelAuthorization module:
```
def authorize_project_deployer!
    unauthorized! unless current_user.is_deployer? || current_user.is_deployer_for?(current_project)
  end
```
Both system-level admins and super-admins will also be system-level deployers, which means, they will have deployer access to a project.

This being said, all I'm asserting in this test is that, regardless of the system-level role the user has, the method `User#is_deployer_for?(project)` will only return true if the User has a project-specific 'Deployer' role for that project, in the table `UserProjectRole`.

steved: Mmm yeah, now I understand. It is a *little* odd, but is more like a naming problem than behavioral.

test/support/timeout.rb
sbrnunes: Hum... something failed here in the rebase with master, this shouldn't be here. It was removed in master.

test/test_helper.rb
sbrnunes: It can. I only kept it separated because the first group of Users represent system-level roles, and this second group are meant for testing Users with project-specific roles.

I'll refactor it.

app/controllers/project_roles_controller.rb
sbrnunes: Ok, I'll take a look.

app/models/concerns/permalinkable.rb
sbrnunes: If it doesn't have a big impact I'll see what I can do. This is being used everywhere we need to load a project before project-related controller actions are executed. ;)

app/models/project.rb
sbrnunes: Ok, I'll double check it and remove it if not needed. Thanks.

app/models/user.rb
sbrnunes: Oh, true, well spotted... :)

app/models/user_project_role.rb
sbrnunes: Ok. I'll move it back to the controller if you guys prefer to see this in the controller.

app/controllers/project_roles_controller.rb
sbrnunes: Added JSON serialiser to ProjectRole and removed this custom JSON from the controller.
Thanks for the heads up.

app/models/concerns/permalinkable.rb
jonmoter: I believe it's valid for a Project to have a name (and thus a permalink) of "12345".  If you branch at the top of the method, you'll never be able to find a project with that name.  The `find_by_id!` is a fallback so that we can create urls of the form `/project/{id}/stages` as well as `/project/{permalink}/stages`.  I can't think of a way of doing that without rescuing an Exception.

jonmoter:
I think this is "close enough" to right.  @sbrnunes, once you handle the merge conflicts, it's good to go.

:+1:

PR #561, opened 2015-10-02
Request additional access rights

msufa:
When a "You are not authorized..." popup is presented to the user, we add a link to request further access rights. The link takes the user to a separate page where they can provide their manager's email and a reason for the additional capabilities. Next, depending on the configuration, an email is sent to the manager and other configured email addresses (one use case could be a JIRA email trigger that would automatically create and assign tickets to the relevant people). The next time the popup appears, the user is informed that an access request is pending. The pending state is reset when an admin updates the user's role, so that further access rights may be requested.

This is a 2nd incarnation of #554 

**The flow:**

The link to request access

screen shot 2015-10-02 at 11 03 58

The access request form

screen shot 2015-10-02 at 11 01 52

Success popup

screen shot 2015-10-02 at 11 03 11

Request pending info

screen shot 2015-10-02 at 11 03 33

Sample email content:

```
Date: Fri, 02 Oct 2015 11:45:30 +0100
From: deploys@samson-deployment.com
To: jira@example.com, 
 managers@example.com, 
 my.manager@example.com
Message-ID: <560e604abc42_24a83fe51250b7e872a8@11139-storage.mail>
Subject: [SAMSON ACCESS] Grant deployer access rights to Bill Lumbergh
Mime-Version: 1.0
Content-Type: text/plain;
 charset=UTF-8
Content-Transfer-Encoding: 7bit

Please bump access rights to deployer on http://localhost:3000 for Bill Lumbergh (bill.lumbergh@example.com)

Reason:
Deployer access needed.
```

TODO:
- [x] Unit Tests
- [x] Readme
- [x] Make sure the PR works well with #553 & #560 and optionally include the target project name/ID in the request email.

/cc @zendesk/samson

### References

### Risks
 - None


app/helpers/flash_helper.rb
```diff
@@ -1,6 +1,6 @@
 module FlashHelper
   # Flash type -> Bootstrap alert class
   def flash_mapping
-    { notice: :info, error: :danger, success: :success }
+    { notice: :info, error: :danger, authorization_error: :danger, success: :success }
```
fneves: Wouldn't it  make sense to have a `:warning` here, that could be used by other places here instead of a specific `:authorization_error`?

app/controllers/access_requests_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def new
+    session[:access_request_back_to] ||= request.referer
+  end
+
+  def create
+    RequestAccessMailer.request_access_email(
+        request.base_url, current_user, params[:manager_email], params[:reason]).deliver_now
+    current_user.access_request_pending = true
+    current_user.save
```
fneves: Use `update`?
```ruby 
  current_user.update!(access_request_pending: true)
```

app/helpers/flash_helper.rb
sbrnunes: What would happen if, then, we needed the warning for a more generic "warning" message?
Because, as far as I understood, the reason why @msufa introduced a new custom error type is because he needs to check if the error was caused by an `:authorization_error` when rendering the flash message in the view, to decide wether the link to request access should be displayed or not (see `_alerts.html.erb` below).

So, I'm curious here: is there another way, to keep the default error types provided by Rails and display the link as part of the error message only when needed, according to the conditions given by the `display_access_request_link?` helper?

The only alternative I found would be something in the likes of: http://stackoverflow.com/questions/26538891/flash-message-with-html-safe-from-the-controller-in-rails-4-safe-version

app/views/shared/_alerts.html.erb
```diff
@@ -4,6 +4,9 @@
       <div class="alert alert-<%= klass %>">
         <button type="button" class="close" data-dismiss="alert">&times;</button>
         <%= message %>
+        <% if display_access_request_link?(type) %>
+          <%= link_to_request_access %>
+        <% end %>
```
sbrnunes: If we could do something like what's described in here: http://stackoverflow.com/questions/26538891/flash-message-with-html-safe-from-the-controller-in-rails-4-safe-version, we could benefit from it a lot in this code snippet, as we could keep it completely generic, i.e., agnostic from the errors being displayed.

Still, personally, I don't like to see html being built on the controller, so, is there another, better, way to accomplish this same behaviour?

app/controllers/access_requests_controller.rb
msufa: Good point @fneves, I need to get more intimate with the model API.

app/helpers/flash_helper.rb
msufa: What @sbrnunes said :) This new flash type should be specific to only those cases where we want to display the access request link, I don't believe this should be generic functionality.

I'm not sure that the stackoverflow solution improves the situation much, however you want to proceed, in the end you still need to add a new flash type.

app/views/shared/_alerts.html.erb
msufa: It's all a matter of taste and preference I think @sbrnunes. The code is actually generated in a helper and I did this deliberately because I, on the other hand, don't like to see Ruby conditionals in templates :)

I've seen this approach in a bunch of templates in Samson so I decided to go for something familiar here as well.

Also, I think it's easier to follow what's going on when you have one dedicated flash type in this case, instead of using 2 separate ones. Did you have a specific solution in mind to keep this code generic @sbrnunes?

fneves: Yeah, I just don't like conditionals in templates as well. But I can't find a better solution. Leave it as it is.

sbrnunes: Sorry @msufa , I was actually suggesting the solution highlighted in the Stackoverflow, unless someone else (i.e., @fneves, @henders , etc.) came up with a better solution or with an opinion against that solution. :-D

The reason why I was saying that I don't like to see HTML generated in the controllers is because, usually, it's not the safest option, as we may end up with XSS issues when generating HTML in the backend. But, in this case, you're probably 100% sure of what's going to be generated and what needs to be escaped (if anything). The other reason why I tend to dislike is because HTML should be rendered in the View. :)

I think the main advantages of having the link being returned as part of the flash message, from the UnauthorizedController, is that the alerts template will remain generic and all the logic regarding what to display in this specific message, and how to display it, will be centralised in one single place and we can keep using the default error types from Rails.

msufa:
I added some helpers to the application and tests based on comments from @jonmoter.

I decided to handle the case where the email addresses are not configured, so that it is possible to use the new feature to only send emails to the manager without any other integrations/distros.

jonmoter:
Looks good.  Let's merge.  :+1:

PR #566, opened 2015-10-04
Add support for project level deployer role [SAMSON-169]

sbrnunes:
Adding support to allow an User with project role 'Deployer' to have deployer access to a project regardless of his system level role.

/cc @zendesk/samson 

### References

### Risks

### DIFF
--- a/app/controllers/concerns/current_project.rb
+++ b/app/controllers/concerns/current_project.rb
@@ -12,6 +12,7 @@ module CurrentProject
   protected
 
   def find_project(param)
+    # Will this return not found if project does not exist ?
     @project = Project.find_by_param!(param)
   end
 end
--- a/app/controllers/locks_controller.rb
+++ b/app/controllers/locks_controller.rb
@@ -1,5 +1,11 @@
 class LocksController < ApplicationController
-  before_action :authorize_deployer!
+  include ProjectLevelAuthorization
+
+  before_action unless: :for_global_lock? do
+    find_project
+    authorize_project_deployer!
+  end
+
   before_action :authorize_admin!, if: :for_global_lock?
 
   def create
@@ -7,26 +13,39 @@ class LocksController < ApplicationController
       permit(:description, :stage_id, :warning).
       merge(user: current_user)
     Lock.create!(attributes)
-    redirect_to :back, notice: "Locked"
+    redirect_to :back, notice: 'Locked'
   end
 
   def destroy
     lock.try(:soft_delete)
-    redirect_to :back, notice: "Unlocked"
+    redirect_to :back, notice: 'Unlocked'
   end
 
   protected
 
   def for_global_lock?
     case action_name
-    when "create" then !params[:lock].try(:[], :stage_id)
-    when "destroy" then !lock.stage_id
+    when 'create' then
+      !params[:lock].try(:[], :stage_id) || !params[:lock][:stage_id].present?
+    when 'destroy' then
+      !lock.stage_id
     else
-      raise "Unsupported action"
+      raise 'Unsupported action'
     end
   end
 
   def lock
     @lock ||= Lock.find(params[:id])
+  end
+
+  def find_project
+    case action_name
+    when 'create' then
+      @project = Stage.find(params[:lock][:stage_id]).project
+    when 'destroy' then
+      @project = lock.stage.project
+    else
+      raise 'Unsupported action'
+    end
   end
 end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -1,5 +1,6 @@
 class Project < ActiveRecord::Base
   include Permalinkable
+  include Searchable
 
   has_soft_deletion default_scope: true
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,6 +3,7 @@ require 'digest/md5'
 
 class User < ActiveRecord::Base
   include HasRole
+  include Searchable
 
   has_soft_deletion default_scope: true
 
@@ -60,11 +61,11 @@ class User < ActiveRecord::Base
   end
 
   def is_admin_for?(project)
-    project_role_for(project).try(:is_project_admin?)
+    project_role_for(project).try(:is_admin?)
   end
 
   def is_deployer_for?(project)
-    project_role_for(project).try(:is_project_deployer?)
+    project_role_for(project).try(:is_deployer?)
   end
 
   def project_role_for(project)
--- a/test/controllers/builds_controller_test.rb
+++ b/test/controllers/builds_controller_test.rb
@@ -111,4 +111,85 @@ describe BuildsController do
       end
     end
   end
+
+  as_a_project_deployer do
+    describe '#index' do
+      it 'works with no builds' do
+        get :index, project_id: project.to_param
+        assert_response :ok
+      end
+
+      it 'displays basic build info' do
+        project.builds.create!(label: 'test branch', git_ref: 'test_branch')
+        project.builds.create!(label: 'master branch', git_ref: 'master')
+        get :index, project_id: project.to_param
+        assert_response :ok
+        @response.body.must_include 'test branch'
+        @response.body.must_include 'master branch'
+      end
+    end
+
+    describe '#show' do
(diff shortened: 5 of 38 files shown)


app/controllers/locks_controller.rb
```diff
@@ -1,32 +1,51 @@
...
+  end
+
   before_action :authorize_admin!, if: :for_global_lock?
 
   def create
     attributes = params.require(:lock).
-      permit(:description, :stage_id, :warning).
-      merge(user: current_user)
+        permit(:description, :stage_id, :warning).
+        merge(user: current_user)
```
henders: indent too big! :)

app/controllers/locks_controller.rb
```diff
@@ -1,32 +1,51 @@
...
   protected
 
   def for_global_lock?
     case action_name
-    when "create" then !params[:lock].try(:[], :stage_id)
-    when "destroy" then !lock.stage_id
-    else
-      raise "Unsupported action"
+      when 'create' then
+        !params[:lock].try(:[], :stage_id) || params[:lock].try(:[], :stage_id).try(:empty?)
```
henders: Why did this line have to change? if params[:lock][:stage_id] is potentially an empty string then
```!params[:lock].try(:[], :stage_id).presence``` should cover it

Comment (henders) [inline app/controllers/locks_controller.rb]: also, indent less :)

File: app/models/project.rb [hunk 3]
```diff
@@ -125,6 +125,20 @@ def last_deploy_by_group(before_time)
...
   end
 
+  def self.search_by_criteria(criteria)
+    scope = Project
+    scope = scope.search(criteria[:search]) if criteria[:search]
+    scope.order("#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
+  end
+
+  def self.sort_column(column)
+    column_names.include?(column) ? column : "created_at"
```
henders: Not sure if hiding bad input is the way to go... if user wanted to sort by a field they typo'd they should probably get an error.

app/models/user.rb
```diff
@@ -60,18 +60,32 @@ def gravatar_url
...
   end
 
+  def self.search_by_criteria(criteria)
+    scope = User
+    scope = scope.search(criteria[:search]) if criteria[:search]
+    scope.order("#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
+  end
+
+  def self.sort_column(column)
+    column_names.include?(column) ? column : "created_at"
```
henders: Same as above for Project

test/controllers/builds_controller_test.rb
```diff
@@ -98,4 +98,85 @@ def stub_git_reference_check(returns: false)
       end
     end
   end
+
+  as_a_viewer_project_deployer do
```
henders: now that's a confusing name :)

app/controllers/locks_controller.rb
sbrnunes: This one was hard to remember why did I changed it. :-P
But yes, I noticed that `stage_id` is always present in the parameters, and when to provided, it comes as an empty string.

I was not aware of the `presence` method. Looks much better, thanks. I'll check it out.

app/models/project.rb
sbrnunes: So, would something like this be better:
```
def self.sort_column(column)
  column.presence || "created_at"
end
```

By default, when not provided, we need to sort by created_at.

test/controllers/builds_controller_test.rb
sbrnunes: Oh man, I know, couldn't come up with a better name. :)

Here I'd like to make it clear that this user is a User with 'Viewer' system level role, but he also has the 'Deployer' role on a specific project.

Would you prefer just `as_a_project_deployer`? Can you remember of a better name for this?

app/controllers/locks_controller.rb
sbrnunes: Fixed. Damn RubyMine code styles. :)

app/models/project.rb
```diff
@@ -126,6 +126,20 @@ def last_deploy_by_group(before_time)
     releases.map { |group_id, deploys| [ group_id, deploys.sort_by(&:updated_at).last ] }.to_h
   end
 
+  def self.search_by_criteria(criteria)
+    scope = Project
+    scope = scope.search(criteria[:search]) if criteria[:search]
+    scope.order("#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
```
steved: without the strict check against format, this is injection waiting to happen

app/models/user.rb
```diff
@@ -60,17 +60,31 @@ def gravatar_url
...
 
   def project_role_for(project)
     user_project_roles.find_by(project: project)
   end
 
+  def self.search_by_criteria(criteria)
+    scope = User
+    scope = scope.search(criteria[:search]) if criteria[:search]
+    scope.order("#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
+  end
```
steved: This probably could be in a concern `Searchable` for user and project

app/controllers/concerns/current_project.rb
```diff
@@ -12,6 +12,7 @@ def current_project
   protected
 
   def find_project(param)
+    # Will this return not found if project does not exist ?
```
steved: :scissors:

app/controllers/locks_controller.rb
```diff
@@ -7,26 +13,39 @@ def create
...
   end
 
   protected
 
   def for_global_lock?
     case action_name
-    when "create" then !params[:lock].try(:[], :stage_id)
-    when "destroy" then !lock.stage_id
+    when 'create' then
+      !params[:lock].try(:[], :stage_id) || !params[:lock].try(:[], :stage_id).presence
```
steved: presence includes nil, so you just need the second check

test/controllers/locks_controller_test.rb
```diff
@@ -1,21 +1,37 @@
...
   let(:global_lock) { Lock.create! user: users(:deployer) }
 
   before { request.headers['HTTP_REFERER'] = '/back' }
 
   as_a_viewer do
     unauthorized :post, :create
-    unauthorized :post, :create, lock: {stage_id: 1}
-    unauthorized :delete, :destroy, id: 1
+
+    it('responds with unauthorized when doing a post to create a local lock') do
```
steved: parens are inconsistent with other tests, only use them for single lines

test/controllers/locks_controller_test.rb
```diff
@@ -82,4 +98,51 @@
       end
     end
   end
+
+  as_a_project_deployer do
+    unauthorized :post, :create
+
+    it("responds with unauthorized when doing a post to create a global lock") { post :create, id: global_lock.id }
```
steved: way too long for a single line and it doesn't test anything

app/models/project.rb
sbrnunes: Damn, you're right, I missed that.

app/controllers/concerns/current_project.rb
sbrnunes: Removed in another PR.

app/controllers/locks_controller.rb
sbrnunes: I'm not worried with the nil, I'm worried with the `stage_id` being sent as an empty string even for a global lock, which I noticed was happening and the tests were not covering it.

What about: `!params[:lock].try(:[], :stage_id) || !params[:lock][:stage_id].presence`

This should return true if:
- params does not contain `stage_id` (first check)
- params contains `stage_id` but `stage_id` is an empty string (second check)

steved: Right. `params[:lock].try(:[], :stage_id).present?` will cover both of those cases.

test/controllers/locks_controller_test.rb
sbrnunes: Indeed. Something went really weird with this line.

Fixed.

PR #567, opened 2015-10-05
Refactoring filters responsible for loading a project before actions get executed

sbrnunes:
Following the code review from pull request [#553](https://github.com/zendesk/samson/pull/553), before_action filters loading the "current project" have been removed and the "current project" is now a lazily initialised attribute, accessible through the helper method `current_project`, which has been defined in the module `CurrentProject`.

/cc @zendesk/samson 

### References
Related to (or consequence of):

### Risks
- Changes to the behaviour regarding how the current project is loaded on several controllers (make sure all tests are passing and no functionality has been broken)

### DIFF
--- a/app/controllers/concerns/current_project.rb
+++ b/app/controllers/concerns/current_project.rb
@@ -2,17 +2,19 @@ module CurrentProject
   extend ActiveSupport::Concern
 
   included do
+    before_action :require_project
+
     helper_method :current_project
   end
 
   def current_project
-    @project
+    @project ||= require_project
   end
 
   protected
 
-  def find_project(param)
-    # Will this return not found if project does not exist ?
-    @project = Project.find_by_param!(param)
+  def require_project
+    @project = (Project.find_by_param!(params[:project_id]) if params[:project_id])
   end
+
 end
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -1,19 +1,14 @@
 class DeploysController < ApplicationController
-  include CurrentProject
   include ProjectLevelAuthorization
 
-
-  before_action except: [:active, :active_count, :recent, :changeset] do
-    find_project(params[:project_id])
-  end
+  skip_before_action :require_project, only: [:active, :active_count, :recent, :changeset]
 
   before_action :authorize_project_deployer!, only: [:new, :create, :confirm, :buddy_check, :pending_start, :destroy]
-
   before_action :find_deploy, except: [:index, :recent, :active, :active_count, :new, :create, :confirm]
   before_action :stage, only: :new
 
   def index
-    @deploys = @project.deploys.page(params[:page])
+    @deploys = current_project.deploys.page(params[:page])
 
     respond_to do |format|
       format.html
@@ -44,7 +39,7 @@ class DeploysController < ApplicationController
   end
 
   def new
-    @deploy = @project.deploys.build(params.except(:project_id).permit(:stage_id, :reference))
+    @deploy = current_project.deploys.build(params.except(:project_id).permit(:stage_id, :reference))
   end
 
   def create
@@ -54,7 +49,7 @@ class DeploysController < ApplicationController
     respond_to do |format|
       format.html do
         if @deploy.persisted?
-          redirect_to [@project, @deploy]
+          redirect_to [current_project, @deploy]
         else
           render :new
         end
@@ -76,7 +71,7 @@ class DeploysController < ApplicationController
       @deploy.confirm_buddy!(current_user)
     end
 
-    redirect_to [@project, @deploy]
+    redirect_to [current_project, @deploy]
   end
 
   def pending_start
@@ -84,7 +79,7 @@ class DeploysController < ApplicationController
       @deploy.pending_start!
     end
 
-    redirect_to [@project, @deploy]
+    redirect_to [current_project, @deploy]
   end
 
   def show
@@ -93,7 +88,7 @@ class DeploysController < ApplicationController
       format.text do
         datetime = @deploy.updated_at.strftime "%Y%m%d_%H%M%Z"
         send_data @deploy.output,
-          filename: "#{@project.repo_name}-#{@deploy.stage.name.parameterize}-#{@deploy.id}-#{datetime}.log",
+          filename: "#{current_project.repo_name}-#{@deploy.stage.name.parameterize}-#{@deploy.id}-#{datetime}.log",
           type: 'text/plain'
       end
     end
@@ -113,7 +108,7 @@ class DeploysController < ApplicationController
       flash[:error] = "You do not have privileges to stop this deploy."
     end
 
-    redirect_to [@project, @deploy]
+    redirect_to [current_project, @deploy]
   end
 
   protected
@@ -127,7 +122,7 @@ class DeploysController < ApplicationController
   end
 
   def stage
-    @stage ||= @project.stages.find_by_param!(params[:stage_id])
+    @stage ||= current_project.stages.find_by_param!(params[:stage_id])
   end
 
   def deploy_params
@@ -139,6 +134,6 @@ class DeploysController < ApplicationController
   end
 
   def active_deploy_scope
-    @project ? @project.deploys.active : Deploy.active
+    current_project ? current_project.deploys.active : Deploy.active
   end
 end
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -1,10 +1,8 @@
 class JobsController < ApplicationController
-  include CurrentProject
   include ProjectLevelAuthorization
 
-  before_action except: [:enabled] do
-    find_project(params[:project_id])
-  end
+  skip_before_action :require_project, only: [:enabled]
+
   before_action :authorize_project_admin!, only: [:new, :create, :destroy]
   before_action :find_job, only: [:show, :destroy]
 
@@ -74,6 +72,6 @@ class JobsController < ApplicationController
   end
 
   def find_job
-    @job = @project.jobs.find(params[:id])
+    @job = current_project.jobs.find(params[:id])
   end
 end
--- a/app/controllers/macros_controller.rb
+++ b/app/controllers/macros_controller.rb
@@ -1,10 +1,6 @@
 class MacrosController < ApplicationController
-  include CurrentProject
   include ProjectLevelAuthorization
 
-  before_action do
-    find_project(params[:project_id])
-  end
   before_action :authorize_project_deployer!
   before_action :authorize_project_admin!, only: [:new, :create, :edit, :update, :destroy]
   before_action :find_macro, only: [:edit, :update, :execute, :destroy]
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,18 +1,16 @@
 class ProjectsController < ApplicationController
-  include CurrentProject
   include ProjectLevelAuthorization
   include StagePermittedParams
 
-  attr_reader :project
-  
-  before_action except: [:index, :new, :create] do
-    find_project(params[:id])
-  end
+  skip_before_action :require_project, only: [:index, :new, :create]
+
   before_action :authorize_admin!, only: [:new, :create, :destroy]
   before_action :authorize_project_admin!, except: [:show, :index, :deploy_group_versions]
   before_action :get_environments, only: [:new, :create]
 
   helper_method :project
+
+  alias_method :project, :current_project
 
   def index
     respond_to do |format|
@@ -108,5 +106,10 @@ class ProjectsController < ApplicationController
 
   def get_environments
     @environments = Environment.all
+  end
+
+  # Overriding require_project from CurrentProject
+  def require_project
+    @project = (Project.find_by_param!(params[:id]) if params[:id])
   end
 end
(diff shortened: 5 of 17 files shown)


app/controllers/deploys_controller.rb
```diff
@@ -1,18 +1,13 @@
 class DeploysController < ApplicationController
-  include CurrentProject
   include ProjectLevelAuthorization
 
-
-  before_action except: [:active, :active_count, :recent, :changeset] do
-    find_project(params[:project_id])
-  end
-
```
jonmoter: How about we keep this `before_action`, but have it call `@project = current_project`, rather than having that line in all the actions below?  That's the point of the before_actions; to extract repeated code.

app/controllers/jobs_controller.rb
```diff
@@ -1,22 +1,21 @@
 class JobsController < ApplicationController
-  include CurrentProject
   include ProjectLevelAuthorization
 
-  before_action except: [:enabled] do
-    find_project(params[:project_id])
-  end
```
jonmoter: Same as my comment on `DeploysController`.

app/controllers/macros_controller.rb
```diff
@@ -1,23 +1,22 @@
 class MacrosController < ApplicationController
-  include CurrentProject
   include ProjectLevelAuthorization
 
-  before_action do
-    find_project(params[:project_id])
-  end
```
jonmoter: Same as my comment on `DeploysController`.

app/controllers/projects_controller.rb
```diff
@@ -97,7 +95,7 @@ def project_params
   end
 
   def project
-    @project
+    current_project
   end
```
jonmoter: Can we get rid of this method?

app/controllers/projects_controller.rb
```diff
@@ -111,4 +109,8 @@ def projects_for_user
   def get_environments
     @environments = Environment.all
   end
+
+  def current_project
+    @project ||= (Project.find_by_param!(params[:id]) if params.try(:[], :id))
+  end
```
jonmoter: Add a comment saying you're overriding the method in the `CurrentProject` module, due to the different param.

app/controllers/releases_controller.rb
```diff
@@ -1,28 +1,27 @@
 class ReleasesController < ApplicationController
-  include CurrentProject
   include ProjectLevelAuthorization
 
-  before_action do
-    find_project(params[:project_id])
-  end
-
```
jonmoter: Same as my comment on `DeploysController`.

app/controllers/stages_controller.rb
```diff
@@ -1,16 +1,11 @@
...
-  include CurrentProject
   include ProjectLevelAuthorization
   include StagePermittedParams
 
   skip_before_action :login_users, if: :badge?
 
-  before_action do
-    find_project(params[:project_id])
-  end
-
```
jonmoter: Same as my comment on `DeploysController`.

app/controllers/projects_controller.rb
sbrnunes: I did attempt to remove it, but his method has been registered as a helper_method, and it's being used in a lot of places (controller, erbs, helpers as well as far as I remember).

So, I ended up not going along that path to keep this PR small and to avoid a bigger risk of breaking something.

app/controllers/deploys_controller.rb
sbrnunes: This refactoring has the ultimate goal of removing this before action. In the code review from pull request #553, It was suggested that I could remove this before_actions in favour of a lazy initialised method (current_project).

Despite knowing I would end up with this repeated code, the reason why I added a line in each action where we need to initialise @project was, first, to make it clear where we need it to be initialised, and second, to minimise the impact from my changes and the size of this PR.

But there's a way to prevent the code duplication and still get rid of the before_action, so, what do you think about:
- Removing the before_actions, but...
- In all the actions where we need the @project, instead of initialising it explicitly, just call the current_project method?

Example:
```
def index
  @project = current_project
  @deploys = @project.deploys.page(..)
end
```
would end up as:
```
def index
  @deploys = current_project.deploys.page(..)
end
```

app/controllers/projects_controller.rb
sbrnunes: Good point.

app/controllers/deploys_controller.rb
jonmoter: If you look in the routes.rb file, you'll find a lot of controllers are defined inside of the `resources :project` block.  That means that those endpoints should only ever get called if a valid project is specified.

So I propose:

1. In the `CurrentProject` module, have it add a `before_filter :find_project`.  That method throws an error if it can't find a valid project.
2. For any actions in controllers where we don't want to get the project, we add a `skip_before_filter` instruction.
3. The `find_project` method assumes `params[:project_id]`, since that's what is used in all but one controller.  The `ProjectsController` can override the method.

That way if there's an invalid project, that is detected immediately, rather than at some random point wherever `current_project` is first called.

That being said, I do like the idea of using `current_project` in the code rather than the `@project` instance variable.

What do you think of that idea?

sbrnunes: > If you look in the routes.rb file, you'll find a lot of controllers are defined inside of the resources :project block. That means that those endpoints should only ever get called if a valid project is specified.

Yep, noticed that.

> That way if there's an invalid project, that is detected immediately, rather than at some random point wherever current_project is first called.
(...)
What do you think of that idea?

I understand your point about failing at some random point when `current_project` is first used, but I still have some doubts in here. When PR [#553](https://github.com/zendesk/samson/pull/553) was first reviewed, and the guys "complained" about the way this before_actions were being used, I started googling for it, and some of the first hits I got kind of explain why should we avoid using before_action callbacks for loading data before actions get executed, and most of the arguments actually make sense, specially the arguments about:
- Sacrificing clarity for the sake of DRY-ness (authorisation filters depending on the `find_project` filter follows the same example described in the articles). However, we already got rid of this by introducing the `current_project` method.
- Filters coupled with controller actions (actions logic, and view templates, depending on the execution of the `find_project` method).
- Changes to the controller's state before the action gets executed (but this will happen anyway when the `current_project` method gets executed by the authorisation filters)

You can find those references in here:
[Don't use before_action to load data](http://craftingruby.com/posts/2015/05/31/dont-use-before-action-to-load-data.html)
[Better alternative to Rails' before_action to load objects](http://jgaskins.org/blog/2014/08/25/better-alternative-to-rails-before_action-to-load-objects)

This being said, I know articles like this can be a little opinionated and you have much more experience with Rails than me. So, if you believe that's the way to go, then your suggestion might be the best one.

What's your opinion on these articles?

jonmoter: Rails is an opinionated framework. It does a whole lot of magic for you behind the scenes, and sacrifices explicitness for ease of use, and needing minimal boilerplate code. When I'm thinking of inserting some "magic" code like this, I ask myself "does this follow patterns that Rails developers are familiar with, or would this be surprising to them?"

The pattern of loading certain required records as part of controllers is pretty standard. For example, you don't have to call a method to define `current_user` in every action of every controller. That's just done for you as part of authorization.

Similarly, a large number of our controller actions don't make sense unless you have a `current_project`.  If you tried going to http://samson.example.com/projects/NOT_A_PROJECT/deploys, it should immediately error or redirect you without ever running the action. So given that's a necessary check, and since it shows up in so many controllers, it makes sense to me to have it in a `before_filter`. Actually, maybe call the filter something like `require_project` to make it more explicit that it's doing a check, and not just loading something from the DB.

So in this case, in my opinion, abstracting out the loading and checking of `current_project` into a module and having that happen automatically is worth the slight loss of explicitness.  But it's obviously not something with a clear-cut answer to it.

app/controllers/projects_controller.rb
fneves: you can use `alias_method` to get rid of this and keep the compatibility with the rest if you don't want to touch the rest of the code.

app/controllers/projects_controller.rb
```diff
@@ -109,4 +111,8 @@ def projects_for_user
   def get_environments
     @environments = Environment.all
   end
+
+  def current_project
+    @project ||= (Project.find_by_param!(params[:id]) if params.try(:[], :id))
```
steved: params is never nil

app/controllers/users_controller.rb
```diff
@@ -1,32 +1,15 @@
...
-  end
-
   before_action :authorize_project_admin!
 
   def index
-    scope = User
-    scope = scope.search(params[:search]) if params[:search]
-    @users = scope.order("#{sort_column} #{sort_direction}").page(params[:page])
+    @project = current_project
+    @users = User.search_by_criteria(params)
```
steved: where is this defined? I can't seem to find it

steved: Nevermind, didn't follow the correct trail of PRs.

app/controllers/projects_controller.rb
sbrnunes: Good point. Fixed.

app/controllers/concerns/current_project.rb
```diff
@@ -6,13 +6,6 @@ module CurrentProject
...
   def current_project
-    @project
-  end
-
-  protected
-
-  def find_project(param)
-    # Will this return not found if project does not exist ?
-    @project = Project.find_by_param!(param)
+    @project ||= (Project.find_by_param!(params[:project_id]) if params.try(:[], :project_id))
```
steved: params is never nil here either

steved:
As long as @jonmoter's comments are addressed (I think they are, I just can't really tell) +1

app/controllers/concerns/current_project.rb
sbrnunes: Roger that.

sbrnunes:
@jonmoter's comments are on the way. I'll commit it in a few moments. Thanks for the review @steved.

sbrnunes:
@jonmoter, I just committed the changes, following what you suggested regarding the `before_action` and `skip_before_action`. I also changed the code to use `current_project`.

Let me know if everything is ok.

app/controllers/projects_controller.rb
sbrnunes: Used the `alias_method`. Thanks for the suggestion @fneves .

app/controllers/deploys_controller.rb
```diff
@@ -139,6 +134,6 @@ def find_deploy
   end
 
   def active_deploy_scope
-    @project ? @project.deploys.active : Deploy.active
+    current_project ? current_project.deploys.active : Deploy.active
```
jonmoter: Calling `current_project` here to know if it exists will call `require_project`, which will throw an error if there's no project.  I think in this case, you want to leave it as @project.

jonmoter:
Looks great to me, other than that one comment above.

Btw, I just noticed this PR is to merge this branch into `sbrnunes/add-deployer-project-level-access` rather than `master`.  Is that on purpose?

app/controllers/deploys_controller.rb
sbrnunes: @jonmoter , this line has confused me a little bit. I created this branches from your initial branch jonmoter/add-project-admins. In that branch, you'll find a difference to master in the way how find_project is used:
- In master:
```
before_action :find_project
```
- In your original branch:
```
before_action :find_project, except: [:active, :active_count, :recent, :changeset]
```
So, in the way this was implemented in the branch, I believe that when the `active` and `active_count` actions were executed, `@project` was never being initialised, and we were always returning `Deploy.active`, unless I missed something. This was probably hard to notice because the fallback was to return all active deploys.

I kept your changes, because I noticed we need to be able to load all active deploys when outside the context of a project, but, in this line, I think we should first make an attempt to load the project, if `project_id` is present in the params, and fallback to list all active deploys, otherwise. That's why I used the `current_project` method.

Can you double check if I'm missing something?

jonmoter: Oh, sorry, my bad. I didn't see the `if params[:project_id]` conditional in `require_project`.  So I thought that calling `current_project` would either return a valid Project, or throw an error.

Nevermind, carry on.

sbrnunes:
Yeah... I had to split my first PR in multiple, smaller, PRs. The lads here went crazy with the size of the PR. :)

So, answering your question, I'm planning on merging this PRs gradually, as soon as I get a :+1: in all of them:
1. sbrnunes/refactoring-find-project-filters into sbrnunes/add-deployer-project-level-access
2. sbrnunes/add-deployer-project-level-access into sbrnunes/add-project-level-access-control
3. sbrnunes/add-project-level-access-control into master

app/controllers/deploys_controller.rb
sbrnunes: :)

jonmoter:
Would it not work to merge this into master?  In your current plan, the merge in step 3 is still going to be gigantic

sbrnunes:
I can't merge this one directly into master, unless I merge the other two branches first. This branch was branched from sbrnunes/add-deployer-project-level-access just to address the refactoring of this filters, so, it's tightly coupled with that branch.

I've been rebasing from master whenever I notice the branches getting too many commits behind master. Hopefully, it will make it easier.

jonmoter:
Okay, give it a go.  :+1:

PR #569, opened 2015-10-05
remove revision from Dockerfile

steved:
@zendesk/samson 

I've also set up a dedicated zendesk/samson registry build.
https://hub.docker.com/r/zendesk/samson/


jonmoter:
:+1:

PR #570, opened 2015-10-05
Fix for Slack authentication errors

steved:
@zendesk/samson

### DIFF
--- a/plugins/slack/app/decorators/stage_decorator.rb
+++ b/plugins/slack/app/decorators/stage_decorator.rb
@@ -34,14 +34,14 @@ Stage.class_eval do
   def channel_for(name)
     return nil unless name
 
-    response = Rails.cache.fetch(slack_channels_cache_key, expires_in: 5.minutes) do
-      Slack.channels_list(exclude_archived: 1)
-    end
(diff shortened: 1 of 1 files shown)


plugins/slack/app/decorators/stage_decorator.rb
```diff
@@ -34,9 +34,14 @@ def channel_exists?
...
+    response = Rails.cache.read(slack_channels_cache_key) || Slack.channels_list(exclude_archived: 1)
+
+    if response['ok']
+      Rails.cache.write(slack_channels_cache_key, response, expires_in: 5.minutes)
+    else
+      Rails.logger.error("Slack API error: #{response.inspect}")
     end
-    response['channels'].select { |c| c['name'] == name }.first
+
+    response.fetch('channels', []).select { |c| c['name'] == name }.first
```
jonmoter: I think it would be a little more clear to put that in the `if response['ok']` logic block, and explicitly say you return `nil` in the else statement.

steved: sure

jonmoter:
:+1: on :green_heart:

sandlerr:
The old cache key had a 5 minute expiry, will this persist indefinitely?

PR #571, opened 2015-10-05
update dockerfile

steved:
* CMD syntax to exec preferred form
* Remove database constraints
* Update ruby version

@zendesk/samson


sandlerr:
:+1:

Issue #531, opened 2015-08-05
Small UI issue


steved:
![screen shot 2015-08-04 at 5 43 13 pm](https://cloud.githubusercontent.com/assets/13248/9075921/4f85274e-3ad0-11e5-9e5d-3a3b1b12b847.png)

henders:
You may have to spell out for me what the UI issue is? Blue vs white buttons? Too close?

steved:
Sorry yeah, it should be have a small margin between.

PR #572, opened 2015-10-05
UI modifications

steved:
* fix for #531 
* unify stop buttons so jobs and deploys are relatively similar

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -50,13 +50,13 @@ class JobsController < ApplicationController
   end
 
   def destroy
-    if @job.started_by?(current_user) || current_user.is_admin?
+    if @job.can_be_stopped_by?(current_user)
       @job.stop!
-
-      head :ok
     else
-      head :forbidden
+      flash[:error] = "You do not have privileges to stop this job."
     end
+
+    redirect_to [@project, @job]
   end
 
   private
--- a/app/views/jobs/_output.html.erb
+++ b/app/views/jobs/_output.html.erb
@@ -7,12 +7,11 @@
         <button class="btn btn-default" id="output-grow">Expand</button>
       </div>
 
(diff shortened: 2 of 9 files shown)


app/controllers/jobs_controller.rb
```diff
@@ -50,13 +50,13 @@ def enabled
...
+    if @job.can_be_stopped_by?(current_user)
       @job.stop!
-
-      head :ok
     else
-      head :forbidden
+      flash[:error] = "You do not have privileges to stop this job."
     end
+
+    redirect_to [@project, @job]
```
jonmoter: Does the `JobsController#destroy` method ever get called via AJAX? If so, a `:forbidden` response makes more sense than a redirect.

steved: Not anymore, it used to be

app/views/jobs/_output.html.erb
```diff
@@ -7,12 +7,11 @@
...
-      <div class="pull-right only-finished">
-        <button class="btn btn-default" id="output-grow-toggle">Expand</button>
-        <%= link_to "Download log", project_job_path(@project, job, format: :text), class: "btn btn-primary" %>
-      </div>
+      <button class="btn btn-default only-finished" id="output-grow-toggle">Expand</button>
+      <%= link_to "Download log", project_job_path(@project, job, format: :text), class: "btn btn-primary only-finished" %>
+
       <% if job.can_be_stopped_by?(current_user) %>
-        <%= stop_button(class: "btn btn-danger pull-right only-active") %>
+        <%= stop_button(deploy: job, class: "btn btn-danger only-active") %>
```
jonmoter: That's a little weird to pass in a job when it appears to be expecting an instance of `Deploy`, but I see what you're doing, and can't think of a better name for the parameter off the top of my head.

jonmoter:
:+1:

PR #574, opened 2015-10-06
use process groups and kill

steved:
pkill with a parent id doesn't work for single processes (i.e. sleep infinity).
Find out the pgid, and then use kill on that.

This stuff still doesn't work as well as I'd like, let's try a new tactic.

@zendesk/samson

### DIFF
--- a/app/models/terminal_executor.rb
+++ b/app/models/terminal_executor.rb
@@ -13,7 +13,7 @@ require 'shellwords'
 #   output.string #=> "hello\r\nworld\r\n"
 #
 class TerminalExecutor
-  attr_reader :pid, :output
+  attr_reader :pid, :pgid, :output
 
   def initialize(output, verbose: false)
     @output = output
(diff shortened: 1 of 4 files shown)


app/models/terminal_executor.rb
```diff
@@ -36,7 +36,11 @@ def execute!(*commands)
   def stop!(signal)
     # Need pkill because we want all
     # children of the parent process dead
-    system "pkill", "-#{signal}", "-P", pid.to_s if pid
+    if pgid
+      system('kill', "-#{signal}", "-#{pgid}")
+    elsif pid
+      system "pkill", "-#{signal}", "-P", pid.to_s
+    end
```
jonmoter: Consistent use of parens, please, for those of us with mild OCD tendencies.  :-)

jonmoter:
My knowledge of how PGID works is limited.  Is there any scenario when the Puma process would have the same PGID as the spawned process?  I want to be careful we don't accidentally kill the Samson web server when trying to stop a job.

steved:
PTY#spawn resets the process group: https://github.com/ruby/ruby/blob/trunk/ext/pty/pty.c#L100-L102

jonmoter:
Okay, let's give this a try.  With any luck, this will fix [SAMSON-157](https://zendesk.atlassian.net/browse/SAMSON-157).

:+1:

PR #575, opened 2015-10-06
unify status panels

steved:
@zendesk/samson


jonmoter:
Looks like a good refactor. Thanks for the cleanup!  :+1:

PR #576, opened 2015-10-06
remove random unneccessary files + bundle update

steved:


steved:
@zendesk/samson

test/integration/cleanliness_test.rb
```diff
@@ -11,14 +11,4 @@
     # if it is not avoidable see http://stackoverflow.com/questions/7163264
     File.read('Gemfile.lock').wont_include 'rails-assets-bootstrap '
   end
-
-  if ActiveRecord::Base.connection.adapter_name == "Mysql2"
-    it "uses the right row format in mysql" do
```
sandlerr: why remove this?

steved: I don't really understand what it's testing, the travis config? People can run samson without utf8mb4, so why do we care? I'd argue the monkey patch is also unnecessary and should be optionally included by users.

sandlerr: you're right, if utf8mb4 is not used then the index size limits won't be exceeded. It did take us a while to figure out why migrations were failing and how to fix it though. Is it reasonable to use the monkey patch in the test environment and keep this spec, but make it conditional on an envar?

steved: ok, added

steved: I'm pretty sure we're still using utf8 in our deployment too.

config/initializers/active_record.rb
```diff
@@ -2,12 +2,17 @@
 # We also want our mysql databases to use utf8_mb4 encoding (the real unicode)
 # This monkey patch is from the following Github comment:
 # https://github.com/rails/rails/issues/9855#issuecomment-57665404
-if Gem.loaded_specs.has_key? 'mysql2'
-  require 'active_record/connection_adapters/mysql2_adapter'
+config = ActiveRecord::Base.configurations[Rails.env]
+
+if ENV.fetch('USE_UTF8M4', Rails.env.test?) && config['adapter'] == 'mysql2'
```
sandlerr: nitpick: insert the B before the `4`

sandlerr:
:+1:

config/initializers/active_record.rb
steved: whoops, thanks

test/integration/cleanliness_test.rb
sandlerr: Yes, our deployment uses Compact

PR #577, opened 2015-10-07
Add audit trail of config changes to projects [SAMSON-176]

sbrnunes:
Adding support for auditing config changes on projects. At the moment, only the following changes will be audited:
- Adding/removing Webhooks on a Project
- Adding/removing commands on a Stage
- Editing a command, on a Stage or through the admin menu

/cc @zendesk/samson 

### Steps to reproduce
When one of the above operations take place, the audit log will be written into the stdout with the following format:
```
{
  "logtype": "AUDIT",
  "logged_at": <some timestamp>,
  "user": <user name and email>,
  "object": <object type: Stage | Command | Webhook>,
  "action": <action type: create | update | destroy>,
  "before": <json string with the previous state of the object>,
  "after": <json string with the new state of the object>
}
```

### References
 - [SAMSON 176](https://zendesk.atlassian.net/browse/SAMSON-176)

### Risks
 - Logging sensitive information (make sure to filter out any sensitive data or to log only what's relevant for auditing purposes)

### DIFF
--- a/app/controllers/concerns/samson_audit.rb
+++ b/app/controllers/concerns/samson_audit.rb
@@ -1,0 +1,57 @@
+module SamsonAudit
+  extend ActiveSupport::Concern
+  include CurrentUser
+
+  mattr_accessor :original_attributes
+
+  def prepare_audit(subject)
+    self.original_attributes = subject.to_auditable_json if is_auditable(subject)
+  end
+
+  def audit(subject)
+    case action_name
+    when 'create' then
+      audit_create(action_name, subject)
+    when 'destroy' then
+      audit_destroy(action_name, subject)
+    when 'update' then
+      audit_update(action_name, subject)
+    else
+      raise "Unsupported action: #{action_name}"
+    end
+  end
+
+  private
+
+  def audit_create(action, subject)
+    audit_action(action, subject, after: subject.to_auditable_json) if is_auditable(subject)
+  end
+
+  def audit_destroy(action, subject)
+    audit_action(action, subject, before: subject.to_auditable_json) if is_auditable(subject)
+  end
+
+  def audit_update(action, subject)
+    audit_action(action, subject, before: original_attributes, after: subject.to_auditable_json) if is_auditable(subject)
+  end
+
+  def audit_action(action, subject, before: {}, after: {})
+    Rails.logger.info(JSON.pretty_generate(audit_object(action, after, before, subject)))
+  end
+
+  def is_auditable(subject)
+    subject.respond_to? :to_auditable_json
+  end
+
+  def audit_object(action, after, before, subject)
+    {
+      logtype: 'AUDIT',
+      logged_at: "#{Time.now.getutc}",
+      user: "#{current_user.name_and_email}",
+      object: "#{subject.class.name}",
+      action: "#{action}",
+      before: "#{before}",
+      after: "#{after}"
+    }
+  end
+end
--- a/app/models/command.rb
+++ b/app/models/command.rb
@@ -27,4 +27,12 @@ class Command < ActiveRecord::Base
   def self.usage_ids
     MacroCommand.pluck(:command_id) + StageCommand.pluck(:command_id)
   end
+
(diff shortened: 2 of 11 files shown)


jonmoter:
Why make the `before` and `after` values be strings, rather than JSON sub-elements?

app/controllers/concerns/samson_audit.rb
```diff
@@ -0,0 +1,57 @@
...
+    audit_action(action, subject, before: original_attributes, after: subject.to_auditable_json) if is_auditable(subject)
+  end
+
+  def audit_action(action, subject, before: {}, after: {})
+    Rails.logger.info(JSON.pretty_generate(audit_object(action, after, before, subject)))
+  end
+
+  def is_auditable(subject)
+    subject.respond_to? :to_auditable_json
+  end
```
jonmoter: General ruby pattern: if you're going to make a method that returns true or false, use a question make.  e.g. call this method `auditable?(subject)` rather than `is_audible(subject)`.

sbrnunes:
@jonmoter , maybe it was the way how I described it. So, currently, what I'm logging in the `before` and `after` values is the result of calling `to_json` on the ActiveModel object that we want to audit, which returns a JSON string representing the object.

Example from stage.rb:
```
to_json(only: [:id, :name, :project_id, :webhooks],
     include: [
       { commands: { only: [:id, :command, :project_id] } },
       { stage_commands: { only: [:id, :command_id] } }
])
```

I'm not, at the moment at least, pretty printing this JSON representation of this objects, just the overall audit log object.

Is this good enough or do you want me to review it?

app/models/command.rb
```diff
@@ -27,4 +27,12 @@ def usages
...
     MacroCommand.pluck(:command_id) + StageCommand.pluck(:command_id)
   end
+
+  def to_auditable_json
+    to_json(
+      only: [:id, :command, :project_id],
+      include: [
+        stages: { only: [:id] }
+      ])
+  end
```
jonmoter: How about we call the method `to_auditable_hash` and then use `as_json` instead of `to_json`.

As a rule of thumb, you should hardly ever need to call `to_json` except at the rendering step.  Models should return hashes, and then have something else be responsible for serializing the data.

Some more info: http://jonathanjulian.com/2010/04/rails-to_json-or-as_json/

steved: Please use serializers for this, rails' built in as_json/to_json are poor presenters.

steved:
I'm butting in, but why not just use an off the shelf auditing gem? I think this behavior belongs closer to the model. [paper_trail](https://github.com/airblade/paper_trail) is popular and @grosser wrote [cia](https://github.com/grosser/cia).

jonmoter:
@sbrnunes, I was looking over the [cia](https://github.com/grosser/cia) gem that @grosser wrote, and that would actually work pretty well in this case.  It would store changes in the DB, so we could (eventually) write a UI to see the revision history of things we care about.

I think it would be a pretty small amount of work to integrate that. Wanna try opening up a separate PR with that implementation?

Sorry for not thinking of this sooner.  If I were smart, I would've looked around for audit trail gems before suggesting rolling our own.

grosser:
prefer paper_trail it stores full audits / is much more popular ... this might help https://github.com/zendesk/samson/pull/577

jonmoter:
@grosser did you mean to link to a different PR?

grosser:
ups :dancer: I remembered someone doing paper_trail and searched for the PR ... guess I found the right one :D

sbrnunes:
@jonmoter, sounds good. I should have asked if it made sense to write it to the DB. When I was searching for gems for auditing purposes, most of them were actually writing to the DB, but I assumed we didn't want that. :)

I'm going to create a new PR for this, then, and I'm going to look at paper_trail.

jonmoter:
If we want something more immutable than a DB record, we could probably figure out a way to write a line out to a log file every time a row is added to the table(s) that paper_trail uses.

PR #582, opened 2015-10-08
user must specify project when requesting access rights

msufa:
This PR is building on #561.

The user needs to choose projects for which the additional access rights should be granted. Project names will be added to the email request. This functionality is added to align the feature with #560 

screen shot 2015-10-09 at 11 19 24

Sample email:

screen shot 2015-10-09 at 11 20 41

/cc @zendesk/samson

### References

### Risks
 - None

### DIFF
--- a/app/views/access_requests/new.html.erb
+++ b/app/views/access_requests/new.html.erb
@@ -20,10 +20,10 @@
     </fieldset>
 
     <div class="form-group">
-      <label class="col-lg-2 control-label">Project</label>
+      <label class="col-lg-2 control-label">Projects</label>
       <div class="col-lg-4">
-        <%= select_tag :project_id, options_from_collection_for_select(Project.all.order(:name), 'id', 'name'),
-                       class: 'form-control', include_blank: true, required: 'required' %>
+        <%= select_tag :project_ids, options_from_collection_for_select(Project.all.order(:name), 'id', 'name'),
+                       class: 'form-control', multiple: true, required: 'required' %>
       </div>
     </div>
 
--- a/app/views/access_requests/new.html.erb
+++ b/app/views/access_requests/new.html.erb
@@ -19,6 +19,14 @@
       </div>
     </fieldset>
 
+    <div class="form-group">
+      <label class="col-lg-2 control-label">Project</label>
+      <div class="col-lg-4">
+        <%= select_tag :project_id, options_from_collection_for_select(Project.all.order(:name), 'id', 'name'),
+                       class: 'form-control', include_blank: true, required: 'required' %>
+      </div>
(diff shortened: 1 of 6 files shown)


pswadi-zendesk:
Would be good to allow users to choose multiple projects. That way in one request user can get access to all the project she is interested in.

msufa:
Good point @pswadi-zendesk, I'll tweak the PR to make this possible.

app/views/access_requests/new.html.erb
```diff
@@ -19,6 +19,14 @@
       </div>
     </fieldset>
 
+    <div class="form-group">
+      <label class="col-lg-2 control-label">Projects</label>
+      <div class="col-lg-4">
+        <%= select_tag :project_ids, options_from_collection_for_select(Project.all.order(:name), 'id', 'name'),
```
jonmoter: In general, it's an anti-pattern to make database calls in the view layer.  Better to load the data in the controller and set it to an instance variable.

test/controllers/access_requests_controller_test.rb
```diff
@@ -53,7 +53,7 @@
       describe 'enabled' do
         let(:manager_email) { 'manager@example.com' }
         let(:reason) { 'Dummy reason.' }
-        let(:request_params) { {manager_email: manager_email, reason: reason} }
+        let(:request_params) { {manager_email: manager_email, reason: reason, project_ids: Project.all.map(&:id)} }
```
jonmoter: can say `Project.all.pluck(:id)`

test/mailers/access_request_mailer_test.rb
```diff
@@ -24,37 +22,58 @@
...
 
-    it 'includes reason in body' do
-      subject.body.to_s.must_match /#{reason}/
+    describe 'single project' do
+      before { AccessRequestMailer.access_request_email(
+          hostname, user, manager_email, reason, [projects(:test).id]).deliver_now }
+
+      it 'includes target project name in body' do
+        subject.body.to_s.must_match /#{projects(:test).name}/
+      end
```
jonmoter: FWIW, I don't really believe in being strict about "have a separate test for everything" philosophy.  Well, I would be more in favor of that, if it weren't for the fact that each test is making at least one database insert.  This adds up and slows down the test suite.

I'd probably do this by having one test each for the "to" address, the subject, and the body, and more than one assertion in each test.

jonmoter:
Now that #561 is merged, can open a new PR targeted at `master`.

msufa:
Thanks @jonmoter. Closing this PR. I'll address your comments in a new one targeted for master.

Issue #11, opened 2014-05-30
Job status tool


halcyonCorsair:
We should have a small utility script that allows us to easily see information about running jobs, eg:
- what jobs are running
- for how long
- pid

It should also allow us to kill long running or stalled jobs.

xakraz:
:+1:

PR #584, opened 2015-10-09
job queueing

steved:
samson-queue


steved:
@zendesk/samson

app/models/job_execution.rb
```diff
@@ -146,24 +154,22 @@ def setup!(dir)
...
-      commands.unshift("export STAGE=#{@stage.permalink.shellescape}")
+    env = {
+      DEPLOY_URL: @job.full_url,
+      DEPLOYER: @job.user.email,
+      DEPLOYER_EMAIL: @job.user.email,
+      DEPLOYER_NAME: @job.user.name,
+      REVISION: @reference,
+      TAG: (@job.tag || @job.commit).to_s,
+      CACHE_DIR: artifact_cache_dir
+    }.merge(@env)
```
jonmoter: Are those constants defined somewhere?  Don't you mean something like:

```
env = {
  'DEPLOY_URL' => @job.full_url,
  ...
}
```

steved: It's just an uppercase symbol

```
irb(main):001:0> { A: 1 }
=> {:A=>1}
```

jonmoter:
Hey @steved, could you give a little context on the purpose of this PR?  Is this purely refactoring and cleanup, or does this enable new functionality, or fix existing bugs?

steved:
This allows our integrations to enqueue multiple deploys per stage instead of just returning "unprocessable entity" when one is already running. It technically also allows users to do the same, but the UI prevents it unless you go directly to the deploy creation page.

app/models/job_execution.rb
jonmoter: Oh, right.  I'm thinking of `{ DEPLOY_URL => @job.full_url }`.  That explains why it works.

I don't think I've ever seen uppercase symbols in Ruby, so I was thrown off by that.  I think it sacrifices understandability a bit, so I'd prefer the construct I suggested.  But not that huge a deal.

app/controllers/streams_controller.rb
```diff
@@ -10,20 +10,29 @@ def show
...
-    streamer = EventStreamer.new(response.stream, &method(:event_handler))
-
-    return response.stream.close unless @job.active? && @execution
-
-    @execution.viewers.push(current_user)
+    execution.viewers.push(current_user)
     ActiveRecord::Base.clear_active_connections!
-    streamer.start(@execution.output)
+
+    sleep(0.1) until execution.active?
```
jonmoter: Is there any scenario where this could end up as an infinite loop?  Like if the `JobExecution` was canceled or removed from the queue?  (I don't know if that's possible right now, but it seems like a reasonable enhancement.)

Or if the thread starts and ends in less than 1/10 of a second?

Maybe something more like `sleep(0.1) while execution.pending?`

app/models/deploy_service.rb
```diff
@@ -19,7 +19,13 @@ def deploy!(stage, attributes)
   def confirm_deploy!(deploy)
     send_before_notifications(deploy)
 
-    job_execution = JobExecution.start_job(deploy.reference, deploy.job)
+    stage = deploy.stage
+
+    job_execution = JobExecution.start_job(
+      deploy.reference, deploy.job,
+      construct_env(stage).merge(key: stage.id)
```
jonmoter: Any reason why you merge `key` here, rather than including it in `construct_env`?

steved: It's not really part of the env, just the general 'options'

app/controllers/streams_controller.rb
steved: Yes, but only if you queue a job, the stream is opened, and then the server is restarted. Pushed a commit to fix it.

app/models/job_execution.rb
```diff
@@ -11,33 +11,43 @@ class JobExecution
...
   cattr_accessor(:stop_timeout, instance_writer: false) { 15.seconds }
 
-  cattr_reader(:registry, instance_accessor: false) { {} }
+  cattr_reader(:registry, instance_accessor: false) { JobQueue.new }
   private_class_method :registry
 
   attr_reader :output, :job, :viewers, :stage, :executor
 
-  def initialize(reference, job)
+  def initialize(reference, job, **env, &block)
```
jonmoter: Why a double-splat for env?  Shouldn't it just be a regular hash that gets passed in?

app/controllers/streams_controller.rb
steved: Hmm, the best solution is probably to just block on the event streamer output. I didn't really want to  ouch it, but it's cleaner.

app/models/job_execution.rb
steved: Yeah, that's left over from when I had a `key:` option argument.

app/models/deploy_service.rb
```diff
@@ -19,7 +19,13 @@ def deploy!(stage, attributes)
...
+    )
+
     send_sse_deploy_update('start', deploy)
 
     job_execution.on_complete do
```
jonmoter: I just noticed there's a race condition here.  The job actually gets started when you call `start_job` above, which spawns off a thread.  Here in the code it's adding a callback via `on_complete`, but just assuming that the JobExecution hasn't gotten to that point yet.

app/models/job_execution.rb
```diff
@@ -95,7 +104,6 @@ def run!(&block)
     @job.update_output!(OutputAggregator.new(@output).to_s)
     @subscribers.each(&:call)
```
jonmoter: FYI, we should probably wrap each call in a try/catch block.  I've seen instances where there jobs in the global registry that completed, but were never removed because some code in one of the subscribers raised an exception, and the remaining subscriptions (and any code below this point) was never called.

jonmoter:
Looks good on the whole.

One comment is that in the past, I've seen scenarios where Samson refuses to restart because it still thinks there are running jobs in its job registry, but those jobs are actually dead.  I think they were never removed from that in-memory hash because an exception was raised in the `ensure` block of `JobExecution#run!`.  But there was no way I could modify that in-memory data structure via Rails console, or rake, 'cause that would spawn a different process.  I had to force-kill and restart Samson in that scenario.

I was thinking it might be wise to have a thread that just wakes up every minute or so, scans the in-memory active job set, and confirms that all of them are in a valid state.  We can do that as a separate PR, but I wanted to run the idea by you, in case you have any ideas.

app/models/deploy_service.rb
steved: Good call, I wonder if this and the subscriber errors may be why we've seen issues where emails and notifications aren't sent out

steved:
I feel like a tool and API as described in #11 would be a better solution. I hope we can fix the underlying problems though.

config/initializers/job_execution.rb
```diff
@@ -1,25 +1,28 @@
...
       JobExecution.enabled = true
-    end
-  end
 
-  Signal.trap('SIGUSR1') do
-    if JobExecution.enabled
-      # Disable new job execution
-      JobExecution.enabled = false
+      if Job.table_exists?
+        Job.running.each(&:stop!)
```
sandlerr: this is surprising

app/models/job_execution.rb
```diff
@@ -179,26 +201,27 @@ def lock_project(&block)
...
+      registry.active?(key, id)
+    end
+
+    def queued?(id, key: id)
+      registry.queued?(key, id)
+    end
+
+    def start_job(reference, job, key: job.id, **options, &block)
+      registry.add(key, reference, job, options, &block)
     end
```
jonmoter: Last comment, I think.  :-)

I find the method signature here confusing.  The `**options` double-splat has one special option (`:on_complete`), and all other values in the hash are env variables.  There's the `:on_complete` block that is run afterwards, and then another block that can be the main execution block of the job.  How about we just put everything in a hash, like:

```ruby
def start_job(reference, job, key: job.id, env: {}, on_exec: nil, on_complete: nil)
```

app/models/signal_handler.rb
```diff
@@ -0,0 +1,31 @@
+class SignalHandler
+  def initialize
+    @read, @write = IO.pipe
+    Thread.new { run }
```
jonmoter: The old signal handling code didn't run in a separate thread.  Is there a reason we're now spawning a thread to execute this logic?

config/initializers/job_execution.rb
```diff
@@ -1,25 +1,28 @@
...
       end
 
-      puts "Passing SIGUSR2 on."
-
-      # Pass USR2 to the underlying server
-      Process.kill('SIGUSR2', $$)
+      if Deploy.table_exists?
+        Deploy.active.each do |deploy|
+          next unless deploy.pending_non_production?
+          deploy.pending_start!
```
jonmoter: Would be simpler as `deploy.pending_start! if deploy.pending_non_production?`

jonmoter:
Looks good.  I have the two suggestions on code readability, and the one question about the Thread in the Signal handling code, but other than that it seems like a good improvement.

Agreed it would be good to have a tool or API as mentioned in #11, and that we can fix the underlying problems.

app/models/signal_handler.rb
steved: JobQueue#clear locks a mutex, which isn't allowed inside signal handlers. This is also slightly better practice, since signal handlers are always preemptible (even if USR2 probably won't be called all that often).

steved: I added a comment to the class

config/initializers/job_execution.rb
steved: sure

config/initializers/job_execution.rb
steved: If a job is running after a restart, that means the execution thread is gone. So just cancel it.

app/models/job_execution.rb
steved: Yeah, I agree. I just don't like. I made start_job take an execution instance instead, then on_complete blocks are still done on the instance and it's cleaner for JobQueue to know nothing about instantiating JobExecution.

app/assets/javascripts/deploys.js
```diff
@@ -225,16 +224,28 @@ $(function () {
...
     }
   });
-
 });
 
 function toggleOutputToolbar() {
   $('.only-active, .only-finished').toggle();
 }
+
+function waitUntilEnabled(path) {
```
sandlerr: is it strange to go from a stream to polling? Couldn't we just send down an enabled event?

steved: We need to break all connections for the restart to happen. I couldn't find a way to auto-reconnect EventSources we've closed, though I guess just instantiating a new object might work.

app/models/job.rb
```diff
@@ -25,6 +25,10 @@ def self.running
     where(status: 'running')
   end
 
+  def self.active
```
sandlerr: doesn't look like this is used

app/models/job_execution.rb
```diff
@@ -179,26 +197,27 @@ def lock_project(&block)
...
-      end
+    def active?(id, key: id)
+      registry.active?(key, id)
+    end
+
+    def queued?(id, key: id)
+      registry.queued?(key, id)
+    end
+
+    def start_job(job_execution, key: job_execution.id)
```
sandlerr: this may be better as an instance method rather than a class method that accepts one instance as an argument

steved: I don't think the instance needs to know about the queue

steved:
bump, I'd like to merge.

jonmoter:
Looks good enough to merge. If we need to tweak or clean up a little, we can do it post-merge.

:+1:

PR #586, opened 2015-10-09
support for multiple streaming domains

steved:
@zendesk/samson 

Probably want to add support to SseRailsEngine, but any comments?


henders:
So is this trying to fix the max connections problem with teh SSEs? Or just tightening up security for JS?

steved:
Trying to fix the max connections. Right now you can only open three deploy output windows. This would allow six.

henders:
Gotcha, I couldn't figure out how this header fixes the connection count issue by looking at the specification for it: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

Mind educating me? :)

steved:
Ah, sorry. So, chrome has a connection pool limit of 6. But not per top level domain (i.e. zende.sk or samson.zende.sk), but per actual domain. If we move streaming and SSE to separate domains, we can bump the limit of windows open to always be 6 instead anywhere from 3-6 depending on which pages are open. But, browers aren't allowed to make client side requests to domains this isn't the one they're executing on (so samson.zende.sk can't request sse.samson.zende.sk) without CORS.

steved:
Allow-Credentials sends cookies with the request (not required for sse-*, but required for deploy streaming).

steved:
Here's our internal jira ticket: https://zendesk.atlassian.net/browse/OP-14777

henders:
Ah cool, thanks for the info!  I also pushed the PRs for the SSE gem that you created.

henders:
Did you want to finish this now that the dependent changes are available in the SSE gem?

steved:
It can be merged since there are sensible defaults, yeah. The actual domains and configuration are still in flight.

steved:
Bump, changes will be pushed out soon. I'ld like to merge.

config/application.rb
```diff
@@ -73,7 +73,12 @@ class Application < Rails::Application
     config.samson.docker = ActiveSupport::OrderedOptions.new
     config.samson.docker.registry = ENV['DOCKER_REGISTRY'].presence
 
-    config.samson.uri = URI( ENV["DEFAULT_URL"] || 'http://localhost:3000' )
+    config.samson.uri = URI(ENV["DEFAULT_URL"] || 'http://localhost:9080')
```
henders: Why the port change to 9080? Should probably be 3000, since that's still the development default.

config/application.rb
```diff
@@ -73,7 +73,12 @@ class Application < Rails::Application
...
+    config.sse_rails_engine.access_control_allow_origin = config.samson.uri.to_s
+
+    config.samson.stream_origin = ENV['STREAM_ORIGIN'].presence || config.samson.uri.to_s
+    config.samson.deploy_origin = ENV['DEPLOY_ORIGIN'].presence || config.samson.uri.to_s
```
henders: If the developer doesn't set these, do the SSE streams still work?

henders:
:+1: after default port change back.

PR #588, opened 2015-10-15
Trim the Git reference for the Deploy model

andreionut:
@zendesk/samson 

The Git reference is not trimmed and can result in user confusion.

screenshot 2015-10-15 10 29 15


henders:
:+1:

PR #589, opened 2015-10-15
fix docker builder service breakage from #584

steved:
@zendesk/samson 

No clue how to test this besides actually building an image.


jonmoter:
Yeah, testing this stuff is tough.  Maybe stubbing out the Docker commands, and making sure that all the surrounding code works properly?

:+1:

PR #591, opened 2015-10-16
Request additional access rights from profile page

msufa:
Viewers are instructed that they lack access rights when trying to view restricted content. Deployers have read access to everything so it's not possible for a deployer to request additional access rights from a popup. A link to request access rights is added in the profile page for every user except super_admin.

screen_shot_2015-10-08_at_11_13_25

/cc @zendesk/samson @jonmoter 

### References
 - Jira link: 

### Risks
 - None


jonmoter:
:+1:

PR #592, opened 2015-10-16
Fix a few small bugs that occurred running tests locally

jonmoter:
A few small fixes related to automated tests:

* Updates Rakefile so that `rake` and `rake test` now do the same thing
* Fixes fixtures problem where running `mtest` on a single test would trigger an error like:

```
ActiveRecord::StatementInvalid: Mysql2::Error: Unknown column 'user' in 'field list': INSERT INTO `kubernetes_release_groups` (`user`, `build`) VALUES ('deployer', 'docker_build')
    test/support/timeout.rb:13:in `block in capture_exceptions'
    test/support/timeout.rb:12:in `capture_exceptions'
```

* Fixes a problem that if there was an error loading fixtures, each test would also spit out an error like:

```
TypeError: no implicit conversion of nil into Array
    test/test_helper.rb:83:in `extra_threads'
    test/test_helper.rb:79:in `kill_extra_threads'
    test/test_helper.rb:75:in `ensure in fail_if_dangling_threads'
    test/test_helper.rb:75:in `fail_if_dangling_threads'
    test/test_helper.rb:69:in `block in <class:TestCase>'
```

/cc @zendesk/samson, @msufa, @sbrnunes

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,10 +6,7 @@ require File.expand_path('../config/application', __FILE__)
 Samson::Application.load_tasks
 
 Rake::Task["default"].clear
-
-Rails::TestTask.new(:default) do |t|
-  t.pattern = "{test,plugins/*/test}/**/*_test.rb"
-end
+task default: :test
 
 namespace :plugins do
   Rails::TestTask.new(:test) do |t|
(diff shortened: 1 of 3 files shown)


msufa:
:+1:

PR #593, opened 2015-10-16
When using Pipeline, deploy the reference rather than the SHA to the next stage

pschambacher:
/cc @zendesk/samson 

### Description
At the moment when you configure pipeline deployments, it will try to deploy the SHA of the commit to the next stage. For one it is a bit ugly and confusion when seing all stages but also some of our stages prevent to deploy a SHA and you have to deploy a TAG so the deploy is failing. This will try to deploy the reference rather than the SHA.

### Steps to reproduce
 - setup pipeline
 - deploy "master" to the first stage
 - see that "master" was deployed to the next stage

### Steps to merge
* [x] :+1: from the team

### Risks
* Low: pipeline deploy failing because there is no reference?

### DIFF
--- a/plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
+++ b/plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
@@ -17,7 +17,7 @@ module SamsonPipelines
     private
 
     def start_next_stage(next_stage, current_job, deploy_service, output)
-      deploy = deploy_service.deploy!(next_stage, reference: current_job.deploy.commit)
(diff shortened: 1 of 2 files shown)


henders:
Nice, mind adding a unit test for this change?

henders:
:+1:  on :green_heart:

jonmoter:
If you deploy a branch (like `master`), isn't it possible that by the time the next stage is executed, `master` points to a different commit?

pschambacher:
@jonmoter Yeah that could happen actually...

PR #594, opened 2015-10-16
Do not override next_stage method in pipeline

pschambacher:
/cc @zendesk/samson 

### Description
By redefining the ```next_stage``` method in pipeline, when deploying to a stage, the first stage listed in the pipeline will be displayed on the "Deploy to next stage" button. This is probably not what we want because the deploy is gonna happen automatically anyway.
Without this method redefined, the button will display the next stage in the list for this project.

### Steps to reproduce
 - have 3 stages, A, B and C in this order
 - have A pipeline deploy to C
 - deploy to stage A
 - when deploy is successful, you should have a button to deploy to stage B

### Steps to merge
* [x] :+1: from the team

### Risks
* None


henders:
Fair point! :)
:+1:

PR #596, opened 2015-10-16
aditional exports hook

lfalcao:
Hello,

We're using Samson here for deployment (512 deploys until now) and it's running pretty well. But we also want to run ansible playbooks, database migrations and blue&green deploys on it.

This PR aims to let you export additional variables (through plugins) during deployment. We're using it on an ansible plugin here to handle different tags for each deploy. Thanks :)

### DIFF
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -24,7 +24,7 @@ module Samson
       :before_docker_build,
       :after_docker_build,
       :after_job_execution,
-      :additional_exports
+      :job_additional_vars
     ].freeze
 
     INTERNAL_HOOKS = [ :class_defined ]
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -23,7 +23,8 @@ module Samson
       :after_deploy,
       :before_docker_build,
       :after_docker_build,
-      :after_job_execution
+      :after_job_execution,
+      :additional_exports
     ].freeze
 
     INTERNAL_HOOKS = [ :class_defined ]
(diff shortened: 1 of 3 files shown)


jonmoter:
Hi, thanks for the contribution.

Would you add a test for this that confirms the hooks work properly?  Maybe something like [this test](https://github.com/zendesk/samson/blob/master/test/models/job_execution_test.rb#L117) in `job_executinon_test.rb`, except it adds (or overwrites) some environment variable via the hook, and asserts they appear in the job output.

lfalcao:
Sure, i'll update the PR later today.

lfalcao:
PR got updated by mistake. The commit #169260b53ae9dffcce846dbdfea36c7a4a7a6aca adds correct icons and urls to the projects.

screen shot 2015-10-19 at 12 41 05 pm

lib/samson/hooks.rb
```diff
@@ -23,7 +23,8 @@ class UserError < StandardError
       :after_deploy,
       :before_docker_build,
       :after_docker_build,
-      :after_job_execution
+      :after_job_execution,
+      :additional_exports
```
jonmoter: Rather than `:additional_exports`, would you rename the hook to something like `:job_env_vars`.  I'd like to have `job` in the name, so it's clear it's related to job execution, and that makes it clear to me that it involves the environment variables you set in the job's process.

jonmoter:
Could you break the icon change into a separate PR, to make it clear it's separate functionality?

Other than my one comment on renaming the hook, this looks good.  Thanks!

lfalcao:
done. :wink:

jonmoter:
Thanks!  :+1:

PR #597, opened 2015-10-16
Update status of Kubernetes release objects based on watching API

jonmoter:
After ReplicationControllers are crated in Kubernetes, watch the status of the Pods until they're in a "Ready" state, and update the status of the `Kubernetes::Release*` objects.

We still need to add logic to watch the `Events` stream for failures to schedule pods.
This also needs tests.  The Kubernetes plugin code is definitely under-tested.

/cc @msufa, @sbrnunes, @henders 

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-39

### Risks
 - Minimal risk, since logic is all localized to Kubernetes plugin.

### DIFF
--- a/plugins/kubernetes/app/models/kuber_deploy_service.rb
+++ b/plugins/kubernetes/app/models/kuber_deploy_service.rb
@@ -44,13 +44,48 @@ class KuberDeployService
     # TODO: hunt down this issue and fix it properly
     Watchers::PodWatcher ; Watchers::ReplicationControllerWatcher ; Watchers::EventWatcher
 
-    # Commented out for now, to avoid creating threads without ever destroying them
-    # Thread.new { kuber_release.watch_pods }
-    # Thread.new { kuber_release.watch_rcs }
+    Thread.new do
+      kuber_release.watch_pods(&method(:handle_pod_notice))
+    end
+
     # Thread.new { sleep(5) ; kuber_release.watch_pod_events }
   end
 
   private
+
+  def handle_pod_notice(notice, watcher)
+    release_doc = kuber_release.docs_by_role[notice.role]
+
+    if release_doc.nil?
+      Kubernetes::Util.log "ERROR: could not find role", role: notice.role, available_roles: kuber_release.docs_by_role.keys
+      return
+    end
+
+    ready_count = watcher.num_ready
+    if ready_count > release_doc.replicas_live
+      release_doc.update_replica_count(ready_count)
+      release_doc.save!
+      update_release_status
+      log 'Another replica is live', role: notice.role, count: ready_count, r_status: kuber_release.status, r_doc_status: release_doc.status
+    end
+  end
+
+  def update_release_status
+    statuses = kuber_release.release_docs.map(&:status)
+
+    if statuses.all? { |s| s == 'live' }
+      deploy_is_live
+    elsif statuses.include?('spinning_up')
+      kuber_release.update_attribute(:status, 'spinning_up') unless kuber_release.status == 'spinning_up'
+    end
+  end
+
+  def deploy_is_live
+    kuber_release.release_is_live
+    kuber_release.save!
+
+    kuber_release.stop_watching   # this should kill the #watch_pods Thread spawned above
+  end
 
   def log(msg, extra_info = {})
     extra_info.merge!(
--- a/plugins/kubernetes/app/models/kuber_deploy_service.rb
+++ b/plugins/kubernetes/app/models/kuber_deploy_service.rb
@@ -71,12 +71,10 @@ class KuberDeployService
   end
 
   def update_release_status
-    statuses = kuber_release.release_docs.map(&:status)
-
-    if statuses.all? { |s| s == 'live' }
+    if kuber_release.release_docs.all?(&:live?)
       deploy_is_live
-    elsif statuses.include?('spinning_up')
-      kuber_release.update_attribute(:status, 'spinning_up') unless kuber_release.status == 'spinning_up'
+    elsif kuber_release.release_docs.any?(&:spinning_up?)
+      kuber_release.update_attribute(:status, 'spinning_up') unless kuber_release.spinning_up?
     end
   end
 
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -2,7 +2,7 @@ module Kubernetes
   class Release < ActiveRecord::Base
     self.table_name = 'kubernetes_releases'
 
-    VALID_STATUSES = %w[created spinning_up live spinning_down dead]
+    STATUSES = %w[created spinning_up live spinning_down dead]
 
     belongs_to :release_group, class_name: 'Kubernetes::ReleaseGroup', foreign_key: 'kubernetes_release_group_id', inverse_of: :releases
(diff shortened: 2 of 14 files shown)


plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -44,14 +44,49 @@ def watch_deployment
...
+    if ready_count > release_doc.replicas_live
+      release_doc.update_replica_count(ready_count)
+      release_doc.save!
+      update_release_status
+      log 'Another replica is live', role: notice.role, count: ready_count, r_status: kuber_release.status, r_doc_status: release_doc.status
+    end
+  end
+
+  def update_release_status
+    statuses = kuber_release.release_docs.map(&:status)
```
msufa: Some time ago somebody suggested that kuber_release.release_docs.pluck(:status) looks better ;)

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -44,14 +44,49 @@ def watch_deployment
...
+
+    if statuses.all? { |s| s == 'live' }
```
msufa: It would be cleaner to have an enum for all possible statuses.

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -10,10 +10,15 @@ class Release < ActiveRecord::Base
...
     validates :release_group, presence: true
     validates :deploy_group, presence: true
-    validates :status, inclusion: VALID_STATUSES
+    validates :status, inclusion: STATUSES
 
     after_initialize :set_default_status, on: :create
 
+    def release_is_live
+      self.status = 'live'
+      self.deploy_finished_at = Time.now
```
sbrnunes: Is `self.` required in here? We're affecting a concrete instance, right?

plugins/kubernetes/app/models/kuber_deploy_service.rb
henders: Or at least symbols?

plugins/kubernetes/app/models/kubernetes/release.rb
henders: This is generally done with activerecord models, but I'm not sure of the reasoning behind it. I think it's because activerecord defines a operator= method override for all columns in table?

henders:
:+1:  after minor quibbles.

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -44,14 +44,49 @@ def watch_deployment
...
+      deploy_is_live
+    elsif statuses.include?('spinning_up')
+      kuber_release.update_attribute(:status, 'spinning_up') unless kuber_release.status == 'spinning_up'
```
sbrnunes: Should we be asking the`kuber_release` if it's on a `spinning up` state instead of making this comparison, i.e., `kuber_release.spinning_up?`.

If we need to make this same comparison somewhere else, it would be nice to have this logic centralised in the `Kubernetes::Release` object.

plugins/kubernetes/app/models/kubernetes/release.rb
sbrnunes: Nevermind, I just checked that, yes, it is required when calling a self write accessor. My java background was playing tricks to me.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -79,6 +81,16 @@ def stop_watching
       @watcher.stop_watching if @watcher
     end
 
+    def update_replica_count(new_count)
+      self.replicas_live = new_count
+
+      if replicas_live >= replica_target
```
sbrnunes: In practice, we should never get more than the `replica_target`, right?

plugins/kubernetes/app/views/kubernetes_release_groups/new.html.erb
```diff
@@ -37,7 +37,7 @@
               <%= build.project.name %> - <%= build.label %>
             </p>
           <% elsif @project.builds.any? %>
-            <%= form.select :build_id, @project.builds.order('id desc').limit(20).map { |b| [b.label, b.id] } %>
+            <%= form.select :build_id, @project.builds.order('id desc').limit(20).map { |b| [b.label, b.id] }, {}, { class: 'form-control' } %>
```
sbrnunes: The use of `@project` in here made me realise that the whole Kubernetes plugin is still not up-to-date with the changes introduced to the project level access control. We'll need to update the `before_action` filters in the controllers and remove the local `project` methods.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
jonmoter: In theory, we shouldn't.

plugins/kubernetes/app/views/kubernetes_release_groups/new.html.erb
jonmoter: Okay, I'll make a follow-up PR to bring the Kubernetes plugin up to date with those changes.

plugins/kubernetes/app/models/kuber_deploy_service.rb
jonmoter: `pluck` works differently. It always causes a database query, and doesn't create any `ActiveRecord` models.  It's equivalent to just saying `SELECT status from kuberentes_release_docs WHERE kubernetes_release_id = #{release_id}`, and turning the results into an array.

If I say `kuber_release.release_docs` here, it can take care of the association caching that Rails does. Once it has pulled the rows out of the DB and constructed the models, it keeps them in a cache, and doesn't need to go back to the database for subsequent calls.

So they're both useful, depending on the scenario.  :-)

plugins/kubernetes/app/models/kuber_deploy_service.rb
jonmoter: The set of valid values for status are defined in `Kubernetes::Release::STATUSES`.  Ruby doesn't have a built-in Enum construct in the way that some other languages do.  And in the database, the `status` field is stored as a string.

So I have a few options:
1. Define constants, so it would look like `statuses.all? { |s| s == Kubernetes::ReleaseDoc::STATUS_LIVE }`
2. Create a method for each status on the object itself, e.g. `statuses.all?(&:live)`
3. Override the `#status` method of ReleaseDoc so it returns a symbol instead of a string, and the `#status=` method so it accepts either a symbol or a string.

All 3 options seem like extra work for not a whole lot of gain, but I'm willing to be convinced otherwise.

plugins/kubernetes/app/models/kuber_deploy_service.rb
jonmoter: Yeah, we could do that, which is like option number 2 in my comment above.

plugins/kubernetes/app/models/kuber_deploy_service.rb
msufa: I was thinking about something similar to what we have for `Role` model.

jonmoter:
Added methods for checking status of `Release` and `ReleaseDoc` as suggested.

Will merge once tests are green.

PR #599, opened 2015-10-16
Add support for project level deployer role [SAMSON-169]

sbrnunes:
Adding support to allow an User with project role 'Deployer' to have deployer access to a project regardless of his system level role.

(This is effectively the same PR as #566, but merging to `master` instead of a different branch.)

/cc @zendesk/samson 

### References

### Risks

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -1,0 +1,21 @@
+module Searchable
+  extend ActiveSupport::Concern
+
+  included do
+    def self.search_by_criteria(criteria)
+      scope = self
+      scope = scope.search(criteria[:search]) if criteria[:search]
+      scope.order("#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
+    end
+
+    private
+
+    def self.sort_column(column)
+      column_names.include?(column) ? column : 'created_at'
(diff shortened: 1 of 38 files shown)


app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,21 @@
...
+    private
+
+    def self.sort_column(column)
+      column_names.include?(column) ? column : 'created_at'
+    end
+
+    def self.sort_direction(direction)
+      %w[asc desc].include?(direction) ? direction : 'asc'
+    end
+  end
```
jonmoter: FYI, if you want to define Class methods using `ActiveSupport::Concern`, the standard pattern is:

```ruby
module Searchable
  extend ActiveSupport::Concern

  module ClassMethods
    def search_by_criteria(criteria)
      # ...
    end
  end
end
```

See the `Permalinkable` module for an example.

jonmoter:
This code change was already thoroughly reviewed when it was PR #566. I gave it a quick once-over, and it looks good to me.  The style suggestion above would be a nice tweak, but is not critical.

:+1:

app/models/concerns/searchable.rb
sbrnunes: Changes applied.

PR #601, opened 2015-10-17
Send access request as current user, cc manager

msufa:
Use currently logged in user's email address as the sender for access request emails. It can then be used by tools like JIRA to properly set the ticket reporter.

Move manager email to CC field.

/cc @zendesk/samson @jonmoter 

### References

### Risks
 - None


henders:
:+1:

PR #602, opened 2015-10-19
use HTTPS to retrieve badges from img.shields.io

jonmoter:
Requests to img.shields.io via HTTP get redirected to HTTPS, and that generates a `RuntimeError: redirection forbidden`

/cc @zendesk/samson 

### References
 - Airbrake: https://zendesk.airbrake.io/projects/95346/groups/1533483716423658788


steved:
+1

PR #603, opened 2015-10-19
add some more badges [ci skip]

steved:
Dockerhub status has been flaky at best. Let's use shields.io. And add our codeclimate badges.

@zendesk/samson


jonmoter:
:+1:

PR #604, opened 2015-10-20
fix job status after execution setup fails

steved:
@zendesk/samson


jonmoter:
Nice catch.  Thanks.  :+1:

PR #605, opened 2015-10-20
Create Kubernetes Services as part of deploy

jonmoter:
When a project is deployed to Kubernetes for the first time, a Service should be created for every `Role` that defines a service.

This PR creates a `Kubernetes::Service` object that wraps some functionality of the Service abstraction. Right now it's just a ruby class, but I'm considering adding a table in the DB to keep track of them. I need to give that a little more thought.

It also adds logic to `KuberDeployService` to create the Service as part of deployment.

Sadly, this is still lacking in tests. I need to give some thought on how to make all this code more testable. I think we'll need some kind of Kubernetes testing framework, and maybe some integration tests.

/cc @henders, @msufa, @sbrnunes 

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-36

### Risks
 - Only affects Kubernetes plugin code, so minimal risk to primary Samson functionality.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role.rb
@@ -19,5 +19,13 @@ module Kubernetes
     def ram_with_units
       "#{ram}Mi" if ram.present?
     end
+
+    def has_service?
+      service_name.present?
+    end
+
+    def service_for(deploy_group)
+      Kubernetes::Service.new(role: self, deploy_group: deploy_group) if has_service?
+    end
   end
 end
--- a/plugins/kubernetes/app/models/kubernetes/service.rb
+++ b/plugins/kubernetes/app/models/kubernetes/service.rb
@@ -1,0 +1,95 @@
+module Kubernetes
+  # Convenience wrapper around the 'Service' object we can get from the K8S API
+  class Service
+    attr_reader :deploy_group, :role, :name, :ip_address, :port_info
+
+    def initialize(deploy_group: nil, role: nil)
+      @deploy_group = deploy_group
+      @role = role
+      @port_info = []
+      @name = role.try(:service_name)
+    end
+
+    def namespace
+      deploy_group.kubernetes_namespace
+    end
+
+    def running?
+      service_object.present?
+    end
+
+    def proxy_address
+      "#{example_node_ip}:#{node_ports.first}" if running?
+    end
+
+    def example_node_ip
+      node = client.get_nodes.first
+      node.spec.externalID if node
+    end
+
+    # The set of ports that are exposed on each Kubernetes minion. Traffic sent
+    # to this port on any minion will be proxied to the Pods in this service.
+    def node_ports
+      fetch_service if @service_object.nil?
+      @port_info.map(&:nodePort)
+    end
+
+    # Returns an array of OpenStructs containing the list of Pods that this
+    # service routes traffic to. Each object return has the following attributes:
+    #   `pod_ip` : IP address of the Pod
+    #   `pod_name` : name of the Pod
+    #   `ready?` : true if the Pod is ready and accepting traffic
+    def endpoints
+      return [] if endpoint_object.nil?
+
+      results = []
+
+      # https://htmlpreview.github.io/?https://github.com/kubernetes/kubernetes/HEAD/docs/api-reference/v1/definitions.html#_v1_endpoints
+      endpoint_object.subsets.each do |subset|
+        [:addresses, :notReadyAddresses].each do |attr|
+          next if subset[attr].blank?
+
+          subset[attr].each do |address|
+            if address.targetRef.kind == 'Pod'
+              results << OpenStruct.new(pod_ip: address.ip,
+                                        pod_name: address.targetRef.name,
+                                        ready?: (attr == :addresses))
+            end
+          end
+        end
+      end
+
+      results
+    end
+
+    def service_object
+      @service_object || fetch_service
+    end
+
+    def endpoint_object
+      @endpoint_object || fetch_endpoint
+    end
+
+    private
+
+    def fetch_service
+      @service_object = client.get_service(name, namespace)
+      @name = @service_object.metadata.name
+      @ip_address = @service_object.spec.clusterIP
+      @port_info = @service_object.spec.ports || []
+      @service_object
+    rescue KubeException => e
+      raise e unless e.error_code == 404
+      nil
+    end
+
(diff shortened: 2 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/service.rb
```diff
@@ -0,0 +1,95 @@
...
+      "#{example_node_ip}:#{node_ports.first}" if running?
+    end
+
+    def example_node_ip
+      node = client.get_nodes.first
+      node.spec.externalID if node
+    end
+
+    # The set of ports that are exposed on each Kubernetes minion. Traffic sent
+    # to this port on any minion will be proxied to the Pods in this service.
```
msufa: these ports

plugins/kubernetes/app/models/kubernetes/service.rb
```diff
@@ -0,0 +1,95 @@
...
+    def node_ports
+      fetch_service if @service_object.nil?
+      @port_info.map(&:nodePort)
+    end
+
+    # Returns an array of OpenStructs containing the list of Pods that this
+    # service routes traffic to. Each object return has the following attributes:
```
msufa: "Each object returned"

plugins/kubernetes/app/models/kubernetes/service.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    def running?
+      service_object.present?
+    end
+
+    def proxy_address
+      "#{example_node_ip}:#{node_ports.first}" if running?
+    end
+
+    def example_node_ip
```
msufa: What does "example" mean here? We just pick one node from the ones that are available and we don't care which?

plugins/kubernetes/app/models/kubernetes/service.rb
```diff
@@ -0,0 +1,95 @@
...
+    #   `pod_name` : name of the Pod
+    #   `ready?` : true if the Pod is ready and accepting traffic
+    def endpoints
+      return [] if endpoint_object.nil?
+
+      results = []
+
+      # https://htmlpreview.github.io/?https://github.com/kubernetes/kubernetes/HEAD/docs/api-reference/v1/definitions.html#_v1_endpoints
+      endpoint_object.subsets.each do |subset|
+        [:addresses, :notReadyAddresses].each do |attr|
```
msufa: "attr" is a Module method, some code editors treat it as a keyword (e.g. the GitHub syntax highlighter), it's a minor issue but you might consider a different name.

plugins/kubernetes/app/models/kubernetes/service.rb
```diff
@@ -0,0 +1,95 @@
+module Kubernetes
+  # Convenience wrapper around the 'Service' object we can get from the K8S API
+  class Service
+    attr_reader :deploy_group, :role, :name, :ip_address, :port_info
+
+    def initialize(deploy_group: nil, role: nil)
+      @deploy_group = deploy_group
+      @role = role
+      @port_info = []
+      @name = role.try(:service_name)
```
msufa: Service is now coupled with Role just to get the service name (role doesn't seem to be used anywhere else). Would it be simpler just to pass the name to the constructor?

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -19,5 +19,13 @@ def label_name
     def ram_with_units
       "#{ram}Mi" if ram.present?
     end
+
+    def has_service?
+      service_name.present?
```
msufa: We have a small inconsistency here. Currently even if a service is defined in the template, but the user does not specify the name when creating a role, we will assume that no service should be deployed. It would be better to pre-populate this from the template, or solely rely on the template.

plugins/kubernetes/app/models/kubernetes/service.rb
```diff
@@ -0,0 +1,95 @@
...
+      @service_object || fetch_service
+    end
+
+    def endpoint_object
+      @endpoint_object || fetch_endpoint
+    end
+
+    private
+
+    def fetch_service
```
henders: Should probably make this a lazy-loading function properly for @service_object, and make name, ip_address, etc... lazy accesses as well.

But I foresee a fair amount of refactoring coming to this code anyway when we start adding tests. So we can possibility ignore this for now.

plugins/kubernetes/app/models/kubernetes/service.rb
henders: Nice catch!

henders:
:+1:  after comments addressed

plugins/kubernetes/app/models/kubernetes/service.rb
jonmoter: True, that would make more sense.  Good call.

plugins/kubernetes/app/models/kubernetes/service.rb
jonmoter: Yeah. In Kubernetes, any of the nodes acts as a proxy/load-balancer to any of the other nodes. There's no special server that has that function.  So if you want to send traffic to the pods that are part of a service, and the service has exposed a `NodePort`, you send a request to that port on any of the nodes (aka minions).

I created this method for the `proxy_address` method, which is nice to display in the UI so that a user in dev mode can see what the address is for the app.

plugins/kubernetes/app/models/kubernetes/role.rb
jonmoter: Yeah, I'm not entirely happy with this either.  It's important that the name of the service stay consistent, because the service name is the value that defines the DNS lookup and the ENV vars that other applications will use. Also, it's important that service names be unique, so we should add a check to make sure no two `Kubernetes::Roles` in Samson try using the same service name. That's why I want it to be a column in the database.

But if someone makes a change to to the template file and gives the service a different name, what should we do?

I' still puzzling over this.

plugins/kubernetes/app/models/kubernetes/service.rb
jonmoter: Possibly.  But then the `Service` object would have no connection to the `Project`, which doesn't seem right to me.

I actually created a version of this code where `Service` was an `ActiveRecord::Base` subclass, and existed as a table in the database.  I'm still going back and forth on whether that's a good idea, so that you could more easily query all the Services running.  But since I wasn't sure about that, I started with the simplest implementation, which was this.

jonmoter:
Comments addressed.  Will merge on green build.

PR #607, opened 2015-10-21
Respect RACK_ENV when choosing to load bootscale

sj26:
`bundle --deployment --without development:test` then loading the rack app with `RACK_ENV=production` would fail because this line required the bootscale gem which is only in the development and test groups and was only checking `RAILS_ENV`.

This changes the logic to be the same as `Rails.env` (as of writing):

https://github.com/rails/rails/blob/0d216d1add9eaaddfc0b02813ccf08fd22910859/railties/lib/rails.rb#L70


steved:
Thanks!

PR #609, opened 2015-10-22
Ensure access request flag gets cleared for project level access updates

msufa:
The flag was cleared only for global access rights changes. This prevented the user to request access rights for additional projects after the initial request got serviced.

/cc @zendesk/samson @sbrnunes 

### References

### Risks
 - None

### DIFF
--- a/app/controllers/project_roles_controller.rb
+++ b/app/controllers/project_roles_controller.rb
@@ -14,7 +14,7 @@ class ProjectRolesController < ApplicationController
 
     if new_role.persisted?
       Rails.logger.info("#{current_user.name_and_email} granted the role of #{new_role.role.display_name} to #{new_role.user.name} on project #{current_project.name}")
-      reset_access_request_flag(new_role.user_id)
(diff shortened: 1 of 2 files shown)


app/controllers/project_roles_controller.rb
```diff
@@ -39,4 +41,8 @@ def create_params
   def update_params
     params.require(:project_role).permit(:role_id)
   end
+
+  def reset_access_request_flag(user_id)
+    User.find(user_id).update!(access_request_pending: false)
```
sbrnunes: I think you might be able to do pass `new_role.user` as argument and then do:
```
user.update!(...)
```

Sorry, I mentioned the user_id before but I just noticed that you should be able to get the User from the UserProjectRole.

sbrnunes:
After bumping the Travis build and checking that little improvement: :+1:

msufa:
Thanks @sbrnunes for the user object suggestion, I posted an update to the PR.

PR #610, opened 2015-10-22
Added caching when loading project roles to prevent delays when rendering the radio buttons

sbrnunes:
Added some caching support when loading the project roles catalog to prevent delays in the UI when showing the radio buttons. Because each row was fetching the project roles from the backend, some annoying delays were visible.

/cc @zendesk/samson 

### Steps to reproduce
By clicking on a User after accessing the Admin -> Users page, one could see how long it was taking to show the radio buttons for each project row.

### Risks
 - None


app/assets/javascripts/services/project_roles_svc.js
```diff
@@ -1,6 +1,26 @@
...
+  /*
+    Creates a cache for project roles to prevent access to the backend each time we need to load the project roles,
+    and corresponding delays to show the data in the browser. Only first request will hit the backend.
+    Data stored in the browser session storage (when available). If not, it will be stored in memory.
+    Cache will expires after 5 minutes (this is a catalog, so, not expect to change that often).
+   */
+  if (!CacheFactory.get('project_roles_cache')) {
+    CacheFactory.createCache('project_roles_cache', {
+      deleteOnExpire: 'passive', //expired items deleted on access
+      maxAge: 5 * 60 * 1000, //expires after 10 minutes
```
msufa: I see a little lie in the comment :) The cache expires after 5 minutes.

henders:
This seems like a complicated solution. Why do you only draw the user roles for each project 1 at a time? 

Was a single request to get all user roles for all projects too slow?

You draw all the projects up front and also get the types of roles up front, so at minimum draw in controls for the role selection even if you don't know the value yet.

sbrnunes:
@henders, what is getting loaded for each project it's the actual role types, i.e., the labels and values we use for each radio button: "Deployer", "Admin".

Answering your question, the radio buttons are rendered as a partial, for each project, and this partial is reused by the Users tab in the project settings, in which case it gets rendered for each User. Inside the partial, I'm creating a new Controller with it's own scope. Think of this as a self contained component, something like an Angular directive (it should have been one, actually). So, for each row on this tables, I need to load the available project roles.

The only alternative I see to achieve what you were suggesting is a parent controller, at the table level, because a child scope can have access to it's parent scope, if the controller is nested. But it would be even more complicated (a whole new controller) and I actually don't like that approach, as we would be moving code to that parent controller that is responsibility of this "components", and we would be coupling the two controllers.

The caching option looked like a good option to me. I started by trying to set the Cache-Control headers at the controller level, taking advantage of the Rails middleware for caching, but the requests are done concurrently, meaning, the first page would always be slow.

Why do you think this is a complicated solution? I can be wrong, but I believe when you have concurrent requests that you need to cache on the client side, this type of solution becomes pretty standard.

app/assets/javascripts/services/project_roles_svc.js
sbrnunes: Damn, I meant to delete this comments after adding the longer description.

Good catch.

sbrnunes:
Actually, there would be another alternative, perhaps more aligned with what you were suggesting: I could load the ProjectRoles up front, before the whole page is rendered, and use that collection to render the radio buttons. That would solve the issue but it would, not only, change the entire implementation of this component, but also force us to use a mix of client-side AngularJS code with server-side rendering in Rails on that little template we use for rendering the radio buttons. I've been trying to avoid this and stick with client-side logic whenever possible.

We can maybe discuss it better in person, but let me know your thoughts on this, because I really don't see another way of doing it without going back to use server-side rendering of this components.

sbrnunes:
@henders, apart from the missing test for the new directive, I've finished the changes we've discussed. When you have some time to spare, can you take a look at this PR, please? Thanks.

app/views/admin/users/show.html.erb
```diff
@@ -13,7 +13,7 @@
 
   <%= render partial: "shared/search_bar", locals: {url: admin_user_path, query: params[:search]} %>
 
-  <table id="users" class="table table-hover table-condensed">
+  <table id="users" class="table table-hover table-condensed" ng-controller="ProjectRolesCtrl" ng-init="loadProjectRoles()">
```
henders: As per the docs: https://docs.angularjs.org/api/ng/directive/ngInit
Just move this initialization into the controller itself. 

I.e., just call the loadProjectRoles() function at the bottom of the controller source itself.

app/views/users/index.html.erb
```diff
@@ -5,7 +5,7 @@
 
   <%= render partial: "shared/search_bar", locals: {url: project_users_path, query: params[:search]} %>
 
-  <table class="table table-hover table-condensed">
+  <table class="table table-hover table-condensed" ng-controller="ProjectRolesCtrl" ng-init="loadProjectRoles()">
```
henders: Same comment as above

app/assets/javascripts/directives/project_roles_directive.js
```diff
@@ -0,0 +1,13 @@
+samson.directive('projectRoles', function(userProjectRoleFactory) {
+  return {
+    restrict: 'E',
+    templateUrl: 'shared/_project_roles.tmpl.html',
+    scope: {
+      roles: '=',
+      roleChanged: '&'
+    },
+    link: function($scope, element) {
+      $scope.project_role = userProjectRoleFactory.build(element[0]);
```
henders: This looks odd. Not sure what element[0] is supposed to correspond to, the root element of this directive? 
Is there another way to code this to be clearer to read?

app/assets/javascripts/services/project_roles_svc.js
```diff
@@ -1,6 +1,26 @@
...
+    Data stored in the browser session storage (when available). If not, it will be stored in memory.
+    Cache will expires after 5 minutes (this is a catalog, so, not expect to change that often).
+   */
+  //if (!CacheFactory.get('project_roles_cache')) {
+  //  CacheFactory.createCache('project_roles_cache', {
+  //    deleteOnExpire: 'passive', //expired items deleted on access
+  //    maxAge: 5 * 60 * 1000, //expires after 10 minutes
+  //    storageMode: 'sessionStorage' //stored
+  //  });
+  //}
```
henders: Remember to delete all this :)

app/assets/javascripts/directives/project_roles_directive.js
sbrnunes: I know, it really looks like a code smell. Tried to find a way to avoid it but with no success. The current element should be just that, the element, not an array with the root element of the directive. :(

app/assets/javascripts/services/project_roles_svc.js
sbrnunes: Damn, thanks. Had forgotten about that.

app/views/admin/users/show.html.erb
sbrnunes: You're absolutely right, I kept it from what I had previously, but there's no need to do it anymore. I'm going to change it.
Thanks for the heads up.

sbrnunes:
@henders, @msufa, I've fixed the things you guys highlighted. Feel free to take a final look at this when you get a few minutes to spare. Thanks.

henders:
:+1:

PR #611, opened 2015-10-22
pending label should match timeline.js

steved:
@zendesk/samson 

The default is info.

https://github.com/zendesk/samson/blob/master/app/assets/javascripts/timeline.js#L30-L40


jonmoter:
:+1:

PR #612, opened 2015-10-23
Add headers to API for deploys

jacobat:
When creating a new deploy it is nice to know the location of the deploy as
well as getting a proper status code (201 Created) so API clients can more
easily be implemented.


jonmoter:
Looks like you need to fix one other test.

:+1: on :green_heart:

PR #613, opened 2015-10-23
Select role when requesting additional access rights

msufa:
Since we are now requesting roles per project, it makes sense to let the user specify which role they need.

screen shot 2015-10-23 at 14 24 10

/cc @zendesk/samson

### References

### Risks
 - None

### DIFF
--- a/app/controllers/access_requests_controller.rb
+++ b/app/controllers/access_requests_controller.rb
@@ -8,7 +8,7 @@ class AccessRequestsController < ApplicationController
   def new
     session[:access_request_back_to] ||= request.referer
     @projects = Project.all.order(:name)
-    @roles = Role.all[1...-1] # ignore viewer (default) and super_admin (no such role per project)
+    @roles = ProjectRole.all
   end
 
   def create
--- a/app/controllers/access_requests_controller.rb
+++ b/app/controllers/access_requests_controller.rb
@@ -8,11 +8,14 @@ class AccessRequestsController < ApplicationController
   def new
     session[:access_request_back_to] ||= request.referer
     @projects = Project.all.order(:name)
+    @roles = Role.all[1...-1] # ignore viewer (default) and super_admin (no such role per project)
   end
 
   def create
     AccessRequestMailer.access_request_email(
-        request.base_url, current_user, params[:manager_email], params[:reason], params[:project_ids]).deliver_now
+        request.base_url, current_user,
+        params.require(:manager_email), params.require(:reason),
+        params.require(:project_ids), params.require(:role_id)).deliver_now
     current_user.update!(access_request_pending: true)
     flash[:success] = 'Access request email sent.'
     redirect_to session.delete(:access_request_back_to)
--- a/app/mailers/access_request_mailer.rb
+++ b/app/mailers/access_request_mailer.rb
@@ -1,10 +1,11 @@
 class AccessRequestMailer < ApplicationMailer
-  def access_request_email(host, user, manager_email, reason, project_ids)
+  def access_request_email(host, user, manager_email, reason, project_ids, role_id)
     @host = host
     @user = user
     @manager_email = manager_email
     @reason = reason
     @projects = Project.order(:name).find(project_ids)
+    @role = Role.find(role_id)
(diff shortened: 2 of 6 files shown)


henders:
For visual PRs like this, it's good to have a screenshot so reviewers have some context! :)

app/controllers/access_requests_controller.rb
```diff
@@ -8,11 +8,14 @@ def self.feature_enabled?
   def new
     session[:access_request_back_to] ||= request.referer
     @projects = Project.all.order(:name)
+    @roles = Role.all[1...-1] # ignore viewer (default) and super_admin (no such role per project)
```
sbrnunes: If here what you want is the Project Roles, then instead of `Role` you'll need to use `ProjectRole`. If you use `ProjectRole`, you won't need to use the range to filter out `Viewer` and `Supe Admin`.

app/mailers/access_request_mailer.rb
```diff
@@ -1,10 +1,11 @@
 class AccessRequestMailer < ApplicationMailer
-  def access_request_email(host, user, manager_email, reason, project_ids)
+  def access_request_email(host, user, manager_email, reason, project_ids, role_id)
     @host = host
     @user = user
     @manager_email = manager_email
     @reason = reason
     @projects = Project.order(:name).find(project_ids)
+    @role = Role.find(role_id)
```
sbrnunes: Same as above. Actually, be careful, because in the context of a project, the available roles are:
- Deployer, with id: 0
- Admin: with id: 1

test/controllers/access_requests_controller_test.rb
```diff
@@ -55,7 +55,9 @@
       describe 'enabled' do
         let(:manager_email) { 'manager@example.com' }
         let(:reason) { 'Dummy reason.' }
-        let(:request_params) { {manager_email: manager_email, reason: reason, project_ids: Project.all.pluck(:id)} }
+        let(:role) { Role::DEPLOYER }
+        let(:request_params) {
+          {manager_email: manager_email, reason: reason, project_ids: Project.all.pluck(:id), role_id: role.id} }
```
henders: Missing space after opening '{'

test/controllers/access_requests_controller_test.rb
```diff
@@ -55,7 +55,9 @@
       describe 'enabled' do
         let(:manager_email) { 'manager@example.com' }
         let(:reason) { 'Dummy reason.' }
-        let(:request_params) { {manager_email: manager_email, reason: reason, project_ids: Project.all.pluck(:id)} }
+        let(:role) { Role::DEPLOYER }
```
sbrnunes: And on the tests, just to keep it consistent, same thing: we need to change it to ProjectRole.

test/mailers/access_request_mailer_test.rb
```diff
@@ -10,6 +10,7 @@
     let(:hostname) { 'localhost' }
     let(:manager_email) { 'manager@example.com' }
     let(:reason) { 'Dummy reason.' }
+    let(:role) { Role::DEPLOYER }
```
sbrnunes: Same here.

test/mailers/previews/access_request_mailer_preview.rb
```diff
@@ -5,7 +5,7 @@ class AccessRequestMailerPreview < ActionMailer::Preview
   def access_request_email
     enable_access_request
     email = AccessRequestMailer.access_request_email(
-        'localhost', User.first, 'manager@example.com', 'Dummy reason.', Project.all.pluck(:id))
+        'localhost', User.first, 'manager@example.com', 'Dummy reason.', Project.all.pluck(:id), Role::DEPLOYER.id)
```
sbrnunes: Same here.

test/controllers/access_requests_controller_test.rb
msufa: Isn't the convention to have spaces for blocks but none for hashes?

sbrnunes:
As soon as you switch over from Role to ProjectRole, :+1:

msufa:
Thanks @sbrnunes, you're a living proof that there's value in code reviews :) Will update the PR shortly...with screenshots!

PR #615, opened 2015-10-28
Fix create Deploy Group when Kubernetes is not enabled

sbrnunes:
When the Kubernetes plugin is disabled, an error is displayed when we try to access the page that allows us to create a new Deploy Group.

/cc @zendesk/samson

### DIFF
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -9,7 +9,7 @@ class Admin::DeployGroupsController < ApplicationController
 
   def new
     @deploy_group = DeployGroup.new
-    @deploy_group.build_cluster_deploy_group if allow_kuber_cluster_assignment?
+    Samson::Hooks.fire(:edit_deploy_group, @deploy_group)
     render 'edit'
   end
 
@@ -25,9 +25,7 @@ class Admin::DeployGroupsController < ApplicationController
   end
 
   def edit
-    if allow_kuber_cluster_assignment?
-      @deploy_group.build_cluster_deploy_group unless @deploy_group.cluster_deploy_group
-    end
+    Samson::Hooks.fire(:edit_deploy_group, @deploy_group)
   end
 
   def update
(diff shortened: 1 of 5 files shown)


sbrnunes:
@henders, @jonmoter, can one of you guys review this fix, please? Thanks.

henders:
Hmm... it seems wrong that Samson core should have to know about a plugin.

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -70,7 +72,7 @@ def build_kuber_cluster
     @deploy_group.build_cluster_deploy_group unless @deploy_group.cluster_deploy_group
   end
```
henders: Looks like this function isn't used anywhere, so easy to refactor out :)

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -9,7 +9,9 @@ def index
 
   def new
     @deploy_group = DeployGroup.new
-    @deploy_group.build_cluster_deploy_group
+    if allow_kuber_cluster_assignment?
+      @deploy_group.build_cluster_deploy_group
+    end
```
henders: ```@deploy_group.build_cluster_deploy_group if allow_kuber_cluster_assignment?```

henders:
:+1:  after 2 minor code changes. Just to get this out to production faster.

app/controllers/admin/deploy_groups_controller.rb
sbrnunes: @henders , which function? `build_cluster_deploy_group`?

What do you mean with not used anywhere?

sbrnunes:
@henders, I've committed on of the minor fixes you suggested. The other I didn't understand what you meant.

I'd like to review with you how to make the core app agnostic of the plugin. But, for now, I'm going to merge it to be able to ship it to production, if that's ok.

app/controllers/admin/deploy_groups_controller.rb
jonmoter: We can probably remove the `Admin::DeployGroupsController#build_kuber_cluster` method, since I don't think that's used anywhere.

jonmoter:
Looks like this will fix the issue.  :+1:

app/controllers/admin/deploy_groups_controller.rb
sbrnunes: Oh, I see it now. I was looking into another line. Sorry.

sbrnunes:
Sorry guys, I know this already had a :+1: , but after discussing it a little further with @henders, I took another approach using Samson hooks to avoid having references to the Kubernetes plugin in the core DeployGroupsController.

Can you guys take a final look at it please?

Tested with and without the Kubernetes plugin enabled.

plugins/kubernetes/app/models/concerns/kubernetes/deploy_group_permitted_params.rb
```diff
@@ -0,0 +1,15 @@
+module Kubernetes::DeployGroupPermittedParams
```
sbrnunes: @henders, I was not sure about this concern, but I think we still need it because we need to resolve which parameters are required and permitted and only then we can append the `_destroy` flag to the parameters. I couldn't find a better way to do this in the `deploy_group_permitted_params` callback bellow.

Well, I could have defined yet another event to append this flag to the parameters list, after resolving the required and permitted parameters. That would avoid the need of the decorator and the concern...

henders:
I :heart:  this much better :+1:

PR #616, opened 2015-10-28
Fix for environment variables

sbrnunes:
Environment Variables, in the Admin menu, was broken due to an obsolete route config that was mapping to an non-existing controller.

/cc @zendesk/samson


henders:
:+1:  on :green_heart:

msufa:
:+1:

PR #617, opened 2015-10-29
fix starring projects

sandlerr:
The `CurrentProject` concern used in the `StarsController` pulls out the `project_id` param, not the `id` param.

@zendesk/samson @miogalang @ryanseddon

### DIFF
--- a/app/controllers/concerns/current_project.rb
+++ b/app/controllers/concerns/current_project.rb
@@ -15,6 +15,7 @@ module CurrentProject
 
   def require_project
     @project = (Project.find_by_param!(params[:project_id]) if params[:project_id])
+    @project ||= (Project.find_by_param!(params[:id]) if params[:id])
   end
 
 end
--- a/app/helpers/projects_helper.rb
+++ b/app/helpers/projects_helper.rb
@@ -2,14 +2,14 @@ module ProjectsHelper
   def star_for_project(project)
     content_tag :span, class: 'star' do
       if current_user.starred_project?(project)
-        path = star_path(project_id: project)
+        path = star_path(id: project)
         options = {
           class: 'glyphicon glyphicon-star',
           method: :delete,
           title: 'Unstar this project'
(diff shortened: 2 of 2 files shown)


xuebinglee:
:+1: on :green_apple:

miogalang:
:+1:

sandlerr:
@xuebinglee @miogalang have changed the implementation (ignore the red from CodeClimate please)

miogalang:
:+1:

app/controllers/concerns/current_project.rb
```diff
@@ -15,6 +15,7 @@ def current_project
 
   def require_project
     @project = (Project.find_by_param!(params[:project_id]) if params[:project_id])
+    @project ||= (Project.find_by_param!(params[:id]) if params[:id])
```
sbrnunes: I think loading the project by id as a fallback might be dangerous and lead us to load the wrong project. In a URL using both `project_id` and `id` as path variables, we wouldn't want this to happen.

In another similar situation, we overrided the `require_project` method to match the specific need. See `ProjectsController` for an example of what I mean.

henders: This might cause problems on the stages controller, as it uses this function as well. I.e. If user tries to access a project page with a bad project_id param, it'll fall back to trying to find a project with the stages' param[:id].

I don't like it, but the solution is probably to do what the LocksController does and override the require_project.

henders: Ah, @sbrnunes got there before me.

msufa:
I've implemented the fix as @sbrnunes & @henders suggested in #620. I've also added some tests so that we catch similar issues early next time. You can close this PR.

henders:
:-1: as I think #620 fixes some edge conditions, and added tests. :smile_cat:

sandlerr:
Thanks guys

PR #618, opened 2015-10-29
dont error out if the manifest is not used for the samson env plugin

sandlerr:
We need to use a `manifest.json` because its part of the spec of a Zendesk Apps Marketplace app, but we don't need the SamsonEnv plugin to parse it at all. This is a quick fix.

@zendesk/quokka


svizzari:
:+1: 🐸

henders:
:+1:

PR #619, opened 2015-10-29
make the tests faster again

sandlerr:
The change in https://github.com/zendesk/samson/pull/604 made the tests run dramatically slower. As an aside, I'd be strongly in favour of removing the conditional [here](https://github.com/zendesk/samson/blob/master/test/support/timeout.rb#L5) so this doesn't happen in future.

screen shot 2015-10-29 at 5 16 07 pm

@zendesk/samson @steved @jonmoter

### DIFF
--- a/test/models/buddy_check_test.rb
+++ b/test/models/buddy_check_test.rb
@@ -65,6 +65,7 @@ describe BuddyCheck do
     before do
       stage.update_attribute(:production, true)
       job_execution.stubs(:execute!)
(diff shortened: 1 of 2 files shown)


sandlerr:
sqlite still seems kinda slow for this build but it is an improvement

henders:
Looks good! :+1:

grosser:
timeout fix https://github.com/zendesk/samson/pull/621

PR #600, opened 2015-10-16
Refactoring filters responsible for loading a project before actions get executed

sbrnunes:
Following the code review from pull request [#553](https://github.com/zendesk/samson/pull/553), before_action filters loading the "current project" have been removed and the "current project" is now a lazily initialised attribute, accessible through the helper method `current_project`, which has been defined in the module `CurrentProject`.

(This is effectively the same PR as #567, but merging to master instead of a different branch.)

/cc @zendesk/samson 

### References
Related to (or consequence of):
- [SAMSON-169](https://zendesk.atlassian.net/browse/SAMSON-169)

### Risks
- Changes to the behaviour regarding how the current project is loaded on several controllers (make sure all tests are passing and no functionality has been broken)

PR #620, opened 2015-10-29
Fix for starring projects (with tests)

msufa:
Fixing broken functionality of adding stars to projects (problem introduced with #600)

We did not catch it earlier because stars controller was lacking tests.

/cc @zendesk/samson @sbrnunes @henders 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/app/controllers/stars_controller.rb
+++ b/app/controllers/stars_controller.rb
@@ -13,4 +13,11 @@ class StarsController < ApplicationController
 
     head :ok
   end
+
+  private
+
+  def require_project
+    # override from CurrentProject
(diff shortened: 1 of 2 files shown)


henders:
:+1:

app/controllers/stars_controller.rb
```diff
@@ -13,4 +13,9 @@ def destroy
 
     head :ok
   end
+
+  def require_project
```
henders: Bonus points if making this method ```private```

msufa:
Made require_project private as suggested by @henders, will merge on :green_heart:

PR #621, opened 2015-10-30
timeout

grosser:
@sandlerr let's see if that works ...
alternative approach would be to use forking_test_runner to at least see what file was hanging ... or running with `-v`

### Risks
 - None


grosser:
@zendesk/samson rebuilt a few times and no random errors ... so I think the setup stubbing fixed the previous issue ...

grosser:
how about always stubbing out setup and only unstubbing it for certain tests?

henders:
This is a nice cleanup anyway and catching tests that leave threads lying around. :+1:

PR #622, opened 2015-11-02
PAAS-80 Create angular framework for K8s Project tab

sbrnunes:
PR providing just the initial framework and corresponding configuration to turn the Kubernetes tab into a full angular, client-side, page.

All existing functionality should be working as before.

/cc @zendesk/samson

### Tasks
 - [ ] Implement functionality above on the client-side with AngularJS
 - [ ] Implement dashboard tab
 - [ ] Finish styling

### References

### Risks


plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
```diff
@@ -0,0 +1,20 @@
...
+  text-transform: none;
+}
+
+.kubernetes .md-tab.md-active {
+  color: rgb(255,205,210);
+}
+
+/*.kubernetes .md-tabs {*/
+  /*transition: none;*/
+/*}*/
```
henders: If we don't need this then delete.

plugins/kubernetes/app/controllers/kubernetes_project_controller.rb
```diff
@@ -1,16 +1,15 @@
 class KubernetesProjectController < ApplicationController
-  before_action :authorize_deployer!
-  before_action :project
+  include ProjectLevelAuthorization
+  before_action :authorize_project_deployer!
```
henders: I'm wondering if the viewer should be able to see the dashboard we'll be making?

plugins/kubernetes/app/views/kubernetes_project/show.html.erb
```diff
@@ -1,10 +1,30 @@
 <%= render 'projects/header', project: @project, tab: 'kubernetes' %>
 
-
-<section class="tabs">
-  <h1>Project Roles</h1>
-  <%= render partial: 'kubernetes_roles/roles_list_show' %>
-
-  <h1>Kubernetes Releases</h1>
-  <%= render partial: 'kubernetes_release_groups/kuber_release_group_table' %>
```
henders: We can delete the old partials now as well, right?

plugins/kubernetes/app/controllers/kubernetes_project_controller.rb
sbrnunes: That's a good point. We could hide all tabs except the Dashboard, or show all the tabs but hide the buttons for creating new Roles and Releases.

Food for thoughts... I think we can squeeze that in easily, anyway, because all this files are ERBs.

plugins/kubernetes/app/views/kubernetes_project/show.html.erb
sbrnunes: They're still being used by the pages that get rendered when you access directly to: `/kubernetes_release_groups` and  `/kubernetes_roles`, e.g., following a `create`.

But yes, they're going to be removed soon.

henders:
Cool, :+1:

PR #623, opened 2015-11-03
Add @team mention in the default Flowdock message

codella:
I found myself to always change the default Flowdock message, in order to mention the @team and get more attention.

I wonder if this is a general felt need - in that case, this PR will fix that.

@zendesk/samson 

#### Risks
 * none

### DIFF
--- a/plugins/flowdock/app/helpers/flowdock_helper.rb
+++ b/plugins/flowdock/app/helpers/flowdock_helper.rb
@@ -1,5 +1,5 @@
 module FlowdockHelper
   def default_flowdock_message(deploy)
-    ":pray: #{deploy.user.name} is requesting approval to deploy **#{deploy.reference}** to production. [Approve this deploy](#{project_deploy_url(deploy.project, deploy)})."
+    ":pray: @team #{deploy.user.name} is requesting approval to deploy **#{deploy.reference}** to production. [Approve this deploy](#{project_deploy_url(deploy.project, deploy)})."
   end
 end


henders:
:+1: on :green_heart:

dasch:
:+1:

pschambacher:
I liked this emoji :pray:

codella:
@pschambacher it is still there, just at the end :^) want me to put it back at the beginning?

bquorning:
:+1:

codella:
@pschambacher I put :pray: back at the beginning, so I guess you're happy again with this change! :)

pschambacher:
:+1:

pschambacher:
I've got a change of my own on its way regarding the message itself

sandlerr:
Not too keen on this. It will be noisy, and your deploy buddy should be a specific person who is up to speed on what is being deployed, not just anyone hanging out in a flowdock channel

codella:
@sandlerr it is a default message specifically targeting flows, so I assumed we should have the most common message in this scenario.

The workflow you describe most likely requires that the message is not broadcasted on a flow, but is actually sent to the specific person via PM or such.

Moreover, it is mentioning @team - not @all or @anybody - so people that are generally not really part of the team can bail.

Hope this clarified my point of view!

henders:
Merging since the majority have spoken and @codella made good points about opt-out of the @team msgs, plus this msg is just the default and if you are targeting a specific person @sandlerr, then you'd have to modify this msg anyway. Now that we have lots of smaller projects, asking the whole team for approval is probably not noise in the majority of cases.

PR #625, opened 2015-11-03
PAAS-80 Create angular framework for K8s Project tab (roles tab)

sbrnunes:
PR following #622, providing all the existing functionality for managing the Kubernetes Role from the client side, with AngularJS.

/cc @zendesk/samson

### Tasks
 - [x] Implement Kubernetes Roles functionality on the client-side with AngularJS
 - [ ] Implement dashboard tab
 - [ ] Finish styling

### References

### Risks

### Screenshots
screenshot 2015-11-03 17 43 58
screenshot 2015-11-03 17 44 14
screenshot 2015-11-03 17 44 29
screenshot 2015-11-03 17 44 45


msufa:
Rails part looks OK. I'd like someone with JS skills to have another look though.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_role_ctrl.js
```diff
@@ -0,0 +1,40 @@
...
+  $scope.cancel = function() {
+    $state.go('kubernetes.roles');
+  };
+
+  (function loadKubernetesRoleDefaults() {
+    kubernetesService.loadKubernetesRoleDefaults($stateParams.project_id).then(function(data) {
+        $scope.role = kubernetesRoleFactory.build(data);
+      }
+    );
+  })();
```
henders: You're not really protecting any variables by using this closure, just execute directly or wrap in 'init()' method:
```
kubernetesService.loadKubernetesRoleDefaults($stateParams.project_id).then(function(data) {
    $scope.role = kubernetesRoleFactory.build(data);
  });
```
or
```
function init() {
  kubernetesService.loadKubernetesRoleDefaults($stateParams.project_id).then(function(data) {
    $scope.role = kubernetesRoleFactory.build(data);
  });
}
init();
```

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_role_ctrl.js
```diff
@@ -0,0 +1,40 @@
...
+  $scope.submit = function(e) {
+    e.preventDefault(); //prevents default form submission
+
+    kubernetesService.createKubernetesRole($stateParams.project_id, $scope.role).then(
+      function() {
+        // Postpones execution of this instruction until the current digest cycle is finished.
+        // This was needed to keep the flash message on the page, otherwise it would disappear
+        // instantly due to the URL state change that follows.
+        $timeout(function() {
+          notificationService.success('Role ' + $scope.role.name + ' has been create successfully.');
```
henders: 'created'

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_roles_ctrl.js
```diff
@@ -1,5 +1,16 @@
-samson.controller('KubernetesRolesCtrl', function($state, $stateParams) {
+samson.controller('KubernetesRolesCtrl', function($scope, $stateParams, kubernetesService, kubernetesRoleFactory) {
+  $scope.project_id = $stateParams.project_id;
 
-//TODO: Load roles asynchronously
+  $scope.roles = [];
 
+  (function loadKubernetesRoles() {
```
henders: Again, the closure syntax is a bit ugly and unncecessary here,  it does save a line instead of creating an 'init' function and calling it, but the syntax doesn't make it worth it IMO.

henders: A lot of angular tutorials out there seem to just create and call an 'init()' method.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_role_ctrl.js
```diff
@@ -0,0 +1,40 @@
+samson.controller('KubernetesCreateRoleCtrl', function($scope, $state, $stateParams, kubernetesService, kubernetesRoleFactory, notificationService, $timeout) {
+
+  $scope.action = 'Create';
```
henders: I don't see where this is used?

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_tabs_ctrl.js
```diff
@@ -1,7 +1,7 @@
-samson.controller('KubernetesTabsCtrl', function($rootScope, $scope) {
+samson.controller('KubernetesTabsCtrl', function($rootScope, $scope, $stateParams) {
```
henders: Not using stateParams in this controller.

plugins/kubernetes/app/assets/javascripts/kubernetes/kubernetes.js
```diff
@@ -5,31 +5,57 @@ angular.module('samson').config(function($stateProvider, $urlRouterProvider) {
...
+      url: "/:role_id/edit",
+      views: {
+        'content@': {
+          templateUrl: 'kubernetes/kubernetes_edit_role.tmpl.html',
+          controller: 'KubernetesEditRoleCtrl'
+        }
+      }
+    })
+    .state('kubernetes.roles.create', {
+      url: "/new",
```
henders: url should probably be /roles/new

plugins/kubernetes/app/assets/javascripts/kubernetes/services/kubernetes_service.js
```diff
@@ -0,0 +1,86 @@
+samson.service('kubernetesService', function($http, $q) {
+
+  var config = {
+    headers: {
+      'Accept': 'application/json'
+    }
+  };
+
+  this.loadKubernetesRoles = function(project_id) {
```
henders: You probably don't need to use 'Kubernetes' in each of the methods, as it belongs to a 'Kubernetes' service. It'll be obvious when using it from other code what it does.

plugins/kubernetes/app/controllers/kubernetes_roles_controller.rb
```diff
@@ -1,74 +1,48 @@
...
-          render :new, status: 422
-        end
-      end
-
-      format.json do
-        render json: {}, status: @kubernetes_role.persisted? ? 200 : 422
-      end
-    end
-  end
+  before_action :authorize_project_deployer!
```
henders: You should probably wrap the new/create/update methods under the project_admin privs.

plugins/kubernetes/app/controllers/kubernetes_roles_controller.rb
```diff
@@ -1,74 +1,48 @@
...
-  def find_role
-    @kubernetes_role = project.roles.find(params[:id])
+  def new_role_params
+    params.require(:kubernetes_role).permit(:name, :config_file, :service_name, :ram, :cpu, :replicas, :deploy_strategy)
   end
 
-  def new_role_params
-    params.require(:kubernetes_role).permit(:name, :config_file, :service_name, :ram, :cpu, :replicas)
+  def update_role_params
+    params.require(:kubernetes_role).permit(:id, :name, :config_file, :service_name, :ram, :cpu, :replicas, :deploy_strategy)
```
henders: No point in having the :id param since you do a find beforehand anyway. And if you remove the :id param, then you might as well use the above 'new_role_params' and rename it to just 'role_params'

plugins/kubernetes/app/views/kubernetes_roles/_roles_list.html.erb
```diff
@@ -0,0 +1,41 @@
...
+        <th>RAM</th>
+        <th>CPU</th>
+        <th>Replicas</th>
+        <th>Actions</th>
+      </tr>
+      </thead>
+
+      <tbody>
+      <tr ng-show="!roles.length" ng-hide="roles.length">
+        <td>No project roles have been created yet.</td>
```
henders: Does this not look weird to have this text squashed into the first column?
If it does look weird, maybe extract this outside the table and only show the table if there are roles.

plugins/kubernetes/test/angular/kubernetes_spec.js
```diff
@@ -0,0 +1,149 @@
...
+    it('State views should reflect current state', function() {
+      expect($state.current.views['content@']['templateUrl']).toEqual('kubernetes/kubernetes_release_groups.tmpl.html');
+      expect($state.current.views['content@']['controller']).toEqual('KubernetesReleaseGroupsCtrl');
+    });
+
+    it('State params should reflect current state', function() {
+      expect($stateParams.project_id).toEqual('some_project');
+    });
+  });
+});
```
henders: W00t, tests! :heart:

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_role_ctrl.js
sbrnunes: Okidokie.
The reason why I went with this approach was was just to execute this function right away. I wanted to define the function, as I don't like to see service calls placed directly in the body of the controller, but I didn't want to write another line just to invoke the function. :)

I'm going to change it, anyway. There's no good argument in favour of doing it. I just found it interesting. :-P

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_role_ctrl.js
sbrnunes: Damn, was an experiment, but ended up passing the label for the button through the rendering of the partial.

This code should have been removed. Thanks.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_tabs_ctrl.js
sbrnunes: Thanks.

plugins/kubernetes/app/assets/javascripts/kubernetes/kubernetes.js
sbrnunes: This url is relative to the parent state, which is 'kubernetes.roles'. So, when you navigate to this state, the url, in the browser, will actually be '/roles/new'.

plugins/kubernetes/app/assets/javascripts/kubernetes/services/kubernetes_service.js
sbrnunes: Fair enough. :)

plugins/kubernetes/app/controllers/kubernetes_roles_controller.rb
sbrnunes: Good point. Hadn't thought about that.

plugins/kubernetes/app/controllers/kubernetes_roles_controller.rb
sbrnunes: Good point. Thanks.

plugins/kubernetes/app/views/kubernetes_roles/_roles_list.html.erb
sbrnunes: I forgot to add the colspan.
I just finished some tweaks around this logic, anyway. In summary:
- while we're waiting for the call to come back, we're showing a loading spinner in place of the table
- after the call comes back, if there's no results, we display the table with the headers and a single row with the message above (going to add the colspan)
- if there's any items, we display them

sbrnunes:
All suggested fixed have been added.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_role_ctrl.js
```diff
@@ -0,0 +1,40 @@
...
+    e.preventDefault(); //prevents default form submission
+
+    kubernetesService.createRole($stateParams.project_id, $scope.role).then(
+      function() {
+        // Postpones execution of this instruction until the current digest cycle is finished.
+        // This was needed to keep the flash message on the page, otherwise it would disappear
+        // instantly due to the URL state change that follows.
+        $timeout(function() {
+          notificationService.success('Role ' + $scope.role.name + ' has been created successfully.');
+        });
```
henders: This looks odd. Can you trigger the state change first and then trigger the flash msg?

plugins/kubernetes/app/assets/javascripts/kubernetes/kubernetes.js
henders: Ah, cool.

henders:
:+1:  after last question answered! :)

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_role_ctrl.js
sbrnunes: I can. Regardless of the order of this instructions, the timeout function will wait until the end of the digest cycle, so, It'll work either way.

PR #626, opened 2015-11-04
Call WatchStream#finish outside the #each loop

msufa:
As per https://github.com/abonas/kubeclient/issues/139, closing the watcher from inside the #each loop is not safe and most probably leads to an exception when accessing the underlying socket.

/cc @jonmoter @sbrnunes @henders 

### References
 - Jira link:

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/app/models/watchers/base_watcher.rb
+++ b/plugins/kubernetes/app/models/watchers/base_watcher.rb
@@ -8,19 +8,20 @@ module Watchers
     def initialize(watch_stream, log: true)
       @watch_stream, @log, = watch_stream, log
       @start_time = Time.now
+      @enabled = false
     end
 
     def start_watching(&block)
+      @enabled = true
       @watch_stream.each do |notice|
         handle_notice(notice, &block)
+        break if stop_watching?
       end
+      cleanup
     end
 
     def stop_watching
-      if @watch_stream
-        @watch_stream.finish
-        @watch_stream = nil
-      end
+      @enabled = false
     end
 
     def time_watching
@@ -45,6 +46,17 @@ module Watchers
 
     def log(msg, extra_info = {})
       Kubernetes::Util.log(msg, extra_info) if @log
+    end
+
+    def stop_watching?
+      not @enabled
+    end
+
+    def cleanup
+      if @watch_stream
+        @watch_stream.finish
+        @watch_stream = nil
+      end
     end
   end
 end
--- a/plugins/kubernetes/app/models/watchers/base_watcher.rb
+++ b/plugins/kubernetes/app/models/watchers/base_watcher.rb
@@ -12,6 +12,7 @@ module Watchers
     end
 
     def start_watching(&block)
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/app/models/watchers/base_watcher.rb
```diff
@@ -8,19 +8,20 @@ class BaseWatcher
...
+      cleanup
     end
 
     def stop_watching
-      if @watch_stream
-        @watch_stream.finish
-        @watch_stream = nil
-      end
+      @enabled = false
     end
```
jonmoter: There's one scenario where this doesn't work.  If a separate thread calls `#stop_watching` on this object, the watcher will continue waiting for new data to come down the stream.  Once it gets another notice, it will break from the `#each` loop.  But if it never gets another notice (let's say it was watching a single pod that got deleted), then it will get stuck in that wait loop forever.

I'm not sure the cleanest way to fix this.  We might need to check whether it's the same thread that is calling `#stop_watching`, or a different thread, and handle it in different ways.

simon3z: @jonmoter in that case from the other thread you should call finish to force the connection to drop (IIRC you can safely call finish multiple times if you protect from races)

plugins/kubernetes/app/models/watchers/base_watcher.rb
```diff
@@ -8,18 +8,24 @@ class BaseWatcher
...
 
     def stop_watching
-      if @watch_stream
-        @watch_stream.finish
-        @watch_stream = nil
+      if @enabled
+        @enabled = false
+        # close the socket and stop receiving updates if called from a different thread than the original
+        cleanup if Thread.current.object_id != @original_thread_id
```
sbrnunes: I'm still puzzled about how can this be happening, I mean, reaching this point from a different Thread. If all this objects (ReleaseGroup, Release, DeployService and PodWatcher) get created within the scope of a request, and then a new Thread is launched from this request to watch for the state of this objects, with no state being shared across different Threads (apparently), how can we end up with this scenario? 

I'm just concerned there's something else we're missing in here that could lead us to some weird and inconsistent state. :(

sbrnunes: @msufa just explained me the actual goal behind this fix, to allow us to stop watching the stream from a different Thread, if we find out that this Thread is stuck or something like that.

I misunderstood the whole point of this PR. I thought it was only to fix the crash we had.

Sorry @msufa. :-D

msufa: In the current state of the plugin there are no other threads involved. However, if we were to have a supervisor thread, or any other thread trying to stop the watcher when the watcher was stuck waiting for events, it would have no effect. The 2nd commit is addressing this (currently hypothetical) scenario.

jonmoter:
Thanks for handling that other scenario.

:+1:

PR #627, opened 2015-11-04
Stop previous deploy before starting a new one

msufa:
We need to scale down all running replication controllers that match, then delete them. Only after the cleanup is done we proceed with the new deploy.

Lessons learned:
 - Deleting a replication controller does not automatically stop the pods it controls
 - You always need to specify a namespace when deleting a resource

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/app/models/kuber_deploy_service.rb
+++ b/plugins/kubernetes/app/models/kuber_deploy_service.rb
@@ -3,7 +3,7 @@ require 'kubeclient'
 class KuberDeployService
   attr_reader :kuber_release
 
-  delegate :build, :client, to: :kuber_release
+  delegate :release_group, :client, to: :kuber_release
 
   def initialize(kuber_release)
     @kuber_release = kuber_release
@@ -13,6 +13,7 @@ class KuberDeployService
     # TODO: handling different deploy strategies (rolling updates, etc.)
     log 'starting deploy'
 
+    delete_replication_controllers!
     create_services!
     create_replication_controllers!
 
@@ -42,6 +43,20 @@ class KuberDeployService
       else
         log 'creating Service', role: role.name, service_name: service.name
         client.create_service(Kubeclient::Service.new(release_doc.service_hash))
+      end
+    end
+  end
+
+  def delete_replication_controllers!
+    kuber_release.release_docs.each do |release_doc|
+      log 'deleting existing ReplicationControllers', role: release_doc.kubernetes_role.name
+      labels = {project: release_group.build.project_name, role: release_doc.kubernetes_role.label_name}
+      client.get_replication_controllers(
+          label_selector: labels.to_kuber_selector,
+          namespace: kuber_release.deploy_group.cluster_deploy_group.namespace).each do |rc|
+        rc.spec.replicas = 0
+        client.update_replication_controller rc
+        client.delete_replication_controller rc.metadata['name'], rc.metadata['namespace']
       end
     end
   end


plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -46,6 +47,20 @@ def create_services!
     end
   end
 
+  def delete_replication_controllers!
+    kuber_release.release_docs.each do |release_doc|
+      log 'deleting existing ReplicationControllers', role: release_doc.kubernetes_role.name
+      labels = {project: release_group.build.project_name, role: release_doc.kubernetes_role.label_name}
```
jonmoter: Style nitpick: for single-line hashes, leave a space around the curly braces.

e.g. `{ foo: bar }`, not `{foo: bar}`

msufa: I read somewhere (can't find the reference now, it might have been the Pickaxe book) that it's good practise to use spaces for block one-liners, but not for hashes, in order to make the two visually distinct. It seemed reasonable, but coherence trumps convention :)

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -46,6 +47,20 @@ def create_services!
...
+      client.get_replication_controllers(
+          label_selector: labels.to_kuber_selector,
+          namespace: kuber_release.deploy_group.cluster_deploy_group.namespace).each do |rc|
```
jonmoter: you could just say `namespace: kuber_release.namespace`

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -46,6 +47,20 @@ def create_services!
...
+        rc.spec.replicas = 0
+        client.update_replication_controller rc
+        client.delete_replication_controller rc.metadata['name'], rc.metadata['namespace']
+      end
+    end
+  end
```
jonmoter: This method would only be used if a role had a `kill_and_restart` strategy, right?  I assume it wouldn't be used if we're doing a rolling update.  (Which we haven't implemented yet.)

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -46,6 +47,20 @@ def create_services!
...
+  def delete_replication_controllers!
+    kuber_release.release_docs.each do |release_doc|
+      log 'deleting existing ReplicationControllers', role: release_doc.kubernetes_role.name
+      labels = {project: release_group.build.project_name, role: release_doc.kubernetes_role.label_name}
+      client.get_replication_controllers(
+          label_selector: labels.to_kuber_selector,
+          namespace: kuber_release.deploy_group.cluster_deploy_group.namespace).each do |rc|
+        rc.spec.replicas = 0
+        client.update_replication_controller rc
+        client.delete_replication_controller rc.metadata['name'], rc.metadata['namespace']
```
jonmoter: If you delete the RC right after updating the number of replicas to zero, does that give Kubernetes enough time to process and kill the pods?  Or do we need to wait to confirm that happens?

plugins/kubernetes/app/models/kuber_deploy_service.rb
msufa: Yup, exactly. I didn't want to focus much on other strategies, the main goal is to not have 50 pods trying to run in my local cluster after I do some tests :)

plugins/kubernetes/app/models/kuber_deploy_service.rb
msufa: Cool!

plugins/kubernetes/app/models/kuber_deploy_service.rb
msufa: I've double checked and we actually need to verify that the number of replicas is 0 before we delete the RC. I'll fix this when we have proper event monitoring in place. BTW, K8S 1.1 should contain graceful pod termination - maybe that will handle the issues we're facing here.

msufa:
Closing this PR. We decided to go a slightly different route.

PR #628, opened 2015-11-05
Add ldap bind_dn

Jeoffreybauvin:
Hi,

I need to set "bind_dn" for LDAP authentication. 

https://github.com/intridea/omniauth-ldap

":bind_dn and :password is the default credentials to perform user lookup."


henders:
:+1:

PR #629, opened 2015-11-05
Added support for the releases list page in AngularJS.

sbrnunes:
PR following #625, supporting the list of Kubernetes releases in the client side in AngularJS.

/cc @zendesk/samson

Tasks
 - [x] Implement Kubernetes Roles functionality on the client-side with AngularJS
 - [ ] Implement Kubernetes Releases functionality on the client-side with AngularJS
 - [ ] Implement dashboard tab
 - [ ] Finish styling

### References

### Risks

screenshot 2015-11-05 17 00 15

screenshot 2015-11-05 17 01 03

### DIFF
--- a/plugins/kubernetes/app/assets/javascripts/kubernetes/factories/build_factory.js
+++ b/plugins/kubernetes/app/assets/javascripts/kubernetes/factories/build_factory.js
@@ -1,0 +1,16 @@
+samson.factory('buildFactory', function() {
+
+  function Build(id, label) {
+    this.id = id;
+    this.label = label;
+  }
+
+  Build.build = function(data) {
+    return new Build(
+      data.id,
+      data.label
+    );
+  };
+
+  return Build;
+});
--- a/plugins/kubernetes/app/controllers/kubernetes_release_groups_controller.rb
+++ b/plugins/kubernetes/app/controllers/kubernetes_release_groups_controller.rb
@@ -1,9 +1,31 @@
 class KubernetesReleaseGroupsController < ApplicationController
+  include ProjectLevelAuthorization
   helper ProjectsHelper
 
-  before_action :project
-  before_action :authorize_deployer!
+  before_action :authorize_project_deployer!
   before_action :load_environments, only: [:new, :create]
+
+  def index
+    render json: current_project.kubernetes_release_groups.order('id desc'), root: false
+  end
+
+  def show
+    @release_group = Kubernetes::ReleaseGroup.find(params[:id])
+
+    respond_to do |format|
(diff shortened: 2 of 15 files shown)


plugins/kubernetes/app/assets/javascripts/kubernetes/factories/build_factory.js
```diff
@@ -0,0 +1,16 @@
+samson.factory('buildFactory', function() {
```
henders: This seems like a very Java-like pattern. I don't think Angularjs factories are meant to be used like Java factories.

plugins/kubernetes/app/controllers/kubernetes_release_groups_controller.rb
```diff
@@ -27,30 +49,29 @@ def create
...
+  #   if release_group.save
+  #     release_group.releases.each do |release|
+  #       KuberDeployService.new(release).deploy!
+  #     end
+  #
+  #     render :created, json: release_group, root: false
+  #   else
+  #     render :bad_request, json: {errors: release_group.errors.full_messages}
+  #   end
+  # end
```
henders: Might as well delete the code if it's not used.

plugins/kubernetes/app/assets/javascripts/kubernetes/factories/build_factory.js
henders: Alright, if that's what the cool kids are doing these days, then fine!  :cry:

henders:
:+1: after deleting commented out code + merge conflicts + :green_apple:  :)

plugins/kubernetes/app/assets/javascripts/kubernetes/factories/build_factory.js
sbrnunes: I know it seems quite a java-like pattern, but people usually say that a lot about AngularJS in general. :)

I've researched quite a bit around how to build proper model objects in Javascript with AngularJS, and all the references I saw mentioned using a Factory to build the object. That's why I went that way.

The advantages I see of using this approach instead of pure JSON objects is that you can leverage the model to support read-only objects (Object.freeze()) and rich client-side validations. I'm still not doing it, though, but I'll probably start doing it soon.

Just a few references:
https://medium.com/opinionated-angularjs/angular-model-objects-with-javascript-classes-2e6a067c73bc
http://www.webdeveasy.com/angularjs-data-model/
https://www.mutuallyhuman.com/blog/2014/05/08/angularjs-services-and-factories-done-right/

PR #630, opened 2015-11-07
use a secure protocol to fetch projects

grosser:
@jonmoter 

can you add the PR url you are waiting on to get merged ?

### Risks
 - None


jonmoter:
Actually, I think the upstream version of the gem works now.  I'll test it out, and if so, put the dependency back in the plugin gemspec.

jonmoter:
@grosser, I opened PR #631, which should make this PR unnecessary.

PR #631, opened 2015-11-09
use upstream version of kubeclient, instead of zendesk custom fork

jonmoter:
Since the [kubeclient](https://github.com/abonas/kubeclient) gem has merged the PRs we submitted, we don't need to use our fork.

We can close PR #630 in favor of this change.

/cc @grosser, @henders, @msufa, @sbrnunes 

### Risks
 - Minimal, changes only impact Kubernetes plugin


sandlerr:
:+1:

PR #633, opened 2015-11-09
do not migrate my MYSQL_URL

grosser:
... turns out MYSQL_URL this is set to another database and then migrating here messes that up ... this should be safer ...

@jonmoter 

### Risks
 - None


jonmoter:
:+1:

PR #635, opened 2015-11-09
deploy all stages to a single group

grosser:
@zendesk/runway 

screen shot 2015-11-09 at 11 28 09 am
screen shot 2015-11-09 at 2 55 51 pm
screen shot 2015-11-09 at 2 56 05 pm

### Links
 - JIRA: https://zendesk.atlassian.net/browse/SAMSON-198

### Risks
 - Medium: people creating a bunch of stages by accident
 - Medium: incorrectly named stages get deployed

### DIFF
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -1,7 +1,7 @@
 class Admin::DeployGroupsController < ApplicationController
   before_action :authorize_admin!
-  before_action :authorize_super_admin!, only: [ :create, :new, :update, :destroy ]
-  before_action :deploy_group, only: [:show, :edit, :update, :destroy]
+  before_action :authorize_super_admin!, only: [ :create, :new, :update, :destroy, :deploy_all, :deploy_all_now ]
+  before_action :deploy_group, only: [:show, :edit, :update, :destroy, :deploy_all, :deploy_all_now]
 
   def index
     @deploy_groups = DeployGroup.all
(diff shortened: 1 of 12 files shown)


grosser:
@jonmoter

grosser:
@jonmoter

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -44,8 +44,52 @@ def destroy
...
+  def matching_stages(environment)
+    Project.all.map do |project|
+      project.stages.detect do |stage|
+        matches_group?(stage) && # direct match
+          stage.last_successful_deploy # was deployed
+      end || project.stages.detect do |stage|
+        stage.command.include?("$DEPLOY_GROUPS") && # is dynamic
+          stage.deploy_groups.where(environment: environment).exists? && # is made to go to this environment
+          stage.last_successful_deploy # was deployed
+      end
```
jonmoter: Any reason you didn't put this logic into the `Stage` model?

grosser: was thinking about it ... only used once so far ... so wanted to keep the craziness locale

jonmoter:
The code looks good, at least for a first pass.

If this proves successful, we may want to have some way to save a custom deploy list, track when they happen, and have a global buddy deploy for the entire thing.

Lemme just run this by Compliance before we merge this, to make sure that doesn't raise any concerns for them.

grosser:
updated to 
 - insist on $DEPLOY_grOUPS
 - do not fail if our specific stage already exists but did not get deployed
 - pick the latest deploy across the target environment

jonmoter:
For the record, I ran this feature by Lisa and Ruby, and they gave me the thumbs up.

jonmoter:
Let's give it a try and see how it looks in practice.

:+1:

PR #636, opened 2015-11-10
add rbtrace for debugging

sandlerr:
@zendesk/samson sometimes samson gets into weird states but we can't inspect things like the JobExecution registry. This is easier than building a debug UI and should be usable even if samson is choking and won't serve http.


jonmoter:
This is not encouraging that the Travis builds are dying when doing a `bundle install`

sandlerr:
that happened locally for me as well, it's a bundler bug that can be worked around by disabling parallel installation with `-j1`. I'll look into fixing it if you're ok with this in principle since it is a bit invasive to go around attaching debuggers to live apps

config/initializers/job_execution.rb
```diff
@@ -1,3 +1,5 @@
+require 'rbtrace'
```
jonmoter: Let's wrap this in an environment variable check, so people running Samson can easily enable/disable this functionality without needing to modify the code.

jonmoter:
I don't have any experience using [rbtrace](https://github.com/tmm1/rbtrace), so I don't know if there are any dangers to it.  @grosser or @steved, could I get a second opinion on this?

In principle, people attach debuggers & run strace on misbehaving processes all the time. If it's a choice between debugging and just killing it, might as well debug. So I don't object to the idea. I just don't know if including these gems could possibly cause more problems than it solves.

sandlerr:
Sure. As long as it doesn't do anything until signalled from a terminal by someone doing `bundle exec rbtrace` it should not cause trouble.

grosser:
:+1: if it works ... try it on production first ... tried a few of these and did not have any luck :/

sandlerr:
I have run into problems using `gdb` in production, not sure if our sudo permissions are too constrained for meaningful debugging. strace is useable.

Definitely agree that this should be behind a guard for configuration or an envar.

PR #637, opened 2015-11-10
Remove Kubernetes plugin release_group table

henders:
Remove Release_group table and move Kubernetes_Release table to top level

This broke the watch functionality, but that had to be changed anyway. 

screen shot 2015-11-10 at 1 39 17 p m
screen shot 2015-11-10 at 1 39 27 p m
screen shot 2015-11-10 at 1 39 44 p m
screen shot 2015-11-10 at 1 39 54 p m

/cc @msufa  @sbrnunes @jonmoter 

### References
 - Jira link: 

### Risks
 - Kubernetes plugin breakage.


plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -81,14 +83,25 @@ def stop_watching
...
       @docs_by_role ||= release_docs.each_with_object({}) do |rel_doc, hash|
         hash[rel_doc.kubernetes_role.label_name] = rel_doc
       end
     end
+
+    def deploy_group_ids
+      release_docs.map(&:deploy_group_id)
+    end
+
+    def deploy_group_ids=(id_list)
```
msufa: Shouldn't there be an assignment here?

msufa:
I'd love to see some screenshots for the new views.

plugins/kubernetes/app/models/kubernetes/release.rb
henders: Nope, we aren't using it anymore. The controller was using this mechanism to create releases under the release_group.

I couldn't figure out how to get rid of this function as the permit params on the controller was not liking me not using the passed in params.

henders:
@msufa 
There aren't any new views in this, I just moved the views around. They are 'mostly' showing the same stuff as before. The post-release page is just a bit different but we're going to recreate that page anyway.

henders:
But there you go anyway @msufa  :smile_cat:

sbrnunes:
From my side, :+1:

msufa:
Thanks for indulging me with the beautiful screenshots @henders You've earned your :+1:

PR #638, opened 2015-11-10
PAAS-81 Ability to auto import Kubernetes Roles from the git sha of a given build

sbrnunes:
Adding the ability to auto import Kubernetes Roles from the git sha of a given build.

/cc @zendesk/samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-81

### Risks
 - None (additional functionality)

screenshot 2015-11-20 01 05 35

screenshot 2015-11-20 01 06 29

screenshot 2015-11-20 01 06 53


app/models/build.rb
```diff
@@ -69,6 +69,31 @@ def file_from_repo(path, ttl: 1.hour)
...
 
+  def import_kubernetes_roles
+    # TODO : Build -> KubernetesRoles association
+    project.roles.destroy_all
+
+    config_files = project.repository.directory_contents(git_sha, path: 'kubernetes')
+
+    config_files.map do |file|
+      file_contents = project.repository.file_contents(git_sha, file)
+      config_file = Kubernetes::RoleConfigFile.new(file_contents, file)
```
msufa: I think it would be safer to only search for yaml files. I wouldn't be surprised if there was a readme file or some garbage in that directory.

For extra credit we could think of some validation on the yaml files (unfortunately currently it's only supported in kubectl)

msufa:
I would love to see some tests for the new functionality :)

app/models/build.rb
sbrnunes: You made a good point regarding additional files in that directory. I can filter the output to get only the .yaml or .json files. Thanks for the heads up.

Regarding the validation, I was thinking about that, but I'm still not sure about where and how to do it, i.e., if we should validate the files contents or validate the required data after parsing the file. Or... validate the files when creating the build.

We can address this is multiple ways, I'm just not sure about how to do it, yet.

sbrnunes:
I know. I know. But I think that writing tests for things that we are still just experimenting it's overkill. In this code, I'm still using the Kubernetes::Role ActiveRecord class and persisting to the database, but this will probably go away, as we discussed yesterday. 

But I guess I could write tests for the file parsing part of this changes. That shouldn't change.

Another think I just realised, I placed the import logic in the wrong class (Build instead of Kubernetes::BuildDecorator). I have to move that code.

sbrnunes:
@msufa, added the ability to filter the files by allowing to pass a block into the method. Also, added tests to this core functionality (git hub integration and parsing of the configuration files).

@msufa , @henders : this PR is now ready to be reviewed.

app/models/git_repository.rb
```diff
@@ -114,6 +114,27 @@ def file_changed?(sha1, sha2, file)
...
+    output
+  end
+
+  def file_contents(git_ref, file_path)
+    #first command disables less as default pager
+    cmds = ["git config core.pager ''", "git show -q #{git_ref}:#{file_path}"]
+    valid, output = run_commands(cmds) { |line| line }
+    Rails.logger.error("Could not retrieve file contents for '#{file_path}'. Git Ref '#{git_ref}'") unless valid
+    output.join
+  end
```
jonmoter: FYI, `Build#file_from_repo` does the same thing. Let's not have two methods that do the same thing.

app/models/git_repository.rb
```diff
@@ -124,10 +145,28 @@ def locally_cached?
...
   end
+
+  # Runs the provided list of commands.
+  # This method DOES NOT filter the output, neither does any sort of sorting.
+  def run_commands(*commands, pwd: repo_cache_dir)
+    tmp_executor = TerminalExecutor.new(StringIO.new)
+    success = tmp_executor.execute!("cd #{pwd}", commands)
+    result = tmp_executor.output.string.lines.map { |line| yield line if block_given? }
+    [success, result]
+  end
```
jonmoter: Could you rewrite `run_single_command` to use this method, to DRY it up?

plugins/kubernetes/app/decorators/build_decorator.rb
```diff
@@ -5,4 +5,31 @@
   def project_name
     @project_name ||= project.name.parameterize('-')
   end
+
+  def import_kubernetes_roles
+    # TODO : Build -> KubernetesRoles association
+    project.roles.destroy_all
```
jonmoter: This is dangerous. `Kubernetes::ReleaseDoc` belongs_to role.  If you destroy all the roles, either the ReleaseDocs will get deleted, orphaned, or with `kubernetes_role_ids` columns pointing to a null row, or the wrong row.

Maybe some logic that checks if the project has any Kubernetes releases, and fails if it does?

plugins/kubernetes/app/decorators/build_decorator.rb
```diff
@@ -5,4 +5,31 @@
...
+
+    config_files = project.repository.directory_contents(git_sha, path: 'kubernetes', ).select { |file|
+      file.ends_with?('.yml', '.yaml', '.json')
+    }
+
+    config_files.map do |file|
+      file_contents = project.repository.file_contents(git_sha, file)
+      config_file = Kubernetes::RoleConfigFile.new(file_contents, file)
+
+      # TODO : Build -> KubernetesRoles association
```
jonmoter: Why do you need an association between the two?  They're connected via Build -> Release -> ReleaseDoc -> Role.

plugins/kubernetes/app/models/kubernetes/role_config_file.rb
```diff
@@ -0,0 +1,173 @@
...
+  # This file represents a Kubernetes configuration file for a specific project role.
+  # A single configuration file can have both a Replication Controller and a Service, as two separate documents.
+  #
+  # For the schema definition of a Replication Controller see:
+  # https://cloud.google.com/container-engine/docs/replicationcontrollers/operations
+  #
+  # For the schema definition of a Service, see:
+  # https://cloud.google.com/container-engine/docs/services/operations
+  #
+  class RoleConfigFile
```
jonmoter: Hmm, this class seems to duplicate a chunk of the logic in ReleaseDoc.  Are you thinking of refactoring to combine them, or keep them separate?

app/models/git_repository.rb
sbrnunes: Oh, so dumb of me, I could have called this method and then run the uniq.sort on the results.
Thanks for the heads up.

plugins/kubernetes/app/decorators/build_decorator.rb
sbrnunes: Sorry @jonmoter, you were caught by surprise here. We were having some thoughts around how to refresh the Kubernetes roles. I'll write you an email explaining some new ideas we're experimenting with. I'll C.c. the guys as well and we can discuss it there.

plugins/kubernetes/app/decorators/build_decorator.rb
sbrnunes: See my comment above.

plugins/kubernetes/app/models/kubernetes/role_config_file.rb
sbrnunes: I know. We'll probably need to review ReleaseDoc when we get there (i.e., when we start dealing with the actual release logic). For now, the main goal for this class is just to represent a specific Role configuration file. At the moment, it only has logic to parse and retrieve data, but, in the mid-term, It could also contain some validation logic as well, but we haven't thought about that yet. Where and how to validate the yml files.

We could then leverage this class in the ReleaseDoc, when creating a Release. The logic to parse the file is already encapsulated in here, so, we just need to refactor the logic that deals with overriding the template with the specifics of the release.

app/models/git_repository.rb
sbrnunes: Yeah... I was wondering about which way to go, now that we're listing each Role config file and fetching it's contents. I'm conscious there's not going to be a lot of Roles, but I was discussing this with @henders as well, and we were wondering if it's not actually faster to use the TerminalExecutor. Fetching the contents of each file will lead to a separate API call for each file. Using the TerminalExecutor, we already have all the files in the local cached repo, so, it should be faster than using the API.

I should also mention that, due to some new ideas we've been thinking about, which I'll explain by email, there's a good chance the User will be triggering this logic when creating a new Release, or creating a new Role. Either way,  in terms of client-side user experience, we'd like to make this calls as soon as possible.

jonmoter: One thing you're leaving out in this workflow is a `git fetch`.  There's no guarantee that the local cache of the repo is going to have the reference that's specified. So before calling `#file_contents`, you should call `#update!`.

So let's say you're fetching the values of 3 different files.  Using the logic in `Build#file_from_repo` would make 3 calls to GitHub, and cache the results in memcache for 1 hour.  If you use this logic, you'd do a single `git fetch` to bring the local repo up to date, and then run 3 local git commands to get the file contents.

I don't know off-hand which would be faster.  My guess is that `git fetch` can take some time, even if the repo is up to date.  But we could do some testing.

sbrnunes: Hummm, that's a good point. I missed that workflow in the previous method as well: `directory_contents`. 
As soon as I fix that workflow, I can run some benchmarks. I actually started by using the Github API, through the `Build#file_from_repo` method you mentioned. But later realised that the local cached repo should already be there, and ended up switching over to this implementation. :)

Thanks for the heads up. I'll let you know if there's any big differences.

jonmoter: Technically, you could have code that looks like this:

```ruby
repo = GitRepository.new(repository_url: 'github.com/foo/bar', repository_dir: '/tmp/foo')
repo.file_contents('master', 'README.md')
```

In that scenario, `setup!` wasn't called, so the local cache doesn't even exist. I'm guessing it would fail somewhere in `run_commands` with a cryptic error message.

Maybe add a check at the beginning of the method like `fail "Not setup" unless locally_cached?`

sbrnunes: Ok. Will do. That's another good point. :)

sbrnunes:
After applying Jon's suggestions, I run a few simple benchmarks to compare the Github API vs running the Git commands against the PTY pseudo terminal, and realised that the API is significantly faster (and without even using the cache). So, ended up switching over to use the API.

Also, made the API responsible for importing the roles from the Github repo read-only for now, until we decide which way to go.

jonmoter:
FYI, if the GitHub API is significantly faster, you might as well use that to get the contents of a directory too.  The output looks like this:

```ruby
[12] pry(main)> GITHUB.contents('zendesk/truth_service', path: 'spec', ref: 'master').first
{
            :name => "spec_helper.rb",
            :path => "spec/spec_helper.rb",
             :sha => "d41d51838b1820c46178e16f2cd8c08a259015ed",
            :size => 2150,
             :url => "https://api.github.com/repos/zendesk/truth_service/contents/spec/spec_helper.rb?ref=master",
        :html_url => "https://github.com/zendesk/truth_service/blob/master/spec/spec_helper.rb",
         :git_url => "https://api.github.com/repos/zendesk/truth_service/git/blobs/d41d51838b1820c46178e16f2cd8c08a259015ed",
    :download_url => "https://raw.githubusercontent.com/zendesk/truth_service/master/spec/spec_helper.rb?token=ABAe-iF579IZQKtQbnn3B811hzW3bVGHks5WRNB-wA%3D%3D",
            :type => "file",
          :_links => {
        :self => "https://api.github.com/repos/zendesk/truth_service/contents/spec/spec_helper.rb?ref=master",
         :git => "https://api.github.com/repos/zendesk/truth_service/git/blobs/d41d51838b1820c46178e16f2cd8c08a259015ed",
        :html => "https://github.com/zendesk/truth_service/blob/master/spec/spec_helper.rb"
    }
}
```

So you could make a variation of `Build#file_from_repo` that returned a list of filenames in a specified directory.  That way, we'd use the same mechanism for both operations.

sbrnunes:
I was forgetting the Contents API could list the files on a directory as well. 
I'm going to change that, then. Thanks.

app/assets/javascripts/directives/git_references_typeahead_directive.js
```diff
@@ -0,0 +1,118 @@
+samson.directive('gitReferencesTypeahead', function() {
```
sbrnunes: @henders, had a few problems with using the existing Git references typeahead with AngularJS. AngularJS directives apparently run after the jQuery document is loaded, meaning, when the jQuery bindings run, the AngularJS templates have not been loaded yet and the target elements are still not in the DOM. For this reason, the typeahead was not being initialized.

Ended up creating a generic component (directive) that searches for the Git reference and delegates the action to the parent Controller (i.e., the parent Controller defines what to do when the search button is clicked).

sbrnunes:
@jonmoter, @henders, @msufa : updated this PR with the suggestions from previous comments and from the decisions we took recently. Looks big but half of it it's tests. :)

Cheers.

plugins/flowdock/test/angular/buddy_request_box_spec.js
```diff
@@ -126,7 +126,7 @@ describe('Buddy requests notifications', function () {
         };
         expect(flowdock.buddyRequest).toHaveBeenCalledWith('Some message');
         buddyNotificationsCtrl.notifyFlowDock();
-      }, 100);
+      }, 500); //if tests fail because of this, increase the delay
```
henders: Nice, glad to see this being less of a random fail.

plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
```diff
@@ -1,37 +1,115 @@
...
+    animation-timing-function: linear;
+}
+
+.spinning.ng-animate {
+    -webkit-animation: none;
+    -moz-animation: none;
+    -ms-animation: none;
+    animation: none;
+}
+
```
henders: May want to alter rubymine's tab setting for CSS files

plugins/kubernetes/app/controllers/kubernetes_roles_controller.rb
```diff
@@ -21,19 +21,12 @@ def update
...
-
-  def create
-    role = current_project.roles.build(role_params)
-    role.save
-
-    if role.persisted?
-      render status: :created, json: role
+  def refresh
+    roles = current_project.refresh_kubernetes_roles!(refresh_params)
+    if roles.try(:any?)
```
henders: Seems a bit odd to enumerate the whole list here to check if it's empty? Why not just ```roles.empty?```

plugins/kubernetes/app/decorators/project_decorator.rb
```diff
@@ -1,4 +1,64 @@
...
+  rescue Octokit::NotFound
+    nil
+  end
+
+  # Imports the new kubernetes roles. This operation is atomic: if one role fails to be imported, none
+  # of them will be persisted and the soft deletion will be rollbacked.
+  def refresh_kubernetes_roles!(git_ref)
+    config_files = directory_contents_from_repo('kubernetes', git_ref)
+
+    if config_files.try(:any?)
```
henders: Again, just use ```empty?```, quicker as it's just a size check, any? will allow nils to exist possibly causing errors further down.

henders:
Epic amount of tests! :heart: 

:+1: , great starting point to iterate on!

plugins/kubernetes/app/controllers/kubernetes_roles_controller.rb
sbrnunes: Oh, damn, I hadn't noticed that 'any?' would enumerate the whole list. I thought it was the opposite from 'empty?'. Good catch. I'll fix that.

plugins/kubernetes/app/decorators/project_decorator.rb
sbrnunes: Same as above. I'll fix that. Good catch.

PR #639, opened 2015-11-16
Shender/kubernetes deploy watcher

henders:
Create a watcher actor that subscribes to events from K8s.
At moment it just fakes some msgs from K8s. But will hook up to stuff from @msufa who's code with publish the events.

hjl9zspapc

/cc @sbrnunes @msufa @jonmoter 

### References
 - Jira link: 

### Risks
 - None


plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -0,0 +1,83 @@
+require 'celluloid'
+require 'celluloid/autostart'
```
msufa: celluloid/autostart already includes celluloid. I would however go for celluloid/current to force ourselves to use the new APIs. Celluloid is on its way to version 1.0 and there are some changes coming, which are better addressed now, see https://github.com/celluloid/celluloid/wiki/DEPRECATION-WARNING

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -0,0 +1,83 @@
...
+  class DeployWatcher
+    include Celluloid
+    include Celluloid::Notifications
+
+    finalizer :on_termination
+
+    def initialize(release)
+      @release = release
+      @current_rcs = {}
+      Rails.logger.info "Start watching K8s deploy: #{@release}"
```
msufa: On my side of things I include `Celluloid::Internals::Logger` and use that one (see https://github.com/celluloid/celluloid/wiki/Logging). It would be good to agree how and which loggers do we use. If we decide to go with Rails logger, we would benefit from plugging it into Celluloid to have traces of actors crashing. Currently I'm creating a new log file when the plugin starts: `Celluloid.logger = ::Logger.new(File.join(Rails.root, 'log', 'celluloid.log'))`

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    def on_termination
+      send_event(msg: 'Finished Watching Deploy!')
+    end
+
+    def update_replica_count(release_doc, pod)
+      rc = rc_pods(release_doc.replication_controller_name)
+
+      if pod.deleted?
+        rc.delete(pod.name) if pod.deleted?
```
msufa: Safer to double check? ;) the 2nd 'if pod.deleted?' is obsolete.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -0,0 +1,83 @@
...
+    def deploy_finished?
+      @release.release_docs.all?(&:live?)
+    end
+
+    def release_doc_from_rc_name(name)
+      @release.release_docs.select { |doc| doc.replication_controller_name == name }.first
+    end
+
+    def on_termination
+      send_event(msg: 'Finished Watching Deploy!')
```
msufa: It would be cleaner to also unsubscribe from notifications. I think we might see some stray messages in the logs in some edge cases.

plugins/kubernetes/samson_kubernetes.gemspec
```diff
@@ -1,7 +1,9 @@
-Gem::Specification.new "samson_kubernetes", "0.0.1" do |s|
-  s.summary = "Allow deploying projects using Kubernetes"
-  s.description = "TBD"
-  s.authors = ["Jon Moter"]
-  s.email = "jmoter@zendesk.com"
-  s.add_runtime_dependency "kubeclient", '>= 0.9'
+Gem::Specification.new 'samson_kubernetes', '0.0.1' do |s|
+  s.summary = 'Allow deploying projects using Kubernetes'
+  s.description = 'TBD'
+  s.authors = ['Jon Moter', 'Maciek Sufo', 'Sergio Nunes', 'Shane Hender']
```
msufa: How could you!

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
henders: Good point, I'll give that a go.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
henders: I'd say either put everything in the rails logger, or keep the k8s log file, but not 3 :)

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
henders: Oops, refactor fail! :)

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
henders: nice, will do!

plugins/kubernetes/samson_kubernetes.gemspec
henders: Oh I could! One gets blamed, we all get blamed! :smiling_imp:

msufa: My outrage was directed at the way you spelled my last name!

henders: Oh sorry man, my humble apologies, I'll make it up in future +1s :)

henders:
@msufa , ready for review again.

I don't need to unsubscribe because Celluloid takes care of it for me by linking to actors and trapping their exit.

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -50,51 +52,31 @@ def project
...
-      deploy_is_live
-    elsif kuber_release.release_docs.any?(&:spinning_up?)
-      kuber_release.update_attribute(:status, 'spinning_up') unless kuber_release.spinning_up?
+  # TODO: Remove this dummy data when proper watchers created.
+  def publish_fake_updates
+    Thread.new do
+      @kuber_release.release_docs.each do |release_doc|
+        release_doc.replica_target.times do
+          sleep 2
+          Celluloid::Notifications.publish("#{release_doc.replication_controller_name}", fake_event)
```
msufa: Just curious how does this work if `celuloid` is not `required`?

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -38,62 +39,21 @@ def nested_error_messages
...
-    end
-
     def docs_by_role
       @docs_by_role ||= release_docs.each_with_object({}) do |rel_doc, hash|
         hash[rel_doc.kubernetes_role.label_name] = rel_doc
       end
     end
 
+    def watch
+      Watchers::DeployWatcher.new(self).async :watch
```
msufa: This is ok for now, but if we ever want to have `DeployWatcher`s supervised, `watch` would need to be called from the initialisation code. Otherwise when this actor is restarted by a supervisor, `watch` will not be omitted. I'm struggling with a similar issue myself ATM.

msufa:
I left 2 additional low priority comments. If you're fine with the issues mentioned then :+1:

plugins/kubernetes/app/models/kuber_deploy_service.rb
henders: Good point, there is a gem that's supposed to autoload everything, but I moved the require into the plugin engine init file.

henders: Anyway, I moved the require into shared area just to be sure.

plugins/kubernetes/app/models/kubernetes/release.rb
henders: Solved this by kicking off the async call from within the actor itself. Celluloid seems to support this.

henders:
Addressed both comments as bonus! :)

msufa:
I sustain my :+1: :)

PR #640, opened 2015-11-16
Fail test script  if intermediate tasks fail.

henders:
Previous builds wouldn't fail if just ```npm run-script jshint``` failed.

/cc @sbrnunes @zendesk/samson @jonmoter 

### References
 - Jira link: 

### Risks
 - None


sandlerr:
nice catch :+1:

PR #642, opened 2015-11-17
move job restart after routes are loaded so flowdock notifications ca…

grosser:
…n use routes

 - rename too generic SignalHandler class
 - move job execution details into JobExecution
 - remove redundant checks, tables and precompile are not set when we are booting as server

@zendesk/runway 

/cc @steved 

### Risks
 - Low: restart failing / deploys not being continued on restart / deploys not halting on restart


config/application.rb
```diff
@@ -87,9 +87,18 @@ class Application < Rails::Application
...
 
-    config.after_initialize do
-      # Token used to request badges
-      unless ENV['PRECOMPILE']
+    if !Rails.env.test? && !ENV['PRECOMPILE'] && ENV['SERVER_MODE'] && Job.table_exists? && Deploy.table_exists?
+      initializer :execute_job, after: :set_routes_reloader_hook do # flowdock uses routes: run after the routes are loaded
+        JobExecution.enabled = true
+        JobExecution.stop_running_jobs
+        JobExecution.start_pending_deploys
+        RestartSignalHandler.listen
```
grosser: afaik this should only happen in server mode too ... so moving this into the block

steved:
How is an initializer better than after_initialize?

sandlerr:
can we do `Rails.application.reload_routes!` ? That would mean doing it twice but the initializer can stay in its own file

jonmoter:
@steved Samson died in production today with this stacktrace:

```
/data/samson/releases/20151116-224535-ead2eaf/plugins/flowdock/app/models/flowdock_notification.rb:5:in `project_deploy_url': undefined method `project_deploy_url' for #<Module:0x007f4ffeae9458> (NoMethodError)
        from /data/samson/releases/20151116-224535-ead2eaf/plugins/flowdock/app/models/flowdock_notification.rb:47:in `deploy_url'
        from /data/samson/releases/20151116-224535-ead2eaf/plugins/flowdock/app/models/flowdock_notification.rb:22:in `deliver'
        from /data/samson/current/plugins/flowdock/lib/samson_flowdock/samson_plugin.rb:19:in `block in <top (required)>'
        from /data/samson/releases/20151116-224535-ead2eaf/lib/samson/hooks.rb:134:in `call'
        from /data/samson/releases/20151116-224535-ead2eaf/lib/samson/hooks.rb:134:in `block in fire'
        from /data/samson/releases/20151116-224535-ead2eaf/lib/samson/hooks.rb:134:in `map'
        from /data/samson/releases/20151116-224535-ead2eaf/lib/samson/hooks.rb:134:in `fire'
        from /data/samson/releases/20151116-224535-ead2eaf/app/models/deploy_service.rb:70:in `send_before_notifications'
        from /data/samson/releases/20151116-224535-ead2eaf/app/models/deploy_service.rb:23:in `confirm_deploy!'
        from /data/samson/releases/20151116-224535-ead2eaf/app/models/deploy.rb:92:in `pending_start!'
        from /data/samson/releases/20151116-224535-ead2eaf/config/initializers/job_execution.rb:16:in `block (2 levels) in <top (required)>'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/activerecord-4.2.4/lib/active_record/relation/delegation.rb:46:in `each'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/activerecord-4.2.4/lib/active_record/relation/delegation.rb:46:in `each'
        from /data/samson/releases/20151116-224535-ead2eaf/config/initializers/job_execution.rb:15:in `block in <top (required)>'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/lazy_load_hooks.rb:36:in `call'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/lazy_load_hooks.rb:36:in `execute_hook'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/lazy_load_hooks.rb:45:in `block in run_load_hooks'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/lazy_load_hooks.rb:44:in `each'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.4/lib/active_support/lazy_load_hooks.rb:44:in `run_load_hooks'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/railties-4.2.4/lib/rails/application/finisher.rb:62:in `block in <module:Finisher>'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/railties-4.2.4/lib/rails/initializable.rb:30:in `instance_exec'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/railties-4.2.4/lib/rails/initializable.rb:30:in `run'
        from /data/samson/current/vendor/bundle/ruby/2.2.0/gems/railties-4.2.4/lib/rails/initializable.rb:55:in `block in run_initializers'
```

One of the pending deploys tried to send a Flowdock notification, which tried to use a route helper. But the route helpers hadn't been loaded yet, so it died.

According to Grosser, an initializer that runs `after: :set_routes_reloader_hook` is executed after the routes are all set, so we don't get that error.

steved:
Yeah, I saw the flow. But explain why after_initialize happens before an initializer.

sandlerr:
fyi after_initialize appears to be after routes https://github.com/rails/rails/blob/v4.2.4/railties/lib/rails/application.rb#L44-L55

grosser:
after_initialize is an initializer itself, it is called before the routes are loaded see:

https://github.com/rails/rails/blob/4-2-stable/railties/lib/rails/application/finisher.rb#L61-L83

app/models/job_execution.rb
```diff
@@ -76,6 +76,20 @@ def on_complete(&block)
     @subscribers << JobExecutionSubscriber.new(job, block)
   end
 
+  def self.stop_running_jobs
```
steved: I don't think any of this belongs in the class

app/models/restart_signal_handler.rb
```diff
@@ -1,11 +1,18 @@
 # JobQueue locks a mutex, hence the need for a separate SignalHandler thread
 # Self-pipe is also best practice, since signal handlers can themselves be interrupted
-class SignalHandler
+class RestartSignalHandler
+  class << self
+    alias listen new
+  end
+
   def initialize
+    Signal.trap('SIGUSR1') { signal }
```
steved: This should be after opening the pipe and the thread in case it gets signalled

app/models/restart_signal_handler.rb
```diff
@@ -1,11 +1,18 @@
 # JobQueue locks a mutex, hence the need for a separate SignalHandler thread
 # Self-pipe is also best practice, since signal handlers can themselves be interrupted
-class SignalHandler
+class RestartSignalHandler
+  class << self
+    alias listen new
```
steved: Why?

jonmoter:
:+1:

app/models/restart_signal_handler.rb
grosser: because calling RestartSignalHandler.new is not very intention conveying :D 
alternatives:
 - make it all static methods ?
 - make listen do a new and then call run ... but that means run will be public or I'll have to use send

app/models/restart_signal_handler.rb
grosser: good catch, changed!

app/models/job_execution.rb
grosser: yeah ... removed

app/models/restart_signal_handler.rb
steved: Why not just leave it like it was?

grosser: splits responsibility/knowledge into 2 places that have nothing to do with each other

grosser: `SignalHandler.new.listen` ?

steved: You're pretty much arguing semantics, it was still very clear what was happening before. The signal handler knows how to read from its own pipe and handle that input. The actual invocation is "customizable" and hence outside the signal handler.

grosser: for me it was not clear what was happening especially that USR1 was trapped outside and USR2 got raised inside ... so keeping them together made it easier to see this behavior

grosser: if we wanted to make it more testable I'd guess it should keep both the trap and the kill in some outside method ... 

`listen 'SIGUSR1' { Process.kill $$, 'SIGUSR2' }`

grosser: more like `listen 'SIGUSR1' { JobExecution.wait_for_jobs_to_finish; Process.kill $$, 'SIGUSR2' }` ... but that can be it's own PR if anyone wants that ...

steved:
You can merge the initializer change, but I think the other refactors are just moving code around for no purpose.

app/models/job_execution.rb
```diff
@@ -76,6 +76,16 @@ def on_complete(&block)
...
 
+  def self.start_pending_deploys
+    Job.non_deploy.pending.each do |job|
+      JobExecution.start_job(JobExecution.new(job.commit, job))
+    end
+
+    Deploy.active.each do |deploy|
+      deploy.pending_start! if deploy.pending_non_production?
+    end
+  end
```
steved: None of this stuff belongs in this file. JobExecution does not need to know about Deploy.

grosser: do you suggest it all in application.rb or job things here and deploy things in deploy ?

steved: Unless you're going to test it, leave it in the initializer

grosser: all in initializer now, good to go ?

PR #645, opened 2015-11-18
Watch pod events per cluster using non-blocking IO

msufa:
- monkey patch `kubeclient` to use a non-blocking http lib in watch streams
 - start a watcher per cluster on init
 - create/restart watchers when k8s clusters added/edited
 - remove obsolete classes from model

### References
 - Jira link: 

### Risks
 - None


plugins/kubernetes/config/initializers/watchers.rb
```diff
@@ -0,0 +1,65 @@
+require 'celluloid/current'
+require 'celluloid/io'
+require 'kubeclient'
+require 'http'
+require 'json'
+require 'logger'
+
+# monkey patching kubeclient to use non-blocking IO in watch streams
```
msufa: This is the original for reference: https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/watch_stream.rb

plugins/kubernetes/app/models/kubernetes/util.rb
```diff
@@ -1,3 +1,5 @@
+require 'celluloid/current'
```
henders: This is now in the lib/samson_plugin.rb file, but it should be harmless.

plugins/kubernetes/app/models/kubernetes/util.rb
```diff
@@ -20,5 +22,20 @@ def self.log(message, extra_info = {})
...
+    def self.start_watcher(cluster)
+      watcher_name = pod_watcher_symbol(cluster)
+      Watchers::ClusterPodWatcher.supervise as: watcher_name, args: [cluster.client]
+    end
+
+    def self.restart_watcher(cluster)
+      watcher = Celluloid::Actor[pod_watcher_symbol(cluster)]
+      watcher.terminate if watcher and watcher.alive?
+      start_watcher(cluster)
+    end
```
henders: This seems an odd place to put this logic.  Why not a class method in the ```Watchers::ClusterPodWatcher``` itself?

henders:
You forgot to remove the line which causes the exception during deploy:
https://github.com/zendesk/samson/pull/645/files#diff-e25235b8653153a4ecb45b1cd3453a0dL22

:)

plugins/kubernetes/app/models/watchers/cluster_pod_watcher.rb
```diff
@@ -0,0 +1,61 @@
...
+      else
+        false
+      end
+    end
+
+    def handle_notice(notice)
+      debug notice.to_s
+      return if handle_error(notice)
+      rc_name = notice.object.metadata.labels['replication_controller']
+      publish rc_name, notice if rc_name
```
henders: Should use parenthesis to make this a bit more readable ```publish(rc_name, notice) if rc_name```

plugins/kubernetes/config/initializers/watchers.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+      end
+    end
+  end
+end
+
+Celluloid.logger = Rails.logger
+$CELLULOID_DEBUG = true
+
+Kubernetes::Cluster.all.each { |cluster| Kubernetes::Util::start_watcher(cluster) }
```
henders: Only run this during web server execution:
https://github.com/zendesk/samson/blob/master/config/initializers/zz.rb#L1

plugins/kubernetes/config/initializers/watchers.rb
```diff
@@ -0,0 +1,65 @@
...
+        @finished = true
+        @http.close unless @http.nil?
+      end
+
+      private
+
+      def get_client
+        !@http_options[:use_ssl] && @http_options[:basic_auth_user] && @http_options[:basic_auth_password] ?
+            HTTP.basic_auth(user: @http_options[:basic_auth_user], pass: @http_options[:basic_auth_password]) :
+            HTTP::Client.new
```
henders: Use an if/else format here, multiline ternaries are a bit hard to parse

PR #649, opened 2015-11-19
Tweaks

steved:
These are two optimizations we've had in other applications, but never got into here.

@zendesk/samson


sandlerr:
:+1:

PR #650, opened 2015-11-20
add a database config to the docker image

sandlerr:
/cc @zendesk/samson


steved:
I left it out specifically because DATABASE_URL should work.

PR #634, opened 2015-11-09
make deploy groups show a bit nicer

grosser:
@jonmoter 

before:
screen shot 2015-11-09 at 11 18 57 am

after:
screen shot 2015-11-09 at 11 18 14 am

### Risks
 - None

PR #652, opened 2015-11-22
[Fix #583] Use Slack's incoming webhook API

### LINKED ISSUE #583: Slack: use webhooks or incorporate groups
https://api.slack.com/incoming-webhooks
https://api.slack.com/methods/groups.list

HashNuke:
This PR improves the Slack plugin by using incoming's webhooks API as mentioned in #583.

Please do let me know if there's any detail I'm missing.

### Summary

* Adds `slack_webhooks` table
* Removes `slack_channels` table
* Updates fields on the "Slack" section on stage add/edit page (screenshot attached)
* Updates to models and decorator
* Removes slack-api gem dependency from slack plugin
* Removes initializer from slack plugin
* Updates tests
* Remove `SLACK_TOKEN` env var from `.env.example`

**[UPDATE]**

> **Channel field is optional because**: Slack webhooks are configured with default channel. If a channel isn't specified, the message ends up in the default channel. Channel field's value can be public channels like `#general` or specific users like `@someone`. https://api.slack.com/incoming-webhooks

> In the webhook, **we can specify the avatar for the message (let me know which one)**. I held back from using the zendesk logo at `http://github.com/zendesk.png` because users might confuse that with the official zendesk integration logo (if any).

### Screenshot of Slack fields on stage add/edit page

screen shot 2015-11-23 at 2 13 09 am

### Screenshot of Slack notification using incoming webhook API

screen shot 2015-11-23 at 2 03 54 am


HashNuke:
I could use some help with the failing `TASK=js` build job; It passes or fails at random. Here's a passing build - https://travis-ci.org/HashNuke/samson/builds/92612093 (the only extra change in the branch was to enable building all branches).

HashNuke:
@grosser Just noticed that you have the same error with the `TASK=js` build job on #634. Same with this PR's build.

HashNuke:
Fixed the build. Angular just needed more time to make this assignment. https://github.com/zendesk/samson/blob/master/plugins/flowdock/app/assets/javascripts/flowdock/application.js#L80. 100ms wasn't enough.

So when `buddyNotificationsCtrl.notificationBox` was accessed in the test, it was undefined. https://github.com/zendesk/samson/blob/master/plugins/flowdock/test/angular/buddy_request_box_spec.js#L129. I increased the timeout from 100ms to 1second.

According to some googling, we should be using `browser.waitForAngular()` instead. This would require that another dependency be added to the tests. I see this stuff as a topic for another PR instead.

HashNuke:
Yay ~! survived an update to master
@zendesk/samson I've resolved conflicts with master and updated the branch. Build is green again.

plugins/slack/app/models/slack_notification.rb
```diff
@@ -4,18 +4,23 @@ def initialize(deploy)
...
   def deliver
-    Slack.chat_postMessage(
-      token: ENV['SLACK_TOKEN'],
-      channel: @stage.slack_channels.first.channel_id,
-      text: content)
+    payload = {text: content, username: "samson-bot"}
+
+    if !@webhook.channel.blank?
+      payload.merge!({channel: @webhook.channel})
+    end
```
henders: ```payload.merge!({channel: @webhook.channel}) unless @webhook.channel.blank?``` to avoid negative 'if's.

henders:
There doesn't seem to be a migration step for people upgrading their Samson version. Their Slack configs will just disappear. I guess there is no way to determine the webhook URL for them.

That's my only concern, and I don't know how to address it.

plugins/slack/app/models/slack_notification.rb
HashNuke: @henders sure :+1: I changed it to `unless`.
I used negative-if because `unless` is a bit off to read. I have nothing against `unless`. I would gladly welcome `unless` to my house for dinner 😊

HashNuke:
@henders Yes there's no way for that.

Since we have the Slack API token on older installations, I checked if we could programmatically create a webhook to replace existing ones. But there's no API endpoint to manage Slack integrations.

HashNuke:
@henders I didn't want code smell by supporting both webhooks & channels. So I removed the older slack token support entirely (effect: removes a lot of code, removes a dependency, reduces a HTTP call and no more caching slack channel list)

henders:
Gotcha, was there a main reason to ditch the slack api gem and use faraday direct?

HashNuke:
@henders webhook API is just a plain HTTP call. Slack gives you a unique-long-url once you configure it (with a default channel). No authentication at all. So no library is required.

henders:
Cool, thanks @HashNuke for the explainations. LGTM, but pinging @jonmoter for a look as well.

HashNuke:
@henders thanks for looking into this PR.

plugins/slack/app/views/samson_slack/_fields.html.erb
```diff
@@ -1,18 +1,22 @@
...
+      </div>
+
+      <div class="col-lg-3 col-lg-offset-2">
+        <%= slack_fields.text_field :channel, class: "form-control", placeholder: "Channel (optional)" %>
       </div>
 
       <% if slack_fields.object.persisted? %>
-        <div class="col-lg-1 checkbox">
+        <div class="col-lg-offset-2 col-lg-10">
           <%= slack_fields.check_box :_destroy if slack_fields.object.persisted? %>
```
steved: Not that you wrote this, but the if statement should already be satisfied here

plugins/slack/app/views/samson_slack/_fields.html.erb
```diff
@@ -1,18 +1,22 @@
...
+        <%= slack_fields.text_field :webhook_url, class: "form-control", placeholder: "Webhook URL" %>
+      </div>
+
+      <div class="col-lg-3 col-lg-offset-2">
+        <%= slack_fields.text_field :channel, class: "form-control", placeholder: "Channel (optional)" %>
       </div>
 
       <% if slack_fields.object.persisted? %>
-        <div class="col-lg-1 checkbox">
+        <div class="col-lg-offset-2 col-lg-10">
```
steved: I'm being super nitpicky, but can we add the paragraph descriptor inside the form-group and have all three divs on the same line?

```
<p class="col-lg-8 col-lg-offset-2"></p>
<div class="col-lg-4 col-lg-offset-2"></div>
<div class="col-lg-3"></div>
<div class="col-lg-1 checkbox"></div>
```

And for this last one, place the actual checkbox inside the label:
```
<%= slack_fields.label ... do %>
  <%= slack_fields.check_box :_destroy %>
  Destroy
<% end %>
```

It'll end up looking like this:
screen shot 2015-11-30 at 10 27 36 am

plugins/slack/app/models/slack_notification.rb
```diff
@@ -4,18 +4,20 @@ def initialize(deploy)
...
+    payload = {text: content, username: "samson-bot"}
+    payload.merge!({channel: @webhook.channel}) unless @webhook.channel.blank?
 
-  rescue Slack::Error => e
+    Faraday.post @webhook.webhook_url, {payload: payload.to_json}
+  rescue Faraday::ClientError => e
     Rails.logger.error("Could not deliver slack message: #{e.message}")
   end
 
+
```
steved: extra newline

plugins/slack/app/models/slack_notification.rb
```diff
@@ -4,18 +4,20 @@ def initialize(deploy)
     @stage = deploy.stage
     @project = @stage.project
     @user = @deploy.user
+    @webhook = @stage.slack_webhooks.first
```
steved: You'll only end up notifying the first webhook, you should iterate inside #deliver

plugins/slack/app/models/slack_notification.rb
```diff
@@ -4,18 +4,20 @@ def initialize(deploy)
...
   end
 
+
   def deliver
-    Slack.chat_postMessage(
-      token: ENV['SLACK_TOKEN'],
-      channel: @stage.slack_channels.first.channel_id,
-      text: content)
+    payload = {text: content, username: "samson-bot"}
+    payload.merge!({channel: @webhook.channel}) unless @webhook.channel.blank?
```
steved: No need for {} when doing it as arguments (`payload.merge!(channel: ..)`)

plugins/slack/app/models/slack_notification.rb
```diff
@@ -4,18 +4,20 @@ def initialize(deploy)
...
 
-  rescue Slack::Error => e
+    Faraday.post @webhook.webhook_url, {payload: payload.to_json}
```
steved: Let's add the parens and drop the `{}`

steved:
Thanks! The code looks great, I'm going to need to think about how we do end up notifying / migrating existing users since we don't have any sort of versioning or mailing list type deal.

plugins/slack/app/models/slack_notification.rb
steved: I just realized it's the same as the old code... I think that's a bug. Unless this whole thing should be singleton.

HashNuke: @steved Yes. This was written as a one-to-many but it's being as a one-to-one. So I just stuck to whatever was present already.

HashNuke:
@steved There's no easy upgrade path. I checked if Slack allows programmatically creating webhook urls using the API (so that we could use the token to do this). But API doesn't have any endpoints for managing integrations.

I thought about supporting both the old Channels API and the new incoming webhooks API. But down the road that would turn out to be a mess (too much to support). So I removed/replaced it entirely.

steved:
I agree, your decision is fine. I just need to figure out a way to notify everyone and make it less painful.

HashNuke:
@steved finished up all changes. Feel free to ping me if there's anything else :)

PR #654, opened 2015-11-23
Allow 'bundle install' to be run inside subdirectories without screwi…

henders:
Allow 'bundle install' to be run inside subdirectories without screwing up plugin deps.
Sometimes a dev (possibly me) doesn't realize they're in a subdirectory while executing bundle install and wonder why various gems are getting deleted from vendor/cache.

/cc @zendesk/samson @sbrnunes @msufa 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -340,10 +340,10 @@ GEM
       rails-assets-angular (= 1.4.7)
     rails-assets-angular-scenario (1.4.7)
       rails-assets-angular (= 1.4.7)
-    rails-assets-angular-ui-bootstrap-bower (0.14.3)
-      rails-assets-angular (>= 1.3.0)
     rails-assets-angular-truncate-2 (0.4.1)
       rails-assets-angular (>= 1.2.0, < 2)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -340,10 +340,10 @@ GEM
       rails-assets-angular (= 1.4.7)
     rails-assets-angular-scenario (1.4.7)
       rails-assets-angular (= 1.4.7)
-    rails-assets-angular-ui-bootstrap-bower (0.14.3)
-      rails-assets-angular (>= 1.3.0)
     rails-assets-angular-truncate-2 (0.4.1)
       rails-assets-angular (>= 1.2.0, < 2)
+    rails-assets-angular-ui-bootstrap-bower (0.14.3)
+      rails-assets-angular (>= 1.3.0)
```
henders: Not sure why existing Gemfile.lock had them ordered wrong. They should have been alphabetized already.

sbrnunes: Weird indeed. I need to pay attention to that next time.

sbrnunes:
Cool. That's going to be handy. :+1:

PR #655, opened 2015-11-23
Replacing Angular Material dependencies

sbrnunes:
This PR replaces Angular Material dependencies by [UI Bootstrap](https://angular-ui.github.io/bootstrap) dependencies. Angular Material offers a richer set of components but, apparently, the CSS conflicts with Bootstrap, so, we should stick with UI Bootstrap.

/cc @steved, @zendesk/samson

### Risks
 - There shouldn't be any risk as UI Bootstrap is just a set of AngularJS directives built in top of Bootstrap.


plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
```diff
@@ -1,87 +1,28 @@
...
-.kubernetes [role="progressbar"]::after {
-  position: absolute;
-  font-size: 14px;
-  transform: scale(2) !important;
-  width: 150px;
-  text-align: center;
-  top: 140px;
-  left: -20px;
+us-spinner.kubernetes-roles-spinner [role="progressbar"]::after {
+  content: "Loading Kubernetes Roles";
```
henders: A bit weird to put non-styling content here, Were you not able to put this text between ```us-spinner``` tags?

app/assets/stylesheets/spinners.scss
```diff
@@ -0,0 +1,78 @@
...
+@keyframes spin {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
+
+
```
henders: Did the spin gem not already have all these settings?

sbrnunes: This is the spinning animation I'm using for the typeahead button. When you type something in the typeahead and then you click on the button to submit the reference, the button gets replaced by a icon that spins while the API call is ongoing. All this is done through CSS, so, to animate it, I need this CSS.

The spinner gem is being used for a different purpose: it offers an angular directive to display a pre-defined spinner that you can show as a modal (default) or in place of the content while the content is not rendered (that's what I'm doing).

But because both have spinning animations, I refactored it to this spinners.scss file.

plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
sbrnunes: I usually treat this "labels" as a part of the styling. It's very handy because you can treat it as part of the main "component" and you don't need to be creating extra elements (spans, paragraphs, etc). For simple things like this, I never read anything discouraging it's use.

But there's a better way to achieve both, which I forgot. You can link the content to an attribute of the element: 'content: attr(data-label)'. This way, the label can be provided from de element, and it's more DRY. We don't need this CSS.

sbrnunes: Fixed as you suggested (my approach wouldn't work because the spinner is actually inject a child element of the tag intercepted by this CSS).

app/assets/stylesheets/spinners.scss
henders: Gotcha, didn't realize that CSS was unrelated to the angular spinner.

henders:
:+1:  on :green_heart:

steved:
Thanks!

PR #656, opened 2015-11-23
Fixing roles spinner (should be visible only while loading the roles).

sbrnunes:
For testing purposes, I made this spinner always visible and then forgot to change it back.

/cc @henders, @zendesk/samson 

### Risks
 - None


henders:
:+1:  on :green_heart:

sbrnunes:
Apparently, Codeclimate was not being able to analyse this commit because the change was focused on a html file only, so, I'm going ahead with the merge.

PR #658, opened 2015-11-26
Fix socket shutdown in pod watcher

msufa:
System IO raises an EOFError when reading from a closed socket, which happens when we close a long running connection listening for pod events. Celluloid transforms EOFError into MailboxShutdown and terminates the actor quietly. We ned to rescue EOFError before it gets to Celluloid.

/cc @zendesk/samson @henders @sbrnunes 

### References
 - Jira link: 

### Risks
 - None


henders:
:+1:

PR #659, opened 2015-11-26
add some more NewRelic custom metrics to hunt down Samson slowness

jonmoter:
Adding some more New Relic instrumentation for parts of the code that might be problematic.

/cc @zendesk/samson

### Risks
 - Minimal risk, just New Relic instrumentation

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -1,6 +1,8 @@
 require 'shellwords'
 
 class JobExecution
+  include ::NewRelic::Agent::Instrumentation::ControllerInstrumentation
+
   # Whether or not execution is enabled. This allows completely disabling job
   # execution for testing purposes.
   cattr_accessor(:enabled, instance_writer: false) do
@@ -13,7 +15,7 @@ class JobExecution
   cattr_reader(:registry, instance_accessor: false) { JobQueue.new }
   private_class_method :registry
 
-  attr_reader :output, :job, :viewers, :executor
+  attr_reader :output, :reference, :job, :viewers, :executor
 
   delegate :id, to: :job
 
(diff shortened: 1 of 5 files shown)


app/models/job_execution.rb
```diff
@@ -125,6 +127,7 @@ def run!
     finish
     ActiveRecord::Base.clear_active_connections!
   end
+  add_transaction_tracer :run!, category: :task, params: '{ job_id: id, project: job.project.try(:name), reference: reference }'
```
sandlerr: did you want to double quote and interpolate here?

sandlerr:
:+1: on :green_apple:

app/models/job_execution.rb
jonmoter: Nope.  Single-quoted so that it will be interpolated when `run!` is called, rather than when `add_transaction_tracer` is called.  It's described in the [New Relic rubydocs](http://www.rubydoc.info/github/newrelic/rpm/NewRelic/Agent/Instrumentation/ControllerInstrumentation/ClassMethods#add_transaction_tracer-instance_method)

jonmoter:
Hmm, confused why Travis isn't running the tests...

jonmoter:
Dunno what's up with Travis, but the tests do pass, so I'll merge.

```
 ~/Code/zendesk/samson 0372626 jonmoter/more-method-tracers [8:19:23]
$ rake
Run options: --seed 23501

# Running:

..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 55.667531s, 19.2213 runs/s, 31.4007 assertions/s.

1070 runs, 1748 assertions, 0 failures, 0 errors, 0 skips
```

PR #660, opened 2015-11-27
handle invalid UTF-8 characters in terminal output

jonmoter:
Fix airbrake error we see when output contains an invalid UTF-8 character.

```
ArgumentError: JobExecutionSubscriber failed: invalid byte sequence in UTF-8
app/models/terminal_output_scanner.rb:29 in scan
app/models/terminal_output_scanner.rb:29 in write
app/models/terminal_output_scanner.rb:16 in block in each
app/models/output_buffer.rb:60 in each
...
```

/cc @zendesk/samson

### References
 - Airbrake error: https://zendesk.airbrake.io/projects/95346/groups/1539486345310184431
 - Inspiration for fix: https://robots.thoughtbot.com/fight-back-utf-8-invalid-byte-sequences

### Risks
 - Minimal risk, one-line change to output parsing code.

### DIFF
--- a/app/models/terminal_output_scanner.rb
+++ b/app/models/terminal_output_scanner.rb
@@ -26,8 +26,10 @@ class TerminalOutputScanner
   private
 
   def write(data)
-    data.scan(/\r?[^\r]*/).each do |part|
-      next if part == ""
+    data
+      .encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
+      .scan(/\r?[^\r]*/).each do |part|
+        next if part == ''
       write_part(part)
     end
   end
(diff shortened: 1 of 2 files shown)


sandlerr:
:+1:

fneves:
I'm a little late to the game but won't `scrub` work here?

`http://ruby-doc.org/core-2.1.0/String.html#method-i-scrub`

jonmoter:
Ah, I wasn't aware of `scrub`.  That would probably work too.

steved:
This PR scrubs the prefix character from this line:
https://github.com/zendesk/samson/blob/master/app/models/terminal_executor.rb#L24

String#scrub doesn't.

```
irb(main):008:0> File.read('test')
=> "»\n"
irb(main):009:0> File.read('test').scrub
=> "»\n"
irb(main):010:0> File.read('test').encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
=> "\n"
```

PR #661, opened 2015-11-27
Fixing plugin migration path (added by mistaken to core migration folder)

sbrnunes:
By mistaken, I added this migration to the core migrations folder when it should have been added to the Kubernetes migrations folder.

/cc @zendesk/samson


jonmoter:
:+1:

PR #662, opened 2015-11-27
test for invalid git refs to avoid InvalidURI crash

jonmoter:
Fix a `URI::InvalidURIError` that was thrown in the `CommitStatusesController` or `JobExecution` if you specify an invalid reference, like `current_account['url']`

/cc @zendesk/samson
### References
- Error in CommitStatusesController: https://zendesk.airbrake.io/projects/95346/groups/1527665412157157182
- Error in JobExecution: https://zendesk.airbrake.io/projects/95346/groups/1563043716556859438

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -9,6 +9,11 @@ class GitRepository
     Rails.application.config.samson.cached_repos_dir
   end
 
+  def self.valid_ref_format?(reference)
+    IO.popen(['git', 'check-ref-format', "refs/head/#{reference}"]) { |_| }
+    $?.success?
+  end
+
   def initialize(repository_url:, repository_dir:, executor: nil)
     @repository_url = repository_url
     @repository_directory = repository_dir
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -70,6 +70,8 @@ class Changeset
   def find_comparison
     if empty?
       NullComparison.new(nil)
+    elsif !GitRepository.valid_ref_format?(commit)
+      NullComparison.new("Invalid git reference: '#{commit}'")
     else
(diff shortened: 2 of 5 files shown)


app/models/git_repository.rb
```diff
@@ -9,6 +9,11 @@ class GitRepository
     Rails.application.config.samson.cached_repos_dir
   end
 
+  def self.valid_ref_format?(reference)
+    IO.popen(['git', 'check-ref-format', "refs/head/#{reference}"]) { |_| }
```
sandlerr: `heads` should be plural here. Are we not checking that it actually exists because that would force us to pull? Can't we hit the Github API in `valid_ref?` instead?

jonmoter: No, `check-ref-format` just confirms that the reference is valid.

I'm checking for the validity of the reference before I make the call to GitHub, because if you pass a reference like `current_account['url']` into the Octokit library, it fails and throws an `URI::InvalidURIError`.

Another way of fixing this bug would be to add `rescue URI::InvalidURIError` clauses in the same places where we're currently rescuing Octokit errors.

sandlerr: it might make more sense to just rescue the error... there's no rule saying a valid git reference will be parsed into a valid URL.
`check-ref-format 'refs/heads/master&asdf#111'; echo $?` printed `0` for me but I suspect anything after the `#` might not even get sent to the server, depending on your choice of HTTP library.

jonmoter: Hmm, okay.  True we might want to rescue the exception.

I also discovered that in production, somehow one of the `Deploy` records in the DB had a commit with a value of `current_account['url'] + '`, which was causing these AirBrake errors.  I think we should also add a validation to objects that expect a commit or a git reference, like `Deploy`, `Release`, or `Job`.

PR #663, opened 2015-11-30
Increasing API version to make the cache compatible with the api 20

fneves:
@henders @andreionut 

Cache in the docker binary plugin is restricted to work with docker api version 19. There are no changes around this [version 20](https://docs.docker.com/engine/reference/api/docker_remote_api/#v1-20-api-changes).
Just bumping the version to make the mounts work with version 20.


andreionut:
:+

PR #664, opened 2015-11-30
Add some tests to K8s deploy watcher.

henders:
Just adding some tests to the K8s plugin deploy watcher class.

/cc @sbrnunes @msufa 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/test/models/watchers/deploy_watcher_test.rb
+++ b/plugins/kubernetes/test/models/watchers/deploy_watcher_test.rb
@@ -1,0 +1,63 @@
+require_relative "../../test_helper"
+require 'celluloid/current'
+
+
+describe Watchers::DeployWatcher do
+  let(:deploying_release) { kubernetes_releases(:test_release_deploying) }
+  let(:complete_release) { kubernetes_releases(:test_release_complete) }
+
+  before do
+    Celluloid.shutdown; Celluloid.boot
+    Watchers::DeployWatcher.any_instance.stubs(:end_deploy)
+  end
+  after { Celluloid.shutdown }
+
+  describe '#initialize' do
+    it 'starts watching incomplete deploys' do
+      Watchers::DeployWatcher.any_instance.expects(:watch).once
+      Watchers::DeployWatcher.new(deploying_release)
+    end
+
+    it 'does not watch completed deployments' do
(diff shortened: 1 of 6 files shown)


plugins/kubernetes/test/models/watchers/deploy_watcher_test.rb
```diff
@@ -0,0 +1,63 @@
...
+  describe '#watch' do
+    it 'sets deploy as finished when all pods active' do
+      watcher = Watchers::DeployWatcher.new(deploying_release)
+      deploying_release.release_docs.each do |release_doc|
+        release_doc.replica_target.times do |i|
+          msg = create_msg(name: "foo-#{i}", ready: 'True')
+          watcher.handle_update(release_doc.replication_controller_name, msg)
+        end
+      end
+      watcher.send(:deploy_finished?).must_equal true
```
msufa: You can just say `watcher.deploy_finished?`

henders: nope, it's a private method.

msufa: D'oh, I thought you were trying to do some async magic.

msufa:
:+1: on :green_heart:

henders:
Yup, that :green_heart: will be harder to come by, will need to change Samson over to use DatabaseCleaner to get rid of the default transactions.  Celluloid tests are failing due to DB locks while the main test transaction is open.

henders:
@msufa , @sbrnunes , @jonmoter 
Added some dodginess to the tests to allow threaded tests to succeed with transactions. Another +1 desired!

sbrnunes:
I haven't seen any problems from this change, so, I wouldn't mind to give you a :+1:. 
However, I would prefer if someone with more experience than me could take a look at it as well: @jonmoter, @steved ?

This PR is required in order to get #717 and #700 merged.

sbrnunes:
Also, @henders, regarding the conflicts, at this stage, I think you should be able to just revert those tests and let the ones in master prevail.

steved:
I know nothing about this code, but I'm not a huge fan of such a wide-reaching patch. The only solutions I can think of are DI and more fine-grained unit tests with stubbing for the threaded updates.

PR #665, opened 2015-11-30
Cluster pod watcher tests

msufa:
Adding tests for the cluster pod watcher.

/cc @henders @sbrnunes 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/test/models/watchers/cluster_pod_watcher_test.rb
+++ b/plugins/kubernetes/test/models/watchers/cluster_pod_watcher_test.rb
@@ -1,0 +1,220 @@
+require_relative '../../test_helper'
+require 'celluloid/current'
+require 'kubeclient'
+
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/test/models/watchers/cluster_pod_watcher_test.rb
```diff
@@ -0,0 +1,220 @@
+require_relative '../../test_helper'
+require 'celluloid/current'
+require 'kubeclient'
+
+class DummyClient
```
henders: Normally we'd put these support classes either into their own files under 'test/support', or else put them down below the tests so that we see the tests first.

henders:
:+1:  with or without addressing the comment.

msufa:
Helper classes moved to the end of the test code.

PR #666, opened 2015-11-30
PAAS-93 Create Release Wizard for deploying to K8s

sbrnunes:
This PR adds support for creating, and deploying, a new Kubernetes Release. Tests will be added in a subsequent PR to avoid making this one even bigger.

Note that previous replication controllers won't be deleted. That functionality is still missing.

/cc @zendesk/samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-93

### Risks
 - None

screenshot 2015-11-30 16 17 48
screenshot 2015-11-30 16 18 06
screenshot 2015-11-30 16 18 39
screenshot 2015-11-30 16 17 22


app/assets/javascripts/app.js
```diff
@@ -16,3 +16,36 @@ var samson = angular.module("samson", [
...
+  },
+
+  /*
+    Returns a new instance of the array with shallow-copied versions of it's objects.
+    See http://underscorejs.org/#clone for more information.
+   */
+  cloneArray: function(arr) {
+    return _.map(arr, _.clone);
+  }
+});
```
henders: I would move these utility functions into a separate file. The app.js is just declaring the angular app.

app/assets/javascripts/services/builds_svc.js
```diff
@@ -0,0 +1,22 @@
+samson.service('buildsService', function($http, $q, httpErrorService) {
+  var config = {
+    headers: {
+      'Accept': 'application/json'
+    }
+  };
```
henders: I think an alternative to setting the json header is to append '.json' to the URL. But I don't mind which you use.

app/assets/javascripts/services/notification_svc.js
```diff
@@ -1,6 +1,7 @@
 samson.service('notificationService', function(messageCenterService) {
-  this.info = function(message, options) {
+  var _this = this;
```
henders: usually this is labelled 'self' instead of '_this'. I think I picked that up from the 'Best parts of JS' book.

app/assets/javascripts/services/builds_svc.js
sbrnunes: Yep, I realised that in the meantime. I usually like to have the backend deciding which format to use depending on the 'Accept' header sent by the client in the HTTP request. This could even be set globally, but I was not sure about the impact, so I'm doing it service by service. :)

But I'm fine both ways as well. This type of "convention over configuration" can be found in many other frameworks as well.

app/assets/javascripts/services/notification_svc.js
sbrnunes: True. Was trying to remember the exact convention for that... :-P

I'm going to change that.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_release_ctrl.js
```diff
@@ -0,0 +1,206 @@
...
+      if ($scope.allToggled()) {
+        _.each($scope.deploy_groups, function(deploy_group) {
+          deploy_group.selected = false;
+        });
+      }
+      else {
+        _.each($scope.deploy_groups, function(deploy_group) {
+          deploy_group.selected = true;
+        });
+      }
```
henders: Collapse this down to:
```
var newState = !$scope.allToggled();
_.each($scope.deploy_groups, function(deploy_group) {
     deploy_group.selected = newState;
});
```

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_release_ctrl.js
```diff
@@ -0,0 +1,206 @@
...
+      kubernetesService.loadRoles($stateParams.project_id).then(
+        function(roles) {
+          $scope.loadingRoles = false;
+          $scope.roles = roles;
+
+          //todo: merge with environment defaults
+
+          deferred.resolve(roles);
+        },
+        function(result) {
```
henders: Isn't the error case already handled in the Service itself? Notifying the error?

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_release_wizard_ctrl.js
```diff
@@ -0,0 +1,31 @@
...
+  $scope.next = function() {
+    showStep($scope.currentStep + 1);
+  };
+
+  $scope.previous = function() {
+    showStep($scope.currentStep - 1);
+  };
+
+  function showStep(step) {
+    _.findWhere($scope.steps, {step: step}).current = true;
```
henders: Using a find is probably overkill here since it's just an ordered array of steps.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_release_ctrl.js
sbrnunes: Good catch.

plugins/kubernetes/app/assets/javascripts/kubernetes/services/kubernetes_service.js
```diff
@@ -15,9 +15,7 @@ samson.service('kubernetesService', function($http, $q, httpErrorService, kubern
 
     $http.get('/projects/' + project_id + '/kubernetes_roles', config).then(
       function(response) {
-        deferred.resolve(response.data.map(function(role) {
-          return kubernetesRoleFactory.build(role);
-        }));
+        deferred.resolve(response.data.map(kubernetesRoleFactory.build));
```
henders: nice cleanup.

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -54,6 +54,28 @@ def deploy_group_ids
     end
 
     def deploy_group_ids=(_id_list)
```
henders: Should not need this function anymore now that you cleaned up the controller.

plugins/kubernetes/app/views/kubernetes_releases/_create_form.html.erb
```diff
@@ -0,0 +1,20 @@
+<script type='text/ng-template' id='kubernetes/kubernetes_release_wizard.tmpl.html'>
+  <div class="dialog kubernetes-release-dialog" ng-controller="KubernetesReleaseWizardCtrl">
+    <header>
+      <h2 class="title">Create a new release</h2>
+      <span class="step">Step {{current}} of 2</span>
```
henders: If you wanted to get fancy you could do ```Step {{current}} of {{$scope.steps.length}}```

plugins/kubernetes/app/views/kubernetes_releases/_create_form_step_1.html.erb
```diff
@@ -0,0 +1,115 @@
...
+
+        <div class="form-group clearfix">
+          <label for="environment">
+            <span class="order">2.</span>
+            <span>Select the environment:</span>
+          </label>
+
+          <div id="environment-dropdown" class="btn-group" uib-dropdown uib-keyboard-nav>
+            <button type="button" class="btn btn-default" uib-dropdown-toggle>
+              {{ environment.name || 'Choose environment' }} <span class="caret"></span>
```
henders: If there are no environments on the system,  I think this will throw an exception when trying to access ```environment.name```

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_release_ctrl.js
sbrnunes: I did some refactoring recently across the Services. Services are now responsible for making the call against the backend, parsing the response and returning the data or a result object with errors / warnings. 

In this case, the service will handle a response error and reject the promise with a result, returned by the `httpErrorService` or by the Service itself, in the case of a warning.

The controllers are now only dealing with the decision regarding what to do with the result object and updating the View accordingly, in this case, by using the `notificationService` to show a flash message for each `warning` / `error` returned from the service.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_release_wizard_ctrl.js
sbrnunes: I can replace it by `$scope.steps[step - 1].current`.

The reason why I went with this approach was to make it safer, i.e., not being dependent on the order those steps are declared. Regarding the performance, there's just 2 steps in there, so, it wouldn't become a problem.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_release_ctrl.js
henders: Ah I see, ignore my comment then! :)

app/controllers/admin/environments_controller.rb
```diff
@@ -1,10 +1,15 @@
 class Admin::EnvironmentsController < ApplicationController
-  before_action :authorize_admin!
+  before_action :authorize_admin!, except: [ :index ]
```
msufa: we usually don't use leading and trailing spaces with arrays

plugins/kubernetes/app/views/kubernetes_releases/_create_form.html.erb
sbrnunes: True. :-D

I'm gonna change it. Honestly, it crossed my mind, but then I thought: we'll never gonna change the number of steps. :-D

plugins/kubernetes/app/views/kubernetes_releases/_create_form_step_1.html.erb
sbrnunes: Humm, I confess I didn't even think about that scenario. Is it a valid scenario?

The other thing I'm not sure about is if AngularJS tries to resolve this when this section is not visible. This entire section will be displayed only if the `wizardReady()` check on the top most parent div succeeds. That check succeeds only if builds and environments were successfully loaded.

I'll give it a try, anyway.

sbrnunes: Nop, nothing fails, but the form will remain in a "Initialising form" state forever, if there's no environments.

So, I'm wondering now if this is valid or if I should do the same as for the builds: ask the User to create an environment first? I was assuming the User would have created an environment by the time he gets to this wizard.

app/controllers/admin/environments_controller.rb
sbrnunes: Good catch. Forgot to use the auto-format thing in this file.
Thanks.

msufa:
In the example screens I only see `Step of 2`. Are those early versions or is the current step number actually missing?

sbrnunes:
I had a bug. Had renamed the name of a variable ($scope.currentStep) and forgot to update it in the HTML (this fails silently). I fixed it when Shane suggested we could get the number of steps from the steps array.

In summary, those screenshots should be displaying: Step 1 of 2, Step 2 of 2.

sbrnunes:
@henders, @msufa, just committed a new change. Realised that we had the ability to deploy to Kubernetes from the build screen, but that's no longer supported. If, later, we'd like to have that ability again, we need to trigger the wizard from that screen and pre-select the build.

For now, I'm just removing the button because it leads to a broken route.

plugins/kubernetes/app/views/kubernetes_releases/_create_form_step_1.html.erb
henders: I was only thinking of the case of a new Samson install, and the user in their eagerness checks out the K8s support being fully configuring Samson to match their setup.

henders:
:+1:

plugins/kubernetes/app/views/kubernetes_releases/_create_form_step_1.html.erb
sbrnunes: We could make it fail earlier, i.e., when creating a new Cluster, we could make it mandatory to select, at least, a Deploy Group. To create a Deploy Group, we already need to select an environment.

My only question here is: does it make sense to create a Cluster without assigning it to at least a Deploy Group (aka namespace in the Kubernetes world)?

I'm just trying to think where would it makes sense to ask the User to create the missing stuff: if during the setup of the Cluster or later, when trying to release.

PR #667, opened 2015-11-30
save 40 character sha for job's commit

anthonywoo:
When saving the commit for a job - we were saving the shortened. (7 character sha).

However, some of the github endpoint require the long 40 character sha such as creating a status: https://developer.github.com/v3/repos/statuses/#create-a-status

If you want to use the shortened sha - you can still use the `short_reference` from a job: https://github.com/zendesk/samson/blob/master/app/models/job.rb#L145

@zendesk/samson

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -161,7 +161,7 @@ class JobExecution
   def setup!(dir)
     locked = lock_project do
       return false unless @repository.setup!(dir, @reference)
-      commit = @repository.commit_from_ref(@reference)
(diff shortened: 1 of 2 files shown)


steved:
Can you check where this is referenced in the UI and update it to short_reference?

anthonywoo:
@steved I couldn't find any references of the job's commit in the UI. It looks like we only display the deploy's reference - which can be a tag or the shortened sha.

steved:
+1

PR #669, opened 2015-12-01
Use the Github API to check if last release already contains a new commit

henders:
There are cases where webhooks cause many Releases/tags to be created because the samson cached_repo isn't updated often enough, so Samson believes there hasn't been a release created for commits that have already been processed.

As a result you get releases with no associated commits showing up. Plus their tags are pushed to Github.

This fix is just a proposal. Alternates were:
- Configure Github to send Samson notifications when commits occur which are picked up by a background thread in Samson and updates the cached_repo. This would keep the cached_repo up-to-date but would be a hog with resources.
- Sync the cached_repo whenever a webhook is processed. This also will heavily affect the resource usage of the Samson process, though not as much as the above.

/cc @bquorning @jonmoter @grosser @sandlerr @msufa @sbrnunes @zendesk/samson 

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-190

### Risks
 - Breaks any bitbucket/gitlab support for webhooks

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -66,7 +66,8 @@ class Project < ActiveRecord::Base
 
   def last_release_contains_commit?(commit)
     last_release = releases.order(:id).last
-    last_release && repository.downstream_commit?(last_release.commit, commit)
+    # status values documented here: http://stackoverflow.com/questions/23943855/github-api-to-compare-commits-response-status-is-diverged
+    last_release && %w(behind identical).include?(GITHUB.compare(github_repo, last_release.commit, commit).status)
   end
 
   def auto_release_stages
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -68,6 +68,9 @@ class Project < ActiveRecord::Base
     last_release = releases.order(:id).last
     # status values documented here: http://stackoverflow.com/questions/23943855/github-api-to-compare-commits-response-status-is-diverged
     last_release && %w(behind identical).include?(GITHUB.compare(github_repo, last_release.commit, commit).status)
+  rescue Octokit::Error => e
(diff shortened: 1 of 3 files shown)


bquorning:
:+1: code looks fine. I can’t wait to give it a try.

grosser:
:+1: surprisingly small fix for such a shitty problem, good job :D

app/models/project.rb
```diff
@@ -66,7 +66,8 @@ def docker_repo
 
   def last_release_contains_commit?(commit)
     last_release = releases.order(:id).last
-    last_release && repository.downstream_commit?(last_release.commit, commit)
+    # status values documented here: http://stackoverflow.com/questions/23943855/github-api-to-compare-commits-response-status-is-diverged
+    last_release && %w(behind identical).include?(GITHUB.compare(github_repo, last_release.commit, commit).status)
```
jonmoter: Should have some handling for Octokit raising an exception.

jonmoter:
Other than the error handling issue, looks good to me.

sandlerr:
would it make sense to just check if the current commit is already tagged? I can't see a situation where we want to generate a release from a tagged commit.

:+1: this looks good!

henders:
@jonmoter , added error handling!

henders:
@sandlerr , the only reason I'd see people wanting multiple tags on a commit is if they want to use  samson auto-tagging v21,v22,v23,.... and periodically use their own tags for milestones: v1.0, v2.0, ...
But they would probably tag it after Samson finished creating its own tag anyway.
I agree it would probably catch more edge cases, but want to see how this performs first.

jonmoter:
:+1:

PR #673, opened 2015-12-02
Use kubeclient with non-blocking IO support

msufa:
Remove the watcher monkey patch once https://github.com/abonas/kubeclient/pull/142 gets merged to kubeclient.

### References
 - Jira link: 

### Risks
 - None


msufa:
The corresponding PR in `kubeclient` has been merged and tagged to `v1.1`. This PR is no longer WIP and is ready to be reviewed.

Gemfile.lock
```diff
@@ -179,6 +178,7 @@ GEM
       dotenv (= 0.11.1)
     erubis (2.7.0)
     eventmachine (1.0.8)
+    eventmachine (1.0.8-java)
```
henders: Don't know why it added this gem! Maybe remove it?

henders:
:+1:

Gemfile.lock
steved: we need to remove java from the platforms once and for all

PR #674, opened 2015-12-03
Fix creating cluster deploy groups

msufa:
Set up proper associations between the deploy group and cluster deploy group.

Fixes:

screen shot 2015-12-03 at 3 55 55 p m

### References
 - Jira link: 

### Risks
 - None


sbrnunes:
Nice one. Thanks for fixing this bug. :+1:

PR #675, opened 2015-12-04
Make compatible with main docker hub registry.

henders:
Allow docker images to be uploaded to the main Docker Hub registry. Required because the new Kubernetes Deployment API is only available on the Google Compute Engine environment, so needed to use a docker registry that hosts in that environment could access.

Now if you use Docker Hub, for example, you just have to set the following ENV vars:
DOCKER_REGISTRY=registry.hub.docker.com/shender
DOCKER_REGISTRY_USER=shender
DOCKER_REGISTRY_PASS=xxxxx
DOCKER_REGISTRY_EMAIL=shender@zendesk.com

/cc @zendesk/samson @jonmoter @sbrnunes @msufa 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -53,7 +53,7 @@ class DockerBuilderService
 
     output_buffer.puts("### Pushing Docker image to #{project.docker_repo}:#{build.docker_ref}")
 
-    result = build.docker_image.push(registry_credentials) do |output_chunk|
+    build.docker_image.push(registry_credentials) do |output_chunk|
       parsed_chunk = handle_output_chunk(output_chunk)
 
       status = parsed_chunk.fetch('status', '')
@@ -61,12 +61,8 @@ class DockerBuilderService
         build.docker_repo_digest = "#{project.docker_repo}@#{matches[1]}"
       end
     end
-    # Fallbacks if registry did not send us back a digest
-    build.docker_repo_digest ||= result.info['RepoDigests'].first if result.info['RepoDigests']
-    build.docker_repo_digest ||= result.info['RepoTags'].first if result.info['RepoTags']
 
     build.save!
-
     build
   rescue Docker::Error::DockerError => e
     output_buffer.puts("Docker push failed: #{e.message}\n")
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -53,7 +53,7 @@ class DockerBuilderService
 
     output_buffer.puts("### Pushing Docker image to #{project.docker_repo}:#{build.docker_ref}")
 
-    result = build.docker_image.push do |output_chunk|
+    result = build.docker_image.push(registry_credentials) do |output_chunk|
       parsed_chunk = handle_output_chunk(output_chunk)
 
       status = parsed_chunk.fetch('status', '')
@@ -99,5 +99,14 @@ class DockerBuilderService
   def send_after_notifications
     Samson::Hooks.fire(:after_docker_build, build)
     SseRailsEngine.send_event('builds', { type: 'finish', build: BuildSerializer.new(build, root: nil) })
+  end
+
+  def registry_credentials
+    return nil unless ENV['DOCKER_REGISTRY_USER'].present? || ENV['DOCKER_REGISTRY_EMAIL'].present?
+    {
+      username: ENV['DOCKER_REGISTRY_USER'],
+      password: ENV['DOCKER_REGISTRY_PASS'],
+      email: ENV['DOCKER_REGISTRY_EMAIL']
+    }
   end
 end
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -48,11 +48,12 @@ class DockerBuilderService
 
   def push_image(tag)
     build.docker_ref = tag || build.label.try(:parameterize) || 'latest'
+    build.docker_repo_digest = nil
     build.docker_image.tag(repo: project.docker_repo, tag: build.docker_ref, force: true)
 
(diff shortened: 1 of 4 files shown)


app/models/docker_builder_service.rb
```diff
@@ -60,6 +61,7 @@ def push_image(tag)
         build.docker_repo_digest = "#{project.docker_repo}@#{matches[1]}"
       end
     end
+    build.docker_repo_digest ||= result.info['RepoTags'].first
```
sbrnunes: @henders, was trying to understand this line. So, as far as I understood, when we upload a docker image, we keep it's digest in the corresponding build. So, in this case, when we're not able to get the digest from the response, what's the meaning of that `RepoTag`? Does it contain a digest of the image, as well, or is it just a reference to identify that image?

The other thing I did not understand is how do we know this `RepoTags`is returned and what does it mean? Was trying to find documentation in the Docker API v1.16, but could only find this info in the List API. :)

henders: It looks like Docker Hub is not sending back a digest when pushing an image. Unlike our private registry which does.
So I fall back to a tagged version in case it didn't.

sbrnunes: Could we make that column name more generic, like `docker_reference` / `docker_repo_reference`? The reference would be a digest, for our private repo or a tag for Docker Hub. Don't know if it's worth the effort, just food for thoughts.

henders: Aye, I agree with sentiment. But don't want to change the column name in this PR.

app/models/docker_builder_service.rb
```diff
@@ -98,4 +102,13 @@ def send_after_notifications
     Samson::Hooks.fire(:after_docker_build, build)
     SseRailsEngine.send_event('builds', { type: 'finish', build: BuildSerializer.new(build, root: nil) })
   end
+
+  def registry_credentials
+    return nil unless ENV['DOCKER_REGISTRY_USER'].present? || ENV['DOCKER_REGISTRY_EMAIL'].present?
```
sbrnunes: Just checking: are we using an `OR` in here because the API requires one of the two, `username` or `email`, plus the password, for authentication? I mean, what happens if `DOCKER_REGISTRY_USER` was not provided but `DOCKER_REGISTRY_EMAIL` was? Will it go through?

app/models/docker_builder_service.rb
sbrnunes: Fair enough. :wink:

sbrnunes:
After checking my last comment, :+1:

jonmoter:
Hold on a sec, guys. We need to run this by security.

Using the digest is the method we use to prove we're actually deploying the code we think we're deploying, and not depending on a label/tag that can be changed.

app/models/docker_builder_service.rb
```diff
@@ -60,6 +61,9 @@ def push_image(tag)
         build.docker_repo_digest = "#{project.docker_repo}@#{matches[1]}"
       end
     end
+    # Fallbacks if registry did not send us back a digest
+    build.docker_repo_digest ||= result.info['RepoDigests'].first if result.info['RepoDigests']
+    build.docker_repo_digest ||= result.info['RepoTags'].first if result.info['RepoTags']
```
jonmoter: No, do not set the digest to anything other than the cryptographic SHA.  That's the entire point of that column.

If Docker Hub doesn't support digests, then we need to consider whether we can safely deploy code from Docker Hub.

jonmoter: Also, the `docker_ref` column is intended to refer to the tag of the image.

:-1:

henders:
Hey @JonMoter, the Docker Hub support was for development purposes and for other people outside Zendesk to use. So it wouldn’t affect compliance/security.

But as an alternative I could leave the digest blank, and use the ```docker_ref``` field to build the docker image path if the digest is blank.

henders:
@jonmoter ,  I removed the digest override. So at least it can push up to registries needing authentication.

app/models/docker_builder_service.rb
henders: It looks like Docker requires both the username and email to be set.  Even GCE which doesn't use the email field says to just fill it with a random string.

sbrnunes:
@henders, without the docker_repo_digest I can't make anything work. The Kubernetes release relies on that, it checks if that field has been set. It won't allow you to go through with creating a release.

I also noticed that when we're building the replication controller specs, the image field of the container specs pulls that digest from the build.

jonmoter:
Okay.  I understand now.

I can see other people using Samson where they don't require a digest for every image they deploy, and just rely on Docker tags.  Such as if they're using GCE.  That would mean adding the ability to use `Build#docker_ref` instead of `Build#docker_repo_digest` to refer to the Docker image.

I'd like the default behavior to be using the digest, if one exists.  Maybe we need some ability to enable/disable that with an ENV var, or some admin setting. If we leave this functionality in when we go live with Kubernetes deployments, we'll need to run this by security & compliance.

henders:
@jonmoter ,

Yup, that was the plan with the Kubernetes plugin to use the docker_ref as the fallback. That's coming in a PR I'm currently working on. @sbrnunes , this PR is setting the groundwork for that small change in the plugin to use docker_ref if no digest is available.

sbrnunes:
@jonmoter, regarding this last comments, can this PR be merged, then? I'm asking because @henders is on holidays but this changes are very useful for our development lifecycle, as we're using Google's GCE.

jonmoter:
Yeah, I'm okay with it, since it's no longer overwriting the digest column.

:+1:

PR #681, opened 2015-12-08
Fetch initial kubernetes dashboard data

msufa:
We need to fetch the initial data for the dashboard. Output example:

```
[
  {
    "name": "App",
    "deploy_groups": [
      {
        "name": "Pod1",
        "releases": [
          {
            "id": "88",
            "build": "kuber-test",
            "target_replicas": 2,
            "live_replicas": 2
          },
          {
            "id": "89",
            "build": "kuber-test",
            "target_replicas": 1,
            "live_replicas": 1
          }
        ]
      }
    ]
  }
]
```

The controller takes the environment from request params, it can be tested locally using `http://localhost:3000/projects/truth_service/kubernetes_dashboard?environment=1`

I needed to add an additional label to each deployed pod: `role_id`

Unit Tests will follow in a separate PR.

### References
 - Jira link: PAAS-85

### Risks
 - None


plugins/kubernetes/app/models/watchers/events/pod_event.rb
```diff
@@ -0,0 +1,26 @@
+module Watchers
+  module Events
+    class PodEvent
+      def initialize(event)
+        @event = event
+        @api_pod = Kubernetes::Api::Pod.new(event.object)
+      end
+
+      def ready?
```
henders: Maybe make @api_pod an ```attr_reader```, then consumers of this class can call ```pod_event.pod.ready?``` instead of having to create wrapping methods for each method you want to expose.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+                                          releases: hash_with(role_id, &default_release) } }
+  end
+
+  def default_release
+    lambda { |release_id, role_id| { id: release_id, build: get_build_label(release_id),
+                                     target_replicas: get_target_replicas(release_id, role_id),
+                                     live_replicas: 0 } }
+  end
+
+  def get_role_name(role_id)
```
henders: no need to prefix these methods with 'get_'

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+        release = roles[labels.role_id][:deploy_groups]\
+                      [cluster_deploy_group.id][:releases][labels.release_id]
+        api_pod = Kubernetes::Api::Pod.new(pod)
+        release[:live_replicas] += 1 if api_pod.ready?
+      end
+    end
+    roles
+  end
+
+  def hash_with(*optional_params)
```
henders: I'm sure there is something clever going on here with blocks and lambdas, but it's very difficult to figure out what that is.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def get_pod_info
+    roles = hash_with(&default_role)
+    Environment.find(params[:environment]).cluster_deploy_groups.each do |cluster_deploy_group|
+      cluster_deploy_group.cluster.client.get_pods(
+          namespace: cluster_deploy_group.namespace,
+          label_selector: "project=#{current_project.name_for_label}").each do |pod|
+        labels = pod.metadata.labels
+        release = roles[labels.role_id][:deploy_groups]\
+                      [cluster_deploy_group.id][:releases][labels.release_id]
```
henders: Why use the ```cluster_deploy_group.id``` for the deploy_group, shouldn't you be using the actual deploy-group id? ```cluster_deploy_group.deploy_group.id```

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
henders: After reading it some more, it's weird to use the 'default' machanics of the Hash class, when there aren't multiple possibilities for each role or deploy_group or release. It feels like you are abusing mechanic to write this logic.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+  include CurrentProject
+
+  def index
+    render json: format_pod_info(get_pod_info).to_json
+  end
+
+  private
+
+  def get_pod_info
+    roles = hash_with(&default_role)
```
henders: maybe ```roles = hash_with(&role_lookup)```?

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+        role[:deploy_groups][deploy_group_id] ||= build_deploy_group(deploy_group_id)
+        deploy_group = role[:deploy_groups][deploy_group_id]
+        deploy_group[:releases][labels.release_id] ||= build_release(labels.release_id, labels.role_id)
+        release = deploy_group[:releases][labels.release_id]
+        api_pod = Kubernetes::Api::Pod.new(pod)
+        release[:live_replicas] += 1 if api_pod.ready?
+      end
+    end
+    roles
+  end
```
sbrnunes: Just a suggestion: could this logic be refactored into a separate method as well, as you did for all the methods below? I mean, something in the likes of:
```
def pod_info
  Environment.find(params[:environment]).cluster_deploy_groups.each do |cluster_deploy_group|
    cluster_deploy_group.cluster.client.get_pods(
      namespace: cluster_deploy_group.namespace,
      label_selector: "project=#{current_project.name_for_label}").
      each(&:extract_pod_info)
  end
  roles
end

def extract_pod_info
   ... #all the logic responsible for parsing the data from the pod as it stands
end

def roles
  @roles ||= {}
end
```

Just to extract the parsing of the data from the actual call.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def pod_info
+    roles = {}
+    Environment.find(params[:environment]).cluster_deploy_groups.each do |cluster_deploy_group|
+      cluster_deploy_group.cluster.client.get_pods(
+          namespace: cluster_deploy_group.namespace,
+          label_selector: "project=#{current_project.name_for_label}").each do |pod|
+        labels = pod.metadata.labels
+        roles[labels.role_id] ||= build_role(labels.role_id)
+        role = roles[labels.role_id]
```
sbrnunes: Could we move this two lines into the `build_role` method and have the`build_role` method setting the role in the hash and returning it? It would simplify this logic.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+        deploy_group_id = cluster_deploy_group.deploy_group.id
+        role[:deploy_groups][deploy_group_id] ||= build_deploy_group(deploy_group_id)
+        deploy_group = role[:deploy_groups][deploy_group_id]
```
sbrnunes: Same as above, but this time regarding the `build_deploy_group`.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+        deploy_group[:releases][labels.release_id] ||= build_release(labels.release_id, labels.role_id)
+        release = deploy_group[:releases][labels.release_id]
```
sbrnunes: Same as above, but this time regarding the `build_release`.

plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -0,0 +1,26 @@
...
+      def name
+        @pod.metadata.name
+      end
+
+      private
+
+      def condition_ready?
+        @pod.status.conditions.present? &&
+            @pod.status.conditions
+                .select { |c| c['type'] == 'Ready' }
```
sbrnunes: Can we expect more that one condition with type 'Ready'? If not, we could use `find` instead of `select`, I think, return that object and then check the status property.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    {
+        id: release_id,
+        build: build_label(release_id),
+        target_replicas: target_replicas(release_id, role_id),
+        live_replicas: 0
+    }
+  end
+
+  def role_name(role_id)
+    Kubernetes::Role.find(role_id).name
```
sbrnunes: Isn't the role name one of the metadata labels we're sending to Kubernetes? Not sure, though...

sbrnunes:
After checking the comments, it's :+1: for me.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
msufa: It's actually `name.parameterize('-')` so we need to get the original in order for the user to see what they typed in when creating the role.

sbrnunes: Oh, good point. Was forgetting that.

plugins/kubernetes/app/models/kubernetes/api/pod.rb
msufa: We should have only one Ready condition: https://github.com/kubernetes/kubernetes/blob/master/docs/user-guide/pod-states.md
I'll update the code.

sbrnunes: Cool. Thanks for clarifying. :wink:

msufa:
@henders @sbrnunes, please have one last look at the PR. I really appreciate your feedback, the code looks much cleaner now, even Code Climate agrees :tada:

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    api_pod = Kubernetes::Api::Pod.new(pod)
+    release[:live_replicas] += 1 if api_pod.ready?
+  end
+
+  def role(roles, role_id)
+    roles[role_id] ||= build_role(role_id)
+  end
+
+  def build_role(role_id)
```
henders: I would probably combine role and build_role:
```
def role(roles, role_id)
  roles[role_id] ||= { 
    name: role_name(role_id), 
    deploy_groups: {} 
  }
end
```

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+        name: role_name(role_id),
+        deploy_groups: {}
+    }
+  end
+
+  def deploy_group(deploy_groups, deploy_group_id)
+    deploy_groups[deploy_group_id] ||= build_deploy_group(deploy_group_id)
+  end
+
+  def build_deploy_group(deploy_group_id)
```
henders: Same as above for combining.

plugins/kubernetes/app/controllers/kubernetes_dashboard_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+        name: deploy_group_name(deploy_group_id),
+        releases: {}
+    }
+  end
+
+  def release(releases, release_id, role_id)
+    releases[release_id] ||= build_release(release_id, role_id)
+  end
+
+  def build_release(release_id, role_id)
```
henders: Likewise.

henders:
Looks much cleaner! Nice!

Up to you whether you combine some of those small methods. :+1:

Issue #680, opened 2015-12-08
Faster stages#show


bquorning:
`StagesController#show` fetches an image from shields.io during the request. Why not just respond with a 302 redirect to the image, and let the browser suffer the latency penalty?

jonmoter:
It only does that if you're making a request to `stage_name.svg`, not the normal HTML request.

But I don't know why it's making the request server-side and proxying the response. @grosser, you implemented that in bce48989; what's the rationale?

grosser:
should be fine to redirect ...

grosser:
https://github.com/zendesk/samson/pull/682

PR #682, opened 2015-12-08
rediret to badge instead of caching it for speed

grosser:
fixes https://github.com/zendesk/samson/issues/680#issuecomment-162953044

@jonmoter @bquorning 

### Risks
 - None


bquorning:
:+1:

PR #683, opened 2015-12-09
Allow webhooks to trigger from any branch

mwerner:
When I originally started work on adding the ability to run a webhook from any branch, I wanted to make the `branch` column `nil` to denote any branch. However this required two things:

1. We have to change the database to remove the nil restriction
2. We have to provide a checkmark or special dropdown in the form to denote you want it to originate from any branch.

By going with `any` we are matching the same pattern as `source`.

If someone pushes their branch named `any`, it would be the correct behavior, since it's "any branch".

If someone wants to cause a webhook to only occur on their branch that happened to be named "any" for some reason, it could be a problem. Not sure if we should just bear that risk or maybe mitigate by putting something in the UI. Thoughts?

@zendesk/samson

### DIFF
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -6,7 +6,7 @@ class Webhook < ActiveRecord::Base
   belongs_to :stage
 
   def self.for_branch(branch)
-    where(branch: ['any', branch])
+    where(branch: ['', branch])
   end
 
   def self.for_source(service_type, service_name)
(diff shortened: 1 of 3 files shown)


sandlerr:
> If someone wants to cause a webhook to only occur on their branch that happened to be named "any" for some reason, it could be a problem.
I'm not sure if we really need to handle that case? You could name a branch `$(rm -rf /)` as well

steved:
Could we use an empty string as the constant? It sort of fits in nicely if you add placeholder text to the UI and just coerce the nil that comes in. Either way WFM, just want some sort of indication in the UI.

mwerner:
Yeah I think that's a good call, otherwise this would be a really undocumented feature. I'll go with an empty string

PR #685, opened 2015-12-09
Kubernetes Dashboard Page

sbrnunes:
PR containing the Kubernetes dashboard page. 

Whenever a deploying is in progress, if a User is in the dashboard page he will see the updates from the cluster coming through.

/cc @msufa, @henders, @jonmoter 

screenshot 2015-12-14 19 07 00

### DIFF
--- a/plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
+++ b/plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
@@ -1,15 +1,184 @@
-samson.controller('KubernetesDashboardCtrl', function($scope, $stateParams, SseFactory) {
-  $scope.project_id = $stateParams.project_id;
-  $scope.messages = [];
+samson.controller('KubernetesDashboardCtrl',
+  function($rootScope, $scope, $stateParams, $q, environmentsService, kubernetesService, SseFactory, notificationService) {
+    $scope.project_id = $stateParams.project_id;
+    $scope.environments = [];
+    $scope.messages = [];
+    $scope.dashboard_data = [];
+    $scope.environment = undefined;
+    $scope.loadingEnvironments = false;
+    $scope.loadingClusterState = false;
 
-  function init() {
-    // Subscribe to the SSE channel for K8s
-    SseFactory.on('k8s', function(msg) {
-      console.log("Got SSE msg: ", msg);
-      $scope.messages.push(JSON.stringify(msg));
-      _.defer(function() { $scope.$apply(); });
-    });
-  }
+    $scope.reloadDashboardData = function() {
+      loadClusterState();
+    };
 
-  init();
-});
+    $scope.environmentChanged = function(env) {
+      $scope.environment = env;
+      loadClusterState();
+    };
+
+    $scope.dashboardReady = function() {
+      return !$scope.loadingEnvironments && !$scope.loadingClusterState;
+    };
+
+    $scope.empty = function(collection) {
+      return _.isUndefinedOrEmpty(collection);
+    };
+
+    $scope.notEmpty = function(collection) {
+      return _.isNotUndefinedOrEmpty(collection);
+    };
+
+    function init() {
+      // Chaining the several promises (they should be resolved sequentially)
+      loadEnvironments()
+        .then(function() {
+            return loadClusterState();
+          }
+        )
+        .then(function() {
+            startListeningForUpdates()
+          }
+        );
+    }
+
+    init();
+
+    /*
+     Controller private functions.
+     */
+
+    function loadEnvironments() {
+      $scope.loadingEnvironments = true;
+
+      var deferred = $q.defer();
+      environmentsService.loadEnvironments($scope.project_id).then(
+        function(environments) {
+          $scope.loadingEnvironments = false;
+          $scope.environments = environments;
+
+          // Selects the production environment by default
+          var default_environment = _.findWhere(environments, {'is_production': true});
+          $scope.environment = _.isDefined(default_environment) ? default_environment : environments[0];
+
+          deferred.resolve();
+        },
+        function(result) {
+          $scope.loadingEnvironments = false;
+          result.messages.map(function(message) {
+            notificationService.error(message);
+          });
+          deferred.reject();
+        }
+      );
+
+      return deferred.promise;
+    }
+
+    function loadClusterState() {
+      $scope.loadingClusterState = true;
+      $scope.dashboard_data = [];
+
+      var deferred = $q.defer();
+      kubernetesService.loadDashboardData($scope.project_id, $scope.environment).then(
+        function(data) {
+          $scope.loadingClusterState = false;
+          if (_.isNotUndefinedOrEmpty(data)) {
+            $scope.dashboard_data = data;
+            deferred.resolve();
+          }
+          else {
+            deferred.reject();
+          }
+        },
+        function(result) {
+          $scope.loadingClusterState = false;
+          result.messages.map(function(message) {
+            notificationService.error(message);
+          });
+          deferred.reject();
+        }
+      );
+
+      return deferred.promise;
+    }
+
+    function startListeningForUpdates() {
+      SseFactory.on('k8s', function(msg) {
+        var role = findOrCreateRole(msg);
+        var deploy_group = findOrCreateDeployGroup(role, msg);
+        var release = findOrCreateRelease(deploy_group, msg);
+        updateReleaseState(release, msg);
+      });
+    }
+
+    function findOrCreateRole(msg) {
+      var role = findRole(msg.role);
+
+      if (_.isUndefined(role)) {
+        role = {
+          name: msg.role,
+          deploy_groups: []
+        };
+        $scope.dashboard_data.push(role);
+      }
+
+      return role;
+    }
+
+    function findOrCreateDeployGroup(role, msg) {
+      if (_.isDefined(role)) {
+        var deploy_group = findDeployGroup(role, msg.deploy_group);
+
+        if (_.isUndefined(deploy_group)) {
+          deploy_group = {
+            name: msg.deploy_group,
+            releases: []
+          };
+          role.deploy_groups.push(deploy_group);
+        }
+
+        return deploy_group;
+      }
+    }
+
+    function findOrCreateRelease(deploy_group, msg) {
+      if (_.isDefined(deploy_group)) {
+        var release = findRelease(deploy_group, msg.release);
+
+        if (_.isUndefined(release)) {
+          release = {
+            id: msg.release,
+            build: msg.build,
+            live_replicas: msg.live_replicas,
+            target_replicas: msg.target_replicas
+          };
+          deploy_group.releases.push(release);
+        }
+        return release;
+      }
+    }
+
+    function updateReleaseState(release, msg) {
+      release.live_replicas = msg.live_replicas;
+      $scope.$apply();
+    }
+
+    function findRole(role_id) {
+      return _.find($scope.dashboard_data, function(role) {
+        return role.name == role_id;
+      });
+    }
+
+    function findDeployGroup(role, deploy_group_name) {
+      return _.find(role.deploy_groups, function(deploy_group) {
+        return deploy_group.name == deploy_group_name;
+      });
+    }
+
+    function findRelease(deploy_group, release_id) {
+      return _.find(deploy_group.releases, function(release) {
+        return release.id == release_id;
+      });
+    }
+  });
--- a/plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
+++ b/plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
@@ -78,7 +78,6 @@ samson.controller('KubernetesDashboardCtrl',
       var deferred = $q.defer();
       kubernetesService.loadDashboardData($scope.project_id, $scope.environment).then(
         function(data) {
-          $scope.loadingClusterState = false;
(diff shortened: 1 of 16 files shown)


plugins/kubernetes/app/assets/javascripts/kubernetes/directives/kubernetes_deploy_progress_directive.js
```diff
@@ -0,0 +1,64 @@
...
+        return release.id == current.id;
+      };
+
+      $scope.progress = function(release) {
+        return ((release.live_replicas * 100) / release.target_replicas) + '%';
+      };
+
+      $scope.deployCompleted = function() {
+        return _.every($scope.deployGroup.releases, function(release) {
+          return $scope.currentRelease(release) ? release.live_replicas == release.target_replicas : release.live_replicas == 0;
```
msufa: You could just call `targetStateReached` instead of comparing replicas.

plugins/kubernetes/app/assets/javascripts/kubernetes/directives/kubernetes_deploy_progress_directive.js
```diff
@@ -0,0 +1,64 @@
...
+        // TODO: figure out if this will ever be used
+        return false;
+      };
+
+      $scope.targetStateReached = function(release) {
+        return release.live_replicas == release.target_replicas;
+      };
+
+      $scope.$on('clusterUpdateReceived', function(e, message) {
+        if ($scope.role.name == message.role && $scope.deployGroup.name == message.deploy_group) {
```
msufa: Since we now set the role_id label for each pod, it would be safer to compare the ids. This would require updates in the deploy watcher though.

plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
```diff
@@ -18,18 +18,48 @@ select {
...
 
 /*
-    Kubernetes Releases
+  Kubernetes screen
 */
-.kubernetes-releases span.label-info {
+
+.kubernetes-section {
+  margin-top: 0 !important;
+  /*border-top-left-radius: 0;*/
```
msufa: Commented out code.

plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
```diff
@@ -18,18 +18,48 @@ select {
...
+}
+
+.kubernetes-section .nav-tabs {
+  border-bottom: 1px solid #f5f5f5;
+  padding: 0 10px 0 15px;
+  background-color: #fcfcfc;
+}
+
+.kubernetes-section .kubernetes-content {
+  /*padding: 10px;*/
```
msufa: Same here.

plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
```diff
@@ -158,3 +196,223 @@ select {
...
+
+.kubernetes-dashboard .no-data-message {
+  padding: 50px 0;
+  text-align: center;
+}
+
+.kubernetes-dashboard .deploy-groups-container {
+  display: inline-flex;
+  flex-flow: row wrap;
+  /* align-items: baseline; */
```
msufa: And here.

plugins/kubernetes/app/assets/javascripts/kubernetes/directives/kubernetes_deploy_progress_directive.js
sbrnunes: Oh, you're right. I created that method afterwards. Going to refactor it, thanks for the heads up.

plugins/kubernetes/app/assets/javascripts/kubernetes/directives/kubernetes_deploy_progress_directive.js
sbrnunes: Definitely. I'd like to change all this checks to use only ids. We need to make those changes on the watchers first. I'll probably do it later, on a separate PR.

plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
sbrnunes: Thanks. Forgot to remove that.

plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
sbrnunes: And that.

plugins/kubernetes/app/assets/stylesheets/kubernetes/application.css
sbrnunes: And also that. :-D

sbrnunes:
@msufa, @henders, @jonmoter the dashboard is fully functional now. Whenever a deploying is in progress, if a User is on the dashboard he will see the updates coming through, from the cluster.

Previous release is still not being updated, though. That will be possible as soon as the integration with the Deployments API is finished.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
```diff
@@ -1,15 +1,184 @@
...
+
+    $scope.dashboardReady = function() {
+      return !$scope.loadingEnvironments && !$scope.loadingClusterState;
+    };
+
+    $scope.empty = function(collection) {
+      return _.isUndefinedOrEmpty(collection);
+    };
+
+    $scope.notEmpty = function(collection) {
```
henders: Seems like a needless wrapper when you can just use !empty()

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
```diff
@@ -1,15 +1,184 @@
...
+  function($rootScope, $scope, $stateParams, $q, environmentsService, kubernetesService, SseFactory, notificationService) {
+    $scope.project_id = $stateParams.project_id;
+    $scope.environments = [];
+    $scope.messages = [];
+    $scope.dashboard_data = [];
+    $scope.environment = undefined;
+    $scope.loadingEnvironments = false;
+    $scope.loadingClusterState = false;
+
+    $scope.reloadDashboardData = function() {
```
henders: Why not make ```loadClusterState``` part of the scope rather then wrapping?

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
```diff
@@ -1,15 +1,184 @@
...
+
+    function init() {
+      // Chaining the several promises (they should be resolved sequentially)
+      loadEnvironments()
+        .then(function() {
+            return loadClusterState();
+          }
+        )
+        .then(function() {
+            startListeningForUpdates()
```
henders: Is it ok that it starts listening before the ```loadClusterState``` is finished?

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
```diff
@@ -1,15 +1,184 @@
...
+    }
+
+    function loadClusterState() {
+      $scope.loadingClusterState = true;
+      $scope.dashboard_data = [];
+
+      var deferred = $q.defer();
+      kubernetesService.loadDashboardData($scope.project_id, $scope.environment).then(
+        function(data) {
+          $scope.loadingClusterState = false;
```
henders: Maybe move this line to the end of this function?

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
```diff
@@ -1,15 +1,184 @@
...
+          };
+          role.deploy_groups.push(deploy_group);
+        }
+
+        return deploy_group;
+      }
+    }
+
+    function findOrCreateRelease(deploy_group, msg) {
+      if (_.isDefined(deploy_group)) {
```
henders: Probably not needed since you're creating the deploy_group if it doesn't exist. Plus it'll throw an exception anyway if the deploy_group is nil later @ https://github.com/zendesk/samson/pull/685/files#diff-1fe6543a8989338e21335956e8dd9daeR163 when it tries to deference an undefined release.

plugins/kubernetes/app/assets/javascripts/kubernetes/directives/kubernetes_deploy_progress_directive.js
```diff
@@ -0,0 +1,62 @@
...
+    restrict: 'E',
+    templateUrl: 'kubernetes/deploy_progress_widget.tmpl.html',
+    scope: {
+      role: '=',
+      deployGroup: '='
+    },
+    controller: function($scope) {
+
+      $scope.currentRelease = function(release) {
+        var current = $scope.lastRelease();
```
henders: What happens if this is the first release ever for the project? Is this function called?

henders: Ah, nvm, I see it'll return the ongoing release. so all good.

plugins/kubernetes/app/assets/javascripts/kubernetes/filters/append_filter.js
```diff
@@ -0,0 +1,7 @@
+samson.filter('append', function(){
+
+  return function(input, str) {
+    return _.isNotUndefinedOrEmpty(input) ? input + str : input;
```
henders: You might as well return an explicit ```''``` if ```input``` is empty or undefined.

plugins/kubernetes/app/assets/javascripts/kubernetes/filters/default_filter.js
```diff
@@ -0,0 +1,7 @@
+samson.filter('default', function() {
+
+  return function(input, str) {
+    return _.isUndefinedOrEmpty(input) ? str : input;
```
henders: Same as above

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
sbrnunes: I tend to use this instead of the negate operator. IMHO, it helps readability, specially when we're mixing Javascript code  with HTML: `<div ng-if="notEmpty(dashboard_data)" ...>` vs `<div ng-if="!empty(dashboard_data)" ...>`

I know it's 3 extra lines of code, but if you guys think there's no gain, I can take it out.

henders:
:+1: after some small comments.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
sbrnunes: Hummm, I don't remember why did I create this extra method. I think I was also re-registering the callback for the SSE events in here before realising that was not needed. 

I'm going to simplify this.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
sbrnunes: Okidokie.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
sbrnunes: Good catch.

This is old code, added before I started creating the data myself when it didn't exist.

plugins/kubernetes/app/assets/javascripts/kubernetes/filters/append_filter.js
sbrnunes: Hum, I don't want to change the input when the input is undefined. The idea behind this filter is just to append a given input value with the provided string suffix when the input is defined, otherwise, return the original input. If input is undefined, it should be ignored, otherwise, we will be changing the original input, which can have impact on another filter when when chaining multiple filters.

plugins/kubernetes/app/assets/javascripts/kubernetes/filters/default_filter.js
sbrnunes: Same as above. If I return `''` instead of the original value when `undefined`, the above filter will have to deal with the condition `defined but empty` as well, because they've been chained in the HTML. I know that's what you suggested above, but I think although they're being used together, each filter should be agnostic of the other, meaning, they should be applied when possible but return the original value otherwise. This way, they can even be reused anywhere.

What do you think?

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
henders: I'm not too concerned, so I'm good with leaving it.

plugins/kubernetes/app/assets/javascripts/kubernetes/filters/append_filter.js
henders: Fair point.

plugins/kubernetes/app/assets/javascripts/kubernetes/filters/default_filter.js
henders: Yup, fair point.

PR #687, opened 2015-12-11
Update readme to include info about requesting access rights

msufa:
Add info about requesting additional access rights from the profile page.

/cc @zendesk/samson

### References
 - Jira link: 

### Risks
 - None


henders:
:+1:

PR #689, opened 2015-12-15
Kubernetes new deployment api

henders:
Switching over from the Replication Controller deployment mechanism to the new Deployment API. 

### Tasks
 - [x] Use the new deployment API

/cc @msufa @sbrnunes @jonmoter 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
+++ b/plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
@@ -1,0 +1,61 @@
+module Kubernetes
+  module DeployYaml
+    def deployment_hash
+      @deployment_yaml ||=
+        if replication_controller_doc.present?
+          JSON.parse(replication_controller_doc).with_indifferent_access
+        else
+          template.spec.replicas = replica_target
+          template.metadata.namespace = deploy_group.kubernetes_namespace
+          add_labels
+          update_docker_image
+          set_resource_usage
+          Rails.logger.info "Created K8S hash: #{template}"
+          template
+        end
+    end
+
+    private
+
+    def template
+      @template ||= begin
+        yaml = YAML.load_stream(raw_template, kubernetes_role.config_file).detect do |doc|
+          doc['kind'] == 'ReplicationController' || doc['kind'] == 'Deployment'
+        end
+        RecursiveOpenStruct.new(yaml, :recurse_over_arrays => true)
+      end
+    end
+
+    def raw_template
+      @raw_template ||= build.file_from_repo(kubernetes_role.config_file)
+    end
+
+    def add_labels
+      labels.each do |key, value|
+        template.metadata.labels[key] = value
+        template.spec.selector[key] = value
+        template.spec.template.metadata.labels[key] = value
+      end
+    end
+
+    def set_resource_usage
+      container.resources = {
+        limits: { cpu: kubernetes_role.cpu, memory: kubernetes_role.ram_with_units }
+      }
+    end
+
+    def labels
+      kubernetes_release.pod_labels.merge(role: kubernetes_role.label_name, role_id: kubernetes_role.id.to_s)
+    end
+
+    def update_docker_image
+      docker_path = build.docker_repo_digest || "#{build.project.docker_repo}:#{build.docker_ref}"
+      # Assume first container is one we want to update docker image in
+      container.image = docker_path
+    end
+
+    def container
+      template.spec.template.spec.containers.first
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -118,7 +126,7 @@ def ssl_options
           client_cert: user.client_cert,
           client_key:  user.client_key,
           ca_file:     cluster.ca_filepath,
-          verify_ssl:  OpenSSL::SSL::VERIFY_PEER
+          verify_ssl:  OpenSSL::SSL::VERIFY_NONE
```
msufa: That's a bit dangerous for prod environment, is there no other workaround?

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -15,12 +15,11 @@ def initialize(release)
...
 
     def watch
-      @release.release_docs.each do |release_doc|
-        subscribe("#{release_doc.replication_controller_name}", :handle_update)
-      end
+      subscribe("#{@release.build.project_name}", :handle_update)
     end
 
     def handle_update(topic, data)
+      info "Got Release Event: #{topic}"
```
msufa: I wonder if `info` is the correct log level. There will be a lot of these events.

plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
```diff
@@ -0,0 +1,61 @@
+module Kubernetes
+  module DeployYaml
+    def deployment_hash
+      @deployment_yaml ||=
+        if replication_controller_doc.present?
+          JSON.parse(replication_controller_doc).with_indifferent_access
```
sbrnunes: In which case can we end up in this situation? When scaling-up / scaling-down an existing release?

plugins/kubernetes/app/models/kubernetes/role_config_file.rb
```diff
@@ -44,98 +47,28 @@ def as_hash(type)
     #
     # INNER CLASSES
     #
+    class Deployment < RecursiveOpenStruct
```
sbrnunes: This `RecursiveOpenStruct` is so good. I wish I knew it before.

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
henders: Ah, didnt mean to commit that! good catch

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
henders: Yup, that was a temporary log while debugging the watcher. 'debug' would be sufficient if leaving this log in.

plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
henders: Yea, when redeploying an old release was the original use case, I just copied this code. Scaling it will probably need to regenerate the YAML rather then using what was there.

plugins/kubernetes/app/models/kubernetes/role_config_file.rb
henders: Yea, I'm loving it. :)

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -107,6 +111,10 @@ def client
         Kubeclient::Client.new(cluster.url, api_version, ssl_options: ssl_options)
       end
 
+      def ext_client
+        Kubeclient::Client.new(cluster.server + '/apis', 'extensions/v1beta1', ssl_options: ssl_options)
+      end
```
jonmoter: FYI, it wasn't immediately clear to me what "ext" was short for.  I guessed "External", before I saw that it involved extensions.

henders: Good point, renamed all those to 'extension_client'

henders:
@sbrnunes @msufa @jonmoter : addressed all concerns?

henders:
This is also only the baseline for @sbrnunes 's PR #698

sbrnunes:
Yes, it's a :+1: for me.

henders:
Merging so we can get the next PR merged easier.

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -21,6 +21,10 @@ def client_for(context_name)
       contexts[context_name].try(:client)
     end
 
+    def extension_client_for_client_for(context_name)
```
msufa: Kudos on the method name :) Since it's not used anywhere I presume this is some rubbish code?

henders: Ugh, thanks for catching that… that was me fat fingering a function rename. It is being used! :/

PR #690, opened 2015-12-16
locks should unlock after 1 hour by default

kintner:
@jonmoter @msufa

When I lock a stage it's normally to prevent automated deploys while I test a feature branch. I occasionally forget to unlock it after I'm done testing and the stage will remain locked until somebody is bold enough to unlock it. This isn't great since it blocks the normal automated deploy workflow. 

This adds an "unlock in" option when creating a lock and I choose an opinionated default that most locks don't need to last for more then 1 hour. 

production_-_example-project

production_-_example-project

Currently we suggest running the `deploys:stop_expired_deploys` task via cron. Should I add the `locks: remove_expired_locks` to the example cron file as well?


kintner:
Ugh..there doesn't appear to be a cross DB way to do time addition in SQL. The actual number of locks is probably low enough you could to the check in ruby land. Any suggestions?

msufa:
cc @zendesk/samson 

@kintner, your tests are failing for Postgres & SQLite DBs.

steved:
Doing the math up front using a delete_at field would be easier, then the delete can be done in directly SQL later.

Use Rails' [time stubbing](http://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html) intead of Timecop.

I'd like to avoid adding rake cron jobs, I'd rather see a new thread that just sleeps, but that can be up for debate.

app/models/lock.rb
```diff
@@ -19,13 +19,19 @@ def global?
...
+
+    "Locked by #{user.try(:name) || 'Unknown user'} #{time_ago_in_words(created_at)} ago #{unlock}"
   end
 
   def reason
     description.blank? ? "Description not given." : description
   end
 
+  def self.remove_expired_locks
+    Lock.where("created_at IS NOT NULL AND delete_in IS NOT NULL AND created_at + delete_in < ?", Time.now).find_each(&:soft_delete)
```
steved: Why would created_at be null?

kintner:
@steved refactored to use a pre calculated `delete_at` field.

app/models/lock.rb
```diff
@@ -1,6 +1,8 @@
 class Lock < ActiveRecord::Base
   include ActionView::Helpers::DateHelper
 
+  attr :delete_in
```
steved: Is this used? (unrelated, but I forgot attr existed)

app/views/locks/_lock.html.erb
```diff
@@ -4,5 +4,12 @@
   <% else %>
     <%= lock_icon %> Deployments to <%= lock.global? ? "ALL STAGES" : "this stage" %> were locked
   <% end %>
-  <%= lock.reason %> by <%= lock.user.name %> <%= time_ago_in_words(lock.created_at) %> ago.
+  <%= lock.reason %> by <%= lock.user.name %> <%= time_ago_in_words(lock.created_at) %>
```
steved: Can this be lock.summary?

app/models/lock.rb
```diff
@@ -19,13 +21,23 @@ def global?
   end
 
   def summary
-    "Locked by #{user.try(:name) || 'Unknown user'} #{time_ago_in_words(created_at)} ago"
+    unlock = "and will unlock in #{time_ago_in_words(delete_at)}" if delete_at
+
+    "Locked by #{user.try(:name) || 'Unknown user'} #{time_ago_in_words(created_at)} ago #{unlock}"
```
steved: probably a tiny bit nicer if the space after `ago` is in the unlock variable

kintner:
@steved I've added a timer task to delete expired locks.

app/views/locks/_lock.html.erb
kintner: Not really. It's a different sentence then `#summary`

app/models/lock.rb
```diff
@@ -1,6 +1,8 @@
 class Lock < ActiveRecord::Base
   include ActionView::Helpers::DateHelper
 
+  attr_reader :delete_in
```
steved: Does AR complain without this? There's no instance variable to access...

Gemfile
```diff
@@ -29,6 +29,8 @@ gem 'active_model_serializers', '~> 0.8.0'
 gem 'sawyer', '~> 0.5'
 gem 'sse-rails-engine', '~> 1.4'
 
+gem 'concurrent-ruby', require: 'concurrent'
```
steved: This can probably live in `no_preload`

app/views/locks/_lock.html.erb
steved: Fair, but there's definitely some DRY opportunity

lib/samson/tasks.rb
```diff
@@ -0,0 +1,5 @@
+module Samson::Tasks
+  def self.start
+    LockCleaner.start
```
steved: Any reason for the indirection? Should this become a real registry class and keep instances?

lib/samson/tasks/lock_cleaner.rb
```diff
@@ -0,0 +1,26 @@
...
+      @task ||= Concurrent::TimerTask.new(run_now: true, execution_interval: 60, timeout_interval: 10) do
+        Lock.remove_expired_locks
+      end.with_observer(self)
+    end
+
+    # called by Concurrent::TimerTask
+    def update(time, result, ex)
+      if ex
+        Rails.logger.error "(#{time}) Samson::Tasks::LockCleaner failed with error #{ex}\n"
+        Rails.logger.error ex.backtrace.join("\n")
```
steved: Airbrake instead?

lib/samson/tasks/lock_cleaner.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def start
+      task.tap(&:execute)
+    end
+
+    def task
+      @task ||= Concurrent::TimerTask.new(run_now: true, execution_interval: 60, timeout_interval: 10) do
+        Lock.remove_expired_locks
+      end.with_observer(self)
```
steved: with_observer will be called after the first execution which might hide errors

app/models/lock.rb
```diff
@@ -19,13 +21,23 @@ def global?
...
+
+    "Locked by #{user.try(:name) || 'Unknown user'} #{time_ago_in_words(created_at)} ago#{unlock}"
   end
 
   def reason
     description.blank? ? "Description not given." : description
   end
 
+  def self.remove_expired_locks
+    Lock.where("delete_at IS NOT NULL and delete_at < CURRENT_TIMESTAMP").find_each(&:soft_delete)
```
steved: Might as well soft_delete! since you're checking exceptions but not the result of this method

test/controllers/locks_controller_test.rb
```diff
@@ -118,6 +120,7 @@
         stage.warning?.must_equal(false)
         stage.locked?.must_equal(true)
         stage.lock.description.must_equal 'DESC'
+        stage.lock.delete_at.to_time.must_equal(Time.now + 3600)
```
steved: Why to_time?

test/models/lock_test.rb
```diff
@@ -49,4 +49,55 @@
...
+    it "leaves unexpired locks alone" do
+      refute_empty Lock.where("delete_at > ?", Time.now)
+    end
+
+    it "leaves indefinite locks alone" do
+      refute_empty Lock.where("delete_at is null")
+    end
+  end
+
+  def create_some_locks
```
steved: inline this since you only use it once now

app/models/lock.rb
kintner: Yeah...I'm using it in the view: `f.select :delete_in`

lib/samson/tasks/lock_cleaner.rb
kintner: Why do you say that? The `TimerTask` doesn't start running until  `execute` is called which is after the observer is added. `run_now: true` schedules the task to run immediately after `execute` instead of waiting until the first `execution_interval` occurs.

app/models/lock.rb
kintner: let me see what I can do

app/views/locks/_lock.html.erb
kintner: let me see what I can do

lib/samson/tasks/lock_cleaner.rb
steved: Yeah, my bad, the docs say "instantiation" but I should've tried it
https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/timer_task.rb#L171-L173

app/helpers/locks_helper.rb
```diff
@@ -0,0 +1,12 @@
+module LocksHelper
+  def delete_lock_options
+    offsets = [
```
steved: offsets is unused

test/lib/samson/tasks/lock_cleaner_test.rb
```diff
@@ -0,0 +1,43 @@
+require_relative '../../../test_helper'
+
+describe Samson::Tasks::LockCleaner do
+  subject { Samson::Tasks::LockCleaner.new }
+
+  describe "#start" do
+    before { Concurrent::TimerTask.any_instance.expects(:execute).once }
```
steved: This can be `subject.task.expects` and is more part of the test than setup

test/models/lock_test.rb
```diff
@@ -49,4 +49,49 @@
...
+      Lock.create!(user: users(:deployer), stage: stages(:test_production), created_at: expired, delete_in: 3600)
+      Lock.create!(user: users(:deployer), stage: stages(:test_staging), delete_in: 3600)
+      Lock.create!(user: users(:deployer), stage: stages(:test_production), delete_in: 3600)
+      Lock.create!(user: users(:deployer), stage: stages(:test_production_pod))
+
+      Lock.remove_expired_locks
+    end
+
+    it "removes expired locks" do
+      assert_empty Lock.where("delete_at < ?", Time.now)
```
steved: You're mixing and matching `must` and `assert`, I'd prefer `must` but just be consistent

test/controllers/locks_controller_test.rb
```diff
@@ -108,8 +108,10 @@
     end
 
     describe 'POST to #create' do
+      before { travel_to Time.now }
+      after { travel_back }
```
steved: add another empty line after

lib/samson/tasks/lock_cleaner.rb
```diff
@@ -0,0 +1,29 @@
+require 'concurrent'
+
+module Samson::Tasks
+  class LockCleaner
+
```
steved: remove this line

kintner:
@steved :haircut:

app/views/locks/_lock.html.erb
```diff
@@ -4,5 +4,12 @@
   <% else %>
     <%= lock_icon %> Deployments to <%= lock.global? ? "ALL STAGES" : "this stage" %> were locked
   <% end %>
-  <%= lock.reason %> by <%= lock.user.name %> <%= time_ago_in_words(lock.created_at) %> ago.
+  <%= lock.reason %> by <%= lock.locked_by %> <%= time_ago_in_words(lock.created_at) %>
+
+  <% if lock.delete_at %>
```
steved: Can this block be `ago<%= lock.unlock_at %>.`?

steved:
:+1:

PR #691, opened 2015-12-16
Use SSL cert store instead of temp file

msufa:
Expired temp files caused SSLError exceptions when trying to connect to k8s cluster API.

### References
 - Jira link: 

### Risks
 - None


jonmoter:
:+1:

PR #694, opened 2015-12-18
Simplify and fix deploy output when restarting

steved:
Pending non-production jobs should never be a thing, so I removed the approve UI for that.

@zendesk/samson


app/helpers/deploys_helper.rb
```diff
@@ -1,8 +1,26 @@
...
+      output << Samson::Hooks.render_views(:deploy_view, self, deploy: @deploy, project: @project)
+
+      if deploy_queued?
+        output_hidden = true
+        output << render('queued')
+      elsif @deploy.waiting_for_buddy?
+        output_hidden = true
+        output << render('buddy_check', deploy: @deploy)
+      end
+    elsif @deploy.active?
```
henders: Was this put in just to allow current tests to pass?

steved: Job execution is disabled while the server is restarting but there are still jobs pending. This is actually the fix, to properly show a queued message when that happens.

steved:
bump

PR #697, opened 2015-12-18
Revert "timeout"

steved:
This reverts commit 3dfbcdd316ea3222aeaaab8ab7a96b24f74e9430.


jonmoter:
It was 5 before the referenced commit, why the change to 15?

steved:
I'm trying to figure out if there's actually a failing test or 5s is just too small. I don't want to be re-running travis.

jonmoter:
Okay.  :+1:

PR #698, opened 2015-12-21
Kubernetes : new Deployment API and adapted DeployWatcher

sbrnunes:
PR created in top of #685 and #689, containing the changes to the DeployWatcher to track an ongoing deployment (i.e., using the Deployments API).

/cc @msufa, @henders, @jonmoter

### Tasks
 - [x] DeployWatcher should track events for any Release regarding the given Project
 - [x] DeployWatcher should synchronise the DB with the current state in the cluster before it starts watching for events
 - [x] When creating a new Release, if there's a ongoing Deployment, the current watcher must be stopped before a watcher is created (to avoid concurrent updates)
 - [ ] Tests 

####  A simple Release: 
https://drive.google.com/open?id=0B8-_opV53zzXVmhuSnIyb0ppdWs
#### Creating a new Release when another is ongoing: 
https://drive.google.com/open?id=0B8-_opV53zzXR0RXUVktX1kyWG8


sbrnunes:
@henders, @msufa : before reviewing this PR, #685 should be merged into master to reduce the changes on this one. I branched from #685 because I needed the UI functionality.

Gemfile.lock
```diff
@@ -199,8 +199,7 @@ GEM
     hashie (3.4.3)
     hike (1.2.3)
     hitimes (1.2.3)
-    hitimes (1.2.3-java)
-    http (0.9.8)
+    http (1.0.1)
```
msufa: We need to fix the `http` version to `0.9.8`. The current `http` versions `>= 1.0.0` don't support `Celluloid::IO` which we need for non blocking IO in our watchers, see: https://github.com/httprb/http/wiki/Parallel-requests-with-Celluloid::IO

plugins/kubernetes/app/decorators/project_decorator.rb
```diff
@@ -54,6 +54,10 @@ def name_for_label
     name.parameterize('-')
   end
 
+  def deploy_watcher_actor_name
```
msufa: Wouldn't it be cleaner if the deploy watcher handled generating it's own name (like the cluster watchers)? This way the project code needs to be kinda aware of the existence of watchers.

plugins/kubernetes/app/models/concerns/kubernetes/api/deleted_pod.rb
```diff
@@ -0,0 +1,9 @@
+module Kubernetes
+  module Api
+    module DeletedPod
```
msufa: Some documentation for this class would be useful. It's not obvious what it's for at first glance.

plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
```diff
@@ -30,10 +34,43 @@ def raw_template
       @raw_template ||= build.file_from_repo(kubernetes_role.config_file)
     end
 
-    def add_labels
-      labels.each do |key, value|
+    # This key replaces the default kubernetes key: 'deployment.kubernetes.io/podTemplateHash'
+    # This label is used bu kubernetes to identify a RC and corresponding Pods
```
msufa: `used by`

plugins/kubernetes/app/models/concerns/kubernetes/has_status.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    STATUSES.each do |s|
+      define_method("#{s}?") { status == s }
+    end
+  end
+
+  def status=(new_status)
+    super new_status.to_s
+  end
+end
```
msufa: Shouldn't this module also host the `fail!` method used to fail release docs?

plugins/kubernetes/app/models/kuber_deploy_service.rb
```diff
@@ -60,4 +60,11 @@ def log(msg, extra_info = {})
 
     Kubernetes::Util.log msg, extra_info
   end
+
+  # Starts a new deploy watcher, forcing it to get in synch with the cluster
+  def start_watching_cluster
+    watcher = Celluloid::Actor[project.deploy_watcher_actor_name]
+    watcher.terminate if watcher and watcher.alive?
+    Celluloid::Actor[project.deploy_watcher_actor_name] = Watchers::DeployWatcher.new(project)
```
msufa: Shouldn't this watcher be supervised?

plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -5,17 +5,39 @@ def initialize(api_pod)
...
+        @pod.metadata.labels.release_id.to_i
+      end
+
+      def deploy_group_id
+        @pod.metadata.labels.deploy_group_id.to_i
+      end
+
+      def role_id
+        @pod.metadata.labels.role_id.to_i
+      end
```
msufa: We could generate all 4 methods to DRY things up.

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -14,9 +15,9 @@ class Release < ActiveRecord::Base
     validates :status, inclusion: STATUSES
     validate :docker_image_in_registry?, on: :create
 
-    STATUSES.each do |s|
-      define_method("#{s}?") { status == s }
-    end
+    scope :not_dead, -> { where.not(status: :dead) }
+    scope :excluding, ->(ids) { where.not(id: ids) }
+    scope :with_not_dead_release_docs, -> { joins(:release_docs).where.not(Kubernetes::ReleaseDoc.table_name => { status: :dead }) }
```
msufa: I'm not yet sure where this is used, but as a precaution, we also have a `failed` status, does it qualify as `dead` as well in this context?

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -74,6 +69,21 @@ def create_release_docs(params)
...
+    def update_status(release_doc)
+      case
+      when release_docs.all?(&:live?) then self.status = :live
+      when release_docs.all?(&:dead?) then self.status = :dead
+      when release_doc.spinning_up? then self.status = :spinning_up
+      when release_doc.spinning_down? then self.status = :spinning_down
+      when release_docs.any?(&:dead?) then self.status = :spinning_down
+      end
+      save!
+    end
```
msufa: What about `failed` release docs, should we handle them here in any way?

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -36,36 +33,44 @@ def service
...
+      when live_pods > replicas_live && !failed? then self.status = :spinning_up
+      when live_pods < replicas_live then self.status = :spinning_down
       end
       save!
     end
 
+    def update_replica_count(new_count)
+      self.replicas_live = new_count
+      save!
+    end
```
msufa: Why not just call `update_attribute`?

plugins/kubernetes/app/models/kubernetes/role_config_file.rb
```diff
@@ -36,7 +29,10 @@ def parse_replication_controller
 
     def parse_service
       service_hash = as_hash('Service')
-      @service = Service.new(service_hash) unless service_hash.nil?
+      @service = Service.new(service_hash, :recurse_over_arrays => true) unless service_hash.nil?
```
msufa: Adding `recurse_over_arrays` here is obsolete as the `Service` initializer handles that.

plugins/kubernetes/app/models/watchers/base_cluster_watcher.rb
```diff
@@ -8,8 +8,9 @@ class BaseClusterWatcher
 
     finalizer :stop_watching
 
-    def initialize(watch_stream)
-      @watch_stream = watch_stream
+    def initialize(args)
+      @cluster = args[:cluster]
+      @watch_stream = args[:watch_stream]
```
msufa: I'm a bit confused here. The initializer expects a hash, while the `supervise` call below passes in just the cluster. How does that even work? :)

msufa: Ok, I see now. Derived `initialize` turns the cluster into a hash with the cluster & stream and passes it to superclass. Maybe composition in this case would be better than inheritance, we seem to be stretching thing a bit.

plugins/kubernetes/app/models/watchers/cluster_pod_watcher.rb
```diff
@@ -2,16 +2,20 @@ module Watchers
...
+      super({
+        cluster: cluster,
+        watch_stream: cluster.client.watch_pods
+      })
     end
 
     private
 
     def handle_notice(notice)
+      debug notice.to_s
```
msufa: The notice is already logged in the base watcher.

plugins/kubernetes/app/decorators/project_decorator.rb
sbrnunes: Humm, I did this because I needed the same Actor name on the KuberDeployService, but you're right, I could have created a class method on the DeployWatcher that, given a project, returns the Actor name.

I'll change that. It was dumb to put this in the project.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -7,113 +6,211 @@ class DeployWatcher
...
+    def initialize(project)
+      @project = project
       @pod_timer = after(ENV.fetch('KUBERNETES_POD_TIMEOUT', 10.minutes).to_i) { pod_timeout }
-      info "Start watching K8s deploy: #{@release}"
+      start_watching
+    end
+
+    def start_watching
+      sync_with_cluster
+      info 'Start watching for deployments'
```
msufa: Adding the project name to the trace would give us more valuable info.

plugins/kubernetes/app/models/concerns/kubernetes/api/deleted_pod.rb
sbrnunes: Fair enough. :)

I actually would like to get @henders opinion on this. I'm dynamically extending the Pod objects with this module when we get a deleted Pod in the DeployWatcher. Is this an acceptable pattern in Ruby?

plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
sbrnunes: Good eyes man... :smile:

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -7,113 +6,211 @@ class DeployWatcher
...
-                        handle_pod_update(release_doc, data)
-                      end
-      if release_updated
-        send_event(role: release_doc.kubernetes_role.name,
-                   deploy_group: release_doc.deploy_group.name,
-                   target_replicas: release_doc.replica_target,
-                   live_replicas: release_doc.replicas_live,
-                   failed: release_doc.failed?)
-        end_deploy if deploy_finished?
+      subscribe("pod-events-project-#{@project.id}", :handle_event)
```
msufa: It would be good to have the topic constructed in a single place. Right now it's built both here and in the pod watcher.

plugins/kubernetes/app/models/concerns/kubernetes/has_status.rb
sbrnunes: Good point! I'll move it there. It didn't occur to me, but makes sense, specially because we use the same method for Release and ReleaseDoc, I think.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -7,113 +6,211 @@ class DeployWatcher
...
+    end
+
+    def handle_event(topic, message)
+      notice = message.notice
+      info "Got message on topic: #{topic}. Type of message: #{notice.object.kind}"
```
msufa: `info` is a bit to broad, we will flood the logs. I'd suggest `debug`.

plugins/kubernetes/app/models/kuber_deploy_service.rb
sbrnunes: I gave up on my intention of having this DeployWatchers always active, because it would mean a live and running instance of the watcher per project. In my opinion, that's a big overhead for Samson, but that's something I wanted to discuss with you and @henders.

If we decide to go along that path then, yes, I think this could become a supervised Actor that we just need to restart each time a new Release is created (to force it to get in-sync with the Cluster).

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -7,113 +6,211 @@ class DeployWatcher
...
+
+      case
+      when notice.object.kind == 'Event' then handle_event_update(message)
+      when notice.object.kind == 'Pod' then handle_pod_event(message)
```
msufa: Maybe rename the handler to `handle_pod_update`. We have "event" in both lines, it might confuse people.

plugins/kubernetes/app/models/kubernetes/api/pod.rb
sbrnunes: Good point.

Personally, I tend to avoid using metaprogramming for this things, because you loose the IDE assistance when you do it, but it appears to be a common approach in Ruby, so... I'll do it. ;)

plugins/kubernetes/app/models/kubernetes/release.rb
sbrnunes: Nop. `dead` represents a old release / release_doc. It's dead, gone, we don't care about it anymore. :)

This is used to reconcile old releases with the Cluster: if there's a new release being deployed, we're fetching and marking as `dead` each previous Release that:
- is not dead
- does not have a pod being shutdown in the cluster

plugins/kubernetes/app/models/kubernetes/release.rb
sbrnunes: I think we shouldn't. From what I understood, ReleaseDocs gets updated to failed when we get a failed Pod. Releases get updated to fail when we reach the timeout. Am I correct? If so, I think we should ignore a `failed` ReleaseDoc and not update the Release.

Were you doing something in here? Did I miss something in the merge process?

plugins/kubernetes/app/models/kubernetes/release_doc.rb
sbrnunes: Because I'm dumb. :)
Honestly, forgot about it.

plugins/kubernetes/app/models/kubernetes/role_config_file.rb
sbrnunes: Oh, good catch.
I added it here first, before adding the flag to the initializer. Forgot to remove it form here.

plugins/kubernetes/samson_kubernetes.gemspec
```diff
@@ -3,7 +3,7 @@ Gem::Specification.new 'samson_kubernetes', '0.0.1' do |s|
   s.description = 'TBD'
   s.authors = ['Jon Moter', 'Maciek Sufa', 'Sergio Nunes', 'Shane Hender']
   s.email = 'jmoter@zendesk.com'
-  s.add_runtime_dependency 'kubeclient', '>= 0.9'
+  s.add_runtime_dependency 'kubeclient', '>= 1.0.0'
   s.add_runtime_dependency 'celluloid'
   s.add_runtime_dependency 'celluloid-io'
   s.add_runtime_dependency 'http'
```
msufa: Need to update this to `s.add_runtime_dependency 'http', '= 0.9.8'`

plugins/kubernetes/app/models/watchers/base_cluster_watcher.rb
sbrnunes: Now I'm the one confused. :)

What do you have in mind with composition in this case? Making `ClusterPodWatcher` and `ClusterErrorPodWatcher` own a ClusterWatcher? I can't see what else could be extracted.
I know Composition tends to be better than inheritance, but in this case... I'm missing how it could make this more clean.

An alternative to creating the hash would be to pass only the cluster and have a template method on the BaseClusterWatcher, overrided in the child classes, that given a cluster returns the required stream. Would you prefer this approach?

plugins/kubernetes/app/models/watchers/cluster_pod_watcher.rb
sbrnunes: Yep, noticed that in the meantime. I already removed this. Thanks for the heads up anyway.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
sbrnunes: Will do.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
sbrnunes: You're right. It's redundant. I'll think about it.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
sbrnunes: This should be debug. Thanks for the heads up.

plugins/kubernetes/samson_kubernetes.gemspec
sbrnunes: Will do.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
```diff
@@ -1,34 +1,42 @@
...
     before do
       Kubernetes::Cluster.any_instance.stubs(:client).returns(Kubeclient::Client.new('http://cluster.localhost'))
-      stub_request(:get, %r{http://cluster\.localhost.*}).to_return(body: '{}')
     end
 
     it 'returns an empty array if no pods are running' do
-      get :index, project_id: :foo, environment: environments(:production).id, format: :json
+      environment.cluster_deploy_groups.each do |cluster_deploy_group|
+        stub_request(:get, %r{http://cluster\.localhost.*/namespaces/#{cluster_deploy_group.namespace}.*})
+          .to_return(body: '{}')
```
msufa: Do we actually need to stub it for every deploy group, can't we do it once for all like it was done originally?

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
```diff
@@ -1,34 +1,42 @@
...
     end
 
     it 'passes correct params to client' do
-      Kubeclient::Client.any_instance.expects(:get_pods).once.with(
-          { namespace: 'test_namespace', label_selector: 'project=project' }).returns([])
-      # expect this call at least once because we might have more deploy groups configured
-      Kubeclient::Client.any_instance.expects(:get_pods).at_least_once.with(
-          { namespace: 'default', label_selector: 'project=project' }).returns([])
-      get :index, project_id: :foo, environment: environments(:production).id, format: :json
+      Kubeclient::Client.any_instance.expects(:get_pods).times(2).returns([])
```
msufa: This test no longer does what it was supposed to do, namely, check if the correct `namespace` and `label_selector` are passed to the client. Now it just checks that the client is called twice.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
```diff
@@ -36,30 +44,112 @@
 
   private
 
-  def api_response
-    response = JSON.parse(parse_json_response_file('kubernetes_pod_api_response'))
```
msufa: Since you decided not to use template files, you might as well remove `parse_json_response_file` from `test_helper.rb` and remove the .json templates themselves from the `samples` dir.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
```diff
@@ -36,30 +44,112 @@
...
-  end
-
   def compare_json(expected, actual)
     expected_json = JSON.parse(expected)
     actual_json = JSON.parse(actual)
-    [expected_json, actual_json].each do |json|
-      delete_ids_from json # drop IDs, they're generated
-    end
-    actual_json.must_equal expected_json
+    expected_json.must_equal actual_json
```
msufa: If I remember correctly, the previous version of this assertion was intentional, so that the test output in case of a failure correctly identifies the expected and actual values.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
sbrnunes: Can't remember, I think I had problems with this and had to stub one request for each namespace. Can try to change it back and see if it still works.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
sbrnunes: You're absolutely right. Can't remember why I changed this. For sure, it was failing because of some change, but should have updated the expectation instead of taking it out. I'll review this.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
sbrnunes: I think I removed them in the branch that is focused solely in the tests. In this PR I just tried to fix the tests that were failing.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
sbrnunes: Humm, I think I got confused here with the order of the expected / actual values.

plugins/kubernetes/app/models/kuber_deploy_service.rb
sbrnunes: I actually gave it another thought and started supervising it. We can restart it, anyway, and having it supervised protects us from any unexpected crash, which is good if we're in the middle of a deployment: if it crashes, Celluloid will bring it back, the watcher will get in-sync with the cluster and proceed with the deployment.

So, good point in here. Thanks.

plugins/kubernetes/app/models/kubernetes/release.rb
msufa: You're right @sbrnunes, I was just making sure we didn't miss anything.

plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
```diff
@@ -30,10 +34,43 @@ def raw_template
       @raw_template ||= build.file_from_repo(kubernetes_role.config_file)
     end
 
-    def add_labels
-      labels.each do |key, value|
+    # This key replaces the default kubernetes key: 'deployment.kubernetes.io/podTemplateHash'
+    # This label is used by kubernetes to identify a RC and corresponding Pods
+    def set_rc_unique_label_key
+      template.spec.uniqueLabelKey = CUSTOM_UNIQUE_LABEL_KEY
```
msufa: This doesn't seem very unique to me :) If there are multiple deployments going on, won't they interfere with each others replication controllers?

sbrnunes: He He this is confusing.

So, that `CUSTOM_UNIQUE_LABEL_KEY`, which I have set to `rc_unique_identifier`, is just the name of the key that is going to be used to store the RC unique identifier which, by default, is `deployment.kubernetes.io/podTemplateHash`. But the identifier itself will be unique and auto-generated in the cluster. We don't have control over that.

The reason why I changed this key name was to make it more meaningful, when looking at the cluster data but also when using it on the code.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -36,36 +33,44 @@ def service
...
+    def update_replica_count(new_count)
+      update_attributes!(replicas_live: new_count)
+    end
+
+    def live_replicas_changed?(new_count)
+      new_count != replicas_live
+    end
+
+    def can_recover?(failed_pods)
+      failed_pods == 0
```
msufa: Why not use `failed_pods.zero?` like a few lines above?

So this means that if a single pod fails we can no longer recover? Why the need for a timeout in such a case?

plugins/kubernetes/app/models/kubernetes/role_config_file.rb
```diff
@@ -21,11 +14,11 @@ def initialize(contents, file_path)
...
-      parse_replication_controller
+      parse_deployment
       parse_service
     end
 
-    def parse_replication_controller
+    def parse_deployment
       deployment_hash = as_hash('Deployment')
       raise 'Deployment specification missing in the configuration file.' if deployment_hash.nil?
       @deployment = Deployment.new(deployment_hash, :recurse_over_arrays => true)
```
msufa: Sorry, just noticed `recurse_over_arrays` is redundant here as well.

plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
```diff
@@ -2,34 +2,38 @@ module Watchers
...
     end
 
-    def project_name(involved_object)
-      pod = @client.get_pod(involved_object.name, involved_object.namespace)
-      pod.metadata.labels ? pod.metadata.labels.project : nil
+    def get_pod(involved_object)
+      api_pod = cluster.client.get_pod(involved_object.name, involved_object.namespace)
+      Kubernetes::Api::Pod.new(api_pod)
     rescue KubeException => e
-      if e.error_code == 404
```
msufa: Checking for the 404 was done to handle events for deleted pods where we could not get the pod through the `get_pod` call anymore. Now it seems we ignore all errors coming from the API.

plugins/kubernetes/app/models/watchers/cluster_pod_watcher.rb
```diff
@@ -2,16 +2,20 @@ module Watchers
...
 
     private
 
     def handle_notice(notice)
-      if notice.object.metadata.labels
-        project = notice.object.metadata.labels['project']
-        publish(project, notice) if project
+      pod_event = Events::PodEvent.new(notice)
+
+      error("Invalid Kubernetes event: #{notice}") unless pod_event.valid?
```
msufa: Why not put that in an `else` clause of the below `if`?

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -1,119 +1,254 @@
...
-        error 'invalid k8s pod event'
-        false
+    # From the target environment for the current Release, fetches all existing Pods from the corresponding
+    # Kubernetes Clusters and updates the internal data structures used by the watcher.
+    def fetch_cluster_data
+      env = release_env(last_release(project))
+      env.cluster_deploy_groups.each do |cdg|
+        cdg.cluster.client.get_pods(namespace: cdg.namespace, label_selector: "project_id=#{project.id}").each do |pod|
+          update_rc_pods(Kubernetes::Api::Pod.new(pod))
+        end
```
msufa: This code looks very similar to what's happening in the dashboard controller. Maybe we could extract fetching pods for a specific environment and project to a single place?

plugins/kubernetes/app/models/kubernetes/release_doc.rb
sbrnunes: This line is being used by the deploy watcher to drive the way how we update the release doc. Each time we get a cluster event (pod error), we calculate the current amount of failed pods. If it reached zero, it means the release doc can recover and we can proceed with a regular update (status, live replicas, etc). But if the current amount is still greater than zero, we update only the live replicas and leave the status as `failed`.

plugins/kubernetes/app/models/kubernetes/role_config_file.rb
sbrnunes: Thanks. Another good catch. ;)

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -1,119 +1,254 @@
...
     end
 
-    def release_doc_from_rc_name(name)
-      @release.release_docs.select { |doc| doc.replication_controller_name == name }.first
+    # Will mark as :dead each previous Release for which there is no Pod in the cluster.
+    # Will mark as :dead each ReleaseDoc belonging to a previous Release for which there is no Pod in the cluster.
+    def reconcile_old_releases
+      excluded_from_sync = release_ids_from_cluster << last_release(project).id
+
+      scope = project.kubernetes_releases.excluding(excluded_from_sync)
```
msufa: `scope` sounds very JS like here, why not simply call it `previous_releases`?

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -1,119 +1,254 @@
...
+      update_rc_pods(pod)
+      release = Kubernetes::Release.find(pod.release_id)
+      release_doc = release.release_doc_for(pod.deploy_group_id, pod.role_id)
+
+      live_pods = count_live_pods(pod.rc_unique_identifier)
+      failed_pods = count_failed_pods(pod.rc_unique_identifier)
 
-      if pod_event.deleted?
-        rc.delete(pod_event.pod.name)
+      if !recoverable_fail?(release_doc, failed_pods)
```
msufa: `unless`?

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -1,119 +1,254 @@
...
+      rcs.map { |_, pods| pods.map { |_, pod| pod.release_id } }.flatten.uniq
+    end
+
+    def last_release(project)
+      project.kubernetes_releases.last
+    end
+
+    def count_live_pods(rc_unique_identifier)
+      rc = rc_pods(rc_unique_identifier)
+      rc.reduce(0) { |count, (_, pod)| count += 1 if pod.live? && !deleted?(pod); count }
```
msufa: Something doesn't feel right if we can have pods that are both live and deleted. Maybe the `Deleted` pod module could override `live?` and set it to `false` so we don't need to check both?

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
```diff
@@ -36,30 +44,112 @@
...
+          }
+        }
+      }
+    }
+  end
+
+  def add_role(response, release_doc)
+    response << role_template(release_doc.kubernetes_role) unless response.any? { |role|
+      role.id == release_doc.kubernetes_role.id
+    }
```
msufa: We should use `do` for multiline blocks I guess.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
```diff
@@ -36,30 +44,112 @@
...
+  end
+
+  def release_template(release, release_doc)
+    RecursiveOpenStruct.new({
+      'id': release.id,
+      'build': release.build.label,
+      'target_replicas': release_doc.replica_target,
+      'live_replicas': 1
+    })
+  end
```
msufa: I used a JSON template file for 2 reasons:
- It's easy to see the actual structure of the data, which could be useful for someone new starting work on this area of the code.
- You could have an error in your code which generates the requests and expected responses, causing the tests to return false positives. I wouldn't be sure if everything works fine just by looking at the test code.

It's all a matter of preference though :)

plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
msufa: Gotcha! I realised that after going through the rest of the code. Sorry! :)

plugins/kubernetes/app/models/kubernetes/release_doc.rb
msufa: OK, got it. Maybe we could call the method `recovered?` in this case to make things more obvious?

plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
sbrnunes: The problem was that any other errors apart from 404 were causing the exception to be raised and the actor to crash. Even though it is supervised, between crashing and restarting, we were loosing events, thus, the UI was not being updated correctly. So, my thinking here is: we make an attempt to get the Pod. If it fails, we log the exception and return `nil`. Then, in the method `handle_notice`, we ignore the event if we were unable to get the Pod.

Example of an error that was causing the watcher to crash: 401 error returned by the Cluster when querying for a Pod that no longer exists, after the internal event history in Kubernetes has been wiped out (apparently, there's a treshhold in Kubernetes after which the obsolete events (events related to non-existent resources) are cleaned.

I was able to see this happening quite a lot when triggering several consecutive releases, some of them with a big number of replicas.

plugins/kubernetes/app/models/watchers/cluster_pod_watcher.rb
sbrnunes: For no reason at all. I'll change it.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
sbrnunes: Yeah... I was also thinking about that. And also that some of the code in this Watcher could, maybe, be refactored out to a separate class (e.g., KuberSynchronizationService).

I won't do it in this PR, but I'll keep a mental note to review this.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
sbrnunes: LOL, I was actually following the pattern I've seen in another places in Samson.
I noticed that when performing searches using `ActiveRecord` scopes, and in particular when building a search using different scopes, we tend to use a scope variable to denote the `ActiveRecord::Relation` returned from calling each scope.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
sbrnunes: Good catch.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
sbrnunes: Humm, that's a good suggestion, yes. That hasn't crossed my mind because I've been treating the `live?` and `deleted?` concepts as completely independent, as this information comes from different sources, but no harm should come from doing that. I'll give it a try.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
sbrnunes: Fair point. I'll change that.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
sbrnunes: I had to change it because I wanted to generate the whole response payload programatically, from all the `releases` and `release_docs` coming from fixtures. The alternative would have been to access each item in the response items array and update manually each of the values that I wanted to make sure that were added to the response. It wouldn't be very flexible.

Regarding your (valid) concerns, any error in the test code that generates the expected responses should be caught right away as the production code, in the controller, is tightly coupled to the structure of the responses coming from the Cluster. Finally, the response from the controller is validated in the test (`expected_json.must_equal actual_json`) so, any error should be easily detected. I find it really hard to get a false positive in this case.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
sbrnunes: Ok, makes sense. Will do that.

plugins/kubernetes/app/models/concerns/kubernetes/has_status.rb
sbrnunes: Just FYI: left the `fail!` method on the Release and `ReleaseDoc` classes because the implementation differs now.

plugins/kubernetes/test/controllers/kubernetes_dashboard_controller_test.rb
sbrnunes: Yeah... The thing is that, now, I'm making a request against each Namespace / DeployGroup (I changed the fixtures used to cover that scenario). That is why I need to stub one request per each namespace.

msufa: Agree, but it's a regex, you can just ignore the namespace & deploy group, or is there a reason why you need to stub these specific patterns?

In the original approach I only cared about the hostname.

sbrnunes: The problem is not the pattern, is the number of times you are registering the stub. I mean, a valid alternative using the original pattern would probably be:

```
stub_request(:get, %r{http://cluster\.localhost.*}).times(environment.cluster_deploy_groups.count).to_return(body: '{}')
```

The advantage of having the explicit namespace in the pattern is that, if some other request is made that does not fall into the registered patterns, the test will fail.

sbrnunes:
@msufa, @henders : I think I have addressed all the comments. :relieved: See if you want to give it a final look or give it a quick run on your machines. 

Looking forward to get the :+1: :wink:

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -1,119 +1,265 @@
...
+    end
+
+    def terminate_watcher
+      info('Deploy finished!')
+      terminate
+    end
 
-      Rails.logger.info("[SSE] Sending: #{base.merge(options)}")
+    def send_event(release_doc, failed_pods)
+      info("[SSE] Sending: #{sse_event_data(release_doc, failed_pods)}")
```
msufa: This should be `debug`

sbrnunes: Added it back here because it might be useful to track deployments. It does not cause too much noise because the events are filtered.

msufa:
I've had one last look and have raised some reservations which were mostly addressed. @sbrnunes, you mentioned you'll have a separate PR with some tests, in that case I won't spend much time on this area here.

This is a huge and complicated PR so I'd like someone else to have a look as well and give the final :+1: 

/cc @henders @jonmoter

sbrnunes:
Yep, just need to fix the tests after this last changes and add a few other tests for the new stuff we introduced this week.

plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
```diff
@@ -2,34 +2,38 @@ module Watchers
   class ClusterPodErrorWatcher < BaseClusterWatcher
     include Celluloid::Notifications
```
msufa: Missing `include Watchers::TopicSubscription`

This watcher will crash on each attempt to publish to a topic. I had my reservations towards that mixin and my hunch was confirmed when I tried running some ad-hoc tests. Maybe it would be safer not to include that module but call the method directly?

sbrnunes: The problem was that I forgot to include the `TopicSubscription` module in this watcher. I'm not monkey patching the watchers, I'm explicitly including the module in the watchers and calling the method when needed. I just forgot to do so in this watcher and, somehow, I didn't catch it in the tests I've been running.

sbrnunes: Nevermind what I said, was not understanding what you meant with calling it directly, I thought you were thinking that I was monkey patching the watchers.

I agree. Including the module was not the best approach in this case. I changed it to call the method directly. Thanks for the suggestion.

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_dashboard_ctrl.js
```diff
@@ -78,7 +80,7 @@ samson.controller('KubernetesDashboardCtrl',
       var deferred = $q.defer();
       kubernetesService.loadDashboardData($scope.project_id, $scope.environment).then(
         function(data) {
-          if (_.isNotUndefinedOrEmpty(data)) {
+          if (_.isDefined(data)) {
```
henders: Ah, nice catch... I think I was running into this problem.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -36,36 +33,52 @@ def service
...
-      if replicas_live >= replica_target
-        self.status ='live'
-      elsif replicas_live > 0 && !failed?
-        self.status ='spinning_up'
+    def update_status(live_pods)
+      case
+      when live_pods == replica_target then self.status = :live
+      when live_pods.zero? then self.status = :dead
+      when live_pods > replicas_live then self.status = :spinning_up
+      when live_pods < replicas_live then self.status = :spinning_down
```
henders: Wondering if this might lead to confusion if user has just deployed this release, and pods are crashing and getting restarted, the status might get set to 'spinning down'?

plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
```diff
@@ -2,34 +2,38 @@ module Watchers
...
-      @client = client
-      super(@client.watch_events(field_selector: 'involvedObject.kind=Pod'))
+    def initialize(cluster)
+      super(cluster.client.watch_events(field_selector: 'involvedObject.kind=Pod'))
+      @cluster = cluster
     end
 
     private
 
+    attr_reader :cluster
```
henders: It's a bit weird to see an attr-reader declaration down here.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -7,113 +6,259 @@ class DeployWatcher
...
+      when event.kind == 'Pod' then handle_pod_event(message)
+      else error "Object kind not yet supported: #{notice.object.kind}"
       end
     end
 
     private
 
-    def handle_event_update(release_doc)
-      release_doc.failed? ? false : release_doc.fail!
+    attr_reader :project
```
henders: Again not really seeing the point behind private accessors instead of just using ```@project```

plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
sbrnunes: Just thought it was cleaner to have this instead of the `@` spread across the code.
Then I was digging around it and saw some code examples with this things declared at the top after the `private` keyword, so I thought it was a common approach.
Regarding the place, I tend to declare all public methods and the the private methods, so, was not sure where to put it to be honest. :)

I'll take it out.

plugins/kubernetes/app/views/kubernetes_dashboard/_deploy_progress_widget.html.erb
```diff
@@ -10,17 +10,19 @@
...
     </div>
-    <div class="deploy-group-release" ng-class="{ 'current-release': currentRelease(release), 'target-reached': targetStateReached(release) }"
-         ng-repeat="release in deployGroup.releases | orderBy:'id':true" ng-if="!deployCompleted()">
-      <div class="progress-wrapper">
-        <label>{{buildLabel(release)}}.{{releaseId(release)}}</label>
+    <div style="flex: auto;" ng-if="!deployCompleted() && !deployFailed()">
+      <div class="deploy-group-release" ng-class="{ 'current-release': currentRelease(release), 'target-reached': targetStateReached(release) }"
+           ng-repeat="release in deployGroup.releases | orderBy:'id':true">
+        <div class="progress-wrapper">
+          <label>{{buildLabel(release)}}.{{releaseId(release)}}</label>
```
henders: Why not just ```{{release.build}}.{{release.id}}```?

henders:
lgtm :+1: regardless of addressing minor comments.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
sbrnunes: Exactly, and that's the reason why I was wondering if we shouldn't just take this out, and update just the `deployed_at` timestamp indication when the deployment finishes. I've never seen a Pod crashing during a release, but assuming it can happen, how do we update the status? How can we distinguish a release that is spinning down from an ocasional crash? Maybe we can ignore it if it's the current release, but I wonder how many edge cases that we don't know yet will we get and if we really need this complexity.

We could simplify the states to just `deployed` and `failed_to_deploy`, and only update them at the end of the deployment. 

Food for thoughts when we come back to this?

henders: Yes, agree to coming back to it!

plugins/kubernetes/app/views/kubernetes_dashboard/_deploy_progress_widget.html.erb
sbrnunes: Initially I was using the objects directly and querying their properties, but after having to change the implementation a few times and having to to search across the html by all the references to the method properties in order to replace it, I switched over to methods.

By extracting it to a method, we can reuse it and if we ever need to change the implementation, we can just change it in the method and it will be applied everywhere it is used. Even if we need to rename the method to reflect what it is returning now, a refactor should do it, it will change the JS code but also the HTML code, as far as I remember.

So, in the end, a matter of preference, I would say, but I can say I got a few benefits from it. ;)

henders: Fair enough!

sbrnunes:
@henders, @msufa, had to commit a small fix in order to make the tests pass in the other PR. The refactoring on the Watchers broke a test because of the way how we were initialising the watch stream. Can you guys take a look into the last commit, please?

In the meantime, addressed all your comments, @henders. As soon as you guys review this last commit, I'll go ahead with the merge.

msufa:
:+1:

henders:
:+1:

PR #699, opened 2015-12-21
Adding feedback that the User clicked "Deploy" to create a Kubernetes release

sbrnunes:
Just a few fixes to the create Kubernetes Release dialog to provide feedback, while locking the form and the buttons, when the User clicks on the Deploy button.

/cc @henders, @msufa, @jonmoter 

screenshot 2016-01-04 17 02 16

### DIFF
--- a/plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_release_ctrl.js
+++ b/plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_release_ctrl.js
@@ -6,6 +6,7 @@ samson.controller('KubernetesCreateReleaseCtrl',
     $scope.loadingRoles = false;
     $scope.loadingEnvironments = false;
     $scope.loadingDeployGroups = false;
+    $scope.submitting = false;
     $scope.roles = [];
     $scope.environments = [];
     $scope.environment = undefined;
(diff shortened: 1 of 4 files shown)


henders:
Pics? :)

sbrnunes:
Just added the screenshots. :)

henders:
:+1:

plugins/kubernetes/app/assets/javascripts/kubernetes/controllers/kubernetes_create_release_ctrl.js
```diff
@@ -75,10 +76,12 @@ samson.controller('KubernetesCreateReleaseCtrl',
     };
 
     $scope.toggleAll = function() {
-      var newState = !$scope.allToggled();
-      _.each($scope.deploy_groups, function(deploy_group) {
-        deploy_group.selected = newState;
-      });
+      if(!$scope.submitting) {
```
msufa: Just a suggestion: instead of performing the background computations to check for an ongoing deploy and disable toggling active deploy groups, we could just disable the `Clear All` link and get the same result. An additional benefit would be that the user would actually know that this functionality is disabled.

sbrnunes: Unfortunately, disabling a link, i.e., setting the `disabled` attribute, is still not a standard across browsers, thus, not supported is some browser:

http://stackoverflow.com/questions/10276133/how-to-disable-html-links

I did try to disable it by setting the attribute `disabled`, but it didn't work, so, the workaround was to prevent the action from happening in the javascript code.

However, I should have greyed out the text when the form is disabled, but forgot to do it. I'll create a separate PR for taking care of the styling part of it.

Thanks for the heads up.

PR #700, opened 2015-12-21
Adding the ability to run the plugins tests as part of a regular build

sbrnunes:
I've noticed the plugins tests are not being run as part of the default build, not even in Travis. Because of that, we've been accumulating a few failed tests.

I do not know if this was intentional or not, so let me know if it was. Nevertheless, this PR intends to make the plugin tests run by default as part of a regular build.

I also noticed that when executing `rake plugins:test` the plugins tests were being executed against the `development` environment, instead of the `test` environment, which was messing up my development database all the time. So, I moved the whole task into the `test` namespace.

This PR is depending on #664 and #717. All plugin tests have been fixed in PR #717.

/cc @zendesk/samson, @jonmoter, @henders, @steved 

### Tasks
 - [x] Changed Rakefile to make plugins tests run as part of the normal build
 - [x] Fix failing tests

### Risks
 - Breaking the default build in the CI server

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,12 @@ env:
   matrix:
     - TASK=js
     - TASK=precompile
-    - DATABASE_URL=mysql2://travis@127.0.0.1/samson_test USE_UTF8MB4=1 TASK=rake
-    - DATABASE_URL=postgresql://postgres@127.0.0.1/samson_test TASK=rake
-    - DATABASE_URL=sqlite3://null$PWD/db/test.sqlite3 TASK=rake
+    - DATABASE_URL=mysql2://travis@127.0.0.1/samson_test USE_UTF8MB4=1 TASK=rake:default
+    - DATABASE_URL=postgresql://postgres@127.0.0.1/samson_test TASK=rake:default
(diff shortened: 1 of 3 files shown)


steved:
Definitely want this, good catch

sandlerr:
would it be better to make it a separate task? That way plugin tests can be a separate job in travis

sbrnunes:
If we don't enhance the default `test` task, it could be run as `test:plugins` in a separate job.

Would you guys prefer to have 3 extra jobs running the plugin tests against each supported database, as it is done for the regular tests?

sbrnunes:
Following @sandlerr thoughts, I removed the enhancement that adds the plugin tests to the default `test` task and updated the travis YAML file and the CI shell script to trigger the plugin tests as separate jobs in travis. It not only gives us the advantage of running the plugin tests in separate jobs, but it also allows us to see, in a much intuitive way, if the tests failing belong to Samson core or to some plugin.

sbrnunes:
Closing this PR as @grosser addressed this same issue in #791. The only difference is that @sandlerr suggested that we could run the plugin tests in separate tasks, in Travis.

This PR was also depending on #664, from @henders (because of the ActiveRecord connections). @grosser also addressed that issue in #791.

PR #702, opened 2015-12-22
fix the intermittent failing flowdock test

steved:


steved:
Also start running jshint against plugins. @zendesk/samson

sbrnunes:
Cool, really helpful this fix. :+1:

PR #703, opened 2015-12-22
fix some cache keys for stages

steved:
Cache keys weren't updated when project level roles were introduced.

@zendesk/samson


jonmoter:
Thanks for the catch!  :+1:

PR #705, opened 2015-12-24
CSV export for admin/users and deploys/recent

rikeoka:
Add CSV export to admin/users and deploys/recent pages to support audit requests without directly exporting from the databases.  No filtering other than default filters for the User and Deploy classes.

Deploys/Recent UI: Add Download as CSV under filter bar
screen shot 2015-12-23 at 5 18 57 pm

Admin/Users UI:  Add Download as CSV under search box
screen shot 2015-12-23 at 5 19 06 pm


app/controllers/users_controller.rb
```diff
@@ -1,14 +1,46 @@
...
+  private
+
+  def user
+    @user ||= User.find(params[:id])
+  end
+
+  def user_params
+    # also reset pending access request, so the user can request further access rights
+    params.require(:user).permit(:role_id).merge(access_request_pending: false)
+  end
```
henders: Undo all these changes, as there is an actual /app/controllers/admin/users_controller.rb file that you should update instead. This controller is only used to list users on a different page.

app/models/deploy.rb
```diff
@@ -121,6 +121,34 @@ def self.expired
...
+    end
+  end
+
+  def self.to_csv
+    per = 1000
+    max = Deploy.count.fdiv(per).ceil
+    CSV.generate do |csv|
+      csv << ["Deploy Number", "Project Name", "Deploy Sumary", "Deploy Updated", "Deploy Created", "Deployer Name", "Buddy Name", "Production Flag", Deploy.count]
+      for page in 1..max
+        @deploys = Deploy.all.page(page).per(per)
```
henders: Use the find_each batching to perform this cleaner: http://guides.rubyonrails.org/active_record_querying.html#retrieving-multiple-objects-in-batches

app/models/user.rb
```diff
@@ -24,6 +24,16 @@ def starred_project?(project)
     starred_projects.include?(project)
   end
 
+  def self.to_csv
+    @users = User.order(:id)
+    CSV.generate do |csv|
+      csv << ["id","name","email","role"]
+      @users.each do |user|
+        csv << [user.id, user.name, user.email, user.role.name]
+      end
```
henders: Probably better to use the batching like above just in case people go crazy creating users: http://guides.rubyonrails.org/active_record_querying.html#retrieving-multiple-objects-in-batches

app/models/deploy.rb
```diff
@@ -121,6 +121,34 @@ def self.expired
...
+    elsif buddy.nil? && pending?
+      "Pending Buddy"
+    elsif buddy.nil?
+      "No Buddy"
+    elsif (user.id == buddy.id)
+      "Buddy Bypassed"
+    else
+      buddy.name
+    end
+  end
```
henders: These seem very wordy for a simple CSV export. The column is called 'Buddy Name', so is each value containing the word 'buddy' not redundant?

app/controllers/users_controller.rb
rikeoka: Reverted and used /app/controllers/admin/users_controller.rb in commit 2ae95a4

app/models/deploy.rb
rikeoka: Used find_each  in commit 5f4d76e

app/models/user.rb
rikeoka: Used find_each instead in commit 3e19e5f and also added project roles to the report.

app/models/deploy.rb
rikeoka: Removed "buddy" from the column output, and also updated first condition to use the deploy_requires_approval? function instead of recreating the condition.

henders:
Nice, much cleaner. Mind creating a unit test for each export type?

henders:
:+1:

PR #706, opened 2015-12-28
[CVE-2015-5312] Updates Nokogiri gem to v1.6.7.1 to avoid vulnerabilities in libxml2.

mariozaizar:
```
Name: nokogiri
Version: 1.6.6.4
Advisory: CVE-2015-5312
Criticality: High
URL: https://groups.google.com/forum/#!topic/ruby-security-ann/aSbgDiwb24s
Title: Nokogiri gem contains several vulnerabilities in libxml2
Solution: upgrade to >= 1.6.7.1
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem 'kaminari'
 gem 'soft_deletion', '~> 0.4'
 gem 'dalli', '~> 2.7.0'
 gem 'active_model_serializers', '~> 0.8.0'
+gem 'nokogiri', '1.6.7.1'
 
 gem 'sawyer', '~> 0.5'
 gem 'sse-rails-engine', '~> 1.4'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -25,6 +25,7 @@ gem 'kaminari'
 gem 'soft_deletion', '~> 0.4'
 gem 'dalli', '~> 2.7.0'
 gem 'active_model_serializers', '~> 0.8.0'
+gem 'nokogiri', '1.6.7.1'
```
sandlerr: isn't it better to leave a comment at least when locking to a specific version? In 6 months time nobody will know why this is here. Alternatively you could just `bundle update nokogiri` and not even change the Gemfile

mariozaizar: Yup, agree. Specific version is only required if the gem is already being locked with the old version... In this case, the Gemfile.lock + the .gem are the only required changes. Will update the PR. Thanks for the feedback!

kintner:
:+1:

henders:
:+1:

PR #708, opened 2015-12-31
Update to use correct casing for GitHub

jacobbednarz:
Where possible (that is outside of class names, etc) use the correct casing for
the word GitHub.


henders:
:+1:

PR #709, opened 2016-01-01
Change command usage display to use lists

adammw:
:v:

/cc @zendesk/samson @sandlerr

### Description

Before: 

screen shot 2016-01-01 at 7 46 26 pm

After:
screen shot 2016-01-01 at 7 44 56 pm

### Risks
* None


henders:
Looks nice! :+1:

PR #711, opened 2016-01-05
Fix typo with function rename.

henders:
Bad typo when renaming a function. 

/cc @sbrnunes @msufa 

### References
 - Jira link: 

### Risks
 - None


msufa:
:+1:

PR #712, opened 2016-01-05
New pod error watcher and base class for cluster watchers

msufa:
- Extracted a common base class for cluster watchers
- Created new watcher to observe pod error notices (scheduling failed, could not download Docker image, etc.)
- Added a new `failed` release status
- Added a timeout for a release
 - default 10minutes, configurable
 - if the release is not successful and no new pod has started within the timeout period, the release is considered failed

/cc @sbrnunes @henders @jonmoter 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -20,6 +20,10 @@ module Kubernetes
       define_method("#{s}?") { status == s }
     end
 
+    def fail!
+      update_attribute(:status, :failed)
+    end
+
     def has_service?
       kubernetes_role.has_service? && service_template.present?
     end
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -56,9 +56,10 @@ module Kubernetes
 
       if replicas_live >= replica_target
         self.status ='live'
-      elsif replicas_live > 0
+      elsif replicas_live > 0 && !failed?
         self.status ='spinning_up'
       end
+      save!
     end
 
     def client
--- a/plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
+++ b/plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
@@ -1,0 +1,35 @@
+module Watchers
+  class ClusterPodErrorWatcher < BaseClusterWatcher
+    include Celluloid::Notifications
+
+    def initialize(client)
+      @client = client
+      super(@client.watch_events(field_selector: 'involvedObject.kind=Pod'))
+    end
+
+    private
+
+    def handle_notice(notice)
+      if error_notice?(notice.object.reason)
+        rc_name = rc_name(notice.object.involvedObject)
+        publish(rc_name, notice) if rc_name
+      end
+    end
+
+    def rc_name(involved_object)
+      pod = @client.get_pod(involved_object.name, involved_object.namespace)
+      pod.metadata.labels ? pod.metadata.labels.replication_controller : nil
+    rescue KubeException => e
+      if e.error_code == 404
+        warn e.to_s
+        nil
+      else
+        raise
+      end
+    end
+
+    def error_notice?(reason)
+      reason == 'failed' || reason == 'failedScheduling'
+    end
+  end
+end
--- a/plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
+++ b/plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
@@ -11,14 +11,14 @@ module Watchers
 
     def handle_notice(notice)
       if error_notice?(notice.object.reason)
-        rc_name = rc_name(notice.object.involvedObject)
-        publish(rc_name, notice) if rc_name
+        project_name = project_name(notice.object.involvedObject)
+        publish(project_name, notice) if project_name
       end
     end
 
-    def rc_name(involved_object)
+    def project_name(involved_object)
       pod = @client.get_pod(involved_object.name, involved_object.namespace)
-      pod.metadata.labels ? pod.metadata.labels.replication_controller : nil
+      pod.metadata.labels ? pod.metadata.labels.project : nil
     rescue KubeException => e
       if e.error_code == 404
         warn e.to_s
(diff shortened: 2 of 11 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -56,9 +56,10 @@ def update_replica_count(new_count)
 
       if replicas_live >= replica_target
         self.status ='live'
-      elsif replicas_live > 0
+      elsif replicas_live > 0 && !failed?
```
sbrnunes: I'm thinking if this should be a safe guard at the beginning of the method, I mean, should we return right away if the "release_doc" has failed?

```
def update_replica_count(new_count)
  return if failed?
  ...
end
```

Because as soon as a pod fails, this status should never be updated again, I think.

plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
```diff
@@ -0,0 +1,35 @@
...
+      @client = client
+      super(@client.watch_events(field_selector: 'involvedObject.kind=Pod'))
+    end
+
+    private
+
+    def handle_notice(notice)
+      if error_notice?(notice.object.reason)
+        rc_name = rc_name(notice.object.involvedObject)
+        publish(rc_name, notice) if rc_name
```
sbrnunes: In here we need to publish to the project topic. I saw you already adapted the ClusterPodWatcher to do so, we just need to do the same here.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -10,6 +10,7 @@ class DeployWatcher
     def initialize(release)
       @release = release
       @current_rcs = {}
+      @pod_timer = after(ENV.fetch('KUBERNETES_POD_TIMEOUT', 600)) { pod_timeout }
```
sbrnunes: This is 10 mins right? Should we increase the default to last a little longer? I'm just thinking that 10 mins might not be too long for a medium-sized deploy. 

Hum, nvm, I just noticed we are updating the timer each time we get a valid update for a pod.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -20,18 +21,44 @@ def watch
...
       release_doc = release_doc_from_rc_name(topic)
-      pod_event = Events::PodEvent.new(data)
-      return error('invalid k8s pod event') unless pod_event.valid?
-      update_replica_count(release_doc, pod_event)
-      send_event(role: release_doc.kubernetes_role.name,
-                 deploy_group: release_doc.deploy_group.name,
-                 target_replicas: release_doc.replica_target,
-                 live_replicas: release_doc.replicas_live)
-      end_deploy if deploy_finished?
+      updates_found = if data.object.kind == 'Event'
```
sbrnunes: May I suggest `release_updated`?

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
```diff
@@ -20,18 +21,44 @@ def watch
...
+      end
     end
 
     private
 
+    def handle_event_update(release_doc)
+      if release_doc.failed?
+        false # no update needed, the release has already failed
+      else
+        release_doc.update_attribute(:status, :failed)
```
sbrnunes: What do you think of refactoring this logic to the `ReleaseDoc`, so that only the 'ReleaseDoc' knows about the possible values for the status?

Then, this would look just like:
```
release_doc.fail!
```

sbrnunes:
After addressing comments, it's a :+1: for me.

Out of curiosity, did you see the UI getting updated with the error message on a failed deployment?

plugins/kubernetes/app/models/kubernetes/release_doc.rb
msufa: There might be a glitch when deploying after which resources for scheduling a pod become available, I've seen this in ad-hoc tests. We kill the deploy watcher only after the deploy is live or there was a timeout, to give it a chance of recovery. The UI will flicker but at least we won't cause panic over a good deploy.

plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
msufa: Good catch, thanks!

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
msufa: This is just a default anyway, it's configurable via environment vars.

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
msufa: Maybe `release_doc_updated` to be more specific?

plugins/kubernetes/app/models/watchers/deploy_watcher.rb
msufa: Good point!

msufa:
@sbrnunes I did see the UI getting updated, I should have added a screenshot. I'll update the PR.

msufa:
@sbrnunes, I've addressed your comments, please have another look.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
sbrnunes: Hummm, good point. With the guard at the beginning of the method we would never be able to recover if resources become available.

I believe that's fine for now, but eventually, what we could do as well, is mark that Pod as 'Failed' when we get an error, in the internal data structure, and remove that "marker" if we get a new event for that Pod with the condition `Ready` set to `True`, in which case we could also update the status of the ReleaseDoc back to 'spinning-up' or 'live', depending on the new number of live replicas. The advantage of doing this is a more accurate update of the UI.

Food for thoughts when we come back to this?

plugins/kubernetes/app/models/watchers/cluster_pod_error_watcher.rb
```diff
@@ -0,0 +1,35 @@
...
+      @client = client
+      super(@client.watch_events(field_selector: 'involvedObject.kind=Pod'))
+    end
+
+    private
+
+    def handle_notice(notice)
+      if error_notice?(notice.object.reason)
+        project_name = project_name(notice.object.involvedObject)
+        publish(project_name, notice) if project_name
```
sbrnunes: When I submit my PR, after merging with this changes, we must double check that I have changed this to `project_id` (which was added to the metadata labels as well).

sbrnunes:
:+1: 

Really :heart: the refactoring on the watchers. :smile:

PR #714, opened 2016-01-05
destroy macros on project destroy

steved:
macros are soft deleted

@zendesk/samson


steved:
For reference: https://zendesk.airbrake.io/projects/95346/groups/1525983833869462659/notices/1591533811303726896

jonmoter:
:+1: 

Do we also need to do something to clean up the Macros associated with deleted projects?

steved:
Probably manually, yeah.

PR #715, opened 2016-01-06
Revoke project level deploy/admin access

rikeoka:
Add a option to project level authorizations to reduce a project level authorization to below deployer.  This allows an admin to revoke a user's project level authorization without having to remove the record in the db.

### DIFF
--- a/app/models/project_role.rb
+++ b/app/models/project_role.rb
@@ -5,6 +5,7 @@ class ProjectRole < ActiveHash::Base
   include ActiveModel::Serializers::JSON
 
   self.data = [
+    { id: -1, name: "default" },
     { id: 0, name: "deployer" },
     { id: 1, name: "admin" }
   ]
--- a/app/models/project_role.rb
+++ b/app/models/project_role.rb
@@ -5,7 +5,7 @@ class ProjectRole < ActiveHash::Base
   include ActiveModel::Serializers::JSON
 
   self.data = [
-    { id: -1, name: "default" },
+    { id: -1, name: "viewer" },
     { id: 0, name: "deployer" },
     { id: 1, name: "admin" }
   ]
--- a/docs/permissions.md
+++ b/docs/permissions.md
@@ -6,11 +6,13 @@ You can manage the roles of all your users from the 'Admin' -> 'Users' menu.
 Role | Description
 --- | ---
 Viewer | Can view all deploys.
-Deployer | Viewer + ability to deploy projects.
-Admin | Deployer + can setup and configure projects.
+Deployer | Viewer + ability to deploy all projects.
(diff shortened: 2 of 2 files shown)


rikeoka:
screen shot 2016-01-06 at 11 30 44 am

sbrnunes:
Hi @rikeoka,
Yeah, I didn't think on adding the option for revoking project level rights as the initial idea was to allow Samson to specify User admins on a project and to be able to remove these admin rights. But the way this ended up, with the introduction of the `Deployer` role at the project level as well, it would have made total sense to have the means to revoke any role.
So, good catch. I have just on single remark (see my comment on the ProjectRole class).

henders:
I would probably change the text back to 'viewer' to keep it consistent with the global roles. 'Default' is a bit ambiguous.

app/models/project_role.rb
```diff
@@ -5,6 +5,7 @@ class ProjectRole < ActiveHash::Base
   include ActiveModel::Serializers::JSON
 
   self.data = [
+    { id: -1, name: "default" },
```
sbrnunes: Should we name it `none` to give the explicit idea that any project level role will be revoked? I'm thinking that 'default' might be too ambiguous, may cause confusion around what's the behaviour of the default role.

An alternative would be `viewer`, to make it consistent with the system level roles. We'll also need to update the documentation to reflect this change.

sbrnunes: Hadn't noticed that @henders had already commented on this. I guess `viewer` gets the most votes. :smile:.

sbrnunes:
:+1: after comments have been addressed and documentation updated. You may find it in: `docs/permissions.md`. Thanks for taking care of this improvement.

msufa:
@sbrnunes, since demoting a user back to `Viewer` will result in a regular call to the `ProjectRolesController`, the `reset_access_request_flag` method should be called as usual, see https://github.com/zendesk/samson/blob/master/app/controllers/project_roles_controller.rb#L45
We should not see any issues here.

The only quirky thing I see is that the user will be able to request `Viewer` access for themselves from their profile page, but maybe that's a good thing since they will be able to give up excessive access rights to projects they're no longer involved with.

rikeoka:
Changed the downgraded role to "viewer". 

Added the role to the permissions.md file, added clarifications about the system-level deploy/admin roles applying to all projects, and under project role indicating that if system level role is more permissive than the project-role, the system-level role applies.  Basically to avoid confusion that the project roles can be used to downgrade access below a user's system-level role.  Example a user needs admin except for project A, so the user is assigned admin at the system-level and given viewer in Project A, would not work.

msufa:
:+1:

PR #716, opened 2016-01-07
Fix for empty referer when creating access request

msufa:
This might happen when someone pastes a link in the browser to create an access request. Take the user to the root page in such case.

/cc @zendesk/samson

### References
 - Jira link: 

### Risks
 - None


henders:
:+1: on :green_apple:

PR #718, opened 2016-01-12
[SAMSOM-123] make reorder work with references to deleted stages

craig-day:
Only check for circular pipelines when next_stage_ids is changed and only allow soft_delete if the stage is not referenced by any other stage

/cc @steved @grosser @jonmoter 

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-123

### Risks
 - Low. Circular pipeline check only runs in limited cases now.

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -28,7 +28,7 @@ class Stage < ActiveRecord::Base
   attr_writer :command
   before_save :build_new_project_command
 
-  before_soft_delete :check_stage_references
+  before_soft_delete :verify_not_part_of_pipeline
 
   def self.reorder(new_order)
     transaction do
@@ -224,14 +224,5 @@ class Stage < ActiveRecord::Base
 
   def permalink_scope
     Stage.unscoped.where(project_id: project_id)
-  end
-
-  def check_stage_references
-    project.stages.each do |s|
-      if s.next_stage_ids.include?(id)
-        errors[:base] << "Stage #{name} is referenced by another stage and cannot be deleted"
-        return false
-      end
-    end
   end
 end
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -28,6 +28,8 @@ class Stage < ActiveRecord::Base
   attr_writer :command
   before_save :build_new_project_command
 
+  before_soft_delete :check_stage_references
+
   def self.reorder(new_order)
     transaction do
       new_order.each.with_index { |stage_id, index| Stage.update stage_id.to_i, order: index.to_i }
@@ -222,5 +224,14 @@ class Stage < ActiveRecord::Base
 
   def permalink_scope
     Stage.unscoped.where(project_id: project_id)
+  end
+
+  def check_stage_references
+    project.stages.each do |s|
+      if s.next_stage_ids.include?(id)
+        errors[:base] << "Stage #{name} is referenced by another stage and cannot be deleted"
+        return false
+      end
+    end
   end
 end
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -28,8 +28,6 @@ class Stage < ActiveRecord::Base
   attr_writer :command
   before_save :build_new_project_command
 
-  before_soft_delete :verify_not_part_of_pipeline
-
   def self.reorder(new_order)
     transaction do
       new_order.each.with_index { |stage_id, index| Stage.update stage_id.to_i, order: index.to_i }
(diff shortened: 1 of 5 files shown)


craig-day:
marked it as WIP because I have a little bit to workout in the tests.

app/models/stage.rb
```diff
@@ -223,4 +225,13 @@ def permalink_base
...
   end
+
+  def check_stage_references
+    project.stages.each do |s|
+      if s.next_stage_ids.include?(id)
+        errors[:base] << "Stage #{name} is referenced by another stage and cannot be deleted"
+        return false
+      end
+    end
+  end
```
steved: Shouldn't this all be in the plugin?

craig-day:
@steved removed WIP, ready for review.

Moved the code to the right spot (I think?) and fixed my tests.

app/models/stage.rb
```diff
@@ -28,6 +28,8 @@ class Stage < ActiveRecord::Base
   attr_writer :command
   before_save :build_new_project_command
 
+  before_soft_delete :verify_not_part_of_pipeline
```
steved: This too

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -30,4 +30,16 @@ def valid_pipeline?(origin_id = id)
...
+  # Make sure that this stage isn't referenced by another stage in a pipeline.
+  # This will stop soft_deletion of the stage.
+  def verify_not_part_of_pipeline
+    project.stages.each do |s|
+      if s.next_stage_ids.include?(id)
+        errors[:base] << "Stage #{name} is in a pipeline from #{s.name} and cannot be deleted"
+        return false
+      end
+    end
+    true
```
steved: I think this is implicit

app/models/stage.rb
craig-day: if I move this to `plugins/pipelines/app/decorators/stage_decorator` I get an undefined method error for `before_soft_delete`.  I could move the `has_soft_deletion` call there too, but I'm not sure that makes sense. Maybe I'm trying to move it to the wrong spot?

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
craig-day: True. Was just following the from of the validation method above, but this isn't a validation. Consider it gone.

app/models/stage.rb
steved: That's probably not good, I'd think the stage was already loaded.

steved: that "Stage" not "the stage".

steved: Ugh yeah it runs on ActiveRecord::Base inheritance.

craig-day: So, we ok to leave this here? I don't know order of execution well enough to know where (if anywhere) else to put this.

steved: I would prefer not to, so let me think about it. This isn't urgent I'm assuming

craig-day: Not at all, was just curious on thoughts.

henders: Yea, unfortunately they made the decision to load the plugin classes before the main app classes, thus you'll see the ```prepend``` directive scattered about to work around this.

henders: I would vote for copying the ```has_soft_deletion default_scope: true``` into the pipeline stage concern, rather then moving it to address this. We're trying really hard not to put plugin specific logic into the core code.

henders:
I'd probably have preferred to allow the soft-deletion and just remove its reference to from the other stages. But this works too.

app/models/stage.rb
craig-day: @steved how do we feel about copying the `has_soft_deletion` into here?

steved: Blah, yeah ok. soft_deletion plugin one day?

craig-day:
@jonmoter @steved @henders applied feedback. This look good?

craig-day:
bump

plugins/pipelines/app/decorators/stage_decorator.rb
```diff
@@ -2,5 +2,10 @@
   prepend SamsonPipelines::StageConcern
   serialize :next_stage_ids, Array
 
-  validate :valid_pipeline?
+  validate :valid_pipeline?, if: :next_stage_ids_changed?
+
+  # duplicate call from models/stage.rb, but this is needed
+  # to load the soft delete methods
```
grosser: hmm decorators should be loaded after the base class is loaded I guess ... but that might be tricky :(

grosser:
:+1: :heart: the tests :)

PR #719, opened 2016-01-15
Fixing some project access rights

sbrnunes:
When the project access right were introduced, allowing Users with system level Viewer access to have special permissions on certain projects, a couple of places were missed, including in the buddy check feature.

/cc @zendesk/samson

### Steps to reproduce
Start a deploy to production. The deploy screen should display the buddy check request.

### Risks
 - None


henders:
:+1:

PR #724, opened 2016-01-20
allow selecting a different environment so we can deploy pod12 which …

grosser:
…is in pre-prod

@jonmoter 

screen shot 2016-01-20 at 10 59 18 am
screen shot 2016-01-20 at 10 59 08 am

https://zendesk.atlassian.net/browse/POD-236

### Risks
 - None

### DIFF
--- a/app/views/admin/deploy_groups/_environment_selector.html.erb
+++ b/app/views/admin/deploy_groups/_environment_selector.html.erb
@@ -1,0 +1,12 @@
+<%= form_tag '?', method: :get do %>
+  Environment: <%= select_tag :environment_id, options_for_select(Environment.all.map { |e| [e.name, e.id] }, params[:environment_id] || @deploy_group.environment_id) %>
+<% end %>
+
+<script>
+  $(function(){
+    // when environment selector is changed, reload the page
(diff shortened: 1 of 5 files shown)


app/views/admin/deploy_groups/_environment_selector.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= form_tag '?', method: :get do %>
+  Environment: <%= select_tag :environment_id, options_for_select([['Matching', '']] + Environment.all.map { |e| [e.name, e.id] }, params[:environment_id]) %>
```
jonmoter: How about the default value is the environment of the deploy group?

jonmoter:
:+1:

app/views/admin/deploy_groups/_environment_selector.html.erb
grosser: Yeah true ... that's what matching was supposed to mean ... but we can be
clearer :)

grosser: updated

PR #726, opened 2016-01-23
Pull Request Webhooks

mwerner:
WIP: First pass at accepting other types of github events. Would love some feedback on the approach.

I'm not totally convinced re-purposing `Changeset::PullRequest` is warranted, but I also didn't want to create a different `PullRequest` class or namespace for this stuff.

Todo:
- Needs some work on sha selection for PR and issue comments.
- No filtering setup yet on issue comment content.

@zendesk/samson


app/models/changeset/issue_comment.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def service_type
+    'pull_request'
+  end
+
+  private
+
+  def pull_request
+    @pull_request ||= Changeset::PullRequest.find(repo, data['number'])
```
steved: Is the PR inlined? Can we avoid another outbound request?

app/controllers/integrations/github_controller.rb
```diff
@@ -20,21 +21,35 @@ def valid_signature?
...
-    params[:ref][/refs\/heads\/(.+)/, 1]
+    webhook_event.branch
   end
 
   private
 
+  def webhook_event
+    return @webhook_event if defined?(@webhook_event)
+
+    @webhook_event = begin
```
steved: Just a nitpick, you can inline case statements

```
x = case ...
when ...
end
```

app/controllers/integrations/github_controller.rb
```diff
@@ -20,21 +21,35 @@ def valid_signature?
...
+        Changeset::PullRequest.find(project.repo_name, params[:number])
+      when 'issue_comment'
+        Changeset::IssueComment.new(project.repo_name, params)
+      end
+    end
+  end
+
   def service_type
-    'code'
+    webhook_event && webhook_event.service_type
```
steved: Maybe we raise in valid_payload or something so we don't have to worry about there being a nil webhook_event

steved:
Direction looks good

app/models/changeset/issue_comment.rb
mwerner: PR info isn't inline unfortunately.

mwerner:
closed in favor of #746

PR #727, opened 2016-01-26
Parse alternate style markdown headers for risks section

jden:
Currently, risks only show up if they are using hash-style headers,
e.g. `## Risks`. Markdown also specifies underline style, which is
useful for certain unixy environments which treat # as a comment-beginning
character and would otherwise ignore the heading, such as when creating
PRs using github's `hub` tool. This alternate style of headers looks like

```md
Heading 1
----

Heading 2
====
```

This PR updates samson to parse the risks section regardless of heading style.

Required
======
- [ ] Approval of @zendesk/samson

Risks
=====
- None

### DIFF
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -3,8 +3,8 @@ class Changeset::PullRequest
   CODE_ONLY = "[A-Z][A-Z\\d]+-\\d+"  # e.g., S4MS0N-123, SAM-456
   PUNCT = "\\s|\\p{Punct}|~|="
 
-  # Matches a section heading named "Risks".
-  RISKS_SECTION = /#+\s+Risks.*\n/i
+  # Matches a markdown section heading named "Risks".
+  RISKS_SECTION = /^\s*#*\s*Risks?\s*#*\s*\n(?:\s*[-=]*\s*\n)?/i
 
   # Matches URLs to JIRA issues.
   JIRA_ISSUE_URL = %r[https?:\/\/[\da-z\.\-]+\.[a-z\.]{2,6}\/browse\/#{CODE_ONLY}(?=#{PUNCT}|$)]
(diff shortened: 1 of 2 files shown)


app/models/changeset/pull_request.rb
```diff
@@ -3,8 +3,9 @@ class Changeset::PullRequest
   CODE_ONLY = "[A-Z][A-Z\\d]+-\\d+"  # e.g., S4MS0N-123, SAM-456
   PUNCT = "\\s|\\p{Punct}|~|="
 
-  # Matches a section heading named "Risks".
+  # Matches a markdown section heading named "Risks".
   RISKS_SECTION = /#+\s+Risks.*\n/i
```
grosser: should both not match `Risksetting` etc ... so ``\s*\n` instead of `.*` for both

grosser: also this one should be `^\s*#+Risks` since we don't want to match `There are some #Risks`

app/models/changeset/pull_request.rb
```diff
@@ -3,8 +3,9 @@ class Changeset::PullRequest
...
+  RISKS_SECTION_ALT = /^\s*Risks.*\n=+\s*$/mi
```
grosser: so this ... `Risks` or `Risk` with `#` in front of it and optionally followed by heading `=`(h1) or `-`(h2)
`/^\s*#*Risks?\s*\n(?:^\s*:[=-]+\s*$)?/mi`

test/models/changeset/pull_request_test.rb
```diff
@@ -310,6 +310,15 @@ def no_risks
       pr.risks.must_equal nil
     end
 
+    it "finds risks with underline style markdown headers" do
+      body.replace(<<-BODY.strip_heredoc)
+        Risks
+        =====
+          - Snakes
```
grosser: Snakes on a PR :trollface:

jden:
Alright, I did some regex fiddling and I think I found one that looks good: http://www.rubular.com/r/H3iJFLXaab

grosser:
looks good!

jden:
Alright, updated the PR, tests are green, I'll merge on your :+1:

grosser:
:+1:

PR #728, opened 2016-01-28
Increase the docker version we support for the build

andreionut:
:globe_with_meridians: :fuelpump: :cloud: 

/cc @fneves @henders 

### Risks
- Level: None


fneves:
:+1:

PR #730, opened 2016-01-29
put new stages at the back, they are usually just 1-off spam

grosser:
ops creates a new podX stage to deploy something new and it messes up the natural order of stages ... lets avoid that ...

@zendesk/runway 

### Risks
 - None


jonmoter:
:+1:

PR #731, opened 2016-01-29
sort often used commands to the front

grosser:
we have lots of commands and we should encourage reused -> often used goes to the top and unused can slowly fade away ...

@zendesk/runway 

screen shot 2016-01-28 at 8 31 05 pm

### Risks
 - None

### DIFF
--- a/test/models/command_test.rb
+++ b/test/models/command_test.rb
@@ -2,6 +2,8 @@ require_relative '../test_helper'
 
 describe Command do
   describe '.for_object' do
+    let(:stage) { stages(:test_staging) }
+
     it "shows global for new stage" do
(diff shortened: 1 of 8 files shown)


grosser:
@zendesk/runway

test/models/command_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    it "sorts own commands in front" do
+      stage.commands.size.must_be :>=, 1
+      Command.for_object(stage).must_equal(
+        stage.commands + (Command.global - stage.commands)
+      )
+    end
+
+    it "hightly used commands to the front" do
```
sandlerr: extra t in highly

sandlerr:
:+1:

PR #732, opened 2016-02-03
new header + corrected clone counters

grosser:
New and clone previously had no link back to the project ... bring them in line with show and edit:
screen shot 2016-02-03 at 1 54 14 pm

Clone has pre-checked checkboxes, so 1 was subtracted from the commands usage count, which means you could inline edit commands without any warning even though it was used by the cloned stage.

screen shot 2016-02-03 at 1 59 54 pm

@zendesk/runway 

### Risks
 - None


sandlerr:
:+1:

PR #733, opened 2016-02-05
allow bypass buddy check

grosser:
some stages need to be deployed automatically or do not affect production code (like check or deploy artifact construction) so having a straight forward way of disabling the buddy check would be great since it would stop the practice of not picking any deploy group, which hurts documentation and command reuse

@zendesk/runway @steved 

screen shot 2016-02-04 at 6 03 55 pm

### Risks
 - None

### DIFF
--- a/app/views/deploy_groups/_deploy_group_select.html.erb
+++ b/app/views/deploy_groups/_deploy_group_select.html.erb
@@ -46,3 +46,16 @@
     <% end %>
   </div>
 </div>
+
+<% if BuddyCheck.enabled? %>
+  <div class="form-group">
+    <div class="col-sm-offset-2 col-sm-10">
+      <div class="checkbox">
+        <%= form.label :bypass_buddy_check do %>
+          <%= form.check_box :bypass_buddy_check %>
+          Bypass buddy check
+        <% end %>
+      </div>
+    </div>
+  </div>
+<% end %>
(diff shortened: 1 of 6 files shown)


sandlerr:
can we error out if that checkbox is checked and no production stage is chosen? it will help to avoid people misunderstanding what this is for

grosser:
screen shot 2016-02-04 at 8 40 06 pm

grosser:
looks good ?

app/views/deploy_groups/_deploy_group_select.html.erb
```diff
@@ -46,3 +46,16 @@
     <% end %>
   </div>
 </div>
+
+<% if form.object.deploy_would_require_approval? %>
```
sandlerr: won't this be hidden when creating a new stage?

grosser: you and your 'logic' ... now I have to rework that again :D

grosser:
@sandlerr all fixed up, looks good ?

jonmoter:
I think we need to make Compliance aware of this.  If they see some UI that says "Bypass buddy check", and they didn't know it exists, that could cause problems with auditors.

grosser:
true, I'll shoot an email their way

rikeoka:
Can we change the text inline 54 of app/views/deploy_groups/_deploy_group_select.html.erb to something like "Bypass buddy check (Does not deploy code)"?

grosser:
`Does not deploy code / disable buddy check`

grosser:
updated + ok-ed by compliance, good to go @sandlerr / @jonmoter  ?

sandlerr:
:+1:

PR #739, opened 2016-02-12
Remove CR in CRLF line endings

bquorning:
Without this change, a CRLF line ending would cause two tokens to be created (one for the line, and one for the LF). With this change, that same line creates just one token.

In the EventStream that my browser is currently receiving from Samson, every second line reads “append `{"msg":"<span class=\"ansible_none\">\n</span>"}`”. Since following the deploy log on large deploys is rather slow, I’d prefer to receive fewer events. Say, around 50% :-)

cc @henders @steved @jonmoter

### DIFF
--- a/app/models/terminal_output_scanner.rb
+++ b/app/models/terminal_output_scanner.rb
@@ -26,7 +26,8 @@ class TerminalOutputScanner
   private
 
   def write(data)
-    data.scrub.scan(/\r?[^\r]*/).each do |part|
+    clean_data = data.scrub.gsub("\r\n", "\n")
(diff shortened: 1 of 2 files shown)


grosser:
:+1:

bquorning:
@henders, @jonmoter, @sandlerr A review, please?

dasch:
:+1: but maybe add a comment.

henders:
Cool! :+1: 

I was initially concerned that you were going to just strip all \r's which would have screwed up progress output from bash scripts, but that's not the case. Nice one!

PR #740, opened 2016-02-12
CSV Report fix for deleted projects

rikeoka:
Fixed issue with deleted projects or stages that caused the report to run into returning attributes of a non-existant object.

The deploy report fix may need to be applied to the UI/json responses where deleted stages causes certain pages to not load.

Adjusted tests to make sure row counts don't include deleted projects/stages.


henders:
:+1: and bonus points for squashing commits :smile:

PR #744, opened 2016-02-16
If command is local, allow project admins to modify.

henders:
Allow a project admin to edit the local project commands.

/cc @zendesk/samson 

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/app/controllers/admin/commands_controller.rb
+++ b/app/controllers/admin/commands_controller.rb
@@ -1,5 +1,8 @@
 class Admin::CommandsController < ApplicationController
-  before_action :authorize_admin!
+  include ProjectLevelAuthorization
+
+  before_action :authorize_admin!, except: [ :update, :edit ]
+  before_action :find_command, only: [ :update, :edit ]
 
   def index
     @commands = Command.order('project_id').page(params[:page])
@@ -22,13 +25,7 @@ class Admin::CommandsController < ApplicationController
     end
   end
 
-  def edit
-    @command = Command.find(params[:id])
-  end
-
   def update
-    @command = Command.find(params[:id])
-
     if @command.update_attributes(command_params)
       successful_response('Command updated.')
     else
@@ -64,5 +61,10 @@ class Admin::CommandsController < ApplicationController
 
       format.json { render json: {} }
     end
+  end
+
+  def find_command
+    @command = Command.find(params[:id])
+    unauthorized! unless current_user.is_admin? || current_user.is_admin_for?(@command.project)
   end
 end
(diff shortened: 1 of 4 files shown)


sbrnunes:
Sorry, another permission tweak that was missed. :+1:

henders:
@sbrnunes : Have another look, I had missed some areas around the command pages as well.

app/controllers/admin/commands_controller.rb
```diff
@@ -1,5 +1,8 @@
 class Admin::CommandsController < ApplicationController
-  before_action :authorize_admin!
+  include ProjectLevelAuthorization
```
sbrnunes: Hummm, this CommandsController should get split between admin (gor global commands) and project specific one.
In the meantime, would it be better to have:

test/controllers/admin/commands_controller_test.rb
```diff
@@ -137,4 +137,49 @@
...
+        assert_response :ok
+      end
+
+      it 'does not update for local command as non-project-admin' do
+        patch :update, id: commands(:echo).id, command: attributes, format: :json
+        @unauthorized.must_equal true, "Request was not marked unauthorized"
+      end
+    end
+  end
 end
```
sbrnunes: I would be nice to have a few extra tests checking for:
- project admins changing global commands (which only system-level admins can change, I think) 
- system level admins changing project specific commands

But I understand if you prefer to do it later to get this fixed ASAP.

sbrnunes:
:+1 after comments have been reviewed

grosser:
:+1: not sure why we need this since they should be able to edit inline, but still a great consistency fix!

test/controllers/admin/commands_controller_test.rb
henders: There are tests there for "project admins changing global commands" which I verify is unauthorized. And there are existing tests for system level admins.

henders:
@grosser , even editing inline needed this because the inline-edit box makes a POST to this admin controller.

app/controllers/admin/commands_controller.rb
henders: Hmm, good call.

test/controllers/admin/commands_controller_test.rb
sbrnunes: Oh, sorry, I took a very brief look but didn't see it. My bad.

app/controllers/admin/commands_controller.rb
henders: Ah, can't seem to do that, seems the before_actions override each other in some weird way. When I tried the above, the before_action for the :index, :create, :destroy was removed and let anyone through.

PR #746, opened 2016-02-23
Enable Webhooks for Pull Requests

mwerner:
This PR will add functionality to the github webhook. It expands the types of webhook types Samson will respond to beyond Code Push. It uses the `ChangeSet` namespace to establish a common protocol to be a webhook handler. The `GithubController` then delegates the method calls to the webhook handler, and uses the returned info to decide how to respond.

1. Adding a "Any Pull Request" webhook will start watching webhooks for pull requests, and the issue comments present on those pull requests.
2. I have hardcoded the requirement of `[samson]` to be present in the body of either the Pull Request or Issue Comment.
3. If the base for the pull request is `production` Samson will ignore the `[samson]` requirement.

@zendesk/samson


grosser:
can you elaborate on what this will be used for / what this enabled or fixes ?

mwerner:
The intent is to be able to build and upload assets on Pull requests that are deemed by those involved as something worth testing. Currently they need to copy their sha to Samson. I am aiming to expedite that process by watching for an explicit `[samson]` reference.

app/models/changeset/pull_request.rb
```diff
@@ -3,6 +3,8 @@ class Changeset::PullRequest
   CODE_ONLY = "[A-Z][A-Z\\d]+-\\d+"  # e.g., S4MS0N-123, SAM-456
   PUNCT = "\\s|\\p{Punct}|~|="
 
+  WEBHOOK_FILTER = /(\[)\s*(samson)\s*(\])/i # [samson]
```
grosser: how about `/\[\s*samson\s*\]/i` ?

app/models/changeset/issue_comment.rb
```diff
@@ -0,0 +1,36 @@
...
+    comment = params[:comment] || {}
+    comment[:body] =~ Changeset::PullRequest::WEBHOOK_FILTER
+  end
+
+  def sha
+    pull_request.sha
+  end
+
+  def branch
+    pull_request.sha
```
grosser: use `pull_request.branch` ?

app/models/changeset/pull_request.rb
```diff
@@ -31,6 +33,17 @@ def self.find(repo, number)
...
 
+  def self.changeset_from_webhook(project, params = {})
+    find(project.github_repo, params[:number])
+  end
+
+  def self.valid_webhook?(params)
+    data = params[:pull_request] || {}
+    return false unless data[:state] == 'open'
+
+    data[:body] =~ WEBHOOK_FILTER || data.fetch(:base, {})[:ref] == 'production'
```
grosser: production PRs are always valid ?

test/controllers/integrations/github_controller_test.rb
```diff
@@ -31,28 +32,57 @@
...
+    end
 
-      hmac = OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha1'), 'test', payload.to_param)
-      request.headers['X-Hub-Signature'] = "sha1=#{hmac}"
+    does_not_deploy 'without "[samson]" in the body' do
+      payload.deep_merge!(pull_request: {body: 'imafixwolves'})
     end
 
-    test_regular_commit "Github", no_mapping: { ref: 'refs/heads/foobar' }, failed: false
+    # test_regular_commit "Github", no_mapping: { ref: 'refs/heads/foobar' }, failed: false do
```
grosser: delete or uncomment ?

app/models/changeset/pull_request.rb
grosser: the description said it would ignore production ... but this does the opposite ?
 - missing test ?
 - will cause problems when using other branches like staging to pick commits ?

grosser:
so when I push multiple times, this will run multiple times ?

sounds like a useful feature ...

app/models/changeset/pull_request.rb
mwerner: I meant in the description that it would ignore the "requirement" of having [samson] present. However, the more I think about this the more I'm thinking we should always require an explicit `[samson]` to be present. Thoughts @shajith?

mwerner:
This is separate from the code push webhook.

In a hypothetical example:
You can open a PR with `[samson]` in the body of the PR description and have your webhook run. Push a few commits and discuss various parts of the code without triggering any webhooks. Then write an issue comment with `[samson]` in it and fire your webhook with the latest sha on the PR.

app/models/changeset/pull_request.rb
shajith: yeah i'm for requiring an explicit user action to kick this off

mwerner:
I noticed some behavior that was slightly different than I expected, but I think I'm in favor of. If you put `[samson]` in your PR description, every subsequent code push will cause a pull request sync webhook which will pass the `[samson]` in the PR body test, causing a deploy.

If you only include `[samson]` in a comment, it will trigger a webhook once for the latest sha on that pull request, but subsequent code pushes will not cause anything.

This could be useful for developers that want to cause all code pushes on a PR to be deployed using `[samson]` in the PR description or simply deploy once using a PR comment with `[samson]`

I'm open to either way, but as it stands I left the behavior as is described here.

mwerner:
Was discussing the usage of `[samson]`. Should we use something that prevents us from accidentally hitting the webhook on a markdown comment mentioning [samson](https://github.com/zendesk/samson/pull/746)?

steved:
charon does `(^|\s)\[charon review\]($|\s)` as the regex match

steved:
wfm :+1:

test/models/changeset/pull_request_test.rb
```diff
@@ -29,6 +30,19 @@
     end
   end
 
+  describe ".changeset_from_webhook" do
+    before do
+      Rails.cache.clear
```
steved: Where does this come from?

mwerner: Leftover from when the `changeset_from_webhook` was making a github api call. thanks.

mwerner:
Decided to follow Charon's pattern using `[samson review]` I didn't want to use "deploy" or "publish" since those are project specific terms. I'm good with "review" and this will keep things consistent.

PR #747, opened 2016-02-24
Move Deploy CSV extract into a background job

rikeoka:
* The deploys report was failing to generate realtime with a large number of records.  The report is now generated in a background job.  The stored files is cleaned up the next time a CSV Job is run, when it is older than the lessor of 12 hours after download or 1 month after creation.

* UI changes.
Header:
screen shot 2016-03-10 at 5 50 01 pm

Recent Deploys:
screen shot 2016-02-24 at 11 50 24 am

New CSV pages:
/csv
screen shot 2016-02-24 at 11 50 59 am

/csv/new
screen shot 2016-02-24 at 11 51 05 am

/csv/status/:id
screen shot 2016-02-24 at 11 51 13 am
screen shot 2016-02-24 at 11 52 08 am

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
N/A

### Risks
- Level: Med


rikeoka:
The aging for cleanup_downloaded in /app/jobs/csv_export_job should probably go into a config, but couldn't figure out which file I should add the config to.

app/controllers/csv_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def index
+    respond_to do |format|
+      @csv_exports = CsvExport.where(user_id: current_user.id)
+      format.html
+      format.json do
+        if @csv_exports.any?
+          render json: @csv_exports
+        else
+          render json: {status: "not-found"}.to_json, :status => 404
```
grosser: extract into `not_found` method ?

app/controllers/csv_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+        else
+          render json: @csv_export
+        end
+      end
+    end
+  end
+
+  def download
+    if CsvExport.where(id: params[:id]).any?
```
grosser: should 404 when not found ?

grosser: also strange to use where and then a regular find

app/controllers/csv_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      csv_export = CsvExport.find(params[:id])
+      sent = false
+      if csv_export.finished? or csv_export.downloaded?
+        filename = "#{Rails.root}/export/#{csv_export.id}"
+        if File.exist?(filename)
+          send_file filename, type: :csv, filename: csv_export.filename
+          csv_export.downloaded!
+          sent = true
```
grosser: this seems strange ... every branch should either send or render something ?

app/controllers/csv_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def download
+    if CsvExport.where(id: params[:id]).any?
+      csv_export = CsvExport.find(params[:id])
+      sent = false
+      if csv_export.finished? or csv_export.downloaded?
+        filename = "#{Rails.root}/export/#{csv_export.id}"
+        if File.exist?(filename)
+          send_file filename, type: :csv, filename: csv_export.filename
+          csv_export.downloaded!
```
grosser: smells like internal state that should be handled in the export ... 
```Ruby
if file = ecv_export.download
  send file
else
   404
end

def download
  if finished?
    ...
  else
    deleted!
  end
end
```

app/models/csv_export.rb
```diff
@@ -0,0 +1,80 @@
...
+  def started?
+    return  status == 'started'
+  end
+
+  def finished?
+    return  status == 'finished'
+  end
+
+  def downloaded?
+    return  status == 'downloaded'
```
grosser: no need for returns anywhere

app/models/user.rb
```diff
@@ -13,6 +13,7 @@ class User < ActiveRecord::Base
   has_many :locks, dependent: :destroy
   has_many :user_project_roles, dependent: :destroy
   has_many :projects, through: :user_project_roles
+  has_many :CsvExports
```
grosser: `:csv_exports`

grosser: `, dependent: :destroy`

app/jobs/csv_export_job.rb
```diff
@@ -0,0 +1,86 @@
+class CsvExportJob < ActiveJob::Base
+  require 'csv'
```
grosser: requires go at the top of the file

app/jobs/csv_export_job.rb
```diff
@@ -0,0 +1,86 @@
...
+  
+  queue_as :csv_jobs
+ 
+  def perform(csv_export_id)
+    ActiveRecord::Base.connection_pool.with_connection do
+      csv_export = CsvExport.find(csv_export_id)
+      create_export_folder
+
+      export_task(csv_export) unless csv_export.nil?
```
grosser: no need for nil check, it will blow up when not found

app/jobs/csv_export_job.rb
```diff
@@ -0,0 +1,86 @@
...
+      cleanup_downloaded
+    end
+  end
+  
+  private
+  
+  def cleanup_downloaded
+    # clean up downloaded files older than 12 hours or any jobs stuck for over 1 month
+    @csv_exports = CsvExport.where("(status = 'downloaded' AND updated_at <= :end_date) OR updated_at <= :timeout_date",
+      {end_date: (Time.now - Rails.application.config.samson.export_job.downloaded_age),
```
grosser: this reads a bit hard ... maybe assign some variables ?

app/jobs/csv_export_job.rb
```diff
@@ -0,0 +1,86 @@
...
+      else
+        csv_export.failed!
+      end
+      
+      export_completed_notify(csv_export)
+    rescue
+      delete_file(get_filename(csv_export))
+      csv_export.failed!
+      Rails.logger.info("Export #{csv_export.filename} failed");
+    end
```
grosser: end / begin can be removed

app/jobs/csv_export_job.rb
```diff
@@ -0,0 +1,86 @@
...
+        deploy_csv_export(filename)
+      else
+        csv_export.failed!
+      end
+      
+      export_completed_notify(csv_export)
+    rescue
+      delete_file(get_filename(csv_export))
+      csv_export.failed!
+      Rails.logger.info("Export #{csv_export.filename} failed");
```
grosser: .error

app/views/deploys/recent.html.erb
```diff
@@ -29,7 +29,7 @@
   <% end %>
   <section class="timeline-content clearfix">
     <div class="deploy-csv">
-      <%= link_to "Download as CSV", recent_deploys_path(format: "csv") %>
+    	<%= button_to "Download as CSV", { controller: "csv", action: "create" }, method: :post, params: {content: 'deploys'} %>
```
grosser: use a path helper ... here `csv_path`

app/views/shared/_header.html.erb
```diff
@@ -49,6 +49,9 @@
             <%= content_tag(:span, '{{ currentActiveDeploys }}', class: 'badge badge-deploys', style: 'display: none', id: 'currentDeploysBadge') %>
           <% end %>
         </li>
+        <li class="<%= 'active' if current_page?(controller: '/csv', action: 'index') %>">
+          <%= link_to "CSV", csv_index_path %>
```
grosser: that should be `csvs_path` ?

config/application.rb
```diff
@@ -86,6 +86,11 @@ class Application < Rails::Application
     }
 
     config.action_controller.action_on_unpermitted_parameters = :raise
+    
+    config.active_job.queue_adapter = :sucker_punch
+    config.samson.export_job = ActiveSupport::OrderedOptions.new
+    config.samson.export_job.downloaded_age = ENV['EXPORT_JOB_DOWNLOADED_AGE'] || 12.hours
```
grosser: env variables are strings ... so need to convert to integers or math above will blow up

`(ENV['EXPORT_JOB_DOWNLOADED_AGE'] || 12.hours).to_i`

config/routes.rb
```diff
@@ -67,7 +67,7 @@
       get :recent
     end
   end
-
+
```
grosser: trailing whitespace

config/routes.rb
```diff
@@ -81,6 +81,10 @@
 
   get '/login', to: 'sessions#new'
   get '/logout', to: 'sessions#destroy'
+  
+  resources :csv, only: [ :index, :new, :create ]
```
grosser: `resources :csvs, only: [:index, :new, :create, :show, :download]`

... maybe download can go away ... json/html = status and when requesting with .csv then it's a download ?

db/migrate/20160218161746_add_csv_export.rb
```diff
@@ -0,0 +1,12 @@
+class AddCsvExport < ActiveRecord::Migration
+  def change
+    create_table :csv_exports do |t|
+      t.integer :user_id,    null: false
+      t.timestamps           null: false
+      t.timestamp :deleted_at
```
grosser: do we need to keep track of when they were deleted ? ... just deleted them ?

db/migrate/20160218161746_add_csv_export.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.string :content,        null: false
```
grosser: what is content ?

db/migrate/20160218161746_add_csv_export.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.string :status
+      t.string :filename
```
grosser: get rid of filename and make the filename the id ?

test/controllers/csv_controller_test.rb
```diff
@@ -0,0 +1,198 @@
...
+        it "renders the template" do
+          assert_template :index
+        end
+      end
+        
+      describe "as JSON for a user with no jobs" do
+        let(:format) { :json }
+        
+        it "renders json" do
+          assert_equal "application/json", @response.content_type
```
grosser: index actions should never return 404 ...

test/controllers/csv_controller_test.rb
```diff
@@ -0,0 +1,198 @@
...
+
+      describe "as html for a user with no jobs" do
+        let(:format) { :html }
+      
+        it "renders the template" do
+          assert_template :index
+        end
+      end
+        
+      describe "as JSON for a user with no jobs" do
```
grosser: test with jobs ?

test/controllers/csv_controller_test.rb
```diff
@@ -0,0 +1,198 @@
...
+        
+        it "renders json" do
+          assert_equal "application/json", @response.content_type
+          assert_response :ok
+          @response.body.must_equal csv_exports(:pending).to_json
+        end
+      end
+    end
+    
+    describe "a GET to :deleted for started" do
```
grosser: `a GET to status with a deleted export` ?

test/controllers/csv_controller_test.rb
```diff
@@ -0,0 +1,198 @@
+require_relative '../test_helper'
+
+describe CsvController do
```
grosser: missing permission tests ... either structure like this (lowest group first, then test as soon as it is allowed) ... or as_a_viewer/dev/admin per action

```
as_a_viewer do
  index should be forbidden
  post should be forbidden
  status should return x
end

as_a_deployer do
  index should be forbidden
  post should create an export
end

as_a_admin do
  index should return xyz
end
```

test/fixtures/csv_exports.yml
```diff
@@ -0,0 +1,35 @@
+pending:
```
grosser: no need for all these fixtures, change the fixture when used ?

grosser:
- good idea / direction! :heart: 
 - lots of little details we should iron out :)
 - fork the zendesk repo, you should have push access, this way anyone can check out your branch more easily

db/migrate/20160218161746_add_csv_export.rb
rikeoka: To identify the type of export.  Was leaving room to move users csv export if that ever becomes thousands of lines and too slow to generate real-time.  With the move to project level authorizations and more deployers, could reach a scenario where the report generates several thousand row entries that may require asynchronous exports.

db/migrate/20160218161746_add_csv_export.rb
rikeoka: Was using this originally, but code climate was complaining about the filename coming from a model attribute with user input.  Switched it to use the id so its safer, and kept it to send a useful filename of the report.  It could have the report generated and completed time appended to the report, or use created_at in the filename.

henders:
Might be also a good idea to move the 'CSV' menu item under the 'Admin' menu... it's not really a top-level functionality for daily-use

rikeoka:
@henders 

Moved the CSV menu item to Admin as Reports.  Reports menu item would need to be visible to all users, unless it should only be for Admins.  The items above the spacer row, will be hidden for non-admins.

screen shot 2016-03-11 at 1 38 20 pm

test/controllers/csv_controller_test.rb
rikeoka: This is currently set to allow anyone with access to generate and download a report.  Not sure what level of permissions is necessary.  

There may be use cases where a viewer should be able to get the report, so they can review the deploys data without being able to deploy.

rikeoka:
No longer the current branch.  Current brnach is /zendesk/samson/rikeoka/asynchronous_csv.  New PR is https://github.com/zendesk/samson/pull/766.

PR #748, opened 2016-02-25
Clean up docker images, add 'latest' tag always, and update docker gem.

henders:
This is to primarily add the 'latest' tag to all docker image pushes up to the registry. Also added a cleanup of docker images which has plagued our docker build hosts, and also updated the gem we use to support newer versions of docker.

/cc @zendesk/samson  

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: 

### Risks
- Level: Low
- Notes for Support: None
- Notes for QA: None


andreionut:
:+1: on green build

PR #752, opened 2016-03-02
command search + project-id filter

grosser:
- search for text inside of commands, for example deprecated calls
 - filter by project to clean it out

/cc @zendesk/samson

screen shot 2016-03-01 at 10 04 15 pm
 

### Risks
- None

### DIFF
--- a/app/views/admin/commands/index.html.erb
+++ b/app/views/admin/commands/index.html.erb
@@ -1,5 +1,20 @@
 <h1>Commands</h1>
 
+<%= form_tag '?', method: :get, class: 'clearfix' do %>
+  <div class="col-sm-6" style="padding-left: 0;">
+    <%= label_tag "Search" %>
+    <%= text_field_tag 'search[query]', params[:search].try(:[], :query), class: "form-control" %>
+  </div>
+  <div class="col-sm-3">
+    <%= label_tag "Project" %><br/>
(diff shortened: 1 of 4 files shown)


grosser:
@zendesk/runway

sandlerr:
are there too many commands to do a fuzzy search in JS?

sandlerr:
perhaps something like https://github.com/atom/fuzzaldrin

grosser:
we have 1500 ... don't want to load all this onto the page ... if we do fuzzy search then I'd prefer it on the backend

app/views/admin/commands/index.html.erb
```diff
@@ -1,5 +1,20 @@
 <h1>Commands</h1>
 
+<%= form_tag '?', method: :get, class: 'clearfix' do %>
+  <div class="col-sm-6" style="padding-left: 0;">
+    <%= label_tag "Search" %>
+    <%= text_field_tag 'search[query]', params[:search].try(:[], :query), class: "form-control" %>
+  </div>
+  <div class="col-sm-3">
+    <%= label_tag "Project" %><br/>
+    <% options = options_for_select(Project.pluck(:name, :id), params[:search].try(:[], :project_id)) %>
```
sandlerr: we should go to Ruby 2.3, then you can `params.dig(:search, :project_id)`

sandlerr:
:+1: as is, we can add fuzzy later

PR #753, opened 2016-03-02
[INFR-1683] pass env vars to Docker build container

msufa:
Pass environment variables configured for a project to the build container.

This PR is just a first naive approach and all comments and suggestions are welcome.

Open issues and considerations:
- With this PR `docker_binary_builder` plugin now depends on `env` plugin, though the builds should still work even if the `env` plugin is not enabled, no env vars will be passed in such scenario.
- We are passing project level env vars, because it is possible to spin up a build outside a deploy with no environment / stage / deploy group associated.
- We will pass in **all global** environment variables configured for a project (configured for all environments and stages)

/cc @zendesk/samson @andreionut @fneves @henders 

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/INFR-1683

### Risks
- Level: Low

### DIFF
--- a/plugins/docker_binary_builder/test/models/binary_builder_test.rb
+++ b/plugins/docker_binary_builder/test/models/binary_builder_test.rb
@@ -48,6 +48,7 @@ describe BinaryBuilder do
         {
           'Cmd' => ['/app/build.sh'],
           'Image' => 'foo_build:abc-19f',
+          'Env' => [],
           'Volumes' => { '/opt/samson_build_cache' => {} },
           'HostConfig' => {
             'Binds' => ['/opt/samson_build_cache:/build/cache'],
@@ -63,6 +64,7 @@ describe BinaryBuilder do
         {
           'Cmd' => ['/app/build.sh'],
           'Image' => 'foo_build:abc-19f',
+          'Env' => [],
           'Mounts' => [
             {
               'Source' => '/opt/samson_build_cache',
@@ -76,6 +78,14 @@ describe BinaryBuilder do
           }
         }
       )
+    end
+
+    it 'sets up global environment variables for the build container' do
+      project.environment_variables.create!(name: 'FIRST', value: 'first')
+      project.environment_variables.create!(name: 'SECOND', value: 'second')
+      project.environment_variables.create!(name: 'THIRD', value: 'third',
+                                            scope_type: 'Environment', scope_id: environments(:production).id)
+      builder.send(:create_container_options)['Env'].must_equal %w(FIRST=first SECOND=second)
     end
 
     it 'throws exception with api < 1.15' do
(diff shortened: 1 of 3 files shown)


fneves:
looks good.
Does the env plugin allow us to override environmental specific properties per stage?

andreionut:
lgtm :+1: Let's give it a spin

msufa:
@fneves I assume it will let us override per stage if the build is started in the context of a deployment. I also assume it should work out of the box. If we bump into problems along the way we can tweak the select to take the stage into account.

msufa:
I had another look and unfortunately this will not work correctly with considerations to environments and stages. I need to get env vars using `EnvironmentVariable.env` which requires a project and a deploy group. We could use a nil group for standalone builds but we need to figure out how to handle multiple deploy groups per a deployment job. I'll look into this and will update the PR shortly.

jonmoter:
I just made a comment in INFR-1683 that given our current Vault implementation, Samson shouldn't be able to read/write to Vault.

msufa:
I've updated the implementation to take all global env vars configured for a project (the ones with `All` set in the env var combo box) as well as global env vars configured in groups belonging to a project. There's no way to configure the vars per stage currently as there's no such concept when running a build outside a deploy. @fneves @andreionut @henders please have another look.
@jonmoter as far as Vault goes we'll go with the current approach until we have that working, let's continue the discussion in INFR-1683.

andreionut:
I don't see the new code

msufa:
It was a force push @andreionut. The main thing that changed is that I'm using `EnvironmentVariable.env` now instead of going through the project object. I was able to see the new code after refreshing the webpage.

andreionut:
:+1: lgtm, but I'm not a Samson expert

plugins/docker_binary_builder/test/models/binary_builder_test.rb
```diff
@@ -78,6 +80,14 @@
       )
     end
 
+    it 'sets up global environment variables for the build container' do
+      project.environment_variables.create!(name: 'FIRST', value: 'first')
+      project.environment_variables.create!(name: 'SECOND', value: 'second')
+      project.environment_variables.create!(name: 'THIRD', value: 'third',
+                                            scope_type: 'Environment', scope_id: environments(:production).id)
+      builder.send(:create_container_options)['Env'].must_equal %w(FIRST=first SECOND=second)
```
fneves: isn't this missing Third?

msufa: That's the actual test, it should pass only global variables and ignore the ones that have scope defined :)

fneves:
:+1:

PR #755, opened 2016-03-03
show deploy groups when they are set

grosser:
Makes it obvious where the deploy is going / if $DEPLOY_GROUPS is set up correctly,
same info as shown on index

@zendesk/runway 

screen shot 2016-03-02 at 5 17 19 pm

### Risks
- None


jonmoter:
Nice addition!  :+1:

PR #757, opened 2016-03-10
Fix typo on command model

tuanmai:
Just fix typo in `command` model


henders:
Huh, good catch. Guess we don't use that relationship! :+1:

tuanmai:
Yeah, I also did find that too :D. I just play around with source code and try to contribute to this project ;)

PR #760, opened 2016-03-16
remove and restore StageCommands so usage numbers match on admin/comm…

grosser:
…ands and commands/edit

atm admins/commands shows that a command is used, but clicking on edit shows no usage ...

screen shot 2016-03-16 at 4 43 52 pm

@zendesk/runway 

### Risks
- Level: Low ... when a command is deleted restoring the stage would result in an invalid StageCommand


sandlerr:
:+1:

PR #763, opened 2016-03-17
INFR-1737 no-release stages do not produce releases / are invisible from dashbo…

grosser:
…ards

atm build stages show in the dashboards, which they should not, since they are not code releases.
This introduces a `release` flag for the deploy, so we know it is a proper code deploy and not some random other tasks like building assets

... in the next PR: kill macros and make them stages that do not generate releases

https://zendesk.atlassian.net/browse/INFR-1737

@zendesk/runway 

screen shot 2016-03-17 at 2 28 12 pm

### Risks
 - Low: deploys failing because db was not migrated
 - Low: users being confused because build stages are missing from the dashboards

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -98,7 +98,7 @@ class Stage < ActiveRecord::Base
   end
 
   def create_deploy(user, attributes = {})
-    deploys.create(attributes) do |deploy|
+    deploys.create(attributes.merge(release: !bypass_buddy_check)) do |deploy|
       deploy.build_job(project: project, user: user, command: command)
     end
   end
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -98,7 +98,7 @@ class Stage < ActiveRecord::Base
   end
 
   def create_deploy(user, attributes = {})
-    deploys.create(attributes.merge(release: !bypass_buddy_check)) do |deploy|
+    deploys.create(attributes.merge(release: !no_code_deployed)) do |deploy|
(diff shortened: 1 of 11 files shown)


zendesknotifications:
&nbsp;<b>Code Climate</b> has <a href="https://codeclimate.com/repos/53340bef6956800b9000675c/pull/763">analyzed this pull request</a>.

grosser:
@sandlerr can you review this one too ?

app/models/stage.rb
```diff
@@ -98,7 +98,7 @@ def confirm_before_deploying?
   end
 
   def create_deploy(user, attributes = {})
-    deploys.create(attributes) do |deploy|
+    deploys.create(attributes.merge(release: !no_code_deployed)) do |deploy|
```
sandlerr: or even `attributes.merge(release: !attributes.delete(:no_code_deployed))`

sandlerr:
I'm not sure if these really are the same thing - this is wrapped in `BuddyCheck.enabled?` but now it provides a function that could be useful for orgs that don't use the buddy check. Do you think it could also have some use outside of production?

app/models/stage.rb
grosser: Why would we need to pass in the `:no_code_deployed` ... a stage is always the same command, so it's always either a deploy or not a deploy

grosser:
unwrapped from BuddyCheck ... since it's useful with and without.

sandlerr:
:+1: on green

PR #765, opened 2016-03-18
[DATAM-226]update docker binary plugin to have additional option to skip untar

waichee:
:ghost:

Add a project settings on the admin screen to allow customisation of whether or not to untar the artifact binary created from docker container when using the docker binary builder plugin..

image

/cc @zendesk/samson, @henders, @zendesk/bunyip  

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/DATAM-226

### DIFF
--- a/db/migrate/20160318170635_projects_add_extract_docker_binary.rb
+++ b/db/migrate/20160318170635_projects_add_extract_docker_binary.rb
@@ -1,0 +1,5 @@
+class ProjectsAddExtractDockerBinary < ActiveRecord::Migration
+  def change
+    add_column :projects, :extract_docker_packaged_artifact, :boolean, default: true, null: false
+  end
+end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -90,6 +90,7 @@ class ProjectsController < ApplicationController
         :permalink,
         :release_branch,
         :deploy_with_docker,
+        :extract_docker_packaged_artifact,
         :auto_release_docker_image,
         stages_attributes: stage_permitted_params
       ] + Samson::Hooks.fire(:project_permitted_params)
--- a/app/views/projects/_form.html.erb
+++ b/app/views/projects/_form.html.erb
@@ -59,6 +59,10 @@
           <%= form.check_box :deploy_with_docker %>
           <%= form.label :deploy_with_docker, 'Deploy with Docker' %>
         </div>
+        <div class="checkbox">
+          <%= form.check_box :extract_docker_packaged_artifact %>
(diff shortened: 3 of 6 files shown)


grosser:
nobody is using this atm ... so make it the default if that's what we want to use ?

waichee:
@grosser  I think the Legion projects are using the plugin and they might like to continue un-tarring the artifact on the deploy server as part of their build and deployments.

grosser:
ENV is global .. so they cannot do that ... needs to be a per-project setting then

waichee:
Oh ya true, I'll change it to project settings

fneves:
Legion, Doorman and Pigeon projects are using this. The current behaviour should be kept, otherwise things will get broken. @grosser I did not understand the part of ENV is global.

waichee:
@fneves current behavior should be kept. the default settings is to keep the current behavior of extracting the artifact.tar once its pulled back to the deploy server from the container. it will only change for projects that specifically unchecked the option on the project settings.

db/migrate/20160318170635_projects_add_extract_docker_binary.rb
```diff
@@ -0,0 +1,5 @@
+class ProjectsAddExtractDockerBinary < ActiveRecord::Migration
+  def change
+    add_column :projects, :extract_docker_packaged_artifact, :boolean, default: true, null: false
```
fneves: when the migration is ran, what will happen to the current existing projects? Are previous values backfilled? if not won't null mean false, causing the default value of the flag to be false?

henders:
The code shouldn't be in the core Samson code considering it only affects functionality added by a plugin. You should move this stuff to the docker_binary_build plugin.

db/migrate/20160318170635_projects_add_extract_docker_binary.rb
waichee: locally when the migration ran, the existing projects were set to 1 and seems to retain the existing behavior of extracting the tar file.

waichee:
Agreed that this should be in docker binary build plugin rather than core. 

After some thoughts,  I realized that this could be achieve by wrapping the tar in the container itself via the existing build shell script of the docker binary plugin and does not require any code change to samson nor the binary plugin for this purpose. 

The initial reason that the skipping of untar was required as I was using it for python virtual envs packaging (with a lot of files and binaries) rather than an uber jar. The resulting symlinks and long links wasn't able to be handled appropriately by the untar in the plugin.

I'll close this PR and not merge it.

PR #767, opened 2016-03-20
update docker binary plugin to support running an optional script bef…

waichee:
:ghost:

Update docker binary plugin to allow an optional script to be run before starting the build of the docker image. The reason that the prebuild is required is because for building python apps with 'closed source' dependencies, pip does not work within the container as we do not have access to the closed source github repo. Thus pre build step is used for checking out the relevant dependencies on the deploy server before triggering the docker binary build.

As the pre-build script is optional, it should not affects existing projects that does not require hooking into the stage before the binary build.

/cc @zendesk/samson @zendesk/bunyip

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: 

### Risks
- Level: None

### DIFF
--- a/plugins/docker_binary_builder/README.md
+++ b/plugins/docker_binary_builder/README.md
@@ -24,7 +24,7 @@ If the directory '/opt/samson_build_cache' exists on the Docker host, it will mo
 docker build image. That way you could then instruct Maven to use '/build/cache/.m2' as the cache directory for all your 
 projects.
 
-You can also provide a script named 'pre_binary_build.sh' to be ran, before the docker binary plugin starts building the image.
+You can also provide a script named 'pre_binary_build.sh' to be ran before the docker binary plugin starts building the image.
 
 The build container will also receive all global (`All` selected in the env var's combo box) environment variables that are configured for a project, assuming the `env` plugin is enabled.
 
--- a/plugins/docker_binary_builder/README.md
+++ b/plugins/docker_binary_builder/README.md
@@ -16,13 +16,15 @@ This plugin makes a few assumptions:
 - You have a build script 'build.sh' that can be executed to compile the needed binaries
   - This is packaged inside the Dockerfile.build image and executed
   - After binaries are compiled the build.sh script finishes by tar'ing all necessary files into an 'artifacts.tar' file.
-  
+
 If you want to share caches between builds, e.g. all your projects are Java and use Maven to build, then sharing the .m2 
 directory will be a huge time saving tactic to share all downloaded dependencies.
 
 If the directory '/opt/samson_build_cache' exists on the Docker host, it will mount it to '/build/cache' inside the 
 docker build image. That way you could then instruct Maven to use '/build/cache/.m2' as the cache directory for all your 
 projects.
+
+You can also provide a script named 'pre_binary_build.sh' to be ran, before the docker binary plugin starts building the image.
 
 The build container will also receive all global (`All` selected in the env var's combo box) environment variables that are configured for a project, assuming the `env` plugin is enabled.
 
--- a/plugins/docker_binary_builder/app/models/binary_builder.rb
+++ b/plugins/docker_binary_builder/app/models/binary_builder.rb
@@ -5,16 +5,22 @@ class BinaryBuilder
   ARTIFACTS_FILE_PATH = "/app/#{ARTIFACTS_FILE}".freeze
   DOCKER_HOST_CACHE_DIR = '/opt/samson_build_cache'.freeze
   CONTAINER_CACHE_DIR = '/build/cache'.freeze
+  TAR_LONGLINK = '././@LongLink'.freeze
+  PRE_BUILD_SCRIPT = 'pre_binary_build.sh'.freeze
 
-  def initialize(dir, project, reference, output)
+  def initialize(dir, project, reference, output, executor = nil)
     @dir = dir
     @project = project
     @git_reference = reference
     @output_stream = output
+    @executor = executor || TerminalExecutor.new(@output_stream, verbose: true)
   end
 
   def build
     return unless @project.try(:deploy_with_docker?) && File.exists?(File.join(@dir, DOCKER_BUILD_FILE))
+
+    run_pre_build_script
+
     @output_stream.puts "Connecting to Docker host with Api version: #{docker_api_version} ..."
 
     @image = create_build_image
@@ -28,6 +34,14 @@ class BinaryBuilder
     @output_stream.puts 'Cleaning up docker build image and container...'
     @container.delete(force: true) if @container
     @image.remove(force: true) if @image
+  end
+
+  def run_pre_build_script
+    file = File.join(@dir, PRE_BUILD_SCRIPT)
+    return unless File.exist?(file)
+
+    @output_stream.puts "Running pre build script..."
+    @executor.execute! file
   end
 
   private
--- a/plugins/docker_binary_builder/app/models/binary_builder.rb
+++ b/plugins/docker_binary_builder/app/models/binary_builder.rb
@@ -38,10 +38,12 @@ class BinaryBuilder
 
   def run_pre_build_script
     file = File.join(@dir, PRE_BUILD_SCRIPT)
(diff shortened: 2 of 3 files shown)


plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -30,6 +40,12 @@ def build
     @image.remove(force: true) if @image
   end
 
+  def run_pre_build_script
```
waichee: Is there an alternative way to hook onto pre-docker binary build phase?

the docker binary build seems to be triggered at start of job execution and hooking on to deploy:build in Capfile doesnt seems to be able to intercept it...

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -5,16 +5,26 @@ class BinaryBuilder
...
+  PRE_BUILD_SCRIPT = 'pre_binary_build.sh'.freeze
 
-  def initialize(dir, project, reference, output)
+  def initialize(dir, project, reference, output, executor = nil)
     @dir = dir
     @project = project
     @git_reference = reference
     @output_stream = output
+    if executor
+      @executor = executor
```
grosser: `@executor = executor || ...` :)

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -5,16 +5,26 @@ class BinaryBuilder
...
+    else
+      @executor = TerminalExecutor.new(@output_stream, verbose: true)
+    end
   end
 
   def build
     return unless @project.try(:deploy_with_docker?) && File.exists?(File.join(@dir, DOCKER_BUILD_FILE))
+
+    run_pre_build_script if File.exists?(File.join(@dir, PRE_BUILD_SCRIPT))
```
grosser: file location is used here and inside, would be drier / spread less knowledge when it is only handled inside

```
def run_pre_build_script
  file = File.join(@dir, PRE_BUILD_SCRIPT)
  return unless File.exist?(file)
  ...
end
```

plugins/docker_binary_builder/test/models/binary_builder_test.rb
```diff
@@ -42,6 +46,23 @@
...
 
+    it 'run pre build shell script if it is available' do
+      File.expects(:exists?).with(File.join(dir, BinaryBuilder::DOCKER_BUILD_FILE)).returns(true)
+      File.expects(:exists?).with(File.join(dir, BinaryBuilder::PRE_BUILD_SCRIPT)).returns(true)
+      executor.expects(:execute!).with(File.join(dir, BinaryBuilder::PRE_BUILD_SCRIPT)).once
+
+      project.update_attributes(deploy_with_docker: true)
+      builder.build
+      output.string.must_equal [
+                                 "Running pre build script...\n",
```
grosser: 2-space indent plz ... wastest less space and does not need to all be changed when first line changes

waichee: Sure. updated as per feedback. :)

waichee:
:haircut:

plugins/docker_binary_builder/README.md
```diff
@@ -16,7 +16,7 @@ This plugin makes a few assumptions:
 - You have a build script 'build.sh' that can be executed to compile the needed binaries
   - This is packaged inside the Dockerfile.build image and executed
   - After binaries are compiled the build.sh script finishes by tar'ing all necessary files into an 'artifacts.tar' file.
-  
+
```
grosser: ahhh a whitespace snuck in :D

plugins/docker_binary_builder/test/models/binary_builder_test.rb
```diff
@@ -31,9 +32,29 @@
 
     it 'builds image if docker flag is set for project and dockerfile.build exists' do
       File.expects(:exists?).with(File.join(dir, BinaryBuilder::DOCKER_BUILD_FILE)).returns(true)
+      File.expects(:exists?).with(File.join(dir, BinaryBuilder::PRE_BUILD_SCRIPT)).returns(false)
+      executor.expects(:execute!).with(File.join(dir, BinaryBuilder::PRE_BUILD_SCRIPT)).never
+
+      project.update_attributes(deploy_with_docker: true)
+      builder.build
+      output.string.must_equal [
+        "Connecting to Docker host with Api version: 1.19 ...\n",
```
grosser: maybe be less exact, this will break when we update docker or any other workflow

grosser:
don't know much about the issue / how to solve this otherwise ... but :+1: for the code ...

plugins/docker_binary_builder/test/models/binary_builder_test.rb
waichee: this should be all right as the test is stubbing the version in 
https://github.com/zendesk/samson/blob/wyau/update-docker-binary-plugin/plugins/docker_binary_builder/test/models/binary_builder_test.rb#L15

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -30,6 +36,14 @@ def build
     @image.remove(force: true) if @image
   end
 
+  def run_pre_build_script
+    file = File.join(@dir, PRE_BUILD_SCRIPT)
+    return unless File.exist?(file)
+
+    @output_stream.puts "Running pre build script..."
+    @executor.execute! file
```
sandlerr: does this kill the build if the return is non zero?

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -30,6 +36,14 @@ def build
     @image.remove(force: true) if @image
   end
 
+  def run_pre_build_script
+    file = File.join(@dir, PRE_BUILD_SCRIPT)
+    return unless File.exist?(file)
```
sandlerr: nitpick: you can use [`File.file?`](http://ruby-doc.org/core-2.2.0/File.html#method-c-file-3F) to be a bit more specific

sandlerr:
Is it possible to pack python dependencies in the repo? That is what we usually do with Ruby dependencies, although I don't think we do this with Javascript ones.

jeffreytheobald:
It might be a bit messy because the python artefacts are eggs within the virtual env and tied quite deeply to the virtual env.

Also, the intention of this step is to package up the deployment to a standalone package.  It might be a bit heavy to package up the dependencies, so we can package up the standalone package?  Not sure.

plugins/docker_binary_builder/README.md
```diff
@@ -24,6 +24,8 @@ If the directory '/opt/samson_build_cache' exists on the Docker host, it will mo
 docker build image. That way you could then instruct Maven to use '/build/cache/.m2' as the cache directory for all your 
 projects.
 
+You can also provide a script named 'pre_binary_build.sh' to be ran, before the docker binary plugin starts building the image.
```
jeffreytheobald: `to be run` (with no comma)

plugins/docker_binary_builder/app/models/binary_builder.rb
waichee: it doesnt....good pickup.
 :haircut:

plugins/docker_binary_builder/README.md
waichee: :haircut:

fneves:
@waichee, we've recently modified the binary plugin so that we are able to send in custom env properties that are specified in the environment variables plugin. Would it work if send in some env properties to the build container that would allow you to fetch those dependencies from within the container?

waichee:
@fneves you mean like passing on the deploy github credentials to the container via env?
would that be all right from security point of view?

fneves:
I was suggesting that you could put an authorised ssh key as an env variable of the project, and write that key to the .ssh folder in the docker build container. 
Similar with what people do with Travis for private gem repositories. 
Not sure if this is compliant with security though.

fneves:
:+1:

PR #770, opened 2016-03-23
no code deployed should be useable even when not using deploy groups

grosser:
can be used to bypass the production check even when not using deploy groups

@sandlerr


sandlerr:
:+1:

PR #776, opened 2016-03-28
Use stringed keys on Github webhook data

mwerner:
This is an attempt to properly assess the validity of a Github webhook. The added functionality was not parsing correctly due to the format of the incoming webhook data. After downloading a copy of the data from the log servers and testing locally, this was the most obvious potential problem.

@zendesk/samson

### Risks
- Level: Low

### DIFF
--- a/app/models/changeset/issue_comment.rb
+++ b/app/models/changeset/issue_comment.rb
@@ -13,7 +13,7 @@ class Changeset::IssueComment
 
   def self.valid_webhook?(params)
     comment = params['comment'] || {}
-    (comment['body'] =~ Changeset::PullRequest::WEBHOOK_FILTER).present?
+    !(comment['body'] =~ Changeset::PullRequest::WEBHOOK_FILTER).nil?
   end
 
   def sha
--- a/app/models/changeset/issue_comment.rb
+++ b/app/models/changeset/issue_comment.rb
@@ -12,8 +12,8 @@ class Changeset::IssueComment
   end
 
   def self.valid_webhook?(params)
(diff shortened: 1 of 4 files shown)


app/models/changeset/issue_comment.rb
```diff
@@ -12,8 +12,8 @@ def self.changeset_from_webhook(project, params = {})
   end
 
   def self.valid_webhook?(params)
-    comment = params[:comment] || {}
-    comment[:body] =~ Changeset::PullRequest::WEBHOOK_FILTER
+    comment = params['comment'] || {}
+    (comment['body'] =~ Changeset::PullRequest::WEBHOOK_FILTER).present?
```
grosser: this is a number or nil ... so `present?` does not make much sense ...

app/models/changeset/pull_request.rb
```diff
@@ -34,15 +34,15 @@ def self.find(repo, number)
...
   end
 
   def self.valid_webhook?(params)
-    data = params[:pull_request] || {}
-    return false unless data[:state] == 'open'
+    data = params['pull_request'] || {}
+    return false unless data['state'] == 'open'
 
-    data[:body] =~ WEBHOOK_FILTER
+    (data['body'] =~ WEBHOOK_FILTER).present?
```
grosser: same here

grosser:
:+1:

app/models/changeset/issue_comment.rb
mwerner: Right... was just trying to catch the `nil`, I'll fix.

PR #777, opened 2016-03-28
SAMSON-193 Fix child processes having access to all of samsons env, including se…

grosser:
https://zendesk.atlassian.net/browse/SAMSON-193

Also fixes race condition where with_clean_env might have just ended in another thread and even .env values get exposed

@jonmoter @steved @dragonfax 

/cc @zendesk/samson 

### Risks
 - High: all kinds of deploys failing

### DIFF
--- a/app/models/terminal_executor.rb
+++ b/app/models/terminal_executor.rb
@@ -1,5 +1,4 @@
-require 'pty'
-require 'shellwords'
+require 'open3'
 
 # Executes commands in a fake terminal. The output will be streamed to a
 # specified IO-like object.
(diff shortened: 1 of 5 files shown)


app/models/terminal_executor.rb
```diff
@@ -34,26 +33,26 @@ def stop!(signal)
...
-    _, status = Process.wait2(pid)
-
-    input.close
+      wait_thr.value
+    end.success?
+  end
 
-    return status.success?
+  def whitelisted_env
+    whitelist = ['TERM', 'SHELL'] + ENV.keys.select { |k| k.end_with?('PATH', 'DIR', '_HOME') }
```
jonmoter: Seems kind of arbitrary that we're keeping env vars that end with `PATH`, `DIR`, or `_HOME`.

I would rather have this whitelist be minimal and documented, with only the ENV vars that would be necessary for the commands to function.

jonmoter:
Thanks for finding a fix for this! It's been bugging me for a while now.

app/models/terminal_executor.rb
grosser: we need path / java_home / tempdir / gem_home / gem_path ... possibly gopath / nodpath etc ... so I was thinking the pattern here is all paths are fine ... either way is fine I guess, just means a few fix-up PRs will be coming soonish :)
so which one do you like more, all paths / exact whitelist ?

grosser:
updated to reduced whitelist + ENV_WHITELIST support

test/models/terminal_executor_test.rb
```diff
@@ -1,6 +1,13 @@
 require_relative '../test_helper'
 
 describe TerminalExecutor do
+  def with_env(k, v)
```
jonmoter: How about having this take a hash of new values, rather than a single k/v pair?

grosser: updates

jonmoter:
:+1:

PR #778, opened 2016-03-29
Samson 194 (adding time formatting options)

irwaters:
* Add description here
Adding ability for user to change the way times are displayed.  This currently works only on the 'Recent Deploys' tab.  If folks like this, I'll wire it up current as well
.  
Changes allow for users to set a default time display preference in their profile and have that apply, and it can be changed on the screen.  

screen shot 2016-03-28 at 5 08 49 pm
screen shot 2016-03-28 at 5 11 09 pm

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:
https://zendesk.atlassian.net/projects/SAMSON/issues/SAMSON-194

### Risks
- Level: Low/Med/High
Low


app/assets/javascripts/timeline.js
```diff
@@ -209,6 +224,20 @@ samson.controller("TimelineCtrl", function($scope, $window, $timeout, Deploys, S
   $scope.timelineDeploys = Deploys;
   $scope.deploys = Deploys.entries;
 
+  if (!$scope.selectedTimeFormat) {
```
grosser: can we render this into the page ?
`User.timezone = <%= current_user.timezone %>` or so ... to avoid another request ?

app/views/profiles/show.html.erb
```diff
@@ -22,6 +22,17 @@
           <%= form.text_field :token, class: "form-control", disabled: true %>
         </div>
       </div>
+      <div class="form-group">
+        <%= form.label :time_format, class: "col-lg-2 control-label" %>
+        <div class="col-lg-4">
+          <%= form.radio_button(:time_format, "relative") %>
+          <%= form.label(:relative, "Relative") %>
+          <%= form.radio_button(:time_format, "local") %>
+          <%= form.label(:local, "Local") %>
```
grosser: does clicking on the labels select the radio buttons ?
if not, maybe do `label { radio_button }`

app/views/deploys/recent.html.erb
```diff
@@ -25,6 +25,12 @@
           <option value="">All</option>
         </select>
       </div>
+      <div class="timeline-filter col-sm-4" >
+        <label for="tlf-status">TIME:</label>
+        	<input type=radio  ng-model="timeFormat" ng-change="changeTime()" value='relative'>relative</input>
```
grosser: strange indent ... make sure to use double-space and not tabs ...

app/models/user.rb
```diff
@@ -91,4 +94,8 @@ def project_role_for(project)
   def set_token
     self.token = SecureRandom.hex
   end
+
+  def set_time_format
+    self.time_format = 'relative'
```
grosser: can this be the column default -> remove method + allow initializing with a different format

config/routes.rb
```diff
@@ -71,6 +71,7 @@
   resources :deploy_groups, only: [:show]
 
   resource :profile, only: [:show, :update]
+  get '/profile/details', to: 'profiles#details'
```
grosser:
```Ruby
resource :profile, only: [:show, :update] do
  member do
    get :details
  end
end

```

db/migrate/20160328221058_add_time_format_to_user.rb
```diff
@@ -0,0 +1,8 @@
+class AddTimeFormatToUser < ActiveRecord::Migration
+  def change
+    add_column :users, :time_format, :string
```
grosser: `default: 'relative', null: false`

test/models/user_test.rb
```diff
@@ -26,6 +27,23 @@
     end
   end
 
+  describe "#time_format" do
+    let(:user) { User.create(name: "jimbob", email: 'test@test.com') }
+    describe 'time_format' do
+      it "should have a default time format of relative" do
+        user.time_format.must_equal('relative')
+      end
+      it "should not update and should raise" do
```
grosser: newlines between ifs plz

test/models/user_test.rb
```diff
@@ -26,6 +27,23 @@
...
+        lambda {user.update_attributes!(:time_format => 'foobar') }.must_raise ActiveRecord::RecordInvalid
+      end
+      it "should update" do
```
grosser: using active forms makes it more readable/simple `it should update` -> `it updates`

test/models/user_test.rb
```diff
@@ -26,6 +27,23 @@
...
   end
 
+  describe "#time_format" do
+    let(:user) { User.create(name: "jimbob", email: 'test@test.com') }
+    describe 'time_format' do
+      it "should have a default time format of relative" do
+        user.time_format.must_equal('relative')
+      end
+      it "should not update and should raise" do
+        lambda {user.update_attributes!(:time_format => 'foobar') }.must_raise ActiveRecord::RecordInvalid
```
grosser: not testing that it did not update ...

test/models/user_test.rb
```diff
@@ -26,6 +27,23 @@
     end
   end
 
+  describe "#time_format" do
+    let(:user) { User.create(name: "jimbob", email: 'test@test.com') }
```
grosser: use `create!` to faster see errors in tests

grosser:
is format in user profile good enough or do we also need the extra boxes in deploy view ?
a select might be less clutter/space than radio buttons

app/views/profiles/show.html.erb
irwaters: does now, everywhere.

app/views/deploys/recent.html.erb
irwaters: vimrc issues, resolved.

app/assets/javascripts/timeline.js
irwaters: is a good idea.

irwaters:
select boxes exist in deploy view (recent).  will add them to current as well if people like this (different code path etc).

app/assets/javascripts/timeline.js
```diff
@@ -96,6 +96,21 @@ samson.filter("userFilter",
...
+  function() {
+    return function(td, timeFormat) {
+      if (timeFormat == 'local') {
+        return moment(td).format('LLL');
+      } else if (timeFormat == 'utc'){
+        return moment(td).utc().format();
+      } else {
+        return moment(td).fromNow();
+      }
+
```
grosser: unnecessary newline 
can it be `else if == now ... else throw "Unknown format xyz"` to make future debugging easier

app/assets/javascripts/timeline.js
```diff
@@ -209,6 +224,17 @@ samson.controller("TimelineCtrl", function($scope, $window, $timeout, Deploys, S
   $scope.timelineDeploys = Deploys;
   $scope.deploys = Deploys.entries;
 
+
+  if (!$scope.selectedTimeFormat) {
+      $scope.selectedTimeFormat = document.getElementById("default_time_format").value;
```
grosser: `$('#default_time_format').val()`

app/assets/javascripts/timeline.js
```diff
@@ -209,6 +224,17 @@ samson.controller("TimelineCtrl", function($scope, $window, $timeout, Deploys, S
...
+      $scope.timeFormat = $scope.selectedTimeFormat;
+  }
+
+
+  $scope.changeTime = function() {
+    $scope.selectedTimeFormat = $scope.timeFormat;
```
grosser: feels like this should be `global_scope. selectedTimeFormat  = $scope.timeFormat` and not assigning on itself ... then we could simply use timeFormat everywhere ... but no idea how this angular stuff works :D

app/controllers/profiles_controller.rb
```diff
@@ -14,9 +14,13 @@ def show
     @user = current_user
   end
 
+  def details
```
grosser: can go away now ?

app/views/deploys/recent.html.erb
```diff
@@ -25,6 +26,21 @@
           <option value="">All</option>
         </select>
       </div>
+      <div class="timeline-filter col-sm-4" >
+        <label for="tlf-status">TIME:</label>
+        <label for=time_foarmat_relative>
```
grosser: <% User::TIME_FORMATS.each do %> and use it for validations too

app/views/profiles/show.html.erb
```diff
@@ -22,6 +22,17 @@
           <%= form.text_field :token, class: "form-control", disabled: true %>
         </div>
       </div>
+      <div class="form-group">
+        <%= form.label :time_format, class: "col-lg-2 control-label" %>
+        <div class="col-lg-4">
+          <%= form.radio_button(:time_format, "relative") %>
```
grosser: same `each do` logic here

config/routes.rb
```diff
@@ -70,7 +70,11 @@
 
   resources :deploy_groups, only: [:show]
 
-  resource :profile, only: [:show, :update]
+  resource :profile, only: [:show, :update] do
+    member do
```
grosser: can go away now ?

db/migrate/20160328221058_add_time_format_to_user.rb
```diff
@@ -0,0 +1,8 @@
+class AddTimeFormatToUser < ActiveRecord::Migration
+  def change
+    add_column :users, :time_format, :string, :default => 'relative', :null => false
+    User.all.each do |user|
```
grosser: no longer needed

grosser: also FYI `all.each` will blow up if there are too many records, so  `find_each` is usually better

db/schema.rb
```diff
@@ -149,7 +149,7 @@
     t.integer  "user_id",    limit: 4,                              null: false
     t.integer  "project_id", limit: 4,                              null: false
     t.string   "status",     limit: 255,        default: "pending"
-    t.text     "output",     limit: 1073741823
+    t.text     "output",     limit: 4294967295
```
grosser: I think this is wrong / dangerous ... try to only add the changes you need (copy and the git co master -- db/schema.rb)

test/models/user_test.rb
```diff
@@ -26,6 +26,30 @@
     end
   end
 
+  describe "#time_format" do
+    let(:user) { User.create!(name: "jimbob", email: 'test@test.com') }
+    describe 'time_format' do
```
grosser: 1 describe is enough :)

test/models/user_test.rb
```diff
@@ -26,6 +26,30 @@
...
+        lambda {user.update_attributes!(:time_format => 'foobar') }.must_raise ActiveRecord::RecordInvalid
+      end
+
+      it "should update" do
+        user.update_attributes!(:time_format => 'utc')
+        user.reload
+        user.time_format.must_equal('utc')
+      end
+
+      let(:local_user) { User.create!(name: "bettysue", email: 'bsue@test.com', time_format: 'local') }
```
grosser: all the lets in 1 spot is usually easier to read

grosser: also don't need a let when it is only used once

app/assets/javascripts/timeline.js
irwaters: ok, did some poking into this.  I think that ideally this is the kinda thing that would go into rootScope after app initialization.  this doesn't seem to really work that way though as we go through the config/init stuff on ever page change.  probably best to leave it here unless we want to re-factor the angular stuff more.

irwaters:
bump

app/assets/javascripts/timeline.js
```diff
@@ -96,6 +96,24 @@ samson.filter("userFilter",
...
+      if (timeFormat == 'local') {
+        return moment(td).format('LLL');
+      } else if (timeFormat == 'utc') {
+        return moment(td).utc().format();
+      } else if (timeFormat == 'relative') {
+        return moment(td).fromNow();
+      } else {
+        throw 'timeFormat should be one of local | utc | relative, ' + timeFormat + " provided";
+      }
+
```
grosser: unnecessary newline

app/models/user.rb
```diff
@@ -4,6 +4,7 @@
 class User < ActiveRecord::Base
   include HasRole
   include Searchable
+  TIME_FORMATS = ['local', 'utc', 'relative'].freeze()
```
grosser: no need for `()`

app/models/user.rb
```diff
@@ -17,6 +18,7 @@ class User < ActiveRecord::Base
   validates :role_id, inclusion: { in: Role.all.map(&:id) }
 
   before_create :set_token
+  validates :time_format, inclusion: { in: TIME_FORMATS }
```
grosser: does `validates :time_format, in: TIME_FORMATS` also work ?

app/models/user.rb
```diff
@@ -91,4 +93,5 @@ def project_role_for(project)
   def set_token
     self.token = SecureRandom.hex
   end
+
```
grosser: unnecessary newline

app/views/deploys/recent.html.erb
```diff
@@ -1,6 +1,7 @@
 <h1><%= title %></h1>
 
 <div class="timeline" ng-controller="<%= controller %>">
+  <input type=hidden  id=default_time_format value="<%= current_user.time_format %>">
```
grosser: select the radio button that is current instead of adding a hidden input ?

grosser: then the js for handlin default_time_format can go away too ...

grosser: add `checked` attribute

app/views/deploys/recent.html.erb
```diff
@@ -25,6 +26,15 @@
           <option value="">All</option>
         </select>
       </div>
+      <div class="timeline-filter col-sm-4" >
+        <label for="tlf-status">TIME:</label>
+        <% User::TIME_FORMATS.each do |time_fmt| %>
```
grosser: I'd recommend spelling out short works, fmt -> format ... seems to be a ruby standard

app/views/profiles/show.html.erb
```diff
@@ -22,6 +22,15 @@
           <%= form.text_field :token, class: "form-control", disabled: true %>
         </div>
       </div>
+      <div class="form-group">
+        <%= form.label :time_format, class: "col-lg-2 control-label" %>
+        <div class="col-lg-4">
+          <% User::TIME_FORMATS.each do |time_fmt| %>
+            <%= form.radio_button(:time_format, "#{time_fmt}") %>
```
grosser: `"#{time_fmt}"` -> `time_fmt`

app/views/profiles/show.html.erb
```diff
@@ -22,6 +22,15 @@
...
+            <%= form.label(:time_format, "#{time_fmt.capitalize}", :value => "#{time_fmt}") %>
```
grosser: same here

config/routes.rb
```diff
@@ -70,7 +70,7 @@
 
   resources :deploy_groups, only: [:show]
 
-  resource :profile, only: [:show, :update]
+  resource :profile, only: [:show, :update]
```
grosser: extra whitespace ... set your editor to remove trailing whitespaces

test/models/user_test.rb
```diff
@@ -26,6 +26,28 @@
     end
   end
 
+  describe "#time_format" do
+  let(:user) { User.create!(name: "jimbob", email: 'test@test.com') }
```
grosser: missing indent

test/models/user_test.rb
```diff
@@ -26,6 +26,28 @@
...
+    it "should have a default time format of relative" do
```
grosser: no need for shoulds

test/models/user_test.rb
```diff
@@ -26,6 +26,28 @@
...
+      user.time_format.must_equal('relative')
+    end
+
+    it "should not update" do
```
grosser: try to write the `it` in a way that ideally makes it unnecessary to read what the test does

`it "does not update to a invalid time format"`

test/models/user_test.rb
```diff
@@ -26,6 +26,28 @@
...
+      lambda {user.update_attributes!(:time_format => 'foobar') }.must_raise ActiveRecord::RecordInvalid
```
grosser: can be simplified to 
```Ruby
user. time_format = 'foobar'
refute user.valid?
```

test/models/user_test.rb
```diff
@@ -26,6 +26,28 @@
...
+    end
+
+    it "should update" do
```
grosser: `it "can update to a valid time format"`

test/models/user_test.rb
```diff
@@ -26,6 +26,28 @@
...
+      user.update_attributes!(:time_format => 'utc')
+      user.reload
+      user.time_format.must_equal('utc')
+    end
+
+    local_user = User.create!(name: "bettysue", email: 'bsue@test.com', time_format: 'local')
```
grosser: creating things outside of it/describe means they will not get rolled back and will stack in the test database -> move this into to `it` or a `before`

app/models/user.rb
irwaters: nope

app/views/deploys/recent.html.erb
irwaters: something about the interplay between the angular filter etc is made sad by this.

grosser: maybe `checked="cachekd"` ... but totally possible that this won't work out ... angular might want to set/unset these boxes :(

irwaters: Yeah, looks like it does.  You see it blink for a sec on page reload…

irwaters:
ready for more feedback

test/models/user_test.rb
```diff
@@ -26,6 +26,29 @@
     end
   end
 
+  describe "#time_format" do
+    let(:user) { User.create!(name: "jimbob", email: 'test@test.com') }
+    it "has a default time format of relative" do
+      user.time_format.must_equal('relative')
+    end
+
+    it "does not update" do
```
grosser: `does not update with invalid values`

grosser:
:+1: :tada: :D

PR #780, opened 2016-03-30
Samson 215 (making pipeline'd deployment honor no_code_deployed)

irwaters:
- Add description here
  currently deploys in a pipeline don't honor the no_code_deployed attibute on the stage.  changes here cause it do so.
- Add any screenshots if you are touching the UI
- Remember to add unit tests

/cc @zendesk/samson
### Tasks
- [ ] :+1: from team
### References

https://zendesk.atlassian.net/projects/SAMSON/issues/SAMSON-215
### Risks
- Level: Low/Med/High
  Med


app/models/stage.rb
```diff
@@ -149,7 +149,10 @@ def send_github_notifications?
 
   def production?
```
grosser: looks like the wrong place to add this logic ... the stage is still `production?` ... patch should be in the pipeline part

grosser:
try to start with a test :)

`a pipeline that includes a no-deploy stage should work` or so

app/models/stage.rb
irwaters: figured out where it should be and added it/tests

irwaters:
bump

app/models/stage.rb
```diff
@@ -150,6 +150,10 @@ def send_github_notifications?
...
     if DeployGroup.enabled?
       deploy_groups.empty? ? super : deploy_groups.any? { |deploy_group| deploy_group.environment.is_production? }
+#      deploy_groups.empty? ? (super && !self.no_code_deployed?): deploy_groups.any? do |deploy_group|
```
grosser: delete comments

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,7 +1,10 @@
 module SamsonPipelines::StageConcern
+  require 'byebug'
```
grosser: remove byebug

plugins/pipelines/test/models/stage_test.rb
```diff
@@ -1,4 +1,5 @@
 require_relative '../test_helper'
+require 'byebug'
```
grosser: remove byebug

grosser:
please sanity check the diff on github before asking for a re-review, saves time for everyone :)

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,7 +1,10 @@
...
   # Return true if any stages in the pipeline are marked production
+  # make sure to check the no_code_deployed flag on the 
+  # next deployment as well
```
grosser: original `production?` method does not check no_code_deployed ... so this is the wrong place to add this check

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -8,6 +10,34 @@ def next_stages
     Stage.find(next_stage_ids)
   end
 
+  # Needs to find all the possible stages in case this is a pipeline of pipelines as each subsequent stage
+  # could have valid next_stage_ids
+  def all_stages
```
grosser:
```Ruby
def recursive_next_stage_ids
  return [] if next_stage_ids.empty?
  next_stage_ids.map(&:to_i) + Stage.find(next_stage_ids).flat_map(&:recursive_next_stage_ids)
end
```

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -8,6 +10,34 @@ def next_stages
...
+    result = []
+    stages_ids = Stage.where(id: parent_stage_id).map(&:next_stage_ids)
+    return [parent_stage_id] if stages_ids.empty?
+    stages_ids.each do |stage_id|
+        result << recursive_stages(stage_id)
+    end
+    result << parent_stage_id
+  end
+
+  def deploy_requires_approval?
```
grosser: never used ... how is this passing the tests ?

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -8,6 +10,34 @@ def next_stages
...
+    stage_collection.map { |id| id.to_i }
+    Stage.find(stage_collection);
+  end
+
+  def recursive_stages(parent_stage_id)
+    result = []
+    stages_ids = Stage.where(id: parent_stage_id).map(&:next_stage_ids)
+    return [parent_stage_id] if stages_ids.empty?
+    stages_ids.each do |stage_id|
+        result << recursive_stages(stage_id)
```
grosser: bad indent ... fix your editor and/or sanity check github diff

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -8,6 +10,34 @@ def next_stages
     Stage.find(next_stage_ids)
   end
 
+  # Needs to find all the possible stages in case this is a pipeline of pipelines as each subsequent stage
+  # could have valid next_stage_ids
+  def all_stages
+    stage_collection = []
+    stages = Stage.find(next_stage_ids)
+    stages.each do |this_stage|
```
grosser: don't name things `this_*` :D

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
irwaters: --> #0  SamsonPipelines::StageConcern.deploy_requires_approval? at /Users/iwaters/co/zendesk/samson/plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb:34
    #1  Deploy.waiting_for_buddy? at /Users/iwaters/co/zendesk/samson/app/models/deploy.rb:83
    #2  DeployService.deploy!(stage#Stage, attributes#Hash) at /Users/iwaters/co/zendesk/samson/app/models/deploy_service.rb:15

is how it get's called seemingly.

irwaters:
should be ready for another go!

plugins/pipelines/test/models/stage_test.rb
```diff
@@ -40,6 +40,20 @@
       stage1.update!(next_stage_ids: [ stage3.id, stage2.id ])
       stage1.production?.must_equal true
     end
+
+    it 'returns true if no_code_deployed is false on a production pipeline stage' do
+      stage2.update(production: true)
```
grosser: could wrap this into a describe so every test only has exactly 1 change

```Ruby
describe "on a production pipeline stage" do
  before do
    stage2.update(production: true)
    stage1.update!(next_stage_ids: [ stage2.id ])
  end

  it "is true if a pipelined state has no_code_deployed" do
    stage2.update(production: true)
    stage1.production?.must_equal true
  end

  it "is false if a pipelined state does not have no_code_deployed" do
    stage1.production?.must_equal true
  end
end
```

grosser: also these tests are wrong ... the `production?` does not change ...

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -8,6 +7,28 @@ def next_stages
...
+    Stage.find(stage_collection);
+  end
+
+  def recursive_next_stage_ids
+    return [] if next_stage_ids.empty?
+    next_stage_ids.map(&:to_i) + Stage.find(next_stage_ids).flat_map(&:recursive_next_stage_ids)
+  end
+
+  def deploy_requires_approval?
+    super || all_stages.any? { |next_stage| (next_stage.production? && !next_stage.no_code_deployed?) }
```
grosser: don't duplicate the logic 

```Ruby
super || next_stages.any?(&:deploy_requires_approval?)
```

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
grosser: this and the rest are all private methods

grosser: all stages might be a bit confusing ... how about `resursive_next_stages` or `all_next_stages` ...

also implementation can be:

```
Stage.find(recursive_next_stage_ids)
```

irwaters:
ready for another pass

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,13 +1,27 @@
 module SamsonPipelines::StageConcern
-  # Return true if any stages in the pipeline are marked production
   def production?
-    super || next_stages.any?(&:production?)
+    super || all_next_stages.any? { |next_stage| next_stage.production? && !next_stage.no_code_deployed? }
```
grosser: don't mix production and no_code_deployed ...

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,13 +1,27 @@
...
   end
 
   def next_stages
     Stage.find(next_stage_ids)
   end
```
grosser: private ?

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,13 +1,27 @@
...
+    Stage.find(recursive_next_stage_ids);
+  end
+
+  def recursive_next_stage_ids
+    return [] if next_stage_ids.empty?
+    next_stage_ids.map(&:to_i) + Stage.find(next_stage_ids).flat_map(&:recursive_next_stage_ids)
+  end
+
+  def deploy_requires_approval?
+      super || all_next_stages.any?(&:deploy_requires_approval?)
```
grosser: indent

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
irwaters: making next_stages private blows up stuff in EnvironmentVariableGroupsController

irwaters:
when you have a moment...

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,13 +1,28 @@
...
+  end
+
+  def recursive_next_stage_ids
+    return [] if next_stage_ids.empty?
+    next_stage_ids.map(&:to_i) + Stage.find(next_stage_ids).flat_map(&:recursive_next_stage_ids)
+  end
+
+  def deploy_requires_approval?
+    super || all_next_stages.any? do |next_stage|
+      next_stage.deploy_requires_approval? && !next_stage.no_code_deployed?
```
grosser: deploy_requires_approval already checks no_code_deployed ... so the && can go away ?

grosser: -> `all_next_stages.any?(&: deploy_requires_approval?)`

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
grosser: these 3 new methods can be private though

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,13 +1,28 @@
...
+    end
+  end
+
+  private :all_next_stages
```
grosser: move into private or protected area

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,13 +1,28 @@
...
+  # Needs to find all the possible stages in case this is a pipeline of pipelines as each subsequent stage
+  # could have valid next_stage_ids
+  def all_next_stages
+    Stage.find(recursive_next_stage_ids);
+  end
+
+  def recursive_next_stage_ids
+    return [] if next_stage_ids.empty?
+    next_stage_ids.map(&:to_i) + Stage.find(next_stage_ids).flat_map(&:recursive_next_stage_ids)
+  end
```
jonmoter: Any possibility of this turning into an infinite loop?

Also, you can reuse the `next_stages` method above to DRY up.

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,13 +1,28 @@
...
-  end
 
   def next_stages
     Stage.find(next_stage_ids)
   end
 
+  # Needs to find all the possible stages in case this is a pipeline of pipelines as each subsequent stage
+  # could have valid next_stage_ids
+  def all_next_stages
+    Stage.find(recursive_next_stage_ids);
```
jonmoter: Unnecessary semi-colon

grosser:
1 more rebase or this might fail on merge with new coverage reporting

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
jonmoter: I see that we have a validation below that checks for cycles. So as long as `recursive_next_stage_ids` is called on a Stage that is already known to be valid, it should be safe.  But if this method is ever called on an stage with cycles (either because the stage isn't `valid?`, or `valid?` was never called), this could fall into an infinite loop.

Would probably be safer to have an iterative method that puts all the stage ids into a `Set`, and only checks for descendants that aren't already in the set.

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
irwaters: seems to cause issues with SseRailsEngine.send_event() when these are all made private (specifically next_stages())

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
irwaters: should be ok as there's a check tor next_stage_ids, and valid_pipeline? should prevent a loop in the pipelines.

irwaters:
@jonmoter check and see if this satisfies your recursion requirements.

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -10,6 +12,27 @@ def next_stages
 
   protected
 
+  # Needs to find all the possible stages in case this is a pipeline of pipelines as each subsequent stage
+  # could have valid next_stage_ids
+  def all_next_stages
+    existing_stages = next_stages.collect(&:id).to_set
```
grosser: can next_stage_ids  to save the lookup if there are no next_stages

grosser: basically doing a quick return and doing the other logic after:
```
return if next_stages.empty?
```

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -10,6 +12,27 @@ def next_stages
...
+    loop do
+      last_count = existing_stages.count
+      existing_stages.each do |stage_id|
+        existing_stages += Stage.find(stage_id).next_stage_ids.map(&:to_i).to_set
+      end
+      break if last_count == existing_stages.count
+    end
+    # if we find ourselves with ourselves in the list, we have a loop.
+    # this shouldn't happen, but this will prevent it
+    if existing_stages.include?(id)
```
grosser: we should handle recursion before saving / in a validation and not let it get stored and then later build code to deal with it

irwaters: we are already.  this is in response to @jonmoter not wanting to do recursion.  valid_pipeline? does the validation in this case, as noted there is a small chance for a race.

grosser: then use valid_pipeline here instead of re-implementing the check ?

irwaters: Honestly I'd like to just to back to the recursive code that was there and rely on the model validation to prevent a loop (as it was before).

thoughts?

grosser: me too! :D

irwaters: Ok, back to whence we came.

irwaters:
put this back to the nice recursive stuff.

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,15 +1,26 @@
...
+    super || all_next_stages.any?(&:deploy_requires_approval?)
   end
 
   def next_stages
     Stage.find(next_stage_ids)
   end
 
+  def recursive_next_stage_ids
+    return [] if next_stage_ids.empty?
+    next_stages.collect(&:id) + Stage.find(next_stage_ids).flat_map(&:recursive_next_stage_ids)
```
grosser: no need for another find:  
```
 next_stages.collect(&:id) + next_stages.flat_map(&:recursive_next_stage_ids)
```

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,15 +1,26 @@
...
-  # Return true if any stages in the pipeline are marked production
-  def production?
-    super || next_stages.any?(&:production?)
+
+  def deploy_requires_approval?
+    super || all_next_stages.any?(&:deploy_requires_approval?)
   end
 
   def next_stages
     Stage.find(next_stage_ids)
```
grosser: and this will need to be cached `@next_stages ||= Stage.find(next_stage_ids)`

irwaters:
feedback processed

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,15 +1,26 @@
...
   end
 
   def next_stages
-    Stage.find(next_stage_ids)
+    @next_stages ||= Stage.find(next_stage_ids)
+  end
+
+  def recursive_next_stage_ids
+    return [] if next_stage_ids.empty?
+    next_stages.collect(&:id) + next_stages.flat_map(&:recursive_next_stage_ids)
```
grosser: FYI most code I see uses `map` instead of `collect` and it is also nicely matches to flat_map + it's shorter :)

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,15 +1,26 @@
...
   end
 
   protected
 
+  # Needs to find all the possible stages in case this is a pipeline of pipelines as each subsequent stage
+  # could have valid next_stage_ids
```
grosser: this comment would fit better to `def recursive_next_stage_ids`

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -1,15 +1,26 @@
 module SamsonPipelines::StageConcern
-  # Return true if any stages in the pipeline are marked production
-  def production?
```
grosser: if you remove production? then `def deploy_buddy` needs to check for `deploy_requires_approval?` ... same for `deploy.production?` and `latest_approved_deploy` and `validate_bypass_used_correctly` (I'll remove that in another PR ... so don't worry about validate_bypass_used_correctly) ... also the `production?` would be  wrong now since the stage does to production ...

irwaters: changes made.  a test is currently failing as a result (validation as you mention).

henders:
Don't lose hope @irwaters! I'm hoping you get to merge this soon! :)

henders:
Is this ready to merge?

grosser:
please squasch before merging ... the number of commits is getting a bit out of hand / will be unhelpful when going through a blame to find out why something was added

henders:
bump bump

PR #785, opened 2016-04-02
intuitive scroll

grosser:
- when you scroll up: stop to follow (you most likely want to read something / not get yanked down a second later)
 - when you scroll down to the end: start to follow (you are at the bottom, so new output is not disruptive + no longer need to click the follow button)

@zendesk/samson 

### Risks
- Low: output scroll breaking

### DIFF
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -13,7 +13,8 @@ $(function () {
       $submit = $form.find('input[type=submit]'),
       $reference = $("#deploy_reference"),
       $ref_problem_list = $("#ref-problem-list"),
-      $ref_status_label = $("#ref-problem-warning");
+      $ref_status_label = $("#ref-problem-warning"),
+      $messages = $("#messages");
 
   $("#deploy-tabs a[data-type=github]").click(function (e) {
       e.preventDefault();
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/deploys.js
```diff
@@ -240,13 +240,24 @@ $(function () {
...
       $output.find('.deploy-check').hide();
     }
   });
+
+  // when user scrolls all the way down, start following
+  // when user scrolls up, stop following since it would cause jumping
+  $messages.scroll(function() {
+    var position = $messages.prop("scrollHeight") - $messages.scrollTop() - $messages.height() - 30;
+    if(position > 0 && following) {
+      $("#output-steady").click();
```
sandlerr: would be nice to extract these to functions

sandlerr:
I'm not sure what the magic number 30 represents but :+1: on green

grosser:
somehow the math does not quiet add up ... I guess it's borders not being counted or so ... so I added a bit of padding :)

sandlerr:
would it help to try http://api.jquery.com/outerheight/ instead of `.height()`?

grosser:
does not add up either way ... correct height would be 532 ... height gives me 529 ... outerHeigh 550 innerHeight 548 ... so it's the closest one ... just not correct :unamused:

PR #787, opened 2016-04-04
added new task for zendesk_samson

ocke:
:v:

/cc @zendesk/vegemite

### Description

I added a new rake task that is used in zendesk_samson. This used to be in script/ci but since you removed that I added it in the Rakefile.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### References
* JIRA: 

### Risks
* [HIGH | medium | low] Does it work across browsers (including IE!)?
* [HIGH | medium | low] Does it work in the different products (Lotus, Zopim, People)?
* [HIGH | medium | low] Are there any performance implications?
* [HIGH | medium | low] Any security risks?
* [HIGH | medium | low] What features does this touch?

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -20,6 +20,11 @@ namespace :plugins do
   Rails::TestTask.new(:test) do |t|
     t.pattern = "plugins/*/test/**/*_test.rb"
   end
+
+  task :clean_test_run do
+    sh "bundle exec rake db:create db:migrate"
+    sh "bundle exec rake"
+  end
 end
 
 namespace :test do


Rakefile
```diff
@@ -20,6 +20,11 @@ namespace :plugins do
   Rails::TestTask.new(:test) do |t|
     t.pattern = "plugins/*/test/**/*_test.rb"
   end
+
+  task :clean_test_run do
+    sh "bundle exec rake db:create db:migrate"
```
grosser: travis seems to be happy without the db:migrate ... do we need it ?

grosser:
how is having an extra task simpler then `rake db:migrate default` ? (have to remember something new anyway)

Rakefile
grosser: can db:create / migrate be a dependency so this does not need to boot rake 2 times ?

PR #788, opened 2016-04-04
added zendesk_app to .gitignore

ocke:
:v:

/cc @zendesk/vegemite

### Description

New method on deploy model used by zendesk_app

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

added 2 new methods to deploy model

### References
* JIRA: 

### Risks
* [HIGH | medium | low] Does it work across browsers (including IE!)?
* [HIGH | medium | low] Does it work in the different products (Lotus, Zopim, People)?
* [HIGH | medium | low] Are there any performance implications?
* [HIGH | medium | low] Any security risks?
* [HIGH | medium | low] What features does this touch?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,5 +37,8 @@
 # Ignore temp devel files
 /routes.txt
 
+# Ignore certain plugins
+/plugins/zendesk_app
+
 # .bak files
 *.bak
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -97,6 +97,16 @@ class Deploy < ActiveRecord::Base
     DeployService.new(user).confirm_deploy!(self)
   end
 
+  def self.in_range(time_value, time_unit)
+    where(created_at: (Time.now - time_value.public_send(time_unit))..Time.now)
+  end
+
+  def self.search_by_project(search)
+    return self unless search
+    joins({stage: 'project'})
+    .where('lower(projects.name) like ?', "%#{sanitize_sql_like(search.downcase)}%")
+  end
+
   def self.active
     includes(:job).where(jobs: { status: Job::ACTIVE_STATUSES })
   end


app/models/deploy.rb
```diff
@@ -97,6 +97,16 @@ def pending_start!
     DeployService.new(user).confirm_deploy!(self)
   end
 
+  def self.in_range(time_value, time_unit)
```
grosser: add tests if you add code ... or move the code into the plugin ?

.gitignore
```diff
@@ -37,5 +37,8 @@
 # Ignore temp devel files
 /routes.txt
 
+# Ignore certain plugins
```
grosser: might be a good idea to ignore all plugins to make experimentation easier ...

grosser: alternatively ignore all /_* folders ... then we can link them to a _ name ...

app/models/deploy.rb
```diff
@@ -97,6 +97,16 @@ def pending_start!
...
+    where(created_at: (Time.now - time_value.public_send(time_unit))..Time.now)
```
grosser: .ago instead of time math

grosser: also can be simplified to the faster `"created_at > ?", xxx.ago` (1 time comparison vs 2)

app/models/deploy.rb
```diff
@@ -97,6 +97,16 @@ def pending_start!
...
+  end
+
+  def self.search_by_project(search)
+    return self unless search
+    joins({stage: 'project'})
+    .where('lower(projects.name) like ?', "%#{sanitize_sql_like(search.downcase)}%")
```
grosser: indent on continuation

app/models/deploy.rb
```diff
@@ -97,6 +97,16 @@ def pending_start!
     DeployService.new(user).confirm_deploy!(self)
   end
 
+  def self.in_range(time_value, time_unit)
+    where(created_at: (Time.now - time_value.public_send(time_unit))..Time.now)
+  end
+
+  def self.search_by_project(search)
+    return self unless search
+    joins({stage: 'project'})
```
grosser: no need for `{}` for option hashes

app/models/deploy.rb
grosser: posibly faster / cleaner to search all the matching projects + `.pluck(:id)` and then select stages

sandlerr:
I'd suggest keeping the `Deploy` scopes in your plugin, that makes it less likely that someone deletes them thinking they are unused code.
[Here](https://github.com/zendesk/samson/blob/master/plugins/env/app/decorators/project_decorator.rb)'s an example of patching core classes in a plugin.

PR #791, opened 2016-04-06
SAMSON-220 make rake and rake test run plugin tests too

grosser:
@jonmoter @irwaters 

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/test/models/watchers/deploy_watcher_test.rb
+++ b/plugins/kubernetes/test/models/watchers/deploy_watcher_test.rb
@@ -9,6 +9,9 @@ describe Watchers::DeployWatcher do
   let(:project) { current_release.project }
 
   before do
+    # Disable multithreading so celluloid stays in the same transaction
+    ActiveRecord::Base.stubs(connection: ActiveRecord::Base.connection)
+
     Celluloid.shutdown; Celluloid.boot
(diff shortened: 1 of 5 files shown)


sandlerr:
:+1: on green

grosser:
that was fun ... multithreading and timeouts ... good times :D

plugins/kubernetes/test/models/watchers/deploy_watcher_test.rb
```diff
@@ -9,6 +9,9 @@
   let(:project) { current_release.project }
 
   before do
+    # Disable multithreading so celluloid stays in the same transaction
+    ActiveRecord::Base.stubs(connection: ActiveRecord::Base.connection)
```
sandlerr: :open_mouth:

grosser: found that on a very reliable blog ... so it must be good ;)
http://grosser.it/2016/04/06/testing-suckerpunchcelluloid-vs-activerecord-transactions/

sandlerr: hmm that blogger sounds familiar, should we report the soft_deletion issue upstream to him as well?

grosser: nono ... only send PRs ;)

PR #794, opened 2016-04-07
SAMSON-222 fail validations when deploy groups is used but no group i…

grosser:
screen shot 2016-04-06 at 5 51 47 pm

@zendesk/samson

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'bundler', '>= 1.8.4'
 
 gem 'rails', '~> 4.2.0'
 gem 'puma'
-gem 'dotenv-rails', '~> 0.9'
+gem 'dotenv'
 
 gem 'dogstatsd-ruby', '~> 1.5.0', require: 'statsd'
 gem 'goldiloader'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -176,8 +176,6 @@ GEM
     dotenv (0.11.1)
       dotenv-deployment (~> 0.0.2)
     dotenv-deployment (0.0.2)
-    dotenv-rails (0.11.1)
-      dotenv (= 0.11.1)
     erubis (2.7.0)
     excon (0.47.0)
     execjs (2.6.0)
@@ -479,7 +477,7 @@ DEPENDENCIES
   docker-api
   dogapi (~> 1.9)
   dogstatsd-ruby (~> 1.5.0)
-  dotenv-rails (~> 0.9)
+  dotenv
   faraday-http-cache (~> 1.1)
   github-markdown (~> 0.6.3)
   goldiloader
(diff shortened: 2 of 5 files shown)


jonmoter:
Let's say a user edits a Command, changing it from `cap production deploy` to `cap $DEPLOY_GROUPS deploy`.  That saves the Command. Does it run validations on every Stage that includes that Command?

Or does adding a Command to a Stage trigger the validations of the Stage itself?

I was thinking that rather than making this a validation on the Stage or Command, just display a warning on the deploy screen if it sees that string in the commands it's about to run, but no deploy groups are specified.

If it's a warning, then you don't have the false positive of having an error if someone just put `$DEPLOY_GROUPS` in a comment, or if they really do want to run the command even though it's empty.

grosser:
I think the issue is most common when editing a stage, but also showing it on the deploy show page would catch even more issues ...
I don't think there will be false-positives ... printing the $DEPLOY_GROUPS is not really a valid usecase ...

jonmoter:
FYI, it looks like there are currently 4 stages that fail this check:

```
irb(main):002:0> class Stage
irb(main):003:1>   def invalid_deploy_groups?
irb(main):004:2>     DeployGroup.enabled? && deploy_groups.empty? && command.include?("$DEPLOY_GROUPS")
irb(main):005:2>   end
irb(main):006:1> end
=> :invalid_deploy_groups?
irb(main):007:0> missing = Stage.includes(:deploy_groups).each_with_object({}) { |stage, hash| hash[stage.id] = stage if stage.invalid_deploy_groups? } ; missing.count
=> 4
irb(main):009:0> missing.values.each { |stage| puts [stage.project.name, stage.name].join(' - ') }
ultragrep - test-drdb
JIG - Staging Restart
Embed Key Registry - Production
Knowledge Bank Client - Production build
```

jonmoter:
We talked about it. Will keep the validations, and add a warning on the "new Deploy" screen if this scenario is detected.

grosser:
screen shot 2016-04-07 at 10 53 54 am

jonmoter:
The "Stage" is unnecessary and awkward in that warning message.

Suggestion for text: "This stage contains at least one command using the $DEPLOY_GROUPS environment variable, but there are no Deploy Groups associated with this stage. This is probably an error."

grosser:
Has to start with Stage, since the error is added to :stage
"This is probably an error."  could be added to every validation ... so
it's kind of redundant.

grosser:
"Validation failed: Stage contains at least one command using the $DEPLOY_GROUPS environment variable, but there are no Deploy Groups associated with this stage."

jonmoter:
:+1: on :green_heart:

grosser:
FYI found a bug where we would load .env in test mode anyway because we still had dotenv-rails in the Gemfile

PR #769, opened 2016-03-23
WIP macros are stages that do not deploy code

grosser:
/cc @zendesk/samson

turning macro into a stage with type=Macro ... which gets rid of a lot of duplication and also make macros audited since they keep track of their deploys

keeping old macro tables for a while to make sure we can roll back

screen shot 2016-03-22 at 8 39 49 pm

### Risks
- Medium: macro thing blowing up
- Medium: some views showing macros when they should not

PR #795, opened 2016-04-07
convert macros to have command associations

grosser:
Get rid of command column in preparation for macros/stage unification ... see https://github.com/zendesk/samson/pull/769

existing `command` column will get converted into a actual `Command` new UI:

screen shot 2016-04-07 at 12 01 01 pm

 - includes a bit of duplication, but that will go away once Macros and Stage are merged
 - migration is using only clean classes to not break once macro becomes a Stage subclass

@zendesk/samson 

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/models/concerns/has_commands.rb
+++ b/app/models/concerns/has_commands.rb
@@ -1,0 +1,37 @@
+# expects command_associations to be defined in base class
+module HasCommands
+  def self.included(base)
+    base.class_eval do
+      before_save :build_new_project_command
+      attr_writer :command
+    end
+  end
+
+  def command
+    commands.map(&:command).join("\n")
+  end
+
+  def command_ids=(new_command_ids)
+    super.tap do
+      reorder_commands(new_command_ids.reject(&:blank?).map(&:to_i))
+    end
+  end
+
+  private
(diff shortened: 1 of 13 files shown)


jonmoter:
Can you fix the CodeClimate issue?

grosser:
extracted into a module ... good to go ?

app/models/concerns/has_commands.rb
```diff
@@ -0,0 +1,36 @@
+module HasCommands
+  def self.included(base)
+    base.class_eval do
+      before_save :build_new_project_command
+      attr_writer :command
+    end
+  end
+
+  def command
```
jonmoter: How about `script` instead?

app/models/concerns/has_commands.rb
```diff
@@ -0,0 +1,36 @@
+module HasCommands
```
jonmoter: Add a comment that it expects `command_associations` to be defined by including class.

jonmoter:
Okay, looks good. :+1:

grosser:
added comment + made migration reversible

PR #796, opened 2016-04-07
privacy for kubernetes

grosser:
only thing ever used from the outside is `restart_watcher`

@jonmoter 

### Risks
- Low: kubernetes breaking


jonmoter:
:+1:

PR #797, opened 2016-04-08
create cache directory before trying to resolve a git ref [SAMSON-207]

jonmoter:
If you try creating a Build before running a stage, you get a crash. This fixes it.

Also handles a scenario where if you create a Build with a branch, it could map that to the wrong SHA if the local cache wasn't up to date.

I tried writing a unit test for this, but all the logic around creating cached repos defeated me. Welcome to have help here.

/cc @zendesk/samson, @grosser 

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-207

### Risks
- Level: Low

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -106,6 +106,13 @@ class BuildsController < ApplicationController
   end
 
   def git_sha
-    current_project.repository.commit_from_ref(new_build_params[:git_ref], length: nil)
+    @git_sha ||= begin
+      # Create/update local cache to avoid getting a stale reference
+      current_project.with_lock(holder: 'BuildsController#create') do
+        current_project.repository.setup_local_cache!
+      end
+
+      current_project.repository.commit_from_ref(new_build_params[:git_ref], length: nil)
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/build.rb
```diff
@@ -79,8 +79,10 @@ def validate_git_reference
 
     return if errors.include?(:git_ref) || errors.include?(:git_sha)
 
-    project.with_lock(holder: 'Build reference validation') do
-      project.repository.setup_local_cache!
+    unless project.repository.last_pulled
+      project.with_lock(holder: 'Build reference validation') do
+        project.repository.setup_local_cache!
+      end
     end
```
jonmoter: Without this check, it would end up doing two `git fetch` operations each time a Build was created, and that's unnecessary overhead.

app/controllers/builds_controller.rb
```diff
@@ -106,6 +106,13 @@ def create_build
   end
 
   def git_sha
-    current_project.repository.commit_from_ref(new_build_params[:git_ref], length: nil)
+    @git_sha ||= begin
+      # Create/update local cache to avoid getting a stale reference
+      current_project.with_lock(holder: 'BuildsController#create') do
```
grosser: can we extract this with_lock behavior / push it into setup_local_cache! ?

grosser:
:+1:

PR #798, opened 2016-04-08
remove user from macros, it was never used

grosser:
@zendesk/samson 

```
>> Macro.where(user_id: nil).count
=> 53
>> Macro.where('user_id is not null').count
=> 0
```

also a bunch of cleanup from @sbrnunes copy-paste-spree in https://github.com/zendesk/samson/pull/560 ... more coming later

### Risks
- None


jonmoter:
:+1: on :green_heart:

PR #799, opened 2016-04-08
Fix a few issues related to Docker builds and Kubernetes

jonmoter:
* fix a crash where Docker build output was not displayed
* fix a crash where submitting Kubernetes info for a DeployGroup crashed
* allow rebuilding a Docker image if it fails the first time

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


app/views/jobs/_output.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="output">
```
grosser: can this re-use the shared output ?

app/models/docker_builder_service.rb
```diff
@@ -11,6 +11,8 @@ def initialize(build)
   end
 
   def run!(image_name: nil, push: false)
+    build.docker_build_job.try(:destroy) # if there's an old build job, delete it
```
grosser: test for this ?

grosser:
👍

app/views/jobs/_output.html.erb
jonmoter: Possibly, I'll try...

app/models/docker_builder_service.rb
jonmoter: `DockerBuilderService` needs more tests in general, since the `run!` method currently doesn't have any tests.  I'll circle back on this next week

jonmoter:
Travis build failure was a timeout on a call to `Celluloid.shutdown` in an unrelated test. It only failed for the sqlite3 build.  I'm chalking that up to a Travis issue and merging.

PR #800, opened 2016-04-09
remove redundant user level tests / actually test project access logic works

grosser:
as with all other permission tests we now only test the lowest level of access and do no repeat on every level 

 - actually tests that access is not allowed when the user does not have access to a project
 - inlines the HasRole modules that were only used once

```Ruby
# before
as_a_project_deployer ... lots of tests
as_a_deployer copy-pasted

# after
as_a_project_deployer ... lots of tests
```

@zendesk/samson

### Risks
- None


jonmoter:
:+1:

PR #801, opened 2016-04-09
roles refactor

grosser:
- 500 lines of angular -> 18 lines of jquery
 - 100 lines of ruby gone
 - 2 routes gone
 - 1 extra request on page load gone / waiting for it to finish
 - less tests needed / less logic
 - js/ruby no longer share routing logic
 - use `Role` for project roles, instead of a different role_id that has different meaning (1=admin / 1=deployer)
 - do not store `None` role assignments, they are useless
 - dead simple to debug

@zendesk/samson 

### Risks
- medium: project roles breaking


app/controllers/project_roles_controller.rb
```diff
@@ -1,48 +1,23 @@
...
-  skip_before_action :require_project, only: [:index]
-
-  before_action :authorize_project_admin!, only: [:create, :update]
-
-  def index
-    render json: ProjectRole.all, root: false
-  end
+  before_action :authorize_project_admin!
 
   def create
```
jonmoter: Comment that this is a combo of create/update/destroy

jonmoter:
:+1:

PR #805, opened 2016-04-10
simple/helpful coverage enforcement with only 5% runtime overhead

grosser:
@zendesk/samson

 - 5% runtime overhead when used vs 30+% for simplecov (for single files)
 - direct inline coverage results when working on a specific file vs having to re-run rake
 - direct information on what lines need coverage
 - no overhead when running by line number / `-n`

```
app/controllers/sessions_controller.rb lower then expected coverage 94% vs 100%
Lines missing coverage:
app/controllers/sessions_controller.rb:84
app/controllers/sessions_controller.rb:85
app/controllers/sessions_controller.rb:86
```

### Risks
- None

### DIFF
--- a/test/support/single_cov.rb
+++ b/test/support/single_cov.rb
@@ -1,0 +1,67 @@
+class SingleCov
+  COVERAGES = []
+
+  class << self
+    def expect(file, percent)
+      COVERAGES << [file, percent]
+    end
+
+    def verify!(result)
+      COVERAGES.each do |file, expected_percent|
+        uncovered!(file) unless coverage = result[File.expand_path(file)]
+
+        line_of_code = coverage.compact.count
+        line_of_covered_code = coverage.count { |l| l && l > 0 }
+        actual_percent = 100 * line_of_covered_code / line_of_code
+        next if actual_percent == expected_percent
+
+        details = "#{actual_percent}% vs #{expected_percent}%"
+        if actual_percent > expected_percent
+          warn "#{file} exceeds expected coverage #{details}, increment expected coverage?"
+        else
+          warn "#{file} lower then expected coverage #{details}"
+          warn "Lines missing coverage:"
+          warn coverage.select { |c| c == 0 }.each_with_index.map { |c, i| "#{file}:#{i+1}" }
+          exit 1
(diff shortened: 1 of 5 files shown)


test/support/single_cov.rb
```diff
@@ -0,0 +1,67 @@
...
+        next if actual_percent == expected_percent
+
+        details = "#{actual_percent}% vs #{expected_percent}%"
+        if actual_percent > expected_percent
+          warn "#{file} exceeds expected coverage #{details}, increment expected coverage?"
+        else
+          warn "#{file} lower then expected coverage #{details}"
+          warn "Lines missing coverage:"
+          warn coverage.select { |c| c == 0 }.each_with_index.map { |c, i| "#{file}:#{i+1}" }
+          exit 1
```
sandlerr: this doesn't actually kill the process immediately because everything is running `at_exit`? It just makes sure the code is not 0?

test/support/single_cov.rb
```diff
@@ -0,0 +1,67 @@
...
+  class << self
+    def expect(file, percent)
+      COVERAGES << [file, percent]
+    end
+
+    def verify!(result)
+      COVERAGES.each do |file, expected_percent|
+        uncovered!(file) unless coverage = result[File.expand_path(file)]
+
+        line_count = coverage.count { |l| l }
```
sandlerr: I guess we don't need to pass the block here? Or are some of these elements `false`?

test/support/single_cov.rb
grosser: yep

test/support/single_cov.rb
grosser: we want to ignore the nils ... just count would not do that

sandlerr: makes sense. Alternatively we could compact and count, not sure what's more readable

sandlerr:
:+1: on green

test/support/single_cov.rb
grosser: updated with .compact.count ... seems to be a bit more intention revealing ... also made the variables more readable

PR #807, opened 2016-04-11
better highlight

sandlerr:
@zendesk/samson 

fixes the problems in the previous PR:

- broken streaming output (horizontal scroll instead of vertical)
- annoying scroll event when clicking on a line of deploy log output

accidentally based this work on #806, I'll rebase if that isn't easily mergable

### DIFF
--- a/app/assets/stylesheets/deploys.scss
+++ b/app/assets/stylesheets/deploys.scss
@@ -136,12 +136,15 @@ td.status {
   margin-top: 10px;
 }
 
-#messages > span {
-  display: inline-block;
-  width: 100%;
-  cursor: pointer;
+#messages {
+  white-space: pre-wrap;
+  > span {
+    display: inline-block;
+    width: 100%;
+    cursor: pointer;
 
-  &.highlighted {
-    background-color: #0700d6;
+    &.highlighted {
+      background-color: #0700d6;
+    }
   }
 }
--- a/app/assets/stylesheets/deploys.scss
+++ b/app/assets/stylesheets/deploys.scss
@@ -135,3 +135,13 @@ td.status {
 #ref-problem-list {
   margin-top: 10px;
 }
+
+#messages > span {
+  display: inline-block;
+  width: 100%;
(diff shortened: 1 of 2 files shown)


grosser:
after stopping a build it inserts newlines and then looks like this:

screen shot 2016-04-11 at 8 07 46 am

highlighting on click / on shift-click does not work ... seeing a js error "Uncaught TypeError: Cannot read property 'length' of undefined"

sandlerr:
so strange, it seems it has 3 states with different styles needed

1. during deploy - `block`
2. stopped deploy - `inline-block`
3. loaded a page that was already stopped - `inline` or `block`

I'll look more at it.

sandlerr:
fixed the JS and the separation. Turns out literal linebreaks are respected inside `<pre>` tags, and we insert them between lines... sometimes. I fixed it with CSS rather than resolving the root cause but it shouldn't create any more problems.

app/assets/stylesheets/deploys.scss
```diff
@@ -135,3 +135,16 @@ td.status {
 #ref-problem-list {
   margin-top: 10px;
 }
+
+#messages {
+  white-space: pre-wrap;
+  & > span {
```
grosser: `& >` .... just ` > ` ... you want all spans inside, right ?

grosser:
minor bugs: 
 - when continuing to shift-click it does not expand, but selects a new starting line
 - selecting line 3-10 breaks ... others work fine ... might be that there is a blank line ... not sure

grosser:
👍

sandlerr:
> when continuing to shift-click it does not expand, but selects a new starting line

That was intentional, but now I've clicked around I see that it should always expand the selection when shift-clicking.

> selecting line 3-10 breaks ... others work fine ... might be that there is a blank line ... not sure

yes, lines with no characters in them have a height of `0` :( I tried setting it to `18px` always but there is still a black gap so going to leave out that change.

PR #810, opened 2016-04-12
show js errors

grosser:
makes debugging easier / errors visible so we don't ship them by accident

screen shot 2016-04-11 at 8 00 16 pm

@zendesk/samson

### DIFF
--- a/app/assets/javascripts/error.js
+++ b/app/assets/javascripts/error.js
@@ -1,0 +1,10 @@
+// show errors immediately so we can catch them early and fix
+$(function(){
+  if($('meta[name=environment]').first().attr('content') == "development") {
+    window.onerror = function(msg, url, line, col, error) {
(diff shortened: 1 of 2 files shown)


grosser:
- travis: random failure
 - CC: complaining about the alert usage

jonmoter:
How much of Samson have you gone through to check for JS errors?  I'm wary of pushing this live, and then having a bunch of alerts popping up in their faces.

Have you considered using [airbrake-js](https://github.com/airbrake/airbrake-js) for reporting client-side Javascript errors to Airbrake?

grosser:
We could limit this to `Rails.env.development?` ...

jonmoter:
Yeah, limiting it to development sounds like a better idea to me.

grosser:
updated to only alert in development

irwaters:
👍

PR #811, opened 2016-04-12
sort deploy groups with natural sorting

grosser:
screen shot 2016-04-11 at 8 53 41 pm
screen shot 2016-04-11 at 9 10 28 pm
screen shot 2016-04-11 at 9 13 10 pm
screen shot 2016-04-11 at 9 13 32 pm

@zendesk/samson

### DIFF
--- a/app/models/deploy_group.rb
+++ b/app/models/deploy_group.rb
@@ -28,6 +28,10 @@ class DeployGroup < ActiveRecord::Base
     "#{name} (#{environment.name})"
   end
 
+  def natural_order
+    name.split(/(\d+)/).each_with_index.map { |x, i| i.odd? ? x.to_i : x }
+  end
+
   private
 
   def permalink_base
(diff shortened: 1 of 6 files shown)


app/models/deploy_group.rb
```diff
@@ -28,6 +28,10 @@ def long_name
     "#{name} (#{environment.name})"
   end
 
+  def natural_order
+    name.split(/(\d+)/).each_with_index.map { |x, i| i.odd? ? x.to_i : x }
```
sandlerr: `name[/\d+/].to_i` ?

sandlerr:
:+1:

app/models/deploy_group.rb
grosser: would only work if all had a number in their name
something like master1 / pod1 should still sort alphabetically

PR #815, opened 2016-04-13
setup/teardown -> before/after

grosser:
@zendesk/samson 

saw a few weird behaviors when mixing before and setup ... so let's unify on 1 way of doing things ...

### Risks
- None

### DIFF
--- a/plugins/kubernetes/test/controllers/kubernetes_releases_controller_test.rb
+++ b/plugins/kubernetes/test/controllers/kubernetes_releases_controller_test.rb
@@ -30,7 +30,7 @@ describe KubernetesReleasesController do
 
   as_a_deployer do
     describe 'a GET to #index' do
-      setup :current_release_count
+      before { current_release_count }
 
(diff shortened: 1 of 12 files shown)


test/models/buddy_check_test.rb
```diff
@@ -65,7 +65,7 @@
     before do
       stage.update_attribute(:production, true)
       job_execution.stubs(:execute!)
-      job_execution.stubs(:setup!).returns(true)
+      job_execution.stubs(:before!).returns(true)
```
pschambacher: not so sure about this one?

pschambacher:
:+1: except for derp find/replace

test/models/buddy_check_test.rb
grosser: good catch! ... apparently the test was green even with that ...

grosser: fixed!

pschambacher:
:+1:

PR #817, opened 2016-04-13
make user project roles unique

grosser:
UI only allows 1 role per user and project ... so let the db enforce that

```
>> [1][0..-2]
=> []
>> [1,2][0..-2]
=> [1]
```

tested locally by disabling the model validation and then migrating

@jonmoter 

/cc @zendesk/samson 

### Risks
- Low: users losing access

### DIFF
--- a/db/migrate/20160413171549_unique_roles.rb
+++ b/db/migrate/20160413171549_unique_roles.rb
@@ -1,0 +1,19 @@
+class UniqueRoles < ActiveRecord::Migration
+  class UserProjectRole < ActiveRecord::Base
+  end
+
+  def up
+    # delete the lowest permission duplicate roles
+    UserProjectRole.all.group_by(&:user_id).each do |_user, roles|
+      roles.sort_by(&:role_id)[0..-2].each(&:destroy)
+    end
+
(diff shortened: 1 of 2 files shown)


db/migrate/20160413171549_unique_roles.rb
```diff
@@ -0,0 +1,14 @@
+class UniqueRoles < ActiveRecord::Migration
+  def up
+    UserProjectRole.all.group_by(&:user_id).each do |user, roles|
+      roles.sort_by(&:role_id)[0..-2].each(&:destroy)
+    end
```
jonmoter: Wary of doing migrations that depend on ActiveRecord models.  If we nuke that class 6 months from now, running `rake db:migrate` on a new DB will fail.

jonmoter: Also, could you add a comment saying what this is doing. I had to read the code a few times to get what was going on

grosser: done and done

jonmoter:
Travis failure looks like an unrelated timeout.

:+1:

PR #819, opened 2016-04-13
setting deploy groups and not using them is a bad practice

grosser:
saw a lot of stages already that select deploy groups but then use `cap production deploy` or `cap pod5 deploy` ... also including some stages having deploy group x selected and deploying to y

@jonmoter 
/cc @zendesk/samson 

### Risks
- low: scripts using $DEPLOY_GROUPS internally and failing -> add a `# uses $DEPLOY_GROUPS` comment to by pass

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -188,8 +188,15 @@ class Deploy < ActiveRecord::Base
   # commands and deploy groups can change via many different paths,
   # so we validate once a user actually tries to execute the command
   def validate_stage_uses_deploy_groups_properly
-    if DeployGroup.enabled? && stage.deploy_groups.none? && stage.command.include?("$DEPLOY_GROUPS")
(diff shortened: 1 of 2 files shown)


jonmoter:
Plz investigate how many stages this would break.

grosser:
```
c=0
Stage.find_each do |stage|
  if stage.command.include?("$DEPLOY_GROUPS") && stage.deploy_groups.empty?
    puts "BAD STAGE #{stage.id}"
    c += 1
  end
end
```

 - 135 legitimate ones that only use deploy groups as a naming contention / to enforce buddy checks
 - 134 bad ones that should use $DEPLOY_GROUPS
... good times :D 

so lot's of breakage ... how about we add a `&& createa_at > 2016-05-01` to make every new stage have to deal with it / stop the spread ... and then we can manually fix all the old ones ?

PR #820, opened 2016-04-13
more coverage and better coverage excemptions

grosser:
keeping percentage of covered lines constant is not good enough:
 - fails when adding new fully covered code
 - fails when removing fully covered code
 - does not catch adding new partially covered code if the percentage does not change

number of uncovered lines is more accurate and what should never go up / will clearly show that a PR adds 1/2/3 new uncovered lines

@zendesk/samson 

### Risks
- None


pschambacher:
:+1: looks good

PR #822, opened 2016-04-13
we need env variables inside of application.rb so we have to load earlier

grosser:
@zendesk/samson

previously this worked because dotenv-rails always loaded the .env file during Bundler.require


jonmoter:
:+1: thanks!

PR #824, opened 2016-04-14
Fix Dockerfile

rata:
Just trying to build the docker image it fails, this file was needed for bundle to work for me.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

Without this, when building the docker image it fails with:

[!] There was an error parsing `Gemfile`: No such file or directory @ rb_sysopen - .ruby-version. Bundler cannot continue.

With this patch, the build works just fine.


grosser:
👍

PR #827, opened 2016-04-15
SAMSON-227 secret storage

grosser:
https://zendesk.atlassian.net/browse/SAMSON-227

@zendesk/paas
@zendesk/samson 
/cc @cericksen 

 - secret values are encrypted using 'aes-256-cbc' with session secret as key and generated salt and stored in the database
 - existing values are never shown to users
 - project admins can create/edit/delete secrets under their projects namespace
 - admins can create/edit/delete secrets under any namespace
 - pluggable backend (keys/read/write/delete) to later support vault and other backends

# TODO:
 - [ ] integration with commands and env plugin

### Risks
- None


app/models/secret_value.rb
```diff
@@ -0,0 +1,61 @@
+require 'attr_encrypted'
+
+class SecretValue
+  class DbBackend < ActiveRecord::Base
```
jonmoter: Add comment describing what SecretValue and DbBackend are supposed to be

grosser:
screen shot 2016-04-14 at 5 45 19 pm
screen shot 2016-04-15 at 10 19 26 am

db/schema.rb
```diff
@@ -302,6 +302,17 @@
   add_index "releases", ["build_id"], name: "index_releases_on_build_id", using: :btree
   add_index "releases", ["project_id", "number"], name: "index_releases_on_project_id_and_number", unique: true, using: :btree
 
+  create_table "secret_values", id: false, force: :cascade do |t|
+    t.string   "id",                 limit: 255
+    t.string   "encrypted_value",    limit: 255, null: false
+    t.string   "encrypted_value_iv", limit: 255, null: false
+    t.string   "encryption_key_sha", limit: 255, null: false
+    t.datetime "created_at",                     null: false
```
jonmoter: Add fields for user_ids for who created and updated it.

app/models/secret_value.rb
jonmoter: Refactor this a bit to be more intuitive.

grosser:
screen shot 2016-04-15 at 3 56 36 pm

app/models/secret_storage.rb
```diff
@@ -0,0 +1,79 @@
...
+      validates :id, format: /\A\S+\/\S*\Z/
+
+      private
+
+      def store_encryption_key_sha
+        self.encryption_key_sha = Digest::SHA2.hexdigest(ENCRYPTION_KEY)
+      end
+    end
+
+    def self.read(key)
```
jonmoter: Return an `OpenStruct` rather than a Hash, so you can say `secret.value` on the response?

app/models/user.rb
```diff
@@ -27,6 +27,15 @@ def starred_project?(project)
     starred_projects.include?(project)
   end
 
+  def administrated_projects
```
jonmoter: Comment that this returns a scope rather than an array

app/views/admin/secrets/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    <thead>
+      <tr>
+        <th>Key</th>
+        <th>Project</th>
+        <th></th>
+      </tr>
+    </thead>
+
+    <tbody>
+      <% @secrets.each do |key| %>
```
jonmoter: `@secret_keys` would be more descriptive

jonmoter:
Okay.  I think the data model works.  Still not convinced about using hashes instead of real objects, but let's give it a try and see how it works when we get a second backend.

:+1:

PR #828, opened 2016-04-15
extract errors partial

grosser:
getting rid of flash messages in favor of displaying detailed error messages

@zendesk/samson

### Risks
- None


pschambacher:
:+1:

PR #833, opened 2016-04-15
make sure there is a test for everything

grosser:
@zendesk/samson

```
either shows this when a new test was added:
Remove ["test/serializers/stage_serializer_test.rb"] from known_missing!

or this when someone added a file without a test:
missing test for lib/warden/strategies/basic_strategy.rb at test/lib/warden/strategies/basic_strategy_test.rb
```
### Risks
- None


jonmoter:
:+1:

PR #838, opened 2016-04-18
gemify single_cov

grosser:
@zendesk/samson 

### Risks
- None


test/test_helper.rb
```diff
@@ -1,5 +1,8 @@
 ENV["RAILS_ENV"] ||= "test"
 
+require 'single_cov'
+SingleCov.setup :minitest
```
pschambacher: This should be `SingleCov.setup :rspec` :trollface:

pschambacher:
👍  on green

PR #839, opened 2016-04-18
more coverage and change project url to include a actual url

grosser:
@zendesk/samson 

### Risks
- Level: consumers of projects json might fail when they always prefixed the host to the url


app/serializers/project_serializer.rb
```diff
@@ -2,6 +2,6 @@ class ProjectSerializer < ActiveModel::Serializer
   attributes :id, :name, :url
 
   def url
-    project_path(object)
+    project_url(object)
```
pschambacher: Is this actually used? You renamed it but I don't see any change elsewhere?

grosser: did not see it being used anywhere ... maybe some angular magic ...

pschambacher: 😨

pschambacher:
👍  on green I guess

PR #841, opened 2016-04-18
resolve secrets when executing commands

grosser:
@zendesk/samson @zendesk/paas 

 - resolve `secret://` usages in commands
 - do not print secrets
 - only allow to access global or project specific secrets

### Risks
- Medium: accidentally exposing secrets


irwaters:
+1

Issue #831, opened 2016-04-15
kubernetes deployment does not delete the old ReplicaSet


rata:
Hi,

When testing the kubernetes plugin, using a 1.2.0 kubernetes cluster, I run into some issues.

The first deployment is done just fine. But when I do a new one, going also to "release" --> new release, etc. The deployment is made, but the old pods continue to be there running because the old replica set is there too. For example, after doing a second deploy, I see:
```
(env-aws) Rodrigos-MacBook-Pro:wido rodrigocampos$ kubectl get rs 
NAME                     DESIRED   CURRENT   AGE
wido-web-rc-3517251930   2         2         15m
wido-web-rc-3998679387   2         2         14m
(env-aws) Rodrigos-MacBook-Pro:wido rodrigocampos$ kubectl get deployments
NAME          DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
wido-web-rc   2         2         2            2           15m
(env-aws) Rodrigos-MacBook-Pro:wido rodrigocampos$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
cadvisor-signalfx-qw9am        1/1       Running   1          2d
wido-web-rc-3517251930-0wg6t   3/3       Running   0          13m
wido-web-rc-3517251930-b7608   3/3       Running   0          15m
wido-web-rc-3998679387-sa9u2   3/3       Running   0          15m
wido-web-rc-3998679387-u5teb   3/3       Running   0          15m
```

The old pods are never deleted because the old ReplicaSet is never deleted. I'm not sure why this is, as the deployment is made server-side using the "Deployment" Kind.

Now I realize that maybe it is not detecting it is a new deploy and it is creating it again, and that's why it creates a new RS instead of having both at the same time and do a rolling update until the deploy finishes and the other ReplicaSet is deleted. When the deployment was running, the 2 new pods were created without waiting for the other to go down, so probably is not doing an update.

Is this possible?

grosser:
@jonmoter

jonmoter:
Hi Rata.  I'm looking into the same code right now.

The Samson/Kubernetes integration was done and tested on Kubernetes 1.1, and since the Deployments API is still in Beta, I think they might've changed a few things between 1.1 and 1.2.  I'm currently going through the code and figuring out what (if anything) needs to be updated.

There are also some issues with the Kubernetes Dashboard page when running against 1.2, which I'm looking at.  I'll cc you on any PRs I make to resolve it.

[...]

PR #842, opened 2016-04-18
Bug/deployment selector

jbarreneche:
Hi, we are using kubernetes 1.2 and we run into the following issue. The definition of labelSelector [changed](http://kubernetes.io/docs/api-reference/extensions/v1beta1/definitions/#_v1beta1_labelselector) and now to match by labels you need to nest `.matchLabels` instead of putting the labels directly into the `selector` key.

This fix only supports kubernetes 1.2. Let me know how if you want to support 1.2, and how you would like to handle the support for multiple kubernetes versions.

This fix is related with #831 . 

Also, the last commit allows the base "kubernetes/deployment.yml" to skip the the "selector.matchLabels" definition.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Med
(Incompatibility with kubernetes 1.1)

### DIFF
--- a/plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
+++ b/plugins/kubernetes/app/models/concerns/kubernetes/deploy_yaml.rb
@@ -63,6 +63,9 @@ module Kubernetes
     # Sets the metadata that is going to be used as the selector. Kubernetes will use this metadata to select the
     # old and new Replication Controllers when managing a new Deployment.
     def set_selector_metadata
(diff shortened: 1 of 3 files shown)


rata:
It's worth noting that the deployment API in k8s 1.1 was in alpha (the beta used is a documented mistake, it was alpha in 1.1 and is beta in 1.2)

jonmoter:
I'm fine with just supporting the most recent version of Kubernetes. I don't know of anyone else using this plugin.

👍

PR #844, opened 2016-04-18
upgrade kubeclient gem to 1.1.3

jonmoter:
* Upgrades to a more recent version of kubeclient

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


grosser:
👍

PR #846, opened 2016-04-18
make deploy buddy info dependent on wether a deploy buddy was required

grosser:
@zendesk/samson 

@irwaters 

... also fixes deleted buddies not showing

### Risks
- None

### DIFF
--- a/test/models/deploy_test.rb
+++ b/test/models/deploy_test.rb
@@ -1,67 +1,101 @@
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 40
+SingleCov.covered! uncovered: 38
 
 describe Deploy do
   let(:project) { projects(:test) }
   let(:user) { users(:deployer) }
   let(:user2) { users(:admin) }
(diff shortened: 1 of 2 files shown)


test/models/deploy_test.rb
```diff
@@ -1,67 +1,101 @@
...
+  describe "#summary" do
+    let!(:deploy) { create_deploy! }
 
-    describe "no buddy message at all" do
-      it "returns no buddy name when BuddyCheck is not enabled" do
-        BuddyCheck.stubs(:enabled?).returns(false)
-        @deploy.summary.must_match(/#{user.name}  deployed/)
-      end
+    it "shows no buddy" do
+      deploy.summary.must_equal "Deployer  deployed baz to Staging"
```
sandlerr: extra whitespace?

sandlerr:
:+1: on green

test/models/deploy_test.rb
grosser: will be rendered as html, so does not really matter (also was there before
..) :)

PR #847, opened 2016-04-19
make sure we do not have accidental non-tests + fix newrelic wtf

grosser:
@zendesk/samson

found a test that did not have a _test.rb extension and did not work because we overwrote the existing namespace ... so it was never loaded

... not sure about the name ... NewRelicHelper is not great, but at least it is unique ...

### Risks
- Low: newrelic things breaking


pschambacher:
👍

PR #851, opened 2016-04-19
mirror lib structure in tests

grosser:
@zendesk/samson less extra code and easier to find where the tests are ...

### Risks
- None


pschambacher:
👍

PR #853, opened 2016-04-19
fix tests messing up global vars and fix deploy nav not rendering

grosser:
@zendesk/samson 

### Risks
- None


pschambacher:
👍  on green

PR #855, opened 2016-04-20
fix migration

sandlerr:
@grosser @zendesk/vegemite 

this is failing our deploys (string indices need this because mysql)


ocke:
👍

grosser:
👎 need a real fix by setting up mysql correctly :(

sandlerr:
the same thing was done in every other string index

grosser:
hmm I guess it's cheaper/faster to have short indices too ... just seems like a hacky solution to something that should work out of the box ...

PR #856, opened 2016-04-20
extract datadog to a plugin

grosser:
... the last custom fields and callbacks component is finally gone :D

@zendesk/samson 

### Risks
- Medium: datadog integration breaking


pschambacher:
👍  on green

PR #857, opened 2016-04-20
prevent accidental manipulation of viewers without syncing output and add tests

grosser:
previously calling shift/unshift etc all worked but would not notify output ... bad pattern ...

@zendesk/samson 

### Risks
- None


pschambacher:
👍  nice find

PR #858, opened 2016-04-20
use common helper for environments

grosser:
### Risks
- None

### DIFF
--- a/app/controllers/admin/environments_controller.rb
+++ b/app/controllers/admin/environments_controller.rb
@@ -4,11 +4,9 @@ class Admin::EnvironmentsController < ApplicationController
   before_action :environment, only: [:edit, :update, :destroy]
 
   def index
-    @environments = Environment.all
-
     respond_to do |format|
       format.html
-      format.json { render json: @environments }
+      format.json { render json: Environment.all }
     end
   end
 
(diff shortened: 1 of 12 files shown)


app/controllers/projects_controller.rb
```diff
@@ -90,7 +89,7 @@ def project_params
         :release_branch,
         :deploy_with_docker,
         :auto_release_docker_image,
-        stages_attributes: stage_permitted_params
+        {stages_attributes: stage_permitted_params}
```
pschambacher: some spaces would be bice

app/controllers/admin/environments_controller.rb
```diff
@@ -4,11 +4,9 @@ class Admin::EnvironmentsController < ApplicationController
   before_action :environment, only: [:edit, :update, :destroy]
 
   def index
-    @environments = Environment.all
-
     respond_to do |format|
       format.html
-      format.json { render json: @environments }
+      format.json { render json: Environment.all }
```
pschambacher: why not use the application helper then?

pschambacher:
👍

app/controllers/admin/environments_controller.rb
grosser: then I'd have to include it in the controller :(

PR #859, opened 2016-04-20
show nice inline error descriptions instead of stuffing everything in to the flash

grosser:
@zendesk/samson

before:
screen shot 2016-04-20 at 3 46 41 pm

after:
screen shot 2016-04-20 at 3 46 16 pm

### Risks
- Low: errors not showing


pschambacher:
👍

PR #860, opened 2016-04-21
cache for the stars

grosser:
@zendesk/samson

 - avoid unnecessary sql requests in json calls
 - avoid star query on every request when db was not updated
 - improve testing for star controller
 - simplify StarsController to not need overrides

### Risks
- Low: star issues


sandlerr:
:+1:

pschambacher:
👍

PR #861, opened 2016-04-21
rename command -> script to make it more obvious what is going on under the hood

grosser:
samson is running a bash script that shows commands and fails when a sub-command fails ... if someone was to copy-paste that, it would need a `set -ex` anyway ... and this would make the inner working of samson more obvious

Also resolves the naming conflict between doing `command=` which expects a single command and `command` returning the full script.

before:
screen shot 2016-04-20 at 5 56 36 pm

after:
screen shot 2016-04-20 at 5 57 22 pm

@zendesk/samson 
/cc @jonmoter 

### Risks
- Low: stage/macros failures


jonmoter:
I like!  👍

PR #862, opened 2016-04-21
Irwaters/paas 114 (adding vault as a secret backend)

irwaters:
Add vault as a secret backend.  

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:
https://zendesk.atlassian.net/browse/PAAS-114
### Risks
- Level: Low/Med/High
Low


.env.example
```diff
@@ -68,3 +68,7 @@ GITHUB_TOKEN={fill me in}
 # REQUEST_ACCESS_EMAIL_PREFIX={optional, email subject prefix}
 
 # SECRET_STORAGE_BACKEND={optional, name of ruby class to use as secret storage}
```
grosser: lets document the 2 choices here now ... and turn it into a section

Gemfile
```diff
@@ -32,6 +32,9 @@ gem 'active_model_serializers', '~> 0.8.0'
 gem 'sawyer', '~> 0.5'
 gem 'sse-rails-engine', '~> 1.4'
 
+# Hashicorp vault
+gem 'vault'
```
grosser: none of the other gems have comments ...

grosser: ... I guess after this one the do ... :D

app/models/secret_storage.rb
```diff
@@ -50,6 +50,43 @@ def self.keys
     end
   end
 
+  class HashicorpVault
+
+    SECRET_BACKEND ='secret/'.freeze
```
grosser: `PREFIX` ... or `FOLDER` ?
also should be `samson/` since everything in vault is secret ?

app/models/secret_storage.rb
```diff
@@ -50,6 +50,43 @@ def self.keys
...
+    # we don't really want other directories in here,
+    # and there may be other chars that we find we don't like
+    ENCODINGS = {"/": "%2F"}
+
+    def self.read(key)
+      result = Vault.logical.read(SECRET_BACKEND  + self.encode(key))
```
grosser: no need for self.

app/models/secret_storage.rb
```diff
@@ -50,6 +50,43 @@ def self.keys
...
+      result.data[:secret_data]
+    end
+
+    def self.write(key, data)
+      Vault.logical.write(SECRET_BACKEND + self.encode(key), secret_data: data)
+    end
+
+    def self.delete(key)
+      Vault.logical.delete(SECRET_BACKEND + key)
```
grosser: using the same code 3 times `SECRET_BACKEND + key` -> extract into a method ... will also fix the bug of delete not being able to remove with slashes ...

app/models/secret_storage.rb
irwaters: the path there actually tells vault what storage backend it's using.  the prefix secret/ is puts the stuff into that secret engine in vault.  I'll rename it to make it more clear.

app/models/secret_storage.rb
```diff
@@ -50,6 +50,43 @@ def self.keys
...
+  class HashicorpVault
+
+    SECRET_BACKEND ='secret/'.freeze
+    # we don't really want other directories in here,
+    # and there may be other chars that we find we don't like
+    ENCODINGS = {"/": "%2F"}
+
+    def self.read(key)
+      result = Vault.logical.read(SECRET_BACKEND  + self.encode(key))
+      result.data[:secret_data]
```
grosser: what happens if that fails ... the it is nil or empty hash ?

grosser: what other metadata is in there ? ... would it make sense to show it in the UI -> return all ?

app/models/secret_storage.rb
```diff
@@ -50,6 +50,43 @@ def self.keys
...
+    end
+
+    def self.write(key, data)
```
grosser: data is a hash with lots of things that are not secret ... feels strange to dump everything into secrets-data ... won't be usable via the cli / from kubernetes after

app/models/secret_storage.rb
```diff
@@ -50,6 +50,43 @@ def self.keys
...
+      Vault.logical.write(SECRET_BACKEND + self.encode(key), secret_data: data)
+    end
+
+    def self.delete(key)
+      Vault.logical.delete(SECRET_BACKEND + key)
+    end
+
+    def self.keys()
+      Vault.logical.list(SECRET_BACKEND).map! { |key| self.decode(key) }
```
grosser: no self. and no empty params for keys

app/models/secret_storage.rb
```diff
@@ -50,6 +50,43 @@ def self.keys
...
+    end
+
+    private
```
grosser: privacy is not enforced on self. methods, use `class << self`

```Ruby
class << self
  def read
  end

  private

  def actually_private
  end
end
```

app/models/secret_storage.rb
```diff
@@ -50,6 +50,43 @@ def self.keys
...
+
+    def self.encode(string)
+      ENCODINGS.each { |k, v| string.gsub!(k.to_s, v.to_s) }
+      string
+    end
+
+    def self.decode(string)
```
grosser: modifies the key -> need a bang ... but ideally don't ... `ENCODINGS.each { |k, v| string = string.gsub(v.to_s, k.to_s) }`

grosser: or start by string = string.dup and then do the gsub!

config/initializers/vault.rb
```diff
@@ -0,0 +1,19 @@
+VAULT_ENABLED = ENV.fetch('SECRET_STORAGE_BACKEND', false)
+if VAULT_ENABLED == 'SecretStorage::HashicorpVault'
+  require 'vault'
+  Rails.logger.info("Vault Client enabled")
+  Vault.configure do |config|
+    config.ssl_pem_file = ENV.fetch("VAULT_SSL_CERT")
```
grosser: makes it required -> change the docs

config/initializers/vault.rb
```diff
@@ -0,0 +1,19 @@
...
+    config.address = ENV.fetch("VAULT_ADDR", 'https://127.0.0.1:8200')
+
+    # Timeout the connection after a certain amount of time (seconds), also read
+    config.timeout = 30
+
+    # It is also possible to have finer-grained controls over the timeouts, these
+    # may also be read as environment variables
+    config.ssl_timeout  = 5
+    config.open_timeout = 5
+    config.read_timeout = 30
```
grosser: that seems crazy high for a key-value store ... 1s should be enough ?

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,44 @@
...
+        stub_request(:put, "https://127.0.0.1:8200/v1/secret%2Ffoo%252Fisbar").
+          to_return(status: 200, body: "", headers: {'Content-Type': 'application/json'})
+      end
+
+      it "deletes a key based on a key with /s" do
+        assert SecretStorage::HashicorpVault.write('foo/isbar', 'somevalue')
+      end
+    end
+  end
+
```
grosser: extra newline

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,44 @@
...
+          to_return(status: 200, body: '', headers: {'Content-Type': 'application/json'})
+      end
+
+      it "deletes a key based on a key with /s" do
+        assert SecretStorage::HashicorpVault.delete('foo/isbar')
+      end
+    end
+
+    describe ".delete" do
+      setup do
```
grosser: use before, not setup

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,44 @@
...
+    # as the model has already loaded it's config
+    # from ENV
+    describe ".read" do
+      setup do
+        data = {foo: 'bar', data: { secret_data:"bar"}}.to_json
+        stub_request(:get, "https://127.0.0.1:8200/v1/secret%2Ffoo%252Fisbar").
+          to_return(status: 200, body: data, headers: {'Content-Type': 'application/json'})
+      end
+
+      it "gets a value based on a key with /s" do
```
grosser: missing negative test

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,44 @@
...
+      end
+    end
+
+    describe ".delete" do
+      setup do
+        stub_request(:delete, "https://127.0.0.1:8200/v1/secret%2Ffoo%2Fisbar").
+          to_return(status: 200, body: '', headers: {'Content-Type': 'application/json'})
+      end
+
+      it "deletes a key based on a key with /s" do
```
grosser: missing negative test

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,44 @@
...
+        assert SecretStorage::HashicorpVault.delete('foo/isbar')
+      end
+    end
+
+    describe ".delete" do
```
grosser: .write

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,44 @@
...
+      setup do
+        stub_request(:put, "https://127.0.0.1:8200/v1/secret%2Ffoo%252Fisbar").
+          to_return(status: 200, body: "", headers: {'Content-Type': 'application/json'})
```
grosser: don't need the .to_return if it's the default

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,44 @@
...
+      end
+
+      it "deletes a key based on a key with /s" do
```
grosser: copy pasted description ?

app/models/secret_storage.rb
grosser: ahh ... should we still prefix samson- or something like that to not mess up the root namespace for others ?

grosser: I guess that would make fetching secrets kind of ugly because we would have to use samson- all the time ...

config/initializers/vault.rb
irwaters: turned it down.  of note, there are no defaults for these, so if we don't set them it'll fall back to whatever openssl/socket timeouts are set.

test/models/secret_storage_test.rb
irwaters: client/api returns true even when deleting a non-existent key.
[4] pry(main)> Vault.logical.delete('secret/nowaynohowman')
=> true

grosser: that's great :/

app/models/secret_storage.rb
irwaters: changed this around so that it'll raise like when the store is mysql etc.

app/models/secret_storage.rb
irwaters: no real choice.  you have directories, directories contain keys.  so there has to be some kinda default key, else we are in the business of trying to search the tree and find keys (and there is no recursive searching etc).

irwaters:
ok, think I got all your feedback processed.

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
 
+  class HashicorpVault
+
+    VAULT_SECRET_BACKEND ='secret/'.freeze
+    # we don't really want other directories in here,
+    # and there may be other chars that we find we don't like
+    ENCODINGS = {"/": "%2F"}
+
+    def self.read(key)
+      result = Vault.logical.read(gen_vault_path(key))
```
grosser: prefer real works like `generate` :)

grosser: or simply `vault_path`

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
+      return false if result.nil?
+      result.data[:secret_data]
+    end
```
grosser: is there more things in data dat we could display ?

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
+
+    def self.write(key, data)
+      Vault.logical.write(gen_vault_path(key), secret_data: data)
```
grosser: should only write `data[:vault]` to make the the read cli work ?

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
+    end
+
+    def self.delete(key)
+      Vault.logical.delete(gen_vault_path(key))
+    end
+
+    def self.keys()
```
grosser: prefer no empty `()`

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
+      Vault.logical.list(VAULT_SECRET_BACKEND).map { |key| encode_path(key) }
```
grosser: can do `map!` ?

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
+    end
+
+    private
+
+    def self.gen_vault_path(key = nil)
```
grosser: don't need the default ... and the nil handling can go away then too ?

app/models/secret_storage.rb
grosser: should remove the VAULT_SECRET_BACKEND prefix ?

grosser: otherwise the index -> edit links will not work

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
+      key.nil? ? false : VAULT_SECRET_BACKEND + encode_path(key, 'encode')
+    end
+
+    def self.encode_path(string = nil, direction = "decode")
+      return false if string.nil?
+      string = string.dup
+      if direction == 'encode'
```
grosser: use symbols for arguments/comparissons, they are faster and don't eat gc

config/initializers/vault.rb
```diff
@@ -0,0 +1,19 @@
...
+if VAULT_ENABLED == 'SecretStorage::HashicorpVault'
+  require 'vault'
+  Rails.logger.info("Vault Client enabled")
+  Vault.configure do |config|
+    config.ssl_pem_file = ENV.fetch("VAULT_SSL_CERT")
+    config.ssl_verify = ActiveRecord::Type::Boolean.new.type_cast_from_user(ENV.fetch("VAULT_SSL_VERIFY", true))
+    config.address = ENV.fetch("VAULT_ADDR", 'https://127.0.0.1:8200')
+
+    # Timeout the connection after a certain amount of time (seconds), also read
+    config.timeout = 30
```
grosser: lower that ?

config/initializers/vault.rb
grosser: ^

test/models/secret_storage_test.rb
grosser: ^

app/models/secret_storage.rb
irwaters: needs it to list all the keys at the root of the storage engine.  nav works as intended.

irwaters:
getting there.

irwaters:
when you have a sec/feel up to it :)

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
+      Vault.logical.list(VAULT_SECRET_BACKEND).map! { |key| encode_path(key) }
+    end
+
+    private
+
+    def self.vault_path(key)
+      VAULT_SECRET_BACKEND + encode_path(key, :encode)
+    end
+
+    def self.encode_path(string = nil, direction = :decode)
```
grosser: don't need the nil default and return guard ... don't need the default direction ...

app/models/secret_storage.rb
```diff
@@ -50,6 +50,49 @@ def self.keys
...
+    def self.write(key, data)
+      Vault.logical.write(vault_path(key), vault: data)
+    end
+
+    def self.delete(key)
+      Vault.logical.delete(vault_path(key))
+    end
+
+    def self.keys()
+      Vault.logical.list(VAULT_SECRET_BACKEND).map! { |key| encode_path(key) }
```
grosser: default for encode_path is to decode ... wut ?

irwaters:
fixed up.

.env.example
```diff
@@ -67,4 +67,8 @@ GITHUB_TOKEN={fill me in}
 # REQUEST_ACCESS_EMAIL_ADDRESS_LIST={optional, space separated list of email addresses (managers mailing list, JIRA, etc.)}
 # REQUEST_ACCESS_EMAIL_PREFIX={optional, email subject prefix}
 
-# SECRET_STORAGE_BACKEND={optional, name of ruby class to use as secret storage}
+# SECRET_STORAGE_BACKEND={required, should be one of: SecretStorage::DbBackend or SecretStorage::HashicorpVault }
```
grosser: it is not required and will default to DbBackend

app/models/secret_storage.rb
```diff
@@ -50,6 +50,47 @@ def self.keys
...
+    ENCODINGS = {"/": "%2F"}
+
+    def self.read(key)
+      result = Vault.logical.read(vault_path(key))
+      return false if result.nil?
+      result.data[:vault]
+    end
+
+    def self.write(key, data)
+      Vault.logical.write(vault_path(key), vault: data)
```
grosser: I don't think we should write the complete hash into vault, that will make reading the value via consul / cli complicated

app/models/secret_storage.rb
```diff
@@ -50,6 +50,47 @@ def self.keys
...
+    end
+
+    private
+
+    def self.vault_path(key)
+      VAULT_SECRET_BACKEND + encode_path(key)
+    end
+
+    def self.encode_path(string, direction = :encode)
+      if direction == :decode
```
grosser: modifies the key, needs a dup or strange things might happen ...

config/initializers/vault.rb
grosser: ^

test/models/secret_storage_test.rb
grosser: ^

app/models/secret_storage.rb
```diff
@@ -50,6 +50,47 @@ def self.keys
...
+
+    VAULT_SECRET_BACKEND ='secret/'.freeze
+    # we don't really want other directories in here,
+    # and there may be other chars that we find we don't like
+    ENCODINGS = {"/": "%2F"}
+
+    def self.read(key)
+      result = Vault.logical.read(vault_path(key))
+      return false if result.nil?
+      {data: result.data[:vault]}
```
grosser: should be `value:` or the read won't be able to get the value

irwaters:
now returning the entire payload from vault.

app/models/secret_storage.rb
```diff
@@ -50,6 +50,51 @@ def self.keys
...
+    # we don't really want other directories in here,
+    # and there may be other chars that we find we don't like
+    ENCODINGS = {"/": "%2F"}
+
+    def self.read(key)
+      result = Vault.logical.read(vault_path(key))
+      return false if result.nil?
+      result = result.dup.to_h
+      # making sure that the delete won't blow up
+      result[:value] = result[:data][:vault]
```
grosser: that now has a hash with another nested hash and the secret is in both ...

result.merge(result.delete(:data))
result[:value] = result.delete(:vault)

app/models/secret_storage.rb
```diff
@@ -50,6 +50,51 @@ def self.keys
...
+      Vault.logical.list(VAULT_SECRET_BACKEND).map! { |key| encode_path(key, :decode) }
+    end
+
+    private
+
+    def self.vault_path(key)
+      VAULT_SECRET_BACKEND + encode_path(key)
+    end
+
+    def self.encode_path(string, direction = :encode)
```
grosser: call this covert_path and make the direction mandatory ? `covert_path key, :encode` + `covert_path key, :decode`

config/initializers/vault.rb
grosser: still the same

test/models/secret_storage_test.rb
grosser: still missing

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,54 @@
...
+    describe ".read" do
+      before do
+        data = {foo: 'bar', data: { vault:"bar"}}.to_json
+        stub_request(:get, "https://127.0.0.1:8200/v1/secret%2Ffoo%252Fisbar").
+          to_return(status: 200, body: data, headers: {'Content-Type': 'application/json'})
+      end
+
+      it "gets a value based on a key with /s" do
+        response = SecretStorage::HashicorpVault.read('foo/isbar')
+        response[:data].must_equal({:vault=>"bar"})
```
grosser: result should be a flash hash with a value key

grosser: -> assert on the full hash to not let anything slip through

irwaters:
with all your spare time...

grosser:
VAULT_SSL_CERT is still required atm and document as optional ... already mentioned it 3 times

grosser:
same for negative read test (fails to read)

irwaters:
there ya go

app/models/secret_storage.rb
```diff
@@ -50,6 +49,51 @@ def self.keys
...
+      result = Vault.logical.read(vault_path(key))
+      return false if result.nil?
+      result = result.to_h
+      result = result.merge(result.delete(:data))
+      result[:value] = result.delete(:vault)
+      result
+    end
+
+    def self.write(key, data)
+      Vault.logical.write(vault_path(key), vault: data[:value])
```
grosser: what do you think about writing updater/creator too ?

test/models/secret_storage_test.rb
```diff
@@ -91,4 +91,61 @@
...
+        stub_request(:get, "https://127.0.0.1:8200/v1/secret%2Fnotgoingtobethere").
+          to_return(status: 200, body: fail_data, headers: {'Content-Type': 'application/json'})
+      end
+
+      it "gets a value based on a key with /s" do
+        SecretStorage::HashicorpVault.read('foo/isbar').must_equal({:lease_id=>nil, :lease_duration=>nil, :renewable=>nil, :auth=>nil, :value=>"bar"})
+      end
+
+      it "fails to read a key" do
+        SecretStorage::HashicorpVault.read('notgoingtobethere').must_equal({:lease_id=>nil, :lease_duration=>nil, :renewable=>nil, :auth=>nil, :value=>nil})
```
grosser: read miss is supposed to raise an `ActiveRecord::RecordNotFound`

grosser:
👍 except for the read-miss ... that should end up rendering our standard 404 page -> ActiveRecord::RecordNotFound

irwaters:
added raise ActiveRecord::RecordNotFound on read miss and test.

app/models/secret_storage.rb
```diff
@@ -50,6 +49,51 @@ def self.keys
...
+
+    def self.vault_path(key)
+      VAULT_SECRET_BACKEND + convert_path(key, :encode)
+    end
+
+    def self.convert_path(string, direction)
+      string = string.dup
+      if direction == :decode
+        ENCODINGS.each { |k, v| string.gsub!(v.to_s, k.to_s) }
+      elsif direction == :encode
```
grosser: ideally either `else` or `elsif + else raise` ... allowing unknown or matching twice is a bit strange ...

grosser:
👍

PR #864, opened 2016-04-21
Fix log styling if using color in log output

mblaschke:
Merge #807 introduced some display issues with colored logs which is fixed with this commit

Before:
image

After:
image

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low (only log output)

### DIFF
--- a/app/assets/javascripts/streams.js
+++ b/app/assets/javascripts/streams.js
@@ -20,7 +20,7 @@ function startStream() {
 
     var addLine = function(data) {
       var msg = JSON.parse(data).msg;
-      $messages.append(msg);
+      $messages.append(msg + "\n");
       if (following) {
         $messages.scrollTop($messages[0].scrollHeight);
       }
--- a/app/assets/stylesheets/deploys.scss
+++ b/app/assets/stylesheets/deploys.scss
@@ -140,7 +140,6 @@ td.status {
   white-space: pre-wrap;
   > span {
     display: inline-block;
-    width: 100%;
     cursor: pointer;
 
     &.highlighted {


grosser:
does hat work when straming / stopped streaming / reload too ? ... @sandlerr

mblaschke:
Doesn't work with streaming :(
image

grosser:
need more creative hacks :D

mblaschke:
@grosser Fixed it :)

grosser:
worked locally ... so let's give that a try :)

sandlerr:
did anyone test if it support this: https://github.com/zendesk/samson/pull/864/files#diff-4efb82dfc1c9d13dd57f0445aee474b2L65 ?
That's when some program is showing interesting progress output by over-writing the same line repeatedly

grosser:
tried to use:

```
echo -e '#####                     (33%)\r\c'
sleep 1
echo -e '#############             (66%)\r\c'
sleep 1
echo -e '#######################   (100%)\r\c'
echo -e '\n'
```

would did not work before or after :(

sandlerr:
so there was always a newline after each line of input was evaluated (which is a minor bug). A single command that overwrote its own line would have worked fine though. You might be able to join those echo statements with `&&` to get it to work.

grosser:
`echo -e '1\r\c' && sleep 2 && echo -e '2\r\c' && sleep 2 && echo -e '3\r\c' && sleep 2 && echo -e '4\r\c' && sleep 2 && echo -e '5\r\c' && sleep 2 && echo -e '6\r\c' && sleep 2 && echo -e '7\r\c'` does not generate any output on any version and the data does not reach the js ... need a better example :D

grosser:
big git clone also shows no progress ... so that's just broken :)

grosser:
might be the switch to IO.popen ...

grosser:
damn ... reverting 2bf25bdfb123b4e12997ee3e49e7eee5789b955c fixes it ... and the \n breaks it :D

grosser:
fix in the works https://github.com/zendesk/samson/pull/867

Issue #768, opened 2016-03-22
Export PROJECT as variable


mblaschke:
REVISION, TAG, STAGE is exported before deployment commands are executed but not the PROJECT.

Would be nice if PROJECT would be also available.

jonmoter:
Could you add a command saying something like `export PROJECT="myprojectname"`, and add that command to each stage of your project?

mblaschke:
`URL`, `DEPLOY_URL`, `DEPLOYER` etc are also exported per default so it would be nice to get the project name, repository url and permalink for eg. automatic chat status reporting.

We would like to add some default tasks in our [webdevops/samson-deployment service](https://github.com/webdevops/samson-deployment) (based on your docker image) eg. for slack and hipchat reporting so that the user don't need to setup any env vars :)

BTW: Thanks for the great work :)

PR #865, opened 2016-04-21
Add project environment variables for deployment

mblaschke:
Add project environment variables for deployment:
* PROJECT_NAME
* PROJECT_PERMALINK
* PROJECT_REPOSITORY

Will or can be used for automatic notification hooks (eg. slack/hipchat).

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

Fixes zendesk/samson#768


grosser:
name / repo kind of make sense ... not sure why project-permalink would be needed ... deploy_url includes that and is what should be used for notifications ?

mblaschke:
We're using the permalink as project key and the project name for a speaking name.

PR #868, opened 2016-04-22
fix heroku deployment not having mysql available

grosser:
running precompile / js without database set up works just fine ...

@zendesk/samson 

### Risks
- Medium: deploy/test issues

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,12 @@ branches:
   only: master
 env:
   matrix:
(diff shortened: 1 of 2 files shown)


pschambacher:
👍

grosser:
fails to deploy ... Could not load database configuration. No such file - ["config/database.yml"]
so we have to set a DATABSE_URL according to the currently available gems :/

PR #871, opened 2016-04-22
SAMSON-226 cannot reuse a recently approved deploy when the stage was changed after the deploy

grosser:
@zendesk/samson

### Risks
- Medium: previously approved deploys being unable to reuse


pschambacher:
👍

PR #872, opened 2016-04-22
make precompile and js testing work when no database is set up

grosser:
@zendesk/samson 

revised fix for https://github.com/zendesk/samson/pull/868

### Risks
- None

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,7 +12,15 @@ task :asset_compilation_environment do
   ENV['SECRET_TOKEN'] = 'foo'
   ENV['GITHUB_TOKEN'] = 'foo'
   ENV['PRECOMPILE'] = '1'
-  ENV['DATABASE_URL'] = 'mysql2://none@none/none'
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -12,7 +12,15 @@ task :asset_compilation_environment do
...
-  ENV['DATABASE_URL'] = 'mysql2://none@none/none'
+
+  config = Rails.application.config
+  def config.database_configuration
+    {}
+  end
+
+  ar = ActiveRecord::Base
+  def ar.establish_connection
+  end
```
pschambacher: That's not going to make any problem with migrations and what not?

pschambacher:
👍

Rakefile
grosser: migrations don't use asset_compilation_environment ... but it would blow up when doing `rake assets:compile db:migrate` ... hopefully nobody does that :D

PR #873, opened 2016-04-22
improve method naming

grosser:
### Risks
- None

### DIFF
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -12,7 +12,7 @@ class DeployService
     if deploy.persisted?
       send_sse_deploy_update('new', deploy)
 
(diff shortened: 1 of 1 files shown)


app/models/deploy_service.rb
```diff
@@ -56,7 +56,7 @@ def latest_approved_deploy(reference, project)
       .detect { |d| d.production? && !d.bypassed_approval? }
   end
 
-  def release_approved?(deploy)
+  def copy_approval_from_last_deploy(deploy)
```
pschambacher: Question mark for convention?

pschambacher:
👍

app/models/deploy_service.rb
grosser: it has side-effects, so I don't want to use `?`

PR #874, opened 2016-04-22
enable running multiple samsons on the same base domain without havin…

grosser:
ng csrf errors from colliding cookies

@zendesk/samson 

# Risks
 - low: all users will be logged out


irwaters:
👍

grosser:
fixed samson-staging access for me ...

PR #875, opened 2016-04-22
SAMSON-231 make bypass unable to click by accident

grosser:
@zendesk/samson 
@pdeuter 

### Risks
- Low: unable to bypass

### DIFF
--- a/app/views/deploys/_buddy_check.html.erb
+++ b/app/views/deploys/_buddy_check.html.erb
@@ -7,7 +7,8 @@
     <div class="deployer-stop">
       <%= stop_button %>
       or
-      <%= link_to "Bypass", buddy_check_project_deploy_path(@project, @deploy), method: :post, class: "btn btn-danger" %>
+      <% confirm = "Are you sure this is an emergency and you cannot find a buddy?\n#{strip_tags(details)}" %>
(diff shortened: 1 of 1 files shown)


grosser:
screen shot 2016-04-22 at 2 18 03 pm
screen shot 2016-04-22 at 2 17 58 pm

app/views/deploys/_buddy_check.html.erb
```diff
@@ -7,7 +7,8 @@
     <div class="deployer-stop">
       <%= stop_button %>
       or
-      <%= link_to "Bypass", buddy_check_project_deploy_path(@project, @deploy), method: :post, class: "btn btn-danger" %>
+      <% confirm = "Are you sure this is an emergency and you cannot find a buddy?\n#{details}" %>
```
sandlerr: since details is marked as html_safe above would probably make sense to [strip_tags](http://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-strip_tags) here

sandlerr:
:+1: on green

app/views/deploys/_buddy_check.html.erb
grosser: good catch!

PR #876, opened 2016-04-23
track stage versions

grosser:
@zendesk/samson

@apanzerj for paperclip hacks

screen shot 2016-04-22 at 6 28 35 pm
screen shot 2016-04-22 at 6 28 21 pm

### Risks
- Medium: stage save failing


config/initializers/paper_trail.rb
```diff
@@ -0,0 +1,27 @@
+# record what was done in the console
+if defined?(::Rails::Console)
+  PaperTrail.whodunnit = "#{`whoami`.strip}: console"
```
sandlerr: do you want this to run if we are inside a rake task also?

app/models/stage.rb
```diff
@@ -162,6 +164,15 @@ def command_updated_at
     ].compact.max
   end
 
+  def record_script_change
+    send :record_update, true
```
sandlerr: don't need to send?

grosser:
fixed up

app/controllers/sessions_controller.rb
```diff
@@ -1,7 +1,7 @@
 require 'omniauth/github_authorization'
 
 class SessionsController < ApplicationController
-  skip_before_action :login_user
```
apanzerj: why is this all over the place?

grosser: change from before to aroud

test/controllers/concerns/current_user_test.rb
```diff
@@ -0,0 +1,42 @@
+require_relative '../../test_helper'
+
+SingleCov.covered! uncovered: 2
+
+describe "CurrentUser Controller" do
```
sandlerr: worth being more obvious that this is testing a module (not an actual controller)?

test/controllers/concerns/current_user_test.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  tests CurrentUserTestController
+  use_test_routes
+
+  as_a_viewer do
+    it "knows who did something" do
+      PaperTrail.with_whodunnit nil do
+        PaperTrail.whodunnit.must_equal nil
+        get :whodunnit, test_route: true
+        PaperTrail.whodunnit.must_equal nil
```
sandlerr: I feel like this assertion could be a different `it` block. `it 'doesnt assign a user to something they didnt do' do ...`

test/controllers/versions_controller_test.rb
```diff
@@ -0,0 +1,31 @@
...
+      PaperTrail.with_logging do
+        stage.update_attribute(:name, 'Fooo')
+      end
+    end
+  end
+
+  let(:stage) { stages(:test_staging) }
+
+  as_a_viewer do
+    describe "#inde" do
```
sandlerr: nitpick `#index`

test/models/stage_test.rb
```diff
@@ -435,4 +435,38 @@
...
+      stage.update_attributes(order: 5, updated_at: 1.second.from_now)
+      stage.versions.size.must_equal 0
+    end
+
+    it "records script" do
+      stage.record_script_change
+      YAML.load(stage.versions.first.object)['script'].must_equal stage.script
+    end
+
+    it "can restore ... but loses script" do
```
sandlerr: how would we restore the script? Can this test use variables like `old_name`, `new_name` to make it more obvious? Otherwise need to look at the fixture to see where `Staging` came from

sandlerr:
comments only on tests so :+1:

is a HTML view the best fit for your usecase? There might be a lot of changes, wondering if JSON might be better so users can filter it programmatically.

test/controllers/concerns/current_user_test.rb
grosser: muahaha `describe "CurrentUser included in controller" do`

test/models/stage_test.rb
grosser: restoring script is tricky since you'd have to split it into commands and create new commands where necessary and set correct ordering ... nothing impossible, but not something I wanted to tackle :)

grosser:
all updated

app/controllers/versions_controller.rb
```diff
@@ -0,0 +1,5 @@
+class VersionsController < ApplicationController
+  def index
+    @versions = PaperTrail::Version.where(item_id: params.require(:item_id), item_type: params.require(:item_type))
```
sandlerr: thinking more about this, how confident are we that PaperTrail won't ever `const_get(item_type)`? Do we need a whitelist?

grosser: if nothing is stored that matches this type then nothing will be found ...
we might have to later lock this down if we only want certain types ...

PR #878, opened 2016-04-23
give timeout tests a second chance

sandlerr:
@zendesk/samson lots of timeout errors in the test run now, let's give them a second chance.


grosser:
I prefer retry to copy-paste ... but not too bad 👍

sandlerr:
Changed to use `retry`, didn't know that existed. :+1:?

grosser:
👍

PR #879, opened 2016-04-25
do not blow up on unconfigured kube config

grosser:
and improve error messages when config is broken

@irwaters


irwaters:
👍

jonmoter:
Thanks for the fix and adding more tests!  👍

PR #880, opened 2016-04-25
blow up cubernetes without docker

grosser:
@irwaters


irwaters:
👍

grosser:
changed to make the controller render the error message so the plugin stays useable/loadable like all the other plugins ...

jonmoter:
👍

PR #884, opened 2016-04-26
current-user + authorization cleanup and unification

grosser:
@zendesk/samson 

### Risks
- Low: unauthorized on some actions


jonmoter:
👍

PR #885, opened 2016-04-26
only check docker when docker is enabled and make it fail fast instea…

grosser:
…d of hanging

reproduce by taking down docker but leaving DOCKER_URL set

Discussion: 
 - should we enforce DOCKER_URL to be set ?
 - DOCKER_REGISTRY env var is not documented

# Risks
 - Failing to boot when docker is too slow

### DIFF
--- a/config/initializers/docker.rb
+++ b/config/initializers/docker.rb
@@ -1,16 +1,15 @@
 require 'docker'
 
-if !Rails.env.test? && !ENV['PRECOMPILE']
-  if ENV['DOCKER_URL'].present?
-    Docker.url = ENV['DOCKER_URL']
-    Docker.options = { read_timeout: 600 }
-
-    # Confirm the Docker daemon is a recent enough version
-    Docker.validate_version!
-  end
-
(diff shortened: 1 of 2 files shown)


grosser:
@zendesk/samson

config/initializers/docker.rb
```diff
@@ -1,16 +1,14 @@
...
+    exit 1
   end
 
-  if ENV['DOCKER_FEATURE'] && !ENV['DOCKER_REGISTRY'].present?
-    puts '*** DOCKER_REGISTRY environment variable must be configured when DOCKER_FEATURE is enabled ***'
-    exit(1)
+  if (url = ENV['DOCKER_URL']).presence
+    Docker.url = url
+    Docker.options = { read_timeout: 600, connect_timeout: 2 }
+    Docker.validate_version! # Confirm the Docker daemon is a recent enough version
```
jonmoter: validate even if DOCKER_URL is not set. (scenario where using a unix socket instead of a TCP connection)

irwaters:
:+1

Issue #849, opened 2016-04-19
Docker integration does not work with Amazon ECR


rata:
Hi,

Amazon ECR does not uses a user/pass auth. You need to run something like: 

```
aws ecr get-login --region=us-west-2
```

in the aws-shell and it gives you a user/pass that is valid for ~12 hours. So, setting the DOCKER_REGISTRY_USER and DOCKER_REGISTRY_PASS env variables at startup isn't really an option.

Is there any plan or idea how to integrate support for Amazon ECR in docker builds?

grosser:
afaik there are no plans to support it ... how about setting `DOCKER_REGISTRY_AUTH` to the command that will return user/password ?

rata:
And use that env var how/when? Like set the user and pass env vars from the
output of that command that is run before running any docker build?

The command returns a docker command, something like:

```
docker login -u AWS -p <some-pass> -e none https://<your-id>.dkr.ecr.us-west-2.amazonaws.com
```

When running this, it stores the credentials in ~/.docker/config.json. What do
you think of using the config.json file?

grosser:
could use that if user/pass is not set in the env ... would that fix your issue / how would you make sure it's filled prior to being needed ?

maybe do a hacky PR with a solution that would work and then we take the discussion thee ?

jonmoter:
We don't have a solution yet, but if we wanted to support it, here's one possibility.

With the [docker gem](https://github.com/swipely/docker-api#global-calls) we're using, you can authenticate with the registry by calling:

```
Docker.authenticate!('username' => 'MY_ACCOUNT', 'password' => `MY_PASSWORD`)
```

and then it will use those credentials when connecting to the registry.

As far as I can tell from the docs, the tokens that you get back from Amazon last for 12 hours, so they need to be regularly updated.  So, if we wanted to support this, we could include the aws-sdk gem to call [get-login](http://docs.aws.amazon.com/sdkforruby/api/Aws/ECR/Client.html#get_authorization_token-instance_method), and have a thread that does something like:

```ruby
ecr_client = Aws::ECR::Client.new(region: 'us-west-2', credentials: somethingsomething)
loop do
  response = ecr_client.get_authorization_token
  auth_token = response.authorization_data[0].authorization_token
  Docker.authenticate!('username' => 'AWS', 'password' => auth_token)
  sleep(6.hours)
end
```

I haven't tested this, but in theory, something like this could work?

[...]

PR #886, opened 2016-04-26
Get user/pass from AWS ECR for each docker build

jbarreneche:
* Allow using AWS ECR as docker registry (the problem with this registry is that it requires that you refresh the password at least every 12hs)

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

Fixes: #849


plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+  end
+end
+
+Samson::Hooks.callback :before_docker_build do
+  begin
+    authorization_token = SamsonAwsEcr::Engine.ecr_client.get_authorization_token
+    authorization_data  = authorization_token.authorization_data.first
+    user, pass = Base64.decode64(authorization_data.authorization_token).split(":")
+    ENV['DOCKER_REGISTRY_USER'] = user
```
grosser: multiple deploys will override each other ?

plugins/aws_ecr/samson_aws_ecr.gemspec
```diff
@@ -0,0 +1,6 @@
+Gem::Specification.new "samson_aws_ecr", "0.0.0" do |s|
+  s.summary = "Add Support for AWS ECR Docker registry"
+  s.authors = [""]
+  s.email = ""
+  s.add_runtime_dependency "aws-sdk"
```
grosser: can you reduce this to `aws-sdk-core` ?

plugins/aws_ecr/test/samson_aws_ecr/samson_plugin_test.rb
```diff
@@ -0,0 +1,40 @@
+require_relative '../test_helper'
+
+SingleCov.covered! unless defined?(Rake) # rake preloads all plugins
+
+describe SamsonDockerb do
+  let(:stage) { stages(:test_staging) }
+  let(:ecr_client) { Aws::ECR::Client.new(stub_responses: true, region: 'us-west-2') }
```
grosser: sweet did not know about `stub_responses` :D

plugins/aws_ecr/test/samson_aws_ecr/samson_plugin_test.rb
```diff
@@ -0,0 +1,40 @@
+require_relative '../test_helper'
+
+SingleCov.covered! unless defined?(Rake) # rake preloads all plugins
+
+describe SamsonDockerb do
```
grosser: wrong plugin

plugins/aws_ecr/test/samson_aws_ecr/samson_plugin_test.rb
```diff
@@ -0,0 +1,40 @@
...
+          authorization_token: base64_authorization_token
+        ]
+      })
+
+      fire
+
+      ENV['DOCKER_REGISTRY_USER'].must_equal username
+      ENV['DOCKER_REGISTRY_PASS'].must_equal password
+    end
+
```
grosser: extra newline

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
```diff
@@ -0,0 +1,21 @@
+require 'aws-sdk'
+
+module SamsonAwsEcr
+  class Engine < Rails::Engine
+    mattr_accessor :ecr_client do
+      Aws::ECR::Client.new(region: ENV['AWS_REGION'] || "us-west-2")
```
grosser: prefer no default region and `ENV.fetch('AWS_REGION')` ?

plugins/aws_ecr/README.md
```diff
@@ -0,0 +1,19 @@
...
+Because Samson only supports to set credentials through the environment, this plugin
+runs a callback before every build requesting AWS for the new credentials. Then sets
+these credentials to the environment so the normal docker build process can use them.
+
+To configure this plugin you need to:
+
+* Enable docker in samson (DOCKER_FEATURE=1)
+* Set your ECR registry (DOCKER_REGISTRY=<account>.dkr.ecr.<aws-region>.amazonaws.com)
+* Set your [AWS credentials](http://docs.aws.amazon.com/sdkforruby/api/#Configuration)
+* Set your ECR registry region (AWS_REGION=<aws-region>)
```
grosser: AWS_REGION is a bit global ... maybe AWS_ECR_REGION ?

grosser:
fyi @jonmoter

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
jonmoter: How about you do something like store the timestamp of the last time you got a token as a `mattr_accessor`, and only get a new token if the old one is older than 8 hours, or something like that?

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
jbarreneche: I had a problem when setting up the tests. 
`AWS_REGION` is the default variable `aws-sdk` uses for selecting the region. But I couldn't find a good place to set this variable (because `mattr_accessor` initializes when this file is loaded). Do you have a preferred way for doing this kind of client configuration?

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
jonmoter: Also, please only run this code if `ENV['DOCKER_REGISTRY']` matches a regex for an ECR registry.

I frequently run Samson with `PLUGINS=all`, and I don't want it to try to authenticate with ECR if I'm not pointing to an ECR registry.

jbarreneche: Yes, but Samson currently supports only a pair of credentials for docker repository (and I checked that AWS returns the same credentials while they are valid). So, I don't think this might be a problem (other than sometimes resquesting new credentials isn't really necessary)

plugins/aws_ecr/samson_aws_ecr.gemspec
jbarreneche: done!

plugins/aws_ecr/test/samson_aws_ecr/samson_plugin_test.rb
jbarreneche: ups!
Thx!

plugins/aws_ecr/README.md
jbarreneche: I went for `AWS_REGION` because it's the default for `aws-sdk` (and I'm pretty sure I saw other places of samson using the default gem env value). Anyway, I can change it to something more specific if you want.

jbarreneche: I changed the behavior (because of another comment), now I take the region from the DOCKER_REGISTRY. I'll update the doc

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
```diff
@@ -0,0 +1,32 @@
...
+        SamsonAwsEcr::Engine.ecr_client = Aws::ECR::Client.new(region: match['region'])
+      end
+    end
+
+  end
+end
+
+Samson::Hooks.callback :before_docker_build do
+  begin
+    if SamsonAwsEcr::Engine.ecr_client && SamsonAwsEcr::Engine.credentials_expire_at < Time.current
```
grosser: maybe add some padding (+5 minutes) so it does not fail mid deploy

jbarreneche: I'm not sure how the AWS credential work. But, if they haven't expired, AWS returns the same credentials (that will have the same expires_at)... so... I'm not sure if that would be of any help. I could wait for the credentials to expire (but it seems too hackish)

grosser: I guess we'll how often it breaks during deploys :)

grosser:
👍  nice docs and tests :)

jbarreneche:
Thanks!! 😄

PR #889, opened 2016-04-26
use socket options for extensions client

grosser:
followup to https://github.com/zendesk/samson/pull/673 ... looks like a bug to me that we support socket in one but not the other client ... @henders / @msufa please discuss and hit merge ...

### Risks
- Low: etension client failing


henders:
lgtm 👍

henders:
I think it wasn’t a bug only because the extension_client didn’t need any streaming http session like the watcher threads in celluloid.

henders:
but good to be consistent

PR #890, opened 2016-04-26
fix create redirecting to index with format: id cluster.3 ... and cre…

grosser:
…ate blowing up when save fails

@zendesk/paas 

# Risks
 - Low: cluster create failing


jonmoter:
👍

PR #895, opened 2016-04-27
fix migration

sandlerr:
@ocke to fix our staging deploy

PR #896, opened 2016-04-27
verify we are not using bad indexes

grosser:
see https://github.com/zendesk/samson/pull/895

@sandlerr 

this fails on 10+ different tables ... why are they ok to use index without length ?

for example builds has git_sha index :(

### DIFF
--- a/test/integration/cleanliness_test.rb
+++ b/test/integration/cleanliness_test.rb
@@ -133,4 +133,22 @@ describe "cleanliness" do
       untested: untested
     )
   end
+
+  it "uses short index on string columns" do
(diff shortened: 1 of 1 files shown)


sandlerr:
probably because they're utf8 in our staging deployment. If we dropped the db and tried to recreate as utf8mb4 it would probably fail. I'm not sure why https://github.com/zendesk/samson/blob/master/config/initializers/active_record.rb isn't helping, maybe we just need to set that variable. We've also already got a test [here](https://github.com/zendesk/samson/blob/master/test/integration/cleanliness_test.rb#L28) that also failed to help :/

PR #897, opened 2016-04-27
bumping docker api version in docker binary builder

fneves:
Bumping the of the docker binary plugin to the docker api 1.22.
This will make the cache work with the new version.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


henders:
👍 

Later we should modify this check so that it matches higher versions automatically.

fneves:
I agree...but... we are manually checking for breaking changes in the api before the bump...
We can always let it crash and then fix it, unless we find a way of doing some integration tests with a docker host.

PR #898, opened 2016-04-27
Making the project search fuzzy, making the time formatting respect what

irwaters:
Few changes that praveen wanted.
make the project name search fuzzy
make the default time respect what the user has in their profile
put the 'Recent Deploys' tab back

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
https://zendesk.atlassian.net/browse/SAMSON-233

### Risks
Low
the user has in their profile (by default), and changing the name of the
tab back (all changes that praveen wanted).


app/views/shared/_header.html.erb
```diff
@@ -41,7 +41,7 @@
           </li>
         <% end %>
         <li class="<%= 'active' if current_page?(controller: '/deploys', action: 'search') %>">
-          <%= link_to "Deploys", search_deploys_path %>
+          <%= link_to "Recent Deploys", search_deploys_path %>
```
grosser: seems redundant ... why do we need extra works here ?

app/helpers/application_helper.rb
```diff
@@ -54,6 +54,8 @@ def relative_time(time)
   end
 
   def render_time(time, format)
+    # grab the time format that the user has in their profile
+    format = current_user.time_format if format.nil?
```
grosser: test ?

app/controllers/deploys_controller.rb
```diff
@@ -55,11 +55,11 @@ def search
     end
 
     if params[:deployer].present?
-      users = User.where("name LIKE ?", "%#{ActiveRecord::Base.send(:sanitize_sql_like, params[:deployer])}%").pluck(:id)
+      users = User.where("users.name LIKE ?", "%#{ActiveRecord::Base.send(:sanitize_sql_like, params[:deployer])}%").pluck(:id)
     end
 
     if params[:project_name].present?
```
grosser: FYI I'd prefer a `if name = params[:project_name].presence` and then reuse name ... but either way is fine

grosser:
👍

app/views/shared/_header.html.erb
irwaters: I was told it was confusing and that people wanted the old one back.

irwaters:
@grosser when you have a sec would appreciate your thoughts on this.  essentially I've moved to grabbing the user's timezone from the browser and setting it as a cookie so rails can get to it.  that way time formatting nicely changes if a user changes their locale

app/views/shared/_header.html.erb
grosser: Is it just confusing because people are used to the old way or because it is actually confusing (to new people) ?
If it's the first I'd vote for keeping it ...

test/helpers/application_helper_test.rb
```diff
@@ -248,11 +248,17 @@
...
     end
 
-    it "formats time in local" do
-      render_time(ts, 'local').must_equal "<time datetime=\"2016-04-18T17:46:10+00:00\">April 18, 2016  5:46 PM</time>"
+    it "formats local time in America/Los_Angeles" do
+      cookies[:timezone] = 'America/Los_Angeles'
+      render_time(ts, 'local').must_equal "<time datetime=\"2016-04-18 10:46:10 -0700\">2016-04-18 10:46:10 -0700</time>"
+    end
+
+    it "formats local time in America/New_York" do
```
grosser: how is this covering current_user / nil format usage ?

app/views/shared/_header.html.erb
```diff
@@ -86,3 +86,5 @@
     </ul>
   </div>
 </div>
+
+<%= javascript_include_tag "cookies" %>
```
grosser: that looks wrong ...

app/assets/javascripts/cookies.js
```diff
@@ -0,0 +1,3 @@
+var tz = jstz.determine();
+Cookies.set("timezone", tz.name());
```
grosser: how is setting a timezone cookie helping ?

use jquery to do it ? `$.cookie("test", 1);
`

grosser:
rails can real the users timezone from the request env ... so we don't need js + cookies ?

app/assets/javascripts/cookies.js
irwaters: we didn't have the jquery cookie plugin, we do now.

app/views/shared/_header.html.erb
irwaters: you are correct, should get picked up by the //= require_tree .

irwaters:
not seeing it in request.env (though I could be blind).

grosser:
hmm not trivial ... some hacky ways with geocoding ... but asking js seems to be the way to go :(
http://stackoverflow.com/questions/12114356/how-to-get-the-request-timezone

app/assets/javascripts/jquery.cookie.js
```diff
@@ -0,0 +1,117 @@
+/*!
```
grosser: prefer using a gem: https://rails-assets.org/#/components/jquery-cookie

Gemfile
```diff
@@ -31,6 +31,7 @@ gem 'active_model_serializers', '~> 0.8.0'
 gem 'paper_trail'
 gem 'sawyer', '~> 0.5'
 gem 'sse-rails-engine', '~> 1.4'
+gem 'jsTimezoneDetect-rails'
```
grosser: prefer rails-assets https://rails-assets.org/#/components/jstimezonedetect

app/helpers/application_helper.rb
```diff
@@ -54,10 +54,12 @@ def relative_time(time)
   end
 
   def render_time(time, format)
+    # grab the time format that the user has in their profile
+    format = current_user.time_format ||= 'utc' if format.nil?
```
grosser: helper is not supposed to change model data `||=` -> `||`

grosser: format ||= current_user.time_format || 'utc'

app/helpers/application_helper.rb
```diff
@@ -54,10 +54,12 @@ def relative_time(time)
...
     if format == 'local'
-      time_tag(time, format: '%B %d, %Y %l:%M %p')
+      content_tag(:time, time.in_time_zone(cookies[:timezone] || 'UTC').to_s, datetime: time.in_time_zone(cookies[:timezone] || 'UTC'))
```
grosser: how about not repeating logic multiple times ?

test/helpers/application_helper_test.rb
grosser: ah, this is the cookie usage ... then a `it "reads timezone from the cookie set via javascript"` might be a better explanation

irwaters:
yeah, I sadly explored the timezone rathole last night.  would sure be useful as part of a request object...

irwaters:
bump

Gemfile
```diff
@@ -31,6 +31,9 @@ gem 'active_model_serializers', '~> 0.8.0'
 gem 'paper_trail'
 gem 'sawyer', '~> 0.5'
 gem 'sse-rails-engine', '~> 1.4'
+gem 'rails-assets-jstimezonedetect', source: 'https://rails-assets.org'
+gem 'rails-assets-jquery-cookie', source: 'https://rails-assets.org'
```
grosser: rails-assets has a section below

app/helpers/application_helper.rb
```diff
@@ -54,10 +54,14 @@ def relative_time(time)
   end
 
   def render_time(time, format)
+    # grab the time format that the user has in their profile
+    format ||= current_user.time_format || 'utc'
```
grosser: never called with nil, so this line can go away ?

grosser:
checked with @pswadi-zendesk and the `Deploys` can stay ;)
maybe turn current deploys into `Deploying` ... then we save even more space :D

irwaters:
yay, on the nav change :)

app/helpers/application_helper.rb
```diff
@@ -54,10 +54,13 @@ def relative_time(time)
   end
 
   def render_time(time, format)
+    # grab the time format that the user has in their profile
```
grosser: comment is no longer true ? -> remove ?

app/views/shared/_header.html.erb
```diff
@@ -41,7 +41,7 @@
           </li>
         <% end %>
         <li class="<%= 'active' if current_page?(controller: '/deploys', action: 'search') %>">
-          <%= link_to "Deploys", search_deploys_path %>
+          <%= link_to "Deploying", search_deploys_path %>
```
grosser: I was suggesting `Deploys` for the search and `Deploying` for the active deploys

irwaters: ahh, makes sense.

grosser:
👍

grosser:
🎉 👍

PR #899, opened 2016-04-27
Fix pulling from private registries

rata:
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

The "serveraddress" field is needed for docker-api to use the
authentication when connecting to that server.


grosser:
👍

jonmoter:
👍

PR #902, opened 2016-04-28
fail stage validation when selecting kubernetes but not having the deploy groups configured

grosser:
@zendesk/paas 

only check out last deploy, first is from another PR

### Risks
- Low: stages failing to save


jonmoter:
👍

PR #903, opened 2016-04-29
create build when none is found

grosser:
@zendesk/paas 

only look at second commit, first is from other PR

### Risks
- Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -18,17 +18,15 @@ module Kubernetes
     end
 
     def execute!(*_commands)
-      build = find_build
+      build = find_or_create_build
+      return false if stopped?
       release = create_release(build)
       ensure_service(release)
       create_deploys(release)
 
       # Wait until deploys are done and show progress
       loop do
-        if @stopped
-          @output.puts "STOPPED"
-          return false
-        end
+        return false if stopped?
 
         status = release.release_docs.map { |release_doc| pod_is_live?(release_doc) }
         if @testing_for_stability
@@ -56,6 +54,13 @@ module Kubernetes
 
     private
 
+    def stopped?
+      if @stopped
+        @output.puts "STOPPED"
+        true
+      end
+    end
+
     # TODO only call once per cluster and filter the output
     def pod_is_live?(release_doc)
       group = release_doc.deploy_group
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -79,13 +79,43 @@ def pod_is_live?(release_doc)
...
+        creator: @job.user,
+        project: @job.project,
+        label: "Automated build triggered via Deploy ##{@job.deploy.id}"
+      )
+      DockerBuilderService.new(build).run!(push: true)
+
+      loop do
+        @output.puts("Waiting for Build ##{build.id} to finish.")
+        sleep 2
+        break if build.docker_build_job.finished?
```
jonmoter: Have a way to break out of this if the user hits the "Cancel" button.

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -79,13 +79,43 @@ def pod_is_live?(release_doc)
...
+      build = Build.create!(
+        git_ref: @job.commit,
+        creator: @job.user,
+        project: @job.project,
+        label: "Automated build triggered via Deploy ##{@job.deploy.id}"
+      )
+      DockerBuilderService.new(build).run!(push: true)
+
+      loop do
+        @output.puts("Waiting for Build ##{build.id} to finish.")
```
jonmoter: Show URL so end user can go to that page.

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -79,13 +79,43 @@ def pod_is_live?(release_doc)
     # TODO: create it and wait ... also show creation logs etc
     def find_build
       if build = Build.find_by_git_sha(@job.commit)
-        @output.puts("Found build #{build.id} for sha #{@job.commit}")
+        ensure_build_is_successful(build)
       else
-        raise Samson::Hooks::UserError, "Build for sha #{@job.commit} does not exist, create it before deploying"
+        build = create_build
+        ensure_build_is_successful(build)
```
jonmoter: DRY up by putting outside the "else"

jonmoter:
👍  other than small comments above

grosser:
updated to also wait when build is already running + stop when stopped

jonmoter:
👍

Issue #837, opened 2016-04-17
Slack webhook fields not showing in stage edit form


michael-bouvy:
Hi,

When creating or editing a stage, Slack webhook fields (URL and Channel) are only visible if at least 1 webhook exists for the stage.

Although I'm not a Ruby developper, I got the expected behavior by changing line `5` in `plugins/slack_webhooks/app/views/samson_slack_webhooks/_fields.html.erb` from:

```rb
<% stage.slack_webhooks.build unless stage.slack_webhooks.last.try(:webhook_url).blank? %>
```

to:

```rb
<% stage.slack_webhooks.build %>
```

I'm wondering if it is the correct way of fixing this. If yes, I'd be pleased submitting a PR. 👍 

Expected behavior in image:

![image](https://cloud.githubusercontent.com/assets/1674029/14588597/6ef97248-04cd-11e6-9465-6e2872143199.png)

grosser:
Yes it's a bug, but the intention was that it should not show a new row
when the last row was empty, so the correct fix would be:

stage.slack_webhooks.build if (last = stage.slack_webhooks.last) &&
last.webhook_url.blank?

please open a PR!

if you feel funky, try adding a test too `it "shows a form when no webhook
exists"`

michael-bouvy:
Thanks for your quick reply!

How could the last row be empty? I tried adding a webhook with only the channel set, but it is not saved.

Will try to write a test 😎

grosser:
maybe when trying to save and it fails the same form is re-rendered ... or
something like that ... if it does not work, then feel free to just use
.build without conditions

michael-bouvy:
Ok. Could you just advise me how to get started with the test? I looked at existing ones but could not find something equivalent.

grosser:
https://github.com/zendesk/samson/blob/master/test/controllers/admin/secrets_controller_test.rb#L34

PR #904, opened 2016-04-29
Display Slack webhook fields in stage edit form

michael-bouvy:
Related to #837 

image

I'm not quite sure that the tests location is correct (tightly coupled), but I couldn't figure where else to place it. Please tell me if it's the wrong place.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### DIFF
--- a/plugins/slack_webhooks/app/views/samson_slack_webhooks/_fields.html.erb
+++ b/plugins/slack_webhooks/app/views/samson_slack_webhooks/_fields.html.erb
@@ -2,7 +2,7 @@
   <legend>Slack</legend>
 
   <% stage = form.object %>
-  <% stage.slack_webhooks.build unless stage.slack_webhooks.last.try(:webhook_url).blank? %>
+  <% stage.slack_webhooks.build %>
 
   <div class="col-lg-offset-2">
     <p>Details of the webhook to be notified on deploys.</p>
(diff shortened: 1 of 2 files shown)


plugins/slack_webhooks/app/views/samson_slack_webhooks/_fields.html.erb
```diff
@@ -2,7 +2,7 @@
   <legend>Slack</legend>
 
   <% stage = form.object %>
-  <% stage.slack_webhooks.build unless stage.slack_webhooks.last.try(:webhook_url).blank? %>
+  <% stage.slack_webhooks.build unless stage.slack_webhooks.size > 0 && stage.slack_webhooks.last.webhook_url.blank? %>
```
grosser: what does this change do ? ... now it does 2 queries instead of 1 :(

grosser: ahhhh it handles missing hook and empty hook differently ... how about:

`unless (last = stage.slack_webhooks.last) && last.webhook_url.blank?`

grosser:
great to see a test :D

plugins/slack_webhooks/app/views/samson_slack_webhooks/_fields.html.erb
michael-bouvy: Works, thanks for the tip! 😉 

How about removing the condition anyways? As previously said in the issue, I couldn't reproduce missing hook URL.

grosser: this is meant to not add a blank hook when a update/create failed, so not super important ...

PR #905, opened 2016-04-29
improve search UI and streamline timeline css

grosser:
before:
screen shot 2016-04-29 at 10 57 31 am

after:
screen shot 2016-04-29 at 10 57 01 am

@zendesk/paas @irwaters 

### Risks
- None


irwaters:
👍  on green

PR #906, opened 2016-04-29
unify access methods into current_user

grosser:
@zendesk/samson 

looks a bit silly to have 1 module doing 2 methods and the rest living in current_user ...

### Risks
- Low auth errors


jonmoter:
Simpler.  Nice.  👍

PR #908, opened 2016-04-29
PAAS-132 Make cluster lookup more efficient

grosser:
instead of doing 1 api call per role, just do 1 and store use the results to find the correct pod

@zendesk/paas 

# Risks
 - Low: breaking kube deploy

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -28,7 +28,9 @@ module Kubernetes
       loop do
         return false if stopped?
 
-        status = release.release_docs.map { |release_doc| pod_is_live?(release_doc) }
+        pods = release.fetch_pods
+        status = release.release_docs.map { |release_doc| pod_is_live?(pods, release_doc) }
+
         if @testing_for_stability
           if status.all?
             @testing_for_stability += 1
@@ -61,22 +63,14 @@ module Kubernetes
       end
     end
 
-    # TODO only call once per cluster and filter the output
-    def pod_is_live?(release_doc)
+    def pod_is_live?(pods, release_doc)
       group = release_doc.deploy_group
-      role = release_doc.kubernetes_role.name
-      pod = group.kubernetes_cluster.client.get_pods(
-        namespace: group.kubernetes_namespace,
-        label_selector: {
-          deploy_group_id: group.id,
-          project_id: @job.project_id,
-          release_id: release_doc.kubernetes_release_id,
-          role: role
-        }.map { |k, v| "#{k}=#{v}" }.join(',')
-      ).first
+      role = release_doc.kubernetes_role
+
+      pod = pods.detect { |pod| pod.role_id == role.id && pod.deploy_group_id == group.id }
 
       live, description = analyze_pod_status(pod)
-      @output.puts "#{group.name} #{role}: #{description}"
+      @output.puts "#{group.name} #{role.name}: #{description}"
       live
     end
 
@@ -142,7 +136,6 @@ module Kubernetes
 
     def analyze_pod_status(pod)
       if pod
-        pod = Kubernetes::Api::Pod.new(pod)
         if pod.live?
           if pod.restarted?
             [false, "Restarted"]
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -56,19 +58,25 @@ def execute!(*_commands)
...
+      release.release_docs.map(&:deploy_group).flat_map do |deploy_group|
+        query = {
+          namespace: deploy_group.kubernetes_namespace,
+          label_selector: {
+            deploy_group_id: deploy_group.id,
+            project_id: @job.project_id,
+            release_id: release.id
+          }.map { |k, v| "#{k}=#{v}" }.join(',')
+        }
+        deploy_group.kubernetes_cluster.client.get_pods(query)
```
jonmoter: Have this return an array of `Kubernetes::Api::Pod` objects, so you can use their helper methods

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -56,19 +58,25 @@ def execute!(*_commands)
...
-    def pod_is_live?(release_doc)
+    def deployed_pods(release)
+      release.release_docs.map(&:deploy_group).flat_map do |deploy_group|
+        query = {
+          namespace: deploy_group.kubernetes_namespace,
+          label_selector: {
+            deploy_group_id: deploy_group.id,
+            project_id: @job.project_id,
+            release_id: release.id
+          }.map { |k, v| "#{k}=#{v}" }.join(',')
```
jonmoter: use `#to_kuber_selector` method

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -56,19 +58,25 @@ def execute!(*_commands)
...
-        namespace: group.kubernetes_namespace,
-        label_selector: {
-          deploy_group_id: group.id,
-          project_id: @job.project_id,
-          release_id: release_doc.kubernetes_release_id,
-          role: role
-        }.map { |k, v| "#{k}=#{v}" }.join(',')
-      ).first
+
+      pod = pods.detect { |pod| pod.metadata.labels.role == role && pod.metadata.labels.deploy_group_id.to_i == group.id }
```
jonmoter: Maybe add a helper method to the `Pod` class to determine whether it's part of a ReleaseDoc?

jonmoter:
Looks good in general.

jonmoter:
👍

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -87,6 +87,20 @@ def update_status(release_doc)
       save!
     end
 
+    def pods
```
jonmoter: Small detail: I'd suggest calling this `#fetch_pods`.

If you call it `#pods`, I'd assume that it's like an ActiveRecord association, and the results are cached.  So I could call `release.pods` several times without doing a round-trip to Kubernetes each time.  If you call it `fetch_pods`, that hints to me that calling it does a network roundtrip.

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -56,22 +57,14 @@ def execute!(*_commands)
...
-        label_selector: {
-          deploy_group_id: group.id,
-          project_id: @job.project_id,
-          release_id: release_doc.kubernetes_release_id,
-          role: role
-        }.map { |k, v| "#{k}=#{v}" }.join(',')
-      ).first
+      role = release_doc.kubernetes_role
+
+      pod = pods.detect { |pod| pod.role_id == role.id && pod.deploy_group_id == group.id }
```
jonmoter: Can avoid doing fetches of `group` and `role`.

```
pod = pods.detect { |pod| pod.role_id == release_doc.kubernetes_role_id && pod.deploy_group_id == release_doc.deploy_group_id }
```

jonmoter: Ah, nevermind. I see that you're outputting their names below.

grosser:
all updated

PR #909, opened 2016-04-29
reduce redundant messages and group roles by deploy group

grosser:
@zendesk/paas 

based on previous PR, only look at the second commit.

before:
screen shot 2016-04-29 at 1 23 34 pm

after:
screen shot 2016-04-29 at 1 23 23 pm


jonmoter:
👍

PR #911, opened 2016-04-29
Irwaters/cachefix changing the cache key for deploys to prevent it from using the cache when the time format changes

irwaters:
changing the cache key for deploys to prevent it from using the cache when the time format changes.

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High
Low


grosser:
not a big fan of caching on all of params ... maybe `params[:timeformat]` ?

irwaters:
that works for me, change made

grosser:
👍

PR #912, opened 2016-04-29
store project_id directly on release for easier querying

grosser:
we are doing unnecessary queries in a few places and adding indirection ... also means we need joins ... just normalize it like most other things already do ...

@zendesk/paas

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -6,14 +6,14 @@ module Kubernetes
 
     belongs_to :user
     belongs_to :build
+    belongs_to :project
     has_many :release_docs, class_name: 'Kubernetes::ReleaseDoc', foreign_key: 'kubernetes_release_id'
     has_many :deploy_groups, through: :release_docs
 
-    delegate :project, to: :build
-
-    validates :build, presence: true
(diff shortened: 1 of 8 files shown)


plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -6,12 +6,11 @@ class Release < ActiveRecord::Base
 
     belongs_to :user
     belongs_to :build
+    belongs_to :project
```
jonmoter: Can you add a validation that `build.project_id == product_id`?

jonmoter: and maybe a `before_validation` filter that `project_id ||= build.try(:project_id)`

grosser: it blows up when it's not set, so makes sure it is always passed in -> not
necessary to prefill it ? / no other model does that
no other model has project integrity validation -> not necessary here ?

grosser:
added project integrity validation

jonmoter:
Did you mean to make all those changes to schema.rb?

grosser:
updated!

jonmoter:
👍 on resolving merge conflicts

PR #913, opened 2016-04-30
remove broken deploy.url

grosser:
@zendesk/samson 

make .url return an actual url ...


jonmoter:
some Travis tests broke on this change

grosser:
should be all green now ...

jonmoter:
👍

PR #914, opened 2016-04-30
get rid of dead angular code

grosser:
disposing of the dead code @irwaters left behind when turning deploys into a regular search endpoint

/cc @zendesk/samson 

### Risks
- Low: active deploys misbehaving


irwaters:
👍

jonmoter:
👍

PR #915, opened 2016-04-30
rewrite project migration since postgres uses different joins and sql…

grosser:
…ite does not support them at all

@jonmoter 
/cc @zendesk/samson


jonmoter:
👍

PR #918, opened 2016-05-02
fetch files directly from git

grosser:
remove github dependency and make file fetching faster

```
 Benchmark.realtime { p.repository.send :file_content, "59aa8ab37f013f4b8fe7d87e30edbc808175b51b", 'Gemfile' }
=> 0.03225039999233559
```

@zendesk/samson 

### Risks
- Low: slow operations when repo is not cached .. which should not happen since we resolve the sha before hitting this codepath anyway

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -50,7 +50,7 @@ class GitRepository
   add_method_tracer :update!
 
   def commit_from_ref(git_reference, length: 7)
-    setup_local_cache! unless locally_cached?
+    ensure_local_cache!
 
     Dir.chdir(repo_cache_dir) do
       # brakeman thinks this is unsafe ... https://github.com/presidentbeef/brakeman/issues/851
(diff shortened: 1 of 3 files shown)


irwaters:
👍

app/models/git_repository.rb
```diff
@@ -114,8 +107,22 @@ def file_changed?(sha1, sha2, file)
     executor.execute!("cd #{pwd}", "git diff --quiet --name-only #{sha1}..#{sha2} #{file}")
   end
 
+  def file_content(sha, file)
+    raise ArgumentError, "Need a sha, but #{sha} (#{sha.size}) given" unless sha.size == 40
```
jonmoter: FYI, the `Build` model has a validation that the `git_sha` attribute is a valid SHA1.  Might want to extract out the logic for checking that something is a valid SHA and use that, rather than checking length here, and using a regex there.

grosser: nice, taken care of in next PR

PR #919, opened 2016-05-02
allow creating a Kubernetes config without an ENV var

jonmoter:
The ENV var KUBE_CONFIG_FILE is not strictly necessary for creating a new cluster, so don't blow up if it's not set.

If another cluster exists, default to the config file that cluster uses.

/cc @zendesk/paas, @grosser 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


jonmoter:
If we merge this, we can set up clusters on samsontest.zende.sk without needing to add a new ENV var.

grosser:
👍

PR #920, opened 2016-05-02
use execute_command more

grosser:
@zendesk/samson 

also unrelated cleanup ... hiding `update!` method and making everyone use update_local_cache which will clone or update as necessary

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -19,13 +19,13 @@ class GitRepository
     raise ArgumentError.new("git_reference is required") if git_reference.blank?
 
     executor.output.write("# Beginning git repo setup\n")
-    return false unless setup_local_cache!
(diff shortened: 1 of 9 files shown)


app/models/git_repository.rb
```diff
@@ -131,26 +122,9 @@ def locally_cached?
...
-  # - array support for input safety
-  # - no forced unique + sort
-  # - simpler
-  def capture_stdout(command)
-    Dir.chdir(repo_cache_dir) do
-      out = IO.popen(command, err: [:child, :out]) do |io|
-        io.read.strip
-      end
-
+  def capture_stdout(*command, dir: repo_cache_dir)
```
jonmoter: Add a comment that it returns `nil` if the command errors.

jonmoter:
👍

PR #923, opened 2016-05-02
don't crash if Kubeconfig file doesn't have any users

jonmoter:
Fix crash I saw in staging when kubeconfig file has no users

/cc @zendesk/pass, @grosser 

### Tasks
 - [ ] :+1: from team

### References
 - Airbrake: https://zendesk.airbrake.io/projects/95346/groups/1677031245491938996

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/client_config_file.rb
+++ b/plugins/kubernetes/app/models/kubernetes/client_config_file.rb
@@ -39,7 +39,7 @@ module Kubernetes
 
     def parse_clusters
       @clusters = {}
-      @config_file.fetch(:clusters).each do |cluster_hash|
+      (@config_file[:clusters] || []).each do |cluster_hash|
         cluster = Cluster.new
         cluster.name = cluster_hash[:name]
         cluster.server = cluster_hash[:cluster][:server]
@@ -56,7 +56,7 @@ module Kubernetes
 
     def parse_users
       @users = {}
-      @config_file.fetch(:users).each do |user_hash|
+      (@config_file[:users] || []).each do |user_hash|
         user = User.new
         user.name = user_hash[:name]
         user.username = user_hash[:user][:username]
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -39,7 +39,7 @@ def parse_file
 
     def parse_clusters
       @clusters = {}
-      @config_file.fetch(:clusters).each do |cluster_hash|
+      (@config_file[:clusters] || []).each do |cluster_hash|
```
grosser: but a file without clusters is useless ?

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -75,7 +75,7 @@ def parse_users
 
     def parse_contexts
       @contexts = {}
-      @config_file.fetch(:contexts).each do |context_hash|
+      (@config_file[:contexts] || []).each do |context_hash|
```
grosser: file without context is also useless ?

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
jonmoter: Possibly, but it's not an invalid config file, and shouldn't cause a crash.

grosser: but then it will crash later when we try to do anything with it ...

grosser:
maybe don't crash but show validation error ?

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
jonmoter: It will fail validation when trying to add a cluster with that config file, because `Kubernetes::Cluster#test_client_connection` will fail.  But it will show up as a validation error, rather than a 500

grosser: then that's fine 👍

PR #926, opened 2016-05-03
docs

grosser:
@zendesk/paas 

documenting a few things I tried to delete and then figured out what they do ...


grosser:
added a rewrite of filter-by in pure jquery ... feels simpler to me ... thoughts ? (second commit)

jonmoter:
👍

PR #928, opened 2016-05-03
move kubernetes angular code into kubernetes folder

grosser:
@zendesk/paas this seems to all be unused / builds tab works fine without ...


grosser:
@sbrnunes

grosser:
this is all only used in kubernetes part ... so moving it there ... how about deleting all that kubernetes js now that we have a proper deploy flow ? :trollface:

jonmoter:
👍

sbrnunes:
Sorry @grosser, I was on PTO last week, I'm just getting through my email. 

Yep, this code was only used by the Kubernetes AngularJS screens. However, because they were not specific for the Kubernetes plugin, i.e., this code could be reused anywhere, I placed it in the core.

PR #930, opened 2016-05-03
take into account the time_format from the users's profile for caching purposes

irwaters:
First time you hit the page, params[:time_format] is always nil.  this leads to strange behavior when the user changes their default time_formatting.  

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
https://zendesk.atlassian.net/browse/SAMSON-235
### Risks
- Level: Low/Med/High
Low


app/views/projects/_deploy.html.erb
```diff
@@ -1,4 +1,4 @@
-<% cache [@project, deploy, params[:time_format]] do %>
+<% cache [@project, deploy, params[:time_format] ? params[:time_format] : current_user.time_format] do %>
```
grosser: `params[:time_format].presence || current_user.time_format`

grosser:
👍

PR #931, opened 2016-05-03
reuse local fetching code for role file lookups

grosser:
@zendesk/samson 

 - reuse git repository local fetching logic in project
 - make repo setup commands raise when they fail
 - use memory store so we get an actual cache
 - don't fail a bunch of commands when checkout/update fails

### Risks
- Low: slowdown due to repo downloads/updates


irwaters:
👍

PR #933, opened 2016-05-04
fix migration deleting all but one of peoples project role permissions

grosser:
@zendesk/samson

Found the reason for PermissionGate ...


jonmoter:
😭 
👍

PR #934, opened 2016-05-04
search users by role and make role display reflect internal role logic

grosser:
@zendesk/samson 

 - search users by role to find all admins/deployers for a project /fyi @joshdellinger @Rchirackal 
 - can no longer select None when user is a >= Deployer
 - can no longer select Deployer when user is a >= Admin

screen shot 2016-05-04 at 9 28 29 am

### Risks
- Low: unable to assign project level roles correctly

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -5,6 +5,9 @@ class UsersController < ApplicationController
 
   def index
     @users = User.search_by_criteria(params)
+    if role_id = params[:role_id]
+      @users = @users.with_role(role_id, current_project.id)
+    end
 
     respond_to do |format|
       format.html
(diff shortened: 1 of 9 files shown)


jonmoter:
In list UI, have some way to distinguish whether someone has a global or project-specific role.

app/controllers/users_controller.rb
```diff
@@ -5,6 +5,14 @@ class UsersController < ApplicationController
 
   def index
     @users = User.search_by_criteria(params)
+    if role_id = params[:role_id]
+      @users = @users.
+        joins('LEFT OUTER JOIN user_project_roles on users.id = user_project_roles.user_id').
+        where(
+          '(user_project_roles.project_id = ? AND user_project_roles.role_id >= ?) OR users.role_id >= ?',
+          current_project.id, role_id, role_id
+        )
```
jonmoter: How about putting this on the model instead of the controller?

grosser:
all shiny and even tested now :)

screen shot 2016-05-05 at 7 42 34 pm

irwaters:
:+1 lgtm

PR #936, opened 2016-05-04
fail when pods take too long to get ready

grosser:
@zendesk/paas


jonmoter:
Probably fine, but we'll need to tweak the watcher logic to handle rolling updates.

👍

PR #937, opened 2016-05-05
make h1 use margins consistent with the rest of the UI

grosser:
@zendesk/samson 

before: inconsistent spacing with rest of the site
screen shot 2016-05-05 at 8 53 30 am

after:
screen shot 2016-05-05 at 8 51 57 am
screen shot 2016-05-05 at 8 54 47 am


jonmoter:
👍

PR #939, opened 2016-05-05
fixing cache key 

irwaters:
last compound cache key used params[..].presence which would be nil in some occastions.  leads to a messed up cache

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High
Low


grosser:
this is exactly the same logic as before ... nil || current_user is the same as testing for nil and then using current_user

irwaters:
you are totally correct.  must just be a messed up cache somehow

PR #940, opened 2016-05-05
since this partial is used in a couple of different places, we need to be more careful about caching

irwaters:
this partial is used by 2 different pages, with 2 differnt column counts.  need to make sure that our caching is selective about those pages so that we don't end up wiht funky results

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High
Low


grosser:
👍

PR #941, opened 2016-05-05
do not let secrets into the lean command executor, it fails when used…

grosser:
… with hub gem on private repos

needs HOME to find ssh keys ... and PATH to find `git`

@pschambacher 
/cc @zendesk/samson


pschambacher:
👍  Looks safe

PR #942, opened 2016-05-06
turn DeployYaml into a object

grosser:
@jonmoter 

looks good ?

... I'll add some basic tests too just for sanity


jonmoter:
Looks good.  Will probably create a `DaemonsetYaml` class that will look similar, and do some refactoring to DRY up code there too.

👍

Issue #943, opened 2016-05-06
Use git worktree instead of cloning cached repos


mariovisic:
Hi team

I'd like to propose using `git worktree` in place of the current system where each time a job is created the cached repo is cloned to a temporary directory. There would be a number of benefits of swapping this over.

Git worktree works by creating a copy of the project folder but symlinking a lot of the git internals, this means that it can see the exact same tree as the cached repo, if you fetch on the cached repo for example the worktree will also see it but they can point to different HEADs. It's as simple as `git worktree add /tmp/path <ref> --force` Then once  the folder is removed simply `git worktree prune`.

Git worktrees are much quicker than cloning, here is a benchmark on a fairly large repo:

```
$ time git clone project project-temp
Cloning into 'project-temp'...
done.
Checking out files: 100% (12544/12544), done.

real	0m9.838s
user	0m0.856s
sys	0m3.351s

$ time git worktree add ../project-temp master --force
Preparing ../project-temp (identifier project-temp)
HEAD is now at b2c6367 Merge pull request #16000 from envato/analytics_constant_resolution

real	0m3.147s
user	0m0.734s
sys	0m1.513s
```

Ir also takes much less space as all of the git objects are symlinked.

```
# with clone
$ du -sh project-temp/
1.7G	project-temp/

# with worktree
$ du -sh project-temp/
239M	project-temp/
```

The code changes should be fairly minimal, only a change to `GitRepository#setup!` would be required and a small amount of connecting parts.

Thoughts?

grosser:
yes please!

mariovisic:
See #944

PR #944, opened 2016-05-06
[WIP] Use git worktree instead of cloning cached repos

mariovisic:
This PR uses `git worktree` to create a copy of cached repos for a job instead of cloning. This is a work in progress. I still need to update tests but wanted to get some feedback early on. In particular i've removed the mirror option from clone, now all repos will be cloned as a mirror, I couldn't see any instances of needing a non mirror clone apart from the case that was replaced with the worktree.

/cc @zendesk/samson and @grosser 
### TODO
- [ ] Update tests
- [ ] Prune worktrees from the cached repo after a job has finished executing (inside the ensure)
### Tasks
- [ ] :+1: from team
### Risks
- Level: Med

Changing the way that job directories are setup so there's some risk involved here.

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -20,8 +20,7 @@ class GitRepository
 
     executor.output.write("# Beginning git repo setup\n")
     return false unless update_local_cache!
-    return false unless clone!(from: repo_cache_dir, to: temp_dir)
-    return false unless checkout!(git_reference, pwd: temp_dir)
+    return false unless setup_worktree!(temp_dir, git_reference)
     true
   end
 
@@ -29,20 +28,21 @@ class GitRepository
     if locally_cached?
       update!
     else
-      clone!(from: repository_url, to: repo_cache_dir, mirror: true)
+      clone!(from: repository_url, to: repo_cache_dir)
     end
   end
 
-  def clone!(from: repository_url, to: repo_cache_dir, mirror: false)
+  def clone!(from: repository_url, to: repo_cache_dir)
     @last_pulled = Time.now if from == repository_url
-    command = if mirror
-      "git -c core.askpass=true clone --mirror #{from} #{to}"
-    else
-      "git clone #{from} #{to}"
-    end
+    command = "git -c core.askpass=true clone --mirror #{from} #{to}"
     executor.execute! command
   end
   add_method_tracer :clone!
+
+  def setup_worktree!(temp_dir, git_reference)
+    command = "git worktree add "
+    executor.execute!("cd #{repo_cache_dir}", "git worktree add #{temp_dir} #{git_reference} --force")
+  end
 
   def commit_from_ref(git_reference, length: 7)
     return unless ensure_local_cache!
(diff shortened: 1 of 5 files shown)


mariovisic:
The `worktree` subcommand was released for git 2.5 from what I can see, which was August 2015, the version of git running on CI doesn't seem to have that command and so that's causing the failures present.

app/models/git_repository.rb
```diff
@@ -20,30 +20,30 @@ def setup!(temp_dir, git_reference)
...
-    else
-      "git clone #{from} #{to}"
-    end
+    command = "git -c core.askpass=true clone --mirror #{from} #{to}"
     executor.execute! command
   end
   add_method_tracer :clone!
 
+  def setup_worktree!(temp_dir, git_reference)
+    command = "git worktree add "
```
grosser: not used ?

test/models/git_repository_test.rb
```diff
@@ -84,17 +84,6 @@
     end
   end
 
-  describe "#cheakout!" do
```
grosser: cheakout ;)

jonmoter:
I wouldn't want to require a super-recent version of git in order to run Samson.

Can you add the logic to be conditional, so that it uses worktree when it detects git version >= 2.5, or if an env var like `GIT_WORKTREE_ENABLED` is set?

grosser:
can we have it blow up if the installed git version is too low so nobody has to hunt down why it does not work ?

grosser:
2.5 is ~10 months old now ... what would be not super-recent ?

maybe make it print deprecation warnings when an old git version is used but still support it ... 

maybe this installs a 2.5+ git version on travis:

```
addons:
  apt:
    packages:
    - git-core

```

jonmoter:
I want to avoid adding a new global dependency to Samson without allowing backward compatibility.

For example, we're currently running git 1.7.x on our servers. So we would break Samson if we merged this changed and deployed it.

rata:
Just curious, what is the benefit of using git worktree in these cases?

grosser:
Faster afaik

rata:
On Sat, May 14, 2016 at 12:29:17PM -0700, Michael Grosser wrote:
> Faster afaik

I can't find any reference for that. Btw, a local clone uses hardlinks.

What is the motivation for this change?

Also, debian stable (and probably other distros too), have an older git version.
Debian stable is used in Samson's Dockerfile IIRC, so it won't work there.

mariovisic:
@rata the numbers from my testing are here: https://github.com/zendesk/samson/issues/943

grosser:
Do you want to do the refactor so it's an optional feature when git is new
enough ? (maybe triggered by an ENV flag for simplicity ...)

mariovisic:
@grosser yes, I think that's a great idea. I'll need to ensure the testing is solid though, that's my only concern at the moment. I wouldn't mind having it as configurable with an ENV variable, but defaulting to enabled if git is 2.5 or newer, thoughts?

grosser:
Yeah should be opt-out when you have a newer git version ...

Detection is fast ... so could do on startup:
Benchmark.realtime { `git --version` } => 0.004835939995246008
... but ideally on call of the clone method ...

rata:
On Sat, May 14, 2016 at 09:41:09PM -0700, Mario Visic wrote:
> @rata the numbers from my testing are here: https://github.com/zendesk/samson/issues/943

Oh, wow, those are impressive numbers! Great :-)

jonmoter:
PR is a year and a half old. Should we consider this abandoned, or is there still appetite to finish this?

grosser:
I'd like to have this ... seems silly that we use something else when there is something that is clearly better ...

PR #945, opened 2016-05-06
test deploy yaml

grosser:
@jonmoter 

 - validate containers
 - validate matchLabels
 - make cpu readable -> use floats


jonmoter:
👍

PR #946, opened 2016-05-06
support DaemonSet when deploying kubernetes

grosser:
@zendesk/paas 

reusing the deploy parsing logic and sending to a different endpoint
only second commit, first is part of another PR

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -71,14 +71,14 @@ module Kubernetes
     end
 
     def deploy_to_kubernetes
-      deployment = Kubeclient::Deployment.new(deployment_hash)
-      # Create new client as 'Deployment' API is on different path then 'v1'
-      extension_client = deploy_group.kubernetes_cluster.extension_client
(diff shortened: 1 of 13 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -71,14 +71,10 @@ def client
...
-      deployment = Kubeclient::Deployment.new(deployment_hash)
-      # Create new client as 'Deployment' API is on different path then 'v1'
-      extension_client = deploy_group.kubernetes_cluster.extension_client
-      if previous_deploy?(extension_client, deployment)
-        extension_client.update_deployment(deployment)
-      else
-        extension_client.create_deployment(deployment)
-      end
+      deployment = Kubeclient.const_get(deploy_yaml.resource.camelcase)
+      deployment = deployment.new(deploy_yaml.deployment_hash)
```
jonmoter: Call something like `resource_klass` or something like that.

jonmoter: Clearer if just something like `if dameonset?`...

jonmoter:
A little unclear code, but close enough to right to move forward.
👍

PR #947, opened 2016-05-07
SAMSON-236 clean out all unnecessary gemfile requires

grosser:
@zendesk/samson 

enabling us to not install asset gems for production deploys, making nodejs no longer a dependency + the deployment lighter + boot times faster


jonmoter:
👍 on 💚

PR #948, opened 2016-05-07
make staging behave like production

grosser:
@zendesk/samson 

(only diff was staging having FORCE_SSL setting which is redundant since we do that in application_controller -> local /ping did not work in staging when FORCE_SSL was enabled)


sandlerr:
:+1: (or symlink it)

grosser:
leaving it like this so it's easy to add additions / changes

PR #951, opened 2016-05-09
kubernetes role CRUD

grosser:
@zendesk/paas 

 - can only `seed` once to avoid soft-deleting / destroying things
 - can create/update/edit any role as admin and view as deployer
 - turning DOCKER_FEATURE check into a flash error so it's still useable (... don't have to spin up docker to play with the UI)
 - all tested
 - at deploy time fail when role in the repo is not configured (previously we created them automatically)
 - better feedback when multiple deploy group roles are missing


grosser:
No roles: seed
screen shot 2016-05-09 at 1 36 08 pm

index:
screen shot 2016-05-09 at 1 35 45 pm

edit:
screen shot 2016-05-09 at 1 35 09 pm

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -19,6 +19,37 @@ class Role < ActiveRecord::Base
...
+          service_name: config_file.service && config_file.service.metadata.name,
+          ram: config_file.deployment.ram_mi, # TODO: remove this column
+          cpu: config_file.deployment.cpu_m, # TODO: remove this column
+          replicas: config_file.deployment.spec.replicas || 1, # TODO: remove this column
+          deploy_strategy: config_file.deployment.strategy_type
+        )
+      end
+    end
+
+    def self.for_deploy(project, git_sha)
```
jonmoter: Should raise an error if `git_sha` is not a valid ref for that project.

jonmoter: Naming is odd.  Think of something better!

jonmoter: Maybe `#configured_for_project`?

jonmoter:
👍

PR #952, opened 2016-05-09
we must enforce unique service names or different projects will overr…

grosser:
…ide each other

@zendesk/paas


jonmoter:
Add handling for trying to seed when there's a name collision.

grosser:
added seeding support for duplicate service names

irwaters:
👍  lgtm

PR #954, opened 2016-05-10
cleanup gemfile

grosser:
@zendesk/paas 

 - freeing versions for future update (no gem can be downgraded ... so no point in locking anything unless there is a known upgrade restriction)
 - removing unnecessary require false ... nothing is required by default anymore
 - remove duplicate source listing


irwaters:
👍  on green

PR #955, opened 2016-05-10
make deploy group role a real association

grosser:
@zendesk/paas 

now that we no longer delete/create them all the time we can use ids to reference them

before: free form names
screen shot 2016-05-10 at 8 56 24 am

after: associated with role or deleted
screen shot 2016-05-10 at 9 21 31 am

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160508055458) do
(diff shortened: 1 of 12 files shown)


grosser:
making the edit fill the roles according to the currently selected project:

screen shot 2016-05-10 at 10 03 57 am
screen shot 2016-05-10 at 10 04 03 am

role cannot be changed once saved:
screen shot 2016-05-10 at 10 02 48 am

jonmoter:
👍 on 💚

PR #957, opened 2016-05-10
indexes stay on mysql but get auto dropped on postgres

grosser:
@jonmoter 

heroku deploy failed with 

```
Index name 'index_kubernetes_deploy_group_roles_on_project_id' on table 'kubernetes_deploy_group_roles' does not exist/tmp/build_32431f4dab5ea561cd29f831df75536f/zendesk-samson-bcbfdda/vendor/bundle/ruby/2.2.0/gems/activerecord-4.2.5.2/lib/active_record/connection_adapters/abstract/schema_statements.rb:988:in `index_name_for_remove'

```


jonmoter:
👍

PR #959, opened 2016-05-10
bump a few gems to make rails 5 easier

grosser:
@zendesk/samson 

 - webmock has a known issue with basic auth now needing to be in the headers -> convert to explicit headers
 - serializers implements some inheritance tracking that does not call super -> we use prepend

### Risks
- Medium: everything


irwaters:
👍

PR #961, opened 2016-05-11
manually do what 12-factor gem did to avoid confusion

grosser:
@zendesk/samson @bhouse 

we did serve static assets because 12 factor set it up in the background,
which led to assets not being served once the auto-require was removed in #947 

it also sets
 -    `action_dispatch.x_sendfile_header` ... but we already have it as nil
 - logger to stdout ... but we overwrote that with our syslog anyway


sandlerr:
does the docker container still log to stdout?
:+1:

grosser:
seems to still log to stdout ... fixed a few other errors to get it running ...

```
00
D, [2016-05-11T04:36:20.859468 #1] DEBUG -- :
D, [2016-05-11T04:36:20.859595 #1] DEBUG -- :
I, [2016-05-11T04:36:20.860645 #1]  INFO -- : Started GET "/" for 127.0.0.1 at 2016-05-11 04:36:20 +0000
D, [2016-05-11T04:36:21.681433 #1] DEBUG -- :   ActiveRecord::SchemaMigration Load (0.7ms)  SELECT `schema_migrations`.* FROM `schema_migrations`
I, [2016-05-11T04:36:22.369752 #1]  INFO -- : Processing by ProjectsController#index as */*
W, [2016-05-11T04:36:22.379216 #1]  WARN -- : Halted as unauthorized! threw :warden
I, [2016-05-11T04:36:22.380108 #1]  INFO -- : Completed   in 10ms (ActiveRecord: 0.0ms)
```

grosser:
staging/prod don't log in docker/heroku though ... next PR will take care of that ... but this will make staging serve assets which is more important :D

sandlerr:
LGTM

grosser:
test-deploy fixes assets 💃

PR #962, opened 2016-05-11
bump mysql down since it fails to deploy

grosser:
fallout from https://github.com/zendesk/samson/pull/959

No such file or directory @ rb_sysopen - conftest.c


grosser:
also bumping down airbrake since it's config format has changed and blows up on boot in staging/production

grosser:
@zendesk/samson ... will fix currently failing staging deploys

sandlerr:
:+1:

PR #963, opened 2016-05-11
support stdout logging in dev via RAILS_LOG_TO_STDOUT

grosser:
new rails 5 shinyness ...
I was hoping production would work, but somehow syslogger does not go into stdout ... so making it use the same stdout logger as development with docker ... 

Heroku log lines now look like this:
```
2016-05-11T05:23:46.806416+00:00 app[web.1]: method=GET path=/ping format=html controller=ping action=show status=200 duration=0.50 view=0.00 db=0.00 params={}
```

@sandlerr 
/cc @zendesk/samson

### DIFF
--- a/app.json
+++ b/app.json
@@ -43,6 +43,9 @@
     },
     "DEPLOY_TASKS": {
       "value": "db:migrate"
+    },
+    "RAILS_LOG_TO_STDOUT": {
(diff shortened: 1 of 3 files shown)


sandlerr:
:+1:

PR #964, opened 2016-05-11
allow reading kubeconfig TLS data from files on disk [PAAS-125]

jonmoter:
* Adds functionality to read cert data from files, as well as base64-encoded values.
* Logic mostly stolen from https://github.com/abonas/kubeclient/pull/127
* Once that PR gets merged, we can remove this file entirely and use the logic in the gem.

/cc @zendesk/paas

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-125

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/client_config_file.rb
+++ b/plugins/kubernetes/app/models/kubernetes/client_config_file.rb
@@ -1,6 +1,8 @@
 require 'celluloid/io'
 
 module Kubernetes
+  # Parses the kubeconfig file that contains cluster information
+  # http://kubernetes.io/docs/user-guide/kubeconfig-file/
   class ClientConfigFile
     attr_reader :filepath, :api_version, :clusters, :users, :contexts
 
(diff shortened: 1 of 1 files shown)


grosser:
how about using that branch ?
otherwise some comments on how to remove that ... and verify that it works ... since there are no tests ;(

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -43,10 +45,10 @@ def parse_clusters
         cluster = Cluster.new
         cluster.name = cluster_hash[:name]
         cluster.server = cluster_hash[:cluster][:server]
-        ca_data = cluster_hash[:cluster][:'certificate-authority-data']
+        ca_data = fetch_cluster_ca_data(cluster_hash[:cluster])
```
grosser: .fetch ?

plugins/kubernetes/app/models/kubernetes/client_config_file.rb
```diff
@@ -85,6 +90,34 @@ def parse_contexts
       end
     end
 
+    def ext_file_path(path)
+      File.join(File.dirname(@filepath), path)
+    end
+
+    def fetch_cluster_ca_data(cluster)
+      if cluster.key?('certificate-authority')
```
grosser: not a big fan of repeating keys ... how about a if assignment or assign + check ?

```Ruby
if data = cluster['certificate-authority']
  ... parse data ...
end
```

jonmoter:
Yeah, I know the lack of tests suck.  The PR I referenced hasn't been rebased recently, so I wouldn't want to use the branch directly.

We could use the https://github.com/zendesk/kubeclient fork that I made a while back, and create our own branch that is master plus his PR merged.  (And one bug fixed that I found.)  That sound like a better idea?

grosser:
yeah ... we already have deploys + daemonsets ... might be easier to have our own fork and then also keep the extra tests there ...

jonmoter:
Okay, here's a branch we can use: https://github.com/zendesk/kubeclient/tree/samson-gem-branch

PR #965, opened 2016-05-11
prevent like injection

grosser:
@zendesk/samson 

/fyi @cericksen we don't have any detection for this -> potential DOS attack

opened a brakeman issue to hopefully track it there 

https://github.com/presidentbeef/brakeman/issues/876


henders:
lgtm 👍

PR #967, opened 2016-05-11
use the Kubeclient::Config file rather than our own version [PAAS-125]

jonmoter:
* Update Kubeclient gem to use our fork which contains `Kubeclient::Config` class
* Refactor a bit of code to use new gem

/cc @zendesk/paas

### Tasks
 - [ ] :+1: from team
 - [ ] fix bug where Celluloid clients aren't working

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-125

### Risks
- Level: Low


Gemfile
```diff
@@ -43,6 +43,7 @@ gem 'net-http-persistent'
 gem 'concurrent-ruby'
 gem 'vault'
 gem 'docker-api'
+gem 'kubeclient', github: 'zendesk/kubeclient', branch: 'samson-gem-branch'
```
grosser: maybe document the PRs we are waiting for so we can easily check up on them

grosser:
👍 ❤️

PR #968, opened 2016-05-11
allow multiple deployments even though they are not properly supported

grosser:


jonmoter:
Actually, I don't think this is it.  We want to allow > 1 container in a the spec of a single Deployment.  Isn't this validating whether there's more than 1 Deployment in a config file?

jonmoter:
We want to remove the check in `DeployYaml#container`.

```
    def container
      containers = template.spec.template.try(:spec).try(:containers) || []
      if containers.size != 1
        raise Samson::Hooks::UserError, "Template #{@doc.template_name} has #{containers.size} containers, having 1 section is valid."
      end
      containers.first
    end
```

That's the problematic check.

grosser:
yes, got that mixed up :D

grosser:
updated!

jonmoter:
👍 on 💚

PR #970, opened 2016-05-11
fix migration by removing index prior to dropping columns

grosser:
@jonmoter 

resulted in the index being modified to not include the name and the index_exit failed because it did not
reference all columns ... now it should behave properly ...


jonmoter:
👍

PR #971, opened 2016-05-11
do not override latest with branch builds

grosser:
only integration pushes will update latest now ... 
/cc @zendesk/paas 

downside is that builds on master won't ... but that should be fine for our workflows ...

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -83,7 +83,7 @@ class Integrations::BaseController < ApplicationController
   end
 
   def create_docker_image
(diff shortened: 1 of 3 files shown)


irwaters:
👍

app/controllers/integrations/base_controller.rb
```diff
@@ -83,7 +83,7 @@ def create_build_record
   end
 
   def create_docker_image
-    DockerBuilderService.new(@build).run!(push: true)
+    DockerBuilderService.new(@build).run!(push: true, push_to_latest: true)
```
jonmoter: I think `tag_as_latest` is slightly more accurate

PR #972, opened 2016-05-12
Rails 5

grosser:
WIP ... just leaving this here in case someone wants to do the same or help

## TODO
 - [x] make mysql 0.4.4 compile
 - [X] make newrelic api go away https://github.com/zendesk/samson/pull/974
 - [x] mystery nested params issue
 - [x] mystery button disappearing issue
 - [X] streaming broken on rails 5 https://github.com/rails/rails/issues/25672

## TODO after merge
 - [x] use new rails defaults by stealing from new clean app
 - [ ] replace sse-rails-engine
 - [ ] covert every controller test to integration test
 - [x] fix and bump soft_deletion return false on save to remove deprecation


henders:
I created v1.5 of the SSE gem @grosser

grosser:
@sandlerr streaming seems to be the only thing not working ... but there is no clear error message / reason I can see ... can you poke it a bit ?

app/controllers/unauthorized_controller.rb
```diff
@@ -32,8 +32,6 @@ def respond
   private
 
   def redirect_back_or(path)
-    redirect_to :back
-  rescue ActionController::RedirectBackError
-    redirect_to path
+    redirect_back fallback_location: path
```
sandlerr: kinda makes this method obsolete, we can clean it up later to make the diff here smaller

config/environments/test.rb
```diff
@@ -34,7 +34,13 @@
   config.action_mailer.delivery_method = :test
 
   # Print deprecation notices to the stderr.
-  config.active_support.deprecation = :stderr
+  config.active_support.deprecation = lambda do |message, _backtrace|
+    if message =~ /DEPRECATION WARNING: .* HTTP request methods will accept only/
+      # ignore ... we'll have to deal with that later ...
```
sandlerr: let's log this

config/initializers/zz.rb
```diff
@@ -4,7 +4,7 @@
     [
       ActiveRecord::Base.send(:descendants).map(&:name),
       ActionController::Base.descendants,
-      (File.basename($0) != "rake" ? (defined?(Rake) && "rake") : nil),
+      # (File.basename($0) != "rake" ? (defined?(Rake) && "rake") : nil), # https://github.com/rails/rails/issues/25029
```
sandlerr: might be worth logging this as well, else nobody will remember if that situation ever changes

sandlerr:
I can't start a deploy to see the streaming

screen shot 2016-06-03 at 4 08 52 pm
am I doing something dumb?

grosser:
nah, I saw that issue before but then it was gone ... I'll take another
look :/

config/initializers/zz.rb
grosser: don't want warn noise that we cannot do anything about ... just makes people learn to ignore stuff which I don't want :D

config/environments/test.rb
grosser: I'll fix this up after the merge, but would create too much diff atm ... and I want a clean test run :)

grosser:
all good :D 

deploys / buttons / streaming done!

give it a spin ... then we can merge and do the cleanup tasks / remove deprecations

grosser:
@sandlerr need some sanity checking ... travis fails with being unable to get commit 1c2de72f187104149dea254963ac4c472991a8da ... but it does nowhere sat that it wants that ...

sandlerr:
I can't see that failure in the last travis run, can you link to it?

grosser:
grml ... now it works ... maybe the cache clearing did it ... or it's scared of you ...

app/controllers/deploys_controller.rb
```diff
@@ -71,7 +71,8 @@ def search
       stages = Stage
       stages = stages.where(project: projects) if projects
       if params[:production].present?
-        production = ActiveRecord::Type::Boolean.new.type_cast_from_user(params[:production])
+        production = params[:production]
+        production = production.nil? || !ActiveModel::Type::Boolean::FALSE_VALUES.include?(production)
```
sandlerr: you want this to be true if it is nil?

grosser: `ActiveRecord::Type::Boolean.new.type_cast_from_user(nil) -> nil` ... so no ...
so changing to !nil && !include

sandlerr:
Looks good to me - was able to log in and deploy, can work out the rough edges later.

Notes from reading official upgrade guide:

We can probably create `ApplicationRecord < ActiveRecord::Base` and make everything else inherit from that at this point.

We can probably use `rails` on the commandline instead of `rake` in .travis.yml and script/bootstrap

Kill https://github.com/zendesk/samson/blob/master/config/environments/test.rb#L52

`belongs_to` is now required by default, do we have any optional ones? maybe these:

- https://github.com/zendesk/samson/blob/master/plugins/kubernetes/app/models/kubernetes/release.rb#L6 
- https://github.com/zendesk/samson/blob/c11a1c6752f5b7c12095cee028d3cbf41d62c587/app/models/build.rb#L7
- https://github.com/zendesk/samson/blob/2a35ade6c54b035a51b091ee090da3d2dee28156/app/models/lock.rb#L8
- https://github.com/zendesk/samson/blob/master/app/models/command.rb#L7
- https://github.com/zendesk/samson/blob/master/app/models/deploy.rb#L5-L7

Maybe try `rake rails:update` if it's straightforward or scan through http://railsdiff.org/4.2.7/5.0.0

sandlerr:
looks like migrations still not running on sqlite :(

grosser:
silly sqlite works just fine locally :/

grosser:
sledgehammer fix for shitty problem solves it ...

grosser:
@sandlerr good to go with later cleanup or did you see any blockers ?

grosser:
also applied the rails-diff changes ... left out 2 scary sounding options that we can enable later

config/environments/production.rb
```diff
@@ -32,21 +27,23 @@
   # Do not fallback to assets pipeline if a precompiled asset is missed.
   config.assets.compile = false
 
-  # Generate digests for assets URLs.
-  config.assets.digest = true
```
sandlerr: do we care to turn this off for development?

config/environments/production.rb
```diff
@@ -61,6 +58,11 @@
   # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
   # config.assets.precompile += %w( search.js )
 
+  # Use a real queuing backend for Active Job (and separate queues per environment)
+  # config.active_job.queue_adapter     = :resque
+  # config.active_job.queue_name_prefix = "railsdiff_#{Rails.env}"
```
sandlerr: `railsdiff` was probably their app name, if anyone cares this comment can be changed to `samson`

config/initializers/new_framework_defaults.rb
```diff
@@ -0,0 +1,24 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 5.0 upgrade.
+#
+# Read the Rails 5.0 release notes for more info on each option.
+
+# Enable per-form CSRF tokens. Previous versions had false.
+Rails.application.config.action_controller.per_form_csrf_tokens = true
```
sandlerr: works with `form_tag` or need to switch to `form_for`?

config/initializers/new_framework_defaults.rb
```diff
@@ -0,0 +1,24 @@
...
+
+# Enable origin-checking CSRF mitigation. Previous versions had false.
+Rails.application.config.action_controller.forgery_protection_origin_check = true
+
+# Make Ruby 2.4 preserve the timezone of the receiver when calling `to_time`.
+# Previous versions had false.
+ActiveSupport.to_time_preserves_timezone = true
+
+# Require `belongs_to` associations by default. Previous versions had false.
+# Rails.application.config.active_record.belongs_to_required_by_default = true
```
sandlerr: so if this is commented does that mean we get the new default anyway? We should be able to set `required: false` on the optional ones but that can be done later

sandlerr:
looks like it is worth a try, let's give it a run on staging :+1:

config/initializers/new_framework_defaults.rb
grosser: no idea ... suppose it works with both ...

config/environments/production.rb
grosser: new defaults said it's no longer needed ...   page renders with  `<script src="/assets/application-fd416841b5adf3d3928cc66d8fc34e880ee4c38a8abb0154706b716096750bfc.js"></script>` and it works ...

grosser:
hmm failed ... `uninitialized constant #<Class:0x007fca374ee540>::BuddyCheck`

grosser:
https://github.com/zendesk/samson/pull/1159

grosser:
failed again with `uninitialized constant #<Class:0x007f2b5c8bb5d0>::BuddyCheck` ... I'll have to play with staging env then ...

sandlerr:
I think we probably shouldn't autoload `lib` at all?

grosser:
no we should ... otherwise you'll need requires everywhere

grosser:
now fails to boot ...

```
! Unable to load application: NameError: uninitialized constant Generators::NamedBase
/data/samson/releases/20160802-033751-3585f84/lib/generators/plugin/plugin_generator.rb:1:in `<top (required)>': uninitialized constant Generators::NamedBase (NameError)
```

grosser:
all working and did a test-deploy too ... should be good ... I'll merge when I come back from PTo to not let you sit on this timebomb :D

sandlerr:
Oof thanks for pushing this to 🍏 !

grosser:
here goes nothing ...

PR #973, opened 2016-05-12
Fix k8s log

rata:
* Use a better default for KUBER_LOGFILE

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/util.rb
+++ b/plugins/kubernetes/app/models/kubernetes/util.rb
@@ -14,11 +14,14 @@ module Kubernetes
     end
 
     def self.log(message, extra_info = {})
-      @output_file ||= File.open(ENV['KUBER_LOGFILE'] || File.join(Rails.root, 'log', 'kubernetes.log'), 'a')
-
-      @output_file.write({ message: message }.merge(extra_info).to_json + "\n")
-      @output_file.flush
-      message
+      msg_log = { message: message }.merge(extra_info).to_json
+      if ENV['KUBER_LOGFILE']
+        @output_file ||= File.open(ENV['KUBER_LOGFILE'], 'a')
+        @output_file.write(msg_log + "\n")
+        @output_file.flush
+      else
+        Rails.logger.info(msg_log)
+      end
     end
   end
 end


grosser:
why is log not good ... that's where logs should go :D 

how about setting the ENV var ?

rata:
On Thu, May 12, 2016 at 11:43:42AM -0700, Michael Grosser wrote:
> why is log not good ... that's where logs should go :D 

Because it doesn't exist in the docker image, at least. So a deployment fails
because it can't be created (the dir doesn't exist, the 'a' creates only the
file)

> 
> how about setting the ENV var ?

Yes, I'm doing that. But thought another default might help. But it's fine by me
:)

grosser:
hmmm let's create the log folder in docker then ... but that seems kind of weird too ... should go to stdout for docker ? 

@jonmoter thoughts ?

jonmoter:
How about if the `KUBER_LOGFILE` env var isn't set, it just uses the Rails logger?

rata:
On Thu, May 12, 2016 at 11:54:24AM -0700, Jon Moter wrote:
> How about if the `KUBER_LOGFILE` env var isn't set, it just uses the Rails logger?

Oh, that sound great IMHO! Because there are standard/easy ways to handle that
(stdout in my case) in a kubernetes deployment. And we don't have to worry about
the pod disk usage if no logs are created as files.

In our pods, for example, we use a log collector to avoid having files there and
just send them over the network.

Having it use the rails logger, that can use stdout, is great for us at least!
:)

jonmoter:
Yeah, I originally created that as a separate log 'cause I wanted to be easily able to tail the Kubernetes-related stuff while doing development, without seeing all the other noise in the rails logger.  But for most use-cases we probably want it to be combined with the other Samson logs.

rata:
@jonmoter So, should I replace self.log to use Rails.logger? Which severity? Or how do you want to do it?

grosser:
I'd say Rails.logger.info ... so it will still show in production ... check out if it survives the trip through the production logger with all the syslog/lograge conversion

rata:
@grosser I've done that. Is it ok?

grosser:
👍

grosser:
waiting for travis :(

grosser:
fixing travis error + removing the ENV var in https://github.com/zendesk/samson/pull/978

rata:
@grosser sorry is my first time doing web programming (I'm a C programming guy) and never used ruby. Sorry for the bug :-/

grosser:
Pah, your doing great :D

PR #974, opened 2016-05-12
get rid of deprecated newrelic gem

grosser:
@zendesk/samson 

 - activeresource is deprecated and blocks rails 5
 - changed input into a select since we know all the app names
 - using api v2 and json instead of v1 with xml

# Risks
 - Medium: newrelic UI breaking

screen shot 2016-05-12 at 11 28 42 am
screen shot 2016-05-12 at 11 41 35 am


pschambacher:
LGTM 👍

PR #979, opened 2016-05-13
simplify brakeman setup by using its builtin engine support

grosser:
... also caught 1 more false-positive, so views were previously not analyzed 

@zendesk/samson


jonmoter:
👍

PR #980, opened 2016-05-13
mime types needs 0.3s to load ... we do not need all these types

grosser:
@zendesk/samson

standalone time is 0.3s `fork { puts(Benchmark.realtime { require 'mime/types' })}`

when running a benchmark this only yielded 0.1s improvement, but it's a step in the right direction ...

```
without:
runs: 10
total: 54.48
average: 5.45
range: 5.35..5.54

with:
runs: 10
total: 53.86
average: 5.39
range: 5.27..5.82
```


sandlerr:
+1

PR #981, opened 2016-05-13
RFC: Reduce docker image size

rata:
This reduces the docker image size, while being quite not-intrusive, in ~23%. It's useful for us because our internet connection is not really fast. What do you guys think of a change like this? (I've only build tested it, but will test it correctly if you are onboard with such a PR)

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

This reduces the docker image size from 1GB to 790MB, 23% smaller.
Further changes may be possible, but they are probably more intrusive.
These are pretty simple and yet the shrink is considerable. So this
feels good enough for a start, at least.

This is useful for us because our internet connection is pretty slow.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM ruby:2.2.3
+FROM ruby:2.2.3-slim
 
 RUN apt-get update && apt-get install -y wget apt-transport-https
 RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
 RUN echo 'deb https://deb.nodesource.com/node_0.12 jessie main' > /etc/apt/sources.list.d/nodesource.list
-RUN apt-get update && apt-get install -y nodejs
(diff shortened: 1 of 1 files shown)


grosser:
meh ... let's go alpine ? :D

rata:
@grosser What I don't like about alpine is the support for the packages in a version, it's like 6 moths or something like that. But it's not deal breaker for me on samson, I think.

Do you think it's possible to merge this and the full Dockerfile rewrite with alpine come later?

grosser:
sure ... a bit of extra download / possible bugs ... but looks safe enough ...

rata:
@grosser cool, thanks! Will probably work on a even smaller Dockerfile when testing a newer version of samson. Or is there any rush on your side?

grosser:
nah ... we are not deploying samson via docker, so would only be nice for
local development ... no rush :)

PR #983, opened 2016-05-14
avoid duplicates by not selecting all rows for global admins

grosser:
@zendesk/samson

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -5,7 +5,7 @@ module Searchable
     def search_by_criteria(criteria)
       scope = self
       scope = scope.search(criteria[:search]) if criteria[:search]
-      scope.order("#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
+      scope.order("#{table_name}.#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
     end
 
     private
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,11 +27,8 @@ class User < ActiveRecord::Base
     where("name LIKE ? OR email LIKE ?", "%#{query}%", "%#{query}%")
   }
   scope :with_role, -> (role_id, project_id) {
-    joins('LEFT OUTER JOIN user_project_roles ON users.id = user_project_roles.user_id').
-      where(
-        '(user_project_roles.project_id = ? AND user_project_roles.role_id >= ?) OR users.role_id >= ?',
-        project_id, role_id, role_id
-      )
+    joins("LEFT OUTER JOIN user_project_roles ON users.id = user_project_roles.user_id AND user_project_roles.project_id = #{project_id.to_i}").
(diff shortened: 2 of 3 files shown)


app/models/concerns/searchable.rb
```diff
@@ -5,7 +5,7 @@ module ClassMethods
     def search_by_criteria(criteria)
       scope = self
       scope = scope.search(criteria[:search]) if criteria[:search]
-      scope.order("#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
+      scope.order("#{table_name}.#{sort_column(criteria[:sort])} #{sort_direction(criteria[:direction])}").page(criteria[:page])
```
grosser: not needed, but now this can be used with group statements too ...

app/models/user.rb
```diff
@@ -27,10 +27,11 @@ class User < ActiveRecord::Base
...
   scope :with_role, -> (role_id, project_id) {
-    joins('LEFT OUTER JOIN user_project_roles ON users.id = user_project_roles.user_id').
+    joins('LEFT JOIN user_project_roles ON users.id = user_project_roles.user_id').
       where(
-        '(user_project_roles.project_id = ? AND user_project_roles.role_id >= ?) OR users.role_id >= ?',
-        project_id, role_id, role_id
+        '(user_project_roles.project_id IS NULL AND users.role_id >= ?) OR ' \
+        '(user_project_roles.project_id = ? AND user_project_roles.role_id >= ?)',
+        role_id, project_id, role_id
       )
```
grosser: before we selected users multiple times if they had multiple user_project_roles

rikeoka:
Security filed JIRA: SAMSON-240 for this bug as well.

grosser:
looks good / +1 ?

rikeoka:
I noticed the same thing that I was seeing when I tried this approach before.

List of users ordered before filtering:
screen shot 2016-05-16 at 1 51 53 pm

With Search filter:
screen shot 2016-05-16 at 1 52 06 pm

Looking into if my test data was corrupted.

app/models/user.rb
```diff
@@ -27,10 +27,11 @@ class User < ActiveRecord::Base
     where("name LIKE ? OR email LIKE ?", "%#{query}%", "%#{query}%")
   }
   scope :with_role, -> (role_id, project_id) {
-    joins('LEFT OUTER JOIN user_project_roles ON users.id = user_project_roles.user_id').
+    joins('LEFT JOIN user_project_roles ON users.id = user_project_roles.user_id').
```
rikeoka: joins("LEFT OUTER JOIN user_project_roles ON users.id = user_project_roles.user_id AND user_project_roles.project_id = #{project_id.to_i}").
      where('users.role_id >= ? OR user_project_roles.role_id >= ?', role_id, role_id)

grosser:
added test that fails without the changed sql ... good to go ?

rikeoka:
Addresses that disappearing users bug I was noticing.  Not sure if I can give it the ok not being a core team member.

grosser:
You probably know more about the user permissions then anyone else atm :D

rikeoka:
👍

PR #985, opened 2016-05-14
make sse connection lazy and active deploys read value from html

grosser:
@zendesk/samson @henders 

one of the most wasteful things in samson is that we open a streaming connection (stageful / expensive) on every page load ... and do a request to active_count ... for the teeny-tiny benefit of seeing an active deploy count ... this PR is killing most of the overhead while still keeping the functionality intact ... (I'd even go so far as to make this badge completely static since I never use it and clicking on `Deploying` would still show the active count / I don't think anyone cares how many deploys are going on ...)

 - open streaming connection on demand vs always
 - query for deploy updates after the page is open for a while vs always
 - cache active deploy count vs db calls on every page load
 - use html data vs doing additional request

## Risks
 - Low: deploy badge breaking or being out of sync
 - Low: other streaming features breaking

### DIFF
--- a/app/assets/javascripts/controllers/current_deploys_badge_ctrl.js
+++ b/app/assets/javascripts/controllers/current_deploys_badge_ctrl.js
@@ -1,33 +1,37 @@
 samson.controller('currentDeployBadgeCtrl', function($scope, $http, SseFactory) {
   'use strict';
 
-  $scope.currentActiveDeploys = 0;
+  var $badge = $('#currentDeploysBadge');
+  $scope.currentActiveDeploys = parseInt($badge.data('count'), 10);
 
-  SseFactory.on('deploys', function(msg) {
-    if (msg.type === 'new') {
-      $scope.currentActiveDeploys += 1;
-    } else if (msg.type === 'finish') {
-      $scope.currentActiveDeploys -= 1;
-    }
-    updateBadge();
-  });
+  // once a user has the page open for a while:
+  // - get current active count to reduce race condition
+  // - connect to stream for updates
+  setTimeout(function(){
+    $http.get('/deploys/active_count.json').success(function(result) {
+      $scope.currentActiveDeploys = result.deploy_count;
+      updateBadge();
+
+      SseFactory.on('deploys', function(msg) {
+        if (msg.type === 'new') {
+          $scope.currentActiveDeploys += 1;
+        } else if (msg.type === 'finish') {
+          $scope.currentActiveDeploys -= 1;
+        }
+        updateBadge();
+      });
+    });
+  }, 5000);
 
   function updateBadge() {
     _.defer(function() { $scope.$apply(); });
     if ($scope.currentActiveDeploys > 0) {
-      $('#currentDeploysBadge').show();
+      $badge.show();
     } else {
       $scope.currentActiveDeploys = 0;
-      $('#currentDeploysBadge').hide();
+      $badge.hide();
     }
   }
 
-  function init() {
-    $http.get('/deploys/active_count.json').success(function(result) {
-      $scope.currentActiveDeploys = result.deploy_count;
-      updateBadge();
-    });
-  }
-
-  init();
+  updateBadge();
 });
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/controllers/current_deploys_badge_ctrl.js
```diff
@@ -1,33 +1,30 @@
...
+      $badge.show();
     } else {
       $scope.currentActiveDeploys = 0;
-      $('#currentDeploysBadge').hide();
+      $badge.hide();
     }
   }
 
-  function init() {
-    $http.get('/deploys/active_count.json').success(function(result) {
```
henders: You will still need this call otherwise the badge count will be potentially incorrect since it'll miss all the starting/stopping deploy events during the 5 seconds.

grosser: Yeah :/

What do you think about getting rid of the live counter ?

henders: I remember the live counter was explicitly requested by a few folks (I think mostly HC devs?).
So people are/were definitely using it. And I also check it every now and again myself.

We could change the 
- Put back in the ```$http.get('/deploys/active_count.json')``` line when the timer kicks in
- Or I remember the live counter was explicitly requested by a few folks (I think mostly HC devs?).
So people are/were definitely using it. And I also check it every now and again myself.

We could change the deploy SSE msgs to instead of sending a increment/decrement type msg, have it send the current total number of active deploys. Which would be helped by your cache entry you added for the deploy count.

Or else put back in the ```$http.get('/deploys/active_count.json')``` line when the timer kicks in, and at least have the benefit of reducing the load on the initial page load?

grosser:
I think this feature is mostly pointless and is wasting engineering time / server resoruces
 - 99% of users don't need to see things jump from 2-3-2-1 etc
 - the number of current deploys is not meaningful / does not tell any important or accurate information (except during server restarts I guess ...)
 - it is not a feature we `must` have ... just some obscure addon

how about we make it static and see if anyone even notices / complains and then discuss with them why they need it ?

henders:
Hmm... I kinda like it as it is. And I know other people use it. Why don't we just do the 5second delay in establishing the connection and getting the current count at the same time. This will reduce a lot of load anyway since at least I click through lots of screens to get to where I want without staying longer then 5 seconds generally. So same probably holds true for others.

grosser:
ok ...
I don't understand how that counter is useful ... can you explain ?
It shows how many projects are currently deploying ... so no indication of
what your deploy/project is doing ...

henders:
Aye, I use it mainly as a sanity check that if my deploy is going slowly, I can look up at the top and see how much load Samson is under.

grosser:
updated with call to uncached active_count after initial timeout to reduce race condition ... good to go ?

app/controllers/deploys_controller.rb
```diff
@@ -21,18 +21,18 @@ def index
...
-    render json: { deploy_count: active_deploy_scope.count }
-  end
-
   def active
     @deploys = active_deploy_scope
     render partial: 'shared/deploys_table', layout: false if params[:partial]
   end
 
+  def active_count
+    render json: Deploy.active.count
```
henders: Not want to use the cache here?

henders: '''deploy.active_count'''

config/routes.rb
```diff
@@ -17,7 +17,6 @@
     resources :deploys, only: [:index, :show, :destroy] do
       collection do
         get :active
-        get :active_count
```
henders: Need this line back right?

grosser: nope, no per-project active counter

app/controllers/deploys_controller.rb
grosser: Either we get the live count here (0-1s race condition) or we don't do the
http call (0-10s race condition) ...

henders: Good point

config/routes.rb
henders: oh yea! oops.

henders:
👍

PR #986, opened 2016-05-16
bump mini profiler we will need it for rails 5

grosser:
@henders ... just to not spam everyone ;)


henders:
hah, easy one! :)
👍

PR #987, opened 2016-05-16
add git to the docker image, disable interactive host key checking fo…

sandlerr:
@zendesk/samson I think it is reasonable to actually include `git` for a tool that shells out to git sometimes.

The second change I'm not so sure about, but I got an interactive prompt in the logs. Something like

```
The authenticity of host 'github.com (65.74.177.129)' can't be established.
RSA key fingerprint is 16:27:ac:a5:76:28:2d:36:63:1b:56:4d:eb:df:a6:48.
Are you sure you want to continue connecting (yes/no)?
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM ruby:2.2.3-slim
 
-RUN apt-get update && apt-get install -y wget apt-transport-https
+RUN apt-get update && apt-get install -y wget apt-transport-https git
 RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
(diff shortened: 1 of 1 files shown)


grosser:
Number 1 looks good ...

the prompt is annoying ... but shipping insecure ssh seems like a bad idea ... how about seeding the know hosts ?

sandlerr:
you mean adding to the `known_hosts` file? Github uses a /22 ( https://help.github.com/articles/what-ip-addresses-does-github-use-that-i-should-whitelist/ ) so that would be a lot of addresses. Not sure if they all use the same keys either. I pulled that change out of this PR so we can solve that problem later, can I merge?

grosser:
+1

PR #988, opened 2016-05-16
PAAS-150 auto seed + project new fixes

grosser:
@zendesk/paas

### DIFF
--- a/plugins/kubernetes/app/decorators/stage_decorator.rb
+++ b/plugins/kubernetes/app/decorators/stage_decorator.rb
@@ -1,5 +1,6 @@
 Stage.class_eval do
   validate :validate_deploy_groups_have_a_cluster, if: :kubernetes
+  after_create :seed_kubernetes_roles
 
   private
 
@@ -7,5 +8,9 @@ Stage.class_eval do
     if kubernetes && bad = deploy_groups.reject(&:kubernetes_cluster).presence
       errors.add(:kubernetes, "Deploy groups need to have a cluster associated, but #{bad.map(&:name).join(', ')} did not.")
     end
+  end
+
+  def seed_kubernetes_roles
+    Kubernetes::Role.seed! project, 'master' if kubernetes
   end
 end
--- a/plugins/kubernetes/app/models/kubernetes/role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role.rb
@@ -61,7 +61,7 @@ module Kubernetes
     end
 
     def label_name
-      name.parameterize('-')
+      name.parameterize
     end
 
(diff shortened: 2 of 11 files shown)


plugins/kubernetes/app/decorators/stage_decorator.rb
```diff
@@ -8,4 +9,8 @@ def validate_deploy_groups_have_a_cluster
       errors.add(:kubernetes, "Deploy groups need to have a cluster associated, but #{bad.map(&:name).join(', ')} did not.")
     end
   end
+
+  def seed_kubernetes_roles
+    Kubernetes::Role.seed! if kubernetes
```
grosser: this does not produce any bad side-effects since it only adds roles

grosser:
screen shot 2016-05-16 at 12 50 00 pm

grosser:
this PR is already too big ... got a few more issues I need to solve, but it's a start ...

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -24,13 +24,18 @@ class Role < ActiveRecord::Base
     # create initial roles for a project by reading kubernetes/*{.yml,.yaml,json} files into roles
     def self.seed!(project, git_ref)
       kubernetes_config_files_in_repo(project, git_ref).each do |config_file|
+        scope = where(project: project)
+        next if scope.where(config_file: config_file.file_path).exists?
+
         service_name = config_file.service && config_file.service.metadata.name
-        service_name << "-#{rand(9999999)}" if where(service_name: service_name).exists?
+        service_name << "-#{rand(9999999)}" if service_name && scope.where(service_name: service_name).exists?
```
irwaters: do we really want to create a random service name, or should this require user intervention?

grosser: this is happening in the background, so we can either:
 - create an obviously bad name (this PR)
 - fail and block (would block stage/project creation or seeding of other roles ... bad)
 - skip this role ... I think that's bad too since it might go unnoticed ...

maybe we simply do not allow deploying with on of these generated names ... so deploys will fail with an obvious error ...

irwaters: I think this is fine for now, and create a story to investigate that behavior later?

irwaters:
👍

plugins/kubernetes/app/models/kubernetes/role.rb
grosser: I'll make a followup PR with the `do not allow generated services to deploy`

PR #989, opened 2016-05-16
Fix UI Bug with user project roles

rikeoka:
Fix a UI issue where if a user has user_project_roles the user is presented multiple times in the project/user pages. 

Before
screen shot 2016-05-16 at 12 16 34 pm

After
screen shot 2016-05-16 at 12 16 51 pm

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: SAMSON-240

### Risks
- Level: Low

Fix UI Bug with user project roles

I think this is cleaner, also doesn't result in user not appearing if a user has a lower user_project_role.role_id than their user.role_id.

Alternatively could replace “OR users.role_id >=?” with “OR
(user_project_roles.project_id IS NULL AND user.role_id >=?)”, but it would be susceptible to the 'disappearing' bug.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -31,7 +31,7 @@ class User < ActiveRecord::Base
       where(
         '(user_project_roles.project_id = ? AND user_project_roles.role_id >= ?) OR users.role_id >= ?',
         project_id, role_id, role_id
-      )
+      ).distinct
   }
 
   def starred_project?(project)


grosser:
already opened 3 days ago :(
https://github.com/zendesk/samson/pull/983

grosser:
I think my PR is the more efficient fix / solves the root problem ... and it has tests ... if you see a bug in it, let's add a test for it there ...

rikeoka:
ok closing this

grosser:
FYI delete branches to make git faster / github autocomplete faster / repo smaller

PR #994, opened 2016-05-17
Fix k8s deployment

rata:
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
@@ -68,7 +68,7 @@ module Kubernetes
 
     def deployment_labels
       # Deployment labels should not include the ids of the release, role or deploy groups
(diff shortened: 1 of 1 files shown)


grosser:
isn't it nice to be able to query deploys by tag ?

the issue is writing it to the env again ?

rata:
@grosser query how?

The except is to not match it when doing a deploy, if not it will not replace the old version as the revision/tag are different (as they are supposed to, they are different version because they are different deploys). It's just to match which pods to replace, IIUC

grosser:
ah yes

grosser:
should be a .slice then since we don't want to put anything in there by accident ?
also some test that makes sure nothing gets added ...

grosser:
I'll put that into my PR as well ... should be done soon ...

rata:
@grosser oh, thanks! Let me know when it is merged! Thanks =)

PR #995, opened 2016-05-17
fix tag failing to be stored in env when it contains a / #993

grosser:
@rata @irwaters 

 - replaces #994 
 - replaces #993

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
@@ -66,9 +66,11 @@ module Kubernetes
       end
     end
 
+    # labels that are used to match the previous replicasets,
+    # they cannot change or we will create a new replicasets instead of updating the previous one
+    # renaming roles or renaming projects will lead to duplicate replicasets
     def deployment_labels
-      # Deployment labels should not include the ids of the release, role or deploy groups
-      release_doc_metadata.except(:release_id, :role_id, :deploy_group_id)
+      release_doc_metadata.slice(:project, :role)
     end
 
     # Sets the metadata that is going to be used as the selector. Kubernetes will use this metadata to select the
@@ -102,14 +104,14 @@ module Kubernetes
           project: release.project.permalink,
           project_id: release.project_id,
 
-          role: role.name,
+          role: role.name.parameterize,
           role_id: role.id,
 
           deploy_group: deploy_group.env_value.parameterize,
           deploy_group_id: deploy_group.id,
 
           revision: build.git_sha,
-          tag: build.git_ref
+          tag: build.git_ref.parameterize
         }
       end
     end
--- a/plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
@@ -47,26 +47,35 @@ describe Kubernetes::DeployYaml do
       metadata = result.fetch(:metadata)
       metadata.fetch(:namespace).must_equal 'pod1'
       metadata.fetch(:labels).must_equal(
-        project_id: doc.kubernetes_release.project_id.to_s,
-        revision: "1a6f551a2ffa6d88e15eef5461384da0bfb1c194",
-        tag: "v123",
-        deploy_id: "123",
-        project: "foo",
-        role: "app_server",
-        deploy_group: "pod1",
+        project: doc.kubernetes_release.project.permalink,
+        role: doc.kubernetes_role.name
       )
     end
 
     it "works without selector" do
       assert doc.raw_template.sub!('selector:', 'no_selector:')
       result = yaml.to_hash
-      result.fetch(:spec).fetch(:selector).fetch(:matchLabels).keys.sort.must_equal([:deploy_group, :deploy_id, :project, :project_id, :revision, :role, :tag].sort)
+      result = result.fetch(:spec).fetch(:selector).fetch(:matchLabels).keys.sort
+      result.must_equal [:project, :role].sort
     end
 
     it "works without labels" do
       yaml.send(:template).metadata.labels = nil
+      result = yaml.to_hash.fetch(:metadata).fetch(:labels).keys.sort
+      result.must_equal [:project, :role].sort
+    end
+
+    it "escapes things that would not be allowed in labels or environment values" do
+      doc.deploy_group.update_column(:env_value, 'foo:bar')
+      doc.kubernetes_role.update_column(:name, 'Oh boy this is weird')
+      doc.build.update_column(:git_ref, 'user/feature')
+
       result = yaml.to_hash
-      result.fetch(:metadata).fetch(:labels).keys.sort.must_equal([:deploy_id, :project, :project_id, :role, :deploy_group, :revision, :tag].sort)
+      result.fetch(:spec).fetch(:template).fetch(:metadata).fetch(:labels).slice(:deploy_group, :role, :tag).must_equal(
+        tag: "user-feature",
+        role: "oh-boy-this-is-weird",
+        deploy_group: 'foo-bar'
+      )
     end
 
     describe "deployment" do
@@ -112,11 +121,6 @@ describe Kubernetes::DeployYaml do
           [:REVISION, :TAG, :PROJECT, :ROLE, :DEPLOY_ID, :DEPLOY_GROUP, :POD_NAME, :POD_NAMESPACE, :POD_IP].sort
         )
         env.map { |x| x[:value] }.map(&:class).map(&:name).sort.uniq.must_equal(["NilClass", "String"])
-      end
-
-      it "removes : in env values since they would not validate against kubernetes (([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?" do
-        doc.deploy_group.update_column(:env_value, 'foo:bar')
-        container.fetch(:env).detect { |v| break v if v[:name] == :DEPLOY_GROUP }.fetch(:value).must_equal 'foo-bar'
       end
 
       it "fails without containers" do


grosser:
nvm ... the bug was for labels too ... reworking

grosser:
how about now ?

rata:
@grosser Yes, I confirm the parametric fixes it too!

grosser:
does this look good ?

```
   # labels that are used to match the previous deployment,
    # they cannot change or we will create a new deployment instead of updating the previous one
    # updating role names or re-creating the deploy group will lead to duplicate deployments
    def deployment_labels
      release_doc_metadata.slice(:project_id, :role, :deploy_group_id)
    end
```

rata:
@grosser seems fine. But why not role_id instead of role?

grosser:
yeah ok ... should be stable now ... I'm a bit afraid of people doing deletes in the UI ... but should be fine ...

rata:
@grosser whatever you feel more comfortable :)

grosser:
also makes it consistent ... and that makes me comfortable ;)

rata:
@grosser :-D Seems fine by me this way (and I've seen the slice fix code, just in case)

rata:
@grosser Did you detect any problems with this? Or waiting for merging because you are unsure about something in here?

grosser:
just waiting for someone to review the PR ... should be done soon

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -66,9 +66,11 @@ def set_deployment_metadata
       end
     end
 
+    # labels that are used to match the previous deployment,
+    # they cannot change or we will create a new deployment instead of updating the previous one
+    # re-creating roles or re-creating the deploy group will lead to duplicate deployments
     def deployment_labels
-      # Deployment labels should not include the ids of the release, role or deploy groups
-      release_doc_metadata.except(:release_id, :role_id, :deploy_group_id)
+      release_doc_metadata.slice(:project_id, :role_id, :deploy_group_id)
```
jonmoter: I'd like to keep the nice names in the Deployment labels.  It's safe to change them when updating a deployment. The thing that needs to stay consistent is the `metadata.name` field.

grosser: @rata 
we update deployments by name and namespace, so having changing labels should be fine ?

grosser: just tried a few deploys without any label restrictions and it won't take down the pods when deleting the deploy ... so we create duplicate pods and they never get cleaned up

with this .slice implementation the validation failed:

```
JobExecution failed: Deployment.extensions "truth-service-app-server" is invalid: spec.template.metadata.labels: Invalid value: {"deploy_group":"production","deploy_group_id":"1","deploy_id":"111","project":"example-project","project_id":"1","release_id":"79","revision":"617182f02ecdf2b38a46978e00d2ff755811c098","role":"app-server","role_id":"6","tag":"617182f02ecdf2b38a46978e00d2ff755811c098"}: `selector` does not match template `labels
```

since the project name in samson and the kubernetes file did not match ... good times

so the simple solution is to for now only put role and project in there ... since that is the only thing we need to be unique ... and for future PR stop messing with the labels altogether ...

rata: Sorry, was coming home.

Not sure I understand, what is to "keep the nice names in the Deployment labels"?

I think this is used to match the pods to replace only (as the previous comment also says). Without this change (or technically, the one I did that is very similar), when creating a new release, the old pods continue to be there running. Looking quickly at the file, it seems because it is used in set_selector_metadata() (like the comment in that func says, it is used for matching).

The tag, revision, etc. labels are in the pods. I confirm I run this and the labels were there after a new deployment. This change does not remove those labels from the pods, if I tested correctly.

grosser: yep, just confirmed that too ... updating it to only use project and role ... does that work for you too ?

rata: @grosser Oh, the difference is probably that I tested the "except" and you used an "slice"

grosser: we decided on stopping to mess with labels altogether and only add annotations (which will avoid any kind of matching issues), sounds good ?

rata: On Tue, May 17, 2016 at 04:03:58PM -0700, Michael Grosser wrote:
> >      def deployment_labels
> > -      # Deployment labels should not include the ids of the release, role or deploy groups
> > -      release_doc_metadata.except(:release_id, :role_id, :deploy_group_id)
> > +      release_doc_metadata.slice(:project_id, :role_id, :deploy_group_id)
> 
> yep, just confirmed that too ... updating it to only use project and role ... does that work for you too ?

Hmm, not sure that will be enough. Maybe it is because staging/production, if
they are in the same cluster, will be in different namespaces. But with the
environment (deploy_group I think?) for sure it is unique

grosser: we use a different namespace for every environment, so role and project are unique enough ... and it allows us to not have to change any labels around and still be able to deploy + use local kubectl create too ... works for you too ?

grosser: updated PR with the `only role and project` change + confirmed it works on our cluster

rata: On Tue, May 17, 2016 at 04:05:46PM -0700, Michael Grosser wrote:
> >      def deployment_labels
> > -      # Deployment labels should not include the ids of the release, role or deploy groups
> > -      release_doc_metadata.except(:release_id, :role_id, :deploy_group_id)
> > +      release_doc_metadata.slice(:project_id, :role_id, :deploy_group_id)
> 
> we decided on stopping to mess with labels altogether and only add annotations (which will avoid any kind of matching issues), sounds good ?

I think labels are more easy to obtain and query with kubectl (and maybe show in
a UI too).

I think this is becoming messy because there are several labels in the pod and
we are not treating them differently. One thing is the selector (spec.selector),
that uses labels to match which pods should be replaced, and one other is the
labels we use in a spec.template.metadata.labels that the labels there should be
the labels that a pod has when it's been created with this deployment.

We can have project, role and maybe deploy_group in the selector, and add the
revision, tag, etc. in spec.template.metadata.labels and, therefore, it should
be in the pods running as labels.

I think it will be better to have the selector and the labels used for the pods
separated (maybe one is the others plus some, so the selector will for sure
match, and we add stuff about this particular release we want to have as
labels).

What do you think of something like this?

I think it can be pretty clean and have all the advantages of using labels

rata: On Tue, May 17, 2016 at 04:09:24PM -0700, Michael Grosser wrote:
> >      def deployment_labels
> > -      # Deployment labels should not include the ids of the release, role or deploy groups
> > -      release_doc_metadata.except(:release_id, :role_id, :deploy_group_id)
> > +      release_doc_metadata.slice(:project_id, :role_id, :deploy_group_id)
> 
> we use a different namespace for every environment, so role and project are unique enough ... and it allows us to not have to change any labels around and still be able to deploy + use local kubectl create too ... works for you too ?

Yes, role and project should be enough. Not sure what is the problem of using
the deploy_group too in the slice. It's just for sure unique  (doesn't depend on
the configuration of the cluster) and doesn't cause any problems with updating
labels on deploy. Right?

grosser: yes, having base labels for matching and then adding to them sounds clean.

Do you need labels other than project and role (what can be statically added to the yml files) in the deployment or selector ? 

If yes, then we can stop adding labels and avoid all the mess (maybe still have additional labels for pods, but no longer override labels on anything else)

grosser: ah perfect, so I'll keep this PR as it is and make a new one that no longer modifies deployment and selector and only adds (never overrides) labels to pods -> yml file must be valid / we will never make it invalid :D

jonmoter:
👍

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
rata: On Tue, May 17, 2016 at 04:26:26PM -0700, Michael Grosser wrote:
> >      def deployment_labels
> > -      # Deployment labels should not include the ids of the release, role or deploy groups
> > -      release_doc_metadata.except(:release_id, :role_id, :deploy_group_id)
> > +      release_doc_metadata.slice(:project_id, :role_id, :deploy_group_id)
> 
> yes, having base labels for matching and then adding to them sounds clean.
> 
> Do you need labels other than project and role (what can be statically added to the yml files) in the deployment or selector ? 

No, I don't :-)

> 
> If yes, then we can stop adding labels and avoid all the mess (maybe still have additional labels for pods, but no longer override labels on anything else)

There is one issue that we may want to ignore, but at least decide to ignore and
not without noticing.

The selector is optional, see [1] (it says required: false). If no selector is
specified in the yaml, then the ones from the template are used IIRC.

So, we may want to add a selector if there is none, specifying the role and
project labels (maybe deploy_group too).

That way we are really nice with everyone, I guess.

Or we can force to use a selector and be done with it (but validate it, because
no selector is valid for k8s). The ugly thing about this is that not all yamls valid
for k8s are valid for samson.

[1]: http://kubernetes.io/docs/api-reference/extensions/v1beta1/definitions/#_v1beta1_deploymentspec

grosser: Good point, I'd lean on the side of validating ... I want to stay out of label management business :)

rata: On Tue, May 17, 2016 at 04:55:10PM -0700, Michael Grosser wrote:
> >      def deployment_labels
> > -      # Deployment labels should not include the ids of the release, role or deploy groups
> > -      release_doc_metadata.except(:release_id, :role_id, :deploy_group_id)
> > +      release_doc_metadata.slice(:project_id, :role_id, :deploy_group_id)
> 
> Good point, I'd lean on the side of validating ... I want to stay out of label management business :)

Fare enough :-)

Thanks a lot!!

PR #997, opened 2016-05-17
encapsulate service logic in release doc and validate that service is correct

grosser:
Includes https://github.com/zendesk/samson/pull/996 ... only review second commit

@zendesk/paas


jonmoter:
👍

PR #999, opened 2016-05-18
stop changing labels, the template is responsible for getting these r…

grosser:
…ight

 - validate deploy/template/service labels are in sync (at deploy time)
 - validate labels make the service/deploy unique (at deploy time)

@zendesk/paas @rata

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
@@ -13,8 +13,6 @@ module Kubernetes
         set_rc_unique_label_key
         set_namespace
         set_replica_target
-        set_deployment_metadata
-        set_selector_metadata
         set_spec_template_metadata
         set_docker_image
         set_resource_usage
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -128,15 +96,20 @@ def set_docker_image
...
     end
 
+    # helpful env vars, also useful for log tagging
     def set_env
       env = (container.env || [])
 
       # static data
       metadata = release_doc_metadata
-      [:REVISION, :TAG, :PROJECT, :ROLE, :DEPLOY_ID, :DEPLOY_GROUP].each do |k|
+      [:REVISION, :TAG, :DEPLOY_ID, :DEPLOY_GROUP].each do |k|
```
rata: Not sure I follow, but does means that the rev, tag, etc. will be added as labels, and the yaml file should have the project and roles labels in the template for the pods that will be created?
Not that I have a problem with that, just want to be sure I understand (sorry, I'm a newby in web and ruby :))

grosser: - rev/tag/etc will only be added to the pods and not the deployments/services/selectors
 - existing tags will not be overirtten
 - template is responsible for setting consistent role/project labels everywhere

sounds good ?

rata: On Wed, May 18, 2016 at 08:55:20AM -0700, Michael Grosser wrote:
> >      def set_env
> >        env = (container.env || [])
> >  
> >        # static data
> >        metadata = release_doc_metadata
> > -      [:REVISION, :TAG, :PROJECT, :ROLE, :DEPLOY_ID, :DEPLOY_GROUP].each do |k|
> > +      [:REVISION, :TAG, :DEPLOY_ID, :DEPLOY_GROUP].each do |k|
> 
>  - rev/tag/etc will only be added to the pods and not the deployments/services/selectors
>  - existing tags will not be overirtten
>  - template is responsible for setting consistent role/project labels everywhere
> 
> sounds good ?

Perfect! :-)

jonmoter:
Sure, let's give it a try.  👍

PR #1000, opened 2016-05-18
rubocop

grosser:
@zendesk/samson 

some todos left over ... but let's get the base merged first :)


jonmoter:
I'm not reviewing a PR with 282 files that changed and the CI builds are failing.

grosser:
all fixed now ... let me know if you want chunked PRs instead to reduce the size/scope ...

sandlerr:
can we turn off `Style/IfUnlessModifier`?

config/initializers/app_routes.rb
```diff
@@ -2,6 +2,6 @@ module AppRoutes
   # memoize this due to a bug in MRI that shows up in multi-threaded code
   # see https://github.com/puma/puma/issues/647, search for "bug in MRI"
   def self.url_helpers
-    @@routes_url_helpers ||= Rails.application.routes.url_helpers
+    @routes_url_helpers ||= Rails.application.routes.url_helpers
```
sandlerr: any behaviour change here? looks to me like we can delete the entire file now?

plugins/env/app/models/environment_variable.rb
```diff
@@ -40,7 +41,7 @@ def matches?(ev, deploy_group)
     def resolve_dollar_variables(env)
       env.each_value do |value|
         value.gsub!(/\$\{(\w+)\}|\$(\w+)/) do |original|
-          env[$1 || $2] || original
+          env[Regexp.last_match(1) || Regexp.last_match(2)] || original
```
sandlerr: not sure this is an improvement

config/initializers/app_routes.rb
grosser: it is a class method, so `@` should be fine / a global cache, why would we want to delete the file ?

plugins/env/app/models/environment_variable.rb
grosser: yeah true ... did not see that ...

grosser:
do you have an example for `Style/IfUnlessModifier` ? / why you don't like it ?

grosser:
@sandlerr broke out the app changes here https://github.com/zendesk/samson/pull/1004 ... let's start with that :)

config/initializers/app_routes.rb
sandlerr: the Ruby bug seems to have been closed in 2.2

grosser:
PR 1000 is ready now @sandlerr ... but are you ready for PR 1000 ?

plugins/datadog/app/models/datadog_notification.rb
```diff
@@ -23,7 +24,7 @@ def deliver
     )
 
     client = Dogapi::Client.new(api_key, nil, "")
-    status, _ = client.emit_event(event)
+    status, = client.emit_event(event)
```
sandlerr: this is weird syntax, is it equivalent to `status = client.emit_event(event)[0]` ?

sandlerr:
this has merge conflicts & codeclimate errors now :/

plugins/datadog/app/models/datadog_notification.rb
grosser: yeah ... theoretically correct, but looks fishy :D

grosser:
rebased and fixed syntax

grosser:
codeclimate is 5 fixed, 4 new ones ... so overall net positive :)

sandlerr:
not sure how CC works, :+1: if this isn't going to make every subsequent PR red as well

grosser:
cc only shows warnings / we got a ton of these ... just trying to add less then I remoe :D

PR #1001, opened 2016-05-19
show/inspect status of all pods for a given release

grosser:
before it only logged the first pod per role

```
Production:
  app-server: Waiting (Pending, not Ready)
  app-server: Waiting (Pending, not Ready)
  worker: Waiting (Pending, not Ready)
```


jonmoter:
👍

PR #1003, opened 2016-05-19
dry up secrets

grosser:
best review each commit, first is only a bunch of indent

 - use same path logic for db and vault
 - avoid duplicating key ordering knowledge
 - make initial `new` view blank / do not pre-select anything
 - cleanup tests to have mocks per test and not global
 - move test helpers into kubernetes folder

screen shot 2016-05-19 at 1 51 29 pm

screen shot 2016-05-19 at 2 03 46 pm

### DIFF
--- a/plugins/kubernetes/test/test_helper.rb
+++ b/plugins/kubernetes/test/test_helper.rb
@@ -2,6 +2,61 @@ require_relative '../../../test/test_helper'
 require_relative '../lib/samson_kubernetes/hash_kuber_selector'
 require 'celluloid/test'
 
+# Mock up vault client
+class VaultClient
+  def logical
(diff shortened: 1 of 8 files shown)


plugins/kubernetes/test/test_helper.rb
```diff
@@ -2,6 +2,61 @@
 require_relative '../lib/samson_kubernetes/hash_kuber_selector'
 require 'celluloid/test'
 
+# Mock up vault client
```
grosser: moving this to the correct test_helper and getting rid of stub data

grosser:
@zendesk/paas

irwaters:
👍

PR #1004, opened 2016-05-19
rubocop fixes for the app

grosser:


app/controllers/builds_controller.rb
```diff
@@ -23,9 +23,7 @@ def create
     @build.creator = current_user
     @build.save
 
-    if @build.persisted? && params[:build_image].present?
-      start_docker_build
-    end
+    start_docker_build if @build.persisted? && params[:build_image].present?
```
sandlerr: this is an example of `Style/IfUnlessModifier`. When the condition is a bit long or complex I find it harder to parse in one line.

sandlerr:
:+1: on green

app/controllers/builds_controller.rb
grosser: k, reverted that line and added `Style/IfUnlessModifier` to the other PR

PR #1005, opened 2016-05-20
use simpler test client with expectations

grosser:
- avoiding the to/from http conversion
 - asserting that all calls were used

@zendesk/paas


irwaters:
👍 on green

PR #1007, opened 2016-05-20
more rubocop

grosser:
@sandlerr it never stops ...

### DIFF
--- a/config/initializers/app_routes.rb
+++ b/config/initializers/app_routes.rb
@@ -2,6 +2,6 @@ module AppRoutes
   # memoize this due to a bug in MRI that shows up in multi-threaded code
   # see https://github.com/puma/puma/issues/647, search for "bug in MRI"
   def self.url_helpers
-    @@routes_url_helpers ||= Rails.application.routes.url_helpers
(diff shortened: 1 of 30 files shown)


config/initializers/app_routes.rb
```diff
@@ -2,6 +2,6 @@ module AppRoutes
   # memoize this due to a bug in MRI that shows up in multi-threaded code
   # see https://github.com/puma/puma/issues/647, search for "bug in MRI"
```
sandlerr: the linked bug is: https://bugs.ruby-lang.org/issues/10871 seems closed in ruby 2.2

sandlerr:
what do you think of just excluding all old migrations from rubocop?

sandlerr:
:+1:

config/initializers/app_routes.rb
grosser: let's do a separate PR for that ...

grosser:
too late on migrations ... hard to exclude all old without all listing them 1-by-1 as exclude ... so I'd rather fix them up

PR #1008, opened 2016-05-20
document and auto-disable routes hack

grosser:
@sandlerr how about that ?
(don't feel like bumping ruby ... 2.3.1 is a bit too new ...)


sandlerr:
:+1:

didn't realise it was only 2.3.1, I thought it was fixed in 2.2

PR #1009, opened 2016-05-20
daemon sets cannot be updated, so we have to delete and create them

grosser:
@zendesk/paas 

could possibly get into a bad state when a deploy fails and then the daemonset is missing ... 
following the process kuberctl uses to delete a daemonset

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -54,15 +54,21 @@ module Kubernetes
       deploy_group.kubernetes_cluster.client
     end
 
-    def deploy_to_kubernetes
-      resource = case deploy_yaml.resource_name
-      when 'deployment' then Kubeclient::Deployment.new(deploy_yaml.to_hash)
-      when 'daemon_set' then Kubeclient::DaemonSet.new(deploy_yaml.to_hash)
-      else raise "Unknown resource #{deploy_yaml.resource_name}"
+    def deploy
+      case deploy_yaml.resource_name
+      when 'deployment'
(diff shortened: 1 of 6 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -54,15 +54,21 @@ def client
...
+      when 'deployment'
+        deploy = Kubeclient::Deployment.new(deploy_yaml.to_hash)
+        if resource_running?(deploy)
+          extension_client.update_deployment deploy
+        else
+          extension_client.create_deployment deploy
+        end
+      when 'daemon_set'
+        daemon = Kubeclient::DaemonSet.new(deploy_yaml.to_hash)
+        delete_daemon_set(daemon)
```
jonmoter: `delete_daemon_set(daemon) unless daemon_set_exists?(daemon)`

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -105,6 +111,33 @@ def resource_running?(resource)
...
+      loop do
+        sleep 2
+        current = extension_client.get_daemon_set(*daemon_set_selector)
+        break if current.status.currentNumberScheduled == 0 && current.status.numberMisscheduled == 0
+      end
+
+      # delete it
+      extension_client.delete_daemon_set *daemon_set_selector
+    rescue KubeException
+      false # daemon set was already deleted
```
jonmoter: Remove rescue if you make the change above.

jonmoter:
👍  with 2 suggestions above

PR #1013, opened 2016-05-24
fix tests not working when gpg signing is globally enabled

grosser:
@zendesk/paas 
```
gpg: skipped "Test User <test@example.com>": secret key not available
gpg: signing failed: secret key not available
error: gpg failed to sign the data
fatal: failed to write commit object
```


irwaters:
👍

PR #1014, opened 2016-05-24
remove unused role columns, they make the new UI break

grosser:
@zendesk/paas

we no longer use these columns because we read from deploy_group_role ... and not having them set made the `role new` UI break

## Risks:
 - low: angular releases UI will deploy 1/1/50 instead of role defaults


jonmoter:
:+1:

PR #1017, opened 2016-05-24
try adding hakiri security scanner

grosser:
@zendesk/samson @cericksen 

just a try ... did not find anything so far ... not sure how good it is ...


grosser:
looks like it's just a fancy wrapper around brakeman ... removing for now ...

PR #1018, opened 2016-05-24
reuse parsing logic for secret resolution

grosser:
@zendesk/paas 

making it obvious that we are missing environment verification + drying parsing logic


irwaters:
👍  lgtm

PR #1019, opened 2016-05-25
Irwaters/paas 147

irwaters:
This adds dynamic insertion of the secret-sidecar and appropriate volumes when the feature is enabled

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low (needs to go do staging so we can wire it up to the staging k8s cluster and staging vault)


plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,46 @@ def template
...
+      secret_sidecar = {
+        image: ENV.fetch("SECRET_SIDECAR_IMAGE").to_s,
+        name: "secret-sidecar",
+        volumeMounts: [
+          secret_vol,
+          { mountPath: "/vault-auth", name: "vaultauth" },
+          { mountPath: "/secretkeys", name: "secretkeys" }
+        ]
+      }
+      #TODO, do we want resource limits on this, or just give it what it needs?
```
grosser: should need next to nothing ? ... maybe deploy on it's own and see what the minimum is ... 100m seems like too much for a simple script ... 
ideally we'd boot it and then shut it down since it will not be needed again ... maybe even better if we can make it a daemonset so we don't need 1 per pod ... but having something that actually works is a good first step :)

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,46 @@ def template
...
+      secret_sidecar[:resources] =  { limits: { cpu: 0.1, memory: "100Mi" } }
+
+      # also inject the secrets FS into the primary container so that the
+      # secrets can be shared
+      containers = template.spec.template.spec.containers.dup
```
grosser: that seems dangerous ... why do we have to dup ? ... prefer a clean container definition ?

grosser: ah nvm ... it's just adding to the copy  ....

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,46 @@ def template
...
+      containers.first.volumeMounts = [secret_vol]
```
grosser: preferably only add to the mounts since it might have other mounts ?

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,46 @@ def template
...
+      containers << secret_sidecar
+
+      template.spec.template.spec.containers = containers
+
+      #lastly, define the volumes in the pod
```
grosser: prefer space in front of comments for readability `# lastly`

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,46 @@ def template
...
+      if template.spec.template.spec[:volumes].nil?
```
grosser: use `.` access consistently ? `if template.spec.template.spec.volumes`

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,46 @@ def template
...
+        template.spec.template.spec.volumes = pod_volumes
+      else
+        pod_volumes.each { |pv| template.spec.template.spec.volumes << pv }
```
grosser:
```Ruby
template.spec.template.spec.volumes ||= []
template.spec.template.spec.volumes.concat pod_volumes
```

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -126,6 +167,20 @@ def set_env
        }
       end
 
+      if ENV.fetch("SECRET_SIDECAR_IMAGE", false)
```
grosser: don't need a default ? `if ENV["SECRET_SIDECAR_IMAGE"]`

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -142,5 +197,12 @@ def container
         containers.first
       end
     end
+
+    def sidecar_container
+      @sidecar ||= begin
+        containers = template.spec.template.try(:spec).try(:containers) || []
```
grosser: can a template have no spec ? ... maybe just bad test stubs ?

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -142,5 +197,12 @@ def container
...
+        containers.map { |possible_container| return possible_container if possible_container.name == 'secret-sidecar' }
```
grosser: prefer `each` since nothing is returned

grosser: also if nothing is found this would return the map ... missing a testcase ?

```Ruby
containers.detect { |possible_container| break possible_container if possible_container.name == 'secret-sidecar' }
```

plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
```diff
@@ -131,6 +131,38 @@
...
+        end
+        e.message.must_include "has 0 containers, having 1 section is valid"
+      end
+
+      it "adds to existing volume definitions" do
+        doc.raw_template.gsub!("containers:\n      - {}\n", "containers:\n      - {}\n      volumes:\n      - {}\n      - {}\n")
+        yaml.to_hash[:spec][:template][:spec][:volumes].count.must_be(:>=, 2)
+      end
+
+      after do
```
grosser: prefer to keep before and after together so it's easy to check symmetry ... even better to use around + with_env

```Ruby
around { |t| with_env(''VAULT" => 'bla', "BAR" => 'foo', &test) }
```

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
irwaters: I don't think a template w/o a spec is valid.

grosser: maybe just blow up then ... but I guess either way is fine ... will just
blow up somewhere else then ...

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,55 @@ def template
...
+        ]
+      }
+
+      # also inject the secrets FS into the primary container so that the
+      # secrets can be shared
+      containers = template.spec.template.spec.containers.dup
+      if containers.first.volumeMounts.nil?
+        containers.first.volumeMounts = [secret_vol]
+      else
+        containers.first.volumeMounts ||= []
```
grosser: can the if/else go away now since the code in the else should work for both ?

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,55 @@ def template
...
+        containers.first.volumeMounts << secret_vol
+      end
+      containers << secret_sidecar
+      template.spec.template.spec.containers = containers
```
grosser: would injecting directly into the containers not work / avoid the duping ?

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -46,6 +47,55 @@ def template
...
+
+      # lastly, define the volumes in the pod
+      if template.spec.template.spec.volumes.nil?
+        template.spec.template.spec.volumes = pod_volumes
+      else
+        template.spec.template.spec.volumes ||= []
```
grosser: same here, the else code works for both ?

plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
```diff
@@ -131,6 +131,45 @@
       end
     end
 
+    describe "secret-sidecar-containers" do
+      let(:container) { result.fetch(:spec).fetch(:template).fetch(:spec).fetch(:containers).first }
+      before do
```
grosser: around + with_env ?

plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
```diff
@@ -131,6 +131,45 @@
...
+      it "creates a sidecar" do
+        yaml.to_hash[:spec][:template][:spec][:containers].last[:name].must_equal('secret-sidecar')
+      end
+
+      it "fails if there is no primary container" do
+        assert doc.raw_template.sub!("      containers:\n      - {}", '')
+        e = assert_raises Samson::Hooks::UserError do
+          yaml.to_hash
+        end
+        e.message.must_include "has 0 containers, having 1 section is valid"
```
grosser: that is re-testing different code ... not necessary ?

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
grosser: code above already raises when there are no containers -> no necessary ?

grosser:
only style nitpicks left, so 👍 on fixing them ... please squash this time ;)

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -142,5 +198,13 @@ def container
         containers.first
       end
     end
+
+    def sidecar_container
+      @sidecar ||= begin
+        template.spec.template.spec.containers.each do |possible_container|
+          return possible_container if possible_container.name == 'secret-sidecar'
```
grosser: returning means it will not get cached -> detect + break

plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
```diff
@@ -131,6 +131,36 @@
       end
     end
 
+    describe "secret-sidecar-containers" do
+      before do
```
grosser: around and with_env ?

PR #1020, opened 2016-05-25
resolve global environment/deploy-group secrets

grosser:
@zendesk/paas 

next step is turning it just into keys, but for now allow global for environment+project+deploy_group

 - no longer clears the form when save fails
 - no longer allows empty value for key
 - no longer only able to pick a single deploy group
 - uses js for new and edit so we verify sure it always works / do not duplicate logic

screen shot 2016-05-25 at 1 30 31 pm
screen shot 2016-05-25 at 1 33 45 pm
screen shot 2016-05-25 at 1 30 40 pm


irwaters:
👍 on green

PR #1022, opened 2016-05-26
lookup secrets by key

grosser:
@zendesk/paas 

Problem: 
secret keys need to be customized for each deploy group/project/environment which is tedious and blocks sharing.

Solution:
Instead of hardcoding production/foo-project/zone1/my-key everywhere and having to edit stages,
only use `secret:my-key` and the backend will figure out what you wanted.

Downsides:
worst case this will mean 2**3=8 reads ... could optimize that for db backend by doing a read_multi callback

### TODO:
 - [ ] do not use raise/rescue as way of finding out if a key exists
 - [ ] use read_multi
 - [ ] figure out how to support this in secret sidecar
 - [ ] make admin UI show key components separately
 - [ ] update admin UI secret:// instructions

### DIFF
--- a/app/models/terminal_executor.rb
+++ b/app/models/terminal_executor.rb
@@ -39,21 +39,54 @@ class TerminalExecutor
 
   private
 
-  # TODO: pick dynamic secret from just the key
   def resolve_secrets(command)
-    deploy_groups = @deploy.try(:stage).try(:deploy_groups) || []
-    allowed = {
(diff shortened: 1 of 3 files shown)


irwaters:
So I think this is a reasonable plan.  The secret puller script will need to implement some of the same logic if it's only to be given the key, and we'll need to make sure that the rest of the vars necessary to construct the keyspace get injected into the sidecar for it to use.

I'd prefer to see the existing stuff work before we start to change it :)

PR #1023, opened 2016-05-26
SAMSON-247 refactor next_stage tests and make sure they work with soft-deleted s…

grosser:
@irwaters 

https://zendesk.atlassian.net/browse/SAMSON-247


irwaters:
👍

PR #1024, opened 2016-05-27
swap includes for joins in several places on the Deploy model

sandlerr:
### Description

In our private plugin we have call `Deploy.successful.where(...)`, which pulls back the longtext `jobs.output`. This is having an extreme effect on query performance causing requests to timeout.

The `where` syntax seems to be supported for `joins` as it is used in the `self.expired` method.

/cc @zendesk/samson @zendesk/vegemite @seancaffery 

### Risks
* [low] performance problems. Some of the patched methods are unused.

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -105,19 +105,19 @@ class Deploy < ActiveRecord::Base
   end
 
   def self.pending
-    includes(:job).where(jobs: { status: 'pending' })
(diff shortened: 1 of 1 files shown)


seancaffery:
👍  on 🍏

grosser:
👍

ocke:
👍

PR #1025, opened 2016-05-27
fix login failing with no implicit conversion of Fixnum into String

grosser:
@jonmoter 

... unable to reproduce locally ... does not make any sense … other actions that trigger Papertrail work fine too ...


jonmoter:
😕 
👍

grosser:
did not help :(

PR #1029, opened 2016-05-30
better debugging for webhooks

jonmoter:
Merge to master on Volunteer app wasn't triggering releases, but I have no idea why. Trying to add some better logging to hunt down what's up.

Still WIP.

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -1,7 +1,7 @@
 class Integrations::BaseController < ApplicationController
   skip_around_action :login_user
   skip_before_action :verify_authenticity_token
-  before_action :validate_webhook
+  before_action :validate_request
 
   def create
     return head(:ok) unless deploy?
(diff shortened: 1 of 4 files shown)


app/controllers/integrations/base_controller.rb
```diff
@@ -22,6 +23,20 @@ def create
...
   protected
 
+  # These methods can/must be overridden by subclasses
+
+  def validate_webhook
+    true # can be overridden in subclasses
+  end
+
+  def commit
+    raise NotImplementedError.new("#commit must be overridden in a subclass")
```
grosser: prefer `raise Xyz, message`

app/controllers/integrations/base_controller.rb
```diff
@@ -22,6 +23,20 @@ def create
 
   protected
 
+  # These methods can/must be overridden by subclasses
+
+  def validate_webhook
```
grosser: would `validate_request` be better ?

app/controllers/integrations/github_controller.rb
```diff
@@ -10,8 +10,12 @@ class Integrations::GithubController < Integrations::BaseController
 
   protected
 
+  def validate_webhook
+    head(:unauthorized) unless valid_signature?
```
grosser: maybe something more descriptive like `text: 'Invalid signature'`

jonmoter:
@grosser now ready to merge

grosser:
👍 so the problem was missing logs ?
... if anything else is fixed now can we add a test ?

jonmoter:
No, this didn't actually fix the problem, just the first bit of refactoring to add debugging and try to figure out what went wrong. May add another log message or two before merging.

PR #1030, opened 2016-05-31
comments for secret resolver

grosser:
@jonmoter 

does that help ?
trying to not change the code to avoid conflicts with https://github.com/zendesk/samson/pull/1022


jonmoter:
Eh.  Let's hold off on this for now, and I'll take a look at the Secret code to figure out what would clarify it best.

PR #1032, opened 2016-05-31
make gitlab auth optional like ldap

grosser:
@zendesk/paas 

/fyi @shouta-dev


jonmoter:
any reason we don't make all of them "enable"?

grosser:
for now I want to get back to the previous state / not confuse everyone with a gitlab button ... 
for next PR I was thinking no auth should be enabled and auth should show a "Hey pick an auth method" ... since having an auth method showing that does not work ... or potentially works by accident would be bad

jonmoter:
👍

PR #1034, opened 2016-05-31
do not show not working auth methods

grosser:
@zendesk/samson 

Problem:
 - we show auth methods that do not work, leading to potential confusion
 - every new auth method will be a second-class citizen since we cannot show it by default

Solution:
 - no auth method is shown by default, everything must be enabled
 - bootstrap still still sets up the example github auth

screen shot 2016-05-31 at 10 51 06 am

# Risks
 - will disable existing samson login screens that relied on the default setup


jonmoter:
👍

jonmoter:
What's the ops change that's needed for this?  Set AUTH_GOOGLE=1?

grosser:
ops change is in the works, waiting on merging that first

grosser:
made `true` also mean enabled since 1 is a weird way of saying `enabled`

grosser:
all rolled out, should be good to go :)

PR #1035, opened 2016-05-31
fix gitlab being enabled when set to 0 ...

grosser:
quickfix for https://github.com/zendesk/samson/pull/1032 ... since https://github.com/zendesk/samson/pull/1034 is a bit bigger / needs some ops work before it can be merged ...

@jonmoter


jonmoter:
👍

PR #1036, opened 2016-05-31
Irwaters/paas 139 (make samson write to multiple vault servers)

irwaters:
Adds ability for samson to write to multiple vault servers.  Also adds a config file in case people wana use it for different ENVs.  Also ads retry logic around deletes and writes (since we are talking about multiple servers now).

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-139

### Risks
- Level: Low


config/initializers/vault.rb
```diff
@@ -1,11 +1,10 @@
-VAULT_ENABLED = ENV.fetch('SECRET_STORAGE_BACKEND', false)
-if VAULT_ENABLED == 'SecretStorage::HashicorpVault'
+VAULT_CONFIG = Rails.application.config_for(:vault).symbolize_keys.freeze
+if ENV.fetch("SECRET_STORAGE_BACKEND", false) == "SecretStorage::HashicorpVault"
```
grosser: no need for fetch

config/initializers/vault.rb
```diff
@@ -1,11 +1,10 @@
...
   require 'vault'
   Rails.logger.info("Vault Client enabled")
   Vault.configure do |config|
-    config.ssl_pem_file = ENV.fetch("VAULT_SSL_CERT")
-    config.ssl_verify = ActiveRecord::Type::Boolean.new.type_cast_from_user(ENV.fetch("VAULT_SSL_VERIFY", true))
-    config.address = ENV.fetch("VAULT_ADDR", 'https://127.0.0.1:8200')
+    config.ssl_pem_file = Rails.root.join(VAULT_CONFIG[:pem_path])
```
grosser: is this required ? -> use fetch

config/initializers/vault.rb
```diff
@@ -37,7 +36,7 @@ def initialize
       # if we are testing, just return here.  We'll let super configure
       # the rest of the client
       return if Rails.env.test?
-      uri = URI.parse(Vault.address)
+      uri = URI.parse(get_host)
```
grosser: no get_ / set_ methods -> `def host`

config/initializers/vault.rb
```diff
@@ -46,5 +45,38 @@ def initialize
         raise "Failed to get auth token from vault server"
       end
     end
+
+    def read(key)
+      Vault.address = get_host
```
grosser: - check that this is threadsafe (uses Thread.current in the gem) or we this needs to live in Thread.exclusive
 - do not change global config and then leave it like that

```Ruby
def with_host
  old = Vault.address
  Vault.address = host
  yield
ensure
  Vault.address = old
end
```

config/initializers/vault.rb
```diff
@@ -46,5 +45,38 @@ def initialize
...
+    end
+
+    def list(path)
+      Vault.address = get_host
+      Vault.logical.list(path)
+    end
+
+    # make darn sure on deletes and writes that we try a couple of times.
+    def write(key, data)
+      VAULT_CONFIG[:hosts].split(',').each do |vault_server|
```
grosser: .fetch since this will blow up on nil

config/initializers/vault.rb
```diff
@@ -46,5 +45,38 @@ def initialize
...
+      Vault.address = get_host
+      Vault.logical.read(key)
+    end
+
+    def list(path)
+      Vault.address = get_host
+      Vault.logical.list(path)
+    end
+
+    # make darn sure on deletes and writes that we try a couple of times.
```
grosser: `Retry to ensure writes get persisted / servers stay in an consistent state`

config/initializers/vault.rb
grosser: prefer `.split(/, ?/)` so users can use spaces `foobar, barfoo`

config/initializers/vault.rb
```diff
@@ -46,5 +45,38 @@ def initialize
...
+    def delete(key)
+      VAULT_CONFIG[:hosts].split(',').each do |vault_server|
+        Vault.address = vault_server
+        Vault.with_retries(Vault::HTTPConnectionError, attempts: 5) do
+          Vault.logical.delete(key)
+        end
+      end
+    end
+
+    def get_host
```
grosser: private ?

config/initializers/vault.rb
```diff
@@ -46,5 +45,38 @@ def initialize
...
+      VAULT_CONFIG[:hosts].split(',').first
```
grosser: copy-pasted ... make a method for `hosts`

config/initializers/vault.rb
```diff
@@ -46,5 +45,42 @@ def initialize
         raise "Failed to get auth token from vault server"
       end
     end
+
+    def read(key)
+      Vault.address = vault_host
```
grosser: thread safety / with_address ?

config/vault.yml
```diff
@@ -0,0 +1,13 @@
+development: &development
```
grosser: this seems counter-productive ... we try to configure everything via env vars, so why add another yml file ?

config/initializers/vault.rb
```diff
@@ -1,50 +1,72 @@
...
-      if response.code == "200"
-        @token = JSON.parse(response.body).fetch("auth")["client_token"]
-      else
-        raise "Failed to get auth token from vault server"
+      @writers = []
+      # since we have a bunch of servers, don't use the client singlton to
+      # talk to them
+      vault_hosts.each_with_index do |vault_server, idx|
+        @writers[idx] = Vault::Client.new(DEFAULT_CLIENT_OPTIONS)
+        @writers[idx].address = vault_server
```
grosser: using idx makes it more complicated:

```Ruby
@writers = vault_hosts.map do |host|
  writer = Vault::Client.new(DEFAULT_CLIENT_OPTIONS)
  writer.address = host
  ....
  writer
end
```

config/initializers/vault.rb
```diff
@@ -1,50 +1,72 @@
...
+
+        uri = URI.parse(vault_server)
+        @http = Net::HTTP.start(uri.host, uri.port, DEFAULT_CLIENT_OPTIONS)
```
grosser: this seems counter intuitive ... call some method on the writer instead of re-implementing connection logic ?

config/initializers/vault.rb
```diff
@@ -1,50 +1,72 @@
...
+
+    def list(path)
+      @reader.logical.list(path)
+    end
+
+    # make darn sure on deletes and writes that we try a couple of times.
+    # not going to catch any other excpeptions as we want this to blow up
+    # if anything fails
+    def write(key, data)
+      @writers.each do |vault_server|
```
grosser: nice!

config/initializers/vault.rb
```diff
@@ -1,50 +1,72 @@
...
+
+    def delete(key)
+      @writers.each do |vault_server|
+        Vault.with_retries(Vault::HTTPConnectionError, attempts: 5) do
+          vault_server.logical.delete(key)
+        end
+      end
+    end
+
+    def vault_hosts
```
grosser: private ?

grosser:
👍

PR #1038, opened 2016-05-31
PAAS-186 do not show projects that we would never select / do not have roles

grosser:
@zendesk/paas

### DIFF
--- a/plugins/kubernetes/app/views/admin/kubernetes/deploy_group_roles/_form.html.erb
+++ b/plugins/kubernetes/app/views/admin/kubernetes/deploy_group_roles/_form.html.erb
@@ -12,7 +12,8 @@
     <div class="form-group">
       <%= form.label :project_id, 'Project', class: "col-lg-2 control-label" %>
       <div class="col-lg-2">
-        <%= form.select :project_id, current_user.administrated_projects.pluck(:name, :id), {}, class: "form-control", disabled: @deploy_group_role.persisted? %>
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/app/views/admin/kubernetes/deploy_group_roles/_form.html.erb
```diff
@@ -12,7 +12,8 @@
     <div class="form-group">
       <%= form.label :project_id, 'Project', class: "col-lg-2 control-label" %>
       <div class="col-lg-2">
-        <%= form.select :project_id, current_user.administrated_projects.pluck(:name, :id), {}, class: "form-control", disabled: @deploy_group_role.persisted? %>
+        <% projects = current_user.administrated_projects.where(id: Kubernetes::Role.pluck(:project_id)).pluck(:name, :id) %>
```
jonmoter: How does it work to do a class-level `pluck`?

grosser: class level pluck means all projects that have any roles configured ... this endpoint is not scoped by project, so needs to be global

jonmoter: How about creating a scope or class-level method on Project, so you can say `Project.with_kubernetes_roles` or something like that?

grosser: done

jonmoter:
👍 on 💚

PR #1039, opened 2016-06-01
request regular logs when previous logs are not available

grosser:
@zendesk/paas 

first commit is cleanup, second is the change

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -87,29 +87,37 @@ module Kubernetes
     def show_failure_cause(release)
       bad_pods(release).each do |pod, client, deploy_group|
         @output.puts "\n#{deploy_group.name} pod #{pod.name}:"
-        namespace = deploy_group.kubernetes_namespace
-        print_events(client, namespace, pod)
+        print_events(client, pod)
         @output.puts
-        print_logs(client, namespace, pod)
+        print_logs(client, pod)
       end
     end
 
     # logs - container fails to boot
-    def print_logs(client, namespace, pod)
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -85,29 +85,45 @@ def fetch_pods(client, query)
...
-      bad_pods = release.clients.flat_map do |client, query, deploy_group|
-        bad_pods = fetch_pods(client, query).select { |p| p.restarted? || !p.live? }
-        bad_pods.map { |p| [p, client, deploy_group] }
-      end
-
-      bad_pods.each do |pod, client, deploy_group|
-        namespace = deploy_group.kubernetes_namespace
+      bad_pods(release).each do |pod, client, deploy_group|
         @output.puts "\n#{deploy_group.name} pod #{pod.name}:"
+        namespace = deploy_group.kubernetes_namespace
```
jonmoter: FYI, if you're getting the pod back from Kubernetes, the namespace is in the pod itself.  `pod.metadata.namespace`, i believe.

jonmoter: Might want to add that as a helper method to `Kubernetes::Api::Pod`

grosser:
I think this is wrong anyways, the issue was that the pod had multiple containers :/

grosser:
now handles multiple containers as well + added pod#namespace

jonmoter:
`DeployExecutor` is starting to get kindof confusing to follow, and is probably due for a refactor soon. But we're still iterating on this, so I'm fine with holding off on that 'til we get the functionality more-or-less right.

👍 on 💚

PR #1040, opened 2016-06-01
got broken along the way

irwaters:
injection of secrets volume into primary docker container got busted.

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


grosser:
so it was missing a test :trollface:

grosser:
👍 with test that fails without that patch

PR #1041, opened 2016-06-02
do not blow up when connection to github fails

grosser:
@zendesk/paas

just saw this blow up while on the train ... not sure why it is not an Octokit error, but it should not blow up :)


jonmoter:
Does this not throw an Airbrake error? I don't see it in Airbrake.

grosser:
might never happen in production ... just on laptop in trains

jonmoter:
Ah, when running locally.  Got it.

👍

PR #1042, opened 2016-06-02
do not show kubernetes roles when not using kubernetes on a stage

grosser:
@zendesk/paas

atm we show missing roles etc UI that is not useful / confusing when the stage is not using kubernetes ... so let's hide it

before: showing kubernetes and commands
screen shot 2016-06-02 at 8 58 00 am


jonmoter:
👍

Issue #1043, opened 2016-06-02
Samson deletes docker image just built


rata:
Hi,

In app/models/docker_builder_service.rb there is a ```build.docker_image.remove(force: true)``` that is done by default. This makes our docker builds take longer, as there is always no cache but the base image.

Although it seems really on purpose. Can I ask why do you do that? To keep the disk clean on the server? And do your builds always never use the cache, then? Or do you do some other trick to reuse the cache when building images?

We would like to use the cache and clean the disk somehow (probably running some other command to clean daily/weekly/something). Do you think it's an option to not do that if some env var is defined or something?

Thanks!
Rodrigo

jonmoter:
@henders added that logic, so cc'ing him.

I believe the issue that our docker hosts were filling up 'cause we never cleaned up the images.  I get your point that it's nice to have some cached images on disk, especially if you're using the same ones repeatedly.  I've looked into https://github.com/spotify/docker-gc briefly, but haven't tried running it before.

We could add an environment variable that makes it optional whether to delete the images after pushing them. Would that work for you?

rata:
@jonmoter haha, I was looking at the same thing from spotify too! :-D

Yes, that will work I think. Do you want me to do the patch? Or will you do it? :)

jonmoter:
You're welcome to create a PR that adds an ENV var.

rata:
This is already fixed, closing

rata:
@jonmoter @grosser we are using spotify/docker-gc to cleanup the images (run the builder and docker-gc in different containers in a pod, in the docker-gc container I set the variable "DOCKER_HOST" to "tcp://localhost:2375").
I configured it yesterday and seems to be working fine for now. I did some changes (the docker image for docker-gc doesn't build in the spotify repo), you can see them in rata/docker-gc

PR #1044, opened 2016-06-02
Add env var to keep images built with docker

rata:
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

Tested and seems to work fine here with the env var defined to 1


rata:
Closes: #1043

grosser:
.env.example docs for this setting and some docs on why/when this will be useful ?

rata:
@grosser Oh, sorry! How about now? :)

grosser:
❤️

PR #1048, opened 2016-06-03
adding scopes for secrets based

irwaters:
this addes scopes for secrets that are defined in a template

example:
      annotations:
        secret/TEST: 'staging/example-project/local-kubernets/boo'
        secret/ENVTEST: '$ENV/truth_service/$DEPLOY_GROUP/my_key'

the secret ENVTEST will be changed for each stage of the deployment with the correct env and deploy group name.  this allows for the use of a single template across multiple deploy groups/envs.

Also validates that all secrets found in annotations exist before letting the deployment happen.

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
https://zendesk.atlassian.net/browse/PAAS-188

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
@@ -16,7 +16,11 @@ module Kubernetes
         set_spec_template_metadata
         set_docker_image
         set_resource_usage
-        set_secret_sidecar if ENV.fetch("SECRET_SIDECAR_IMAGE", false)
+        if ENV["SECRET_SIDECAR_IMAGE"]
+          set_secret_sidecar
+          expand_secret_annotations
+          verify_secret_annotations
+        end
         set_env
 
         hash = template.to_hash
@@ -45,6 +49,40 @@ module Kubernetes
           )
         end
       end
+    end
+
+    # expand $ENV and $DEPLOY_GROUP in annotation that start with 'secret/'
+    def expand_secret_annotations
+      annotations.to_h.each do |annotation_name, secret_key|
+        if annotation_name.to_s.start_with?(SecretStorage::VAULT_SECRET_BACKEND)
+          secret_key.gsub!(/\${ENV}/, @doc.deploy_group.environment.permalink)
+          secret_key.gsub!(/\${DEPLOY_GROUP}/, @doc.deploy_group.permalink)
+        end
+      end
+    end
+
+    # verify that each secret really exists and inform the user
+    # if it does not as the deployment will fail
+    def verify_secret_annotations
+      errors = []
+      annotations.to_h.each do |annotation_name, secret_key|
+        next unless annotation_name.to_s.start_with?(SecretStorage::VAULT_SECRET_BACKEND)
+        begin
+          SecretStorage.read(secret_key)
+        rescue ActiveRecord::RecordNotFound, NoMethodError
+          errors << "Secret #{annotation_name} with key #{secret_key} could not be found."
+        end
+      end
+      if errors.any?
+        raise(
+          Samson::Hooks::UserError,
+          "Missing Secret Keys:\n\t#{errors.join("\n\t")}"
+        )
+      end
+    end
+
+    def annotations
+      @template.spec.template.metadata.annotations
     end
 
     # Sets up the secret_sidecar and the various mounts that are required
--- a/plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
@@ -161,6 +161,37 @@ describe Kubernetes::DeployYaml do
           "containers:\n      - :name: foo\n        :volumeMounts:\n")
         yaml.to_hash[:spec][:template][:spec][:containers].first[:volumeMounts].count.must_equal 1
       end
+
+      it "fails to find a secret needed by the sidecar" do
+        old_metadata = "role: app-server\n    "
+        new_metadata = "role: app-server\n      annotations:\n        secret/FOO: production/foo/snafu/bar\n    "
+        doc.raw_template.gsub!(old_metadata, new_metadata)
+        e = assert_raises Samson::Hooks::UserError do
+          yaml.to_hash
+        end
+        e.message.must_include "secret/FOO with key production/foo/snafu/bar could not be found"
+      end
+
+      describe "scopes" do
+        before do
+          SecretStorage.stubs(:read).returns(true)
+        end
+
+        it "secret is scoped correctly" do
+          old_metadata = "role: app-server\n    "
+          new_metadata = "role: app-server\n      annotations:\n        "
+          new_metadata += "secret/FOO: ${ENV}/foo/${DEPLOY_GROUP}/bar\n    "
+          doc.raw_template.gsub!(old_metadata, new_metadata)
+          yaml.to_hash[:spec][:template][:metadata][:annotations][:'secret/FOO'].must_equal "production/foo/pod1/bar"
+        end
+
+        it "does not effect a non secret annotation" do
+          old_metadata = "role: app-server\n    "
+          new_metadata = "role: app-server\n      annotations:\n        annotation_key: somevalueforthekey\n    "
+          doc.raw_template.gsub!(old_metadata, new_metadata)
+          yaml.to_hash[:spec][:template][:metadata][:annotations][:annotation_key].must_equal "somevalueforthekey"
+        end
+      end
     end
 
     describe "daemon_set" do
--- a/app/models/secret_storage.rb
+++ b/app/models/secret_storage.rb
@@ -1,6 +1,7 @@
 module SecretStorage
   SECRET_KEYS_PARTS = [:environment_permalink, :project_permalink, :deploy_group_permalink, :key].freeze
   SEPARATOR = "/".freeze
+  VAULT_SECRET_BACKEND = 'secret/'.freeze
 
   require 'attr_encrypted'
   class DbBackend
@@ -53,7 +54,6 @@ module SecretStorage
   end
 
   class HashicorpVault
-    VAULT_SECRET_BACKEND = 'secret/'.freeze
     # we don't really want other directories in the key,
     # and there may be other chars that we find we don't like
     ENCODINGS = {"/": "%2F"}.freeze


grosser:
this means we have to copy the same secret to each deploy group ... can't we do the other way around where we set the key only and then the backend finds it like https://github.com/zendesk/samson/pull/1022 did ?

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -16,7 +16,10 @@ def to_hash
         set_spec_template_metadata
         set_docker_image
         set_resource_usage
-        set_secret_sidecar if ENV.fetch("SECRET_SIDECAR_IMAGE", false)
+        if ENV.fetch("SECRET_SIDECAR_IMAGE", false)
```
grosser: FYI simpler with [] instead of fetch

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -16,7 +16,10 @@ def to_hash
...
+          set_secret_sidecar
+          set_annotations
```
grosser: `expand_and_verify_secret_annotations`

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
       end
     end
 
+    # process the annotations and do some string substition
```
grosser: `expand $ENV and $DEPLOY_GROUP in annotations that start with '/secret/' and verify the secret exists before trying to deploy`

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
grosser: splitting this into 2 methods might actually make it simpler to digest ...

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
...
+    # on the keys we find in there.  this allows a single template
+    # to be used for deployments to multiple ENVs and deploy groups
+    # also ask the secret service if the key is really there, if it's not
+    # then inform the user and bail out
+    def set_annotations
+      deploy_group = DeployGroup.find(@release_doc_metadata[:deploy_group_id])
```
grosser: I'd prefer the more direct `@doc.deploy_group`

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
...
+      environment_permalink = deploy_group.environment.permalink
+      deploy_group_permalink = deploy_group.permalink
+      annotations = @template.spec.template.metadata.annotations
+      annotations.to_h.map do |annotation|
+        if annotation.first =~ /secret\//
+          annotation.last.gsub!(/\$ENV/, environment_permalink) if annotation.last =~ /\$ENV/
```
grosser: abbreviating ENVIRONMENT to ENV but not DEPLOY_GROUP is a bit asymmetric ... but I guess ENV is pretty common ...

prefer closed sub targets like ${ENV} so we never do $ENVIRONMENT -> productionIRONMENT etc foobar :)

grosser: doing the matching twice (and repeating the regex) does not help ... `annotation.last.gsub!(/\$ENV/, environment_permalink)` is good enough :)

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
...
+    # to be used for deployments to multiple ENVs and deploy groups
+    # also ask the secret service if the key is really there, if it's not
+    # then inform the user and bail out
+    def set_annotations
+      deploy_group = DeployGroup.find(@release_doc_metadata[:deploy_group_id])
+      environment_permalink = deploy_group.environment.permalink
+      deploy_group_permalink = deploy_group.permalink
+      annotations = @template.spec.template.metadata.annotations
+      annotations.to_h.map do |annotation|
+        if annotation.first =~ /secret\//
```
grosser: .start_with?('secret/')

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
...
+          annotation.last.gsub!(/\$ENV/, environment_permalink) if annotation.last =~ /\$ENV/
+          annotation.last.gsub!(/\$DEPLOY_GROUP/, deploy_group_permalink) if annotation.last =~ /\$DEPLOY_GROUP/
```
grosser: no real need to use a local variable, the lookup is cheap, also makes env and deploy_group code more symmetric

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
...
+        end
+        begin
+          SecretStorage.read(annotation.last)
+        rescue
```
grosser: rescue a specific error ? `ActiveRecord::RecordNotFound` ?

grosser: also prefer not using exceptions to control flow, it is slow / noisy

...

ideally resolve all secrets and then do a read_multi to get all secrets in 1 go AND tell the user that `these 5 secrets are missing` instead of bam this 1 is missing ... bam this other one is missing ... bam ... etc

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
...
+          raise "Secret #{annotation.first} with key #{annotation.last} could not be found.
```
grosser: regular errors go to airbrake -> `raise Samson::Hooks::UserErrror`

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
...
+            Deploy cannot cannot continue"
```
grosser: `Deploy cannot cannot continue` is kind of obvious / we could say that after each error ... so I'd rather not say that here ... the deploy should take care of that for us by stopping/showing that it failed nicely

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +50,31 @@ def template
...
+    # on the keys we find in there.  this allows a single template
+    # to be used for deployments to multiple ENVs and deploy groups
+    # also ask the secret service if the key is really there, if it's not
+    # then inform the user and bail out
+    def set_annotations
+      deploy_group = DeployGroup.find(@release_doc_metadata[:deploy_group_id])
+      environment_permalink = deploy_group.environment.permalink
+      deploy_group_permalink = deploy_group.permalink
+      annotations = @template.spec.template.metadata.annotations
+      annotations.to_h.map do |annotation|
```
grosser: why do we use map when we do not use the return value ?

grosser: also expand for better readability:

```
annotations.to_h.each do |env_name, key|
```

plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
```diff
@@ -161,6 +161,28 @@
...
+        before do
+          SecretStorage.stubs(:read).returns(true)
+        end
+        it "secret is scoped correctly" do
+          old_metadata = "role: app-server\n    "
+          new_metadata = "role: app-server\n      annotations:\n        secret/FOO: $ENV/foo/$DEPLOY_GROUP/bar\n    "
+          doc.raw_template.gsub!(old_metadata, new_metadata)
+          yaml.to_hash[:spec][:template][:metadata][:annotations][:'secret/FOO'].must_equal "production/foo/pod1/bar"
+        end
+      end
```
grosser: missing test for `it "does not replace or test non-secret annotations"`

plugins/kubernetes/test/models/kubernetes/deploy_yaml_test.rb
```diff
@@ -161,6 +161,28 @@
...
+        doc.raw_template.gsub!(old_metadata, new_metadata)
+        e = assert_raises RuntimeError do
+          yaml.to_hash
+        end
+        e.message.must_include "secret/FOO with key likewhateverman could not be found"
+      end
+
+      describe "scopes" do
+        before do
+          SecretStorage.stubs(:read).returns(true)
```
grosser: create a real key for deeper testing ?

irwaters:
ok, ready for another pass.

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +51,41 @@ def template
       end
     end
 
+    # process the annotations, and expand $ENV and $DEPLOY_GROUP for any annotation
```
grosser: `process/handle/manage` all the same empty fill works :D 

this reads nicely :)
`# expand $ENV and $DEPLOY_GROUP in annotation that start with 'secret/'

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +51,41 @@ def template
...
+    # that begins with 'secret/'
+    def expand_secret_annotations
+      annotations.to_h.each do |annotation_name, secret_key|
+        if annotation_name.to_s.start_with?('secret/')
```
grosser: reuse VAULT_SECRET_BACKEND ?

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +51,41 @@ def template
...
+      errors = []
+      annotations.to_h.each do |annotation_name, secret_key|
+        next unless annotation_name.to_s.start_with?('secret/')
+        begin
+          SecretStorage.read(secret_key)
+        rescue
+          errors << "Secret #{annotation_name} with key #{secret_key} could not be found."
+        end
+      end
+      unless errors.empty?
```
grosser: prefer positive logic: `if errors.any?`

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -47,6 +51,41 @@ def template
...
+
+    # verify that each secret really exists and inform the user
+    # if it does not as the deployment will fail
+    def verify_secret_annotations
+      errors = []
+      annotations.to_h.each do |annotation_name, secret_key|
+        next unless annotation_name.to_s.start_with?('secret/')
+        begin
+          SecretStorage.read(secret_key)
+        rescue
```
grosser: rescue a specific error instead of everything ?

irwaters: ok, got the last batch.

grosser:
👍

grosser:
and squasch :)

PR #1049, opened 2016-06-07
RFC: Make slack output less verbose

rata:
Hi,
Wanted to ask you what do you think of something like this. Do you use the slack plugin? Do you prefer a detailed output? Here people don't like so detailed output, so if you prefer it more verbose maybe I can add a config for it?

What do you think?

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/plugins/slack_webhooks/app/views/samson_slack_webhooks/notification.text.erb
+++ b/plugins/slack_webhooks/app/views/samson_slack_webhooks/notification.text.erb
@@ -4,15 +4,9 @@ There are no new commits since last time.
 <% else %>
 _<%= pluralize changeset.commits.count, "commit" %> by <%= changeset.author_names.to_sentence %>._
 
-*Files changed*
-
-<% changeset.files.each do |file| %>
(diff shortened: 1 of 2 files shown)


grosser:
I don't think we use it atm ... maybe soon ...

And I'd guess `@steved` does not use it either ;)

@gstamp are you still using it ? / does this look good ?

gstamp:
Yep. We still use it but we're using a customised version anyway.  Our version shows PR's first then commit messages only if there are no PRs. We got rid of the files section entirely.

The standard version is very verbose so it makes sense to cut it down.

grosser:
want to bring your version back as a PR ... so we have something that is useable :)

grosser:
@rata 
can you fix the test `mtest plugins/slack_webhooks/test/models/slack_webhook_notification_renderer_test.rb:6
`

rata:
@grosser sorry, fixed now!

PR #1052, opened 2016-06-08
wait when the expected number of pods is not reached fixes #1051

### LINKED ISSUE #1051: Kubernetes deployment runs stability test before all pods started
Hi,

The kubernetes deployment starts the stability test even if not all replicas are started and, if that happens, it detects the new replicas as a restart and says the deployment failed when in fact it went just fine.

The problem seems to be in plugins/kubernetes/app/models/kubernetes/deploy_executor.rb, inwait_for_deploys_to_finish(). When it does:

```
          if statuses.all?(&:live)
            @output.puts "READY, starting stability test"
            @testing_for_stability = 0
```

It should do something like, IIUC:

```
          if statuses.all?(&:live) && replica_target == replica_live
            @output.puts "READY, starting stability test"
            @testing_for_stability = 0
```

I'm not sure how to get the replica_target and replica_live (it seems from the release_doc, but not sure which one as a release has many).

grosser:
@zendesk/paas @rata 

first commit is test organization, second is the actual feature ... easier to review 1-by-1

Risks:
 - deploys might get stuck because this logic is wrong ...

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -38,6 +38,8 @@ module Kubernetes
     def wait_for_deploys_to_finish(release)
       start = Time.now
       stable_ticks = CHECK_STABLE / TICK
+      expected = release.release_docs.to_a.sum(&:desired_pod_count)
+      @output.puts "Waiting for #{expected} pods to be created"
 
       loop do
         return false if stopped?
@@ -59,7 +61,7 @@ module Kubernetes
           end
         else
           print_statuses(statuses)
-          if statuses.all?(&:live)
+          if statuses.all?(&:live) && statuses.count == expected
             @output.puts "READY, starting stability test"
             @testing_for_stability = 0
           elsif statuses.map(&:details).include?(RESTARTED)
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -57,14 +57,14 @@ module Kubernetes
       case deploy_yaml.resource_name
       when 'deployment'
         deploy = Kubeclient::Deployment.new(deploy_yaml.to_hash)
-        if resource_running?(deploy)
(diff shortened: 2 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -38,6 +38,8 @@ def execute!(*_commands)
     def wait_for_deploys_to_finish(release)
       start = Time.now
       stable_ticks = CHECK_STABLE / TICK
+      expected = release.release_docs.to_a.sum(&:desired_pod_count)
```
rata: I tried this, but it seems the desired_pod_count doesn't exist. At least in my version (I tried a small back port, maybe it was that)

rata: Ignore this, it was that :-D

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -107,8 +117,12 @@ def deploy_yaml
       @deploy_yaml ||= DeployYaml.new(self)
     end
 
-    def resource_running?(resource)
-      extension_client.send("get_#{deploy_yaml.resource_name}", resource.metadata.name, resource.metadata.namespace)
+    def deployment
+      extension_client.send(
```
jonmoter: Couldn't this just be `extension_client.get_deployment`, rather than using `#send`?

jonmoter: Or is it also possible for this to be a DaemonSet?  In which case, the method shouldn't be `deployment`.

jonmoter:
👍

plugins/kubernetes/app/models/kubernetes/release_doc.rb
grosser: yes ... needs to be deployish ... or something like that

grosser: going with `deployed`

PR #1053, opened 2016-06-09
role verification

grosser:
bottom of the roles page now has a verify link:
screen shot 2016-06-08 at 5 03 10 pm

which returns a big box to paste the config into:
screen shot 2016-06-08 at 5 02 21 pm

@zendesk/paas

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -221,36 +221,12 @@ module Kubernetes
       if build && kubernetes_role
         if raw_template.blank?
           errors.add(:build, "does not contain config file '#{template_name}'")
-        elsif !project_and_role_consistent?
-          errors.add(:build, "config file '#{template_name}' does not have consistent project and role labels")
-        end
-      end
-    end
-
-    def project_and_role_consistent?
-      labels = parsed_config_file.flat_map do |resource|
-        kind = resource.fetch('kind')
-
-        label_paths =
-          case kind
-          when 'Service'
-            [['spec', 'selector']]
-          when 'Deployment', 'DaemonSet'
-            [
-              ['spec', 'template', 'metadata', 'labels'],
-              ['spec', 'selector', 'matchLabels'],
-            ]
-          else
-            [] # ignore unknown / unsupported types
+        elsif problems = RoleVerifier.new(raw_template).verify
+          problems.each do |problem|
+            errors.add(:build, "#{template_name}: #{problem}")
           end
-
-        label_paths.map do |path|
-          path.inject(resource) { |r, k| r[k] || {} }.slice('project', 'role')
         end
       end
-
-      labels = labels.uniq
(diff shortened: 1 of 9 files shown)


grosser:
also making the release doc use the verifier so we get some consistency / enforcement

jonmoter:
Nice!  I like the structure!  Looks good so far.

grosser:
got all the verifications I could think of ... ready to merge ...

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -179,38 +179,12 @@ def validate_config_file
       if build && kubernetes_role
         if raw_template.blank?
           errors.add(:build, "does not contain config file '#{template_name}'")
-        elsif !project_and_role_consistent?
-          errors.add(:build, "config file '#{template_name}' does not have consistent project and role labels")
+        elsif problems = RoleVerifier.new(raw_template).verify
+          errors.add(:build, "config file '#{template_name}' has errors: #{problems.join("\n")}")
```
jonmoter: Better to add each problem as a separate entry in the `errors` collection?

plugins/kubernetes/app/models/kubernetes/role_verifier.rb
```diff
@@ -0,0 +1,106 @@
+module Kubernetes
+  class RoleVerifier
+    DEPLOYISH = ['Deployment', 'DaemonSet'].freeze
+
+    def initialize(role)
```
jonmoter: Not clear what `role` is supposed to be here. I would've assumed a `Kubernetes::Role` object.  But it's actually supposed to be a string, right?

plugins/kubernetes/app/models/kubernetes/role_verifier.rb
```diff
@@ -0,0 +1,106 @@
...
+      return if containers.all? { |c| c.is_a?(Array) && c.size >= 1 }
+      @errors << "Deployments and DaemonSets need at least 1 container"
+    end
+
+    def all?(accessor)
+      map(accessor).all?
+    end
+
+    # only checks the first element of arrays ... but should be good enough
+    def map(accessor, elements: @elements, array: :all)
```
jonmoter: Can you think of better names than `map` and `accessor`?  Was really confused what this method is intended to do.

plugins/kubernetes/app/models/kubernetes/role_verifier.rb
grosser: role_definition ... the content of the kubernetes/xyz.yml

plugins/kubernetes/app/models/kubernetes/role_verifier.rb
grosser: renamed to map_attributes + path

grosser:
looks good ?

jonmoter:
👍 after fixing merge issues

PR #1058, opened 2016-06-10
Updated Slack Webhooks plugin to support more customization and buddy approval request

zehua:
There are still a bunch of TODOs. But I am sending this for early review of the UI and design before I dive too deep.

- updated the Slack webhook configuration in stage config UI to support configuration of whether to send notification pre and / or post deployment. And also to indicate whether the hook is to be used for buddy approval request
image

- added buddy approval request related UI, similar to that of flowdock
screen shot 2016-06-10 at 5 15 49 pm

- buddy approval request on Slack looks like this:
screen shot 2016-06-10 at 5 15 29 pm

- added a new ENV var SLACK_API_TOKEN for fetching Slack users for @ mentions in buddy approval

- There are quite some amount of duplication of codes. I am not sure what the best way is to share codes between two plugins.

### Tasks
 - [ ] :+1: from team
 - [x] improve the deploy notification message to include PR info
 - [x] add / update tests
 - [x] should I check in changes to `db/schema.rb`?
 - [x] replace slack-ruby-client with raw Faraday calls
 - [x] refactor shared template

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-249

### Risks
- Level: Med. 
- Should affect only the Slack plugin.


plugins/slack_webhooks/app/models/slack_webhook_notification.rb
```diff
@@ -1,14 +1,17 @@
 class SlackWebhookNotification
-  def initialize(deploy)
+  def initialize(deploy, deploy_phase=:for_buddy)
```
grosser: prefer keyword args ... and no default value

plugins/slack_webhooks/app/decorators/stage_decorator.rb
```diff
@@ -11,4 +11,8 @@ def send_slack_webhook_notifications?
   def no_webhook_url?(slack_webhook_attrs)
     slack_webhook_attrs['webhook_url'].blank?
   end
+
+  def for_buddy_channels_names
```
grosser: `def slack_buddy_channels` ?

plugins/slack_webhooks/app/views/samson_slack_webhooks/_notify_buddy_box.html.erb
```diff
@@ -0,0 +1,29 @@
+<% if @deploy.waiting_for_buddy? && @deploy.stage.send_slack_webhook_notifications? %>
+  <slack-buddy-request-box default-buddy-request-message="<%= default_slack_message(@deploy) %>"
+                     channels="<%= @deploy.stage.for_buddy_channels_names.join(', ') %>"
```
grosser: strange indent

plugins/slack_webhooks/app/models/slack_webhook_notification.rb
```diff
@@ -1,14 +1,17 @@
...
     @stage = deploy.stage
     @project = @stage.project
     @user = @deploy.user
   end
 
-  def deliver
+  def deliver(message=content)
     @stage.slack_webhooks.each do |webhook|
-      _deliver(webhook)
+      if webhook.send(@deploy_phase)
```
grosser: public_send

plugins/slack_webhooks/app/controllers/slack_webhooks_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    users = Rails.cache.fetch(:slack_users, expires_in: 5.minutes, race_condition_ttl: 5) do
+      slack_webhooks_service.users
+    end
+    render json: { users: users }
+  rescue Slack::Web::Api::Error => e
+    Rails.logger.error('Could not fetch Slack users! Please set the SLACK_API_TOKEN env variable')
+    render json: { error: "Could not get the users from Slack: #{e.message}" }, status: 404
+  end
+
+  def notify
```
grosser: maybe `def buddy_request` ?

plugins/slack_webhooks/app/models/slack_webhook_notification.rb
```diff
@@ -1,14 +1,17 @@
...
+  def initialize(deploy, deploy_phase=:for_buddy)
     @deploy = deploy
+    @deploy_phase = deploy_phase
     @stage = deploy.stage
     @project = @stage.project
     @user = @deploy.user
   end
 
-  def deliver
+  def deliver(message=content)
```
grosser: do we need the optional argument / can it be a required argument ?

plugins/slack_webhooks/app/views/samson_slack_webhooks/_notify_buddy_box.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <div class="panel panel-info">
+      <div class="panel-heading">
+        <h3 class="panel-title">Request a buddy via Slack</h3>
+      </div>
+      <div class="panel-body">
+        <div class="col-sm-10">
+          <div class="mentions-input-box">
+            <textarea id="buddy_request_box"></textarea>
+          </div>
+		  <em>This will send a buddy request to the <strong>{{ channels }}</strong> channels.</em>
```
grosser: strange indent

plugins/slack_webhooks/db/migrate/20160609164219_add_checkboxes_to_slack_webhooks.rb
```diff
@@ -0,0 +1,7 @@
+class AddCheckboxesToSlackWebhooks < ActiveRecord::Migration
+  def change
+    add_column :slack_webhooks, :before_deploy, :boolean, default: false
```
grosser: `null: false` for all of them

plugins/slack_webhooks/lib/samson_slack_webhooks/samson_plugin.rb
```diff
@@ -14,14 +15,20 @@ class Engine < Rails::Engine
...
+notify_before_deploy = -> (deploy, _buddy) do
   if deploy.stage.send_slack_webhook_notifications?
-    SlackWebhookNotification.new(deploy).deliver
+    SlackWebhookNotification.new(deploy, :before_deploy).deliver
   end
 end
 
-Samson::Hooks.callback :before_deploy, &notify
-Samson::Hooks.callback :after_deploy, &notify
+notify_after_deploy = -> (deploy, _buddy) do
```
grosser: no need to create lambdas when they are not being reused

```
Samson::Hooks.callback :before_deploy do
 . ... code here ...
end
```

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
```diff
@@ -0,0 +1,32 @@
...
+module SamsonSlackWebhooks
+  class SlackWebhooksService
+    delegate :stage, :user, to: :@deploy
+
+    def initialize(deploy)
+      @deploy = deploy
+    end
+
+    def users
+      slack_client.users_list().members.map do |user|
```
grosser: avoid empty params `()`

grosser:
overall looks good, but missing tests :)

plugins/slack_webhooks/app/assets/javascripts/slack_webhooks/application.js
```diff
@@ -0,0 +1,111 @@
...
+      $rootScope.$emit('slack_webhooks_users', users);
+    }).error(function () {
+      console.log('Could not fetch the Slack users!');
+    });
+
+    return users;
+  };
+
+  self.buddyRequest = function (deploy, message) {
+    return $http.post('/slack_webhooks/notify', { deploy_id: deploy, message: message });
```
grosser: prefer reading urls from data-xyz attributes so they do not get hardcoded in js land

plugins/slack_webhooks/app/assets/javascripts/slack_webhooks/application.js
```diff
@@ -0,0 +1,111 @@
+//= require jquery_mentions_input/jquery.elastic.source.js
```
grosser: more docs on what this is / what it does / how it works would be nice

grosser:
can we use jquery for the js part ... render the user names onto the page to make it more responsive ... the filter through them ?

zehua:
> can we use jquery for the js part ... render the user names onto the page to make it more responsive ... then filter through them ?

Most of the js / Angular codes are a direct copy and paste from the corresponding flowdock implementation. I will look into whether the mentionsInput jquery plugin works with a pre supplied user list. At the backend, the api to fetch the full list of users appears to be cached, though.

zehua:
Can we treat buddy approval request feature as a core feature and move some part of the buddy approval request UI and logic out of the flowdock and this slack plugin into the core? That would help reduce code duplication.

zehua:
related to #1049 /cc @rata

grosser:
yeah sure, if multiple plugins need it then it should be in core

rata:
@zehua seems great the functionality, thanks!

plugins/slack_webhooks/app/models/slack_webhook_notification.rb
zehua: done

plugins/slack_webhooks/app/decorators/stage_decorator.rb
zehua: yes

plugins/slack_webhooks/app/views/samson_slack_webhooks/_notify_buddy_box.html.erb
zehua: not relevant now

plugins/slack_webhooks/app/models/slack_webhook_notification.rb
zehua: done

plugins/slack_webhooks/app/controllers/slack_webhooks_controller.rb
zehua: yes

plugins/slack_webhooks/app/models/slack_webhook_notification.rb
zehua: Tried to tweak it but gave up. The default value `content()` is for deployment notification message, which was computed in this class. We only want to explicit specify the `message` for buddy request.

plugins/slack_webhooks/app/views/samson_slack_webhooks/_notify_buddy_box.html.erb
zehua: not relevant. I updated it to use the new template in #1064

plugins/slack_webhooks/db/migrate/20160609164219_add_checkboxes_to_slack_webhooks.rb
zehua: done

plugins/slack_webhooks/lib/samson_slack_webhooks/samson_plugin.rb
zehua: done

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
zehua: yeap. This is in fact nicely reported by rubocop.

.env.example
```diff
@@ -57,6 +57,8 @@ GITHUB_TOKEN={fill me in}
 
 # FLOWDOCK_API_TOKEN= # required for the flowdock integration user mention autocomplete
 
+# SLACK_API_TOKEN= # only required for the slack integration to fetch users in the buddy approval request form (when BUDDY_CHECK_FEATURE=1)
```
grosser: I think we sh flowdock and this one optional ... also flowdock needs this `only when buddy` comment

Gemfile.lock
```diff
@@ -431,6 +435,14 @@ GEM
       simplecov-html (~> 0.10.0)
     simplecov-html (0.10.0)
     single_cov (0.2.2)
+    slack-ruby-client (0.7.4)
```
grosser: can we use https://github.com/stevenosloan/slack-notifier instead ... has 0 dependencies ...

plugins/slack_webhooks/app/views/samson_slack_webhooks/_notify_buddy_box.html.erb
```diff
@@ -0,0 +1,67 @@
+<% if @deploy.waiting_for_buddy? && @deploy.stage.send_slack_webhook_notifications? %>
```
grosser: can we render a shared view here ?
move this into a app/view/shared/_buddy_request and reuse it

zehua: yeah

plugins/slack_webhooks/app/models/slack_webhook_notification.rb
zehua: ok. I gave it a second try, separating the `deliver` and `buddy_request` methods. Should be better now.

zehua:
Do we usually fix the codeclimate detected issues? https://codeclimate.com/repos/53340bef6956800b9000675c/pull/1058

grosser:
nah, unless they are valid/bad things :)

Gemfile.lock
zehua: This library only supports sending message to a channel via webhook, which we do it via direct `Faraday` POSTs at the moment. It does not support the other slack apis, e.g., the `users_list` api that we need.

grosser: how about we use faraday gets for these then ? :)

zehua: ok, removed slack-ruby-client and used direct faraday instead.

plugins/slack_webhooks/app/views/samson_slack_webhooks/notification.text.erb
```diff
@@ -1,12 +1,6 @@
...
 There are no new commits since last time.
 <% else %>
-_<%= pluralize changeset.commits.count, "commit" %> by <%= changeset.author_names.to_sentence %>._
-
-*Commits*
-
-<% changeset.commits[0..4].each do |commit| %>
-> <<%= commit.url + '|' + commit.summary %>> (<%= commit.author_name %>)
-<% end %>
+_<<%= changeset.github_url %>|<%= pluralize changeset.commits.count, "commit" %>> and <%= pluralize changeset.pull_requests.count, "pull request" %> by <%= changeset.author_names.to_sentence %>._
```
sandlerr: I know you didn't create this, but can we replace it by the html entities? Or does this actually generate html tags?

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
```diff
@@ -0,0 +1,49 @@
+module SamsonSlackWebhooks
+  class SlackWebhooksService
+    # users in the format jquery mentions input needs
+    # see _notify_buddy_box.html.erb
+    def users
+      Rails.cache.fetch(:slack_users, expires_in: 5.minutes, race_condition_ttl: 5) do
+        if slack_api_token
```
sandlerr: can we early return instead of nesting the rest of the function body in a conditional? Asking because it is quite deeply nested

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
```diff
@@ -0,0 +1,49 @@
...
+              body['members'].map do |user|
+                {
+                  id: user['id'],
+                  name: user['name'],
+                  avatar: user['profile']['image_48'],
+                  type: 'contact'
+                }
+              end
+            else
+              Rails.logger.error("Error fetching slack users: #{body['error']}")
```
sandlerr: not sure if we care about the HTTP status as well

plugins/slack_webhooks/app/views/samson_slack_webhooks/notification.text.erb
zehua: It actually follow some markdown like syntax, not html: https://api.slack.com/docs/message-formatting

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
zehua: It's not explicitly documented. But it seems to imply that the api would always return 200. https://api.slack.com/web#basics

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
zehua: yeah, done

plugins/slack_webhooks/test/samson_slack_webhooks/slack_webhooks_service_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    it "returns empty users when there's an error" do
+      stub_request(:post, "https://slack.com/api/users.list").
+        to_return(body: {ok: false, error: "some error"}.to_json)
+      Rails.logger.expects(:error).with('Error fetching slack users: some error')
+      service.users.must_equal([])
+
+      # is cached
+      service.users.object_id.must_equal service.users.object_id
```
sandlerr: this is an interesting way to test! I like the idea, but I think it might be a little misleading if it is only really valid for the cache used in the test environment. Probably not worth changing, another way to express this might be to assert that the `Faraday` method isn't invoked.

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
```diff
@@ -0,0 +1,49 @@
+module SamsonSlackWebhooks
+  class SlackWebhooksService
+    # users in the format jquery mentions input needs
+    # see _notify_buddy_box.html.erb
+    def users
+      Rails.cache.fetch(:slack_users, expires_in: 5.minutes, race_condition_ttl: 5) do
+        if !slack_api_token
+          Rails.logger.error('Set the SLACK_API_TOKEN env variable to enabled user mention autocomplete.')
+          []
```
sandlerr: sorry to nitpick, in my comment before I meant something like

```ruby
def users
  unless slack_api_token
    Rails.logger.error('Set the SLACK_API_TOKEN env variable to enabled user mention autocomplete.')
    return []
  end
  body = JSON.parse(Faraday.post("https://slack.com/api/users.list", token: slack_api_token).body)
  if body['ok']
    ...
end
```

due to the early return there's no need to indent the rest of the function in an `else` block (and I think we can get away with not using `begin` / `end` here as well

sandlerr:
only other change I'd suggest is maybe moving the JS code to an asset file. Also this has a merge conflict.

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
zehua: doing a `return []` seems to break the caching. I will leave it for now.

lib/samson/hooks.rb
```diff
@@ -144,8 +144,10 @@ def fire(name, *args)
       end
 
       def render_views(name, view, *args)
-        hooks(name).each_with_object("".html_safe) do |partial, html|
-          html << view.render(partial, *args)
+        hooks(name).each_with_object("".html_safe) do |value, html|
+          partial, params_generator = value
+          new_args = params_generator ? [params_generator.call(*args)] : args
```
zehua: Adding this optional context generator to give the plugin hook a chance to customize the render parameters.

zehua:
Did the refactoring to move the duplicated `_notify_buddy_box` template to `shared`.

However, I was not able to use https://rails-assets.org/#/components/jquery-mentions-input, as its 1.6.0 version has a different jquery dependency that bundler complains about. It ends up with 1.5.0 version which is buggy (reporting JS errors on console). So I ended up moving the jquery.mentionsInput js/css into `app/assets`. Hopefully this is ok.

grosser:
did the mentions input in flowdock not work ?

lib/samson/hooks.rb
grosser: that looks super funky ... why do we need that ?

plugins/flowdock/app/models/flowdock_notification.rb
```diff
@@ -25,6 +25,13 @@ def deliver
     Airbrake.notify(error, error_message: 'Could not deliver flowdock message')
   end
 
+  def default_buddy_request_message
+    project = @deploy.project
```
grosser: `@deploy` exists in the model ?

plugins/flowdock/lib/samson_flowdock/samson_plugin.rb
```diff
@@ -4,7 +4,19 @@ class Engine < Rails::Engine
 end
 
 Samson::Hooks.view :stage_form, "samson_flowdock/fields"
-Samson::Hooks.view :deploy_view, 'samson_flowdock/notify_buddy_box'
+Samson::Hooks.view :deploy_view, "shared/notify_buddy_box" do |deploy:, project:|
+  {
```
grosser: render a template that renders the shared template with custom parameters

zehua:
> did the mentions input in flowdock not work ?

It's the same JS/css. Just that I had to move them into core because the template that uses it is now in core.

plugins/slack_webhooks/test/decorators/stage_decorator_test.rb
```diff
@@ -27,4 +27,32 @@
       stage.slack_webhooks.size.must_equal 0
     end
   end
+
+  describe "#slack_buddy_channels" do
+    it "returns channels when for_buddy is true" do
+      stage.slack_webhooks = [SlackWebhook.new(channel: "test", webhook_url: "http://slack.com/abcdef",
```
grosser: 2-space indent plz

plugins/slack_webhooks/test/models/slack_webhook_notification_test.rb
```diff
@@ -3,29 +3,45 @@
...
   let(:user) { stub(name: "John Wu", email: "wu@rocks.com") }
   let(:endpoint) { "https://slack.com/api/chat.postMessage" }
-  let(:webhook) { stub(webhook_url: endpoint, channel: nil) }
-  let(:stage) { stub(name: "staging", slack_webhooks: [webhook], project: project) }
-  let(:deploy) { stub(summary: "hello world!", user: user, stage: stage) }
-  let(:notification) { SlackWebhookNotification.new(deploy) }
+
+  def stub_notification(before_deploy: false, after_deploy: true, for_buddy: false)
+    webhook = stub(webhook_url: endpoint, channel: nil,
+                   before_deploy: before_deploy, after_deploy: after_deploy, for_buddy: for_buddy)
```
grosser: 2-space indent plz

grosser:
I like it, looks pretty dry :)

plugins/flowdock/app/models/flowdock_notification.rb
zehua: Yeah. It's passed in during `initialize`.

plugins/slack_webhooks/test/decorators/stage_decorator_test.rb
zehua: do you mean 2 space indent in the next line?

grosser: yes

db/schema.rb
```diff
@@ -336,6 +336,9 @@
     t.integer  "stage_id",    limit: 4,     null: false
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.boolean  "before_deploy",               default: false, null: false
```
zehua: Because the migration file was generated two weeks ago, the schema version (aka the timestamp) is now not the latest, due to other PRs merged. If we haven't deployed to production, since two weeks ago, it's probably fine. Otherwise, I could re-generate the migrations file to give it a newer timestamp.

lib/samson/hooks.rb
zehua: The same template is shared by two different plugins which have different ways of computing the template parameters. e.g., one determines whether to show the form based on whether there's flowdock configuration and the other decides based on slack configuration.

Is there a different way to do it?

grosser: make it render a unique template and then that template can render the
shared template

zehua: will give it a try

plugins/slack_webhooks/test/decorators/stage_decorator_test.rb
zehua: Getting complaints from rubocop when I move the next line to 2-space indent instead of aligning with the hash key in the previous line:
```
plugins/slack_webhooks/test/decorators/stage_decorator_test.rb:34:9: C: Align the elements of a hash literal if they span more than one line.
        stage_id: stage.id, for_buddy: true)]
        ^^^^^^^^^^^^^^^^^^
```

app/assets/javascripts/jquery-mentions-input/jquery.mentionsInput.js
```diff
@@ -248,9 +248,10 @@
...
-      return c = document.createElement("div"), $(c).css(g), $(i).after(c), b = document.createTextNode(i.value.substring(0, i.selectionEnd)), a = document.createTextNode(i.value.substring(i.selectionEnd)), d = document.createElement("span"), d.innerHTML = "&nbsp;", c.appendChild(b), c.appendChild(d), c.appendChild(a), c.scrollTop = i.scrollTop, f = $(d).position(), $(c).remove(), f
+      };
+      h = ["boxSizing", "fontFamily", "fontSize", "fontStyle", "fontVariant", "fontWeight", "height", "letterSpacing", "lineHeight", "paddingBottom", "paddingLeft", "paddingRight", "paddingTop", "textDecoration", "textIndent", "textTransform", "width", "word-spacing"];
+      for (j = 0, k = h.length; j < k; j++) { e = h[j]; g[e] = $(i).css(e); }
+      return c = document.createElement("div"), $(c).css(g), $(i).after(c), b = document.createTextNode(i.value.substring(0, i.selectionEnd)), a = document.createTextNode(i.value.substring(i.selectionEnd)), d = document.createElement("span"), d.innerHTML = "&nbsp;", c.appendChild(b), c.appendChild(d), c.appendChild(a), c.scrollTop = i.scrollTop, f = $(d).position(), $(c).remove(), f;
```
zehua: Had to do these to make jshint happy: https://travis-ci.org/zendesk/samson/jobs/139819098

Looks like we only do jshint for the core JS, not the JS from plugins.

plugins/slack_webhooks/test/decorators/stage_decorator_test.rb
grosser: foo(
  bar: 1,
  foo: 2
)

db/schema.rb
zehua: I've re-created the file with a newer timestamp.

plugins/slack_webhooks/test/decorators/stage_decorator_test.rb
zehua: fixed

plugins/slack_webhooks/test/models/slack_webhook_notification_test.rb
zehua: done

lib/samson/hooks.rb
zehua: done.

could not figure out how to test the parameters passed to sub template rendering. so I removed the tests. :|

sandlerr:
were you not able to import https://rails-assets.org/#/components/jquery-mentions-input ?

grosser:
I tried that and it was broken / messed up ... some other existed and that
wanted coffee-script and not even then it worked ... I'd prefer to keep
this PR simple ... we can worry about the js source in another PR :)

zehua:
> were you not able to import https://rails-assets.org/#/components/jquery-mentions-input ?

I was able to. But as I mentioned in the other comment https://github.com/zendesk/samson/pull/1058#issuecomment-228117693, due to jquery version conflict, bundler decides to pull down only 1.5.0, which is buggy. 1.6.0 is the latest but the way its has a stricter jquery dependency of `1.6.x`, which is not compatible with other jquery plugins. :|

grosser:
no need to solve that right now, we use the same version for flowdock ...
great for another PR ... updating jquery would be nice anyway :)

zehua:
Slack now supports better interaction UI in the messages. https://api.slack.com/docs/message-buttons

I am a bit worried that if we embed the approve button in the buddy request message, buddies will just blindly click on approve without really reviewing what's going to be deployed.

Maybe discussion and considerations for another PR.

grosser:
approve is meant to be a review ...
 - needs to be authorized
 - needs to have seen the diff / sanity checked

... same as giving a +1 via email, defeats the purpose of doing this whole
thing :)

zehua:
all TODOs are done. Please review again.

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
grosser: keeping the return outside is fine / will make the caching not necessary (faster to check for token then to ask memcached)

plugins/slack_webhooks/app/decorators/stage_decorator.rb
```diff
@@ -6,6 +6,14 @@ def send_slack_webhook_notifications?
     slack_webhooks.any?
   end
 
+  def slack_buddy_channels
+    slack_webhooks.for_buddy.map(&:channel)
```
grosser: `.pluck(:channel)` ?

grosser: also memoize this since it will be used multiple times on 1 page ?

plugins/slack_webhooks/lib/samson_slack_webhooks/samson_plugin.rb
```diff
@@ -14,14 +15,17 @@ class Engine < Rails::Engine
...
+Samson::Hooks.callback :before_deploy do |deploy, _buddy|
   if deploy.stage.send_slack_webhook_notifications?
-    SlackWebhookNotification.new(deploy).deliver
+    SlackWebhookNotification.new(deploy).deliver(:before_deploy)
   end
 end
 
-Samson::Hooks.callback :before_deploy, &notify
-Samson::Hooks.callback :after_deploy, &notify
+Samson::Hooks.callback :after_deploy do |deploy, _buddy|
```
grosser: might be able to do `[:before_deploy, :after_deploy].each do |callback|`

grosser:
looks great ... @sandlerr are you ok with the current state too ?
(after caching fixes)

plugins/slack_webhooks/app/decorators/stage_decorator.rb
zehua: > also memoize this since it will be used multiple times on 1 page ?

Yep, this is used twice on this page. The log shows that the SQL is hit twice:
```
  SlackWebhook Load (1.2ms)  SELECT "slack_webhooks".* FROM "slack_webhooks" WHERE "slack_webhooks"."stage_id" = ? AND "slack_webhooks"."for_buddy" = ?  [["stage_id", 2], ["for_buddy", "t"]]
  CACHE (0.2ms)  SELECT "slack_webhooks".* FROM "slack_webhooks" WHERE "slack_webhooks"."stage_id" = ? AND "slack_webhooks"."for_buddy" = ?  [["stage_id", 2], ["for_buddy", true]]
```
what's the best practice in memoizing this?

sandlerr: conventionally

```ruby
def slack_buddy_channels
  @slack_buddy_channels ||= slack_webhooks.for_buddy.map(&:channel)
end
```

sandlerr: although, if you check out section 1.7 of http://edgeguides.rubyonrails.org/caching_with_rails.html#low-level-caching and look at the timings reported in the log you posted, maybe its not essential.

plugins/slack_webhooks/lib/samson_slack_webhooks/samson_plugin.rb
zehua: done

plugins/slack_webhooks/app/decorators/stage_decorator.rb
zehua: cool, did not realize that the `stage` object is created for every request.

zehua: done both `pluck` and memoization.

grosser:
Looks good to me, can you add the early return when there is not token, so we don't have to go to the cache ?

otherwise 👍 
@sandlerr any other concerns left ?

sandlerr:
:+1: on green + early return

zehua:
ok, updated to do early return.

grosser:
:heart: thx for the great work :D

zehua:
Thanks guys.

Follow up tasks that I leave out from this PR:
* update jquery-mentions-input 1.6 on rails-assets so that it works with other jquery plugins, mostly about relaxing the dependency on 1.6.x of jquery, optionally adding a `jquery-mentions-input.js` file to make `require` easier
* move the JS in `shared/_notify_buddy_box` out into the `deploy.js`

grosser:
FYI had to fix the migration since the name did not match the class https://github.com/zendesk/samson/commit/b3d250470fd752a75f7318e57f9c982b768776d6

zehua:
ah, I messed it up when updating the timestamp of the migration file. thanks for the quick fix.

PR #1059, opened 2016-06-10
fix crash that was happening on KubernetesCluster#show page

jonmoter:
Error was:

```
ActionView::Template::Error: undefined method `use_ssl?' for #<Kubeclient::Config::Context:0x007fd7dc2b0550>
    plugins/kubernetes/app/views/admin/kubernetes/clusters/show.html.erb:28:in `_plugins_kubernetes_app_views_admin_kubernetes_clusters_show_html_erb___2430964412720577784_70282544303400'
```

Added a test for the controller/view

/cc @grosser, @irwaters 

### Tasks
 - [x] :+1: from team

### Risks
- Level: Low


grosser:
👍

PR #1060, opened 2016-06-10
show latest webhook request and response

grosser:
@zendesk/samson 

screen shot 2016-06-10 at 11 26 27 am

## Risks
 - potentially some secret data in there ... request is only shown to admins

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -2,11 +2,18 @@ class Integrations::BaseController < ApplicationController
   skip_around_action :login_user
   skip_before_action :verify_authenticity_token
   before_action :validate_request
+  after_action :record_webhook
 
   def create
-    return head(:ok) unless deploy?
+    unless deploy?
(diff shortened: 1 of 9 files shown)


app/controllers/integrations/base_controller.rb
```diff
@@ -2,11 +2,18 @@ class Integrations::BaseController < ApplicationController
   skip_around_action :login_user
   skip_before_action :verify_authenticity_token
   before_action :validate_request
+  after_action :record_webhook
 
   def create
-    return head(:ok) unless deploy?
+    unless deploy?
+      record_log :error, "Request is not a deploy"
```
jonmoter: Not really an error. We'll probably get notifications for a bunch of stuff that shouldn't be a deploy, like a PR against a branch that isn't master.

jonmoter:
Looks good. 👍 on 💚

app/controllers/integrations/base_controller.rb
grosser: changed to info

PR #1061, opened 2016-06-10
Feature/kubernetes jobs

gonzalorey:
### Description
An initial integration with Kubernetes, allowing us to schedule tasks such as db migrations. It was added as a plugin without disrupting the rest of the service.

The UI now includes the Kubernetes tab, here are a few screenshots showing the new flow.

- The summary of the available Roles and the creation of one:
screen shot 2016-06-10 at 12 08 38 pm
screen shot 2016-06-10 at 1 47 49 pm
The yaml to load must have a format as follows:
````
apiVersion: batch/v1
kind: Job
metadata:
  labels:
    project: app
    task: db-migrate
spec:
  template:
    spec:
      containers:
      - name: app
        image: registry/app:latest
        command: ["bin/rake", "db:migrate"]
      restartPolicy: Never
````

- The Tasks summary, its creation and the execution of a Job.
screen shot 2016-06-10 at 11 11 14 am
screen shot 2016-06-10 at 11 12 08 am

- The Job execution and its progress:
pasted image at 2016_06_15 05_14 pm

- A summary of the executed jobs:
screen shot 2016-06-10 at 11 15 39 am

- A detail of the scheduled releases and the creation of one:
screen shot 2016-06-10 at 1 53 13 pm
screen shot 2016-06-10 at 1 52 05 pm

We are **missing the unit tests**, though we thought we could submit the pull request for review and in the meantime, we start adding the things that we know are missing, such as a better task execution summary, a proper error management, some UI glitches and of course the unit tests.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 https://github.com/zendesk/samson/issues/869

### Risks
- Level: **Medium**

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160621174658) do
+ActiveRecord::Schema.define(version: 20160617223512) do
 
   create_table "builds", force: :cascade do |t|
     t.integer  "project_id",                       null: false
@@ -187,6 +187,35 @@ ActiveRecord::Schema.define(version: 20160621174658) do
   add_index "kubernetes_deploy_group_roles", ["deploy_group_id"], name: "index_kubernetes_deploy_group_roles_on_deploy_group_id", using: :btree
   add_index "kubernetes_deploy_group_roles", ["project_id", "deploy_group_id", "kubernetes_role_id"], name: "index_kubernetes_deploy_group_roles_on_project_id", using: :btree
 
+  create_table "kubernetes_job_docs", force: :cascade do |t|
+    t.integer  "job_id",          limit: 4,                       null: false
+    t.integer  "deploy_group_id", limit: 4,                       null: false
+    t.string   "status",          limit: 255, default: "created", null: false
+    t.datetime "created_at",                                      null: false
+    t.datetime "updated_at",                                      null: false
+  end
+
+  add_index "kubernetes_job_docs", ["deploy_group_id"], name: "index_kubernetes_job_docs_on_deploy_group_id", using: :btree
+  add_index "kubernetes_job_docs", ["job_id"], name: "index_kubernetes_job_docs_on_job_id", using: :btree
+
+  create_table "kubernetes_jobs", force: :cascade do |t|
+    t.integer  "stage_id",           limit: 4,                         null: false
+    t.integer  "kubernetes_task_id", limit: 4,                         null: false
+    t.integer  "user_id",            limit: 4,                         null: false
+    t.integer  "build_id",           limit: 4
+    t.string   "status",             limit: 255,   default: "pending", null: false
+    t.string   "commit",             limit: 255
+    t.string   "tag",                limit: 255
+    t.text     "output",             limit: 65535
+    t.datetime "started_at"
+    t.datetime "created_at",                                           null: false
+    t.datetime "updated_at",                                           null: false
+  end
+
+  add_index "kubernetes_jobs", ["kubernetes_task_id"], name: "index_kubernetes_jobs_on_kubernetes_task_id", using: :btree
+  add_index "kubernetes_jobs", ["stage_id"], name: "index_kubernetes_jobs_on_stage_id", using: :btree
+  add_index "kubernetes_jobs", ["user_id"], name: "index_kubernetes_jobs_on_user_id", using: :btree
+
   create_table "kubernetes_release_docs", force: :cascade do |t|
     t.integer  "kubernetes_role_id",          limit: 4,                         null: false
     t.integer  "kubernetes_release_id",       limit: 4,                         null: false
@@ -227,6 +256,17 @@ ActiveRecord::Schema.define(version: 20160621174658) do
 
   add_index "kubernetes_roles", ["project_id"], name: "index_kubernetes_roles_on_project_id", using: :btree
   add_index "kubernetes_roles", ["service_name", "deleted_at"], name: "index_kubernetes_roles_on_service_name_and_deleted_at", unique: true, length: {"service_name"=>191, "deleted_at"=>nil}, using: :btree
+
+  create_table "kubernetes_tasks", force: :cascade do |t|
+    t.integer  "project_id",  limit: 4,   null: false
+    t.string   "name",        limit: 255, null: false
+    t.string   "config_file", limit: 255
+    t.datetime "deleted_at"
+    t.datetime "created_at",              null: false
+    t.datetime "updated_at",              null: false
+  end
+
+  add_index "kubernetes_tasks", ["project_id"], name: "index_kubernetes_tasks_on_project_id", using: :btree
 
   create_table "locks", force: :cascade do |t|
     t.integer  "stage_id",    limit: 4
--- a/plugins/kubernetes/app/controllers/kubernetes/tasks_controller.rb
+++ b/plugins/kubernetes/app/controllers/kubernetes/tasks_controller.rb
@@ -1,0 +1,56 @@
+class Kubernetes::TasksController < ApplicationController
+  include CurrentProject
+
+  DEPLOYER_ACCESS = [:index, :show].freeze
+  before_action :authorize_project_deployer!, only: DEPLOYER_ACCESS
+  before_action :authorize_project_admin!, except: DEPLOYER_ACCESS
+  before_action :find_task, only: [:show, :update, :run, :destroy]
+
+  def index
+    @tasks = ::Kubernetes::Task.not_deleted.where(project: current_project).order('name desc')
+  end
+
+  def seed
+    Kubernetes::Task.seed!(@project, params.require(:ref))
+    redirect_to action: :index
+  end
+
+  def new
+    @task = Kubernetes::Task.new
+  end
+
+  def create
+    @task = Kubernetes::Task.new(task_params.merge(project: @project))
+    if @task.save
+      redirect_to action: :index
+    else
+      render :new
+    end
+  end
+
+  def show
+  end
+
+  def update
+    if @task.update_attributes(task_params)
+      redirect_to action: :index
+    else
+      render :show
+    end
+  end
+
+  def destroy
+    @task.soft_delete!
+    redirect_to action: :index
+  end
+
+  private
+
+  def find_task
+    @task = Kubernetes::Task.not_deleted.find(params[:id])
+  end
+
+  def task_params
+    params.require(:kubernetes_task).permit(:name, :config_file)
+  end
+end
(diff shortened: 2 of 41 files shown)

### CODE REVIEW

File: db/schema.rb [hunk 1]
```diff
@@ -187,6 +187,41 @@
   add_index "kubernetes_deploy_group_roles", ["deploy_group_id"], name: "index_kubernetes_deploy_group_roles_on_deploy_group_id", using: :btree
   add_index "kubernetes_deploy_group_roles", ["project_id", "deploy_group_id", "kubernetes_role_id"], name: "index_kubernetes_deploy_group_roles_on_project_id", using: :btree
 
+  create_table "kubernetes_job_docs", force: :cascade do |t|
+    t.integer  "kubernetes_task_id", limit: 4,                       null: false
+    t.integer  "job_id",             limit: 4,                       null: false
+    t.integer  "build_id",           limit: 4,                       null: false
+    t.integer  "deploy_group_id",    limit: 4,                       null: false
+    t.string   "status",             limit: 255, default: "created"
```
grosser: null false ?

db/schema.rb
```diff
@@ -187,6 +187,41 @@
...
+    t.datetime "created_at"
```
grosser: null false

plugins/kubernetes/db/migrate/20160606202534_create_kubernetes_jobs.rb
```diff
@@ -0,0 +1,16 @@
+class CreateKubernetesJobs < ActiveRecord::Migration
+  def change
+    create_table :kubernetes_jobs do |t|
+      t.references :stage, null: false, index: true
+      t.references :kubernetes_task, null: false, index: true
+      t.references :user, null: false, index: true
+      t.references :build
```
grosser: more columns null false ?

plugins/kubernetes/config/routes.rb
```diff
@@ -18,6 +18,12 @@
           post :seed
         end
       end
+      resources :tasks, except: :edit, path: :tasks do
```
grosser: why is the :path needed ?

plugins/kubernetes/config/routes.rb
```diff
@@ -18,6 +18,12 @@
...
+        collection do
+          post :seed
+        end
+        resources :kubernetes_jobs, only: [:new, :create, :index], path: :jobs
```
grosser: put the controller under kubernetes folder

grosser:
tests

plugins/kubernetes/app/controllers/kubernetes/kubernetes_jobs_controller.rb
```diff
@@ -0,0 +1,32 @@
+class Kubernetes::KubernetesJobsController < ApplicationController
```
grosser: can be renamed to jobs_controller

plugins/kubernetes/app/controllers/kubernetes/tasks_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  DEPLOYER_ACCESS = [:index, :show].freeze
+  before_action :authorize_project_deployer!, only: DEPLOYER_ACCESS
+  before_action :authorize_project_admin!, except: DEPLOYER_ACCESS
+  before_action :find_task, only: [:show, :update, :run, :destroy]
+
+  def index
+    @tasks = ::Kubernetes::Task.not_deleted.where(project: current_project).order('name desc').to_a
+    respond_to do |format|
+      format.html
+      format.json { render json: @tasks, root: false }
```
grosser: why do we need json ?

plugins/kubernetes/app/models/kubernetes/job_executor.rb
```diff
@@ -0,0 +1,244 @@
+# executes a deploy and writes log to job output
```
grosser: either extract static methods to share logic or some other shared object to avoid all this copy-paste ...

plugins/kubernetes/app/models/kubernetes/job_service.rb
```diff
@@ -0,0 +1,25 @@
...
+      job_execution = JobExecution.new(job.commit, job) do |execution, tmp_dir|
+        @output = execution.output
+        job_executor = JobExecutor.new(@output, job: job)
+        job_executor.execute!
+      end
+
+      JobExecution.start_job(job_execution)
+    end
+
+    private
```
grosser: can go away

plugins/kubernetes/app/models/kubernetes/job_yaml.rb
```diff
@@ -0,0 +1,146 @@
+module Kubernetes
```
grosser: same here, lots of copy-paste

grosser:
the approach looks fine ... does it somehow chain tasks to releases ?

make sure to clean up that copy-paste too ... either some static methods or inheritance or service objects should do the trick ...

FYI one screenshot is using the js release UI ... try using regular stages/deploys ... that should give much better feedback

grosser:
it's kind of sad that it does not solve the `real` problem of `run this task before deploying this role` ... but I guess it brings us 1 step closer / unblocks not having migrations at all :)

plugins/kubernetes/app/models/kubernetes/job.rb
```diff
@@ -0,0 +1,103 @@
...
+      status!("running")
+    end
+
+    def update_output!(output)
+      update_attribute(:output, output)
+    end
+
+    def update_git_references!(commit:, tag:)
+      update_columns(commit: commit, tag: tag)
+    end
```
jonmoter: I don't get the purpose of these `update` methods? Why not just normal mutators like `output=`, `commit=`, etc?

jonmoter:
Could you explain the purpose of the models you've added?  I see KubernetesJob, KubernetesJobDoc, and KubernetesTask, but I'm not quite clear on the distinction.

db/schema.rb
pabloh: Which column is the problem, `build_id` ?

plugins/kubernetes/app/models/kubernetes/job.rb
pabloh: We more or less followed what there was here: https://github.com/zendesk/samson/blob/master/app/models/job.rb#L109

gonzalorey:
Hi @jonmoter, the difference between them lays in the following:
A `KubernetesTask` is a sort of release that is executed through a `KubernetesJob` which comprises the entire execution across pods, while the `KubernetesJobDoc` relates to a single pod.
We chose those names in order to follow your naming conventions like `Release` and `ReleaseDoc`.

plugins/kubernetes/db/migrate/20160606202534_create_kubernetes_jobs.rb
gonzalorey: Just like in `Kubernetes::Release`, first the job is created and then the build, therefore, we cannot set it to be null: false.

gonzalorey:
Hi @jonmoter, @grosser, here we are adding the required changes in order to get this PR approved, as well as improving a few of the things we had pending, such as the progress of the job being executed and the possibility of accessing the executed jobs from a task. Were you able to review this latest changes, and in case you have any suggestion, It would be more than welcome.

We also saw in this thread https://github.com/zendesk/samson/pull/1076 that some of the work we've been doing might get duplicated, even some clases and name are colliding.

db/schema.rb
```diff
@@ -11,20 +11,20 @@
...
     t.string   "docker_repo_digest",  limit: 255
-    t.integer  "docker_build_job_id"
+    t.integer  "docker_build_job_id", limit: 4
     t.string   "label",               limit: 255
     t.string   "description",         limit: 1024
-    t.integer  "created_by"
+    t.integer  "created_by",          limit: 4
     t.datetime "created_at"
     t.datetime "updated_at"
   end
```
jonmoter: Could you revert the changes in schema.rb that aren't related to your migrations, and are just whitespaces changes?

I know that when schema.rb get regenerated it frequently changes whitespace. But it becomes difficult to look at git history and see when the schema actually changed when there are a bunch of whitespace-only changes.

gonzalorey: Done! 🌵

db/schema.rb
```diff
@@ -34,7 +34,7 @@
   add_index "builds", ["project_id"], name: "index_builds_on_project_id", using: :btree
 
   create_table "commands", force: :cascade do |t|
-    t.text     "command",    limit: 10485760
+    t.text     "command",    limit: 16777215
```
jonmoter: Please also revert the changes to the limits that are scattered through schema.rb.  Depending on the DB configuration of the host where you ran the migrations, these limits are different.

If the migrations you added didn't change a table in this PR, we shouldn't see any changes to the table in schema.rb. Sorry if I wasn't clear in my earlier comment.

gonzalorey: No worries, done as well!

db/schema.rb
gonzalorey: Hey @jonmoter, do you want me to remove all those **limit** changes added to the `db/schema.rb` file as well?

jonmoter: Yes, all the limits too.  Since your migrations didn't make any changes to the `builds` table, there should be no changes to it in schema.rb

grosser:
starting a cleanup/testing run on this, will open a new PR when done since I cannot push here ...

gonzalorey:
Great @grosser, should we keep adding changes ourselves or do you prefer doing it yourself and ping us on anything that comes up?

grosser:
I bet I'll move a bunch of stuff around, so better hold off on changes till
I'm done with that ...

gonzalorey:
Ok, keep us updated 😃

grosser:
can we use metadata.name instead of metadata.labels.task for the name ?

grosser:
the more I look at this the less I like it :(

this adds a copy-pasty way of executing a different kind of config a completely different way that is not integrated into anything ...

what we really want is: "run db migrations before every deploy to production" ...
 - make a job just another role
 - when deploying we will detect that this role is a job and run it before the other roles start

-> 
advantage:
 - no new UI
 - everything is automated/integrated
 - existing cap workflows are easy to transition
 - nobody needs to learn a new workflow

disadvantage:
 - we still cannot execute 1-off jobs ... but that could easily be added by having a stage that does not deploy all `roles`

grosser:
started work on this in a new branch ... I'll see how far I get before running into any walls ...

refactor/cleanup branch of this PR is here ... stopped working on it once I realized how complicated this all is ... https://github.com/zendesk/samson/pull/1091

gonzalorey:
OK, let us know if we can do anything in the meantime.

grosser:
closing this since https://github.com/zendesk/samson/pull/1092 is ready to roll :)

gonzalorey:
Sweet 💃

Issue #1062, opened 2016-06-10
New user access rights


ibrahima:
It seems like out of the box, anyone with a valid github account can sign into an instance of Samson and have an account. This is weird because you can access the "Deploys" tab by default, and see a history of who has deployed and full log output. Is there a way to restrict this? It seems like new user signups should need to be approved by default, and additionally new users should not have access to anything until they're actually added to a role.

I guess currently we could put Samson behind HTTP basic auth for minimal protection, but this seems like a pretty big oversight to allow anyone to view deploys by default.

Thanks!

grosser:
2 other options would be:
 - pick a GITHUB_ORGANIZATION to restrict access
 - use other auth that is more restrictive

Having everyone be able to see past deploys might be a security issue, but it's a nice default for open-source projects / to get started ... we could make GITHUB_ORGANIZATION required so it needs to be `all` to allow this / remind people to set it ... I'll update the .env.example to make it more obvious for now ...

ibrahima:
Oh, I missed that section of the configuration options. I can see that it'd definitely be a useful default for open source organizations, but I like the idea of requiring an explicit setting so that it's clear what's happening. Thanks for the clarification!

PR #1063, opened 2016-06-10
docs for making github auth not allow everyone in

grosser:
@zendesk/paas 

fixes https://github.com/zendesk/samson/issues/1062


irwaters:
👍

PR #1065, opened 2016-06-13
automatic ecr repo creation

grosser:
@zendesk/paas @rata 

 - show user a nicer error when unable to authenticate to the registry
 - only create ecr client when trying to push an image
 - add tests for previously uncovered ecr_client

FYI Error that shows when repo was not yet created:

```
status: The push refers to a repository [123.dkr.ecr.us-west-2.amazonaws.com/truth_service]
errorDetail: {"message"=>"name unknown: The repository with name 'truth_service' does not exist in the registry with id '1234'"} | error: name unknown: The repository with name 'truth_service' does not exist in the registry with id '1234'
```

### Risks
 - ecr support breaking

### DIFF
--- a/plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
+++ b/plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
@@ -3,29 +3,55 @@ require 'aws-sdk-core'
 module SamsonAwsEcr
   class Engine < Rails::Engine
     AMAZON_REGISTRY = /\A.*\.dkr.ecr.(?<region>[\w\-]+).amazonaws.com\z/
-    mattr_accessor :ecr_client
-    mattr_accessor(:credentials_expire_at) { Time.current }
 
-    initializer :ecr_client do
-      if match = AMAZON_REGISTRY.match(ENV['DOCKER_REGISTRY'])
-        SamsonAwsEcr::Engine.ecr_client = Aws::ECR::Client.new(region: match['region'])
+    class << self
+      # we make sure the repo exists so pushes do not fail
+      # - ignores if the repo already exists
+      # - ignores if the repo cannot be created due to permission problems
+      def ensure_repository(repository)
+        return unless ecr_client
+        name = repository.split('/', 2).last
+        ecr_client.describe_repositories(repository_names: [name])
+      rescue Aws::ECR::Errors::RepositoryNotFoundException
+        ecr_client.create_repository(repository_name: name)
+      rescue Aws::ECR::Errors::AccessDenied
+        Rails.logger.info("Not allowed to create or describe repositories")
+      end
+
+      # aws credentials are only valid for a limited time, so we need to refresh them before pushing
+      def refresh_credentials
+        if credentials_stale?
+          authorization_data = ecr_client.get_authorization_token.authorization_data.first
+
+          self.credentials_expire_at = authorization_data.expires_at || raise("NO EXPIRE FOUND")
+          user, pass = Base64.decode64(authorization_data.authorization_token).split(":", 2)
+          ENV['DOCKER_REGISTRY_USER'] = user
+          ENV['DOCKER_REGISTRY_PASS'] = pass
+        end
+      rescue Aws::ECR::Errors::InvalidSignatureException
+        raise Samson::Hooks::UserError, "Invalid AWS credentials"
+      end
+
+      private
+
+      attr_accessor :credentials_expire_at
+
+      def ecr_client
+        return @ecr_client if defined?(@ecr_client)
+        @ecr_client = if match = AMAZON_REGISTRY.match(Rails.application.config.samson.docker.registry)
+          Aws::ECR::Client.new(region: match['region'])
+        end
+      end
+
+      def credentials_stale?
+        ecr_client && (!credentials_expire_at || credentials_expire_at < Time.current)
       end
     end
   end
 end
 
-Samson::Hooks.callback :before_docker_build do
-  begin
-    if SamsonAwsEcr::Engine.ecr_client && SamsonAwsEcr::Engine.credentials_expire_at < Time.current
-      authorization_token = SamsonAwsEcr::Engine.ecr_client.get_authorization_token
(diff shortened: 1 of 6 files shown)


rata:
@grosser thanks! Seems fine. Have you tried pulling from a private ECR repository also? Like the docker file starts with "FROM <private ecr repo image here>"?

grosser:
I guess the weirdness would be that the FROM would be foobar and then that would not match the test/foobar that samson generates ... so you could not pull something from the base namespace when using a prefix this way ... I think I'll put the \z back ...

rata:
@grosser sorry, not sure I follow. Didn't get the prefix commit :-/

But I think it's important the ECR auth works when pulling an image (like in FROM on a Dockerfile) and when pushing. Probably you got that before, but just in case :)

grosser:
ahhhh that is a good point ... got to move that back to the before then!

grosser:
ok, back to before_build :)

rata:
@grosser cool! I'm at home now, can try to test it tomorrow (although if you tested the "FROM" and push, it should be fine for me too :))

Are you using ECR too, then? =)

grosser:
Yeah, switching to it soon since we don't want to have our own registry
anymore ...

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
```diff
@@ -3,29 +3,57 @@
...
+    class << self
+      # we make sure the repo exists so pushes do not fail
+      # - ignores if the repo already exists
+      # - ignores if the repo cannot be created due to permission problems
+      def ensure_repository(repository)
+        if ecr_client
+          ecr_client.create_repository(
+            repository_name: repository.split('/', 2).last
+          )
+        end
```
jonmoter: Could you actually do a check here if the registry already exists?

We do monitoring of our CloudTrail logs, and I think we look for failed API calls. Would be nice if we didn't have a whole bunch of false positive failures because Samson keeps trying to create a repo that already exists.

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
```diff
@@ -3,29 +3,57 @@
...
+
+      def ecr_client
+        return @ecr_client if defined?(@ecr_client)
+        @ecr_client = if match = AMAZON_REGISTRY.match(Rails.application.config.samson.docker.registry)
+          Aws::ECR::Client.new(region: match['region'])
+        end
+      end
+
+      def credentials_stale?
+        ecr_client && (!credentials_expire_at || credentials_expire_at < Time.current)
```
jonmoter: Might want to give an extra buffer of 10 minutes.  If a credentials expire at 2:00:00, and a build starts at 1:59:50, it's probably not going to get to the "push" phase until after the credentials have expired.

rata: Why not refresh the credentials before using them? This way, there is no magic number that might sometimes fail and is always refreshed when needed and not otherwise

grosser: Hmm just the overhead of doing it ... so I'd prefer 1.hour or so just to
not be wasteful :)

rata: But if you call refresh, it only refresh them when expired. If not, the same are used. So you only have the overhead of communicating with amazon when needed.

But, in any case, I'm fine with either way

grosser:
updated to do:
 - describe and then create
 - add 1 hour grace period to expiration

good to go ?

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
```diff
@@ -3,29 +3,55 @@
...
+    class << self
+      # we make sure the repo exists so pushes do not fail
+      # - ignores if the repo already exists
+      # - ignores if the repo cannot be created due to permission problems
+      def ensure_repository(repository)
+        return unless ecr_client
+        name = repository.split('/', 2).last
+        ecr_client.describe_repositories(repository_names: [name])
+      rescue Aws::ECR::Errors::RepositoryNotFoundException
+        ecr_client.create_repository(repository_name: name)
```
rata: Is it possible to have permissions to describe but not create? What will happen in that case? Or we can just document that the credentials should have permission to create?

rata:
@grosser fine by me. I did just one small comment, but if it applies **and** we want to, it can be addressed later :)

plugins/aws_ecr/lib/samson_aws_ecr/samson_plugin.rb
grosser: it is possible, but in both cases it will end up in a AccessDenied error
which is caught below

added docs for needed permissions

rata: On Tue, Jun 14, 2016 at 03:58:59PM -0700, Michael Grosser wrote:
> >  
> > -    initializer :ecr_client do
> > -      if match = AMAZON_REGISTRY.match(ENV['DOCKER_REGISTRY'])
> > -        SamsonAwsEcr::Engine.ecr_client = Aws::ECR::Client.new(region: match['region'])
> > +    class << self
> > +      # we make sure the repo exists so pushes do not fail
> > +      # - ignores if the repo already exists
> > +      # - ignores if the repo cannot be created due to permission problems
> > +      def ensure_repository(repository)
> > +        return unless ecr_client
> > +        name = repository.split('/', 2).last
> > +        ecr_client.describe_repositories(repository_names: [name])
> > +      rescue Aws::ECR::Errors::RepositoryNotFoundException
> > +        ecr_client.create_repository(repository_name: name)
> 
> it is possible, but in both cases it will end up in a AccessDenied error
> which is caught below

Ohh, I wasn't sure if in Ruby that will be caught. Great! (I will need to read
something about ruby, I can't avoid it for much longer :-D)

> added docs for needed permissions

Awesome! :)

jonmoter:
👍

PR #1067, opened 2016-06-13
allow setting a docker registry namespace

grosser:
@zendesk/paas @rata


rata:
Not sure I understand the use case. Is this to have a (registry, repo) per stage (like prod, staging, etc.) and be the repo named like: registry/stage/permalink_base?

rata:
But it seems fine! :+1:

grosser:
to have a namespace like samson or test or apps that can have finer access control / not mess with everything else

rata:
Perfect :)

jonmoter:
For example, we want to have a repositories like registry.zendesk.blah/official/* that are only writeable by Samson, but registry.zendesk.blah/dev/* that can be written to by any developer.

👍

PR #1069, opened 2016-06-14
avoid showing Bigdecimal / make output copy-pasteable

grosser:
@jonmoter 

```
Config: [{"id":166765909,"roles":[{"id":45408734,"replicas":5,"cpu":"0.5","ram":512},{"id":1064220792,"replicas":4,"cpu":"0.5","ram":1024}]}]
```


jonmoter:
👍

PR #1070, opened 2016-06-14
show nested errors to be able to debug stage

grosser:
@zendesk/samson 

screen shot 2016-06-14 at 9 59 26 am

ideally we should have infinite nesting, but this is a good start ... working on a test now ...


jonmoter:
:+1:

PR #1071, opened 2016-06-14
using the same service name on multiple projects is fine

grosser:
@zendesk/paas


jonmoter:
Wait, what?  Services need to have unique names.

Actually, now that I think about it, Deployments and DaemonSets also need to have unique names.

grosser:
damn you are right :D 

need to fix that logic the other way around ... closing this

PR #1072, opened 2016-06-14
service names need to be globally unique

grosser:
logic while creating was not checking across projects so it did not match what the validation did and lead to stage creation failing

@zendesk/paas


jonmoter:
👍

PR #1073, opened 2016-06-14
ignore soft deleted roles in a few places

grosser:
atm blocks deploys since the deploy complains that there is no deploy-group-role for a deleted role

@zendesk/paas 

 - no longer showing in the matrix
 - no longer showing in deploy group role select
 - no longer able to select projects that have deleted roles


jonmoter:
👍

PR #1074, opened 2016-06-14
use config from current cluster when on edit/update

grosser:
@zendesk/paas atm we use the latest cluster when editing, which is wrong but never blew up ... preemptively fixing this bug ...


jonmoter:
👍

PR #1075, opened 2016-06-14
do not add the secret sidecar when we do not need to resolve secrets to save resources

grosser:
@zendesk/paas


irwaters:
👍

grosser:
confirmed working :)

PR #1076, opened 2016-06-15
Support Kubernetes Job to build Docker image

hdt91:
This PR add the basic support to create a job on a K8s cluster running a predefined image. This image contains a script to clone the specified repository, commit, tag, etc. passed as environment variables, then optionally push the newly built image and reports the result log.

/cc @jonmoter @grosser 

### Tasks
- Fix CodeClimate and code style violations
- Create the build job image and push to registry
- Test the whole flow on the staging cluster
- Fix any issue after code review ?

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-141

### Risks
- Level: Low

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -85,7 +85,9 @@ class BuildsController < ApplicationController
   end
 
   def new_build_params
-    params.require(:build).permit(:git_ref, :label, :description)
+    params.require(:build).permit(
+      *[:git_ref, :label, :description] + Samson::Hooks.fire(:build_params)
+    )
   end
 
   def edit_build_params
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -2,6 +2,7 @@ require 'docker'
 
 class DockerBuilderService
   DIGEST_SHA_REGEX = /Digest:.*(sha256:[0-9a-f]+)/i
+  DOCKER_REPO_REGEX = /^BUILD DIGEST: (.*@sha256:[0-9a-f]+)/i
   include ::NewRelic::Agent::MethodTracer
 
   attr_reader :build, :execution
@@ -21,7 +22,9 @@ class DockerBuilderService
       @output = execution.output
       repository.executor = execution.executor
 
-      if build_image(tmp_dir)
+      if build.kubernetes_job
+        run_build_image_job(job, image_name, push: push, tag_as_latest: tag_as_latest)
+      elsif build_image(tmp_dir)
         push_image(image_name, tag_as_latest: tag_as_latest) if push
         build.docker_image.remove(force: true) unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
       end
@@ -30,6 +33,31 @@ class DockerBuilderService
     job_execution.on_complete { send_after_notifications }
 
     JobExecution.start_job(job_execution)
+  end
+
+  def run_build_image_job(local_job, image_name, push: false, tag_as_latest: false)
+    k8s_job = Kubernetes::BuildJobExecutor.new(output, job: local_job)
+    docker_ref = docker_image_ref(image_name, build)
+
+    success, build_log = k8s_job.execute!(build, project,
+      tag: docker_ref, push: push,
+      registry: registry_credentials, tag_as_latest: tag_as_latest)
+
+    build.docker_ref = docker_ref
+    build.docker_repo_digest = nil
+
+    if success
+      build_log.each_line do |line|
+        if (match = line[DOCKER_REPO_REGEX, 1])
+          build.docker_repo_digest = match
+        end
+      end
+    end
+    if build.docker_repo_digest.blank?
+      output.puts "### Failed to get the image digest"
+    end
+
+    build.save!
   end
 
   def build_image(tmp_dir)
@@ -52,7 +80,7 @@ class DockerBuilderService
   add_method_tracer :build_image
 
   def push_image(tag, tag_as_latest: false)
-    build.docker_ref = tag.presence || build.label.try(:parameterize).presence || 'latest'
+    build.docker_ref = docker_image_ref(tag, build)
     build.docker_repo_digest = nil
     output.puts("### Tagging and pushing Docker image to #{project.docker_repo}:#{build.docker_ref}")
 
@@ -91,6 +119,20 @@ class DockerBuilderService
     @build.project
   end
 
+  def registry_credentials
+    return nil unless ENV['DOCKER_REGISTRY'].present?
+    {
+      username: ENV['DOCKER_REGISTRY_USER'],
+      password: ENV['DOCKER_REGISTRY_PASS'],
+      email: ENV['DOCKER_REGISTRY_EMAIL'],
+      serveraddress: ENV['DOCKER_REGISTRY']
+    }
+  end
+
+  def docker_image_ref(image_name, build)
+    image_name.presence || build.label.try(:parameterize).presence || 'latest'
+  end
+
   def push_latest
     output.puts "### Pushing the 'latest' tag for this image"
     build.docker_image.tag(repo: project.docker_repo, tag: 'latest', force: true)
@@ -119,15 +161,5 @@ class DockerBuilderService
   def send_after_notifications
     Samson::Hooks.fire(:after_docker_build, build)
     SseRailsEngine.send_event('builds', type: 'finish', build: BuildSerializer.new(build, root: nil))
-  end
-
-  def registry_credentials
-    return nil unless ENV['DOCKER_REGISTRY_USER'].present? || ENV['DOCKER_REGISTRY_EMAIL'].present?
-    {
-      username: ENV['DOCKER_REGISTRY_USER'],
-      password: ENV['DOCKER_REGISTRY_PASS'],
-      email: ENV['DOCKER_REGISTRY_EMAIL'],
-      serveraddress: ENV['DOCKER_REGISTRY']
-    }
   end
 end
--- a/app/views/builds/new.html.erb
+++ b/app/views/builds/new.html.erb
@@ -33,6 +33,8 @@
         </div>
       </div>
 
+      <%= Samson::Hooks.render_views(:build_new, self, form: form) %>
+
       <div class="form-group">
         <%= form.label :build_image, "Build Docker Image?", class: "col-lg-2 control-label" %>
         <div class="col-lg-4">
--- a/plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
@@ -1,0 +1,144 @@
+# Execute a job that can build/push Docker images
+# and write progress to the local job output.
+module Kubernetes
+  class BuildJobExecutor
+    WAIT_FOR_JOB = 5.minutes
+    TICK = 3.seconds
+
+    attr_reader :job
+    delegate :id, to: :job
+
+    def initialize(output, job:)
+      @output = output
+      @job = job
+      @registry = {}
+    end
+
+    def execute!(build, project, tag:, push: false, registry:, tag_as_latest: false)
+      job_log = job_name = job_namespace = ""
+      k8s_job = nil
+
+      unless valid_registry_info?(registry)
+        @output.puts "### Registry server should not be empty. Aborting..."
+        return false, job_log
+      end
+      @registry = registry
+
+      k8s_job = job_config(build, project, tag: tag, push: push, tag_as_latest: tag_as_latest)
+      job_name = k8s_job[:metadata][:name]
+      job_namespace = k8s_job[:metadata][:namespace]
+
+      @output.puts "### Running a remote Docker build job: #{job_name}"
+      success, job_log = create_and_wait_for_job(k8s_job, @output)
+
+      message = (success ? "completed successfully" : "failed or timed out")
+      @output.puts "### Remote build job #{job_name} #{message}"
+
+      return success, job_log
+    ensure
+      # Jobs will still be there regardless of their statuses
+      begin
+        @output.puts "### Cleaning up the remote build job"
+        extension_client.delete_job(job_name, job_namespace) if k8s_job
+      rescue KubeException
+        @output.puts "### Failed to clean up the remote build job"
+      end
+    end
+
+    private
+
+    def build_cluster
+      Kubernetes::Cluster.first!
+    end
+
+    def client
+      build_cluster.client
+    end
+
+    def extension_client
+      build_cluster.extension_client
+    end
+
+    def valid_registry_info?(registry)
+      registry[:serveraddress].present?
+    end
+
+    def build_job_config_path
+      ENV['KUBE_BUILD_JOB_FILE'] || File.join(Rails.root, 'plugins', 'kubernetes', 'config', 'build_job.yml')
+    end
+
+    def fill_job_details(k8s_job, build, project, tag:, push: false, tag_as_latest: false)
+      # Fill in some information to easily query the job resource
+      project_name = project.permalink.tr('_', '-')
+      labels = { project: project_name, role: "docker-build-job" }
+      k8s_job[:metadata][:name] = "#{project_name}-docker-build-#{build.id}-#{SecureRandom.hex(7)}"
+      k8s_job[:metadata][:labels].update(labels)
+      k8s_job[:spec][:template][:metadata][:labels].update(labels)
+
+      # Necessary to query the pod running the job later
+      k8s_job[:spec][:template][:metadata][:labels][:job_name] = k8s_job[:metadata][:name]
+
+      # Pass all necessary information so that remote container can build the image
+      container_params = {
+        env: [{name: 'DOCKER_REGISTRY', value: @registry[:serveraddress] }],
+        args: [
+          project.repository_url, build.git_sha, project.docker_repo, tag,
+          push ? "yes" : "no",
+          tag_as_latest ? "yes" : "no"
+        ]
+      }
+      k8s_job[:spec][:template][:spec][:containers][0].update(container_params)
+    end
+
+    def job_config(build, project, tag:, push: false, tag_as_latest: false)
+      # Read the external config path and create a new job config instance
+      contents = File.read(build_job_config_path)
+      k8s_job = Kubernetes::RoleConfigFile.new(contents, build_job_config_path).job
+      fill_job_details(k8s_job, build, project, tag: tag, push: push, tag_as_latest: tag_as_latest)
+      k8s_job
+    end
+
+    def create_and_wait_for_job(k8s_job, output)
+      extension_client.create_job(k8s_job.deep_symbolize_keys)
+      start = Time.now
+      logs = pod_name = ""
+      job_name = k8s_job[:metadata][:name]
+      job_namespace = k8s_job[:metadata][:namespace]
+
+      # We need to loop here in case the job/pod status is not ready yet.
+      # For that the API still reports the data but the job status is
+      # neither success nor failure.
+      loop do
+        sleep TICK # Sleep a bit to wait for the server
+        begin
+          job = Kubernetes::Api::Job.new(
+            extension_client.get_job(job_name, job_namespace)
+          )
+          selector = "job_name=#{job_name}" # A build job only has one pod
+          pod = Kubernetes::Api::Pod.new(
+            client.get_pods(namespace: job_namespace, label_selector: selector).first
+          )
+          # If the pod name for the job is unchanged from previous loops, it is likely that the pod finishes
+          # but the job status is not yet updated. In this case, we will get old/duplicated logs if we start
+          # a new watch stream.
+          if pod_name != pod.name
+            pod_name = pod.name
+            client.watch_pod_log(pod.name, pod.namespace).each do |line|
+              logs << line << "\n"
+              output.puts line
+            end
+          end
+        rescue StandardError
+          # Something wrong happened. It is better to reset all memorized variables for logs and previous pod name.
+          # Might get duplicated/old logs if the job finished successfully but the failure came from other sources.
+          pod_name = logs = ""
+          output.puts "### Cannot get job #{job_name}'s pod log."
+        end
+
+        return false, "" if job.failure?
+        return false, "" if start + WAIT_FOR_JOB < Time.now
+        return true, logs if job.complete?
+      end
+    end
+  end
+end
--- a/plugins/kubernetes/config/build_job.yml
+++ b/plugins/kubernetes/config/build_job.yml
@@ -1,0 +1,38 @@
+---
+# Sample job configuration to build and optionally
+# push a Docker image. It is used in build_job_executor.rb
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: image-build-job
+  namespace: default
+  labels:
+    project: project-name
+    role: docker-build-job
+spec:
+  autoSelector: true
+  activeDeadlineSeconds: 60
+  template:
+    metadata:
+      name: image-build-job
+      labels:
+        project: project-name
+        role: docker-build-job
+    spec:
+      restartPolicy: Never
+      nodeSelector:
+        role: docker-build-job
+      containers:
+      - name: image-build-job
+        # image: docker-registry.zende.sk/repository-build:latest
+        imagePullPolicy: Always
+        command: ["/opt/build_scripts/repository-build.rb"]
+        args: []
+        volumeMounts:
+          - name: docker-sock
+            mountPath: /var/run/docker.sock
+            readOnly: true
+      volumes:
+        - name: docker-sock
+          hostPath:
+           path: /var/run/docker.sock
--- a/plugins/kubernetes/test/models/kubernetes/build_job_executor_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/build_job_executor_test.rb
@@ -1,0 +1,164 @@
+require_relative "../../test_helper"
+
+# Skip the following lines:
+# - Simple resetting tracking variables for the next loop when
+#   there is an error creating a build job or getting/watching pod log
+SingleCov.covered! uncovered: 2
+
+describe Kubernetes::BuildJobExecutor do
+  let(:output) { StringIO.new }
+  let(:out) { output.string }
+  let(:registry_info) do
+    {
+      serveraddress: Rails.application.config.samson.docker.registry,
+      username: 'foo', password: 'bar', email: 'moo@cow.com'
+    }
+  end
+
+  let(:build) { builds(:docker_build) }
+  let(:job) { jobs(:succeeded_test) }
+  let(:project) { job.project }
+
+  let(:executor) { Kubernetes::BuildJobExecutor.new(output, job: job) }
+
+  describe '#execute!' do
+    def execute!(push: false, registry_info: nil)
+      executor.execute!(build, project,
+        tag: 'latest', push: push, registry: registry_info)
+    end
+
+    let(:client) { stub }
+    let(:extension_client) { stub }
+
+    before do
+      Kubernetes::Cluster.any_instance.stubs(:client).returns client
+      Kubernetes::Cluster.any_instance.stubs(:extension_client).returns extension_client
+    end
+
+    it 'fails when an invalid registry is passed in' do
+      invalid_registry = registry_info.dup
+      invalid_registry[:serveraddress] = ''
+      extension_client.expects(:create_job).never
+      extension_client.expects(:delete_job).never
+      success, job_log = execute!(registry_info: invalid_registry)
+      refute success
+      assert_empty job_log
+    end
+
+    describe 'when there is a valid config and registry information' do
+      let(:build_job_config_path) { kubernetes_sample_file_path('kubernetes_job.yml') }
+      let(:job_config_file) { read_kubernetes_sample_file('kubernetes_role_config_file.yml') }
+      let(:labels) { {foo: 'bar'} }
+      let(:build_config) do
+        stub(job: {
+          metadata: {
+            name: 'test_job',
+            namespace: 'test_ns',
+            labels: labels
+          },
+          spec: {
+            template: {
+              metadata: {
+                name: 'test_job',
+                namespace: 'test_ns',
+                labels: labels
+              },
+              spec: {
+                containers: [{ name: 'test-job' }]
+              }
+            }
+          }
+        })
+      end
+
+      before do
+        extension_client.expects(:delete_job).once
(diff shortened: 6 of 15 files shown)


app/models/docker_builder_service.rb
```diff
@@ -21,9 +22,35 @@ def run!(image_name: nil, push: false, tag_as_latest: false)
...
+            if (match = line[DOCKER_REPO_REGEX, 1])
+              build.docker_repo_digest = match
+            end
+          end
+        end
+        if build.docker_repo_digest.blank?
+          output.puts "### Failed to get the image digest"
+        end
+
+        build.save!
```
jonmoter: Refactor this logic into a separate method, since `run!` is getting pretty big.

plugins/kubernetes/app/models/kubernetes/job_executor.rb
```diff
@@ -0,0 +1,96 @@
+# Execute a job that can build/push Docker images
+# and write progress to the local job output.
+module Kubernetes
+  class JobExecutor
```
jonmoter: Call this something like `BuildJobExecutor`, since it's specifically for executing a job that builds Docker images, rather something that executes jobs generically.

plugins/kubernetes/app/models/kubernetes/job_executor.rb
```diff
@@ -0,0 +1,96 @@
...
+    attr_reader :job
+    delegate :id, to: :job
+
+    def initialize(output, job:)
+      @output = output
+      @job = job
+    end
+
+    def build!(reference, repo:, tag:, push: false, registry:)
+      job_log = ""
```
jonmoter: Seems strange to have this write to `@output`, and also to return a string with the job output.  Why not just write all output to `@output`?

plugins/kubernetes/app/models/kubernetes/job_executor.rb
```diff
@@ -0,0 +1,96 @@
...
+        return false, job_log
+      end
+
+      contents = File.read(build_job_config_path)
+      k8s_job = Kubernetes::JobConfigFile.new(contents, build_job_config_path).job
+      k8s_job.metadata.name = "docker-build-job-#{SecureRandom.hex(16)}"
+      k8s_job.first_container.args = [@job.project.repository_url, reference, repo, tag, push ? "yes" : "no"]
+      k8s_job.first_container.env = [
+        {name: 'DOCKER_REGISTRY_USER', value: registry[:username] || ""},
+        {name: 'DOCKER_REGISTRY_PASS', value: registry[:password] || ""},
```
jonmoter: Not a good practice to put sensitive information into the ENV of an object. Should really use a secret, or put something into Vault.

Also, Kubernetes has its own mechanism to authenticate Docker with remote registries. We can probably leverage that. http://kubernetes.io/docs/user-guide/images

plugins/kubernetes/app/models/kubernetes/job_executor.rb
```diff
@@ -0,0 +1,96 @@
...
+      k8s_job = nil
+
+      if !valid_registry_info?(registry)
+        @output.puts "### Registry server should not be empty. Aborting..."
+        return false, job_log
+      end
+
+      contents = File.read(build_job_config_path)
+      k8s_job = Kubernetes::JobConfigFile.new(contents, build_job_config_path).job
+      k8s_job.metadata.name = "docker-build-job-#{SecureRandom.hex(16)}"
```
jonmoter: Might be nice to name the job based on the project and build id.  Like "docker-build-help-center-12421", so that it's more obvious what the job is for.

Also, add some labels like the name of the project, and something labeling it as a docker build job, so we can easily query on these kind of jobs.

plugins/kubernetes/app/models/kubernetes/job_executor.rb
```diff
@@ -0,0 +1,96 @@
...
+      k8s_job.first_container.env = [
+        {name: 'DOCKER_REGISTRY_USER', value: registry[:username] || ""},
+        {name: 'DOCKER_REGISTRY_PASS', value: registry[:password] || ""},
+        {name: 'DOCKER_REGISTRY_EMAIL', value: registry[:email] || ""},
+        {name: 'DOCKER_REGISTRY', value: registry[:serveraddress]}
+      ]
+
+      @output.puts "### Running a remote Docker build job: #{k8s_job.metadata.name}"
+      extension_client.create_job(k8s_job)
+      success, job_obj = wait_for_job_status(k8s_job)
```
jonmoter: Enhancement: use `watch_pod_log` and write the output to the `@output` buffer, so that the end-user gets feedback before the job completes.

plugins/kubernetes/app/models/kubernetes/job_executor.rb
hdt91: The intention is to return the specific job log to the caller so it can parse only the necessary information.

plugins/kubernetes/app/models/kubernetes/job_executor.rb
hdt91: I can remove this since it is optional. The authentication happens at the job container so just need to change the image. Is there any documentation about the recommended way that we use auth/secret in the cluster, such as just assume the cluster already has the secret object?

plugins/kubernetes/app/models/kubernetes/job_executor.rb
hdt91: Makes sense and a nice addition. Is there any guideline about the convention to attach labels to different Kubernetes resources?

hdt91:
There is another pending PR that actually is sort of a superset for this one. What is your opinion on this?
https://github.com/zendesk/samson/pull/1061

jonmoter:
I think that the other PR has some code that is more general-purpose for using Kubernetes jobs.  However, there's also a lot of copy-pasted code there that needs cleanup.

So my hunch is that we'd be best off cleaning up and merging the other PR, and then adapting your code to fit in with that.

Btw, I'm sorry it's taken so long to review and merge your code here.  I've been really heads-down trying to get the cluster running in our staging environments, wrangling chef cookbooks, etc.  So without a stable Kubernetes cluster, it's not useful to merge in code that relies on the cluster for building. (In general, this project has required a lot more DevOps work than Dev work, for better or worse.)

hdt91:
No worries, that PR actually covers a few TODO that I wanted originally but decided not to avoid spilling the story scope and explore where a job might fit in. I would prefer to have that one get merged nicely, since the adoption should be quite smooth based on what I have read. When everything is good, I will clean up this PR and ping you guys again :D

grosser:
FYI job support is ready in #1092 will hopefully be merged shorty, not sure how helpful that will be though ...

hdt91:
I have cleaned up the PR based on your recent commits and #1092. Please take a look. Btw how to view the Code Climate result?

plugins/kubernetes/app/models/kubernetes/config_file.rb
```diff
@@ -0,0 +1,33 @@
+module Kubernetes
```
grosser: this should not be necessary ...

plugins/kubernetes/app/models/kubernetes/config_verifier.rb
```diff
@@ -0,0 +1,112 @@
+module Kubernetes
+  class ConfigVerifier
```
grosser: already have this too ...

grosser:
why do we need multiple verifiers 1 should be able to handle all resource types ...

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,133 @@
...
+      registry[:serveraddress].present?
+    end
+
+    def build_job_config_path
+      ENV['KUBE_BUILD_JOB_FILE'] || File.join(Rails.root, 'plugins', 'kubernetes', 'config', 'build_job.yml')
+    end
+
+    def fill_job_details(k8s_job, build, project, tag:, push: false, tag_as_latest: false)
+      # Fill in some information to easily query the job resource
+      project_name = project.name.downcase.tr('_', '-')
```
grosser: use `.permalink` ?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,133 @@
...
+      labels = { project: project_name, role: "docker-build-job" }
+      k8s_job[:metadata][:name] = "docker-build-#{project_name}-#{SecureRandom.hex(7)}"
```
grosser: put the build id in there too ?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,133 @@
...
+      k8s_job[:metadata][:labels].update(labels)
+      k8s_job[:spec][:template][:metadata][:labels].update(labels)
+
+      # Necessary to query the pod running the job later
+      k8s_job[:spec][:template][:metadata][:labels][:jobName] = k8s_job[:metadata][:name]
```
grosser: I think so far we used underscopre labels -> job_name

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,133 @@
...
+
+      # Pass all necessary information so that remote container can build the image
+      container_params = {
+        env: [{name: 'DOCKER_REGISTRY', value: @registry[:serveraddress] }],
+        args: [
+          project.repository_url, build.git_sha, project.docker_repo,
+          tag, push ? "yes" : "no", tag_as_latest ? "yes" : "no"
```
grosser: this looks a bit crazy ... maybe parentheses can help  tag, (push ? ...), (tag_as_latest ? ...)

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,133 @@
...
+          selector = "jobName=#{job.name}" # A build job only has one pod
+          pod = Kubernetes::Api::Pod.new(
+            client.get_pods(namespace: k8s_job[:metadata][:namespace], label_selector: selector).first
+          )
+          # If the pod name for the job is unchanged from previous loops, it is likely that the pod finishes
+          # but the job status is not yet updated. In this case, we will get old/duplicated logs if we start
+          # a new watch stream.
+          if pod_name != pod.name
+            pod_name = pod.name
+            client.watch_pod_log(pod.name, pod.namespace).each do |line|
```
grosser: I like the streaming part, should give instant feedback

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,133 @@
...
+      extension_client.create_job(k8s_job.deep_symbolize_keys)
+      start = Time.now
+      logs = ""
+      pod_name = ""
+
+      loop do
+        sleep TICK # Sleep a bit to wait for the server
+        begin
+          job = Kubernetes::Api::Job.new(
+            extension_client.get_job(k8s_job[:metadata][:name], k8s_job[:metadata][:namespace])
```
grosser: maybe extract `job_name` it's used 3 times

plugins/kubernetes/config/build_job.yml
```diff
@@ -0,0 +1,34 @@
+---
```
grosser: could add a top-level comment here explaining what it does

grosser:
that looks reasonable ...

things I'd like to change:
 - extract as much view logic as possible into new plugin hooks
 - ideally reuse existing verifier/config loader
 - ideally extract remove_job logic into kubernetes ... but that might be tricky ...
 - maybe rename remote_job attribute -> kubernetes_job for clarity

hdt91:
Thanks for all those suggestions. May I know more about what is your thought on the third point?

grosser:
kubernetes related things should live in the kubernetes plugin 
 - for simpler grepping/overview
 - to not distract people working on 'regular' samson
 - to be able to remove it if kubernetes ever goes out of fashion
 - to be able to use the same hooks to plug in another container framework

hdt91:
I have reused the config reader/verifier, cleaned up the variable name and moved the view parts into kubernetes plugin directory.

It is indeed a bit tricky to move the part where the build job execution is done in the docker image service as there is a coupling in parsing the output data and we are not having the full picture of a remote job system. For now the new code chunk looks fairly isolated and can be taken out later when this local build service can be moved out as an external plugin and we have a better integration of remote job concepts.

app/controllers/builds_controller.rb
```diff
@@ -92,8 +92,18 @@ def edit_build_params
...
+      hash[p] = params[p].present?
+    end
+    hash
+  end
+
   def start_docker_build
-    DockerBuilderService.new(@build).run!(push: true)
+    docker_build_params = { push: true }
+    docker_build_params.update(extra_docker_build_params)
+    DockerBuilderService.new(@build).run!(docker_build_params)
```
grosser: this can be part of the plugin via a `build_params` hook

app/controllers/builds_controller.rb
```diff
@@ -92,8 +92,18 @@ def edit_build_params
     params.require(:build).permit(:label, :description)
   end
 
+  def extra_docker_build_params
+    hash = {}
+    Samson::Hooks.fire(:extra_build_params).each do |p|
+      hash[p] = params[p].present?
```
grosser: this sets true to some keys ... looks strange ... could merge extra_build_params ?

app/models/docker_builder_service.rb
```diff
@@ -32,6 +35,31 @@ def run!(image_name: nil, push: false, tag_as_latest: false)
     JobExecution.start_job(job_execution)
   end
 
+  def run_build_image_job(local_job, image_name, push: false, tag_as_latest: false)
```
grosser: can this code live somewhere else ? (not mix kubernetes into samson core)

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,140 @@
...
+        return false, job_log
+      end
+      @registry = registry
+
+      # Read the external config path and create a new job config instance
+      contents = File.read(build_job_config_path)
+      k8s_job = Kubernetes::RoleConfigFile.new(contents, build_job_config_path).job
+      fill_job_details(k8s_job, build, project, tag: tag, push: push, tag_as_latest: tag_as_latest)
+
+      if k8s_job
```
grosser: if these is no job file this would all blow up ... now if needed ?

also these past 5-6 lines can be a method ?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,140 @@
...
+      end
+
+      @output.puts "### Running a remote Docker build job: #{job_name}"
+      success, job_log = create_and_wait_for_job(k8s_job, @output)
+
+      if success
+        @output.puts "### Remote build job #{job_name} completed sucessfully"
+      else
+        @output.puts "### Remote build job #{job_name} failed to run or timed out"
+      end
```
grosser: little dry-up with `message = (success ? "completed successfully" : "failed or timed out")`

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,140 @@
...
+    delegate :id, to: :job
+
+    def initialize(output, job:)
+      @output = output
+      @job = job
+      @registry = {}
+    end
+
+    def execute!(build, project, tag:, push: false, registry:, tag_as_latest: false)
+      job_log = job_name = job_namespace = ""
```
grosser: will be nil anyway ... no need to assign ?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,140 @@
...
+      else
+        @output.puts "### Remote build job #{job_name} failed to run or timed out"
+      end
+
+      return success, job_log
+    ensure
+      if k8s_job
+        # Jobs will still be there regardless of their statuses
+        @output.puts "### Cleaning up the remote build job: #{job_name}"
+        extension_client.delete_job(job_name, job_namespace)
```
grosser: this wil blow up with a 404 when the job failed to create ?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+
+    def create_and_wait_for_job(k8s_job, output)
+      extension_client.create_job(k8s_job.deep_symbolize_keys)
+      start = Time.now
+      logs = pod_name = ""
+      job_name = k8s_job[:metadata][:name]
+      job_namespace = k8s_job[:metadata][:namespace]
+
+      loop do
```
grosser: does a job have multiple pods ? ... not sure what's going on here ... some docs might help :)

app/controllers/builds_controller.rb
hdt91: What do you mean by ```build_params``` hook?

app/controllers/builds_controller.rb
hdt91: This only does the minimum since for this task there is only one extra argument. If there are extra arguments the function can change to suit the need with extra conversions from string values to proper parameter values.

app/models/docker_builder_service.rb
hdt91: I would prefer to do this as well but as I said in the earlier comment, this piece of code has a coupling with ```DockerBuilderService``` in parsing the final output and to certain extent kinda of intertwine a local running job and a remote running job with logs content and job status passed around.

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
hdt91: Just my habit to declare all variables before using them to deal with code coverage or normal static analysis tool. Is this a bad thing in Ruby/Rails?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
hdt91: If there is no job file some succeeding code will likely trigger an exception and be caught in the ```ensure``` block. Is there anything wrong that I am not aware of?

What is your thought on grouping other lines of code into a method? These line look like an one-time glue code between the ```fill_job_details``` and ```create_and_wait_for_job``` to me.

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
hdt91: Yes, if it fails to create it will just throw 404 and be caught by the outer try/catch block of job executor, sending a proper message to the log. This ```ensure``` block is the best effort attempt to clean up resource remnants from the build job, what else do you think is good to add here?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
hdt91: In this context I mean the build job should only have one container for simplicity. I do not know what kind of use case we need a build job with more than one container.

app/controllers/builds_controller.rb
grosser: atm it only supports booleans which is a bit strange ...

grosser: idk ... good enough for now ...

app/views/builds/new.html.erb
```diff
@@ -40,6 +40,8 @@
         </div>
       </div>
 
+      <%= Samson::Hooks.render_views(:build_new_use_kubernetes_job, self, form: form) %>
```
grosser: hook names should be universal, something like build_new for example

app/views/builds/show.html.erb
```diff
@@ -61,6 +61,7 @@
   <% unless docker_build_running? @build %>
     <%= form_for :build, url: build_docker_image_project_build_path(@build.project, @build), html: { class: "form-inline" } do |form| %>
       <%= form.submit (job ? 'Rebuild Docker Image' : 'Build Docker Image'), class: "btn btn-primary" %>
+      <%= Samson::Hooks.render_views(:build_show_use_kubernetes_job, self, form: form) %>
```
grosser: same here -> build_show

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
grosser: nah not really just looks a bit strange ...

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
grosser: 28-35 looks like one concern to me something like 'job_config'

grosser: the File.read will already blow up, so I don't think we need a if k8s_job check ...

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
grosser: blowing up in the ensure will clobber the original error message, so I'd prefer a begin; delete_job; rescue KubeClient::Exception;end or so to silence it

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
grosser: not asking about containers, but why we need to loop here ... I was thinking we'd only have 1 pod created / don't need any looping logic

plugins/kubernetes/test/models/kubernetes/build_job_executor_test.rb
```diff
@@ -0,0 +1,152 @@
+require_relative "../../test_helper"
+
+SingleCov.covered! uncovered: 3
```
grosser: can you add a comment on why there are uncovered lines or possibly add some test to hit them ?

plugins/kubernetes/test/models/kubernetes/build_job_executor_test.rb
```diff
@@ -0,0 +1,152 @@
...
+
+        it 'job is unsuccessful' do
+          job_api_obj.stubs(:failure?).returns true
+
+          success, job_log = execute!(registry_info: registry_info)
+          refute success
+          assert_empty job_log
+        end
+
+        it 'job gets timeout' do
```
grosser: strange test case name ... how about something like `it does xyz when the job times out`

plugins/kubernetes/test/models/kubernetes/build_job_executor_test.rb
```diff
@@ -0,0 +1,152 @@
...
+          assert_equal v[:role], 'docker-build-job'
+          assert_equal v[:foo], 'bar'
+        end
+
+        assert_equal build_config.job[:spec][:template][:spec][:containers][0][:args],
+          [project.repository_url, build.git_sha, project.docker_repo, 'latest', 'no', 'no']
+        assert_equal build_config.job[:spec][:template][:spec][:containers][0][:env].length, 1
+      end
+
+      describe 'can create job' do
```
grosser: describes usually are something about a certain state like "describe 'when all params are good' do ... it 'creates a job'"

plugins/kubernetes/test/test_helper.rb
```diff
@@ -106,7 +106,11 @@ def kubernetes_fake_raw_template
     Kubernetes::ReleaseDoc.any_instance.stubs(raw_template: read_kubernetes_sample_file('kubernetes_deployment.yml'))
   end
 
+  def sample_file_path(file_name)
```
grosser: this is a global method so it needs to be namespaced to kubernetes ... kubernetes_sample_file_path

test/models/docker_builder_service_test.rb
```diff
@@ -1,6 +1,6 @@
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 19
+SingleCov.covered! uncovered: 20
```
grosser: what line did not get covered ... can we hit it somehow ?

test/models/docker_builder_service_test.rb
```diff
@@ -23,6 +23,33 @@
...
+    let(:local_job) { stub }
+    let(:k8s_job) { stub }
+    let(:repo_digest) { 'sha256:5f1d7c7381b2e45ca73216d7b06004fdb0908ed7bb8786b62f2cdfa5035fde2c' }
+    let(:build_log) { ["status: Random status", "BUILD DIGEST: #{project.docker_repo}@#{repo_digest}"].join("\n") }
+
+    before do
+      Kubernetes::BuildJobExecutor.expects(:new).returns k8s_job
+    end
+
+    it 'calls the job executor and the job completes successfully' do
```
grosser: something more descriptive like `it "does xyz when the job completes"`

test/models/docker_builder_service_test.rb
```diff
@@ -23,6 +23,33 @@
...
+      k8s_job.expects(:execute!).returns([true, build_log])
+      build.label = "Version 123"
+
+      service.run_build_image_job(local_job, nil)
+      assert_equal('version-123', build.docker_ref)
+      assert_equal("#{project.docker_repo}@#{repo_digest}", build.docker_repo_digest)
+    end
+
+    it 'calls the job executor but it is not successful' do
```
grosser: it 'does xyz when the job fails'

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
hdt91: Got it. I added a comment why it needs a loop. Sometimes the job is not yet ready and the pod log is not available so we need to retry a few seconds later.

test/models/docker_builder_service_test.rb
hdt91: The ```run!``` method's body is uncovered and originally had 19 lines. I added an if check for running the remote job and now the body has one extra line. Do you have any idea how to test the code block passed to the job execution inside this method's body?

grosser: nah ... just sanity checking whenever this increases ... you are welcome to
test it, but that can be a separate PR :)

hdt91:
I updated this PR with better test case description and other minor cleanups. Still there is a red flag in CodeClimate, do you mind checking what is the issue there?

grosser:
"Analysis results are provided for code written by licensed individuals, and Hai Dang Tran does not appear to be a member of your organization ." ... so no idea ... maybe it's just not doing the report and that's why it's red ...

app/controllers/builds_controller.rb
grosser: nvm ... build_params would be attributes for the build ... this is something else ...

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,144 @@
...
+        extension_client.delete_job(job_name, job_namespace) if k8s_job
+      rescue KubeException
+        @output.puts "### Failed to clean up the remote build job"
+      end
+    end
+
+    private
+
+    def build_cluster
+      Kubernetes::Cluster.first!
```
grosser: this most likely needs some tweeking ... something like picking a context / cluster via env vars ... but good enough for now ...

plugins/kubernetes/app/views/samson_kubernetes/_build_show.erb
```diff
@@ -0,0 +1,4 @@
+<div class="form-group">
+  <%= check_box_tag :kubernetes_job%>
+  <%= form.label :kubernetes_job, "Use Remote Kubernetes Job", class: "control-label" %>
```
grosser: should this be a column instead ? ... then we don't need extra_build_params ... just build_params and let it persist ... and no additional UI for build/show ...

grosser:
looking good ❤️

plugins/kubernetes/app/views/samson_kubernetes/_build_show.erb
hdt91: If this is persisted as a build's parameter, it becomes a part of core Samson workflow, doesn't it?

grosser: nah ... just an extra column (default false) nobody needs to know about :)

hdt91: Sounds good to me. Only one small issue is to switch between local build and remote build option without an UI. What is your suggestion?

grosser: the new build form gets checkbox ... rebuild does not really need it ... can always make a new build if someone wants to build without remote

hdt91: Updated with the new column :)

grosser:
looks good to me ... @jonmoter / @rata want to give this a sanity check before I merge ?

jonmoter:
Just did a quick review.  Nothing jumped out to me as a blocker.  Let's merge, try it out for a bit, and iterate if need be.

👍

hdt91:
Thanks for very quick feedback guys :D

app/models/docker_builder_service.rb
```diff
@@ -91,6 +119,20 @@ def project
     @build.project
   end
 
+  def registry_credentials
+    return nil unless ENV['DOCKER_REGISTRY'].present?
+    {
+      username: ENV['DOCKER_REGISTRY_USER'],
+      password: ENV['DOCKER_REGISTRY_PASS'],
+      email: ENV['DOCKER_REGISTRY_EMAIL'],
```
rata: email is deprecated and will be removed in docker 1.13 IIRC. Maybe we can remove it now?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -0,0 +1,144 @@
...
+
+    def build_job_config_path
+      ENV['KUBE_BUILD_JOB_FILE'] || File.join(Rails.root, 'plugins', 'kubernetes', 'config', 'build_job.yml')
+    end
+
+    def fill_job_details(k8s_job, build, project, tag:, push: false, tag_as_latest: false)
+      # Fill in some information to easily query the job resource
+      project_name = project.permalink.tr('_', '-')
+      labels = { project: project_name, role: "docker-build-job" }
+      k8s_job[:metadata][:name] = "#{project_name}-docker-build-#{build.id}-#{SecureRandom.hex(7)}"
```
rata: why not generateName, instead of name, that does this?

rata:
@grosser thanks, seems fine!

PR #1078, opened 2016-06-15
fix staging breaking

grosser:
@rbayerl unable to reproduce locally, but this might fix it ... test-deploying on staging atm


rbayerl:
:+1:

grosser:
confirmed working

PR #1079, opened 2016-06-16
Fix dashboards view to scroll massive tables

rikeoka:
* Fix the dashboards UI {host}/dashboards/production to scroll massive tables within the screen view.
* UI fix with no logic changes.
* Use DataGrid jQuery to pin first column and header to view
* Render page without Angular 1 API call per project was excessive

Before:
screen shot 2016-06-16 at 10 00 46 am

After:
screen shot 2016-06-17 at 2 09 47 pm

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: N/A

### Risks
- Level: Low


rikeoka:
It would be nice to get the header row and first column fixed when scrolling, but not getting any success with cloned tables first column/header row with width/height set since data's being populated from Angular.

If desired, may need to use ui-grid angular directive or another angular addon directive.

app/assets/stylesheets/global.scss
```diff
@@ -290,3 +290,10 @@ fieldset {
   height: 50px;
   border-radius: 25px;
 }
+
+/* Very large tables */
+.large-table-responsive {
+  height: 100%;
+  max-height: 65vh;
```
grosser: how about letting the browser handle the scrolling / make it as big as it wants to be ?

app/views/dashboards/show.html.erb
```diff
@@ -12,17 +12,19 @@
...
-        <th><%= link_to deploy_group.name, deploy_group %></th>
-      <% end %>
-    </tr>
-    </thead>
-    <tbody>
-      <tr ng-repeat="project in projects" class="{{ project.css.tr_class }}" style="{{ project.css.style }}" project_deploy_groups/>
-    </tbody>
-  </table>
+  <div class="large-table-responsive">
+    <table class="table">
```
grosser: can we add the class to the table instead of wrapping it ?

rikeoka: This is needed for the better fix, header that stays on screen and first column that stays in view.

grosser:
👍

app/views/dashboards/show.html.erb
grosser: ah, fancy

rikeoka:
@grosser I'll need some guidance on if I need to write tests for the javascript.  Not sure what test frameworks are used for js and if it's even possible to test repositioning.  The UI positioning works across Chrome, Firefox and Safari.  

Not sure if there's a better way to deal with the timeout hack to wait for asynchronous loading, I did with a large timeout before creating the extra tables for positioning.  I assume I could run a check on the scope.parent.parent.projects until css is present and then call the jQuery, but not sure if there are other ideas.

grosser:
why do we need all this js, can we make the table body scroll instead ?
alternatively use some jquery library ... should be tons available ...

rikeoka:
Until CSS4 or chrome turns back on support for position:sticky, Safari supports it with -webkit and Firefox supports it on certain versions, It's not possible to make the table body scroll naturally and have the first column/header stay fixed.  Also position:sticky seems to fail with the fixed first column.  When it does fail, it does so gracefully and is treated as position:relative.

Most of the jQuery plugins I tried had a bunch of bugs to get both the static header and first column to work since we use bootstrap auto width calculations.  Depending on browser there needed to be custom code to fix some display issues.  Another reason for more JS is timing issues, when the jQuery is run to clone elements of the main table to overlay, a lot of content is not loaded and having the hide projects with no deploy to have the classes and styles transfer over to the clones.

FixedTable http://www.novasoftware.com/download/jquery_fixedtable/jquery_fixedtable.aspx probably could be added in to replace the custom jQuery code, once the timing problem is solved.  If getting rid of the toggle projects with no deploys is an option, it could replace the custom jQuery easily, but all they provide is just the code with no real package.  There still is going to be a bit of JS to wait on the data to load.

These 2 I like better and has far nicer features, but it won't be really compatible with the hide projects with no deploys option, and it will require the server to produce the data in a single json file. https://github.com/paramquery/grid/ jQuery 
http://ui-grid.info/docs/#/tutorial/203_pinning ui-grid for Angular.

grosser:
how about adding a `title` with the row / column name to every cell or making a jquery thing that shows that ... so we don't have to mess with this scrolling magic ?

grosser:
I'd preferably also get rid of all that angular on this page ... no reason this cannot be a big static table ...

grosser:
atm visiting this page does 20+ api requests ... seems kind of silly inefficient ...

rikeoka:
I agree on a large static table.  It takes 1 API call per line so it actually makes a lot more than 20:)

It'll also make it a lot easier to just use an existing jQuery plugin.

rikeoka:
Made rails render the static table, and used the DataTables jQuery plugin to handle the table scrolling.  

There is a momentary off-screen presentation of the table, but once the plugin loads it looks normal

app/helpers/dashboards_helper.rb
```diff
@@ -2,4 +2,20 @@ module DashboardsHelper
   def project_has_different_deploys?(deploy_group_versions)
     deploy_group_versions.values.map(&:reference).uniq.count > 1
   end
+
+  def project_style(project_id)
```
grosser: this will be a global method in all templates, so something more meaningful / unique would be good

grosser:
👍 looks more normal, I like it :)

app/helpers/dashboards_helper.rb
rikeoka: Updated to dashboard_project_row_style to avoid name conflicts

grosser:
🚢

PR #1080, opened 2016-06-16
Redirect user to login page if not logged in

ibrahima:
This seemed to work in our samson deployment a few days ago, but currently we just experience redirect loops if no user is logged in instead of redirecting to the login page. Looking through the code, we couldn't figure out how that was ever happening so we added this. Not sure if it's necessary or what might have broken.

/cc @zendesk/samson

### Risks
- Level: Low

### DIFF
--- a/app/controllers/unauthorized_controller.rb
+++ b/app/controllers/unauthorized_controller.rb
@@ -20,7 +20,7 @@ class UnauthorizedController < ActionController::Metal
     respond_to do |format|
       format.html do
         flash[:authorization_error] = "You are not authorized to view this page."
-        redirect_back_or(root_path)
+        redirect_back_or(login_path)
       end
 
       format.json do
--- a/test/controllers/unauthorized_controller_test.rb
+++ b/test/controllers/unauthorized_controller_test.rb
@@ -23,11 +23,11 @@ describe 'Unauthorized' do
       end
 
       describe 'without a referer' do
-        it 'redirects to the root path' do
+        it 'redirects to the login path' do
           last_response.must_be(:redirect?)
 
           # Really just '/', but Rack insists on using the full SERVER_NAME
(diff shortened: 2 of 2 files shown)


grosser:
so reproduction steps would be this ?
 - login
 - go to an action that you are not authorized to do
 - get redirected to root

an integration test might be able to reproduce this ...

ibrahima:
Oh sorry, no, the reproduction steps would be

- (not logged in)
- Go to the homepage

Expected: Get redirected to login page

Actual: Get stuck in a redirect loop to the root page

grosser:
when I follow these steps I land on http://localhost:3000/login?origin=%2F
... is there anything else I need to do / setup ?
can you reproduce it locally ?

ibrahima:
The weird thing is that that worked for me before, but it broke recently after we reprovisioned the server with the newest version of samson, and we couldn't figure out where in the samson codebase it's actually redirecting to the login page. Could you point out where that's supposed to happen if you get a chance? Thanks!

grosser:
`throw(:warden) # Warden::Manager middleware catches this and calls
UnauthorizedController`

so it goes into UnauthorizedController and in there the def
redirect_back_or might be causing the trouble ...

ibrahima:
Mhm, that's what I understood so far. But to me it looks like, if you navigate to the Samson root, (eg. deploy.example.com), it'll hit UnauthorizedController, then redirect_back_or_to root_path will just send you into an infinite loop.

The nice thing about this patch is that the login path redirects to root if you are already logged in.

The weird thing is that I distinctly recall going to the root path sending me to the login page before, and nothing related to this has changed recently.

grosser:
even if we cannot reproduce the loop, this patch saves a redirect ... so let's just merge that ;)

ibrahima:
Thanks!

PR #1081, opened 2016-06-17
PAAS-210 make builds optional

grosser:
@zendesk/paas 

when a repo has no Dockerfile, allow it to be deployed, but do not try to build it.

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -57,10 +57,6 @@ class Build < ActiveRecord::Base
     @docker_image = image
   end
 
-  def file_from_repo(path)
-    project.repository.file_content path, git_sha
-  end
-
   def url
     AppRoutes.url_helpers.project_build_url(project, self)
   end
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -156,7 +156,7 @@ class JobExecution
     ActiveSupport::Notifications.instrument("execute_shell.samson", payload) do
       payload[:success] =
         if stage.try(:kubernetes)
-          @executor = Kubernetes::DeployExecutor.new(@output, job: @job)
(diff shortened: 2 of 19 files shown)


irwaters:
lgtm 👍

grosser:
local builds seems to work fine too :)

rata:
@grosser seems fine. But I'm not sure I follow. When will you mark a project to deploy with kubernetes and have no Dockerfile?
I understand that you may want to build the docker image outside samson, but this is not what this patch is doing, right?

grosser:
We have some services that will use prebuilt images where we don't have the
original source ... and this way we can deploy them the same way to all our
servers

rata:
On Sat, Jun 18, 2016 at 11:19:09AM -0700, Michael Grosser wrote:
> We have some services that will use prebuilt images where we don't have the
> original source ... and this way we can deploy them the same way to all our
> servers

Oh, cool. But this doesn't make possible to deploy a project with a dockerfile
which has been built in some other place (like travis or the devs pc), right?

Just to understand, the change seems totally fine! :)

grosser:
Nah ... but could be en enhancement down the line ... we might as well have
a 'don't build' flag on the project ... for now I think this is a nice
change anyway since it would just fail when there was no dockerfile
otherwise

rata:
On Sat, Jun 18, 2016 at 12:54:26PM -0700, Michael Grosser wrote:
> Nah ... but could be en enhancement down the line ... we might as well have
> a 'don't build' flag on the project ... for now I think this is a nice
> change anyway since it would just fail when there was no dockerfile
> otherwise

Totally agree

grosser:
update to also make builds ref and sha not-null since that is what the migration implicitly depends on anyway

PR #1082, opened 2016-06-20
revision is a sha and not a ref like master

grosser:
@zendesk/samson


app/models/job_execution.rb
```diff
@@ -194,8 +194,8 @@ def commands(dir)
       PROJECT_NAME: @job.project.name,
       PROJECT_PERMALINK: @job.project.permalink,
       PROJECT_REPOSITORY: @job.project.repository_url,
-      REVISION: @reference,
-      TAG: (@job.tag || @job.commit).to_s,
+      REVISION: @job.commit,
+      TAG: (@job.tag || @job.commit),
```
grosser: commit cannot be empty / job execution fails when trying a reference that cannot be resolved

irwaters:
👍

PR #1084, opened 2016-06-20
Add email notification to Projects#destroy

rikeoka:
Compliance team wanted an email notification when projects are deleted

* Add an email notification to Projects#destroy
* Uses PROJECT_CREATED_NOTIFY_ADDRESS for Projects#create if a separate PROJECT_DELETED_NOTIFY_ADDRESS is not setup

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:  SAMSON-253

### Risks
- Level: Low

### DIFF
--- a/app/mailers/project_mailer.rb
+++ b/app/mailers/project_mailer.rb
@@ -1,8 +1,15 @@
 class ProjectMailer < ApplicationMailer
   def created_email(user, project)
-    address = ENV['PROJECT_CREATED_NOTIFY_ADDRESS']
+    address = Rails.application.config.samson.project_created_email
     subject = "Samson Project Created: #{project.name}"
     body = "#{user.name_and_email} just created a new project #{project.name}"
+    mail(to: address, subject: subject, body: body)
+  end
+
+  def deleted_email(user, project)
+    address = Rails.application.config.samson.project_deleted_email
+    subject = "Samson Project Deleted: #{project.name}"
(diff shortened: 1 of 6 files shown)


app/mailers/project_mailer.rb
```diff
@@ -5,4 +5,11 @@ def created_email(user, project)
     body = "#{user.name_and_email} just created a new project #{project.name}"
     mail(to: address, subject: subject, body: body)
   end
+
+  def deleted_email(user, project)
+    address = ENV['PROJECT_DELETED_NOTIFY_ADDRESS'].presence || ENV['PROJECT_CREATED_NOTIFY_ADDRESS']
```
grosser: set this in config/application.rb to reuse the logic ?

test/controllers/projects_controller_test.rb
```diff
@@ -185,6 +185,12 @@
       it "sets the flash" do
         request.flash[:notice].wont_be_nil
       end
+
+      it "notifies about creation" do
```
grosser: deletion

app/mailers/project_mailer.rb
rikeoka: Moved it to config/application.rb

To be consistent also moved the ENV['PROJECT_CREATED_NOTIFY_ADDRESS'] to a config value as well.

grosser:
👍

PR #1085, opened 2016-06-21
remove old releases controller / dashboard / watchers

grosser:
@zendesk/paas 

there are more 1-off methods here and there that are unused now, but that's for next PR :)


jonmoter:
If we need the functionality in the future, we can always bring some of this back.

👍

PR #1086, opened 2016-06-21
more cleanup around old releases controller and logic

grosser:
@zendesk/paas 

also bringing us up to 100% coverage for release and release doc


plugins/kubernetes/app/decorators/admin/deploy_groups_controller_decorator.rb
```diff
@@ -1,3 +1,18 @@
 Admin::DeployGroupsController.class_eval do
-  prepend Kubernetes::DeployGroupPermittedParams
```
grosser: inlining this since it's weird to have a concern that is not reused

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -95,7 +57,8 @@ def ensure_service
     end
 
     def raw_template
-      @raw_template ||= kubernetes_release.project.repository.file_content(template_name, kubernetes_release.git_sha)
+      return @raw_template if defined?(@raw_template)
```
grosser: caching nils now

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -220,10 +184,10 @@ def parsed_config_file
     def validate_config_file
       if build && kubernetes_role
         if raw_template.blank?
-          errors.add(:build, "does not contain config file '#{template_name}'")
+          errors.add(:kubernetes_release, "does not contain config file '#{template_name}'")
```
grosser: build might not exist, so putting error on an object that always exists

irwaters:
👍

PR #1088, opened 2016-06-22
Allow creation of deployments on behalf of another user

ben:
This allows an authenticated user to create a deployment on behalf of another user. The use case here is a chat bot, where the interaction would look like this:

> [user]: @bot, deploy inbox to production
> [bot]: @user: 👍  https://samson.domain.tld/projects/foobar/deploys/2342

…and that deployment would belong to `user`, not to the robot. All existing permissions checks are obeyed, and this does not bypass the buddy check.

/cc @zendesk/samson

### Tasks
 - [x] Determine deploy creator by token if provided
 - [x] Tests
 - [ ] :+1: from team

### Risks
- Level: Low, pending review from security team

### DIFF
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -3,7 +3,7 @@ class DeploysController < ApplicationController
 
   skip_before_action :require_project, only: [:active, :active_count, :changeset]
 
-  before_action :authorize_project_deployer!, only: [:new, :create, :confirm, :buddy_check, :destroy]
+  before_action :authorize_project_deployer!, only: [:new, :confirm, :buddy_check, :destroy]
   before_action :find_deploy, except: [:index, :active, :active_count, :new, :create, :confirm, :search]
   before_action :stage, only: :new
 
@@ -94,7 +94,9 @@ class DeploysController < ApplicationController
   end
 
   def create
-    creator = params[:on_behalf] ? User.find_by_token(params[:on_behalf]) : current_user
+    puts params[:on_behalf] if params[:on_behalf].present?
+    creator = params[:on_behalf].present? ? User.find_by_token(params[:on_behalf]) : current_user
+    unauthorized! if creator.nil? || !creator.deployer_for?(current_project)
     deploy_service = DeployService.new(creator)
     @deploy = deploy_service.deploy!(stage, deploy_params)
 
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -94,7 +94,8 @@ class DeploysController < ApplicationController
   end
 
   def create
-    deploy_service = DeployService.new(current_user)
+    creator = params[:on_behalf] ? User.find_by_token(params[:on_behalf]) : current_user
+    deploy_service = DeployService.new(creator)
     @deploy = deploy_service.deploy!(stage, deploy_params)
 
     respond_to do |format|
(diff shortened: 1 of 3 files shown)


grosser:
tests or it did not happen ;)

grosser:
how is this preventing me (an evil user) from marking my deploys as yours (a nice user) ?

ben:
I'm working on the tests now, but I wanted to get feedback on the general idea before going all the way down that road.

You can't create deploys on my behalf unless I trust you with my token. Tokens are only visible in the UI to the user who owns them, and they can be reset at any time. This is also behind auth, so a non-user still can't create a deployment, even if they have a highly-privileged token.

grosser:
ah yes, that seems safe ... and then you'd give your token to the chat bot ?

ben:
Yeah, that's the idea. The first time you asked for a deploy, the bot would open up a DM with you and ask for a token, which it stores somewhere safe, and pulls it out whenever you ask it to deploy something.

grosser:
sounds good, check with compliance on that though ... they might not like a
bot deploying things / not going through google auth + 2-factor auth

jonmoter:
Yeah, we need to get compliance approval for this.

We have a bunch of controls around Samson for SOC2 reasons, including how authentication happens. For example, if an employee leaves Zendesk, their Google account is disabled, so they can't log into Samson. But I don't know whether their token gets invalidated.

/cc @joshdellinger or @cericksen

ben:
Added tests. Is there anything else I can do to find out if this is secure enough?

grosser:
wait for compliance to respond ... or hunt them down ... 
one big blocker would be that this token can be used indefinitely even if the google login no longer exists ... so not good :(

ben:
Just trying to articulate a possible attack:

1. Attacker gets credentials for a low-privileged account. *(These are protected in the same way we protect other runtime secrets)*
1. Attacker gets a high-privileged token. *(These are transmitted in a Slack DM stream, and stored in the production data environment)*
1. Attacker can then combine these to escalate create deployments at a higher level. *(This requires access to Slack)*

Regarding the token outliving the user's tenure at the company: the attacker would still need to have access to the robot, which is in Slack, which is behind Google auth with 2FA. Slack access would (I believe) turn off as soon as the Google account is deactivated.

jonmoter:
If you call the `DeploysController#create` method, and if you pass in the `on_behalf` param, it doesn't authenticate the current user.  So if you have a token, you wouldn't need access to Slack at all, or another authorized account. You'd just need to be able to reach Samson and post with that token.

As far as I can tell, there isn't any logic saying that only the Slackbot user has the ability to impersonate another user.

How are the Samson user tokens stored? By the bot?

ben:
@jonmoter, auth is still happening. I've added tests, and I've verified with Postman.

It's true that on-behalf deploys are currently available to all users. I'll add it as a user-level flag.

The user tokens will be stored by the bot, but this code hasn't been written yet. My original plan was to use Redis as a storage backend, but if there's a security concern I could redirect to MySQL or something else.

grosser:
Best wait for compliance to chime in before doing more work :)

Maybe a compromise could be to have these tokens expire after 1-x months ...

jacobat:
I would love to see something like this. Dream scenario is deployments entirely from Slack not having to open Samson at all.

jden:
As a point of comparison, this is how the Trello slackbot works, w/r/t OAuth - Trello is acting as an Identity Provider, the Trello Slack Bot is acting as an OAuth App registered with Trello, and the first time I interact with the bot, it links me to a typical OAuth flow where I as a user grant access to the bot (to certain authorization scopes). The bot then saves the token, and can act on behalf of me to the main Trello API until either 1) I revoke the token 2) the Trello API revokes the token.

In the context of this PR, Samson could act as an OAuth provider, the bot could act as an app, and the user would delegate authorization to the bot. Note that in this way, the tokens are passed directly from the provider to the app, without the user ever having to see or copy / paste tokens around: typical 3-legged oauth flow.

ben:
Closing in favor of #1130.

PR #1089, opened 2016-06-22
ignore refname is ambigouse and other warnings ... we only want the s…

grosser:
@jonmoter @jrettinghouse-zd


jrettinghouse-zd:
I'm not an SME, but this seems like it will help. 👍

rbayerl:
:+1:

PR #1093, opened 2016-06-23
show renamed files

grosser:
currently it only shows the new file name and no icon ... now it shows:

screen shot 2016-06-22 at 8 58 15 pm

@zendesk/paas


jonmoter:
👍

PR #1094, opened 2016-06-23
show repository redirects instead of failing with strange errors

grosser:
before:
comparison returns a Changeset object that has nil files/commits and causes all kinds of errors

after:
changeset causes an exception and is converted to a null-comparisson and then shows this in the UI
```
GitHub: Get https://api.github.com/repos/bar/compare/9fb1144014345bc5c4daf2fb2b87456b2d0301e9...grosser/move: 301 moved permanently // see: https://developer.github.com/v3/#http redirects
```

@zendesk/paas

### DIFF
--- a/config/initializers/octokit.rb
+++ b/config/initializers/octokit.rb
@@ -1,10 +1,27 @@
 require 'octokit'
 require 'faraday-http-cache'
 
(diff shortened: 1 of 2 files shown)


jonmoter:
👍

PR #1095, opened 2016-06-23
Set reference max width and remove old ng-controller ref

rikeoka:
* Limit width of displayed reference to 7REMS and overflow of ellipsis.  Hovering over reference should show full value
* Remove old ng-controler reference

Old Look:
screen shot 2016-06-23 at 10 42 46 am

New Look:
screen shot 2016-06-23 at 11 41 30 am

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/assets/stylesheets/dashboards.scss
+++ b/app/assets/stylesheets/dashboards.scss
@@ -1,0 +1,6 @@
+.dashboard-cell {
+  max-width: 7rem;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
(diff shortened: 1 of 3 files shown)


rikeoka:
If desired could make the width configurable so it could be set to like 8, 10, etc without a code change.

grosser:
can we truncate via css instead ? ... if it's a sha (40 characters a-f0-9 then cut down to 7 without the ...)

jonmoter:
Agreed on truncating in CSS, so we're not removing actual useful data from the page.

rikeoka:
Updated to use CSS max-width to 7rems for font size scalability. 

Also removed ng-controller reference that was causing a JS warning to show up in console.

grosser:
👍

PR #1096, opened 2016-06-23
dry up/clean up RoleConfigFile

grosser:
- make role config simpler and only return hashes (no more recursive structs)
 - ignore configs without deploys / with broken deploys or services while seeding
 - reuse role config in release-doc

@zendesk/paas


irwaters:
👍 lgtm

PR #1097, opened 2016-06-24
make sure that job.commit is always specified

zehua:
After the change in https://github.com/zendesk/samson/pull/1082, Samson fails to deploy from a commit hash, with the following error:

```
AIRBRAKE: NoMethodError - undefined method `shellescape' for nil:NilClass - /Users/zliu/src/zendesk/samson/app/models/job_execution.rb:212:in `block in commands'
/Users/zliu/src/zendesk/samson/app/models/job_execution.rb:210:in `each'
/Users/zliu/src/zendesk/samson/app/models/job_execution.rb:210:in `map'
/Users/zliu/src/zendesk/samson/app/models/job_execution.rb:210:in `commands'
/Users/zliu/src/zendesk/samson/app/models/job_execution.rb:148:in `execute!'
/Users/zliu/src/zendesk/samson/app/models/job_execution.rb:118:in `block in run!'
```

The value of `REVISION` is `nil`. It comes from `job.commit`, which happens to be `nil` if the job is created from `stage.create_deploy`.

- supply commit to job when creating a deploy for a stage
- make commit_from_git return long hash when given a hash
- only update git references of a job when either commit or tag is not nil

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -47,7 +47,7 @@ class GitRepository
 
   def commit_from_ref(git_reference, length: 7)
     return unless ensure_local_cache!
-    command = ['git', 'describe', '--long', '--tags', '--all', "--abbrev=#{length || 40}", git_reference]
+    command = ['git', 'describe', '--always', '--long', '--tags', '--all', "--abbrev=#{length || 40}", git_reference]
     return unless output = capture_stdout(*command)
     output.split('-').last.sub(/^g/, '')
   end
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -45,11 +45,10 @@ class GitRepository
   end
   add_method_tracer :clone!
 
-  def commit_from_ref(git_reference, length: 7)
+  def commit_from_ref(git_reference)
     return unless ensure_local_cache!
-    command = ['git', 'describe', '--always', '--long', '--tags', '--all', "--abbrev=#{length || 40}", git_reference]
-    return unless output = capture_stdout(*command)
-    output.split('-').last.sub(/^g/, '')
+    command = ['git', 'rev-parse', "#{git_reference}^{commit}"]
+    capture_stdout(*command)
   end
 
   def tag_from_ref(git_reference)
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -173,7 +173,7 @@ class JobExecution
       return false unless @repository.setup!(dir, @reference)
       commit = @repository.commit_from_ref(@reference, length: nil)
       tag = @repository.tag_from_ref(@reference)
-      @job.update_git_references!(commit: commit, tag: tag)
+      @job.update_git_references!(commit: commit, tag: tag) if commit || tag
     end
 
     if locked
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -171,9 +171,9 @@ class JobExecution
   def setup!(dir)
     locked = lock_project do
       return false unless @repository.setup!(dir, @reference)
-      commit = @repository.commit_from_ref(@reference, length: nil)
+      commit = @repository.commit_from_ref(@reference)
       tag = @repository.tag_from_ref(@reference)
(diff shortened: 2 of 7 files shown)


grosser:
can we rather make sure it is always set ? / make stage.create_deploy set it ?

zehua:
> can we rather make sure it is always set ? / make stage.create_deploy set it ?

yeah, that's what https://github.com/zendesk/samson/pull/1097/files#diff-0a8601dff8aa30a2fcff0101fc900fceL97 does. I think this was the only case with `job.commit` not set. All other code paths that create `job` are good.

app/models/stage.rb
```diff
@@ -94,7 +94,7 @@ def confirm_before_deploying?
 
   def create_deploy(user, attributes = {})
     deploys.create(attributes.merge(release: !no_code_deployed)) do |deploy|
-      deploy.build_job(project: project, user: user, command: script)
+      deploy.build_job(project: project, user: user, command: script, commit: deploy.reference)
```
grosser: but why is this needed if we always have a commit ?

grosser: wrongline ...

app/models/job_execution.rb
```diff
@@ -173,7 +173,7 @@ def setup!(dir)
       return false unless @repository.setup!(dir, @reference)
       commit = @repository.commit_from_ref(@reference, length: nil)
       tag = @repository.tag_from_ref(@reference)
-      @job.update_git_references!(commit: commit, tag: tag)
+      @job.update_git_references!(commit: commit, tag: tag) if commit || tag
```
grosser: but why is this needed if we always have a commit ?

grosser:
stage/git_repo changes look good, just confused by the `if commit` change

app/models/git_repository.rb
```diff
@@ -47,7 +47,7 @@ def clone!(from: repository_url, to: repo_cache_dir, mirror: false)
 
   def commit_from_ref(git_reference, length: 7)
     return unless ensure_local_cache!
-    command = ['git', 'describe', '--long', '--tags', '--all', "--abbrev=#{length || 40}", git_reference]
+    command = ['git', 'describe', '--always', '--long', '--tags', '--all', "--abbrev=#{length || 40}", git_reference]
```
zehua: This method seems to try to obtain the full commit hash. Why don't we use `git rev-parse`?

app/models/job_execution.rb
zehua: I added this before I made the change to `commit_from_ref` in an attempt to avoid overwriting `commit` to `nil`.

It might still help in case the user passes in an invalid git ref, causing `@repository.commit_from_ref` above to return `nil`.

app/models/git_repository.rb
grosser: haha yeah ... not sure why all this foobar is here, change away :D

app/models/job_execution.rb
zehua: ok, invalid commit hash will not even get to this step, as it will fail at the `checkout` step https://github.com/zendesk/samson/blob/51fc8afd30746b3dfba730431da686da9c8fda0d/app/models/git_repository.rb#L24

I will undo this change.

app/models/git_repository.rb
zehua: done

app/models/job_execution.rb
zehua: done

grosser: pusssssh it :D

grosser: great ... github shoed me 7 hour old cache :/

app/models/git_repository.rb
```diff
@@ -45,11 +45,10 @@ def clone!(from: repository_url, to: repo_cache_dir, mirror: false)
   end
   add_method_tracer :clone!
 
-  def commit_from_ref(git_reference, length: 7)
+  def commit_from_ref(git_reference)
```
grosser: good catch on the param never being used ❤️

PR #1099, opened 2016-06-26
initial work on REST API

neerfri:
Hi Guys,
Great project! Love what you've done here!

I wanted to trigger our builds using a chat bot and thought that a general api to do so would be very useful for other people as well.
This is more of a direction then a complete PR.
Would love to get feedback regarding the idea and implementation.
In case you are interested I would also make this code guarded by a feature flag [like those](https://github.com/zendesk/samson/blob/master/docs/extra_features.md) at least until we are certain about this.

I also think that in the long run it's much better to have an "API key" feature that will allow users to generate (and later revoke) API keys to be used in the REST api instead of the email-token pair.

/cc @zendesk/samson
### Tasks
- [ ] :+1: from team
### References
- Jira link:
### Risks
- Level: Low/Med/High


grosser:
how about having json responses in the normal controllers ?
... it's not very elegant, but super simple ...

might be able to just add new routes to the existing controllers and make that api/v1/deploys ...

neerfri:
@grosser Thanks for the feedback.
Using the json responses from the normal controller was my first attempt too.

After trying it out I felt it has a few drawbacks:
1. While controllers that serve the webapp can easily change (because the webapp is deployed along with them) the REST APIs cannot since the API clients are expecting a certain response. Which also means the API needs versioning (hence the `v1`).
2. The normal controllers are using cookies (session), which are not a very good idea for APIs since they make XSRF attacks possible.
3. I think it's a very good idea to move such an API to be authenticated with tokens. This makes it much easier to manage credentials with all of the integrations. This will mean further change from the normal controllers (or adding bloat to them)

I thought in this case the small code duplication could provide more benefit then harm as it helps to keep the concerns separate (as in single concern - not activesupport's concern).

What do you think?

grosser:
having the api use sessions means it's easy to debug and easy to consume from web-widgets ... cross-origin requests should not be allowed though ... 

I like the stability this would provide / seems like the right long-term direction / easier to maintain down the road ... @zendesk/samson got any feedback here ?

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Api::V1::ApplicationController < ActionController::Base
+  include CurrentUser
+  
+  def login_user
+    warden.authenticate(:basic) || unauthorized!
```
grosser: how about `ALLOWED_AUTH_METHODS = [:basic]` so it's more explicit and less copy-paste ?

neerfri:
@grosser I am cannot see the code climate errors. Do you want me to check them out (to make the build pass)?

grosser:
`Analysis results are provided for code written by licensed individuals, and Neer Friedman does not appear to be a member of your organization .`

mauahaha ... I guess we need some open-source version of CC ... either way it mostly finds duplication so I'm not worried

app/controllers/api/v1/application_controller.rb
neerfri: Not sure I understand...
Do you mean to define a way for a class that includes the `CurrentUser` concern to override the "allowed auth methods"? so there is no need to override `login_user`?

What do you think about:
```ruby
class Api::V1::ApplicationController < ActionController::Base
  include CurrentUser

  def warden_auth_strategies
    [:basic]
  end
end
```

where the default implementation of `warden_auth_strategies` will return an empty array (which means the value of `default_strategies` as it was set in `config/initializers/warden.rb` will be used)

grosser: yes!

maybe even CurrentUser looks for `ALLOWED_AUTH_METHODS` and application_controller and api controller both set them ... no defaults needed ... but must access them via `self.class.ALLOWED_AUTH_METHODS` for it to work

neerfri:
after a bit of fighting the build in Travis is passing :-)

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Api::V1::ApplicationController < ActionController::Base
+  include CurrentUser
+
+  def warden_strategies
```
grosser: should be private ... how about a constant ?

app/controllers/api/v1/deploys_controller.rb
```diff
@@ -0,0 +1,48 @@
+class Api::V1::DeploysController < Api::V1::ApplicationController
+  include CurrentProject
+
+  before_filter :authorize_project_deployer!
```
grosser: now we are copy-pasting permission logic .... also the reference logic looks very similar ...

grosser: extract a DeployControllerHelper or so to hold that logic ?

neerfri: Good point.
The `authorize_project_deployer!` could be refactored to a concern, but it uses different actions in both controllers so it might be more complicated than to explicitly call it here.

Regarding the reference logic, this is actually working differently in the main deploys controller.
In the UI you first show the changes of a branch and a warning if it was not found, but the deploys controller itself will not stop you from trying to deploy a non-existing branch (maybe this is desired...)

In the API, a request with a wrong `reference` will respond with 422.
It might be a good idea to add this error to the returned JSON (like `{ errors: { reference: ["does not exist"] } }`)

If you think we should change also the behavior of the main deploys controller to enforce the reference then I can move it into the `DeployService`.
The best way I know to prevent this kind of logic duplications is using a "form object" as the guys from [CodeClimate explain here](http://blog.codeclimate.com/blog/2012/10/17/7-ways-to-decompose-fat-activerecord-models/)
I wasn't sure if this is "inline" with the code style of the rest of the app.

WDYT?

app/controllers/api/v1/application_controller.rb
neerfri: Correct!

Regarding the constant I think that overriding a method is somewhat more mainstream way of changing behavior in ruby inheritance (and mixins).
Constants have some weird effects depending on which module you are currently at (this is why we need things like `self.class::CONSTANT`).
There's also things like the ActiveSupport hooks that try to load files to auto detect constants.
In general they also limit the way I can decide what to return to class "compile" time instead of runtime.

I'll make this method private as you suggest.

app/controllers/api/v1/deploys_controller.rb
grosser: something simple like a module should work for starters ... the only duplication would be the before_filter calls ...
the json rendering could be done by handlin all response types in that module too like request.json -> unprocessable_entity etc ... or raising a error and then using rescue_from and show the error message as json in the api ... 

for the authorize_project_deployer a `authorize_deploy_modification :create` helper could do the trick until we come up with something nicer

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V1::ApplicationController < ActionController::Base
+  include CurrentUser
+
+  private
+
+  def warden_strategies
```
grosser: a comment like this might help: `# overrides CurrentUser version to only allow basic auth`

apanzerj:
Hey,

Just a heads up, working on a similar PR but adding samson as an OAuth Provider using Doorkeeper.

PR is inbound.

ajp

grosser:
FYI we have a api now

PR #1101, opened 2016-06-27
fix reference test failing locally

grosser:
@zendesk/paas 

silently failed for me because:
 - I have gpg signing active
 - it did not raise when it failed

### DIFF
--- a/test/lib/references_service_test.rb
+++ b/test/lib/references_service_test.rb
@@ -3,26 +3,22 @@ require_relative '../test_helper'
 SingleCov.covered!
 
 describe ReferencesService do
+  include GitRepoTestHelper
+
   let!(:repository_url) do
-    tmp = Dir.mktmpdir
-    cmds = <<-SHELL
-      git init
-      git config user.email "test@example.com"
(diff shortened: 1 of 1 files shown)


test/lib/references_service_test.rb
```diff
@@ -3,28 +3,24 @@
...
-    execute_on_remote_repo(tmp, cmds)
-    tmp
+    create_repo_with_tags
+    execute_on_remote_repo("git checkout -b test_user/test_branch")
   end
 
-  let!(:project) { Project.create!(name: 'test_project', repository_url: repository_url) }
+  let!(:project) { Project.create!(name: 'test_project', repository_url: repo_temp_dir) }
 
   before do
```
irwaters: where is repo_temp_dir defined?

irwaters:
looks fine other than my question.

👍

test/lib/references_service_test.rb
grosser: comes with GitRepoTestHelper

irwaters:
thought as much, thanks.

PR #1102, opened 2016-06-27
make controllers handle creation

grosser:
@zendesk/samson 

Having controllers taking care for creation seems like a natural thing for me ... so I'd propose to remove this from the Service objects ... so all our controllers can look the same (save -> success then do sth / fail -> render) ... ideally leads to more future unification/simplification

drive-by-refactor: 
 - make command finding logic only do 1 sql query instead of 1 per command_id
 - add test for command ordering which was previously untested


jonmoter:
👍

PR #1104, opened 2016-06-27
use deploy_strategy column

grosser:
@zendesk/paas 

storing this but never using it ...


jonmoter:
How about we just remove the column?  The column was necessary when we were using `ReplicationControllers`, but isn't needed for `Deployments`.

grosser:
works for me ... less code less bugs ;)

PR #1105, opened 2016-06-27
remove deploy strategy, we do not use it

grosser:
@zendesk/paas 

alternative to https://github.com/zendesk/samson/pull/1104


irwaters:
👍

PR #1106, opened 2016-06-27
urls need to have a protocol, also use https by default and not http …

grosser:
…which is easy to intercept

/cc @zendesk/samson 

FYI @shouta-dev saw this in your PR :)


irwaters:
👍

PR #1107, opened 2016-06-27
inline MacroService

grosser:
@zendesk/paas 

lot's of ceremony without much work ... not work the overhead ...

looked a bit into the other services, but they are either very complex or used multiple times


jonmoter:
👍

PR #1108, opened 2016-06-28
Added reference to the exposed ENV vars in deploy

zehua:
After https://github.com/zendesk/samson/pull/1082, the `REVISION` is now derived from the user entered git reference, instead of just using that git reference directly. When the git reference is a branch name or a tag on a commit that has multiple tags, the derived `REVISION` would be a git commit hash, which is different from the original git reference. Because we only expose `REVISION` to the ENV during the deploy, the original git reference entered by the deployer might not be available to deploy script.

Sometimes the deploy script does need to know the original reference. Thus, this PR explicitly adds the original reference to the list of ENV Vars exposed by job execution context.

image

/cc @zendesk/samson @zendesk/vulcan @tgohn @yihongang 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Very Low

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -194,6 +194,7 @@ class JobExecution
       PROJECT_NAME: @job.project.name,
       PROJECT_PERMALINK: @job.project.permalink,
       PROJECT_REPOSITORY: @job.project.repository_url,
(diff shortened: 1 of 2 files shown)


grosser:
❤️ 👍

PR #1109, opened 2016-06-28
validate every template we read, removing overhead of checks accross …

grosser:
…the codebase

@zendesk/paas 

reading a config file now means verifying it, so we will no longer have any invalid files or need ad-hoc checks all over the place, also simplifies testing

 - enforce only known 'kind' combinations ... Deployment, DaemonSet, Deployment+Service,Job
 - enforce metadata.labels being set for everything


irwaters:
👍  on green

grosser:
also adding more detailed label errors:

```
Missing label or role for Job metadata.labels
Missing label or role for Job spec.template.metadata.labels
```

PR #1110, opened 2016-06-28
make role names valid in kubernetes / make resources have a unique name

grosser:
@zendesk/paas @rata 

atm we pick whatever name is the metadata.name ... which can lead to conflicts.
 -> use role name to generate a unique resource name (for job/deployment/daemonset)

also includes:
 - validate that all labels are strings
 - validate that all labels match dns regex
 - validate that role name matches dns regex

### Risks
 - will make all old deployments need to be manually replaced

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -216,16 +216,18 @@ ActiveRecord::Schema.define(version: 20160629232753) do
   add_index "kubernetes_releases", ["build_id"], name: "index_kubernetes_releases_on_build_id"
 
   create_table "kubernetes_roles", force: :cascade do |t|
-    t.integer  "project_id",   limit: 4,   null: false
-    t.string   "name",         limit: 255, null: false
-    t.string   "config_file",  limit: 255
-    t.string   "service_name", limit: 255
-    t.datetime "created_at",               null: false
-    t.datetime "updated_at",               null: false
+    t.integer  "project_id",    limit: 4,   null: false
+    t.string   "name",          limit: 255, null: false
+    t.string   "config_file",   limit: 255
+    t.string   "service_name",  limit: 255
+    t.datetime "created_at",                null: false
+    t.datetime "updated_at",                null: false
     t.datetime "deleted_at"
+    t.string   "resource_name", limit: 255, null: false
   end
 
   add_index "kubernetes_roles", ["project_id"], name: "index_kubernetes_roles_on_project_id", using: :btree
+  add_index "kubernetes_roles", ["resource_name", "deleted_at"], name: "index_kubernetes_roles_on_resource_name_and_deleted_at", unique: true, length: {"resource_name"=>191, "deleted_at"=>nil}, using: :btree
   add_index "kubernetes_roles", ["service_name", "deleted_at"], name: "index_kubernetes_roles_on_service_name_and_deleted_at", unique: true, length: {"service_name"=>191, "deleted_at"=>nil}, using: :btree
 
   create_table "locks", force: :cascade do |t|
--- a/plugins/kubernetes/app/controllers/kubernetes/roles_controller.rb
+++ b/plugins/kubernetes/app/controllers/kubernetes/roles_controller.rb
@@ -55,6 +55,6 @@ class Kubernetes::RolesController < ApplicationController
   end
 
   def role_params
-    params.require(:kubernetes_role).permit(:name, :config_file, :service_name)
+    params.require(:kubernetes_role).permit(:name, :config_file, :service_name, :resource_name)
   end
 end
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -44,7 +44,7 @@ module Kubernetes
         # FYI per docs it is supposed to use batch api, but extension api works
         job = Kubeclient::Job.new(resource_template)
         if deployed
-          extension_client.delete_job job.metadata.name, job.metadata.namespace
+          extension_client.delete_job kubernetes_role.resource_name, job.metadata.namespace
         end
         extension_client.create_job job
       else
@@ -115,7 +115,7 @@ module Kubernetes
     def deployed
       extension_client.send(
         "get_#{resource_template.fetch('kind').underscore}",
-        resource_template.fetch('metadata').fetch('name'),
+        kubernetes_role.resource_name,
         deploy_group.kubernetes_namespace
       )
     rescue KubeException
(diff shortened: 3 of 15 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
```diff
@@ -127,6 +128,10 @@ def set_replica_target
       template[:spec][:replicas] = @doc.replica_target if template[:kind] == 'Deployment'
     end
 
+    def set_name
+      template[:metadata][:name] = @doc.kubernetes_role.resource_name
```
rata: This changes the name used in ```deployment.spec.template.metadata.name``` in the yaml given to kubernetes? So, basically, it's override what the user configures there?

rata:
@grosser not sure what the risk is, sorry :-/. If I deploy this version, what will happen to old deploys? Will still be there and when a new deploy is done in the project, the old will be replaced?

I mean, we have samson in production and we want to upgrade to use jobs and those features. But I'm not sure I understand what is needed to be done on upgrade.

Sorry to bother with these questions :-/

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
grosser: yes!

grosser:
np ... that's exactly why I pinged you :D

Merging this PR means old deploys will get abandoned ... which is kind of
shitty ... not sure what the best workaround is ... could deploy the new
version and then take down the old deployment manually ?

jonmoter:
Make it a column in the DB?

grosser:
Yeah I guess ... safest way and then it can be messed with manually when
something goes wrong ...

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
rata: And I'm kind of lost of why this is needed. It doesn't seem nice if we change what the user already specified.

If ```deployment.spec.template.metadata.name``` is not specified (is not mandatory, in fact I don't have it in any of my projects :)), kubernetes will use something for us. And if the user specified it, I guess he/she might have a reason.

Why do we want to set this to something in particular? Sorry, I don't understand :-/

rata:
@grosser abandoned in which way? If the yaml selector still matches the old pods, they will be replaced, right? Or what do you mean with abandoned? Samson will not find the running pods and will think there are no pods running for a project?

Sorry again :-/

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
grosser: this is about metadata.name, not deployment.spec.template.metadata.name

atm if you have 2 projects that both have metadata.name=appserver or
something like that they will override each other ... so we need to enforce
some kind of uniqueness to make sure that does not happen

grosser:
there would be 1 deployment with the old selector -> old pods running
and 1 deployment with the new selector -> new pods running

the service will be unique and only route to the new pods

grosser:
I'll update this PR so it seeds the resource name on create / migrate the old ones ... also reduces risk of someone messing up the deploy by renaming a role ...

jonmoter:
The `metadata.name` of an object in Kubernetes needs to be unique in the scope of a Namespace. That's why if you have a Deployment named "myapp", it will create Pods with names like "myapp-bc31a", "myapp-872ca", etc.

We have lots of different teams that use Samson. Let's say that two different teams have a project that requires them to do a DB migration. They give the name "db-migration" to their Job, which make sense. But you can't have two Jobs exist at the same time with that name. Same possible issue with Deployments.

If Samson manages the naming of these objects, then it can be sure there aren't collisions. But you bring up a good point that this breaks backwards compatibility, since y'all have objects running without this convention.

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
rata: What is metadata.name? deployment.metadata.name? (or job.metadata.name or daemonset...)?

I mean, in terms of ```kubectl explain```, what do you refer to? Just to be sure I understand

rata:
@jonmoter you can use ```metadata.generateName``` and they will be unique

The deployment has unique name too, but just happens that different projects usually don't want to use the same deployment name. But what you say are the name of the pods of a deployments, those are created similar to ```generateName``` (they use the deployment name as prefix)

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
grosser: deployment.metadata.name / daemonset.metadata.name
<http://deployment.metadata.name/> / job.metadata.name

grosser:
not worried about the name of the pods, just the deployment/job/daemonset

rata:
@grosser exactly. Something like ```job.metadata.generateName``` can be used and might be cleaner

plugins/kubernetes/app/models/kubernetes/deploy_yaml.rb
rata: Thanks!

grosser:
but won't that always generate a new name and never clean up/replace the old jobs/deployments/daemonsets ?

rata:
@grosser yeah, for deployments you want to have the same name. So you update them. For jobs, generateName is probably better and easier, if it can be integrated nicely in samson.

But not sure if it's easy to integrate, I think there should be a way to do it and not break all the old things

grosser:
updated to now:
 - backfill roles resource_name with metadata.name from their templates ... warn about all kinds of errors and pick a unique name when there is an error
 - generate a resource name when it is not unique during seeding

rata:
@grosser awesome!
Maybe for jobs it's better to use "generateName" anyways, as that enables to have several jobs for one project. But I don't have a use case for that now, and it seems better to add it (if needed) later :)

db/schema.rb
```diff
@@ -215,16 +215,18 @@
...
+    t.string   "config_file",   limit: 255
+    t.string   "service_name",  limit: 255
+    t.datetime "created_at",                null: false
+    t.datetime "updated_at",                null: false
     t.datetime "deleted_at"
+    t.string   "resource_name", limit: 255, null: false
   end
 
   add_index "kubernetes_roles", ["project_id"], name: "index_kubernetes_roles_on_project_id", using: :btree
+  add_index "kubernetes_roles", ["resource_name"], name: "index_kubernetes_roles_on_resource_name", unique: true, using: :btree
```
jonmoter: Limit, and add `deleted_at` to index

grosser:
updated the index, good to go ?

jonmoter:
Okay, let's give it a go. @rata, you probably want to be careful about applying this change (in particular the migration) to your instance of Samson, if it's already running in production.

👍

rata:
@jonmoter thanks for the warning! I will be, then :-)

PR #1113, opened 2016-06-29
show what version the console is on and add helper for fake login

grosser:
@zendesk/samson 

```
rails c
Samson version: 4ba837f
Loading development environment (Rails 4.2.6)
>> login
=> "logged in as Michael Grosser"
>> app.get '/'
=> 200
```

### DIFF
--- a/lib/samson/console_extensions.rb
+++ b/lib/samson/console_extensions.rb
@@ -1,0 +1,19 @@
+module Samson
+  module ConsoleExtensions
+    # used to fake a login while debugging in a `rails c` console session
+    # so app.get 'http://xyz.com/protected/resource' works
+    def login(user)
+      CurrentUser.class_eval do
+        define_method :current_user do
+          user
+        end
+
+        # this would call warden and cause a redirect
+        define_method :login_user do |&block|
+          block.call
+        end
+        "logged in as #{user.name}"
+      end
+    end
+  end
+end
--- a/config/initializers/console.rb
+++ b/config/initializers/console.rb
@@ -1,0 +1,4 @@
+Rails.application.console do
+  Rails::ConsoleMethods.send(:prepend, Samson::ConsoleExtensions)
+  puts "Samson version: #{Rails.application.config.samson.revision.to_s.first(7)}"
+end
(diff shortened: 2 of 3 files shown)


lib/samson/console_extensions.rb
```diff
@@ -0,0 +1,19 @@
+module Samson
+  module ConsoleExtensions
+    def login(role = Role::SUPER_ADMIN)
```
jonmoter: How does this get called?

lib/samson/console_extensions.rb
```diff
@@ -0,0 +1,19 @@
...
+      user = User.where(role_id: role.id).last!
```
jonmoter: So you'd wind up as whatever user happened to be the last super admin that was created?  That doesn't sound right.

lib/samson/console_extensions.rb
grosser: by the user on the console

lib/samson/console_extensions.rb
grosser: it's for testing/reproducing ... alternatively create a new user, but that would also be a bit strange ...

jonmoter: Just seems weird to specify a Role, and wind up as whoever the last person with that role is.

I'd be more inclined to do something like `login_as "jmoter@zendesk.com"`, or `login_as User.find(1234)`, to set current user as the user of your choice.

jonmoter: Maybe if you just name it `login_as_role` then it would be more clear what you're doing.

And add some docs on the purpose of the method.  When I read it, I thought the idea was that if you did something like trigger a deployment, it would log the correct user as doing it.  For compliance reasons it's important that we keep track of who's doing what, so I want to be careful about mechanism that allow impersonation.

grosser: When I try to reproduce things I usually don't care what user I am ...
being able to pass specific users sounds like a nice feature though ...
How about user=User.first ... which should be a super admin and then it can
be overwritten

grosser: yeah true ... let's go with you have to give a user ... then you give
yourself and everything would be logged/executed correctly

grosser:
updated

jonmoter:
Thanks for the changes!  👍

grosser:
thanks for the feedback ;)

PR #1114, opened 2016-06-29
make lograge+syslog easily testable and not the default

grosser:
parallel to RAILS_LOG_TO_STDOUT this introduces RAILS_LOG_TO_SYSLOG ... 
when nothing is given we use the default log format 

TODO: rollout env update to set RAILS_LOG_TO_SYSLOG

@jwswj / @jonmoter  looks good ?

/cc @zendesk/samson

### DIFF
--- a/config/initializers/logging.rb
+++ b/config/initializers/logging.rb
@@ -1,0 +1,20 @@
+config = Rails.application.config
+
+if ENV["RAILS_LOG_TO_STDOUT"].present?
+  # heroku and docker: dump everything to stdout
+  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
+elsif ENV["RAILS_LOG_TO_SYSLOG"]
+  # log 1 message per request to syslog in json format
+  config.lograge.enabled = true
+
+  config.lograge.custom_options = lambda do |event|
+    # show params for every request
+    unwanted_keys = %w[format action controller]
+    params = event.payload[:params].reject { |key, _| unwanted_keys.include? key }
+    { params: params }
+  end
+
+  require 'syslog/logger'
+  config.logger = Syslog::Logger.new('samson')
+  config.lograge.formatter = Lograge::Formatters::Logstash.new
+end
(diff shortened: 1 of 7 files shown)


config/initializers/logging.rb
```diff
@@ -0,0 +1,20 @@
+config = Rails.application.config
+
+if ENV["RAILS_LOG_TO_STDOUT"].present?
+  # heroku and docker: dump everything to stdout
+  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
+elsif ENV["RAILS_LOG_TO_SYSLOG"]
```
jonmoter: Wouldn't this trigger even if you had `RAILS_LOG_TO_SYSLOG=0` in your config?

grosser: yes, but '0'.present? is also true so we should adjust this and RAILS_LOG_TO_STDOUT I guess ...

jonmoter: I want to be consistent in how we handle boolean environment variables.  In our .env.example we have some values that are 1 or 0, some that have values like "true", and apparently some that are enabled if the environment variable is set at all.  That's confusing.

This is outside the scope of this PR, but I'm bringing it up so that we can try to get some consistency.

grosser: yeah good point, saw that a few times too ... I'll add a boolean env reader
or so ...

grosser:
added env-check ... looks good ? / know more places where I should add it ?

config/initializers/logging.rb
```diff
@@ -0,0 +1,20 @@
+config = Rails.application.config
```
jonmoter: Should this file be named something like `00_logging.rb`, so it's one of the first things loaded?

If any of the other initializers log something, it might go to the wrong place.

jonmoter:
❤️  on EnvCheck.  Other than possibly renaming initializer, 👍

config/initializers/logging.rb
grosser: yeah that would help ... might have some warnings etc coming from the
initializers we'd otherwise lose

PR #1116, opened 2016-06-29
use body of cached 304 response

grosser:
atm 304 is shown as github error ... which is wrong ...

screen shot 2016-06-29 at 9 59 56 am

@jonmoter


jonmoter:
👍

PR #1117, opened 2016-06-30
rename DeployYaml to ResourceTemplate

grosser:
@zendesk/paas 

it's used for more then deploy now and it's not a yaml ... so rename :)


Gemfile.lock
```diff
@@ -228,13 +228,13 @@ GEM
       nokogiri (>= 1.5.9)
     mail (2.6.4)
       mime-types (>= 1.16, < 4)
-    maxitest (2.0.0)
-      minitest (>= 5.0.0, < 5.9.0)
+    maxitest (2.0.1)
```
grosser: random update because I saw an error ...

PR #1118, opened 2016-06-30
save generated resource template on release doc creation

grosser:
bringing back a feature we had before but that got accidentally deleted by erm someone ;)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -21,31 +25,31 @@ def client
     end
 
     def job?
-      resource_template.resource_kind == 'job'
+      resource_template.fetch('kind') == 'Job'
     end
```
jonmoter: Also have `deployment?` and `daemonset?`

jonmoter:
👍

PR #1120, opened 2016-07-01
Bring back vis timeline in deploy_groups page

neroleung:
It was not showing up after we updated vis.js in this commit - 0d7e3d63ce162fced54e0fc3d8dcafd078365852. It's probably because of their javascript or css changes in the new version. Rolling back `rails-assets-vis` to `4.10.0` fixed the issue.

screen shot 2016-06-30 at 4 59 45 pm

### Tasks
 - [x] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


grosser:
So far I never found this View helpful ... it doew Tons of requests too ...
has lots of Code ... do we need it ? Maybe remove it ?

neroleung:
Praveen seems to think this view is helpful though.. I feel like whether to take away this view or make it more efficient should be a separate concern, and can be easily done in a separate PR. So maybe treat this one as just fixing the `vis.js` issue? What do you think, @grosser?

grosser:
don't care much ... was hoping nobody actually uses it ... if someone does
then :+1:

PR #1121, opened 2016-07-01
log user changes

grosser:
@jonmoter 

not tracking token changes ... but everything else ... next PR: UserRole changes

/cc @zendesk/samson 

screen shot 2016-06-30 at 5 21 32 pm


jonmoter:
👍

PR #1122, opened 2016-07-01
audit user role changes on user

grosser:
makes them part of the users history since we don't have a real view for them anyway ... especially when they are already deleted ...

screen shot 2016-07-01 at 1 57 55 pm


jonmoter:
Rebased branch to make the diffs easier to read.

jonmoter:
👍 on 💚

PR #1124, opened 2016-07-01
delete abandomed deploy group stages

grosser:
@apanzerj new UI blew up when with deploy-group-stages without stages ...


apanzerj:
👍 on 🍏

PR #1128, opened 2016-07-07
show when automatic unlock is broken instead of showing negative time…

grosser:
screen shot 2016-07-07 at 2 08 32 pm

when lock service is busted atm it shows negative time as relative time ... which is confusing ...

@zendesk/samson

### DIFF
--- a/app/models/lock.rb
+++ b/app/models/lock.rb
@@ -20,20 +20,26 @@ class Lock < ActiveRecord::Base
     stage_id.blank?
   end
 
+  # short summary used in helpers ... keep in sync with locks/_lock.html.erb
   def summary
-    "Locked by #{locked_by} #{time_ago_in_words(created_at)} ago#{unlock_at}"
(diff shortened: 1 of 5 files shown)


app/models/lock.rb
```diff
@@ -21,15 +21,20 @@ def global?
...
+    "Locked by #{locked_by} #{time_ago_in_words(created_at)} ago#{unlock_summary}"
   end
 
   def locked_by
     (user.try(:name) || 'Unknown user').to_s
   end
 
-  def unlock_at
-    " and will unlock in #{time_ago_in_words(delete_at)}" if delete_at
+  def unlock_summary
```
grosser: also renaming since this is not a Time object

irwaters:
👍

grosser:
looked a bit further into this ... the task seems to not run every 1 minute ... but it eventually works ... not sure how to debug that further ... but at least it's obvious now ...

PR #1130, opened 2016-07-08
Slack slash-command

ben:
This is a better approach than #1088 for allowing deployments to be triggered from Slack.

Here's the setup flow:

1. A Samson admin would create a [new Slack app](https://api.slack.com/apps/new) for their team and configure it to connect to their Samson instance. This would involve mostly setting three URLs (one for OAuth, one for receiving slash-commands, and one for message interaction post-backs).
1. The first time any user invokes the `/deploy` command from Slack, the app would prompt them to authenticate the app on their user account. (The `/trello` command is a good example of this flow.)
  1. The app replies with an "only you can see this" message with a link.
  1. The user clicks the link, which takes them to a confirmation page on their Samson instance.
  1. The user confirms that they want to connect their Samson account with their Slack account, which stores their Slack user ID in their user record.

Once the setup flow is complete, the user can then do `/deploy project[/ref] to stage` to create a deployment. If a buddy is required, something like this will pop up in the chat room:

image

Once someone clicks the :+1: (or triggers a non-buddy deployment), the message content is changed to look like this:

image

And when the deployment completes, a new message is posted to the channel:

image

And if the deployment fails, it would look like this:

image

(Maybe. It's possible editing an existing message won't cause notifications to be sent out. In that case, that mention will be sent as a new message.)

It should be noted that all authorization paths would be respected regarding who can click the button, who can create deployments, etc, since Samson would map Slack users back to Samson users in order to perform actions. Read permissions get a bit muddied, since anyone in that channel can see that a deployment has happened and which PRs are going into it, but full details are still on Samson behind auth.

/cc @zendesk/samson

### Tasks
 - [x] Migration/model for storing Slack->Samson user mapping
 - [x] Implement OAuth flow
 - [x] Implement slash-command handler
 - [x] Implement interactive-button handler
 - [x] Documentation
 - [x] Tests
 - [ ] :+1: from team

### Risks
- Level: Low. All sensitive information passes through SSL and regular authenticated paths.


jacobat:
:heart: wonderful!

One other thing that would be helpful for HelpCenter is to have Samson prompt users in Slack to deploy when the "Production Build" stage completes instead of having the users request a deploy.

Context: In HelpCenter we automatically run the "Production Build" stage when there is a merge to master (actually when Travis marks it as green, but this would be nice to change). This stage produces and distributes a tarball so the "Production deploy" phase runs quickly.

ben:
Interesting, our team does something similar. Maybe a stage that's configured for Slack should always get a message to its room when a deployment finishes? The format could be the "success" or "error" cases shown above.

jacobat:
The sequence I would prefer is:

* Libo merges to master
* Production build kicks off
* Production build successfully finished
* Slack prompt: "@ libo v9563 is ready for "Production deploy", proceed?"
* Libo clicks "Yes"
* Slack prompt: "@ here, seeking approval to deploy v9563 to "Production deploy""
* Mauro clicks "Approved"
* The Production deploy stage is started

app/controllers/slack_controller.rb
```diff
@@ -0,0 +1,12 @@
+class SlackController < ApplicationController
```
grosser: should be a plugin ?

ben: Can plugins add GET/POST endpoints? We need an OAuth round-trip to map users, and two POST endpoints for the slash-command and the buttons.

ben: Whoops, never mind. Found [this](https://github.com/zendesk/samson/blob/ben/slash-command/plugins/slack_webhooks/config/routes.rb).

db/schema.rb
```diff
@@ -11,20 +11,20 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160701215002) do
```
grosser: I'd rather see a clean PR with schema.rb cleanup the mixed in with other PRs ... can you limit this to the slack changes ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,12 @@
+class SlackAppController < ApplicationController
+  skip_before_action :verify_authenticity_token
```
grosser: verify that user is a deployer ?

plugins/slack_app/app/models/slack_identifier.rb
```diff
@@ -0,0 +1,8 @@
+class SlackIdentifier < ActiveRecord::Base
+  belongs_to :user
+
+  def self.app_token
+    record = SlackIdentifier.find_by_user_id nil
```
grosser: need a user ?

plugins/slack_app/app/views/slack_app/oauth.html.erb
```diff
@@ -0,0 +1,65 @@
+<h1>Connect to Slack</h1>
+
+<p>
+  Flash an error if we get <code>?error=access_denied</code>.
+</p>
+<p>
+  If params are empty, send user to
+  <a href="https://slack.com/oauth/authorize?client_id=<%= ENV['SLACK_CLIENT_ID'] %>&redirect_uri=<%=u url_for only_path: false %>&scope=<%=u 'chat:write:bot,commands,identify' %>">
```
grosser: show error when SLACK_CLIENT_ID is not configured

plugins/slack_app/config/routes.rb
```diff
@@ -0,0 +1,3 @@
+Samson::Application .routes.draw do
+  get '/slack/oauth', to: 'slack_app#oauth'
```
grosser: maybe slack_app as route since the slack notification plugin might use slack already

plugins/slack_app/db/migrate/20160709001231_create_slack_app_table.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSlackAppTable < ActiveRecord::Migration
+  def change
+    create_table :slack_identifiers do |t|
+      t.integer :user_id
```
grosser: null: false too ?

grosser:
concept looks ok ... check with compliance that this is good to go ... looks alright too me since you can see what is deployed before approving ...

plugins/slack_app/app/controllers/slack_app_controller.rb
ben: That's coming! The CSRF check needs to be skipped here because the Slack API isn't a logged-in user.

plugins/slack_app/app/models/slack_identifier.rb
ben: Yeah, this is a bit of a hack. We need a Slack API token in order to post chat messages, and I'm storing that in this table with `user_id = nil`. I'd like this to be smooth when the Slack admin authorizes the app to the team, and not stop the whole process until she can get an ops person to change the server configuration to store this value. Is there a better way?

plugins/slack_app/config/routes.rb
ben: I checked, it's using [`/slack_webhooks`](https://github.com/zendesk/samson/blob/ben/slash-command/plugins/slack_webhooks/config/routes.rb#L2). But matching the namespace to the plugin name is probably better.

plugins/slack_app/db/migrate/20160709001231_create_slack_app_table.rb
ben: See above. I'd love to know a better way.

plugins/slack_app/app/models/slack_identifier.rb
grosser: sounds ok ... maybe document this usecase/flow a bit

plugins/slack_app/app/models/slack_identifier.rb
```diff
@@ -0,0 +1,8 @@
...
+    record.nil? ? nil : record.identifier
```
sandlerr: this can be `record.try(:identifier)` or on ruby 2.3 `record&.identifier`

sandlerr:
the other slack plugin just uses Faraday, if we will include a library it may make sense to use it in both plugins

db/schema.rb
ben: I rebased out most of these changes, leaving just the two table definitions. Is it good practice to avoid updating the schema version as well?

grosser: version updates are fine

ben:
At this point the plugin works. If you want to give it a shot, pull down this branch and follow the instructions in [the readme](https://github.com/zendesk/samson/blob/ben/slash-command/plugins/slack_app/README.md). I've connected my own app to my own Slack team, and it seems to work mostly how I expected. There are still some usability improvements that could be made, but it gets the job done.

image

Going to work on writing tests now, the coverage is pretty low.

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,149 @@
...
+    #   "user"=>{"id"=>"U0HAGH3AB", "name"=>"ben"}, <-- who clicked
+    #   "action_ts"=>"1468259216.482422",
+    #   "message_ts"=>"1468259019.000006", <-- need this to maybe update the message
+    #   "response_url"=>"https://hooks.slack.com/actions/T0HAGP0J2/58691408951/iHdjN2zjX0dZ5rXyl84Otql7"
+    # }
+    payload = JSON.parse(params[:payload])
+    check_slack_inputs!(payload)
+    render json: interact_response(payload) # This replaces the original message
+  end
+
```
rikeoka: private

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,149 @@
...
+  def interact_response(payload)
+    # This means someone clicked the "Approve" button in the channel
+    deploy = Deploy.find_by_id payload['callback_id']
+    return '_(Whoops, this deploy disappeared.)_' unless deploy
+
+    # Is this user connected?
+    buddy_identifier = SlackIdentifier.find_by_identifier payload['user']['id']
+    buddy = buddy_identifier.try(:user)
```
rikeoka: It may be better to standardize getting the current slack user as a before_action and perform validation that user exists before getting into the main logic of handling the message.

It isn't too hard to figure out what's going on, but a lot of exit points in this that may be difficult to explain.

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,149 @@
...
+    #   "response_url"=>"https://hooks.slack.com/actions/T0HAGP0J2/58691408951/iHdjN2zjX0dZ5rXyl84Otql7"
+    # }
+    payload = JSON.parse(params[:payload])
+    check_slack_inputs!(payload)
+    render json: interact_response(payload) # This replaces the original message
+  end
+
+  def check_slack_inputs!(params)
+    if params[:ssl_check]
+      render json: 'ok' if params['ssl_check']
```
zehua: Why do this when it will be replaced later by `interact_response`?

db/schema.rb
```diff
@@ -332,6 +339,13 @@
   add_index "slack_channels", ["stage_id"], name: "index_slack_channels_on_stage_id", using: :btree
   add_index "secrets", ["id"], name: "index_secrets_on_id", unique: true, length: {"id"=>191}, using: :btree
 
+  create_table "slack_identifiers", force: :cascade do |t|
+    t.integer  "user_id",    limit: 4
+    t.text     "identifier", limit: 65535, null: false
```
zehua: Should we define indices for these two columns, since we run a number of queries on either of them? Or is it magically done by rails?

plugins/slack_app/README.md
```diff
@@ -0,0 +1,11 @@
+# Slack App Plugin
+
+Here's what you need to do in order to turn on the Slack App plugin for your Samson.
+
+1. Go over to Slack and create a [new app](https://api.slack.com/apps/new). Many of the fields are completely up to you, but these are important:
+  - Grab the Client ID and secret and verification token, and record them in your environment as `SLACK_CLIENT_ID`, `SLACK_CLIENT_SECRET`, and `SLACK_VERIFICATION_TOKEN`.
```
zehua: Should we include these env var in `.env.example`?

db/schema.rb
grosser: nope, have to do that

plugins/slack_app/README.md
grosser: yes!

ben:
OK, I think I've added a reasonable amount of tests, but CC is failing on duplicated test code, and Travis is complaining about missing coverage. What do y'all think about these?

Gemfile.lock
```diff
@@ -418,6 +430,11 @@ GEM
       simplecov-html (~> 0.10.0)
     simplecov-html (0.10.0)
     single_cov (0.2.2)
+    slack-api (1.2.3)
```
grosser: hmm 7 or so gems to do a bit of api calling ... could plain faraday be enough ?

db/schema.rb
```diff
@@ -70,6 +70,15 @@
   add_index "deploy_groups_stages", ["deploy_group_id"], name: "index_deploy_groups_stages_on_deploy_group_id", using: :btree
   add_index "deploy_groups_stages", ["stage_id"], name: "index_deploy_groups_stages_on_stage_id", using: :btree
 
+  create_table "deploy_response_urls", force: :cascade do |t|
+    t.integer  "deploy_id",    limit: 4,   null: false
+    t.string   "response_url", limit: 255, null: false
+    t.datetime "created_at"
```
grosser: these should be `null: false`

plugins/slack_app/README.md
```diff
@@ -0,0 +1,11 @@
+# Slack App Plugin
+
+Here's what you need to do in order to turn on the Slack App plugin for your Samson.
```
grosser: lots of works to say `To activate:`

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
+require 'slack'
+
+class SlackAppController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: [:command, :interact]
+  skip_around_action :login_user, except: :oauth
+
+  before_action do
```
grosser: prefer named method for clarity ...

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    if params[:ssl_check]
+      render json: 'ok'
```
grosser: ... and `return` ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    end
+
+    @payload = params[:payload] ? JSON.parse(params[:payload]) : params
```
grosser: why are there 2 different ways of calling this ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    end
+  end
+
+  def oauth
+    @need_to_connect_app = @app_token.nil?
+    @need_to_connect_user = @current_user_from_slack.nil?
+
+    if params[:code]
+      # Got an OAuth code, let's fetch our tokens
+      resp = slack.oauth_access client_id: ENV['SLACK_CLIENT_ID'],
```
grosser: 2-space indent plz

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+  skip_before_action :verify_authenticity_token, only: [:command, :interact]
+  skip_around_action :login_user, except: :oauth
+
+  before_action do
+    if params[:ssl_check]
+      render json: 'ok'
+    end
+
+    @payload = params[:payload] ? JSON.parse(params[:payload]) : params
+    raise "Invalid Slack token" unless @payload['token'] == ENV['SLACK_VERIFICATION_TOKEN']
```
grosser: ENV.fetch ... or maybe a before_action that verifies ENV is setup correctly ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+
+    @current_user_from_slack = if current_user
+      SlackIdentifier.find_by_user_id(current_user).try(:user)
+    else
+      SlackIdentifier.find_by_identifier(params[:user_id]).try(:user)
+    end
+  end
+
+  def oauth
+    @need_to_connect_app = @app_token.nil?
```
grosser: how could `@app_token` ever be set here ... there is no other before_action ...

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    @need_to_connect_user = @current_user_from_slack.nil?
+
+    if params[:code]
+      # Got an OAuth code, let's fetch our tokens
+      resp = slack.oauth_access client_id: ENV['SLACK_CLIENT_ID'],
+                                client_secret: ENV['SLACK_CLIENT_SECRET'],
+                                code: params[:code],
+                                redirect_uri: url_for(only_path: false)
+      if @need_to_connect_app
+        @app_token = SlackIdentifier.create!(identifier: resp['access_token']).identifier
```
grosser: resp.fetch for faster debugging

plugins/slack_app/app/controllers/slack_app_controller.rb
grosser: resp -> response

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+      end
+      SlackIdentifier.create!(user_id: current_user.id, identifier: resp['user_id'])
+
+      redirect_to
+    end
+  end
+
+  def command
+    render json: command_response
```
grosser: feels a bit strange that this method does nothing ... maybe inline command_response ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    #   "actions"=>[
+    #     {"name"=>"one", "value"=>"one"} <-- which button the user clicked
+    #   ],
+    #   "callback_id"=>"deploy-123456", <-- maps to a deployment
+    #   "user"=>{"id"=>"U0HAGH3AB", "name"=>"ben"}, <-- who clicked
+    #   "action_ts"=>"1468259216.482422",
+    #   "message_ts"=>"1468259019.000006", <-- need this to maybe update the message
+    #   "response_url"=>"https://hooks.slack.com/actions/T0HAGP0J2/58691408951/iHdjN2zjX0dZ5rXyl84Otql7"
+    # }
+    render json: interact_response # This replaces the original message
```
grosser: same here ... does nothing ... inline interact_response ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+
+  def command_response
+    # Interesting parts of params:
+    # {
+    #   "user_id"=>"U0HAGH3AB", <-- look up Samson user using this
+    #   "command"=>"/deploy",
+    #   "text"=>"foo bar baz", <-- this is what to deploy
+    #   "response_url"=>"https://hooks.slack.com/commands/T0HAGP0J2/58604540277/g0xd4K2KOsgL9zXwR4kEc0eL"
+    #    ^^^^ This is how to respond later, we can also directly render some JSON
+    # }
```
grosser: this is nice for future diggers/bug-fixers :)

plugins/slack_app/app/controllers/slack_app_controller.rb
grosser: hmm nvm ... it uses returns for control flow ...

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    return unknown_user unless @current_user_from_slack
+
+    # Parse the command
+    projectname, branchname, stagename = @payload['text'].match(/([^\s\/]+)(?:\/(\S+))?\s*(?:to\s+(.*))?/).captures
+
+    # Sanity checks
+    begin
```
grosser: avoid exceptions as flow control ... 

```Ruby
unless project = Project.find_by_param project_name
  return xxx
end
```

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    #   "user_id"=>"U0HAGH3AB", <-- look up Samson user using this
+    #   "command"=>"/deploy",
+    #   "text"=>"foo bar baz", <-- this is what to deploy
+    #   "response_url"=>"https://hooks.slack.com/commands/T0HAGP0J2/58604540277/g0xd4K2KOsgL9zXwR4kEc0eL"
+    #    ^^^^ This is how to respond later, we can also directly render some JSON
+    # }
+    return unknown_user unless @current_user_from_slack
+
+    # Parse the command
+    projectname, branchname, stagename = @payload['text'].match(/([^\s\/]+)(?:\/(\S+))?\s*(?:to\s+(.*))?/).captures
```
grosser: project_name / stage_name ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+
+    deploy_service = DeployService.new(@current_user_from_slack)
+    deploy = deploy_service.deploy!(stage, reference: branchname || 'master')
+    DeployResponseUrl.create! deploy: deploy, response_url: @payload[:response_url]
+
+    SlackMessage.new(deploy).message_body
+  end
+
+  def interact_response
+    # This means someone clicked the "Approve" button in the channel
```
grosser: maybe I lost track of the conversation but never heard that we cleared this with compliance ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    deploy = Deploy.find_by_id @payload['callback_id']
+    return '_(Whoops, this deploy disappeared.)_' unless deploy
```
grosser: `"Unable to find deploy from callback_id"` ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    return {
+      replace_original: false,
+      text: unknown_user
+    } unless @current_user_from_slack
+
+    # Can this user approve the deployment?
+    return {
+      replace_original: false,
+      text: "You can't approve your own deploys, silly!"
+    } unless @current_user_from_slack != deploy.user
```
grosser: unless != -> if ==

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    # Is this user connected?
+    return {
+      replace_original: false,
+      text: unknown_user
+    } unless @current_user_from_slack
+
+    # Can this user approve the deployment?
+    return {
+      replace_original: false,
+      text: "You can't approve your own deploys, silly!"
```
grosser: `You cannot bypass deploys via slack or approve your own deploy.`

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    } unless @current_user_from_slack != deploy.user
+    return {
+      replace_original: false,
+      text: "It doesn't look like you have permissions to approve that deploy."
```
grosser: `It doesn't look like you` -> `You don't`

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    } unless @current_user_from_slack.deployer_for?(deploy)
+
+    # Buddy up
+    deploy.confirm_buddy!(@current_user_from_slack)
+    SlackMessage.new(deploy).message_body
+  end
+
+  def unknown_user
+    "Sorry, I don't recognize you. Perhaps you should visit " \
```
grosser: You don't have slack bla set up, please visit to xxx to connect your slack account to samson

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+    "#{url_for action: :oauth, only_path: false} " \
+    "to connect your account."
+  end
+
+  def unknown_project(project)
+    "Sorry, I don't know a project called `#{project}`. Try your project's permalink."
```
grosser: Could not find a project with permalink xxx.

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,152 @@
...
+  end
+
+  def unauthorized_deployer(project)
+    "Sorry, it doesn't look like you have permission to create a deployment" \
+    "on `#{project}`."
+  end
+
+  def slack
+    @slack_token ||= SlackIdentifier.app_token
+    Slack.configure do |config|
```
grosser: we cannot change the global token on every request ... so maybe do something like `with_slack_token do` that resets it after

plugins/slack_app/app/helpers/slack_app_helper.rb
```diff
@@ -0,0 +1,12 @@
+require 'uri'
+
+module SlackAppHelper
+  def oauth_url(scopes)
+    redir_url = URI.encode url_for(controller: :slack_app, action: :oauth, only_path: false)
```
grosser: use url helpers

grosser: use builtin .to_query
```Ruby
query = {
  redirect_uri: xxx,
  scope: yyy
}.to_query
```

plugins/slack_app/app/models/slack_identifier.rb
```diff
@@ -0,0 +1,7 @@
+class SlackIdentifier < ActiveRecord::Base
+  belongs_to :user
+
+  def self.app_token
```
grosser: maybe global_app_token or something like that ?

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
+class SlackMessage
+  def initialize(deploy)
+    @deploy = deploy
+  end
+
+  def deliver
+    url = DeployResponseUrl.find_by(deploy_id: @deploy.id).try(:response_url)
+    Faraday.new(url).post do |req|
+      req.headers['Content-Type'] = 'application/json'
+      req.body = ActiveSupport::JSON.encode message_body
```
grosser: why not use plain `JSON` ?

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
+class SlackMessage
+  def initialize(deploy)
+    @deploy = deploy
+  end
+
+  def deliver
+    url = DeployResponseUrl.find_by(deploy_id: @deploy.id).try(:response_url)
+    Faraday.new(url).post do |req|
```
grosser: Faraday.new with nil sounds like it would give weird errors ... blow up nicely instead ?

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
...
+      req.headers['Content-Type'] = 'application/json'
+      req.body = ActiveSupport::JSON.encode message_body
+    end if url.present?
```
grosser: avoid trailing ifs ... makes things harder to read imo

plugins/slack_app/app/models/slack_message.rb
grosser: req -> request

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
+class SlackMessage
+  def initialize(deploy)
+    @deploy = deploy
+  end
+
+  def deliver
+    url = DeployResponseUrl.find_by(deploy_id: @deploy.id).try(:response_url)
```
grosser: fail hard instead ? ... this should not happen ? / will be hard to debug without any feedback

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
...
+    Faraday.new(url).post do |req|
+      req.headers['Content-Type'] = 'application/json'
+      req.body = ActiveSupport::JSON.encode message_body
+    end if url.present?
+  end
+
+  def message_body
+    ret = if @deploy.waiting_for_buddy?
```
grosser: `body = `

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
...
+    end
+    ret['response_type'] = 'in_channel'
+    ret
+  end
+
+  private
+
+  def waiting_for_buddy_body
+    {
+      text: title_str,
```
grosser: title_str -> title ?

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
...
+      attachments: [
+        {
+          text: pr_str,
```
grosser: pr_str -> pr_title ?

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def succeeded_body
+    {
+      text: ':tada: ' + title_str('successfully deployed')
+    }
+  end
+
+  def title_str(ended = nil)
+    if @deploy.buddy
+      owner = SlackIdentifier.find_by_user_id(@deploy.user.id)
```
grosser: `find_by_user_id!` ?

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
...
+        'is deploying '
+      end
+    end
+
+    stage = @deploy.stage
+    str << "<#{@deploy.url}|*#{stage.project.name}* to *#{stage.name}*>."
+  end
+
+  def pr_str
+    return '' unless @deploy.changeset.pull_requests.any?
```
grosser: unless any -> if none / if empty

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
...
+    "PRs:\n" + @deploy.changeset.pull_requests.map do |pr|
+      "<#{pr.url}|##{pr.number}> - #{pr.title}"
+    end.join("\n")
+  end
+
+  def button(text, value, options = {})
+    ret = {
```
grosser: `button =` or some other meaningful name ?

plugins/slack_app/db/migrate/20160709001231_create_slack_app_table.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.text :identifier, null: false
+
+      t.timestamps
```
grosser: `null: false`

plugins/slack_app/db/migrate/20160711222417_create_deploy_response_url_table.rb
```diff
@@ -0,0 +1,10 @@
+class CreateDeployResponseUrlTable < ActiveRecord::Migration
+  def change
+    create_table :deploy_response_urls do |t|
+      t.integer :deployment_id, null: false
+      t.string :response_url, null: false
+
+      t.timestamps
```
grosser: null: false

plugins/slack_app/lib/samson_slack_app/samson_plugin.rb
```diff
@@ -0,0 +1,14 @@
+module SamsonSlackApp
+  class Engine < Rails::Engine
+  end
+end
+
+Samson::Hooks.callback :before_deploy do |deploy, _buddy|
+  Rails.logger.debug(deploy)
```
grosser: remove this ... if we want that info then the hooks.rb should send it ...

plugins/slack_app/lib/samson_slack_app/samson_plugin.rb
```diff
@@ -0,0 +1,14 @@
...
+  SlackMessage.new(deploy).deliver
+end
+
+Samson::Hooks.callback :after_deploy do |deploy, _buddy|
+  Rails.logger.debug(deploy)
```
grosser: same here

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,100 @@
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+def post_as(id, params)
```
grosser: don't define global methods

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,100 @@
...
+    user1_identifier.update_column(:user_id, user1.id)
+    user2_identifier.update_column(:user_id, user2.id)
+  end
+
+  as_a_viewer do
+    describe '#oauth' do
+      with_env SLACK_CLIENT_ID: 'client-id', SLACK_CLIENT_SECRET: 'client-secret' do
+        it 'sends the user to Slack to connect the app' do
+          get :oauth
+          assert_match 'https://slack.com/oauth/authorize?client_id=', @response.body
```
grosser: `@response.body.must_include`

grosser: assert_match produces nasty output when it fails ... include is much nicer there

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,100 @@
...
+          assert_match '/slack_app/oauth', @response.body
+          assert_match 'chat:write:bot,commands,identify', @response.body
+        end
+
+        it 'accepts an app token from Slack' do
+          stub_request(:post, "https://slack.com/api/oauth.access").
```
grosser: put this in a `when something access bla` and but the stub in the before ?

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,100 @@
...
+            to_return(status: 200, body: '{"access_token": "iamatoken", "user_id": "Uabcdef"}')
+          get :oauth, code: 'iamaslackcode'
+          assert SlackIdentifier.app_token
+          identifier = SlackIdentifier.find_by_user_id users(:viewer).id
+          assert_equal identifier.identifier, 'Uabcdef'
+        end
+
+        it 'accepts a user token from Slack' do
+          SlackIdentifier.create! identifier: 'xoxz-i-am-an-app-token'
```
grosser: easier to read without the added xoxz

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,100 @@
...
+          identifier = SlackIdentifier.find_by_user_id users(:viewer).id
+          assert_equal identifier.identifier, 'Udefghi'
+        end
+      end
+    end
+  end
+
+  describe '#command' do
+    it "raises if secret token doesn't match" do
+      assert_raises RuntimeError do
```
grosser: errors can happen for all kinds of reasons ... make sure it's for the right one :)
e = assert_raises
e.message.must_include "xyz"

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,100 @@
...
+        post :command, token: 'thiswontmatch'
+      end
+    end
+
+    describe 'without Slack linkage' do
+      it "returns a private error if the user isn't matched up" do
+        post :command, user_id: 'notconnected', token: nil
+        assert "don't recognize you".in? @response.body
```
grosser: this would be a pretty unreadable error message when it fails -> `@response.body.must_include "don't recognize you"`

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    describe 'with Slack linkage' do
+      before do
+        stub_request(:get, "https://api.github.com/repos/bar/foo/branches/master").
+          to_return(status: 200, body: "", headers: {})
+      end
+
+      it 'checks the verification token' do
+        assert_raises do
```
grosser: e.message.must_include 'xyz' just to be sure it's the right one ?

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,100 @@
...
+          post :command, token: 'wups'
+        end
+      end
+
+      it 'succeeds on SSL check' do
+        post :command, ssl_check: true
+        assert @response
```
grosser: there is always a response .. `assert_response :success`

plugins/slack_app/test/models/slack_message_test.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:user1) { users(:deployer) }
+  let(:user2) { users(:deployer_buddy) }
+  let(:user1_identifier) { slack_identifiers(:user1) }
+  let(:user2_identifier) { slack_identifiers(:user2) }
+  before do
+    user1_identifier.update_column(:user_id, deploy.user.id)
+    user2_identifier.update_column(:user_id, user2.id)
+  end
+
+  describe 'a deploy waiting for a buddy' do
```
grosser: this should all sit in `describe "#message_body"` ... and then also have a test for deliver ?

plugins/slack_app/test/samson_slack_app/samson_plugin_test.rb
```diff
@@ -0,0 +1,42 @@
...
+      describe :after_deploy do
+        it "sends notification on after hook" do
+          SlackMessage.any_instance.expects(:deliver)
+          Samson::Hooks.fire(:after_deploy, deploy, nil)
+        end
+      end
+    end
+  end
+
+  describe 'while not configured' do
```
grosser: prefer 1 test group per thing so it's easier to later go in and ignore everything expect what you are working on

```Ruby
describe :before_deploy
  it x
  it y
end

describe :after_deploy
  it x
  it y
end
```

grosser:
these files are not covered ... so add tests ?
```
slack_app_controller.rb new uncovered lines introduced (26 current vs 0 configured)
plugins/slack_app/app/controllers/slack_app_controller.rb:16
...
plugins/slack_app/app/models/slack_identifier.rb new uncovered lines introduced (1 current vs 0 configured)
...
plugins/slack_app/app/models/slack_message.rb new uncovered lines introduced (4 current vs 0 configured)
...
```

grosser:
looks good

Gemfile.lock
ben: Good point. Removed.

grosser: :heart:

db/schema.rb
ben: These were generated by `t:timestamps` in [the migration](https://github.com/zendesk/samson/blob/ben/slash-command/plugins/slack_app/db/migrate/20160711222417_create_deploy_response_url_table.rb#L7). Is it standard to override this once it's generated?

plugins/slack_app/app/controllers/slack_app_controller.rb
ben: Weirdness in the Slack API. The `command` endpoint is called with the fields we want right in the parameters, but the `interact` endpoint has a double-stringified JSON payload in the `payload` field of the main body. Not sure why they did this, but I didn't make the world.

plugins/slack_app/app/controllers/slack_app_controller.rb
ben: Fixed.

plugins/slack_app/app/controllers/slack_app_controller.rb
ben: Yes, that conversation is still happening. Until then I'm not going to rip it out.

plugins/slack_app/app/controllers/slack_app_controller.rb
ben: Well, this goes straight to Slack, so something with a bit of personality seemed in order. The average user will have _zero_ idea how this slash-command stuff works.

plugins/slack_app/app/controllers/slack_app_controller.rb
ben: Ah, I see. You want to stomp out all the funny business. Fine, have it your way. 😀

plugins/slack_app/app/controllers/slack_app_controller.rb
ben: This went away with the switch to Faraday.

plugins/slack_app/app/helpers/slack_app_helper.rb
ben: Nice tip on `to_query`. Can you help me out with the routing stuff? These aren't mounted as resources, so I'm not sure how to get to `/slack_app/oauth` in a more succinct way (fairly new to Rails here).

plugins/slack_app/app/models/slack_message.rb
ben: In short: cargo culting. Didn't discover `unparse` until after I wrote this.

plugins/slack_app/app/models/slack_message.rb
ben: Only deploys triggered by the slash-command will have an entry in this table, so we should fail softly if we don't have a place to send the message. This seemed like a reasonable way to deal with that situation.

plugins/slack_app/app/models/slack_message.rb
ben: Hmm. Here's what it looks like in Slack if one of the `command` or `interact` endpoints throws:

image

I don't really like that experience. And I'm not sure how to replace it, so I'm doing a lot of error checking.

plugins/slack_app/db/migrate/20160711222417_create_deploy_response_url_table.rb
ben: Ah, I see. Disregard previous.

db/schema.rb
grosser: you can change the migration and schema ... ideally change the migration
and run VERSION=1211221212112 rake db:migrate:down db:migrate:up

plugins/slack_app/app/controllers/slack_app_controller.rb
grosser: Funny is nice ... but I'd prefer people understanding why it does not work
:D

plugins/slack_app/app/helpers/slack_app_helper.rb
grosser: arg yes ... that wont work ... need to use url_for

plugins/slack_app/app/models/slack_message.rb
grosser: maybe leave that as comment in the file

plugins/slack_app/app/models/slack_message.rb
grosser: yeah that's ugly :D
... maybe there is something nice around displaying json errors that we can
do there .... good enough for now ...

plugins/slack_app/app/controllers/slack_app_controller.rb
rikeoka: This isn't cleared with compliance for use in our instance yet, but I think people wanted to move forward with the PR.  Before we enable the plugin for our instance it'll have to be cleared with compliance.  Having the PR cleaned up will make it easier to reviews when we go through and clear it or add some security requirements.

plugins/slack_app/app/models/slack_message.rb
```diff
@@ -0,0 +1,116 @@
...
+      ]
+    }
+  end
+
+  def running_body
+    {
+      text: title_str,
+      attachments: [
+        {
+          text: pr_str + "\nDeploying…",
```
zehua: When approving a deploy, we usually look at the change set and risks. It will be good to include the risks in the summary directly on Slack. If displaying the full risk section is too verbose, we could at least show how many risk items are included in this deploy with a link to point to the section.

plugins/slack_app/test/samson_slack_app/samson_plugin_test.rb
ben: Yeah, that did improve readability.

plugins/slack_app/app/models/slack_message.rb
ben: **Great** idea.

ben:
At @zehua's suggestion, I've changed the output to include risks as well. It looks something like this:

image

ben:
Full coverage with passing tests. Thanks for the great review, @grosser!

grosser:
hehe, thx  for fixing!
80 commits ... a squash would be nice :)

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    # Sanity checks
+    unless project = Project.find_by_permalink(project_name)
+      return unknown_project(project_name)
+    end
+
+    return unauthorized_deployer(project) unless @current_user_from_slack.deployer_for?(project)
+
+    begin
+      stage = project.stages.find_by_param!(stage_name || 'production')
```
grosser: can use find_by_param and avoid the raise/rescue

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,158 @@
...
+    "<#{project_path(project)}|#{project.name}> " \
+    "doesn't have a stage named `#{stage_name}`."
+  end
+
+  def unauthorized_deployer(project)
+    "Sorry, it doesn't look like you have permission to create a deployment on `#{project}`."
+  end
+
+  def verify_environment
+    raise "Invalid configuration" unless ENV['SLACK_VERIFICATION_TOKEN'].present?
```
grosser: maybe `raise "SLACK_VERIFICATION_TOKEN is not set"` to make it more obvious what needs fixing

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,158 @@
...
+  end
+
+  def handle_slack_inputs
+    if params[:ssl_check]
+      # Slack does this when you're configuring your application
+      render json: 'ok'
+      return false
```
grosser: I think `return false` is no longer a thing ... a simple `return` should have the same effect

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,158 @@
+class SlackAppController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: [:command, :interact]
+  skip_around_action :login_user, except: :oauth
+
+  before_action :handle_slack_inputs
+  before_action :verify_environment, except: [:oauth]
```
grosser: this seems weird ... handle_slack_inputs needs SLACK_VERIFICATION_TOKEN but verify_environment is not called for oauth ... best to make the usage and check be part of the same filter ?

plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -0,0 +1,158 @@
...
+  def handle_slack_inputs
+    if params[:ssl_check]
+      # Slack does this when you're configuring your application
+      render json: 'ok'
+      return false
+    end
+
+    @payload = params[:payload] ? JSON.parse(params[:payload]) : params
+    token = @payload['token']
+    raise "Invalid Slack token" if token && token != ENV['SLACK_VERIFICATION_TOKEN']
```
grosser: should do constant time comparison here otherwise attackers can guess your token ... (unlikely but good practice)
http://apidock.com/rails/ActiveSupport/SecurityUtils/secure_compare/class

plugins/slack_app/app/helpers/slack_app_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+    query = {
+      client_id: ENV['SLACK_CLIENT_ID'],
```
grosser: .fetch so this blows up early ?

plugins/slack_app/app/helpers/slack_app_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+      redirect_uri: redir_url,
```
grosser: redir_url  -> redirect_url

plugins/slack_app/app/helpers/slack_app_helper.rb
grosser: .to_query will handle the encoding ... atm this will be double encoded ... are the tests not catching that ?

plugins/slack_app/app/models/slack_identifier.rb
grosser: ^

plugins/slack_app/lib/samson_slack_app/samson_plugin.rb
```diff
@@ -0,0 +1,12 @@
...
+  class Engine < Rails::Engine
+  end
+end
+
+Samson::Hooks.callback :before_deploy do |deploy, _buddy|
+  SlackMessage.new(deploy).deliver
+end
+
+Samson::Hooks.callback :after_deploy do |deploy, _buddy|
+  SlackMessage.new(deploy).deliver
```
grosser: could make this a lambda to dry this up ...

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,167 @@
...
+    opts = params.merge(
+      token: 'token',
+      user_id: id.identifier,
+      response_url: 'http://example.com/blah'
+    )
+    with_env(SLACK_VERIFICATION_TOKEN: 'token') { post method, opts }
+  end
+
+  as_a_viewer do
+    describe :oauth do
```
grosser: prefer '#auth' for consistency

plugins/slack_app/test/controllers/slack_app_controller_test.rb
```diff
@@ -0,0 +1,167 @@
...
+        post_as :command, deployer_identifier, text: project.permalink
+        first_attachment = body['attachments'][0]
+        first_attachment['fields'][0]['value'].must_include '#123'
+        first_attachment['fields'][0]['value'].must_include 'sample PR'
+      end
+    end
+  end
+
+  describe :interact do
+    # before { deploy.stubs(:waiting_for_buddy?).returns(true) }
```
grosser: delete ?

plugins/slack_app/app/helpers/slack_app_helper.rb
```diff
@@ -0,0 +1,13 @@
+require 'uri'
+
+module SlackAppHelper
+  def oauth_url(scopes)
```
grosser: this needs to be renamed ... helpers are all included into the view -> global -> slack_oauth_url

plugins/slack_app/test/helpers/slack_app_helper_test.rb
```diff
@@ -0,0 +1,13 @@
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+describe SlackAppHelper do
+  it "gives a proper OAuth URL with scopes" do
```
grosser: `describe '#oauth_url' do` for consistency

plugins/slack_app/test/models/slack_message_test.rb
```diff
@@ -0,0 +1,151 @@
...
+  let(:deployer_buddy) { users(:deployer_buddy) }
+  let(:deployer_identifier) { slack_identifiers(:deployer) }
+  let(:deployer_buddy_identifier) { slack_identifiers(:deployer_buddy) }
+  before do
+    deployer_identifier.update_column(:user_id, deploy.user.id)
+    deployer_buddy_identifier.update_column(:user_id, deployer_buddy.id)
+    deploy.changeset.stubs(:pull_requests).returns([])
+  end
+
+  describe :message_body do
```
grosser: `describe '#message_body'`

plugins/slack_app/test/models/slack_message_test.rb
```diff
@@ -0,0 +1,151 @@
...
+            )
+          ]
+        )
+      end
+
+      before do
+        deploy.stubs(:waiting_for_buddy?).returns(true)
+      end
+
+      it 'includes a 👍 button' do
```
grosser: ❤️  this shows up as emoji :D

test/fixtures/slack_identifiers.yml
```diff
@@ -0,0 +1,11 @@
+deployer:
+  identifier: 'U1234567890'
```
grosser: maybe `ident-deployer` or something like that to make it easier to decypher what is going on in tests ?

plugins/slack_app/test/models/slack_message_test.rb
ben: UTF-8 ftw 😀

plugins/slack_app/app/controllers/slack_app_controller.rb
ben: Ha. You suggested the opposite earlier. 😀 I'll fix.

grosser: Don't listen to that other panda ...

test/fixtures/slack_identifiers.yml
ben: These are referenced like `slack_identifiers(:deployer)`, which seems pretty clear to me. Is there another form that I haven't seen?

grosser: what I meant is the U1234567890 -> ident-deployer so the tests with hardcoded values are more readable

ben:
Squashed into one commit and tried to address feedback.

test/fixtures/slack_identifiers.yml
ben: Ah, I see. I can do that.

rikeoka:
@grosser, it looks like this is configurable so I'll leave it up to the core team to decide whether this should be merged into master or not.  From a compliance standpoint we can't enable the config yet to turn the feature on, but this is a standalone plugin that doesn't appear to break the main app if the plugin is disabled.

@ben Not sure if you had a chance to do the follow-ups from the last review, not going to post the list here since this is public.  

Looking at the pr, curious how the slack_identifiers.yml fixture is tied to the user for the controller tests.  Wanted to see if the slack_identifiers(:deployer) is using the users(:deployer) or users(:project_deployer) user.

Also where current_user gets populated from in #handle_slack_inputs in the controller.

ben:
@rikeoka, I'm working on the compliance pieces, just wait. To your other questions:

- I tried to use the same names in the [slack identifiers fixtures](https://github.com/zendesk/samson/blob/ben/slash-command/test/fixtures/slack_identifiers.yml) as in the [users fixtures](https://github.com/zendesk/samson/blob/ben/slash-command/test/fixtures/users.yml). So `slack_identifiers(:deployer)` gets you an identifier that's supposed to be associated with `users(:deployer)`. I just realized how to make that more succinct, so it should be more clear now.
- `current_user` comes from [the `CurrentUser` concern](https://github.com/zendesk/samson/blob/ben/slash-command/app/controllers/concerns/current_user.rb#L16-L18), and it's the currently-logged-in user as verified by Warden. This is the same approach used by every other controller.

ben:
Oh, I uploaded a [video of this feature in action](https://www.youtube.com/watch?v=EUmMXidzFAc).

rikeoka:
Oh I get the current_user now.  It's for the oauth.  I was wondering since in the context of slack contacting samson, current_user would never be set.

test/fixtures/slack_identifiers.yml
```diff
@@ -0,0 +1,15 @@
+deployer:
```
jonmoter: Could you put this file into `plugins/slack_app/test/fixtures`?

vitalis:
looks like a great feature!!! merge anytime soon? :)

jonmoter:
I'd say this is pretty well-reviewed.  Let's merge this, and if we need to make further tweaks we can do that in separate PRs.

👍

vitalis:
👏

plugins/slack_app/README.md
```diff
@@ -0,0 +1,12 @@
+# Slack App Plugin
+
+This plugin adds a `/deploy` command to Slack which creates Samson deployments.
+To activate:
+
+1. Go over to Slack and create a [new app](https://api.slack.com/apps/new). Many of the fields are completely up to you, but these are important:
+  - Grab the Client ID and secret and verification token, and record them in your environment as `SLACK_CLIENT_ID`, `SLACK_CLIENT_SECRET`, and `SLACK_VERIFICATION_TOKEN`.
```
vitalis: I can't find verification token.
Crated new app, and entered app credentials tab.

ben: That's odd. Mine looks like this:

image

Maybe you have to enable interactive messages or slash-commands before it shows up?

vitalis: Yes, now I see it

vitalis: I think it's good idea to update the README and move the credentials part right before `2.`

ben: @vitalis, why don't you submit that PR? I'd love for you to get another contributor credit for it.

plugins/slack_app/README.md
```diff
@@ -0,0 +1,12 @@
...
+  - Add `https://your.samson.tld/slack_app/oauth` as a redirect URI.
+  - Enable interactive messages, and set the URI to `https://your.samson.tld/slack_app/interact`.
+  - Add a `/deploy` slash-command, and set its URL to `https://your.samson.tld/slack_app/command`.
+2. Visit https://your.samson.tld/slack_app/oauth, and click the "Connect to Slack" button.
+3. Switch back over to Slack, and try out the `/deploy` command!
```
vitalis: Hi @ben, there is an issue when I work by this README:
if I try the `/deploy` command there is an exception:
```bash
2016-08-07T08:01:52.804249+00:00 app[web.1]: Completed 500 Internal Server Error in 15ms (ActiveRecord: 4.5ms)
2016-08-07T08:01:52.809853+00:00 app[web.1]: 
2016-08-07T08:01:52.809855+00:00 app[web.1]: NoMethodError (undefined method `captures' for nil:NilClass):
2016-08-07T08:01:52.809856+00:00 app[web.1]:   plugins/slack_app/app/controllers/slack_app_controller.rb:55:in `command_response'
2016-08-07T08:01:52.809857+00:00 app[web.1]:   plugins/slack_app/app/controllers/slack_app_controller.rb:34:in `command'
```
I didn't find any example on how to use the `/deploy` command.

ben: So it's failing on [this line](https://github.com/zendesk/samson/blob/216e3c5cf297f5e97dd65713586fab1bb2f4c7c3/plugins/slack_app/app/controllers/slack_app_controller.rb#L56), and it's a bug. If the syntax doesn't match the regex, we should respond with a private message that helps the user figure out how to do it properly. Want to submit a PR with a readme update and a fix?

PR #1131, opened 2016-07-08
Include soft deleted in Deploy Reports

rikeoka:
* Include soft-deleted projects in Deploy CSV Report filters
* Include soft-deleted items in Deploy CSV Report
* Bug fixes in Jira
* Add real-time reports for deploys#search limited to 1000 records for RT download of what you see
* Add stage name and deploy groups names to the report
* Test that soft deleted project deploys are included in report and in the dropdown

/cc @zendesk/samson

### Tasks
 - [X] Add capability to report on soft_deleted projects and select soft_deleted projects
 - [x] Update filter to support deploy_groups.production
 - [x] Update filter to change ending date filter to be 23:59:59 instead of 00:00:00
 - [x] Work out job failures caused by soft_deleted environment.  
 - [x] Add in deploy groups & stage name to report
 - [x] Fast response deploys#search.csv reports limited to a maximum of 1000 deploys, enough for quick analysis reviews.  If over 1000, it indicates that you should go to the job executed report.
 - [X] Change "no code deployed" to "code deployed" in report
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-263

### Risks
- Level: Medium
- Modification to default scope for Deploy, Environment, and Stage.  Global variable $INCLUDE_DELETED removes the soft deletion scope when csv report is being generated in CsvExportJob.  It is unset upon completion and failures.


rikeoka:
@grosser / @jonmoter 

Updated the report to include soft deleted deploys in the report.

If there is a better way to do this let me know.  I tried other methods but it kept on applying the scope for the associations.

app/models/deploy.rb
```diff
@@ -130,11 +130,31 @@ def self.expired
...
+    user.id == buddy_id ? "bypassed" : r_buddy.try(:name)
   end
 
   def buddy_email
-    user.id == buddy_id ? "bypassed" : buddy.try(:email)
+    user.id == buddy_id ? "bypassed" : r_buddy.try(:email)
+  end
+
+  def r_buddy
+    User.unscoped { buddy_id > 0 ? User.find(buddy_id) : nil }
```
grosser: uncachabel ... how about `buddy || (buddy_id && User.with_deleted { User.find(buddy_id) })`

app/models/deploy.rb
```diff
@@ -130,11 +130,31 @@ def self.expired
...
+  def r_project
+    Project.unscoped { Project.find(r_stage.project_id) }
+  end
+
+  def r_stage
+    Stage.unscoped { Stage.find(stage_id) }
+  end
+
+  def r_summary
+    "#{summary_action} #{short_reference} to #{r_stage.name}"
```
grosser: add a option to summary to use either stage or stage_with_deleted

app/models/deploy.rb
```diff
@@ -130,11 +130,31 @@ def self.expired
...
   def buddy_name
-    user.id == buddy_id ? "bypassed" : buddy.try(:name)
+    user.id == buddy_id ? "bypassed" : r_buddy.try(:name)
   end
 
   def buddy_email
-    user.id == buddy_id ? "bypassed" : buddy.try(:email)
+    user.id == buddy_id ? "bypassed" : r_buddy.try(:email)
+  end
+
```
grosser: also all private ... and tests ...

app/models/deploy.rb
grosser: also r_buddy is a strange name ... buddy_with_deleted ?

app/views/csv_exports/new.html.erb
```diff
@@ -30,7 +30,7 @@
       </div>
       <div class="form-group">
         <%= label_tag :project, "Project", class: "col-lg-2 control-label" %>
-        <% projects = [['', 0]] + Project.pluck(:name, :id) %>
+        <% projects = [['', 0]] + Project.unscoped.pluck(:name, :id) %>
```
grosser: mark deleted projects with (deleted) and maybe sort them to the back

test/models/deploy_test.rb
```diff
@@ -241,6 +255,32 @@ def deploy!
...
+    let(:deployer) { users(:super_admin) }
+    let(:other_user) { users(:deployer_buddy) }
+    let(:prod) { stages(:test_production) }
+    let(:prod_deploy) { deploys(:succeeded_production_test) }
+    let(:job) { jobs(:succeeded_production_test) }
+
+    before { Stage.any_instance.stubs(:deploy_requires_approval?).returns true }
+
+    it "returns array with deleted object values" do
+      project.update_attributes(repository_url: "https://github.com/samson-test-org/example-project")
```
rikeoka: Had to do this to get it to pass the test.  This probably means that because of the validations in projects it will fail if the project is soft deleted and the repo is deleted.

Not sure if its possible to preload all the associated data into the object without associations where it skips any validations for reporting.

app/jobs/csv_export_job.rb
```diff
@@ -34,22 +34,19 @@ def deploy_csv_export(csv_export)
     filename = csv_export.path_file
     filter = csv_export.filters
 
-    @deploys = Deploy.joins(:stage, :job).where(filter)
+    @deploys = Deploy.joins('INNER JOIN stages ON stages.id = deploys.stage_id INNER JOIN jobs ON jobs.id =' \
+                            ' deploys.job_id').unscope(where: :deleted_at).where(filter)
```
grosser: 2-space indent ?

grosser: or
```
.joins(
   'long string
)
```

app/models/deploy.rb
```diff
@@ -130,19 +135,41 @@ def self.expired
...
+    user.id == buddy_id ? "bypassed" : buddy_with_deleted.try(:email)
   end
 
   def url
     AppRoutes.url_helpers.project_deploy_url(project, self)
   end
 
+  def to_csv
+    [id, project_with_deleted.name, summary(with_deleted: true), commit, job.status, updated_at,
+     start_time, user_with_deleted.try(:name), user_with_deleted.try(:email), buddy_name, buddy_email,
```
grosser: 2-space indent plz

app/models/deploy.rb
```diff
@@ -23,8 +23,13 @@ def cache_key
     [super, commit]
   end
 
-  def summary
-    "#{job.user.name} #{deploy_buddy} #{summary_action} #{short_reference} to #{stage.name}"
+  def summary(with_deleted: false)
+    if with_deleted
+      "#{job.user.name} #{deploy_buddy(with_deleted: true)} #{summary_action} #{short_reference} to"\
```
grosser: always call `deploy_buddy(with_deleted: with_deleted)`

app/models/deploy.rb
```diff
@@ -179,8 +206,12 @@ def validate_stage_uses_deploy_groups_properly
     end
   end
 
-  def deploy_buddy
-    return unless stage.deploy_requires_approval?
+  def deploy_buddy(with_deleted: false)
+    if with_deleted
+      return unless stage_with_deleted.deploy_requires_approval?
```
grosser: always call `stage_with_deleted(with_deleted: with_deleted)`

app/views/csv_exports/new.html.erb
```diff
@@ -30,7 +30,8 @@
       </div>
       <div class="form-group">
         <%= label_tag :project, "Project", class: "col-lg-2 control-label" %>
-        <% projects = [['', 0]] + Project.pluck(:name, :id) %>
+        <% projects = [['', 0]] + Project.pluck(:name, :id) +
+           Project.unscoped.where("deleted_at IS NOT NULL").map { |p| ["(deleted) #{p.name}", p.id] } %>
```
grosser: still pluck and then map so it pulls down less data

test/models/deploy_test.rb
grosser: `update_column(:deleted_at, Time.now)` ?

rikeoka: The project fixture had an invalid repo url and was giving an invalid repo url error whenever the project was called.  Not sure if this will error in a real instance when the project and repo is deleted.  I can't reproduce it in a live account, probably because GitRepository already exists for the project.

app/models/deploy.rb
rikeoka: removed with_deleted: with_deleted option since to have valid information it should pull deleted stages.

test/models/deploy_test.rb
rikeoka: Ah got it now, it was doing the validation on the project, when I was soft_deleting it.

app/jobs/csv_export_job.rb
```diff
@@ -34,22 +34,20 @@ def deploy_csv_export(csv_export)
     filename = csv_export.path_file
     filter = csv_export.filters
 
-    @deploys = Deploy.joins(:stage, :job).where(filter)
+    @deploys = Deploy.joins(
+      'INNER JOIN stages ON stages.id = deploys.stage_id INNER JOIN jobs ON jobs.id = deploys.job_id'
+    ).unscope(where: :deleted_at).where(filter)
```
grosser: this select all deleted deploys now ... and not both deleted and not deleted ?

app/models/deploy.rb
```diff
@@ -130,19 +130,43 @@ def self.expired
...
+      id, project_with_deleted.name, summary, commit, job.status, updated_at,
+      start_time, user_with_deleted.try(:name), user_with_deleted.try(:email), buddy_name, buddy_email,
+      stage_with_deleted.production, stage_with_deleted.no_code_deployed, project_with_deleted.deleted_at
+    ]
+  end
+
   private
 
+  def buddy_with_deleted
+    buddy || (buddy_id && User.with_deleted { User.find(buddy_id) })
```
grosser: need to cache all of these so they do not cause multiple queries ... also some tests for 'does no query when there is no id' ... 

```
return @foo if defined?(@foo)
@foo = bla || with_delete.find_by_id(xxx)
```

most likely also find_by_id so we don't blow up if something was hard-deleted ...

app/models/deploy.rb
```diff
@@ -130,19 +130,43 @@ def self.expired
...
+  end
+
+  def project_with_deleted
+    Project.unscoped { Project.find(stage_with_deleted.project_id) }
+  end
+
+  def stage_with_deleted
```
grosser: should do the same `stage ||` dance to leverage the pre-loaded stages -> avoid n+1 queries

app/views/csv_exports/new.html.erb
```diff
@@ -30,7 +30,8 @@
       </div>
       <div class="form-group">
         <%= label_tag :project, "Project", class: "col-lg-2 control-label" %>
-        <% projects = [['', 0]] + Project.pluck(:name, :id) %>
+        <% projects = [['', 0]] + Project.pluck(:name, :id) +
+           Project.unscoped.where("deleted_at IS NOT NULL").pluck(:name, :id).map { |p| ["(deleted) #{p[0]}", p[1]] } %>
```
grosser: `unscoped.where("deleted_at IS NOT NULL")` -> `scope :deleted`

app/models/deploy.rb
```diff
@@ -130,19 +130,43 @@ def self.expired
...
+    Stage.unscoped { Stage.find(stage_id) }
+  end
+
+  def user_with_deleted
+    User.unscoped { job.user_id > 0 ? User.find(job.user_id) : nil }
+  end
+
```
grosser: instead of repeating this logic maybe do a concern like:
```Ruby
def with_deleted(association)
  @with_deleted ||= {}
  @with_deleted.fetch(object) do
    found = 
    @with_deleted[object] = send(object) || (id = send("#{object}_id") && association(object).klass.unscoped.find_by_id(id) }
end
```

grosser: I guess for now the copy-paste is good enough ... but at least some tests for the edge cases

rikeoka: would it make sense to do an association instead?
belongs_to :object_with_deleted, -> { unscope(where: 'deleted_at') }, -> class: 'Object', foreign_key: 'object_id

At least it probably makes sense in certain cases like buddy to remove complexity

app/jobs/csv_export_job.rb
rikeoka: This works the same as unscoped, but instead of removing the entire default_scope it just removes from the query "WHERE `deleted_at` IS NULL" condition.  This was when I was trying to join with the associations using joins(:jobs, :stages) but fails since this only applies on the main Model and not associations.

app/models/deploy.rb
grosser: that could work ... and would cache nicely ...

app/jobs/csv_export_job.rb
rikeoka: With the different association, optimized the query to prefetch associations during the join except for users and buddy.  Otherwise the in-memory table probably gets a bit large.

app/jobs/csv_export_job.rb
```diff
@@ -34,9 +34,8 @@ def deploy_csv_export(csv_export)
     filename = csv_export.path_file
     filter = csv_export.filters
 
-    @deploys = Deploy.joins(
-      'INNER JOIN stages ON stages.id = deploys.stage_id INNER JOIN jobs ON jobs.id = deploys.job_id'
-    ).unscope(where: :deleted_at).where(filter)
+    @deploys = Deploy.includes(:job, {stage_with_deleted: :project}).joins(:job, { stage_with_deleted: :project })
```
rikeoka: Should this be done with select to grab only what's needed from the db to reduce amount of data in memory?

app/models/deploy.rb
rikeoka: Testing each of these associations is needed?  I think it's a bit overkill to do that.

grosser: test on in-depth and the rest can be copy-paste ?

app/jobs/csv_export_job.rb
grosser: I'd imagine it only instantiates each project once, so this might be not that bad

rikeoka:
@grosser fixed all the comments.  Let me know if its ready for a squash.

app/controllers/csv_exports_controller.rb
```diff
@@ -71,9 +71,18 @@ def deploy_filter
     end
 
     if production = params[:production].presence
-      case production
-      when 'Yes' then filter['stages.production'] = true
-      when 'No'  then filter['stages.production'] = false
+      if production == 'Yes'
+        if DeployGroup.enabled?
+          filter['environments.production'] = true
```
grosser: logic duplication, can dry up with:
`filter_key = (DeployGroup.enabled? ? 'environments.production' : 'stages.production')`

and then 

```
filter[filter_key] = true
```

grosser: some docs like '# we join different tables when using deploy groups' might be good too

app/controllers/deploys_controller.rb
```diff
@@ -178,4 +184,19 @@ def find_deploy
   def active_deploy_scope
     current_project ? current_project.deploys.active : Deploy.active
   end
+
+  # Creates a CSV for @deploys as a result of the search query limited to 1000 for speed
+  def realtime_csv
+    csv_limit = (params[:limit].presence || 1000).to_i
```
grosser: users can send in 99999 ->
````
max = 1000
csv_limit = [(params[:limit].presence || max).to_i, max].min
```

app/controllers/deploys_controller.rb
```diff
@@ -178,4 +184,19 @@ def find_deploy
...
+    CSV.generate do |csv|
+      csv << Deploy.csv_header
+      @deploys.limit(csv_limit).each { |deploy| csv << deploy.csv_line }
```
grosser: .as_csv could be a nice method name ...

app/controllers/deploys_controller.rb
```diff
@@ -178,4 +184,19 @@ def find_deploy
...
+      csv << ['-', 'count:', @deploys.limit(csv_limit).count]
+      csv << ['-', 'params:', params]
+      if @deploys.count > csv_limit
```
grosser: could save that .count query (which will be expensive with lots of deploys) and ask for 1001 and then check if we got 1001 ... and always pop 1

app/jobs/csv_export_job.rb
```diff
@@ -34,28 +34,44 @@ def deploy_csv_export(csv_export)
     filename = csv_export.path_file
     filter = csv_export.filters
 
-    @deploys = Deploy.joins(:stage, :job).where(filter)
+    get_deploys(filter)
```
grosser: avoid get_ / set_ methods ... we are not in java land :)

app/jobs/csv_export_job.rb
```diff
@@ -34,28 +34,44 @@ def deploy_csv_export(csv_export)
...
-        ]
+        csv << deploy.csv_line
       end
       csv << summary
       csv << filters_applied
     end
   end
 
+  def get_deploys(filter)
+    if filter.keys.include?('environments.production')
```
grosser: can do a `@deploys = filter_deploys` so this method does not need to know about `@deploys` and dry it up a bit

app/jobs/csv_export_job.rb
```diff
@@ -34,28 +34,44 @@ def deploy_csv_export(csv_export)
...
+      production_value = filter.delete('environments.production')
```
grosser: csv_exports_controller sets 2 different keys and this only supports 1 key ... potential bug ?

app/jobs/csv_export_job.rb
```diff
@@ -34,28 +34,44 @@ def deploy_csv_export(csv_export)
...
+      # To match logic of stages.production? True when any deploy_group environment is true or
+      # deploy_groups environment is empty and stages is true
+      production_query = "(A.production = ? OR (A.production IS NULL AND stages.production = ?))"
+      production_query = "NOT " + production_query if production_value == false
```
grosser: FYI `production_query.prepend "NOT "`

grosser: A for a table name is not very intuitive ... maybe deploys_left or something like that ?

app/jobs/csv_export_job.rb
grosser: ah nvm checking that above

app/jobs/csv_export_job.rb
grosser: so this is the `searching for deleted` test ? ... can we make that more obvious ? :)

app/jobs/csv_export_job.rb
```diff
@@ -34,28 +34,44 @@ def deploy_csv_export(csv_export)
...
+      # so it is important this is run only if enviornments.production was set
+      @deploys = Deploy.includes(:buddy, job: :user, stage_with_deleted: :project).
+        joins("INNER JOIN jobs ON jobs.id = deploys.job_id INNER JOIN stages ON stages.id = deploys.stage_id").
+        joins("LEFT JOIN " \
+          "(SELECT DISTINCT deploy_groups_stages.stage_id, environments.production FROM deploy_groups_stages " \
+          "INNER JOIN deploy_groups ON deploy_groups.id = deploy_groups_stages.deploy_group_id " \
+          "INNER JOIN environments ON environments.id = deploy_groups.environment_id) A ON A.stage_id = stages.id").
+        unscope(where: :deleted_at).where(filter).where(production_query, true, true)
+    else
+      @deploys = Deploy.includes(:buddy, job: :user, stage_with_deleted: :project).
```
grosser: does that stage_with_deleted work ... I though you said it does not ....

app/jobs/csv_export_job.rb
grosser: no wait ... they both query only for deleted records ???

app/jobs/csv_export_job.rb
rikeoka: This works for preloading the associated objects.  What doesn't work is the use of it to automatically join the tables for the filter to work.  So i can't do Deploy.includes(:stages_with_deleted).joins(:stages_with_deleted) but I can do Deploy.includes(:stages_with_deleted).joins("INNER JOIN ...")

app/models/csv_export.rb
```diff
@@ -31,11 +31,18 @@ def filters
...
     if filter['deploys.created_at']
       dates = filter['deploys.created_at'].scan(/-?\d+-\d+-\d+/)
-      filter['deploys.created_at'] = (Date.parse(dates[0])..Date.parse(dates[1]))
+      filter['deploys.created_at'] = (DateTime.parse(dates[0] + "T00:00:00Z")..DateTime.parse(dates[1] + "T23:59:59Z"))
     end
     filter
   end
 
+  def filters_project
+    if filters['stages.project_id']
```
grosser: can avoid duplication with `if id = filters['stages.project_id']`

app/models/csv_export.rb
```diff
@@ -31,11 +31,18 @@ def filters
...
+      proj = Project.unscoped.where(id: filters['stages.project_id']).first.try(:permalink)
+      proj + '_' unless proj.nil?
```
grosser: `unless proj.nil?` -> `if proj`

app/jobs/csv_export_job.rb
rikeoka: With this I was getting the expected output report, but had a performance improvement in generation time 10s for 20,000 deploys vs 28 seconds without.

app/models/deploy.rb
```diff
@@ -4,7 +4,8 @@ class Deploy < ActiveRecord::Base
   belongs_to :stage, touch: true
   belongs_to :build
   belongs_to :job
-  belongs_to :buddy, class_name: 'User'
+  belongs_to :buddy, -> { unscope(where: "deleted_at") }, class_name: 'User'
+  belongs_to :stage_with_deleted, -> { unscope(where: "deleted_at") }, class_name: 'Stage', foreign_key: 'stage_id'
```
grosser: isn't this `deleted_stage` ... since it only filters those that are deleted ?

app/jobs/csv_export_job.rb
rikeoka: Moving the subquery out to a pre-constructed string and labeling it "StageProd" 

stage.production? returns true if any DeployGroup.Environment.production is true or if no Deploy Groups are associated with the stage, Stage.production.

production = true, The query is returning StageProd.production is true or if StageProd is not present the stages.production is true.

production = false, probably could use the same logic with the false values, but there may be some conditions where NOT could give the undefined values as well.

stage_prod null = no associated environment.production for the stage (can never be null from db level validation)
alt logic is not using not and passing false on instead of true

stage | stage_prod | expected from stage.production? | this logic | alt logic
----- | ------------ | --------------------------------- | --------- | --------
any | true | true | true | true
any | false | false | false | false
true | null | true | true | true
false | null | false | false | false
null | null | false | false | never applicable

app/jobs/csv_export_job.rb
rikeoka: Adding another grouping around the tests to make it clear with DeployGroups and without.  Wasn't the searching for deleted test.

app/jobs/csv_export_job.rb
rikeoka: For some reason this didn't work out the way it should, so going to redo the logic on the filter to get expected.

app/models/deploy.rb
rikeoka: It removes the "WHERE stages.deleted_at IS NULL" from the query.  So it would get both non-deleted and deleted stages.  If the desired is deleted stage, it would need unscope(where: "deleted_at").where("stages.deleted_at IS NOT NULL") instead.  

I didn't want to update the regular stages scope, since that would cause some wierd UI bugs, but changing the buddy association was fine, since generally if people are looking back in time, they would still want to know who the buddy is even though the buddy is no longer an active user in Samson.

grosser: uh nice ... so it removes the scope on that column ... funky

app/controllers/deploys_controller.rb
```diff
@@ -178,4 +184,22 @@ def find_deploy
   def active_deploy_scope
     current_project ? current_project.deploys.active : Deploy.active
   end
+
+  # Creates a CSV for @deploys as a result of the search query limited to 1000 for speed
+  def as_csv
+    max = 1000
+    csv_limit = [(params[:limit].presence || max).to_i, max].min
+    # may be expensive to count an unfiltered @deploys so do count once
+    deploy_count = @deploys.limit(csv_limit + 1).count
```
grosser: don't think you can limit a count ... also still doing 2 queries ... 

deploys = @deploys.limit(x+1).to_a
deploys[0..-2].each do |deploy|
if deploys.length > limit

app/jobs/csv_export_job.rb
```diff
@@ -34,28 +34,50 @@ def deploy_csv_export(csv_export)
...
       end
       csv << summary
       csv << filters_applied
     end
   end
 
+  def filter_deploys(filter)
+    if filter.keys.include?('environments.production')
+      production_value = filter.delete('environments.production')
+      # To match logic of stages.production? True when any deploy_group environment is true or
```
grosser: best add a comment to `def production?` too ... `# this is reproduced via sql in xxx.rb, make them sync when editing`

app/jobs/csv_export_job.rb
```diff
@@ -34,28 +34,50 @@ def deploy_csv_export(csv_export)
...
+      # deploy_groups environment is empty and stages is true
+      production_query = "(StageProd.production = ? OR (StageProd.production IS NULL AND stages.production = ?))"
```
grosser: should be in a if/else block since it is just overwritten when `production_value == false`

app/models/job.rb
```diff
@@ -1,6 +1,6 @@
 class Job < ActiveRecord::Base
   belongs_to :project
-  belongs_to :user
+  belongs_to :user, -> { unscope(where: 'deleted_at') }
```
grosser: does `unscoped(where: :deleted_at)` also work ?

grosser:
hmm any chance we can do this whole with_deleted without adding new methods everywhere and do `Stage.with_deleted { ... generate csv ... }` instead ?

app/models/stage.rb
```diff
@@ -174,6 +180,10 @@ def destroy
     super
   end
 
+  def deploy_group_names
+    deploy_groups.pluck(:name).join('|')
```
grosser: this seems weird ... should return an array ... then users can join them as they want ...
also needs `.sort_by(&:natural_order)` ... could make `natural_order` static and call it or do a `deploy_groups.select(:name).sort_by(&:natural_order).map!(&:name)`

test/controllers/deploys_controller_test.rb
```diff
@@ -168,6 +168,19 @@ def self.with_and_without_project(&block)
         assert_response :ok
       end
 
+      it "it renders csv" do
```
grosser: `it it`

test/controllers/csv_exports_controller_test.rb
```diff
@@ -211,21 +211,28 @@ def self.show_test(state)
...
+      def self.prod_filter_test(prod, groups)
+        it "with production filter #{prod == "Yes"} and DeployGroup enabled #{groups} creates correct filter" do
+          DeployGroup.stubs(:enabled?).returns(groups)
+          post :create, production: prod
+          csv_filter = CsvExport.last.filters
+          csv_filter[groups ? "environments.production" : "stages.production"].must_equal prod == "Yes"
+        end
       end
 
+      prod_filter_test("Yes", true)
```
grosser: could be nicer with `it_filters_production "Yes", true`

app/models/job.rb
rikeoka: unscoped doesn't work.  It works as a class method to remove all scopes.  In associations unscoped doesn't seem to work at all, and only the query methods seems to work.

rikeoka:
@grosser updated to address the comments.

Unfortunately with the default scope set on the model, the associations will always apply the default scope on associations even with AssociatedClass.with_deleted or unscoped.
Ways around this would be Class.unscoped.find(class_id) but where data is being gathered from a method, it would likely require duplicating the logic of the method.  This probably is the cleanest way as it fails nicely when the data can't be found.  Ie can't do Project.with_deleded { stage.project.name } the project query will still apply the WHERE projects.delete_at IS NULL statement. 

For deploy.csv_line it would mean to get deploy.project_with_deleted.name attribute, it would need a statement like Project.unscoped.where(id: Stage.unscoped.where(id: stage_id).project_id).first.try(:name) unless Stage.unscoped.where(id: stage_id).first.project_id.nil?

I'm open to suggestions.  Maybe raw SQL query?  I don't know if postgree and sqllite supports the GROUP_CONCAT syntax MySQL does.  Support for multiple db engines makes doing a complex query difficult to work across all of the supported db types.  BTW, when I ran the following raw SQL extract from mySQL it can dump 20k records in about 3 seconds versus 16-8 seconds through the Rails way.

SELECT deploys.id, projects.name, jobs.status, jobs.commi, deploys.updated_at, deploys.created_at, deployer.name, deployer.email, buddy.name, buddy.email, stages.name, (stages.production OR StageProd.production) as production, NOT stages.no_code_deployed as code_deployed, projects.deleted_at, DEPLOY_GROUPS.deploy_group_names
FROM deploys
INNER JOIN jobs ON jobs.id = deploys.job_id 
INNER JOIN stages ON stages.id = deploys.stage_id
INNER JOIN projects on projects.id = stages.project_id
LEFT JOIN (SELECT DISTINCT deploy_groups_stages.stage_id, environments.production 
  FROM deploy_groups_stages
  INNER JOIN deploy_groups ON deploy_groups.id = deploy_groups_stages.deploy_group_id
  INNER JOIN environments ON environments.id = deploy_groups.environment_id
) StageProd ON StageProd.stage_id = stages.id
LEFT JOIN (SELECT DISTINCT deploy_groups_stages.stage_id, GROUP_CONCAT(DISTINCT deploy_groups.name ORDER BY deploy_groups.name ASC SEPARATOR '|') as deploy_group_names
  FROM deploy_groups_stages 
  INNER JOIN deploy_groups on deploy_groups.id = deploy_groups_stages.deploy_group_id
  GROUP BY deploy_groups_stages.stage_id
) DEPLOY_GROUPS ON DEPLOY_GROUPS.stage_id = stages.id
LEFT JOIN users as deployer ON deployer.id = jobs.user_id
LEFT JOIN users as buddy ON buddy.id = deploys.buddy_id

grosser:
man that's some ugly query :D

... hmm since the default scope is a block we might be able to completely
disable it via some global setter ...

default_scope { $GIMME_ALL ? where(nil) : where(deleted_at: nil) }

$GIMME_ALL = true
export
$GIMME_ALL = false

rikeoka:
@grosser 

This works, but it may be a bit dangerous and cause UI glitches while a report is being generated.  At a minimum deleted Deploy, Environment, Project and Stage would be affected.  There is an off chance for instance that when the report is running a deleted stage could appear and a deploy is created for one.

Because of tweaking the default scope going to bump the risk for this to medium as well since it could cause unknown effects while reports are being run.

The test to show how this works for 3 cases nil, false, true is in models/deploy_test #stage  Stage had the quirkiest effects, I think paper_trail may add the soft_deletion scope on the model, but fortunately the unscope(with: 'deleted_at') works.  I didn't bother updating deploys#buddy or jobs#user since I think it makes sense that association is always unscoped for UI purposes.

grosser:
has_soft_deletion unless self < SoftDeletion::Core
+
same in plugins/pipelines/app/decorators/stage_decorator.rb
+
default_scope -> { Thread.current[:stage_with_deleted] ? where(nil) :
where(deleted_at: nil) }

and it works :D

-> no funky with_deleted methods / able to use joins(:environments) etc

... I'll update soft_deletion to use this since it's more universal then
unscoped/with_deleted ...

"SELECT `stages`.* FROM `stages` INNER JOIN `projects` ON `projects`.`id` =
`stages`.`project_id` AND `projects`.`deleted_at` IS NULL WHERE
`stages`.`deleted_at` IS NULL  ORDER BY `stages`.`order` ASC"
>> Thread.current[:stage_with_deleted] = true
=> true
>> Stage.joins(:project).to_sql
=> "SELECT `stages`.* FROM `stages` INNER JOIN `projects` ON
`projects`.`id` = `stages`.`project_id`  ORDER BY `stages`.`order` ASC"


... whereas the current approach does not work ...

Stage.with_deleted { Project.with_deleted { Stage.joins(:project).to_sql } }
=> "SELECT `stages`.* FROM `stages` INNER JOIN `projects` ON
`projects`.`id` = `stages`.`project_id` AND `projects`.`deleted_at` IS NULL"

rikeoka:
Cool.  It also works with this so a where(nil) isn't even needed.
default_scope -> { where(deleted_at: nil) unless Thread.current[:stage_with_deleted] }

grosser:
bump soft_deletion to 0.8.0 and you can get rid of all the extra model helpers and just use `User.with_deleted { Stage.with_deleted { ...` around your code

rikeoka:
Updated.  Squashed the last few experimental commits to show the cleanup of the unscoping calls better.

I'll leave it up to you if we want to apply the same to the Job.user and Deploy.buddy association, since it does make sense from a UI perspective to show "John Doe" deployed XYZ is better than "Deleted User" deployed XYZ.  Rest of them should be cleaned up.

app/jobs/csv_export_job.rb
```diff
@@ -30,32 +30,66 @@ def generate_csv(csv_export)
     Airbrake.notify(e, error_message: "Export #{csv_export.id} failed.")
   end
 
+  # Relper method to removes the default soft_deletion scope for these models for the report
+  def remove_deleted_scope_and_create_report(csv_export)
+    Deploy.with_deleted do
```
grosser: this looks fun :D

grosser:
👍 so much better :D 
crazy sql is still a bit scary .... but at least it is contained ...

That buddy deletion might come back to bite us when someone joins on buddy and then something does not work as expected, but looks like a sane change ...

PR #1132, opened 2016-07-12
Update users report

rikeoka:
* Update user access reports for compliance
* Add new_users and users to CsvExportController
* Refactor User.to_csv to support requested options
* User report now work on deleted users, report effective project permissions, report on specific project's users or specific user's projects
* Admins can access the reports the same way under /admins/users#index 
* New links for admins are available to report on user's projects at /admin/users#show or project's users at /project/users#index
* Non-admins but authenticated users (compliance, etc) have the ability to generate the report at /csv_exports#new_users.  
* Relaxed permissions restriction to any logged in user for user reports since no sensitive data is present in report
* Updated tests

screen shot 2016-07-12 at 11 56 51 am
Options at csv_exports#new_users

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-261

### Risks
- Level: Low


app/controllers/csv_exports_controller.rb
```diff
@@ -19,6 +19,19 @@ def new
...
   end
 
+  def new_users
+  end
+
+  def users
+    options = user_filter
+    respond_to do |format|
+      format.csv do
+        options[:datetime] = Time.now.strftime "%Y%m%d_%H%M"
```
grosser: move this into user_filter method ?

app/controllers/csv_exports_controller.rb
```diff
@@ -52,6 +65,15 @@ def download
     end
   end
 
+  def user_filter
+    options = {}
+    options[:inherited] = params[:inherited] == "true"
+    options[:deleted] = params[:deleted] == "true"
```
grosser: prefer a single hash

```
{
  deleted: xyz,
  foobar: xyz
```

grosser: ah nvm ... project and user should not be set ...

app/models/user.rb
```diff
@@ -54,22 +54,53 @@ def administrated_projects
     scope
   end
 
-  def self.to_csv
-    @users = User.order(:id)
+  def self.to_csv(
+    inherited: false, deleted: false, project_id: nil, user_id: nil,
+    datetime: (Time.now.strftime "%Y%m%d_%H%M")
+  )
+    inherited = true if project_id || user_id
```
grosser: feels strange that it overrides that option ...

app/models/user.rb
```diff
@@ -54,22 +54,53 @@ def administrated_projects
...
+    users = if user_id.nil?
```
grosser: looks like it's finding deleted users when giving id and not deleted ... how about:

```Ruby
users = (deleted ? User.unscoped : User)
users = users.order(:id)
users = users.where(id: user_id) if user_id
```

app/models/user.rb
```diff
@@ -136,4 +176,12 @@ def object_attrs_for_paper_trail(attributes)
   def set_token
     self.token = SecureRandom.hex
   end
+
+  private_class_method def self.permissions_count(project_id, inherited, users, permissions_projects = [])
```
grosser: don't need a default value for permissions_projects

app/models/user.rb
```diff
@@ -54,22 +54,53 @@ def administrated_projects
...
+    datetime: (Time.now.strftime "%Y%m%d_%H%M")
+  )
+    inherited = true if project_id || user_id
+    users = if user_id.nil?
+      deleted ? User.unscoped.order(:id) : User.order(:id)
+    else
+      user_id && User.unscoped.where(id: user_id)
+    end
+    if inherited
+      permissions_projects = project_id ? Project.where(id: project_id) : Project.all
```
grosser: no need for `.all` ?

app/models/user.rb
```diff
@@ -54,22 +54,53 @@ def administrated_projects
...
+            csv << user.csv_line(user_project_role.project)
+          end
         end
       end
+      csv << summary
+      csv << options_applied
+    end
+  end
+
+  def csv_line(project = nil)
```
grosser: no need for default

app/models/user.rb
```diff
@@ -54,22 +54,53 @@ def administrated_projects
...
+    if project
+      [id, name, email, project.id, project.name, effective_project_role(project), deleted_at]
+    else
+      [id, name, email, "", "SYSTEM", role.name, deleted_at]
```
grosser: I'd prefer something like this to avoid the duplication

```Ruby
[
  id,
  name,
  (project ? project.id : "")
  ...
]
```

app/models/user.rb
```diff
@@ -121,6 +152,15 @@ def project_role_for(project)
     user_project_roles.find_by(project: project)
   end
 
+  def effective_project_role(project)
+    project_role = project_role_for(project)
```
grosser: project_role_for can move into the else

app/models/user.rb
```diff
@@ -121,6 +152,15 @@ def project_role_for(project)
...
+    if role_id == Role::SUPER_ADMIN.id
+      Role::ADMIN.name
+    else
+      role_id >= (project_role.try(:role_id) || 0) ? role.name : project_role.role.name
```
grosser: this logic already exists somewhere ... please reuse :)

app/models/user.rb
grosser: this look copy-pasty ... seems to duplicate logic from above ... can we reuse that somehow ?

app/views/csv_exports/new_users.html.erb
```diff
@@ -0,0 +1,46 @@
...
+      <div class="form-group">
+        <%= label_tag :inherited, "Show inherited project permissions", class: "col-lg-2 control-label" %>
+        <div class="col-lg-2">
+          <%= select_tag :inherited, options_for_select([['No', false], ['Yes', true]], false), class: "form-control" %>
+        </div>
+      </div>
+      <div class="form-group">
+        <%= label_tag :deleted, "Include deleted users", class: "col-lg-2 control-label" %>
+        <div class="col-lg-2">
+          <%= select_tag :deleted, options_for_select([['No', false], ['Yes', true]], false), class: "form-control" %>
```
grosser: checkbox instead of a select ?

app/views/csv_exports/new_users.html.erb
```diff
@@ -0,0 +1,46 @@
+<h1>Request a CSV Report</h1>
+
+<section>
+  <div class="csv-users-filters clearfix">
+    <%= form_tag users_csv_exports_path(format: :csv), method: :get, class: "form-horizontal" do %>
+      <div><strong>User Permission Reports</strong></div>
+      <div class="form-group">
+        <%= label_tag :inherited, "Show inherited project permissions", class: "col-lg-2 control-label" %>
+        <div class="col-lg-2">
+          <%= select_tag :inherited, options_for_select([['No', false], ['Yes', true]], false), class: "form-control" %>
```
grosser: checkbox instead of select ?

app/views/csv_exports/new_users.html.erb
```diff
@@ -0,0 +1,46 @@
...
+      <div class="form-group">
+        <%= label_tag :project_id, "Project", class: "col-lg-2 control-label" %>
+        <% projects = [['', nil]] + Project.pluck(:name, :id) %>
+        <div class="col-lg-4">
+          <%= select_tag(:project_id, options_for_select(projects), class: "form-control") %>
+        </div>
+      </div>
+      <div class="form-group">
+        <%= label_tag :user_id, "User", class: "col-lg-2 control-label" %>
+        <% users = [['', nil]] + User.pluck(:name, :id) +
```
grosser: should we sort by name ?

config/routes.rb
```diff
@@ -86,7 +86,12 @@
   get '/login', to: 'sessions#new'
   get '/logout', to: 'sessions#destroy'
 
-  resources :csv_exports, only: [:index, :new, :create, :show]
+  resources :csv_exports, only: [:index, :new, :create, :show] do
```
grosser: this looks unsymmetric ... can we re-use new and index with some kind of parameter ... otherwise create a new controller and make the existing one

test/models/user_test.rb
```diff
@@ -187,9 +187,56 @@
...
+    end
+
+    def csv_test(options = {}, expected = {})
+      meta_rows = 3
+      User.to_csv(options).split("\n").size.must_equal expected + meta_rows
+      User.to_csv(options).split("\n")[-2].split(",")[-1].to_i.must_equal expected
+    end
+  end
+
+  describe "#csv_line and #effective_project_role" do
```
grosser: split these into 1 block per method ?

test/models/user_test.rb
```diff
@@ -187,9 +187,56 @@
     end
   end
 
-  describe ".to_csv" do
-    it "generates csv" do
-      User.to_csv.split("\n").size.must_equal User.count + 1 + UserProjectRole.count
+  describe ".to_csv and .permissions_count" do
```
grosser: permissions_count is private so no need to mention it here

app/views/csv_exports/new_users.html.erb
```diff
@@ -0,0 +1,46 @@
+<h1>Request a CSV Report</h1>
```
grosser: `Request a User CSV Report` ?

app/views/csv_exports/new_users.html.erb
```diff
@@ -0,0 +1,46 @@
...
+
+<section>
+  <div class="csv-users-filters clearfix">
+    <%= form_tag users_csv_exports_path(format: :csv), method: :get, class: "form-horizontal" do %>
+      <div><strong>User Permission Reports</strong></div>
```
grosser: this should be in the h1

app/models/user.rb
```diff
@@ -54,22 +54,53 @@ def administrated_projects
...
 
-  def self.to_csv
-    @users = User.order(:id)
+  def self.to_csv(
+    inherited: false, deleted: false, project_id: nil, user_id: nil,
+    datetime: (Time.now.strftime "%Y%m%d_%H%M")
+  )
+    inherited = true if project_id || user_id
+    users = if user_id.nil?
+      deleted ? User.unscoped.order(:id) : User.order(:id)
```
sandlerr: `unscoped` can do surprising things sometimes, I think it reflects your intention better to do `User.with_deleted { User.order(:id) }`

app/models/user.rb
```diff
@@ -54,22 +54,53 @@ def administrated_projects
...
-      ]
-      @users.find_each do |user|
-        csv << [user.id, user.name, user.email, "", "SYSTEM", user.role.name]
-        UserProjectRole.where(user_id: user.id).joins(:project).find_each do |user_project_role|
-          csv << [
-            user.id, user.name, user.email, user_project_role.project_id,
-            user_project_role.project.name, user_project_role.role.name
-          ]
+      csv << ["id", "name", "email", "projectiD", "project", "role", "deleted at"]
+      users.find_each do |user|
```
sandlerr: did you have problems doing this in a single sql query `users.each` rather than `find_each`?

app/views/csv_exports/new_users.html.erb
```diff
@@ -0,0 +1,46 @@
...
+      </div>
+      <div class="form-group">
+        <%= label_tag :deleted, "Include deleted users", class: "col-lg-2 control-label" %>
+        <div class="col-lg-2">
+          <%= select_tag :deleted, options_for_select([['No', false], ['Yes', true]], false), class: "form-control" %>
+        </div>
+      </div>
+      <div class="form-group">
+        <%= label_tag :project_id, "Project", class: "col-lg-2 control-label" %>
+        <% projects = [['', nil]] + Project.pluck(:name, :id) %>
```
sandlerr: I kinda prefer using `permalink` over `id` here, I don't think we use project IDs much anywhere else

app/views/csv_exports/new_users.html.erb
```diff
@@ -0,0 +1,46 @@
...
+        <div class="col-lg-4">
+          <%= select_tag(:project_id, options_for_select(projects), class: "form-control") %>
+        </div>
+      </div>
+      <div class="form-group">
+        <%= label_tag :user_id, "User", class: "col-lg-2 control-label" %>
+        <% users = [['', nil]] + User.pluck(:name, :id) +
+            User.unscoped.where("deleted_at IS NOT NULL").pluck(:name, :id).map { |u| ["(deleted) #{u[0]}", u[1]] } %>
```
sandlerr: does this need to be unscoped? Probably another good place to use `User.with_deleted { User.pluck(:name, :id) }`

sandlerr:
I think it would be awesome if it was clear to users which buttons (if any) will immediately generate a csv file and which ones will just link you to a menu page where you customise your export. Maybe using a slightly different style, symbol or text in those buttons.

app/models/user.rb
rikeoka: I thought with_deleted just passed unscoped onto the Model so this would be simplier?  I might be wrong through https://github.com/grosser/soft_deletion/blob/master/lib/soft_deletion/core.rb#L20

app/models/user.rb
rikeoka: Had this since when reporting on a specific project it would be a bit off to see anywhere between less than 100% of the users.  
Same with reporting on a specific user to see less than 100% of the projects, since most of their privileges are associated with their system permissions.

app/models/user.rb
sandlerr: didn't realise that, no reason to change it then

app/models/user.rb
rikeoka: find_each saves a few percent of the total queries, but I guess it doesn't matter too much.

rikeoka:
The download links on the various user pages are immediate with predefined params based on the page you're on.

Create buttons starts a job
Download buttons immediately generates and sends

Create links go to menu options:)

app/views/csv_exports/new_users.html.erb
rikeoka: Kept it id to be consistent with the deploys report, since that reports the permalink.  

If we really should be using permalink instead, I can update both reports in a separate PR.

rikeoka:
@grosser updated.

@sandlerr user reports are generated real time.  From the respective user pages it'll generate the default reports for that page.  
From admin/users it'll generate report for all users no implied project permissions
From admin/users/:id, it'll generate report of a user's system role + implied project permissions (basically what's seen in the UI)
From project/:permalink/users, it'll generate report of all users for the project with implied project permissions (basically what's seen in the UI)
From csv_exports/new.html?type=users it'll ask users for report generation parameters and generate the report realtime.

test/controllers/csv_exports_controller_test.rb
```diff
@@ -76,16 +76,87 @@
     end
 
     describe "#new" do
-      before do
-        p = projects(:test).dup
-        p.name = "Other"
-        p.save!(validate: false)
+      describe "as html" do
+        before do
+          p = projects(:test).dup
```
sandlerr: this should work without `dup`

test/controllers/csv_exports_controller_test.rb
```diff
@@ -76,16 +76,87 @@
...
+            csv_test({user_id: users(:super_admin).id}, expected)
+          end
+
+          it "returns csv with multiple options" do
+            expected[:inherited] = true
+            expected[:deleted] = true
+            csv_test({inherited: "true", deleted: "true"}, expected)
+          end
+
+          def csv_test(options = {}, expected = {})
```
sandlerr: could you set the default value for `expected` to `expected` rather than empty?

sandlerr:
:+1:

test/controllers/csv_exports_controller_test.rb
rikeoka: This was to create a second project to see if it appears in the dropdown.  I had a previous where it only showed one project in the UI.  It may be overkill to test UI logic.

test/controllers/csv_exports_controller_test.rb
rikeoka: since it only is called in the tests and tests should provide expected values I just removed the options.

test/controllers/csv_exports_controller_test.rb
sandlerr: Cool, would you mind adding a comment to explain that? This is a slightly unconventional way to mock ActiveRecord objects

rikeoka:
@grosser / @sandlerr 
Updated the algorithm for the inherited option for the users report.  By manually caching UserProjectRole for each user it reduces the number of single result queries from O(N*M) to O(N).

It requires the User.to_csv, User.csv_line and User.effective_project_role to have some logic that would otherwise natively be handled by Rails to be manually done, but I think in this case since it has a major performance improvement when generating a full report with a large number of projects to justify this approach.

Inclusive
99projects 1001 users
127.722s, 145.680s, 131.808s

Optimized
99 projects 1001 users
5.662s, 6.492s, 5.956s, 5.889s, 5.792s

app/views/csv_exports/new.html.erb
```diff
@@ -30,7 +30,7 @@
       <div class="form-group">
         <%= label_tag :project, "Project", class: "col-lg-2 control-label" %>
         <% projects = [['', 0]] + Project.pluck(:name, :id) +
-           Project.deleted.map { |p| ["(deleted) #{p[0]}", p[1]] } %>
+           Project.deleted.pluck(:name, :id).map { |p| ["(deleted) #{p[0]}", p[1]] } %>
```
grosser: FYI can do `.map { |name, id| }` for readability

app/models/user.rb
```diff
@@ -54,25 +54,62 @@ def administrated_projects
...
+    users = users.order(:id)
+    users = users.where(id: user_id) if user_id
+    if inherited
+      permissions_projects = project_id ? Project.where(id: project_id) : Project
+      total = project_id ? users.count : (1 + permissions_projects.count) * users.count
+    else
+      total = users.count + UserProjectRole.joins(:user, :project).count
+    end
+    summary = ["-", "Generated At", datetime, "Users", users.count.to_s, "Total entries", total.to_s]
+    options_applied = [
```
grosser: nice to see this in the dump for later reproduction :)

app/views/csv_exports/index.html.erb
```diff
@@ -1,8 +1,9 @@
 <h1>Here are your active reports</h1>
 
 <section id="csv-exports" class="clearfix">
-  <div class="deploy-csv">
-    <%= link_to "Start a CSV export job", new_csv_export_path %>
+  <div>
+    <p><%= link_to "Create Deploys Report", new_csv_export_path %></p>
```
grosser: could remove the `Create` part since hitting that link does not create it ... `New Deploys Report` or just `Deploys Report`

app/models/user.rb
```diff
@@ -54,25 +54,62 @@ def administrated_projects
     scope
   end
 
-  def self.to_csv
-    @users = User.order(:id)
+  def self.to_csv(
```
grosser: this whole section might make a nice `UserCsvPresenter` class or something like that ... but can refactor that after it's merged ...

app/views/csv_exports/new.html.erb
```diff
@@ -43,11 +42,9 @@
...
-    <div class="clearfix">
-      <hr />
-      <div><strong>Admin reports</strong></div>
-      <div><%= link_to "Download Users CSV Report", admin_users_path(format: "csv") %></div>
-    </div>
-  <% end %>
+  <div class="clearfix">
+    <hr />
+    <div><strong>Other reports</strong></div>
+    <div><%= link_to "Create Users Report", new_csv_export_path(type: :users) %></div>
```
grosser: same here: `Request Users Report` or `Users Report` or `New Users Report`

app/views/csv_exports/new_users.html.erb
```diff
@@ -0,0 +1,47 @@
...
+        <div class="col-lg-offset-2 col-lg-10">
+          <%= submit_tag "Download Users CSV Report", class: 'btn btn-default' %>
+        </div>
+      </div>
+    <% end %>
+  </div>
+  <div class="clearfix">
+    <hr />
+    <div><strong>Other reports</strong></div>
+    <div><%= link_to "Create Deploys Report", new_csv_export_path %></div>
```
grosser: same with the create

grosser:
This N+1 optimization work might get lost when someone refactors the code ... so adding a `assert_sql_queries` test could make sure it does not degrade.

👍 looks scary, but reasonable ...

app/models/user.rb
rikeoka: should the csv_line and effective_project_role also be moved to the presenter given they aren't using the native model methods.  It'll keep the optimized logic outside of the model, that won't matter for other use cases.

grosser: sounds good!

rikeoka:
Seeing if this turns green.  
Refactored all of the User Csv methods into the UserCsvPresenter class in /app/presenters/.  Not sure if we'll use API presenters if so maybe it should be in /app/presenters/csv/?  

Added a test for accuracy on the entire to_csv method for inherited option to see if it matches the slow algorithm.

rikeoka:
All tests green now.

app/presenters/user_csv_presenter.rb
```diff
@@ -0,0 +1,99 @@
...
+    ["id", "name", "email", "projectiD", "project", "role", "deleted at"]
+  end
+
+  ## Returns csv line for CSV Report
+  ## Params
+  ## - user: User Object
+  ## - project: Project object or nil for System level role
+  ## - project_role_id: prefetched user_project_role for project
+  ## - project_role_id is prefetched inside of to_csv to optimize to O(N) from O(N*M) using native
+  ##   ActiveRecord methods
```
grosser: ❤️ great docs!

grosser:
👍 looks great :)

PR #1134, opened 2016-07-15
Human readable job/deploy duration

adammw:
:v:

/cc @sandlerr @danielbreves @zendesk/samson

### Description
Replace process duration with human-readable started time, updated every minute (client side).

**before**:
screen shot 2016-07-15 at 2 46 27 pm

**after**:
screen shot 2016-07-15 at 2 43 49 pm
screen shot 2016-07-15 at 2 44 49 pm

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: 

### Risks
- Level: Low - UI change

### DIFF
--- a/app/assets/javascripts/streams.js
+++ b/app/assets/javascripts/streams.js
@@ -36,6 +36,7 @@ function startStream() {
       var data = JSON.parse(e.data);
 
       $('#header').html(data.html);
+      timeAgoFormat();
       window.document.title = data.title;
       if (doNotify && data.notification !== undefined) {
         var notification = new Notification(data.notification, {icon: '/favicon.ico'});
--- a/app/assets/javascripts/streams.js
+++ b/app/assets/javascripts/streams.js
@@ -8,7 +8,10 @@ function timeAgoFormat() {
   });
 }
 
-$(document).ready(timeAgoFormat);
+$(document).ready(function() {
+  timeAgoFormat();
+  setInterval(timeAgoFormat, 60000); // update times every 60s
+});
 
 function startStream() {
   $(document).ready(function() {
--- a/app/helpers/status_helper.rb
+++ b/app/helpers/status_helper.rb
@@ -22,7 +22,16 @@ module StatusHelper
   def status_panel(deploy)
     content = h deploy.summary
     if deploy.active?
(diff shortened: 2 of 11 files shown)


app/helpers/status_helper.rb
```diff
@@ -22,7 +22,16 @@ def status_label(key)
   def status_panel(deploy)
     content = h deploy.summary
     if deploy.active?
-      content << content_tag(:ul, content_tag(:li, deploy.summary_for_process))
+      content << content_tag(:br)
+      content << "Started "
+      content << content_tag(
```
grosser: don't we have some helper for that ? ... maybe create one so it's more descriptive ?

grosser:
👍 ... pah humans ... why do we have to optimize for them ...

app/assets/javascripts/streams.js
```diff
@@ -8,7 +8,10 @@ function timeAgoFormat() {
   });
 }
 
-$(document).ready(timeAgoFormat);
+$(document).ready(function() {
+  timeAgoFormat();
+  setInterval(timeAgoFormat, 60000); // update times every 60s
```
sandlerr: could clear this when the job is done

adammw: this seems to be global, time ago is used in a couple of places, and it doesn't particularly hurt them actually being up-to-date... (is my thinking)

app/helpers/status_helper.rb
```diff
@@ -22,7 +22,16 @@ def status_label(key)
...
+        :span,
+        deploy.start_time.rfc822,
+        data: { time: datetime_to_js_ms(deploy.start_time) },
+        class: 'mouseover'
+      )
+      job = deploy.respond_to?(:job) ? deploy.job : deploy
```
sandlerr: I wonder if there is a more clear way to say that this can be called with either a job or a deploy

sandlerr:
:+1:

adammw:
💇  @grosser - I did find the helper, but not where I expected so cleaned it up a bit...

app/assets/javascripts/streams.js
sandlerr: makes sense, the value will continue to change until it hits `about an hour`

grosser:
👍

PR #1135, opened 2016-07-15
Fix the @here notification for Slack

andreionut:
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/plugins/slack_webhooks/app/models/slack_webhook_notification.rb
+++ b/plugins/slack_webhooks/app/models/slack_webhook_notification.rb
@@ -17,7 +17,8 @@ class SlackWebhookNotification
 
   def default_buddy_request_message
     project = @deploy.project
-    ":pray: @here _#{@deploy.user.name}_ is requesting approval" \
+    # https://api.slack.com/docs/message-formatting
+    ":pray: <!here> _#{@deploy.user.name}_ is requesting approval" \
       " to deploy #{project.name} *#{@deploy.reference}* to #{@deploy.stage.name}.\n"\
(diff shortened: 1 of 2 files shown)


grosser:
👍 weird ...

andreionut:
what's weird?

grosser:
that it's not @here ... but some strange html syntax

andreionut:
`slackml`

plugins/slack_webhooks/app/models/slack_webhook_notification.rb
```diff
@@ -17,7 +17,8 @@ def buddy_request(message)
 
   def default_buddy_request_message
     project = @deploy.project
-    ":pray: @here _#{@deploy.user.name}_ is requesting approval" \
+    # https://api.slack.com/docs/message-formatting
+    ":pray: <!here> _#{@deploy.user.name}_ is requesting approval" \
```
zehua: This does not seem to work on the Slack desktop client. It is showing verbatim `<!here>`. Might be due to html encoding of `<` `>`.

grosser: can you fix it ?

zehua: I attempted to format the the deploy url nicely with proper slack message formatting but failed due to the same issue. I could give it a second try. :)

andreionut:
Proper fix is in here: https://github.com/zendesk/samson/pull/1208

PR #1137, opened 2016-07-19
Irwaters/paas 268

irwaters:
re-factor the vault client so that it can use a token or a cert/key-pair.  also change it so that it can consume a config file.  this is required so that we can have multile vault servers each with their own configs/tokens/certs etc.

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:
https://zendesk.atlassian.net/browse/PAAS-268
### Risks
Low

### DIFF
--- a/config/initializers/vault.rb
+++ b/config/initializers/vault.rb
@@ -15,7 +15,7 @@ if ENV["SECRET_STORAGE_BACKEND"] == "SecretStorage::HashicorpVault"
 
     def initialize
       return if Rails.env.test?
-      config_valid?
+      ensure_config_exists
       # since we have a bunch of servers, don't use the client singlton to
       # talk to them
       # as we configure each client, check to see if the config has a token in it,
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/test/test_helper.rb
```diff
@@ -64,6 +66,12 @@ def to_h
       instance_values.symbolize_keys
     end
   end
+
+  private
+
+  def config_valid?
+    raise "vault.json is required for #{ENV["SECRET_STORAGE_BACKEND"]}" unless File.exist?("vault.json")
```
grosser: use CONFIG_PATH here twice

plugins/kubernetes/test/test_helper.rb
```diff
@@ -64,6 +66,12 @@ def to_h
       instance_values.symbolize_keys
     end
   end
+
+  private
+
+  def config_valid?
```
grosser: not really a boolean method ... how about `def ensure_config_exists`

config/initializers/vault.rb
```diff
@@ -70,8 +81,14 @@ def self.auth_token(vault_server)
 
     private
 
+    def config_valid?
+      unless File.exist?(Rails.root.join(CONFIG_PATH))
+        raise "config/vault.json is required for #{ENV["SECRET_STORAGE_BACKEND"]}"
```
grosser: use CONFIG_PATH here to avoid repeating

grosser:
👍

plugins/kubernetes/test/test_helper.rb
```diff
@@ -64,6 +66,12 @@ def to_h
       instance_values.symbolize_keys
     end
   end
+
+  private
+
+  def ensure_config_exists
+    raise "vault.json is required for #{CONFIG_PATH}" unless File.exist?(CONFIG_PATH)
```
grosser: vault.json is required for vault.json ? :D

PR #1138, opened 2016-07-20
during a deploy we should blow up when a config is invalid instead of…

grosser:
@jonmoter @irwaters 

atm when a role config is invalid that role is silently not deployed ...

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role.rb
@@ -123,11 +123,7 @@ module Kubernetes
         files = files.split("\n").grep(/\.(yml|yaml|json)$/).map { |f| "#{path}/#{f}" }
         files.map do |path|
           file_contents = project.repository.file_content path, git_ref
-          begin
-            Kubernetes::RoleConfigFile.new(file_contents, path)
-          rescue Samson::Hooks::UserError
-            nil
-          end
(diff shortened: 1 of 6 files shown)


plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -40,7 +40,7 @@ class Role < ActiveRecord::Base
 
     # create initial roles for a project by reading kubernetes/*{.yml,.yaml,json} files into roles
     def self.seed!(project, git_ref)
-      kubernetes_config_files_in_repo(project, git_ref).each do |config_file|
+      kubernetes_config_files_in_repo(project, git_ref, ignore_errors: true).each do |config_file|
```
irwaters: do we want to ignore errors here?

irwaters:
lgtm other than my inline quetsion.

👍

plugins/kubernetes/app/models/kubernetes/role.rb
grosser: while seeding we take what we can, users cannot deploy with missing roles anyway ...
but I guess we might as well show the error right there instead of letting you get into a weird state where you cannot deploy ...

grosser: updated

PR #1141, opened 2016-07-20
Remove undockerizable characters from image names

ben:
This is happening when I try to deploy my project using the ref `ben/my-feature`. Docker tags [cannot have slashes](https://docs.docker.com/engine/reference/commandline/tag/):

> A tag name may contain lowercase and uppercase characters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

Happily, Rails’ `parameterize` function will produce results that fit these restrictions.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low. This happens at build time, and should prevent certain classes of poorly-surfaced error.

### DIFF
--- a/plugins/docker_binary_builder/app/models/binary_builder.rb
+++ b/plugins/docker_binary_builder/app/models/binary_builder.rb
@@ -140,7 +140,7 @@ class BinaryBuilder
   end
 
   def image_name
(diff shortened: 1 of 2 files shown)


grosser:
👍

PR #1142, opened 2016-07-21
allow post/delete requests from the console

grosser:
@henders used this to debug macros issue

### DIFF
--- a/lib/samson/console_extensions.rb
+++ b/lib/samson/console_extensions.rb
@@ -12,6 +12,9 @@ module Samson
         define_method :login_user do |&block|
           block.call
         end
+
+        define_method :verify_authenticity_token do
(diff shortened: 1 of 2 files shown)


lib/samson/console_extensions.rb
```diff
@@ -12,6 +12,9 @@ def login(user)
         define_method :login_user do |&block|
           block.call
         end
+
+        define_method :verify_authenticity_token do
```
irwaters: should this be empty?

irwaters:
👍  other than inline comment.

lib/samson/console_extensions.rb
grosser: yes, it's a noop now

PR #1143, opened 2016-07-21
show errors on staging

grosser:
debugging staging


sandlerr:
is something eating errors before they make it to the log file?

grosser:
this is showing them in the UI, not the log files -> faster debugging

sandlerr:
alright, thought this might have been a workaround for inadequate logs :+1:

grosser:
logs were fine ... but constantly going to the server is annoying ...
the airbrake UI change alone would fix that ... but why not be even lazier :D

PR #1144, opened 2016-07-21
Show the airbrake ID to end-users for easier debugging

grosser:
Loads airbrake and insert middleware on staging/production

screen shot 2016-07-21 at 11 31 12 am

To test locally:
 - set AIRBRAKE_API_KEY
 - remove if staging/production checks in application.rb
 - uncomment config.development_environments
 - raise an exception in a controller
 - set config.consider_all_requests_local = false

@zendesk/samson 

users without airbrake account will see a login message ... but I think that's fine and not crazy confusing but gives us the great benefit of being able to click right through to the backtrace

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -1,8 +1,10 @@
 from_cap = !defined?(Rails)
-if from_cap || Rails.env.staging? || Rails.env.production?
-  require 'airbrake'
+if from_cap || defined?(Airbrake)
   Airbrake.configure do |config|
     config.api_key = ENV['AIRBRAKE_API_KEY']
+    config.user_information = # replaces <!-- AIRBRAKE ERROR --> on 500 pages
+      "<br/><br/>Error number: <a href='https://airbrake.io/locate/{{error_id}}'>{{error_id}}</a>"
+    # config.development_environments = ['test'] # uncomment to report in development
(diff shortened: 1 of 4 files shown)


config/initializers/airbrake.rb
```diff
@@ -1,8 +1,10 @@
 from_cap = !defined?(Rails)
-if from_cap || Rails.env.staging? || Rails.env.production?
-  require 'airbrake'
+if from_cap || defined?(Airbrake)
   Airbrake.configure do |config|
     config.api_key = ENV['AIRBRAKE_API_KEY']
```
irwaters: Add AIRBRAKE_API_KEY to env.example?

irwaters:
looks good other than the docs note.  not sure if tests are required etc...

👍

config/initializers/airbrake.rb
grosser: good idea!

grosser:
we don't have the middlewares setup in integration, so it's not very well testable ... tried it and the integration test did not even render the 500 page, but raised ... so I don't know how to test this ...

irwaters:
it won’t break anything I’m guessing, worst case would be some funny UI
stuff (asking for login as you point out).  I don’t think it’s necessary
(IMHO).

PR #1145, opened 2016-07-21
PAAS-278 Show an example YAML file for a project they can copy/paste

grosser:
@zendesk/paas


grosser:
screen shot 2016-07-21 at 1 10 05 pm

grosser:
screen shot 2016-07-21 at 1 33 48 pm

plugins/kubernetes/app/views/kubernetes/roles/example.html.erb
```diff
@@ -0,0 +1,69 @@
...
+      project: <%= project_label %>
+      role: app-server
+  strategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      name: <%= project_label %>-app-server
+      labels:
+        project: <%= project_label %>
+        role: app-server
```
jonmoter: Could you add some example annotations here for secrets, with a comment on how they can be used?

jonmoter:
Thanks!  👍

PR #1146, opened 2016-07-22
update dependencies, refactor dockerfile

sandlerr:
cc @zendesk/samson @grosser 

The Rails changes are quite small

### Risks
- [low] deploy failures due to changed javascript libs


Dockerfile
```diff
@@ -1,38 +1,36 @@
...
 
 # Mostly static
-ADD config.ru /app/
-ADD Rakefile /app/
-ADD bin /app/bin
-ADD public /app/public
-ADD db /app/db
-ADD .env.bootstrap /app/.env
-ADD .ruby-version /app/.ruby-version
+COPY config.ru /app/
```
grosser: does the same, right ?

grosser:
👍

Dockerfile
sandlerr: `ADD` can be surprising because it will decompress tarballs into the layer. [`COPY` is preferred](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/add-or-copy)

PR #1148, opened 2016-07-22
Additional checks for a valid pull request

anthonywoo:
When using [samson review] - we're inspecting github `pull_request` events. However - we would sometimes get multiple deploys with the same `sha`.

screen shot 2016-07-22 at 11 39 59 am

This happens because github will send a pull request event to samson if the pr description gets updated, if someone adds a label, and much more. 

We only want samson to trigger a deploy when there is a new code push, or if someone explicitly adds the text [samson review]. To achieve this - the only events we will now consider are:  `opened`, `edited` or `synchronized`. In addition, if it's an edit event - we'll make sure the previous description did not have `[samson review]` in the description.

/cc @zendesk/samson @mwerner 

### Tasks
 - [x] :+1: from team

### Risks
- Level: Low. Samson could potentially not deploy the pull requests we expect

### DIFF
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -50,7 +50,7 @@ class Changeset::PullRequest
   # should only be when the edit is related to adding the text [samson review]
   def self.valid_webhook?(params)
     data = params['pull_request'] || {}
-    action = params.fetch('github').fetch('action')
+    action = params.dig('github', 'action')
     return false unless data['state'] == 'open' && (VALID_ACTIONS.include? action)
 
     if action == 'edited'
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -16,6 +16,12 @@ class Changeset::PullRequest
 
   # Matches "VOICE-1234" only
   JIRA_CODE = /(?<=#{PUNCT}|^)(#{CODE_ONLY})(?=#{PUNCT}|$)/
+
+  # Github pull request events can be triggered by a number of actions such as 'labeled', 'assigned'
+  # Actions which aren't related to a code push should not trigger a samson deploy.
+  # Docs on the pull request event: https://developer.github.com/v3/activity/events/types/#pullrequestevent
+  VALID_ACTIONS = ['opened', 'edited', 'synchronized'].freeze
+
   # Finds the pull request with the given number.
   #
   # repo   - The String repository name, e.g. "zendesk/samson".
@@ -38,9 +44,19 @@ class Changeset::PullRequest
     new(project.github_repo, data)
   end
 
+  # Webhook events that are valid should be related to a pr code push or someone adding [samson review]
+  # to the description. The actions related to a code push are 'opened' and 'synchronized'
+  # The 'edited' action gets sent when the PR description is edited. To trigger a deploy from an edit - it
+  # should only be when the edit is related to adding the text [samson review]
   def self.valid_webhook?(params)
     data = params['pull_request'] || {}
-    return false unless data['state'] == 'open'
+    action = params.fetch('github').fetch('action')
+    return false unless data['state'] == 'open' && (VALID_ACTIONS.include? action)
+
+    if action == 'edited'
+      previous_desc = params['github']['changes']['body']['from']
+      return false if previous_desc =~ WEBHOOK_FILTER && data['body'] =~ WEBHOOK_FILTER
+    end
 
     !(data['body'] =~ WEBHOOK_FILTER).nil?
   end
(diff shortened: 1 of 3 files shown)


app/models/changeset/pull_request.rb
```diff
@@ -40,11 +40,20 @@ def self.changeset_from_webhook(project, params = {})
 
   def self.valid_webhook?(params)
     data = params['pull_request'] || {}
-    return false unless data['state'] == 'open'
+    return false unless data['state'] == 'open' && valid_action?(params['github']['action'])
+
+    if params['github']['action'] == 'edited'
```
grosser: `action = params['github']['action']` to reuse it ... maybe even 
`params.require(:github).require(:action)` for nicer errors

app/models/changeset/pull_request.rb
```diff
@@ -40,11 +40,20 @@ def self.changeset_from_webhook(project, params = {})
...
+      previous_desc = params['github']['changes']['body']['from']
+      return false if previous_desc =~ WEBHOOK_FILTER && data['body'] =~ WEBHOOK_FILTER
+    end
 
     !(data['body'] =~ WEBHOOK_FILTER).nil?
   end
 
+  def self.valid_action?(action)
+    ['opened', 'edited', 'synchronized'].include? action
```
grosser: extract into VALID_ACTIONS ... and then .include? on it ?

test/models/changeset/pull_request_test.rb
```diff
@@ -41,6 +41,40 @@
...
+          state: 'open',
+          body: 'pr description [samson review]'
+        },
+        github: {
+          action: 'opened'
+        }
+      }.with_indifferent_access
+    end
+
+    it "is not valid if the github action is labeled" do
```
grosser: would be nice to have more user-level description like `is invalid for a PR that had its label changed`

app/models/changeset/pull_request.rb
grosser: some comment here on why we ignore some actions could be helpful for future archeologists

app/models/changeset/pull_request.rb
```diff
@@ -40,11 +40,20 @@ def self.changeset_from_webhook(project, params = {})
 
   def self.valid_webhook?(params)
```
grosser: comments here on why/what we ignore would also be helpful

grosser:
code looks ok ... just needs more docs :)

app/models/changeset/pull_request.rb
mwerner: It might be worthwhile to either put this `include?` in the `valid_webhook?` method or move the `data['state'] == 'open'` check here. Not sure we're getting a lot of benefit out of moving the single `include?` call into its own method

app/models/changeset/pull_request.rb
```diff
@@ -38,9 +44,19 @@ def self.changeset_from_webhook(project, params = {})
...
   end
 
+  # Webhook events that are valid should be related to a pr code push or someone adding [samson review]
+  # to the description. The actions related to a code push are 'opened' and 'synchronized'
+  # The 'edited' action gets sent when the PR description is edited. To trigger a deploy from an edit - it
+  # should only be when the edit is related to adding the text [samson review]
   def self.valid_webhook?(params)
     data = params['pull_request'] || {}
-    return false unless data['state'] == 'open'
+    action = params.fetch('github').fetch('action')
```
sandlerr: we just moved to ruby 2.3, so you can use [this](http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig) if you like

anthonywoo:
💇 Updated PR description with why we're only looking at certain PR events now

@mwerner @grosser

grosser:
👍

mwerner:
+1

app/models/changeset/pull_request.rb
```diff
@@ -38,9 +44,19 @@ def self.changeset_from_webhook(project, params = {})
...
+  # The 'edited' action gets sent when the PR description is edited. To trigger a deploy from an edit - it
+  # should only be when the edit is related to adding the text [samson review]
   def self.valid_webhook?(params)
     data = params['pull_request'] || {}
-    return false unless data['state'] == 'open'
+    action = params.dig('github', 'action')
+    return false unless data['state'] == 'open' && (VALID_ACTIONS.include? action)
+
+    if action == 'edited'
+      previous_desc = params['github']['changes']['body']['from']
```
sandlerr: nitpick: you can also `dig` this value although it won't crash if one of the nested hashes is unexpectedly `nil` - depends on the behaviour you prefer

PR #1151, opened 2016-07-22
show errors directly in build logs

grosser:
@zendesk/samson

not seeing why a build failed is a bit annoying ... now we handle the backtrace output the same way we handle server errors


grosser:
screen shot 2016-07-22 at 4 14 38 pm

irwaters:
👍  lgtm

PR #1153, opened 2016-07-23
show aibrake url in build logs

grosser:
when an error happens in production / when backtraces are hidden we still want users to be able to talk to developers efficiently

screen shot 2016-07-22 at 7 33 20 pm

@zendesk/samson


irwaters:
like the "do not spam us with users issues" bit

👍

PR #1154, opened 2016-07-23
Allow specifying a CA cert for Vault clients

jonmoter:
* Add `ca_cert` attribute to `vault.json` file
* Allow specifying a vault config file via ENV var, rather than hard-coding

/cc @zendesk/samson, @irwaters 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


grosser:
:+1:

PR #1157, opened 2016-07-24
improve the output when deploying Kubernetes stages

jonmoter:
* Don't write output every 2 seconds
* Make the output of each line different, so you can tell it's outputting new info
* Limit output of container logs to last 50 lines

(Will create screenshots after I test it out.)

/cc @zendesk/paas

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -44,7 +44,7 @@ module Kubernetes
     private
 
     def wait_for_resources_to_complete(release, release_docs)
-      start = Time.now
+      @wait_start_time = Time.now
       stable_ticks = CHECK_STABLE / TICK
       expected = release_docs.to_a.sum(&:desired_pod_count)
       @output.puts "Waiting for #{expected} pods to be created"
@@ -76,7 +76,7 @@ module Kubernetes
           elsif statuses.map(&:details).include?(RESTARTED)
             unstable!
             return false
-          elsif start + WAIT_FOR_LIVE < Time.now
+          elsif seconds_waiting > WAIT_FOR_LIVE
             @output.puts "TIMEOUT, pods took too long to get live"
             return false
           end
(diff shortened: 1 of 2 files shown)


irwaters:
should probably add KUBE_WAIT_FOR_LIVE docs, otherwise 👍

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -76,7 +76,7 @@ def wait_for_resources_to_complete(release, release_docs)
           elsif statuses.map(&:details).include?(RESTARTED)
             unstable!
             return false
-          elsif start + WAIT_FOR_LIVE < Time.now
+          elsif seconds_waiting > WAIT_FOR_LIVE
```
grosser: nice refactor

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -320,5 +324,9 @@ def success
       @output.puts "SUCCESS"
       true
     end
+
+    def seconds_waiting
+      (Time.now - @wait_start_time).to_i if @wait_start_time
```
grosser: do we need that if / would that ever happen ?

plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
```diff
@@ -424,7 +424,7 @@ def worker_is_unstable
 
         refute execute!
 
-        out.must_include "LOGS:\nLOG-1\n"
+        out.must_match /LOGS:\s+LOG-1/
```
grosser: I liked the '\n' assertion more ... this might just be a space and would look weird ...

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -181,10 +182,13 @@ def release_statuses(pods, release_doc)
     end
 
     def print_statuses(status_groups)
+      return if @last_status_output && @last_status_output > 10.seconds.ago
```
grosser: this would also not print the status if it changed ?

PR #1158, opened 2016-07-25
Always show the redeploy button

sandlerr:
@zendesk/samson @timbertson 

- move `jquery-ui` higher so its tooltip method doesn't conflict with bootstrap
- show a redeploy button with different styling on successful deploys

### DIFF
--- a/app/helpers/deploys_helper.rb
+++ b/app/helpers/deploys_helper.rb
@@ -99,12 +99,13 @@ module DeploysHelper
     end
   end
 
-  def redeploy_button(already_succeeded = false)
+  def redeploy_button
+    return if @deploy.active?
     html_options = {
       class: 'btn btn-danger',
       method: :post
     }
-    if already_succeeded
+    if @deploy.succeeded?
       html_options[:class] = 'btn btn-default'
       html_options[:data] = {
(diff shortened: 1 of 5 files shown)


app/helpers/deploys_helper.rb
```diff
@@ -99,6 +99,28 @@ def stages_select_options
     end
   end
 
+  def redeploy_button(already_succeeded = false)
```
grosser: call with true/false instead of having a default argument ?

grosser:
screenshot ?

... I often wanted to deploy the same branch again, so this would be useful, I just felt like the UI would be confusing since it offers an option that 99% of the time makes no sense :(

sandlerr:
screen shot 2016-07-26 at 9 29 15 am

sandlerr:
@grosser refactored. Agree that 99% of the time we don't want to redeploy a successful deploy, but I don't want to put an 'are you sure' modal in the way since the whole point is to allow iteratively testing a deploy script as easily as possible.

grosser:
how about we do not show it when ref is `/^v\d+/` ? :)

grosser:
👍

grosser:
a test ?

PR #1159, opened 2016-07-25
adding every folder in lib to the load path makes them reachable via …

grosser:
@bolddane 

imagine lib/bar/foo.rb and lib/baz/foo.rb ... now foo.rb can be loaded with 2 different requires ... bad ... also more load paths to look through -> slow (we have bootscale, so that does not matter, but for other projects it would be bad)


pschambacher:
👍  but better test on staging before production deploy

PR #1160, opened 2016-07-25
Fix docker job status

### LINKED ISSUE #1150: Docker Job Status is shown as failed even when it runs ok
Hi,

I'm using samson on tag 771 and the build status is always shown as failed when it finishes okay. The container is created okay and the image is pushed just fine, but the ```Docker Job Status: failed``` is shown on the `builds` tab (and when entering the build and seeing the details).

The build can be deployed just fine when it finished okay (although it is marked as failed), but the status is shown incorrectly.

This happens in our older version of samson too. We are using docker 1.11.2 to build docker images (so the output is from that docker version) and we are not using kubernetes remote job for building.

Do you see it too? Can I ask which docker version are you using? It's not easy for us to try another version, as it was hitting a kernel bug :-/

rata:
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

Fixes #1150

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -25,8 +25,10 @@ class DockerBuilderService
       if build.kubernetes_job
         run_build_image_job(job, image_name, push: push, tag_as_latest: tag_as_latest)
       elsif build_image(tmp_dir)
-        push_image(image_name, tag_as_latest: tag_as_latest) if push
+        ret = true
+        ret = push_image(image_name, tag_as_latest: tag_as_latest) if push
         build.docker_image.remove(force: true) unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
+        ret
       end
     end
 


grosser:
hmm it's false if we don't push ?

rata:
On Mon, Jul 25, 2016 at 01:54:05PM -0700, Michael Grosser wrote:
> hmm it's false if we don't push ?

I thought you wanted that. `true` is better? :)

grosser:
I'd think so ... if we don't want to push it's still successful ...

Did you try adding a test ?

rata:
@grosser fixed the styling issue, but not sure how to properly write tests (never done that before). Found the `test/models/docker_builder_service_test.rb` file and tried to hack something based on what already is there, but didn't make it work :-/

grosser:
added tests https://github.com/zendesk/samson/pull/1164

PR #1161, opened 2016-07-25
take user back after creating a deploy group role via stage UI

grosser:
screen shot 2016-07-25 at 2 45 11 pm

click create -> new + redirect_to param
create -> success ? redirect-to-param : show error
-> arrive back where you started

@jonmoter 

/cc @zendesk/samson for a generally useful helper

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -23,4 +23,18 @@ class ApplicationController < ActionController::Base
     super
     payload["params"] = request.params
   end
+
+  def redirect_back_or(fallback)
+    if param_location = params[:redirect_to].to_s.presence
+      if param_location.is_a?(String) && param_location.start_with?('/')
+        redirect_to URI("http://nope.nope#{param_location}").request_uri # using URI to silence Brakeman
+      else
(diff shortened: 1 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -23,4 +23,18 @@ def append_info_to_payload(payload)
     super
     payload["params"] = request.params
   end
+
+  def redirect_back_or(fallback)
+    if param_location = params[:redirect_to].to_s.presence
+      if param_location.is_a?(String) && param_location.start_with?('/')
+        redirect_to URI("http://nope.nope#{param_location}").request_uri # using URI to silence Brakeman
```
jonmoter: Why the `nope.nope`?

grosser: need a fake host for URI parsing to work

jonmoter: Ah, I thought `#request_uri` returned the the host too.  That's a little weird, but makes sense.

jonmoter:
👍

app/controllers/application_controller.rb
grosser: yeah .. it's a weird method ...

PR #1162, opened 2016-07-25
allow users to stay in the flow and keep creating multiple secrets

grosser:
@jonmoter @irwaters 

on new page add a new button:

screen shot 2016-07-25 at 3 55 31 pm

and when saving works it will prefill the next form:

screen shot 2016-07-25 at 3 57 06 pm

### DIFF
--- a/app/controllers/admin/secrets_controller.rb
+++ b/app/controllers/admin/secrets_controller.rb
@@ -1,4 +1,6 @@
 class Admin::SecretsController < ApplicationController
+  ADD_MORE = 'Save and add another'.freeze
+
   include CurrentProject
 
   before_action :find_project_permalinks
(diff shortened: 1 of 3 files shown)


app/controllers/admin/secrets_controller.rb
```diff
@@ -58,7 +60,11 @@ def project_permalink
 
   def successful_response(notice)
     flash[:notice] = notice
-    redirect_to action: :index
+    if params[:commit] == ADD_MORE
+      redirect_to new_admin_secret_path(secret: params[:secret].except(:value))
```
jonmoter: How about `except(:key, value)`?  The user probably wants to add a different key.

jonmoter:
One comment and a failing Travis test.

Otherwise, 👍 on 💚

app/controllers/admin/secrets_controller.rb
grosser: yeah ... was thinking that too ... but then I thought they might be adding the same key to multile envs/deploy-groups ... so I opted for keeping it

PR #1163, opened 2016-07-26
OAuth w/ DoorKeeper and First Class API

apanzerj:
### Description
This adds [Doorkeeper](https://github.com/doorkeeper-gem/doorkeeper) to Samson along with an integration with warden called [warden-doorkeeper](https://github.com/kolorahl/warden-doorkeeper). This will allow Samson to be an OAuth provider. In addition, I moved 1 API endpoint as a MVP. Next PR should start porting over as many API/JSON routes.

/cc @zendesk/samson

### Screenshots
New menu items:

screenshot 2016-08-01 12 35 04

Applications List:

screenshot 2016-08-01 12 36 58

Application Detail Page:

screenshot 2016-08-01 12 38 09

### Tasks
 - [x] Restrict token generation to SuperAdmin
 - [x] Restyle pages to look like samson
 - [ ] :+1: from team

### Risks
- Level: Medium
Authentication might not work


Gemfile
```diff
@@ -14,6 +14,8 @@ group :preload do
   gem 'active_model_serializers'
   gem 'paper_trail'
   gem 'soft_deletion'
+  gem 'doorkeeper'
+  gem 'pry'
```
grosser: does this need to be preloaded ?

app/controllers/api/base_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::BaseController < ActionController::Base
+  before_action :doorkeeper_authorize!
+  before_action :log_api
+
+
```
grosser: whitespace

app/controllers/api/base_controller.rb
```diff
@@ -0,0 +1,9 @@
...
+  def log_api
+    Rails.logger.warn("API Request: #{env['REQUEST_PATH']} / #{env['REQUEST_METHOD']}")
+  end
+end
```
grosser: change your editor to auto-add trailing whitespace

app/controllers/api/base_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::BaseController < ActionController::Base
+  before_action :doorkeeper_authorize!
+  before_action :log_api
+
+
+  def log_api
```
grosser: private

app/controllers/api/base_controller.rb
```diff
@@ -0,0 +1,9 @@
...
+    Rails.logger.warn("API Request: #{env['REQUEST_PATH']} / #{env['REQUEST_METHOD']}")
```
grosser: seems useless ... we already log the controller/path

app/controllers/api/deploys_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Api::DeploysController < Api::BaseController
```
grosser: might want to add v1 ...

app/controllers/api/deploys_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  include CurrentProject
+
+  def index
+    scope = current_project.try(:deploys) || Deploy
+    @deploys =
+      if params[:ids]
+        Kaminari.paginate_array(scope.find(params[:ids])).page(1).per(1000)
```
grosser: should fetch all when you give it a set of ids ? ... bad_request when you give it > 100

config/initializers/warden.rb
```diff
@@ -8,8 +8,9 @@
 
 require 'warden/strategies/basic_strategy'
 require 'warden/strategies/session_strategy'
+require "warden/strategies/doorkeeper"
 
 Rails.application.config.middleware.insert_after(ActionDispatch::Flash, Warden::Manager) do |manager|
-  manager.default_strategies :basic, :session
+  manager.default_strategies :doorkeeper, :basic, :session
```
grosser: do we want doorkeeper for regular urls too ?

config/locales/doorkeeper.en.yml
```diff
@@ -0,0 +1,124 @@
+en:
```
grosser: does it fail when you remove this file ? <-> does it not have defaults ?

app/controllers/api/deploys_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Api::DeploysController < Api::BaseController
+  include CurrentProject
+
+  def index
+    scope = current_project.try(:deploys) || Deploy
```
grosser: only reachable under projects, so no try necessary ?

db/schema.rb
```diff
@@ -146,7 +146,7 @@
     t.integer  "user_id",    limit: 4,                              null: false
     t.integer  "project_id", limit: 4,                              null: false
     t.string   "status",     limit: 255,        default: "pending"
-    t.text     "output",     limit: 1073741823
+    t.text     "output",     limit: 4294967295
```
grosser: only update the tables you want or ci/migrate will fail

lib/warden/strategies/doorkeeper_strategy.rb
```diff
@@ -0,0 +1,3 @@
+require "warden"
+
+Warden::Strategies.add(:doorkeeper, Warden::Strategies::Doorkeeper)
```
grosser: this file is strange ... should be in the initializer ?

app/controllers/api/base_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::BaseController < ActionController::Base
```
grosser: add tets for this guy ... especially with the doorkeeper auth :)

config/initializers/warden.rb
apanzerj: how do these two things relate?

config/locales/doorkeeper.en.yml
apanzerj: This is what came with the generator. Do you not I18n in Samson?

app/controllers/api/deploys_controller.rb
apanzerj: This is copy pasta from deploys_controller.rb I need to re-extract to serializers.

db/schema.rb
apanzerj: got it.

lib/warden/strategies/doorkeeper_strategy.rb
apanzerj: Nope. This is the pattern for how it's done. See the basic_strategy.

app/controllers/api/base_controller.rb
apanzerj: Will do!

config/locales/doorkeeper.en.yml
grosser: try removing and see if it is still translated ...
we don't use I18n in samson

grosser: If it no longer translates when this is gone, then just keep it ... just trying to avoid copy-paste and upgrade pains

config/initializers/warden.rb
grosser: just curious if we want doorkeeper to be used for all endpoints or only the api ... either way is fine, but just checking if you did that on purpose or by accident

Gemfile
apanzerj: nope. I'll drop it out.

app/controllers/api/base_controller.rb
apanzerj: it's gone.

app/controllers/api/base_controller.rb
apanzerj: zapped.

config/initializers/warden.rb
apanzerj: We want doorkeeper to only work on api urls. 

I can look into ways of blocking it otherwise.

grosser: the other api PR had a solution for this

apanzerj:
Okay! Next up, skinning!

app/assets/javascripts/controllers/current_deploys_badge_ctrl.js
```diff
@@ -8,7 +8,7 @@ samson.controller('currentDeployBadgeCtrl', function($scope, $http, SseFactory)
   // - get current active count to reduce race condition
   // - connect to stream for updates
   setTimeout(function(){
-    $http.get('/deploys/active_count.json').success(function(result) {
+    $http.get('/api/deploys/active_count.json').success(function(result) {
```
grosser: remove the old endpoint ?

app/controllers/deploys_controller.rb
```diff
@@ -27,10 +27,6 @@ def active
     render partial: 'shared/deploys_table', layout: false if params[:partial]
   end
 
-  def active_count
```
grosser: nvm

config/routes.rb
```diff
@@ -1,6 +1,15 @@
 Samson::Application.routes.draw do
+  use_doorkeeper
   root to: 'projects#index'
```
grosser: remove old active_count route

app/controllers/deploys_controller.rb
apanzerj: I removed the old route

config/routes.rb
apanzerj: ahahahahah yup

config/database.mysql.yml.example
```diff
@@ -1,10 +1,11 @@
 <%
-  uri = URI(ENV['DATABASE_URL'] || 'mysql://root@127.0.0.1:3306')
+  uri = URI(ENV['DATABASE_URL'] || 'mysql://admin@192.168.42.45:3306')
   uri.scheme = 'mysql2'
 %>
 
 development:
   <% uri.path = '/samson_development' %>
+  password: '123456'
```
grosser: noooo

config/database.mysql.yml.example
```diff
@@ -1,10 +1,11 @@
 <%
-  uri = URI(ENV['DATABASE_URL'] || 'mysql://root@127.0.0.1:3306')
+  uri = URI(ENV['DATABASE_URL'] || 'mysql://admin@192.168.42.45:3306')
```
grosser: also noooo

config/initializers/doorkeeper.rb
```diff
@@ -0,0 +1,25 @@
+module Doorkeeper
+  autoload :AccessGrant, 'doorkeeper/orm/active_record/access_grant'
```
grosser: maybe a comment on why we do this

grosser: `will make rails runner fail because the models were preloaded`

config/initializers/omniauth.rb
```diff
@@ -6,8 +6,8 @@
   if Rails.application.config.samson.auth.github
     require 'omniauth-github'
     provider :github,
-      ENV["GITHUB_CLIENT_ID"],
-      ENV["GITHUB_SECRET"],
+      "89eb0e751d4709682a47",
+      "8ef3144b2cac4c3803d4a905299bb83b395823a0",
```
grosser: noooo :D 

... make sure to squash/remove the commit too

config/locales/doorkeeper.en.yml
grosser: so do we need it ?

db/schema.rb
```diff
@@ -270,6 +270,54 @@
...
+    t.string   "secret",       limit: 255,                null: false
+    t.text     "redirect_uri", limit: 65535,              null: false
+    t.string   "scopes",       limit: 255,   default: "", null: false
+    t.datetime "created_at",                              null: false
+    t.datetime "updated_at",                              null: false
+  end
+
+  add_index "oauth_applications", ["uid"], name: "index_oauth_applications_on_uid", unique: true, using: :btree
+
+  create_table "project_environment_variable_groups", force: :cascade do |t|
```
grosser: this looks suspicious

lib/warden/strategies/doorkeeper_strategy.rb
```diff
@@ -0,0 +1,23 @@
...
+    # Hack because we are not using ActionDispatch
+    r = RequestObject.new(request)
+    return unless request.content_type == 'application/json'
+    return unless r.authorization.present?
+
+    # We want to stop authenticating if you didn't provide a full bearer token.
+    # At this point we have already verified it's not basic auth.
+    # w/o this the response would be a redirect to login. Not appropriate for api request.
+
+    return unless r.authorization.include?("Bearer ")
```
grosser: start_with ?

test/controllers/api/base_controller_test.rb
```diff
@@ -0,0 +1,6 @@
+require_relative '../../test_helper'
+
+SingleCov.not_covered!
```
grosser: `.covered!` ? / document how many are uncovered ... ideally add tests application_controller_test style

test/controllers/api/deploys_controller_test.rb
```diff
@@ -0,0 +1,48 @@
...
+    before do
+      Deploy.stubs(:active).returns(['a'])
+      token.save
+      json!
+      auth!("Bearer #{token.token}")
+      get :active_count
+    end
+
+    it 'responds successfully' do
+      response.success?.must_be :==, true
```
grosser: `assert_response :success`

test/controllers/api/deploys_controller_test.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    it 'responds as json' do
+      response.content_type.must_be :==, 'application/json'
```
grosser: `.must_equal`

test/integration/cleanliness_test.rb
```diff
@@ -88,6 +88,7 @@ def check_content(files)
     untested = [
       "app/controllers/concerns/current_project.rb",
       "app/controllers/concerns/stage_permitted_params.rb",
+      "app/controllers/doorkeeper_base_controller.rb",
```
grosser: add empty test instead so it does not ignore people adding new uncovered code

test/test_helper.rb
```diff
@@ -24,6 +24,7 @@
 require 'mocha/setup'
 
 require 'sucker_punch/testing/inline'
+require 'pry'
```
grosser: remove ?

lib/warden/strategies/doorkeeper_strategy.rb
apanzerj: You told me to use include...

test/integration/cleanliness_test.rb
apanzerj: eh?

test/test_helper.rb
apanzerj: Why? Pry is useful in test env

test/integration/cleanliness_test.rb
apanzerj: ahhh
I get it.

app/assets/javascripts/controllers/current_deploys_badge_ctrl.js
apanzerj: done.

app/controllers/api/deploys_controller.rb
apanzerj: nah.. i have a different plan for this, involving swagger

config/initializers/doorkeeper.rb
apanzerj: done.

config/locales/doorkeeper.en.yml
apanzerj: ehhh. I haven't checked. I'll check one sec

test/test_helper.rb
grosser: Benchmark how much boot time this adds then

apanzerj: n/m. I removed it.

test/controllers/api/deploys_controller_test.rb
apanzerj: done

test/controllers/api/deploys_controller_test.rb
apanzerj: done

test/controllers/api/base_controller_test.rb
apanzerj: done

lib/warden/strategies/doorkeeper_strategy.rb
apanzerj: ಠ_ಠ 


fiiiiine

apanzerj:
@grosser I think that covers all your comments.

apanzerj:
@jonmoter can you take a look as well?

jonmoter:
@rikeoka, can I get your eyes on this from a security perspective?

jonmoter:
By adding this logic, does it mean that only subclasses of `DoorkeeperBaseController` use the Doorkeeper auth strategy?  Or is that auth strategy now valid for any controller?

In other words, does this open up any new auth strategies (and thus attack vectors) for the existing Samson functionality?

lib/warden/strategies/basic_strategy.rb
```diff
@@ -18,16 +22,14 @@ def authenticate!
...
     end
   end
 
   # ActionDispatch's
   def authorization
-    @authorization ||= request.env['HTTP_AUTHORIZATION'] ||
-      request.env['X-HTTP_AUTHORIZATION'] ||
-      request.env['X_HTTP_AUTHORIZATION'] ||
-      request.env['REDIRECT_X_HTTP_AUTHORIZATION']
+    RequestObject.new(request).authorization
```
grosser: make it a static method on ActionDispatchPatch ?

grosser: nvm used somewhere else too ...

grosser:
squasch plz :)

👍

grosser:
for all

apanzerj:
You can't access regular pages with this mechanism. See here: https://github.com/zendesk/samson/pull/1163/files#diff-d5ca7c1db3069f37dba56415bb218c50R1

That basically says: for this to be valid it has to be a JSON request. Once we remove all JSON responses from the application controllers then only requests going to the api controller will work.

jonmoter:
So if I authenticate through Doorkeeper, and then create a request to a non-API endpoint with an `application/json` content type, it will authenticate?

apanzerj:
@jonmoter 

Created a mixin that basically allows you to do this:

```
SomeController < ApplicationController
  api_accessible true
end
```

Now SomeController is accessible using OAuth. If you don't specify or specify false, there's a raise that is called and then rescued to return text saying "This resource is not available via the API"

rikeoka:
Looking into the authorization schemes at this point seems fine.  As we start adding more endpoints we'll have to reevaluate authorization scopes.  Perhaps at this point we make the token scope the same as the user's permissions and we reevaluate authorizations as a whole?

When we start adding in POST,PUT,PATCH,DELETE endpoints, we'll need to make sure the permission schemes are followed.  It looks like we can use the same permissions checks, in addition to oauth scope checks.  

Ideally we wouldn't issue oauth scopes for something a user doesn't have permissions for.  ie.  User is deployer, should not get an OAUTH scope user:admin or user:read for example, an admin should not get an OAUTH scope user:admin, but user:read.  A super admin can get OAUTH scope user:admin or user:read depending on what they want scope for.

I could go through all the different permission checks and propose OAuth scope levels in a confluence page or something?

apanzerj:
At the moment, that's not even possible. There's only one API endpoint, the active count endpoint, which is just for MVP purposes. The API Controllers can still use the same permissions models that the regular controllers use and the oauth scopes can be really broad. 

thoughts?

apanzerj:
@jonmoter this should be green now, and we now have explicit blocking of anything using doorkeeper and trying to get at a non-api route.

apanzerj:
Unrelated flakey test. Rebuild should be green.

apanzerj:
There we go. Ready for review I guess. 

cc @pswadi-zendesk

app/controllers/application_controller.rb
```diff
@@ -1,13 +1,26 @@
...
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   force_ssl if: :force_ssl?
 
+  api_accessible! false
+
   include CurrentUser # must be after protect_from_forgery, so that authenticate! is called
 
+  rescue_from(DoorkeeperAuth::DisallowedAccessError) do |message|
+    if Rails.env.test?
```
sandlerr: could put `rescue_from` inside this conditional, that way you don't need to re-raise

app/controllers/application_controller.rb
```diff
@@ -1,13 +1,26 @@
...
+      raise(DoorkeeperAuth::DisallowedAccessError, message)
+    else
+      render text: "This resource is not available via the API", status: 200
```
sandlerr: maybe the status should be 401?

lib/warden/action_dispatch_patch.rb
```diff
@@ -0,0 +1,24 @@
...
+      @request = request
+    end
+
+    def authorization
+      if @request.is_a?(ActionDispatch::Request)
+        @authorization = @request.authorization
+      elsif @request.respond_to?(:env)
+        @authorization ||= @request.env['HTTP_AUTHORIZATION'] ||
+          @request.env['X-HTTP_AUTHORIZATION'] ||
+          @request.env['X_HTTP_AUTHORIZATION'] ||
```
sandlerr: sometimes it is an underscore sometimes a dash?

test/integration/cleanliness_test.rb
```diff
@@ -45,8 +45,8 @@ def check_content(files)
 
   it "does not have actions on base controller" do
     found = ApplicationController.action_methods.to_a
-    found.reject { |a| a =~ /^(_conditional_callback_around_|_callback_before_)/ }
-    (found - ["flash"]).must_equal []
+    found.reject! { |a| a =~ /^(_conditional_callback_around_|_callback_before_|api_)/ }
+    (found - ["flash", "access_denied?", "disallowed?"]).must_equal []
```
sandlerr: can these methods be made private?

sandlerr:
it would be awesome if the active_count endpoint checked an oauth scope since it might be used as an example followed in other controllers

app/controllers/application_controller.rb
apanzerj: Aaah. Nice. Will do.

app/controllers/application_controller.rb
apanzerj: oh yeah. I was testing and the 401 was raising in my client application.

lib/warden/action_dispatch_patch.rb
apanzerj: Hadn't noticed that, but apparently, yes. (－‸ლ)

test/integration/cleanliness_test.rb
apanzerj: ooooh. I was confused by the test. It should say "does not have public methods on base controller"

apanzerj:
@sandlerr so the active_count endpoint is a bad example all around. I need to find a better one.

app/controllers/application_controller.rb
apanzerj: hmmm... it doesn't rescue anymore. I had a lot of problems getting this code to work/test right.

Test failure: https://travis-ci.org/zendesk/samson/jobs/149806164

Am I missing something?

apanzerj: arg. I am missing something...

apanzerj: eh, nevermind, thought it was the assert_response :success but it's not.

apanzerj: in summary: still confused as to why this doesn't work. help!?! @irwaters any thoughts?

sandlerr: actually, why is the test environment special? You're specifically not rescuing in test, so the error is being raised. That's what I would expect.

sandlerr: it's just that Rails.env got evaluated before your mock.

apanzerj:
@sandlerr so I tried, very hard, to figure out how scopes would be used here. I think scopes aren't really necessary at the moment. Our user auth is fairly fleshed out already and doorkeeper-warden takes scopes from warden itself. This doesn't fit with what we are doing right now because we just use a vanilla `warden.authenticate` and we'd need to include those scopes in there to make it work. 

Doorkeeper is better suited (apparently) to devise than warden and I'm running low on time to get this built up. 

At the moment, this works, and if somebody wants to implement scopes, they can, but this meets the needs for the project I'm on right now. 

Thoughts @zendesk/samson ?

apanzerj:
@jonmoter @grosser @sandlerr can you guys take a look?

app/controllers/application_controller.rb
apanzerj: Yeah, the gain isn't worth the trouble at this point.

lib/doorkeeper_auth.rb
```diff
@@ -0,0 +1,28 @@
...
+  def self.included(base)
+    base.extend ClassMethods
+
+    base.class_attribute :api_accessible
+    base.before_action :access_denied?
+  end
+
+  private
+
+  def access_denied?
```
sandlerr: Based on its name I would expect this method to return a boolean, not raise an error.

app/controllers/api/base_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+class Api::BaseController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  include DoorkeeperAuth
+
+  api_accessible! true
+
+  def paginate(scope)
+    if scope.is_a?(Array)
+      Kaminari.paginate_array(scope).page(page).per(1000)
```
grosser: having an array here seems like a sign of trouble/something gone wrong ... maybe blow up ?

app/controllers/api/deploys_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Api::DeploysController < Api::BaseController
+  include CurrentProject
+
+  skip_before_action :require_project, only: [:active_count]
+
+  def index
+    scope = current_project.try(:deploys) || Deploy
+    @deploys = paginate(params[:ids].present? ? [scope.find(params[:ids])] : scope)
```
grosser: this would produce an array in an array ... add a test ;)
```Ruby
@deploys = current_project.try(:deploys) || Deploy
if ids = paginate(params[:ids].presence
  @deploys = @deploys.find(ids.split(','))
end
```

app/controllers/api/stages_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def index
+    render json: @project.stages
+  end
+
+  def clone
+    new_stage = Stage.build_clone(old_stage)
+    new_stage.name = stage_name
+    if new_stage.save
+      if deploy_group_ids.any? && deploy_groups.any?
```
grosser: only need to check `deploy_groups.any?`

grosser: just make it smart/lazy :)

app/controllers/api/stages_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        new_stage.deploy_groups = deploy_groups
+        new_stage.save
```
grosser: `save!` here ?

app/controllers/api/stages_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+
+      render json: new_stage, status: 201
+    else
+      render json: new_stage.errors, status: 422
```
grosser: I think we should `save!` everywhere and then rescue the invalid error to present the error ...

app/controllers/api/stages_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+  private
+
+  def old_stage
```
grosser: will be misleading once we add more actions ...

config/initializers/doorkeeper.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+    end
+  end
+end
+
+Doorkeeper.configure do
+  orm :active_record
+
+  resource_owner_authenticator do
+    env['warden'].user || redirect_to(login_url)
```
grosser: seems wrong for the api to do redirects ... should this `throw :warden` ?

config/routes.rb
```diff
@@ -1,6 +1,19 @@
...
 
+  namespace :api do
+    resources :deploys, only: [:index, :active_count] do
+      collection do
+        get :active_count
+      end
+    end
+
+    resources :stages, only: [:index, :clone] do
```
grosser: remove clone from only

config/routes.rb
```diff
@@ -1,6 +1,19 @@
...
+      post :clone, to: 'stages#clone'
```
grosser: `post :clone` should already do the trick

lib/warden/strategies/doorkeeper_strategy.rb
```diff
@@ -0,0 +1,18 @@
+require "warden/strategies/doorkeeper"
+require "warden/action_dispatch_patch"
+
+class Warden::Strategies::Doorkeeper < ::Warden::Strategies::Base
+  include ActionDispatchPatch
+
+  def valid?
```
grosser: should we not erm test this somehow ?

lib/warden/strategies/session_strategy.rb
```diff
@@ -4,7 +4,11 @@ def valid?
   end
 
   def authenticate!
-    redirect!('/login', origin: request.path)
+    if request.content_type == 'application/json'
```
grosser: got a test that fails without this ?

test/controllers/api/base_controller_test.rb
```diff
@@ -0,0 +1,12 @@
+require_relative '../../test_helper'
+
+SingleCov.covered!
+
+describe Api::BaseController do
+  describe 'Doorkeeper Auth Status' do
+    subject { Api::BaseController }
+    it 'is allowed' do
```
grosser: maybe also do `is allowed for subclass` to make sure it's inherited properly and `is not allowed for superclass`

test/controllers/api/deploys_controller_test.rb
```diff
@@ -0,0 +1,76 @@
...
+    it 'responds as json' do
+      response.content_type.must_equal 'application/json'
+    end
+
+    it 'returns as expected' do
+      response.body.must_be :==, "1"
+    end
+  end
+
+  describe 'get index' do
```
grosser: `#index`

test/controllers/api/deploys_controller_test.rb
```diff
@@ -0,0 +1,76 @@
...
+    it 'responds successfully' do
+      assert_response :success
+    end
+
+    it 'responds as json' do
+      response.content_type.must_equal 'application/json'
+    end
+
+    it 'returns as expected' do
+      response.body.must_be :==, "1"
```
grosser: .must_equal

test/controllers/api/deploys_controller_test.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+
+      it 'renders 1 deploy' do
+        subject.size.must_equal 1
+      end
+    end
+
+    describe 'when an id is passed in' do
+      before do
+        get :index, project_id: project.to_param, ids: deploy.id
```
grosser: maybe pass 2 comma separated to check everything works

test/controllers/api/stages_controller_test.rb
```diff
@@ -0,0 +1,99 @@
+require_relative '../../test_helper'
+
+SingleCov.covered!
+
+describe Api::StagesController do
+  def assert_changed(model, method, amount)
```
grosser: use assert_difference ?

test/controllers/api/stages_controller_test.rb
```diff
@@ -0,0 +1,99 @@
...
+    initial = model.send(method).to_i
+    yield
+    initial.wont_equal model.send(method), "No change detected."
+    initial.must_equal (model.send(method) - amount), "#{model} was expected to change by #{amount}"
+  end
+
+  def assert_not_changed(model, method, amount = 0)
```
grosser: use refute_difference ?

test/controllers/api/stages_controller_test.rb
```diff
@@ -0,0 +1,99 @@
...
+    initial = model.send(method).to_i
+    yield
+    initial.must_equal model.send(method), "#{model} was expected not to change by #{amount}"
+  end
+
+  oauth_setup!
+  let(:project) { projects(:test) }
+  let(:stages) { project.stages }
+
+  describe 'get index' do
```
grosser: `#index`

test/controllers/api/stages_controller_test.rb
```diff
@@ -0,0 +1,99 @@
...
+      assert_response :success
+      response.content_type.must_equal 'application/json'
+    end
+
+    it 'contains a stage' do
+      subject.size.must_equal 1
+    end
+  end
+
+  describe 'post clone' do
```
grosser: `#clone`

test/controllers/api/stages_controller_test.rb
```diff
@@ -0,0 +1,99 @@
...
+        stage.reload.deploy_groups.map(&:id).must_be :==, [dg.id]
+      end
+    end
+
+    describe 'when the cloned stage is invalid' do
+      before do
+        post :clone, stage_id: stages.first.permalink, stage_name: stages.first.name
+      end
+
+      it 'only creates one clone' do
```
grosser: `does not clone` ?

test/controllers/api/stages_controller_test.rb
```diff
@@ -0,0 +1,99 @@
...
+        assert_not_changed(Stage, :count, 0) do
+          post :clone, stage_id: stages.first.permalink, stage_name: stages.first.name
+        end
+      end
+
+      it 'inlcludes the errors' do
```
grosser: `includes`

test/controllers/api/stages_controller_test.rb
```diff
@@ -0,0 +1,99 @@
...
+        post :clone, stage_id: stages.first.permalink, stage_name: stages.first.name
+        response.body.must_match %r{already been taken}
```
grosser: prefer must_include since it will give nicer error messages

test/controllers/application_controller_test.rb
```diff
@@ -4,6 +4,10 @@
 
 describe ApplicationController do
   class ApplicationTestController < ApplicationController
+    def something
```
grosser: something more descriptive like `def raise_doorkeeper_error` ...

test/controllers/application_controller_test.rb
```diff
@@ -59,4 +63,29 @@ def test_redirect_back_or
...
+      subject { @controller }
+
+      it 'is disallowed' do
+        subject.api_accessible.must_equal false
+      end
+
+      describe 'in test env' do
+        it 'does not rescue' do
+          Rails.env.stubs(:test?).returns(true)
+          proc { get :something, test_route: true }.must_raise(DoorkeeperAuth::DisallowedAccessError)
```
grosser: usually using `assert_raises` ... but both work ...

test/integration/authentication_test.rb
```diff
@@ -86,4 +86,123 @@
...
+          code = oauth_app.access_grants.first
+          code.redirect_uri.must_be :==, redirect_uri
+          code.application_id.must_be :==, oauth_app.id
+          code = code.token
+          response.location.must_match %r{#{code}}
+        end
+
+        describe 'getting the token' do
+          let(:new_params) do
+            { client_id: oauth_app.uid,
```
grosser: align looks strange to me ...

```
{
  foo: bar
  ...
}
```

test/integration/cleanliness_test.rb
```diff
@@ -43,9 +43,9 @@ def check_content(files)
     end
   end
 
-  it "does not have actions on base controller" do
+  it "does not have public actions on base controller" do
     found = ApplicationController.action_methods.to_a
-    found.reject { |a| a =~ /^(_conditional_callback_around_|_callback_before_)/ }
+    found.reject! { |a| a =~ /^(_conditional_callback_around_|_callback_before_|api_)/ }
```
grosser: why does api_ need to be public ? ... is it public by accident ?

lib/warden/action_dispatch_patch.rb
```diff
@@ -0,0 +1,24 @@
+# ActionDispatch has an authorization method but Rack::Request does not.
```
grosser: very generic name ... maybe `rack_request_authorization.rb`

test/test_helper.rb
```diff
@@ -242,10 +250,41 @@ def process_with_catch_warden(*args)
...
+      oauth_app.access_tokens.new do |token|
+        token.resource_owner_id = user.id
+        token.application_id = oauth_app.id
+        token.expires_in = 1000
+        token.scopes = :default
+      end
+    end
+
+    before do
+      token.save
```
grosser: `save!` ?

test/test_helper.rb
```diff
@@ -242,10 +250,41 @@ def process_with_catch_warden(*args)
...
+      json!
+      auth!("Bearer #{token.token}")
+    end
+  end
+
   def self.use_test_routes
     before do
       Rails.application.routes.draw do
         match "/test/:test_route/:controller/:action", via: [:get, :post, :put, :patch, :delete]
+        get "testing/foo" => "testing#foo"
```
grosser: do we need that ? add `test_route: true` to make your test routing work

lib/doorkeeper_auth.rb
apanzerj: hmm. fair. I can bread this into two methods: access_denied? and access_denied!

app/controllers/api/base_controller.rb
apanzerj: Nah, the array is for passing in ids.

app/controllers/api/deploys_controller.rb
apanzerj: will do.

app/controllers/api/stages_controller.rb
apanzerj: okay.

app/controllers/api/stages_controller.rb
apanzerj: I ended up doing:

```ruby
if new_stage.save
   do the right thing
else
   say this blew up
end
```

app/controllers/api/stages_controller.rb
apanzerj: why? that seems more expensive. I've always been told that rescuing costs.

app/controllers/api/stages_controller.rb
apanzerj: I'll make this: `stage_to_clone`

config/initializers/doorkeeper.rb
apanzerj: we can't "throw(:warden)" because we want to try session auth as well

I think we can get rid of the redirect actually since there will be an auth'd user when we get here.

config/routes.rb
apanzerj: 🎯

config/routes.rb
apanzerj: roger

lib/warden/strategies/doorkeeper_strategy.rb
apanzerj: It's should be tested in doorkeeper_auth_test actually. We don't test other methods.

lib/warden/strategies/session_strategy.rb
apanzerj: I don't. I'm wondering if I still need this, since we want you to be able to use session auth to access the api. 

Lemme think on this after I address the other comments.

test/controllers/api/deploys_controller_test.rb
apanzerj: :dart:

test/controllers/api/deploys_controller_test.rb
apanzerj: :dart:

test/controllers/api/deploys_controller_test.rb
apanzerj: found. fixed. :dart:

test/controllers/api/stages_controller_test.rb
apanzerj: hmm. looked for that, couldn't find it.

apanzerj: 🎯

test/controllers/api/stages_controller_test.rb
apanzerj: Okay to use assert_difference("Stage.count", 0)?

test/controllers/api/stages_controller_test.rb
apanzerj: :dart:

test/controllers/api/stages_controller_test.rb
apanzerj: 🎯

test/controllers/api/stages_controller_test.rb
apanzerj: 🎯

test/controllers/api/stages_controller_test.rb
apanzerj: :dart:

test/controllers/api/stages_controller_test.rb
apanzerj: :dart:

test/controllers/application_controller_test.rb
apanzerj: :dart:

test/integration/authentication_test.rb
apanzerj: blame rubocop....

test/integration/cleanliness_test.rb
apanzerj: It is public because api_accessible!(true)/(false) works.

lib/warden/action_dispatch_patch.rb
apanzerj: ehhhh. Very low value in fixing this, do you this change is imperative?

test/test_helper.rb
apanzerj: :dart:

test/test_helper.rb
apanzerj: :dart:

apanzerj:
@grosser re-review?

app/controllers/api/stages_controller.rb
grosser: yeah ... was thinking that we might need this general rescue workflow
anyway ... either way is fine

test/controllers/api/stages_controller_test.rb
grosser: yep

test/integration/cleanliness_test.rb
grosser: but that's a private method ... and static ?
maybe the class_attribute needs `instance_accessor: false`

lib/warden/action_dispatch_patch.rb
grosser: nah

test/integration/cleanliness_test.rb
grosser: fixed

test/lib/doorkeeper_auth_test.rb
```diff
@@ -0,0 +1,66 @@
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+class TestingController < ActionController::Base
```
grosser: this will be global, so needs to be named something more unique like DoorkeeperAuthTestingController

grosser:
:+1:

apanzerj:
🎉  thanks @grosser :-D

PR #1169, opened 2016-07-27
Revert  "update dependencies, refactor dockerfile" changes to Dockerfile

rata:
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

This reverts the part of that commit that changed the Dockerfile, except for the
change to ruby 2.3.1, that is of course needed. The bug hit it is this:
https://github.com/npm/npm/issues/9863 (although reported to an older version,
it still applies and bug reports to newer versions of npm where closed as
duplicates of this bug).

This fixes the build that is not working since that change and docker hub fails
to build images automatically since then.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 FROM ruby:2.3.1-slim
 
-RUN curl -sL https://deb.nodesource.com/setup_6.x | bash -
-RUN apt-get update && apt-get install -y nodejs build-essential libmysqlclient-dev libpq-dev libsqlite3-dev git
-RUN npm install -g npm
+RUN apt-get update && apt-get install -y wget apt-transport-https git
+RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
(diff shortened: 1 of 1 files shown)


grosser:
original PR https://github.com/zendesk/samson/pull/1146

can we revert only the first few lines + keep all the apt-get in the head ?

@sandlerr 

/cc @zendesk/samson

rata:
On Wed, Jul 27, 2016 at 08:17:49AM -0700, Michael Grosser wrote:
> original PR https://github.com/zendesk/samson/pull/1146
> 
> can we revert only the first few lines + keep all the apt-get in the head ?

Sure, no problem! Will do that

rata:
@grosser updated the commit msg and code to do that. This fixes the Dockerfile and tests are passing, although probably a test to build the docker image can be added in the future (I think travis needs something special, can look at it later)

grosser:
👍 much saner diff :D

sandlerr:
it's a shame, it builds fine for me locally because I'm not using `aufs`. I think only the `npm i -g npm` needed to be removed, will follow up.

sandlerr:
Thanks for the fix @rata !

PR #1170, opened 2016-07-27
cleanup deleted projects permission or creating new permissions fails

grosser:
@jonmoter @vanchi-zendesk 

alternative would be to make creation not fail ... but I'd rather get rid of the bad data ...


vanchi-zendesk:
lgtm 👍

PR #1171, opened 2016-07-27
PAAS-291 add filtering + totals to deploy group roles UI

grosser:
@jonmoter @irwaters 

screen shot 2016-07-27 at 11 23 59 am


jonmoter:
👍

PR #1173, opened 2016-07-27
do not blow up on stage or project creation when kubernetes config is…

grosser:
screen shot 2016-07-27 at 3 36 51 pm

@irwaters @jonmoter


irwaters:
👍

PR #1175, opened 2016-07-29
silence sprockets vs angular-rails-templates until we have a fix

grosser:
@apanzerj


apanzerj:
the sound of silence....

👍

PR #1176, opened 2016-07-29
eager load everything on travis for improved speed and production par…

grosser:
…ity ... also eager load more

also needed for rails 5 because they broke autoload_paths ...

@sandlerr @pschambacher


pschambacher:
👍  on green

grosser:
that was a fun bug :D 

https://github.com/attr-encrypted/attr_encrypted/issues/226

PR #1181, opened 2016-08-01
warn when trying to enter json, but the json is invalid

grosser:
@jonmoter @irwaters 

we will dump some json into secrets ... validation will be nice once they get complicated

### DIFF
--- a/app/views/admin/secrets/edit.html.erb
+++ b/app/views/admin/secrets/edit.html.erb
@@ -51,6 +51,7 @@
         <%= label_tag 'secret[value]', 'Value', class: "col-lg-2 control-label" %>
         <div class="col-lg-6">
           <%= text_area_tag 'secret[value]', '', class: "form-control", rows: 10 %>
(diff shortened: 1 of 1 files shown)


grosser:
screen shot 2016-08-01 at 1 54 45 pm

app/views/admin/secrets/edit.html.erb
```diff
@@ -107,4 +108,25 @@ $(function () {
...
+
+// when it looks like the user enters json, tell them when it is invalid
+// when they correct it, clear the warning
+$(function () {
+  var $value = $("#secret_value");
+  var $warn = $("#value_json_warning");
+
+  $value.keyup(function () {
+    var val = $value.val();
+    if(/^\s*\{[\s\S]*\}\s*$/.test(val)){
```
jonmoter: Catch arrays as well as hashes as the outermost characters.

grosser:
fixed

jonmoter:
👍 on 💚

PR #1182, opened 2016-08-01
Irwaters/paas 302 (deploy_group to vault relations)

irwaters:
Allows for the assoctionan of a given deploy group to a given vault instance.
vault.json now contains a name for the vault instance.  This is picked up by samson UI, and allows you to add that to a deploy group.  This is then used by the vualt client to store secrets for *that* pod.  global secrets are writen to all pods.  specific pod is checked by samson when doing a k8s deployment to make sure that it contains the correct secrets for that stage of the deployment.  UI changes will only  show up WHEN vault is being used for the secret storage backend. 

screen shot 2016-08-01 at 2 51 24 pm

* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:
https://zendesk.atlassian.net/browse/PAAS-302

### Risks
- Level: Low

### DIFF
--- a/app/models/secret_storage.rb
+++ b/app/models/secret_storage.rb
@@ -81,6 +81,7 @@ module SecretStorage
       def keys
         keys = vault_client.list(VAULT_SECRET_BACKEND + SAMSON_SECRET_NAMESPACE)
         keys = keys_recursive(keys)
+        keys.uniq! # we read from multiple backends that might have the same keys
         keys.map! do |secret_path|
           convert_path(secret_path, :decode) # FIXME: ideally only decode the key(#4) part
         end
--- a/app/views/admin/deploy_groups/_fields.html.erb
+++ b/app/views/admin/deploy_groups/_fields.html.erb
@@ -13,6 +13,15 @@
     </div>
   </div>
 
+  <% if SecretStorage.backend == SecretStorage::HashicorpVault %>
+    <div class="form-group">
+      <%= form.label :vault_instance, "Vault Instance", class: "col-lg-2 control-label" %>
+      <div class="col-lg-4">
+        <%= form.select(:vault_instance, options_for_select(VaultClient.available_instances, @deploy_group.vault_instance), {}, class: "form-control", required: true) %>
+      </div>
+    </div>
+  <% end %>
+
   <div class="form-group">
     <%= form.label :name, "Environment", class: "col-lg-2 control-label" %>
     <div class="col-lg-4">
--- a/config/initializers/vault.rb
+++ b/config/initializers/vault.rb
@@ -39,80 +39,97 @@ if ENV["SECRET_STORAGE_BACKEND"] == "SecretStorage::HashicorpVault"
 
     def initialize
       return if Rails.env.test?
-      ensure_config_exists
+      VaultClient.ensure_config_exists
       # since we have a bunch of servers, don't use the client singlton to
       # talk to them
       # as we configure each client, check to see if the config has a token in it,
       # if it does, then we don't need to worry about going and getting one
-      @writers = vault_hosts.map do |vault_server|
+      @vaults = {}
+      vault_hosts.each do |vault_server|
+        instance = vault_server.fetch("vault_instance")
         if vault_server["vault_token"].present?
-          writer = Vault::Client.new(DEFAULT_CLIENT_OPTIONS.merge(ssl_verify: vault_server["tls_verify"]))
-          writer.token = File.read(vault_server["vault_token"]).strip
+          tls_verification = vault_server.fetch("tls_verify")
+          @vaults[instance] = Vault::Client.new(DEFAULT_CLIENT_OPTIONS.merge(ssl_verify: tls_verification))
+          @vaults[instance].token = File.read(vault_server.fetch("vault_token")).strip
         else
-          pemfile = File.read(vault_server["vault_auth_pem"])
+          pemfile = File.read(vault_server.fetch("vault_auth_pem"))
           client_cert_options = {
             cert: OpenSSL::X509::Certificate.new(pemfile),
             key: OpenSSL::PKey::RSA.new(pemfile),
-            ssl_pem_file: vault_server["vault_auth_pem"],
-            verify_mode: vault_server["tls_verify"]
+            ssl_pem_file: vault_server.fetch("vault_auth_pem"),
+            verify_mode: vault_server.fetch("tls_verify")
           }
           writer = Vault::Client.new(DEFAULT_CLIENT_OPTIONS.merge(client_cert_options))
           writer.token = VaultClient.auth_token(vault_server["vault_address"])
         end
-        writer.address = vault_server["vault_address"]
-        writer.ssl_ca_cert = vault_server["ca_cert"] if vault_server["ca_cert"]
-        writer
+        @vaults[instance].address = vault_server.fetch("vault_address")
+        @vaults[instance].ssl_ca_cert = vault_server.fetch("ca_cert") if vault_server.fetch("ca_cert", false)
       end
-      @reader = @writers.first
     end
 
+    # normally we'll only get a single instance back.  if it's global, just read it once
     def read(key)
-      @reader.logical.read(key)
+      vault_instances(key).first.logical.read(key)
     end
 
     def list(path)
-      @reader.logical.list(path)
+      @vaults.each_value.flat_map { |vault| vault.logical.list(path) }
     end
 
     # make darn sure on deletes and writes that we try a couple of times.
     # not going to catch any other excpeptions as we want this to blow up
     # if anything fails
     def write(key, data)
-      @writers.each do |vault_server|
-        Vault.with_retries(Vault::HTTPConnectionError, attempts: 5) do
-          vault_server.logical.write(key, data)
-        end
+      Vault.with_retries(Vault::HTTPConnectionError, attempts: 5) do
+        vault_instances(key).each { |v| v.logical.write(key, data) }
       end
     end
 
     def delete(key)
-      @writers.each do |vault_server|
-        Vault.with_retries(Vault::HTTPConnectionError, attempts: 5) do
-          vault_server.logical.delete(key)
-        end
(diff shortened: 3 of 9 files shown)


app/models/secret_storage.rb
```diff
@@ -80,7 +80,7 @@ def delete(key)
 
       def keys
         keys = vault_client.list(VAULT_SECRET_BACKEND + SAMSON_SECRET_NAMESPACE)
-        keys = keys_recursive(keys)
+        keys = keys_recursive(keys).uniq # flatten all the keys since the pod/vault association is included now
```
grosser: `# we read from multiple backends that might have the same keys` ... can also call `keys.uniq!` in a separate line to avoid a bit of overhead ...

app/views/admin/deploy_groups/_fields.html.erb
```diff
@@ -13,6 +13,15 @@
     </div>
   </div>
 
+  <% if ENV["SECRET_STORAGE_BACKEND"] == "SecretStorage::HashicorpVault" %>
```
grosser: I think we do a similar check somewhere else ...
something like `SecretStorage.backend == SecretStorage::HashicorpVault` ... or checking against a constant ...

config/initializers/vault.rb
```diff
@@ -1,3 +1,4 @@
+require 'byebug'
```
grosser: remove ?

config/initializers/vault.rb
```diff
@@ -44,10 +45,12 @@ def initialize
       # talk to them
       # as we configure each client, check to see if the config has a token in it,
       # if it does, then we don't need to worry about going and getting one
-      @writers = vault_hosts.map do |vault_server|
+      @vaults = {}
+      vault_hosts.each do |vault_server|
+        instance = vault_server["vault_instance"]
```
grosser: `.fetch` to blow up when it does not exist ?

config/initializers/vault.rb
```diff
@@ -44,10 +45,12 @@ def initialize
...
         if vault_server["vault_token"].present?
-          writer = Vault::Client.new(DEFAULT_CLIENT_OPTIONS.merge(ssl_verify: vault_server["tls_verify"]))
-          writer.token = File.read(vault_server["vault_token"]).strip
+          @vaults[instance] = Vault::Client.new(DEFAULT_CLIENT_OPTIONS.merge(ssl_verify: vault_server["tls_verify"]))
+          @vaults[instance].token = File.read(vault_server["vault_token"]).strip
```
grosser: also fetch

config/initializers/vault.rb
```diff
@@ -59,44 +62,43 @@ def initialize
...
 
+    # normally we'll only get a single instance back.  if it's global, just read it once
     def read(key)
-      @reader.logical.read(key)
+      vault_instances(key).first.logical.read(key)
     end
 
     def list(path)
-      @reader.logical.list(path)
+      @vaults.flat_map { |vault| vault.pop.logical.list(path) }
```
grosser: @vaults is a hash, so how does this work ? ... missing tests ?

grosser: aaa ... the .pop ... `@vaults.each_value` is what you want :)

config/initializers/vault.rb
```diff
@@ -114,5 +116,20 @@ def vault_hosts
     def vault_config_file
       ENV['VAULT_CONFIG_FILE'] || Rails.root.join("config/vault.json")
     end
+
+    # get back the vault instance that's required for this pod.  we'll take
+    # the key, and look up the instance or return all of 'em if it's a global key
+    def vault_instances(key = nil)
```
grosser: `=nil` would mean the split blows up ... so remove the default value ?

config/initializers/vault.rb
```diff
@@ -114,5 +116,20 @@ def vault_hosts
...
+      deploy_group = key.split('/')[4]
```
grosser: reuse the parsing logic from SecretBackend

config/initializers/vault.rb
```diff
@@ -114,5 +116,20 @@ def vault_hosts
...
+      if deploy_group == 'global'
+        @vaults.map { |vault| vault.pop }
```
grosser: `@vaults.values`

config/initializers/vault.rb
```diff
@@ -114,5 +116,20 @@ def vault_hosts
...
+      else
+        deploy_group = DeployGroup.find_by_name(deploy_group)
+        if deploy_group.nil?
+          raise "no vault_instance configured for deploy group #{deploy_group}"
+        end
+        [@vaults[deploy_group.vault_instance]]
```
grosser: .fetch

plugins/kubernetes/test/test_helper.rb
```diff
@@ -41,11 +44,27 @@ def write(key, body)
...
+      'tls_verify' => false, "vault_instance": deploy_group_name
     }
   end
 
+  def ensure_config_exists
+    raise "config file missing" unless File.exist?(CONFIG_FILE)
+  end
+
+  def create_config
+    vault_json = <<END_JSON
```
grosser: for better readability indent normally and use .strip_heredoc

grosser: also using `JSON` means the inner parts get nice highlighting in some editors
```Ruby
  vault_json = <<-JSON.strip_heredoc
    {
      xxx
    }
  JSON
```

grosser: might be simplest to use a ruby hash and do .to_json

plugins/kubernetes/test/test_helper.rb
```diff
@@ -41,11 +44,27 @@ def write(key, body)
...
+{
+  "vault_token": "/foo/bar.token", "vault_address": "https://10.10.10.10:8200", "tls_verify": false, "vault_instance": "superman"
+}
+END_JSON
+    File.write(CONFIG_FILE, vault_json)
+  end
+
+  def remove_config
+#    File.delete(CONFIG_FILE)
```
grosser: remove / uncomment ?

test/models/secret_storage_test.rb
```diff
@@ -170,9 +170,24 @@
     let(:client) { SecretStorage::HashicorpVault.send(:vault_client) }
     let(:secret_namespace) { "secret/apps/" }
 
+    around do |t|
+      @dir = Dir.mktmpdir
```
grosser: this will leave the dir after the test is done

```Ruby
 around { |test| Dir.mktmpdir { |dir| Dir.chdir(dir) { test.call } } }
```

test/models/secret_storage_test.rb
```diff
@@ -170,9 +170,24 @@
...
+    end
+
     before { client.clear }
     after { client.verify! }
 
+    describe "missing config file" do
+      before { client.remove_config }
+      after { client.create_config }
+      it "fails without a config file" do
+        assert_raises RuntimeError do
```
grosser: lots of errors could happen, ideally assert what kind of error you expect:
```
e = assert_raises ...
e.message.must_include "File xyz not found"
```

config/initializers/vault.rb
```diff
@@ -59,44 +62,43 @@ def initialize
...
       end
     end
 
-    def config_for(deploy_group_name)
-      vault_hosts.detect do |hash|
-        hash.fetch('deploy_groups', []).include? deploy_group_name
-      end
+    def self.available_instances
+      client = self.new
+      client.send(:ensure_config_exists)
```
grosser: make the method static so we don't have to create a instance and throw it away ?

config/initializers/vault.rb
grosser: I'd love to see some test that says a write goes to all vault instances when the deploy group is global / goes to 1 when it's local ...

irwaters: will add

app/models/secret_storage.rb
```diff
@@ -80,7 +80,8 @@ def delete(key)
 
       def keys
         keys = vault_client.list(VAULT_SECRET_BACKEND + SAMSON_SECRET_NAMESPACE)
-        keys = keys_recursive(keys)
+        keys = keys_recursive(keys).uniq # we read from multiple backends that might have the same keys
```
grosser: the `.uniq` here can go away

grosser: and comment moved down 1 line

config/initializers/vault.rb
```diff
@@ -39,80 +39,97 @@ def self.client
...
+      @vault_hosts ||= JSON.parse(File.read(VaultClient.vault_config_file))
     end
 
-    def vault_config_file
-      ENV['VAULT_CONFIG_FILE'] || Rails.root.join("config/vault.json")
+    # get back the vault instance that's required for this pod.  we'll take
+    # the key, and look up the instance or return all of 'em if it's a global key
+    def vault_instances(key)
+      # parse_secret_key doesn't know about the namespace etc, so strip it off
+      deploy_group = SecretStorage.parse_secret_key(key.split('/', 3).pop).fetch(:deploy_group_permalink)
```
grosser: prefer .last to .pop since it does not have a side-effect

config/initializers/vault.rb
```diff
@@ -39,80 +39,97 @@ def self.client
...
     def vault_hosts
-      @vault_hosts ||= JSON.parse(File.read(vault_config_file))
+      @vault_hosts ||= JSON.parse(File.read(VaultClient.vault_config_file))
     end
 
-    def vault_config_file
-      ENV['VAULT_CONFIG_FILE'] || Rails.root.join("config/vault.json")
+    # get back the vault instance that's required for this pod.  we'll take
+    # the key, and look up the instance or return all of 'em if it's a global key
+    def vault_instances(key)
```
grosser: any way to test that logic ?

irwaters: added.

grosser:
👍  ... please squasch or merge with squasch ...

grosser:
build is red now ... https://travis-ci.org/zendesk/samson

PR #1187, opened 2016-08-02
make tabs navigation with hashes update the url and show correct tabs…

grosser:
screen shot 2016-08-02 at 1 30 18 pm

@jonmoter @irwaters


jonmoter:
👍

PR #1191, opened 2016-08-03
fixes stdout logs for production env

vitalis:
* fixes stdout logs for production env

Right now the STDOUT logs not working in production, the issue coming form setting the `Rails.application.config` in initializer and using `lograge` gem.
In order to set correct config for `lograge`, `Rails.application.config` should be updated in `application.rb` or `environments/*.rb`.
I'm also not sure that `Syslog::Logger` is working as expected in production. (maybe it is)

Attached screenshots of the fix:
Before: (no logs)
before

After: (with logs)
after

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High


config/application.rb
```diff
@@ -175,3 +175,4 @@ class Application < Rails::Application
 end
 
 require 'samson/hooks'
+require "#{Rails.root}/lib/logging"
```
grosser: can this be `require 'logging'` ?
maybe make it samson_logging so it does not override others by accident ...

grosser:
👍 looks good ... I might have broken this when I refactored logging :D

config/application.rb
vitalis: I can put it under lib/samson/logging

vitalis:
@grosser updated

grosser:
can you add lib/samson/logging.rb to the cleanliness test exceptions so the build is green ?
getting this atm `RuntimeError: missing test for lib/samson/logging.rb`

vitalis:
Done, sorry for the failing test

PR #1195, opened 2016-08-09
Use 1.1.4

apanzerj:
* @grosser @jonmoter Looks like they released the gem you need https://github.com/abonas/kubeclient/pull/127.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Med
No idea what this is, just bumping according to notes in gemfile.


jonmoter:
👍

PR #1196, opened 2016-08-10
bump mini-profiler to avoid security vulnerability spam

grosser:
@apanzerj 

/cc @cericksen caught by that harakiri thing ... not a production issue ... just silencing it


apanzerj:
👍

PR #1198, opened 2016-08-11
Adding More Changes to API

apanzerj:
Description:
- Added the attribute `template` to the Stage model.
- Modified StagesController (both) to allow setting the template stage.
- Nix the permalinks in the API

/cc @zendesk/samson
### Tasks
- [ ] :+1: from team
### References
- Jira link:
### Risks
- Level: Low/Med/High


app/views/shared/_header.html.erb
```diff
@@ -74,8 +74,10 @@
             <li><%= link_to "Secrets", admin_secrets_path %></li>
           <% end %>
           <li><%= link_to "Reports", csv_exports_path %></li>
-          <li><%= link_to "OAuth Applications", oauth_applications_path %></li>
-          <li><%= link_to "OAuth Authorized Applications", oauth_authorized_applications_path %></li>
+          <% if current_user.try(:super_admin?) %>
```
sandlerr: ruby 2.3 means you can use the lonely operator now :D `current_user&.super_admin?`

apanzerj: wuuuuuuut.

crazy, but I like it.

sandlerr:
nitpick: would you mind pulling all the routing tests into a single block at the top of the file since that will map better to the code they're actually testing (i.e. the stuff in `routes.rb` rather than the actual actions)?

test/controllers/api/deploys_controller_test.rb
```diff
@@ -3,17 +3,17 @@
 SingleCov.covered!
 
 describe Api::DeploysController do
+  assert_route "/api/deploys/active_count", 'api/deploys', 'active_count'
```
apanzerj: @sandlerr like this?

apanzerj:
To Do:

- [ ] Tests
- [ ] Enforce one duplicable stage per project
- [ ] Allow duplicable setting to be set via API
- [ ] UI Pictures

sandlerr:
should `duplicable` have a per-project uniqueness validation?

test/controllers/api/deploys_controller_test.rb
sandlerr: yes! one more nitpick though, I would nest it inside 

```ruby
describe 'routing' do
  it 'dispatches correctly' do
    ...
  end
end
```

not sure what minitest's behaviour is if a spec fails outside of an `it` block, maybe it's fine.

sandlerr: never mind, assert_route is an `it` block

apanzerj:
It will

test/controllers/api/deploys_controller_test.rb
apanzerj: actually got it done anyways. sorta...
```ruby
def self.assert_route(verb:, path:, controller:, action:, params: {})
    describe "a #{verb} to #{path}" do
      before do
        @it_name = params.keys.empty? ? "routes to #{controller} #{action}" : "routes to #{controller} #{action} with parameters #{params}"
      end

      it @it_name do
        assert_routing({method: verb, path: path}, {controller: controller, action: action}.merge(params))
      end
    end
  end
```

.rubocop.yml
```diff
@@ -1,3 +1,6 @@
+FrozenStringLiteralComment:
```
apanzerj: cc @grosser is this okay? Without it every file fails.

apanzerj: It would also be a massive pr

apanzerj: Made the massive PR here: https://github.com/zendesk/samson/pull/1200

@zendesk/samson lemme know what you want to do...

apanzerj:
@zendesk/samson Can somebody give this a review?

cc @pswadi-zendesk

.rubocop.yml
grosser: push that big PR through unless there are any big blockers ...

app/controllers/api/base_controller.rb
```diff
@@ -3,6 +3,8 @@
 class Api::BaseController < ApplicationController
   skip_before_action :verify_authenticity_token
 
+  before_action :require_project
```
grosser: not really required ... `find_project` ?

app/controllers/api/stages_controller.rb
```diff
@@ -18,10 +16,19 @@ def clone
     end
   end
 
+  def duplicable
```
grosser: this seems weird ...

grosser: some docs on what it does / why we need that ?

app/controllers/api/stages_controller.rb
```diff
@@ -18,10 +16,19 @@ def clone
...
+    render json: @project.duplicable_stage
+  end
+
+  def put_duplicable
```
grosser: `create_duplicate` ... how is this different from `clone` ... ? ... more docs :D

app/controllers/api/stages_controller.rb
```diff
@@ -18,10 +16,19 @@ def clone
...
+    duplicable_stage.duplicable!
+    head :no_content
+  end
+
   private
 
   def stage_to_clone
-    Stage.find_by_param!(params.require(:stage_id))
+    Stage.find_by_id(params.require(:stage_id))
```
grosser: should we stay with permalinks for consistency ?

app/controllers/api/stages_controller.rb
```diff
@@ -35,4 +42,8 @@ def deploy_group_ids
   def deploy_groups
     DeployGroup.where(id: deploy_group_ids)
   end
+
+  def duplicable_stage
+    @project.stages.where(id: params[:id]).first
```
grosser: `first!` so we get 404s ?

app/models/stage.rb
```diff
@@ -180,6 +180,11 @@ def deploy_group_names
     DeployGroup.enabled? ? deploy_groups.select(:name).sort_by(&:natural_order).map(&:name) : []
   end
 
+  def duplicable!
+    project.duplicable_stage && project.duplicable_stage.update!(duplicable: false)
```
grosser: triggers 2 finds ?

app/views/stages/_fields.html.erb
```diff
@@ -61,6 +61,17 @@
     </div>
   </div>
 
+  <div class="form-group">
+    <div class="col-sm-offset-2 col-sm-10">
+      <div class="checkbox">
+        <%= form.label :duplicable do %>
+          <%= form.check_box :duplicable %>
+          This stage can serve as a production template.
```
grosser: not sure why we need this ... if a stage goes to production it can serve as a production template ?
also I expect people will clone a stage and just leave this on, breaking it ...
... if we still want it, then pick a better column name or description ? `Can be used as template for copying / has not unique customizations`

config/routes.rb
```diff
@@ -3,13 +3,25 @@
...
-    resources :deploys, only: [:index, :active_count] do
+    resources :deploys, only: [:active_count] do
       collection do
         get :active_count
       end
     end
 
-    resources :stages, only: [:index] do
+    resources :projects, param: :project_id, only: [] do
+      get :duplicable, on: :member, to: 'stages#duplicable'
```
grosser: use `member do` .. no need for `to` and `on`

test/controllers/api/deploys_controller_test.rb
```diff
@@ -3,7 +3,11 @@
 SingleCov.covered!
 
 describe Api::DeploysController do
+  assert_route verb: "GET", path: "/api/deploys/active_count", to: 'api/deploys#active_count'
```
grosser: not sure if these are worthwhile, the test that executes them makes sure they exist ...

app/controllers/api/base_controller.rb
apanzerj: I'm just mimic-ing the concern we have. Require everywhere it's included and skip when needed.

app/controllers/api/stages_controller.rb
apanzerj: Situation:

You add a new deploy group(s) for testing. You need to know which stage is a complete stage (ie: all the commands make sense and it deploys the project successfully). Then, when you add your new stage, you simply clone the stage marked duplicable and change the deploy groups.

I wanted to call it a "Gold/Master Stage" but gold stage is meaningless and master stage was confusing.

app/controllers/api/stages_controller.rb
apanzerj: Sorry. `#put_duplicable` is to change which stage is marked duplicable. Only one stage per project is "duplicable". The `#duplicable` method is simply the currently marked stage on the project that is duplicable. Clone is just clone.

app/controllers/api/stages_controller.rb
apanzerj: I thought about that, but permalinks are not RESTful in my opinion. 

The permalink is a nicety, like porcelain, it is essentially an extension of the UI.

app/controllers/api/stages_controller.rb
apanzerj: yessir.

app/models/stage.rb
apanzerj:
```

  def duplicable!
    old_duplicable_stage = project.duplicable_stage
    old_duplicable_stage.update!(duplicable: false) if old_duplicable_stage
    update!(duplicable: true)
  end
```

config/routes.rb
apanzerj: What's wrong with `on: :member` ?

test/controllers/api/deploys_controller_test.rb
apanzerj: @sandlerr thought it was useful. I agree too, it helps make sure you end up with the routes you intend.

app/views/stages/_fields.html.erb
apanzerj: So here's the deal:

1. At one point in time, I cloned a stage and ran with it, not realizing it was somebody just messing around.

2. That stage had `bundle -v foo bar blah` and so it just spit out the current version of bundler. Having a "gold master" stage would fix this problem. If you can come up with a better name, I'm all ears, this was the best thing @sandlerr and I came up with.

app/controllers/api/stages_controller.rb
apanzerj: This changes which stage is duplicable.

config/routes.rb
apanzerj: This needs the to: `stages#duplicable` because without it, the request routes to the projects controller.

See failure: https://travis-ci.org/zendesk/samson/jobs/151951746

app/controllers/api/stages_controller.rb
grosser: chose_duplicable / mark_duplicable ?

config/routes.rb
grosser: hmm yeah works too ... saw `member do` more often ...

app/models/stage.rb
grosser: Label should then also say something like "Maximum of 1 per project"

app/views/stages/_fields.html.erb
grosser: then say it "Duplication template" ... or "Clone template" ... or name the
column clone_template ... just somehow have the description and column in
sync ...

app/controllers/api/stages_controller.rb
apanzerj: mark is better.

app/models/stage.rb
apanzerj: :zap: got it

app/views/stages/_fields.html.erb
apanzerj: I talked to @dragonfax and he had some good ideas, do you like any of the following:

1. Alpha stage (the progenitor stage)
2. Baseline stage
3. Prime stage

apanzerj:
Talked to @grosser and he thinks:

#usable_as_template?
#usable_as_template!

The route name might be:

put api/projects/1/stages/2/template
get api/projects/1/stages/2/template 

Thoughts? cc @grosser

grosser:
+1

apanzerj:
@grosser can you review?

Also: 

So the segfault happened in the build for #0e134c5. I cleared the cache and updated webmock (where the segfault traced to) and it still happened, this time in actioncontroller journey foo.

Then I updated rails and poof, no more segfault. Could be corrupted gem cache or could be unrelated.

apanzerj:
cc @zendesk/samson @dragonfax and @irwaters

app/controllers/api/deploy_groups_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+class Api::DeployGroupsController < Api::BaseController
+  include DeployGroupsCommon
+  skip_before_action :require_project, only: :index
+  before_action :enabled?
+
+  def index
+    render json: paginate(deploy_groups)
+  end
+
+  def project_deploy_groups
```
grosser: that should be index just with a different url ?

app/controllers/api/deploy_groups_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    render json: paginate(stage.deploy_groups)
+  end
+
+  def update_deploy_groups
+    if !production_change?
+      stage.deploy_groups = deploy_groups_to_add
+      stage.save!
+      head :no_content
+    else
+      render json: {message: "Do not modify production via the API"}, status: 403
```
grosser: wut ?

app/controllers/api/deploy_groups_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    render json: paginate(deploy_groups)
+  end
+
+  def project_deploy_groups
+    render json: paginate(stage.deploy_groups)
+  end
+
+  def update_deploy_groups
+    if !production_change?
+      stage.deploy_groups = deploy_groups_to_add
```
grosser: how about stage_controller#update ?

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,43 @@
...
+  def filter
+    if filter_param = params.fetch(:filter, false)
+      filter_param if ["pending", "running", "canceled", "failed", "errored", "succeeded"].include?(filter_param)
+    else
+      false
+    end
+  end
+
+  def search_params
+    if project_id = params.fetch(:project_id, false)
```
grosser: no need to fetch ... `if project_id = params[:project_id]`

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,43 @@
...
+      stage_ids = Project.find(project_id).stages.pluck(:id)
+      search_params = {jobs: {}, deploys: {stage_id: stage_ids}}
+    elsif stage_id = params.fetch(:stage_id, false)
```
grosser: check stage_id first, since if I give project and stage then stage should win ?

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,43 @@
...
   end
 
   def active_count
     render json: Deploy.active.count
   end
+
+  protected
+
+  def filter
+    if filter_param = params.fetch(:filter, false)
```
grosser: no fetch needed

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,43 @@
...
+      filter_param if ["pending", "running", "canceled", "failed", "errored", "succeeded"].include?(filter_param)
```
grosser: reuse some constant that has the list of all value states ?

grosser: also should be bad_request if it is not included instead of returning all silently

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,43 @@
...
+      stage_ids = Project.find(project_id).stages.pluck(:id)
+      search_params = {jobs: {}, deploys: {stage_id: stage_ids}}
+    elsif stage_id = params.fetch(:stage_id, false)
+      search_params = {jobs: {}, deploys: {stage_id: stage_id}}
+    end
+
+    if filter
+      search_params[:jobs] = {status: filter}
+    else
+      search_params.delete(:jobs)
```
grosser: don't set jobs to empty above instead of deleting here ?

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,43 @@
...
+    end
+
+    search_params
+  end
+
+  def deploy_scope
```
grosser: does very little ... maybe inline

app/controllers/api/stages_controller.rb
```diff
@@ -19,10 +17,19 @@ def clone
     end
   end
 
+  def template
+    render json: @project.template
```
grosser: maybe call it template_stage so it's obvious what it is

app/controllers/api/stages_controller.rb
```diff
@@ -36,4 +43,8 @@ def deploy_group_ids
   def deploy_groups
     DeployGroup.where(id: deploy_group_ids)
   end
+
+  def new_template_stage
+    @project.stages.where(id: params[:id]).first!
```
grosser: how about `@project.stages.find(params[:id])`

app/controllers/concerns/deploy_groups_common.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+module DeployGroupsCommon
+  extend ActiveSupport::Concern
```
grosser: don't need concern ?

app/controllers/stages_controller.rb
```diff
@@ -102,4 +99,12 @@ def stage_params
   def find_stage
     @stage = current_project.stages.find_by_param!(params[:id])
   end
+
+  def template_param
+    @template_param ||= ActiveRecord::Type::Boolean.new.type_cast_from_database(stage_params.delete("template"))
+  end
+
+  def update_template
+    template_param ? @stage.project.template!(@stage) : @stage.project.template!
```
grosser: make this a after_save hook ?

not sure what template! does ... maybe pick a more telling name ?

app/models/project.rb
```diff
@@ -43,6 +43,18 @@ class Project < ActiveRecord::Base
 
   scope :search, ->(name) { where("name like ?", "%#{name}%") }
 
+  def template
+    stages.where(template: true)&.first
```
grosser: don't need the `&`

app/controllers/stages_controller.rb
grosser: ... remove default and call with `template!(nil)` ... much more obvious

grosser: this means not selection template will unset the projects template ... seems broken

app/controllers/api/deploy_groups_controller.rb
apanzerj: hmm. I guess we could get rid of it. It was an attempt to protect stages from modification via the API if they are production.

app/views/stages/_fields.html.erb
```diff
@@ -61,6 +61,17 @@
     </div>
   </div>
 
+  <div class="form-group">
+    <div class="col-sm-offset-2 col-sm-10">
+      <div class="checkbox">
+        <%= form.label :template do %>
+          <%= form.check_box :template %>
+          This stage can serve as a production template (maximum of 1 per project).
```
grosser: how about: `Can be cloned and modified to create a new stage (1 per project)`

config/routes.rb
```diff
@@ -4,13 +4,34 @@
...
         get :active_count
       end
     end
 
-    resources :stages, only: [:index] do
+    resources :deploy_groups, only: [:index]
+
+    resources :projects, param: :project_id, only: [] do
+      member do
+        get :template, to: 'stages#template'
```
grosser: make it a collection on stages and do `project/foo/stages/template` ?

grosser: hmmm I guess that could break permalinks ...

config/routes.rb
```diff
@@ -4,13 +4,34 @@
...
+      end
+    end
+
+    resources :projects, only: [:index] do
+      resources :stages, only: [:index] do
+        member do
+          put :template, to: 'stages#mark_template'
+          get :deploy_groups, to: 'deploy_groups#project_deploy_groups'
```
grosser: include deploy groups in normal stage#get ?

config/routes.rb
```diff
@@ -4,13 +4,34 @@
...
+          put :deploy_groups, to: 'deploy_groups#update_deploy_groups'
+        end
+      end
+
+      resources :deploys, only: [:index]
+    end
+
+    resources :stages, only: [] do
+      get :deploys, to: 'deploys#index'
```
grosser: what url is this ? stages/deploys ?

grosser: or same as `resources :deploys, only :index` ?

test/controllers/api/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,155 @@
...
+
+  assert_route verb: "PUT", path: "/api/projects/1/stages/2/deploy_groups", \
+               to: "api/deploy_groups#update_deploy_groups", \
+               params: {project_id: "1", id: "2"}
+
+  oauth_setup!
+
+  describe 'when deploy_groups is disabled' do
+    before do
+      DeployGroup.stubs(:enabled?).returns(false)
```
grosser: one of these stubs is the default and is not needed

test/controllers/api/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,155 @@
...
+      get :index
+    end
+
+    it 'returns precondition_failed' do
+      assert_response :precondition_failed
+    end
+  end
+
```
grosser: I'd prefer having 1 `it 'returns precondition failed when deploy groups are disabled'` instead of wrapping everything in another layer of describes

test/controllers/api/deploys_controller_test.rb
```diff
@@ -4,7 +4,13 @@
 SingleCov.covered!
 
 describe Api::DeploysController do
+  Maxitest.timeout = false
```
grosser: this is a global change ... revert

test/controllers/api/deploys_controller_test.rb
```diff
@@ -25,52 +31,117 @@
...
+        let(:filter) { "green" }
+
+        it 'does not include a filter' do
+          expected = {deploys: {stage_id: stage.id}}
+          subject.send(:search_params).must_equal expected
+        end
+      end
+
+      describe 'if the route is a project' do
+        subject do
```
grosser: modify the params of the controller instead of overloading subject ?

app/controllers/api/deploy_groups_controller.rb
grosser: seems kind of random / makes the api pretty useless ...

app/controllers/api/deploy_groups_controller.rb
apanzerj: So now they both go to index and look like this: 

```ruby
  def index
    scope = current_project.present? ? stage.deploy_groups : deploy_groups
    render json: paginate(scope)
  end
```

app/controllers/api/deploy_groups_controller.rb
apanzerj: This was because we don't want non-humans making production changes.

grosser: but it's an authorized request ... not sure why it matters how it came to
the app ... also some very arbitrary rule that makes the api useless ...

config/routes.rb
apanzerj: eh? there are no permalinks in the api...

apanzerj: but I like the idea of the route being ```project/:id/stages/template```

config/routes.rb
apanzerj: ```api_stage_deploys GET    /api/stages/:stage_id/deploys(.:format)                          api/deploys#index```

config/routes.rb
grosser: should be fine too ... chances are nobody named their stage template :D

config/routes.rb
apanzerj: In order to get the last successful deploy for a given stage.

test/controllers/api/deploy_groups_controller_test.rb
apanzerj: ⚡

test/controllers/api/deploy_groups_controller_test.rb
apanzerj: got it.

test/controllers/api/deploys_controller_test.rb
apanzerj: yup, doing so now.

config/routes.rb
apanzerj: This is now moved per previous comment. Still relevant?

app/views/stages/_fields.html.erb
apanzerj: ehhh. All the stages can be cloned. The point about this is that this particular stage *should* be cloned.

app/controllers/stages_controller.rb
apanzerj: eh?

template!(nil) => Just unset the template stage for the project.
template!(<some stage>) => Unset the existing template for the project (if there is one) and set the one given.

apanzerj: Also: this can't be an after_save hook because the default behavior will still set the passed in stage as template w/o unsetting the previous stage.

app/controllers/concerns/deploy_groups_common.rb
apanzerj: in this case, no, but I might need it later?

Take it out?

app/controllers/api/stages_controller.rb
apanzerj: done

app/controllers/stages_controller.rb
grosser: can you make that more intuitive ? maybe keyword args ?

app/controllers/concerns/deploy_groups_common.rb
grosser: take out anything we don't need :)

app/controllers/stages_controller.rb
grosser: why can't it unset the previous stage when `template` is enabled ?

app/controllers/api/stages_controller.rb
apanzerj: done

app/controllers/api/deploys_controller.rb
apanzerj: roger

app/controllers/api/deploys_controller.rb
apanzerj: done

app/controllers/api/deploys_controller.rb
apanzerj: yup

app/controllers/api/deploys_controller.rb
apanzerj: done.

app/controllers/api/deploys_controller.rb
apanzerj: sure. In the end it's moot, there are three routes that hit this method:

api/stages/:stage_id/deploys
api/projects/:project_id/deploys
api/deploys

app/controllers/api/deploy_groups_controller.rb
apanzerj: done

app/controllers/api/deploy_groups_controller.rb
apanzerj: I got the 'just do it' so I'll back this out.

apanzerj:
@grosser ready when you are

app/controllers/api/deploy_groups_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+class Api::DeployGroupsController < Api::BaseController
+  include DeployGroupsCommon
+  before_action :enabled?
```
grosser: not a boolean method -> `ensure_enabled` or something like that

app/controllers/api/deploy_groups_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def index
+    scope = current_project.present? ? stage.deploy_groups : deploy_groups
```
grosser: no need for `.present?`

grosser: missing `.sort_by(&:natural_order)` for the stage side ... maybe best to copy that bit and not have a concern ?

app/controllers/api/deploy_groups_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    render json: paginate(scope)
+  end
+
+  protected
+
+  def stage
+    current_project.stages.where(id: params[:id]).first!
```
grosser: `current_project.stages.find(params[:id])`

app/controllers/api/deploy_groups_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def enabled?
+    return if DeployGroup.enabled?
+    render json: {message: "DeployGroups are not enabled."}, status: :precondition_failed
+    false
```
grosser: no need for `false`

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,40 @@
...
+    render json: paginate(Deploy.joins(:job).where(search_params))
   end
 
   def active_count
     render json: Deploy.active.count
   end
+
+  protected
+
+  def filter
```
grosser: rename to job_filter ?

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,40 @@
...
+      stage_ids = Project.find(project_id).stages.pluck(:id)
+      search_params = {deploys: {stage_id: stage_ids}}
+    end
+
+    search_params[:jobs] = {status: filter} if filter
+
+    search_params
+  end
+
+  def ensure_filter
```
grosser: no need for true or false

app/controllers/api/deploy_groups_controller.rb
apanzerj: right

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,40 @@
...
+    return true if params[:filter].nil?
```
grosser: `unless params[:filter]`

app/controllers/api/deploys_controller.rb
```diff
@@ -1,16 +1,40 @@
...
+    unless Job.valid_status?(filter)
+      render json: { error: "Filter is not valid. Please use " + Job::VALID_STATUSES.join(", ") }, status: :bad_request
```
grosser: nice!

app/controllers/api/deploy_groups_controller.rb
apanzerj: so just nix the concern?

app/controllers/api/projects_controller.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class Api::ProjectsController < Api::BaseController
+  def index
+    render json: Project.ordered_for_user(current_user).all
```
grosser: does this use the project_serializer or does that render out all attributes ?

app/controllers/api/stages_controller.rb
```diff
@@ -19,10 +18,25 @@ def clone
...
   end
 
+  def update
+    stage.deploy_groups = deploy_groups
+    stage.save!
+    head :no_content
+  end
+
+  def template_stage
+    render json: current_project.template
```
grosser: 404 when there is no template ?

app/controllers/api/stages_controller.rb
```diff
@@ -1,11 +1,10 @@
 # frozen_string_literal: true
 class Api::StagesController < Api::BaseController
-  include CurrentProject
-
-  before_action :require_project
+  skip_before_action :require_project, only: :clone
+  before_action :production_change!, if: :production_change?, only: :update
```
grosser: -> `ensure_non_production` and then inline the checker logic ? ... easier to read and less clutter here ...

app/controllers/api/deploy_groups_controller.rb
apanzerj: changed to: `current_project.stages.find(params[:id])`

app/controllers/api/stages_controller.rb
```diff
@@ -19,10 +18,25 @@ def clone
     end
   end
 
+  def update
+    stage.deploy_groups = deploy_groups
```
grosser: want to accept all params just like stages_controller ?

grosser: ... good enough for now, but I'd expect someone wants to do more then deploy_groups :D

app/controllers/api/stages_controller.rb
```diff
@@ -36,4 +50,17 @@ def deploy_group_ids
   def deploy_groups
```
grosser: params should not be deploy_group_ids, but params[:stage][:deploy_group_ids] -> open for future additions ?

app/controllers/concerns/deploy_groups_common.rb
grosser: take it out :D

app/controllers/stages_controller.rb
grosser: any answer for this ?

app/controllers/api/deploy_groups_controller.rb
apanzerj: roger

app/models/project.rb
```diff
@@ -43,6 +43,18 @@ class Project < ActiveRecord::Base
 
   scope :search, ->(name) { where("name like ?", "%#{name}%") }
 
+  def template
```
grosser: can make this a has_one -> get caching from rails ... also name it `template_stage` ?

app/controllers/api/deploys_controller.rb
apanzerj: sure

app/controllers/api/deploys_controller.rb
apanzerj: zapped

app/controllers/api/deploy_groups_controller.rb
grosser: yeah

app/controllers/api/deploys_controller.rb
apanzerj: yup

app/controllers/api/projects_controller.rb
apanzerj: I'll write a test and let you know

apanzerj: According to test: yes

app/controllers/api/stages_controller.rb
apanzerj: 404 will raise in an api client, so we don't necessarily want to do that... right?

app/controllers/api/stages_controller.rb
apanzerj: zapping the logic. was told it was overkill

app/controllers/api/stages_controller.rb
grosser: I think we want that / it's the normal response for 'not found' :D

app/controllers/stages_controller.rb
apanzerj: sorry. mega pr. 

I'll add kwargs. 

There's two modes because, if you simply want to unset the current template, you need a way to do that without specifying a new stage as template.

app/controllers/concerns/deploy_groups_common.rb
apanzerj: out.

app/models/project.rb
apanzerj: done.

config/routes.rb
apanzerj: even if they did, it wouldn't be project/id/stages/template, it would be project/id/stages/stage_id

test/controllers/api/deploys_controller_test.rb
apanzerj: done

app/controllers/api/deploy_groups_controller.rb
apanzerj: taken out.

app/controllers/api/stages_controller.rb
apanzerj: done.

app/controllers/api/stages_controller.rb
apanzerj: Fixed. New params structure is: 

{ stage: { source_stage_id: , name: , deploy_group_ids: } }

app/models/project.rb
apanzerj: Do you mean:

```
class Project < ActiveRecord::Base
  has_one: :template_stage
end

class TemplateStage < Stage
end
```

grosser: something like:

has_one :template_stage, class_name: 'Stage', -> { where(template: true) }

jonmoter:
@apanzerj could you try to break this work into smaller PRs? This PR now has 48 files changes and 146 comments.

If you made each of those 5 bullet points in the description a separate PR, it would be way easier to manage this.

jonmoter:
Or at least have future PRs be more targeted?

apanzerj:
Yeah. This one got away from me. Sorry!

apanzerj:
This last change should get me a green build. I can try to break this up
once that goes through.

apanzerj:
@jonmoter @grosser last one!

config/initializers/statsd.rb
```diff
@@ -31,7 +31,7 @@ class << Samson
 Samson.statsd = Samson::StatsdLoader.create
 Samson.statsd.namespace = "samson.app"
 
-Samson.statsd.event "Startup", "Samson startup" if ENV['SERVER_MODE']
+Samson.statsd.event "Startup".dup, "Samson startup" if ENV['SERVER_MODE']
```
grosser: no longer need this after statsd update

test/controllers/api/stages_controller_test.rb
```diff
@@ -26,28 +38,8 @@
...
-          @controller.stubs(:params).returns(ActionController::Parameters.new(stage_name: "Foo"))
-        end
-
-        it 'uses the provided name' do
-          @controller.send(:stage_name).must_equal "Foo"
-        end
-      end
-    end
-
     it 'renders a cloned stage' do
```
grosser: `it "clones a stage"` ?

app/controllers/stages_controller.rb
```diff
@@ -102,4 +99,12 @@ def stage_params
...
     @stage = current_project.stages.find_by_param!(params[:id])
   end
+
+  def template_param
+    @template_param ||= ActiveRecord::Type::Boolean.new.type_cast_from_database(stage_params.delete("template"))
+  end
+
+  def update_template
+    template_param ? @stage.project.template_stage!(@stage) : @stage.project.reset_template_stage!
+  end
```
grosser: this still looks wrong to me:
 - removing the template param -> remove the attribute -> simple assignment
 - not changing then template param -> no change ... bug: atm it unsets the template
 - adding template param -> set the attribute -> can be a assignment + after-save remove it from all previous stage

test/controllers/stages_controller_test.rb
```diff
@@ -221,13 +234,29 @@
         end
 
         describe 'valid attributes' do
+          [[true, "0"], [false, "1"]].each do |bool, val|
```
grosser: if you submit as true/false this tests gets a bit simpler and the code makes more sense :D

app/controllers/stages_controller.rb
apanzerj: Yeah. I see what you're saying.

PR #1199, opened 2016-08-11
Change coverage to get private repo green

sandlerr:


sandlerr:
@danielbreves @grosser this would be red if the uncovered count was not correct wouldn't it?

danielbreves:
👍

grosser:
it is only red if the uncovered count is higher then expected ...

should add a comment here to say with it's `3` ...

sandlerr:
@ben @grosser does this look OK by you?

sandlerr:
@zendesk/samson I'm not sure why this was green previously, we get coverage errors in the travis script for a private repo.

grosser:
👍

PR #1200, opened 2016-08-11
FrozenStringLiteralComment Cop

apanzerj:
Since we are now on Ruby 2.3 I had to update Rubocop to use that parser. Doing that added the requirement that we get ready for ruby 3.0. 

Link: http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FrozenStringLiteralComment

Thoughts? 

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team


apanzerj:
So many fails... I'll leave this open but wait for feedback before fixing more...

sandlerr:
:+1: on pushing it through

grosser:
:+1: frozen strings is the future ...

apanzerj:
Okay. The best fix for this is to do:

"Fooo #{bar}" ?

apanzerj:
(I'll take care of this)

apanzerj:
@jonmoter / @grosser you online? Can haz +1?

apanzerj:
I separated the changes out into two pieces: 

1. The actual adding of the header and removing `.freeze` 
2. Fixing issues of string mutation.

app/controllers/integrations/base_controller.rb
```diff
@@ -119,7 +120,7 @@ def latest_release
   end
 
   def record_log(level, message)
-    (@recorded_log ||= "") << "#{level.upcase}: #{message}\n"
+    (@recorded_log ||= "".dup) << "#{level.upcase}: #{message}\n".dup
```
grosser: don't need the second dup here ?

app/models/secret_storage.rb
```diff
@@ -138,9 +139,9 @@ def vault_path(key)
       def convert_path(string, direction)
         string = string.dup
         if direction == :decode
-          ENCODINGS.each { |k, v| string.gsub!(v.to_s, k.to_s) }
+          ENCODINGS.each { |k, v| string = string.gsub(v.to_s, k.to_s) }
```
grosser: string is duped above, why do we need this gsub change ?

config/application.rb
```diff
@@ -168,7 +169,8 @@ class Application < Rails::Application
     unless ENV['PRECOMPILE']
       config.after_initialize do
         # Token used to request badges
-        config.samson.badge_token = Digest::MD5.hexdigest('badge_token' << Samson::Application.config.secret_key_base)
+        config.samson.badge_token = \
+          Digest::MD5.hexdigest('badge_token'.dup << Samson::Application.config.secret_key_base)
```
grosser: use `+` instead of dup+amend ?

grosser:
a few minor stragenesses, but I think 👍 

... I hope there is a better way of doing unfrozen strings then `.dup` ... we can check that out later ...

PR #1201, opened 2016-08-13
Handle Kubernetes deploy failures [PAAS-287]

jonmoter:
* Add kubeclient functionality for doing Deployment rollbacks
* Delete a deployment or daemonset if there's no previous version
* Do a rollback if there's a previous version

/cc @zendesk/paas

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-287

### Risks
- Level: Low (kubernetes only)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -306,9 +314,13 @@ def deploy(release_docs)
 
     def deploy_to_cluster(release, deploys)
       deploy(deploys)
-      success = wait_for_resources_to_complete(release, deploys)
-      show_failure_cause(release) unless success
-      success
+      successful = wait_for_resources_to_complete(release, deploys)
+      unless successful
+        rollback(deploys)
+        show_failure_cause(release)
```
grosser: better show the failure cause first, otherwise if something hangs we will never know

grosser:
I'd say just add a test to deploy-executor that says delete_or_rollback is getting called ... and assert that the output also shows some "Rolling back" message ... then unit-tests for delete_or_rollback

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -241,6 +241,14 @@ def ensure_build_is_successful(build)
       end
     end
 
+    def rollback(release_docs)
+      release_docs.each do |release_doc|
+        action = release_doc.previous_deploy ? 'Rolling back' : 'Deleting'
+        @output.puts "#{action} #{release_doc.deploy_group.name} role #{release_doc.kubernetes_role.name}"
+        release_doc.revert
+      end
+    end
```
jonmoter: At some point we might want to add code to do these in parallel, rather than one at a time

jonmoter:
This is ready for review now, @grosser

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -25,31 +27,62 @@ def client
...
+        if @previous_deploy
+          extension_client.rollback_deployment(resource_name, namespace)
+        else
+          delete_deployment
+        end
+      elsif daemon_set?
+        delete_daemon_set @new_deploy if @new_deploy
+        extension_client.create_daemon_set(@previous_deploy) if @previous_deploy
+      elsif job?
+        extension_client.delete_job(resource_name, namespace)
```
grosser: have an `else` with `raise` here ?

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -25,31 +27,62 @@ def client
...
       elsif job?
         # FYI per docs it is supposed to use batch api, but extension api works
         job = Kubeclient::Job.new(resource_template)
-        if deployed
-          extension_client.delete_job kubernetes_role.resource_name, job.metadata.namespace
+        if @previous_deploy
+          extension_client.delete_job resource_name, namespace
         end
         extension_client.create_job job
-      else
```
grosser: maybe just `else raise`

grosser:
👍  looks like it could work :)

plugins/kubernetes/app/models/kubernetes/release_doc.rb
jonmoter: I now do that check up top, on line 47.

PR #1203, opened 2016-08-15
SAMSON-269 custom dockerfile for projects so uncommon projects can be supported

grosser:
@jonmoter @zendesk/paas @vhiairrassary 

https://zendesk.atlassian.net/browse/SAMSON-269

screen shot 2016-08-15 at 1 38 15 pm

building from a branch that has a custom dockerfile .... works :)

screen shot 2016-08-15 at 1 38 10 pm

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -24,12 +24,9 @@ class DockerBuilderService
       repository.executor = execution.executor
 
       if build.kubernetes_job
-        run_build_image_job(job, image_name, push: push, tag_as_latest: tag_as_latest)
-      elsif build_image(tmp_dir)
-        ret = true
-        ret = push_image(image_name, tag_as_latest: tag_as_latest) if push
-        build.docker_image.remove(force: true) unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
-        ret
+        build_image_via_kubernetes_job(job, image_name, push: push, tag_as_latest: tag_as_latest)
+      else
+        build_image_locally(job, image_name, push: push, tag_as_latest: tag_as_latest, tmp_dir: tmp_dir)
       end
     end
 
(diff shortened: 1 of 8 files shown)


irwaters:
👍 looks sane to me.  may want @jonmoter to chime in too

app/models/docker_builder_service.rb
```diff
@@ -71,9 +80,12 @@ def build_image(tmp_dir)
     output.puts("### Running Docker build")
 
     build.docker_image =
-      Docker::Image.build_from_dir(tmp_dir, {}, Docker.connection, registry_credentials) do |output_chunk|
-        handle_output_chunk(output_chunk)
-      end
+      Docker::Image.build_from_dir(
+        tmp_dir,
```
jonmoter: A thought just occurred to me. We might need to specify what the root directory is for doing the Docker build.

For example, if there's a repo with a file called `$root/scala/Dockerfile`, there's a decent chance the base directory should be `$root/scala`, instead of `$root`.

app/models/docker_builder_service.rb
```diff
@@ -71,9 +80,12 @@ def build_image(tmp_dir)
     output.puts("### Running Docker build")
```
jonmoter: Before we attempt a build, should we validate that the Dockerfile exists?

grosser: it will fail nicely ;)

app/models/docker_builder_service.rb
grosser: true ... useless for bime without

sandlerr:
might want to update https://github.com/zendesk/samson/blob/master/plugins/dockerb/lib/samson_dockerb/samson_plugin.rb#L8 or you could use https://github.com/zendesk/samson/pull/767 and just `mv MyDockerfile Dockerfile`

grosser:
needs rework to change directories and nobody wants it atm ... closing ...

PR #1205, opened 2016-08-15
PAAS-295 do not stage files in .dockerignore

grosser:
shipping extra build context can be slow and is useless since local builds do not have them either

mimic docker cli / .gitignore way of handling ignore files

https://github.com/swipely/docker-api/pull/438

## Risks
 - med: docker builds failing because files are missing from the build context


grosser:
@zendesk/paas

irwaters:
👍 on green (look like travis can't download the gem)

grosser:
finally git that merged ... confirmed working on a local deploy :)

PR #1206, opened 2016-08-16
Slack UX improvements

ben:
Quick: which one of these deploys failed?

image

Hard to tell from a glance. This PR improves the Slack UX for Samson in a few key ways:

- Adds easy-to-recognize status emoji to deploy notifications
- Improves our use of Slack markup
- Fixes the `@here` mention when requesting a buddy
- Adds the nice attachment feature from the [slack_app plugin](https://github.com/zendesk/samson/pull/1130) to the buddy-request message

Now it's easy to tell when a deploy succeeds:

image

/cc @zendesk/samson

### Tasks
 - [x] Status emoji in hook notifications
 - [x] Linkify deploy name in hook notification
 - [x] Fix the `@here` mention on the buddy request
 - [x] Add the PRs/Risks attachment to the buddy request
 - [ ] :+1: from team

### Risks
- Level: Low. This is just a tweak to the output when sending things to Slack.

### DIFF
--- a/app/assets/javascripts/jquery-mentions-input/jquery.mentionsInput.js
+++ b/app/assets/javascripts/jquery-mentions-input/jquery.mentionsInput.js
@@ -470,7 +470,7 @@
     //Resets the text area
     function resetInput(currentVal) {
       mentionsCollection = [];
-      var mentionText = utils.htmlEncode(currentVal);
+      var mentionText = currentVal;
       var regex = new RegExp("(" + settings.triggerChar + ")\\[(.*?)\\]\\((.*?):(.*?)\\)", "gi");
       var match, newMentionText = mentionText;
       while ((match = regex.exec(mentionText)) !== null) {
--- a/plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
+++ b/plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
@@ -31,9 +31,10 @@ module SamsonSlackWebhooks
       end
     end
 
-    def deliver_message_via_webhook(webhook:, message:)
+    def deliver_message_via_webhook(webhook:, message:, attachments:)
       payload = { text: message, username: 'samson-bot' }
       payload[:channel] = webhook.channel unless webhook.channel.blank?
+      payload[:attachments] = attachments if attachments.present?
 
       Faraday.post(webhook.webhook_url, payload: payload.to_json)
     rescue Faraday::ClientError => e
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -13,6 +13,7 @@ engines:
       - 66e8b888d5f032420a906b6210ecf8e1
       - a9e828eeed8e535cc0a9cd801e5ce8c8
       - 828106164429317de33715fe1b5251f2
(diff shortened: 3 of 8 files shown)


app/assets/javascripts/jquery-mentions-input/jquery.mentionsInput.js
```diff
@@ -470,7 +470,7 @@
     //Resets the text area
     function resetInput(currentVal) {
       mentionsCollection = [];
-      var mentionText = utils.htmlEncode(currentVal);
+      var mentionText = currentVal;
```
grosser: this looks like it was some kind of security measure ... do we not need that ?

grosser:
UI looks much better ... code looks good except the 🐟 -y escape removal ...

app/assets/javascripts/jquery-mentions-input/jquery.mentionsInput.js
ben: This plugin is only ever used in the buddy-request view, and this was converting `<!here>` to `&lt;!here&gt;` after document-ready, which is what was breaking the `@here` mention in Slack. The content being filtered here is just going into a textarea, so it should be okay to allow "raw" html tags in there. If nobody is using this mention auto-complete plugin, we could just remove the whole thing.

ben:
I just finished adding PRs and Risks to the buddy attachment. Now it looks like this:

screen shot 2016-08-16 at 10 04 29 am

Also, tests now pass. Can I get a final review?

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
```diff
@@ -31,9 +31,10 @@ def users
       end
     end
 
-    def deliver_message_via_webhook(webhook:, message:)
+    def deliver_message_via_webhook(webhook:, message:, attachments:)
       payload = { text: message, username: 'samson-bot' }
       payload[:channel] = webhook.channel unless webhook.channel.blank?
+      payload[:attachments] = attachments if attachments.present?
```
grosser: any test for this ?

grosser: basically the whole addition of risks seems to be untested ...

grosser:
looks good

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
ben: Added tests.

ben:
Green tests! Can I get a thumb?

grosser:
👍

app/assets/javascripts/jquery-mentions-input/jquery.mentionsInput.js
andreionut: @ben @grosser I removed the need to use the Slack tags by enabling full parsing: https://github.com/zendesk/samson/pull/1208

We probably should reinstate this as it is a security issue.

zehua: The security risk is low given that the value comes from hardcoded values we put into the code. The removal of this also allows us to use the slack link format, which greatly helps improve the UI.

PR #1207, opened 2016-08-16
extract and test secret key resolver / backends

grosser:
@jonmoter @irwaters 

some cleanup for last PR ... no longer mashing up all classes into a single file


irwaters:
👍 lgtm

PR #1208, opened 2016-08-16
Correct fix for Slack @here notifications

andreionut:
* I've enabled Slack message full parsing mode
* This way, Slack can deal with the formatting an we just write the messages like in a chat text box

/cc @zendesk/samson @grosser @zehua 

### Tasks
 - [ ] :+1: from team

### References
 - https://api.slack.com/docs/message-formatting#parsing_modes

### Risks
- Level: Low

### DIFF
--- a/plugins/slack_webhooks/app/models/slack_webhook_notification.rb
+++ b/plugins/slack_webhooks/app/models/slack_webhook_notification.rb
@@ -18,8 +18,7 @@ class SlackWebhookNotification
 
   def default_buddy_request_message
     project = @deploy.project
-    # https://api.slack.com/docs/message-formatting
-    ":pray: <!here> _#{@deploy.user.name}_ is requesting approval to deploy " \
+    ":pray: @here _#{@deploy.user.name}_ is requesting approval to deploy " \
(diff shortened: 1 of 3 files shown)


grosser:
👍

zehua:
I thought we would disable html encoding on the text (see https://api.slack.com/docs/message-formatting#how_to_escape_characters).

But this way also works. Should we use `link_names: 1` instead of `parse: full` so that other formatting such as links would still work?

zehua:
👍  after fixing merge conflicts

andreionut:
@zehua `parse: full` parses all the message, from mentions to links and we want to have that `@here` become a mention.
I tried to disable html encoding, but it was done in a JS function shared by all the buddy check plugins somehow.
I think this is the best approach as it allows people to write the message as they would in a Slack text box, no more magic tags that we have to whitelist.

zehua:
> parse: full parses all the message, from mentions to links and we want to have that @here become a mention.

yeah, `parse: full` will turn `@here` into a mention. But it will also encode `<http://link.com|link text>` and not show it as a link.

andreionut:
Think this: https://github.com/zendesk/samson/pull/1206 conflicts tis PR, so I'm going to revert mine

PR #1209, opened 2016-08-16
PAAS-292 run secret and role verifications before we create a slow build or even a release

grosser:
@jonmoter @irwaters 

creating builds / releases / release docs is slow and pollutes our DB with invalid and unnecessary rows ... so run validations before we try to start a build

ideally this would not duplicate the validation runs, but that's a bit harder to achieve / cheap ...


grosser:
screen shot 2016-08-16 at 11 18 53 am

irwaters:
lgtm 👍

jonmoter:
A little extra overhead, but that will be minimal once we cache the keys.

jonmoter:
👍

PR #1211, opened 2016-08-16
expand wildcard secrets

grosser:
need to do this in samson to get verification support ... also easier to have all the logic in 1 place

tried using vault.list with * ... but that did not return anything ...

https://github.com/zendesk/samson_secret_puller/pull/20

FOO_* with foo_* -> [[FOO_BAR, a/a/a/foo_bar], [FOO_BAZ, a/a/a/foo_baz]]

@irwaters @jonmoter

### DIFF
--- a/lib/samson/secrets/key_resolver.rb
+++ b/lib/samson/secrets/key_resolver.rb
@@ -3,6 +3,8 @@
 module Samson
   module Secrets
     class KeyResolver
+      WILDCARD = '*'
+
       def initialize(project, deploy_groups)
         @project = project
         @deploy_groups = deploy_groups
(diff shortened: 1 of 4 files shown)


lib/samson/secrets/key_resolver.rb
```diff
@@ -11,23 +13,29 @@ def initialize(project, deploy_groups)
 
       # expands a key by finding the most specific value for it
       # bar -> production/my_project/pod100/bar
-      def expand!(secret_key)
-        key = secret_key.split('/', 2).last
+      def expand(env_name, secret_key)
+        if env_name.end_with?(WILDCARD) ^ secret_key.end_with?(WILDCARD)
+          @errors << "#{env_name} and #{secret_key} need to both end with #{WILDCARD} or not include them"
```
irwaters: 2.3.1 :008 > WILDCARD='*'
 => "*" 
2.3.1 :009 > env_name="FOO_*"
 => "FOO_*" 
2.3.1 :010 > secret_key="blah*"
 => "blah*" 
2.3.1 :011 > env_name.end_with?(WILDCARD) ^ secret_key.end_with?(WILDCARD)
 => false 

does this work?

lib/samson/secrets/key_resolver.rb
```diff
@@ -42,6 +50,39 @@ def verify!
...
+          [[env_name, found]]
+        else
+          []
+        end
+      end
+
+      # FOO_* with foo_* -> [[FOO_BAR, a/a/a/foo_bar], [FOO_BAZ, a/a/a/foo_baz]]
+      def expand_wildcard_keys(env_name, secret_key, possible_ids)
+        # look through all keys to check which ones match
+        all = SecretStorage.keys
```
irwaters: this is expensive, will need to get the caching in.

irwaters:
I think this is fine, we should figure out caching asap.  I like the fact this is happening in samson though, as it'll only happen once per deployment.

PR #1213, opened 2016-08-17
Revert "Correct fix for Slack @here notifications"

andreionut:
Reverts zendesk/samson#1208
PR: https://github.com/zendesk/samson/pull/1206 already fixed the `@here` issue and zendesk/samson#1208 only broke the functionality

/cc @zendesk/samson @grosser @ben

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


fneves:
👍

PR #1218, opened 2016-08-17
expand secret annotations and test for it

grosser:
previously this worked because we used replace, but never tested that it works


grosser:
@irwaters

irwaters:
👍

PR #1220, opened 2016-08-17
Api::DeployGroupsController

apanzerj:
* Adding Api::DeployGroupsController

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/controllers/api/deploy_groups_controller.rb
+++ b/app/controllers/api/deploy_groups_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+class Api::DeployGroupsController < Api::BaseController
+  before_action :ensure_enabled
+
+  def index
+    scope = current_project.present? ? stage.deploy_groups : DeployGroup.all
+    render json: paginate(scope.sort_by(&:natural_order))
+  end
+
+  protected
+
+  def stage
+    current_project.stages.find(params[:id])
+  end
+
+  def ensure_enabled
+    return if DeployGroup.enabled?
+    render json: {message: "DeployGroups are not enabled."}, status: :precondition_failed
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,10 +13,16 @@ Samson::Application.routes.draw do
     resources :deploy_groups, only: [:index]
 
     resources :projects, only: [:index] do
+      resources :stages, only: [:index] do
+        member do
+          get :deploy_groups, to: 'deploy_groups#index'
+        end
+      end
+
       resources :deploys, only: [:index]
     end
 
-    resources :stages, only: [:index] do
+    resources :stages, only: [] do
       get :deploys, to: 'deploys#index'
       post :clone, to: 'stages#clone'
     end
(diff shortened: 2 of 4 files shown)


app/controllers/api/deploy_groups_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+class Api::DeployGroupsController < Api::BaseController
+  before_action :ensure_enabled
+
+  def index
+    scope = current_project.present? ? stage.deploy_groups : DeployGroup.all
+    render json: paginate(scope.sort_by(&:natural_order))
```
grosser: paginate does not make much sense if we show all ?

config/routes.rb
```diff
@@ -13,10 +13,16 @@
     resources :deploy_groups, only: [:index]
 
     resources :projects, only: [:index] do
+      resources :stages, only: [:index] do
+        member do
```
grosser: use `resources :deploy_groups, only: [:index]` instead ?

app/controllers/api/deploy_groups_controller.rb
apanzerj: wat? paginate breaks them into clumps of 1000.

if there is less than 1000, then yes, paginate is pointless, but >1000 and it doesn't show them all.

test/controllers/api/deploy_groups_controller_test.rb
```diff
@@ -0,0 +1,61 @@
...
+               to: "api/deploy_groups#index", params: {project_id: "1", id: "2"}
+
+  oauth_setup!
+
+  it 'returns precondition_failed when deploy_groups are disabled' do
+    get :index
+    assert_response :precondition_failed
+  end
+
+  describe 'when deploy_groups are enabled' do
```
grosser: can avoid the wrapping describe with:

```Ruby
before { DeployGroup.stubs(:enabled?).returns(true) }

it 'returns precondition_failed when deploy_groups are disabled' do
  DeployGroup.unstub(:enabled?)
  ...
end


describe "#index" ...
```

config/routes.rb
apanzerj: Yeah, kind of worried about that, I don't want to do three levels of resources foo.

grosser:
👍

app/controllers/api/deploy_groups_controller.rb
grosser: we still fetch them from the db ... but I guess we don't have to serialize them ... I guess nobody has >1000 ... so fine either way ...

test/controllers/api/deploy_groups_controller_test.rb
apanzerj: unstub. neat.

app/controllers/api/deploy_groups_controller.rb
apanzerj: I see. Hmm. Yeah, maybe skip the paginate anyways? I think we only need it for things like deploys which are definitely more than 1000?

grosser: nice for consistency I guess ...

PR #1223, opened 2016-08-18
make secret keys fast for hashiecorp vault and other slow backends

grosser:
@irwaters @jonmoter 

cache keys and expire them on write/delete

alternatively this could be done at the backend level, but this way is universal and easier to test


irwaters:
this should work 👍

PR #1225, opened 2016-08-18
PAAS-339 allow users to see some secrets

grosser:
we have lots of secrets that are not really secret ... so allow to show them

also storing creator_id for vault backend

@irwaters @jonmoter 

screen shot 2016-08-18 at 1 50 44 pm
screen shot 2016-08-18 at 1 49 38 pm

### DIFF
--- a/app/controllers/admin/secrets_controller.rb
+++ b/app/controllers/admin/secrets_controller.rb
@@ -28,10 +28,8 @@ class Admin::SecretsController < ApplicationController
   end
 
   def update
-    attributes = {
-      user_id: current_user.id,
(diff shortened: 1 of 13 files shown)


irwaters:
wfm 👍

grosser:
confirmed working in master :D 

screen shot 2016-08-18 at 2 14 50 pm

jonmoter:
Suggestion: add a tooltip or bit of text explaining what the "Visible" checkbox does in the UI, 'cause it's not immediately obvious to me with that word.

Otherwise 👍

grosser:
added `title: 'Value visible in samson edit UI'`

PR #1227, opened 2016-08-18
plain migration

craig-day:
Something is really broke with the schema file.  All I did was

``` sh
bundle exec rake db:drop
bundle exec rake db:create
bundle exec rake db:migrate
```

/cc @zendesk/samson @zendesk/deploy @jcheatham @adam-hart 
### Risks
- Medium: Schema is in a real weird state.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -14,17 +14,17 @@
 ActiveRecord::Schema.define(version: 20160818194622) do
 
   create_table "builds", force: :cascade do |t|
-    t.integer  "project_id",                                       null: false
+    t.integer  "project_id",          limit: 4,                    null: false
     t.integer  "number",              limit: 4
-    t.string   "git_sha",             limit: 255,                  null: false
+    t.string   "git_sha",             limit: 128,                  null: false
     t.string   "git_ref",             limit: 255,                  null: false
-    t.string   "docker_image_id",     limit: 255
+    t.string   "docker_image_id",     limit: 128
     t.string   "docker_ref",          limit: 255
     t.string   "docker_repo_digest",  limit: 255
-    t.integer  "docker_build_job_id"
+    t.integer  "docker_build_job_id", limit: 4
     t.string   "label",               limit: 255
     t.string   "description",         limit: 1024
-    t.integer  "created_by"
+    t.integer  "created_by",          limit: 4
     t.datetime "created_at"
     t.datetime "updated_at"
     t.boolean  "kubernetes_job",                   default: false, null: false
@@ -35,7 +35,7 @@ ActiveRecord::Schema.define(version: 20160818194622) do
   add_index "builds", ["project_id"], name: "index_builds_on_project_id", using: :btree
 
   create_table "commands", force: :cascade do |t|
-    t.text     "command",    limit: 10485760
+    t.text     "command",    limit: 16777215
     t.datetime "created_at"
     t.datetime "updated_at"
     t.integer  "project_id", limit: 4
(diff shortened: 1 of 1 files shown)


db/schema.rb
```diff
@@ -490,7 +480,7 @@
     t.integer  "item_id",    limit: 4,          null: false
     t.string   "event",      limit: 255,        null: false
     t.string   "whodunnit",  limit: 255
-    t.text     "object",     limit: 1073741823
+    t.text     "object",     limit: 4294967295
```
grosser: this is invalid ...

db/schema.rb
```diff
@@ -377,16 +377,6 @@
 
   add_index "secrets", ["id"], name: "index_secrets_on_id", unique: true, length: {"id"=>191}, using: :btree
 
-  create_table "slack_channels", force: :cascade do |t|
```
grosser: I think this is not having slack enabled ...

grosser: we had some old slack plugin ... I think we forgot to remove it's columns

db/schema.rb
```diff
@@ -156,7 +156,7 @@
     t.integer  "user_id",    limit: 4,                              null: false
     t.integer  "project_id", limit: 4,                              null: false
     t.string   "status",     limit: 255,        default: "pending"
-    t.text     "output",     limit: 1073741823
+    t.text     "output",     limit: 4294967295
```
grosser: invalid

db/schema.rb
```diff
@@ -35,7 +35,7 @@
   add_index "builds", ["project_id"], name: "index_builds_on_project_id", using: :btree
 
   create_table "commands", force: :cascade do |t|
-    t.text     "command",    limit: 10485760
+    t.text     "command",    limit: 16777215
```
grosser: also invalid I'd guess

grosser:
I like this ... maybe clean up slack channels and add some kind of addition to migrate that will convert big  text columns to small ones ?

db/schema.rb
craig-day: I'm not sure what the previous value was, but this is 2^24 which is what a MySQL MEDIUMTEXT column gets set to.

craig-day:
Probably easier to view [without whitespace](https://github.com/zendesk/samson/pull/1227?w=1)

db/schema.rb
grosser: we changed it to 10485760 because anything bigger blew up on travis ... if
you get it green then that's fine ;)

jonmoter: Samson supports using MySQL, Postgres, and SQLite. So some of these settings are the least common denominator of the three.

craig-day:
@grosser @jonmoter I added some code to modify the `db:schema:dump` and `db:schema:load` tasks that will create/use the schema file specific to the adapter you are using.  This should allow people to write and run migrations and not have to manually modify the `schema.rb` file.  I'm working on altering the `db:migrate` tasks so it updates all three schema files, but its kinda a rabbit hole.  Thoughts on the approach?

grosser:
not a fan of multiple schemas ... makes things messier ...
for now how about just merging the clean + hand-corrected schema ?
next PR can be automated cleanup ?

craig-day:
I'm not sure if its worth a merge without automation.  It just generated the file in this format cause I'm using mysql.  If I switched to sqlite then I don't think the file would change at all. Its just a battle of who migrates on what system. Maybe it is a battle that isn't worth fighting.

grosser:
I think this PR already has value in cleaning up the schema ... next PR
would be much smaller ...
don't think anyone uses sqlite ... maybe best to make it produce something
consistent but universal when migration using mysql ...

PR #1228, opened 2016-08-18
PAAS-340 allow commenting on secrets

grosser:
also:
 - make not submitting visible work
 - show '-- Value is hidden --' when value is not shown instead of confusing empty form

screen shot 2016-08-18 at 3 12 12 pm
screen shot 2016-08-18 at 3 24 41 pm

@irwaters @jonmoter

### DIFF
--- a/db/migrate/20160818210955_add_comment_to_secrets.rb
+++ b/db/migrate/20160818210955_add_comment_to_secrets.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+class AddCommentToSecrets < ActiveRecord::Migration
+  def change
+    add_column :secrets, :comment, :string
+  end
+end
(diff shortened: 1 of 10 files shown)


db/migrate/20160818210955_add_comment_to_secrets.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddCommentToSecrets < ActiveRecord::Migration
+  def change
+    add_column :secrets, :comment, :string
```
jonmoter: Think the default 256-character limit is sufficient for a comment?

grosser: I'd hope so ... if not we can expand easily

grosser:
improved UI: invisible secrets have comments/visible/value disabled so they are not updated by accident ... can be enabled via js click event

screen shot 2016-08-18 at 4 39 38 pm
screen shot 2016-08-18 at 4 39 34 pm

jonmoter:
LTGM. 👍

PR #1229, opened 2016-08-19
remove unnecessary vault.json write/read logic

grosser:
noticed vault.json always appearing in my samson folder ... code looked like it's not needed ...

@irwaters


irwaters:
not sure about this.  ideally it should fail if the config isn't there and we are using the vault storage engine.

grosser:
how would it ever fail if we write the config on initialize ? ... did not see anyone else reading the vault.json ... only test that checked that the config exists was testing the test-stub ...

irwaters:
If you think that the initializer blowing up when it can't read the config is enough, then fine :)

👍

PR #1230, opened 2016-08-19
bump gems to silence sprockets

grosser:
@apanzerj


grosser:
@adam-hart maybe since adam is busy with  🍼

adam-hart:
Stinkin' 🍼 👍

PR #1233, opened 2016-08-20
seed roles with empty branch / fail nicely when seeing project withou…

grosser:
screen shot 2016-08-19 at 9 20 59 pm

@jonmoter @irwaters


jonmoter:
👍

PR #1235, opened 2016-08-23
make delete warning a bit more helpful

grosser:
@mariozaizar 

screen shot 2016-08-22 at 9 07 56 pm


grosser:
real +1 would have been better ... but I'll take it ;)

PR #1236, opened 2016-08-24
allow the release number to be specified manually

sandlerr:
@zendesk/samson @grosser @miogalang

### DIFF
--- a/app/views/releases/new.html.erb
+++ b/app/views/releases/new.html.erb
@@ -3,14 +3,19 @@
   <section>
     <fieldset>
       <div class="row">
-        <p class="col-lg-offset-2 col-lg-4">Manual releases should only be created when your CI service is down, and you need to release a new version regardless. The commit will tagged as <%= @release.version %>, which will be pushed to the Git repository.</p>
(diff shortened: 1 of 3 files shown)


grosser:
screenshot ?

... some potential for conflict if 2 ppl open the new form at the same time ... but pretty intuitive ...

👍

app/views/releases/new.html.erb
```diff
@@ -3,14 +3,19 @@
   <section>
     <fieldset>
       <div class="row">
-        <p class="col-lg-offset-2 col-lg-4">Manual releases should only be created when your CI service is down, and you need to release a new version regardless. The commit will tagged as <%= @release.version %>, which will be pushed to the Git repository.</p>
+        <p class="col-lg-offset-2 col-lg-4">Manual releases should only be created when your CI service is down, and you need to release a new version regardless. The commit will tagged and the tag will be pushed to the Git repository.</p>
```
craig-day: s/The commit will tagged/The commit will be tagged

sandlerr:
There is a unique index that should crash in that case. The git push should probably not work either.

sandlerr:
screen shot 2016-08-26 at 9 52 43 am

PR #1237, opened 2016-08-24
project admin is unusable with tons of projects, add search

grosser:
@rikeoka @miogalang 

also moving delete into the admin controller since it is not reachable via normal controller and fit here

screen shot 2016-08-23 at 9 20 14 pm

### DIFF
--- a/test/controllers/admin/projects_controller_test.rb
+++ b/test/controllers/admin/projects_controller_test.rb
@@ -4,19 +4,60 @@ require_relative '../../test_helper'
 SingleCov.covered!
 
 describe Admin::ProjectsController do
+  let(:project) { projects(:test) }
+
+  before do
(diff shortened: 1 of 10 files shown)


test/controllers/admin/projects_controller_test.rb
```diff
@@ -4,19 +4,60 @@
...
+      it "searches" do
+        get :index, search: {query: 'x'}
+        assert_template :index
+        assigns[:projects].must_equal []
+      end
+    end
+
+    describe "#destroy" do
+      before do
+          delete :destroy, id: project.to_param
```
rikeoka: 2 char indent?

rikeoka:
LGTM, it's probably better permissions wise to have delete through admin/projects.
👍

PR #1239, opened 2016-08-25
delete deploy_group_stages once a stage is deleted so they do not bec…

grosser:
…ome invalid / blow up when rendering

https://zendesk.airbrake.io/projects/95346/groups/1760289710393468921/notices/1760290139726124697

Downside: restoring a stage will lose it's deploy_groups ... but that's not too bad ...

@jcheatham


grosser:
FYI @pswadi-zendesk deploygroup show works now ;)

jcheatham:
Not sure how often we delete deploy groups or that they need to be restored.
I've heard good things about this "soft deletion" gem?
👍

PR #1241, opened 2016-08-26
newer vault server no longer gets read as json

grosser:
https://github.com/hashicorp/vault-ruby/issues/88


grosser:
@irwaters @jonmoter

sethvargo:
I've tried really hard to reproduce the incorrect content-type header, and I'm not seeing it. Do you have something different in your setup?

What's the use case for `to_h` - yes it was removed, but it was not an API that was actually by-design (it was a by-product of having inherited from Struct)

grosser:
hmm the to_h patch should be be here ... turned that into a separate PR ...
if it's not a common issue then the header thing is most likely some custom issue on our end :(
... I'll update if we find it ...

mfischer-zd:
@sethvargo the root cause was https://github.com/hashicorp/vault/issues/1791 (incorrect Content-Type response header when using request forwarding), which will be fixed in Vault 0.6.2.

sethvargo:
Ah - okay. I had a feeling it was either: 1. Haproxy/load balance doing some weirdness or 2. the request forwarding

dragonfax:
👍

PR #1242, opened 2016-08-27
Pod-o-matic: Create new stages for deploy group.

dragonfax:
Migrations and UI for managing automatic pod creation.

? should this attribute be `environment_template` rather than `template_environment`?

## Template status show in stages list

On project view page

test_project

## Project Setting for Including in New Pods

samson

## Stage Settings, choose the template environment

editing_master__test_project_

## Deploy Group page, 

Button for creating missing stages, and a dialog for verification before doing so.

fullscreen_8_29_16__4_29_pm

### References
 - Jira link: https://zendesk.atlassian.net/browse/TOOLS-791

### Risks
- Level: Low

We're adding funcionality, not removing or changing it. But this does involve database migrations and new fields in existing forms.


app/views/projects/_form.html.erb
```diff
@@ -54,6 +54,14 @@
       </div>
     </div>
 
+    <div class="form-group">
+      <div class="col-lg-offset-2 col-lg-4">
+        <%= form.check_box :deploy_to_new_pods %>
+        <%= form.label :deploy_to_new_pods, 'Deploy to New Pods'%>
+        <span class="help-block">This a pod-specific service, and should be deployed to each pod, including new pods.</span>
```
grosser: isn't that the same as using as "Stage is using $DEPLOY_GROUPS" ?

app/views/stages/_fields.html.erb
```diff
@@ -61,6 +61,14 @@
     </div>
   </div>
 
+  <div class="form-group">
+    <%= form.label :name, "Template Environment", class: "col-lg-2 control-label" %>
+    <div class="col-lg-4">
+      <%= form.select :template_environment_id, Environment.all.collect {|e| [e.name, e.id] }, { include_blank: true }, {class: 'form-control' } %>
```
grosser: I don't think we have any stages that belong to multiple environments ... so not really useful ?

dragonfax:
Er, I didn't think you'd look at this until I was done, especially since its marked WIP and its a saturday. But thanks.

app/views/projects/_form.html.erb
dragonfax: Nothing says deploy_groups have to be pods. We have deploy groups that aren't pods.

app/views/stages/_fields.html.erb
dragonfax: How do we mark which stages are templates for which environment, then? Maybe `stage.deploy_groups.environment`? I don't know. That would let them choose more than one template per env, and we'd be, like, "which one do we use?"

This makes it explicit. Someone has to look at the stages and say "This stage, this environment". And nothing can go wrong.

I appreciate a normalized database as much as the next guy, but people aren't normalized.

app/views/projects/_form.html.erb
grosser: samson does not know about "pods" ... as long as changing the deploy groups makes the stage copy-able then that should be enough for us ... and that will only work when it uses $DEPLOY_GROUPS ... and so far I have only seen that being used directly via cap `$DEPLOY_GROUPS` or kubernetes ... a project that does not use deploy-groups in the usual way will not be copy-able and will afaik not use $DEPLOY_GROUPS either

app/views/stages/_fields.html.erb
grosser: Ideally let's discuss this on Monday :)

dragonfax:
Okay @grosser , lets do this.

Also, I'd like to deploy this to staging, so Nick D and Praveen can see it in action. Are you done with your testing there?

grosser:
staging needs the branch ... can you review https://github.com/zendesk/samson/pull/1241 ?
then you can rebase on that ...

grosser:
- not a fan of the 'template for' column ... this 1-off usecase should not clutter up the UI ... 99% of users don't need to see it / will be confused
 - samson does not know about "Pods" ... so the label cannot be "Deploy to new pods"
 - template environment selector: not needed ... stage can be template for every deploy group it deploys to -> "Can be cloned" / "Uses only $DEPLOY_GROUPS"

app/assets/stylesheets/admin/deploy_groups.scss
```diff
@@ -0,0 +1,3 @@
+.btn-disabled {
```
grosser: all stylesheets are global, putting this here overrides the other definition ... this was removed to allow titles of disabled buttons to show ... why do you need the events to be none ?

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 class Admin::DeployGroupsController < ApplicationController
   before_action :authorize_admin!
-  before_action :authorize_super_admin!, only: [:create, :new, :update, :destroy, :deploy_all, :deploy_all_now, :edit]
-  before_action :deploy_group, only: [:show, :edit, :update, :destroy, :deploy_all, :deploy_all_now]
+  before_action :authorize_super_admin!, only: [:create, :new, :update, :destroy, :deploy_all, :deploy_all_now,
```
grosser: 2-spacce align please

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -1,8 +1,9 @@
...
+                                                :create_all_stages, :edit]
+  before_action :deploy_group, only: [:show, :edit, :update, :destroy, :deploy_all, :deploy_all_now, :create_all_stages]
```
grosser: none of these stages were changed ... why do they all need the before_filter now ?

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -72,6 +73,22 @@ def deploy_all_now
     redirect_to deploys_path(ids: deploys.map(&:id))
   end
 
+  def create_all_stages
+    deploy_group.environment.template_stages.find_each do |template_stage|
```
grosser: a 1-off way of doing things ... inconsistent with the old way ... will be a confusing UI ... 
reuse the old UI ... only modify the way it gets data ?

app/controllers/dashboards_controller.rb
```diff
@@ -20,13 +20,14 @@ def prepare_dashboard
...
     @projects = Project.all
+    @failed_deploys = params[:failed_deploys] == "true" ? true : false
     @versions = project_versions(@before)
   end
 
   def project_versions(before_time)
     env_deploys = @deploy_groups.map(&:id)
     @projects.map do |project|
-      [project.id, project.last_deploy_by_group(before_time).select { |k| env_deploys.include?(k) }]
+      [project.id, project.last_deploy_by_group(before_time, @failed_deploys).select { |k| env_deploys.include?(k) }]
```
grosser: could use keyword args to make it more readable: `failed: @failed_deploys`

app/models/project.rb
```diff
@@ -144,9 +146,10 @@ def repository_homepage_gitlab
     "#{Rails.application.config.samson.gitlab.web_url}/#{gitlab_repo}"
   end
 
-  def deploys_by_group(before)
+  def deploys_by_group(before, include_failed = false)
     stages.each_with_object({}) do |stage, result|
-      deploy = stage.deploys.successful.where(release: true).where("deploys.updated_at <= ?", before.to_s(:db)).first
+      stage_filter = include_failed ? stage.deploys : stage.deploys.successful.where(release: true)
+      deploy = stage_filter.where("deploys.updated_at <= ?", before.to_s(:db)).first
```
grosser: don't think this needs `to_s(:db)`

app/views/dashboards/show.html.erb
```diff
@@ -1,9 +1,14 @@
...
 <section>
-  <p style="float: right;">
-    <a class="toggle" data-target="tr.no-deploys" href="#">Toggle Projects without Deploys</a>
-  </p>
+  <div style="float: right;">
+    <p>
+      <a class="toggle" data-target="tr.no-deploys" href="#">Toggle Projects without Deploys</a>
+    </p>
+    <p>
+      <a href="?failed_deploys=<%= !@failed_deploys %>">Toggle Failed Deploys</a>
```
grosser: do this the same way the other code works ? ... render all and then toggle via js ?

test/fixtures/deploys.yml
```diff
@@ -15,3 +15,21 @@ succeeded_production_test:
   started_at: 2014-01-02 20:00:00
   updated_at: 2014-01-02 20:10:00
   created_at: 2014-01-02 19:50:00
+
+succeeded_master_test:
```
grosser: modify existing fixtures to suit your usecase, we pay a price on every test run for every fixture we add

grosser:
failed deploys would be a good separate PR

dragonfax:
Still working on tests.

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -91,7 +100,7 @@ def new_stage_with_group(stage)
     stage = Stage.build_clone(stage)
     stage.deploy_groups << deploy_group
     stage.name = deploy_group.name
-    stage.name << " -- copy #{SecureRandom.hex(4)}" if stage.project.stages.where(name: stage.name).exists?
```
dragonfax: Remove the extra name bit, because we won't be creating duplicate stages for the same project/environment.

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -52,26 +53,34 @@ def destroy
...
 
-      stages = stages_in_same_environment(project, environment)
-      next unless deploy = stages.map(&:last_successful_deploy).compact.sort_by(&:created_at).last
-      stages.map { |s| [s, deploy] }
-    end.compact
-  end
+      last_success_deploy = template_stage.last_successful_deploy
+      next unless last_success_deploy
 
-  def deploy_all_now
```
dragonfax: We no longer need the interstitial page, as we won't be selecting and environment anymore, and will be deploying everything listed on the Deploy controller `show.html.erb` page.

app/views/admin/deploy_groups/deploy_all.html.erb
```diff
@@ -1,37 +0,0 @@
-<h1>Deploy to <%= @deploy_group.name %></h1>
```
dragonfax: Again, no need for the interstitial any longer.

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -52,26 +53,34 @@ def destroy
...
-    deploys = params.require(:stages).map do |stage|
-      stage_id, reference = stage.split("-", 2)
-      stage = Stage.find(stage_id)
-      stage = new_stage_with_group(stage) unless only_to_current_group?(stage)
       deploy_service = DeployService.new(current_user)
-      deploy_service.deploy!(stage, reference: reference)
+      deploy_service.deploy!(stage, reference: last_success_deploy.reference)
     end
+
     redirect_to deploys_path(ids: deploys.map(&:id))
```
dragonfax: I don't understand how this works. How does it redirect to a `show` page with multiple ids?

dragonfax:
still fixing tests.

dragonfax:
I cannot tell you whats up with codeclimate.

dragonfax:
Okay @grosser, Its that time again, my friend.

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -52,46 +53,47 @@ def destroy
   end
 
   def deploy_all
-    @stages = Project.all.flat_map do |project|
-      environment = Environment.find(params[:environment_id]) if params[:environment_id]
+    environment = deploy_group.environment
+    deploys = deploy_group.stages.map do |stage|
```
grosser: this makes it dangerous if one of the stages does to multiple deploy groups ... should be fine for our usecase ...

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -52,46 +53,47 @@ def destroy
...
-  def stages_in_same_environment(project, environment)
-    environment ||= deploy_group.environment
-    stages = project.stages.select do |stage|
-      stage.script.include?("$DEPLOY_GROUPS") && # is dynamic
-        stage.deploy_groups.where(environment: environment).exists? # is made to go to this environment
+  def create_all_stages
+    # No more than one stage, per project, per deploy_group
+    # Note: you can call this multiple times, and it will create missing stages, but no redundant stages.
+    environment = deploy_group.environment
+    Project.where(include_new_deploy_groups: true).each do |project|
```
grosser: slightly misleading since the deploy group does not need to be new to do a the deploy-all ...

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -52,46 +53,47 @@ def destroy
...
+      template_stage = environment.template_stages.where(project: project).first
+      deploy_group_stage = deploy_group.stages.where(project: project).first
```
grosser: could avoid N+1 here and fetch all the stages upfront ... same for the line above

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -52,46 +53,47 @@ def destroy
...
+      if template_stage && !deploy_group_stage
+        new_stage_with_group(template_stage)
```
grosser: this is not 'new' but 'create'

dragonfax:
Great, I'll stop by later today, and we have have a quick chat about the issues you find.

dragonfax:
Any idea how to fix code climate?

app/views/admin/deploy_groups/deploy_all.html.erb
grosser: can get rid of the deploy_all route then ?

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -34,6 +34,7 @@
   <div class="form-group">
     <div class="col-lg-offset-2 col-lg-10">
       <%= link_to "Edit", edit_admin_deploy_group_path(@deploy_group), class: "btn btn-primary" %>
+      <%= link_to "Create all stages", create_all_stages_admin_deploy_group_path(@deploy_group), class: "btn btn-default", data: { method: "post", confirm: "Create new stages for all pod-specific projects?" } %>
```
grosser: 'pod' -> 'that should be include in new deploy groups'

app/views/projects/_form.html.erb
```diff
@@ -54,6 +54,13 @@
       </div>
     </div>
 
+    <div class="form-group">
+      <div class="col-lg-offset-2 col-lg-4">
+        <%= form.check_box :include_new_deploy_groups %>
+        <%= form.label :include_new_deploy_groups, 'Include in New Deploy Groups'%>
```
grosser: maybe a `title` here with more details ... or a help bubble ... to make it less scary / mysterious

app/views/stages/_fields.html.erb
```diff
@@ -61,6 +61,19 @@
...
   </div>
 
+  <div class="form-group">
+    <div class="col-sm-offset-2 col-sm-10">
+      <div class="checkbox">
+        <%= form.label :is_template do %>
+          <%= form.check_box :is_template %>
+          Is a Template Stage
+        <% end %>
+        <span class="help-block">This stage can serve as a template to create new stages for its environment.</span>
```
grosser: ... by only changing it's deploy-group

... also hide when deploy-groups are disabled

app/views/projects/_form.html.erb
```diff
@@ -54,6 +54,13 @@
       </div>
     </div>
 
+    <div class="form-group">
```
grosser: ... also hide when deploy-groups are disabled

config/routes.rb
```diff
@@ -130,6 +130,7 @@
       member do
         get :deploy_all
```
grosser: deploy_all route can go away ? ... maybe replace deploy_all_now with it ...

db/migrate/20160827172550_add_deploy_to_new_pods_to_projects.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddDeployToNewPodsToProjects < ActiveRecord::Migration
+  def change
+    add_column :projects, :include_new_deploy_groups, :boolean
```
grosser: default: false, null: false

db/migrate/20160831151133_add_is_template_to_stage.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddIsTemplateToStage < ActiveRecord::Migration
+  def change
+    add_column :stages, :is_template, :boolean
```
grosser: here too

app/controllers/admin/deploy_groups_controller.rb
dragonfax: `template_stages = environment.template_stages
template_stages.select`

dragonfax: bitmoji

app/controllers/admin/deploy_groups_controller.rb
dragonfax: `create_stage_with_group!`

app/views/projects/_form.html.erb
dragonfax: roughly "will add new stages when using create-all for new deploy groups"

config/routes.rb
dragonfax: Button should be a post, not a get.

dragonfax:
Here we go @grosser. Comments addressed.

After this gets into master, Praveen and Nick R had a few more changes requested. But I'm doing things one at a time here.

app/models/project.rb
```diff
@@ -130,8 +130,8 @@ def with_lock(output: StringIO.new, holder:, error_callback: nil, timeout: 10.mi
     MultiLock.lock(id, holder, timeout: timeout, failed_to_lock: callback, &block)
   end
 
-  def last_deploy_by_group(before_time)
-    releases = deploys_by_group(before_time)
+  def last_deploy_by_group(before_time, include_failed_deploys = false)
```
grosser: keyword args ?

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -34,7 +34,17 @@
   <div class="form-group">
     <div class="col-lg-offset-2 col-lg-10">
       <%= link_to "Edit", edit_admin_deploy_group_path(@deploy_group), class: "btn btn-primary" %>
-      <%= link_to "Deploy all projects", deploy_all_admin_deploy_group_path(@deploy_group), class: "btn btn-default" if DeployGroup.enabled? %>
+      <% if DeployGroup.enabled? %>
+        <%= link_to "Create all stages",
+                    create_all_stages_admin_deploy_group_path(@deploy_group),
+                    class: "btn btn-default",
+                    data: {
+                        method: "post",
```
grosser: 4-space indent ?
... also reduce  outside indent to 2-space ?

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -34,7 +34,17 @@
...
+                        confirm: "Create all missing stages for this deploy group? All Project marked to be included in new deploy groups will get a stage for this deploy group."
```
grosser: `All Project marked to be included in new deploy groups that have a template stage for #{environment}`

grosser: ... use it as title instead of the confirm ?

grosser:
👍

PR #1243, opened 2016-08-30
always print a timestamp

sandlerr:
@zendesk/samson @jonmoter 

screen shot 2016-08-30 at 2 19 44 pm


grosser:
ugly!

grosser:
travis does some neat ascii codes to add timing information ... maybe we can steal that ... otherwise make the js end covert the timestamps into end of line bubbles or something less distracting ...

PR #1244, opened 2016-08-30
make the test pass no matter how the environment is configured

sandlerr:
This is causing red tests in a private repo because they are run with the buddy check environment variable turned on.

@danielbreves @zendesk/samson

### DIFF
--- a/test/models/stage_test.rb
+++ b/test/models/stage_test.rb
@@ -451,12 +451,16 @@ describe Stage do
       stage.production = true
     end
 
+    after do
+      BuddyCheck.unstub(:enabled?)
+    end
+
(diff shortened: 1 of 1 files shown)


danielbreves:
👍

test/models/stage_test.rb
```diff
@@ -451,12 +451,16 @@
       stage.production = true
     end
 
+    after do
+      BuddyCheck.unstub(:enabled?)
```
grosser: automatically unstubs after the test

grosser:
the private repo should run with the same base environment ... can enable in the before blocks as necessary ... this will just come up again when something is moved around ...

PR #1245, opened 2016-08-30
fix error saving users

sandlerr:
@zendesk/samson looks like we have some bad data in our database where not all UserProjectRoles have a project. Let's stop crashing.

https://zendesk.atlassian.net/browse/SAMSON-276


grosser:
has already been taken care of and won't happen again ... so don't need this :)

PR #1246, opened 2016-08-31
Show first lines of stacktrace in k8s deploy

ncalca:
Show the first lines of the logs instead of last.
If something fails, the stacktrace is printed and it's useful to see the top of the stack (i.e., exception name) instead of the bottom of the stack.

@grosser @jonmoter


grosser:
idk ... top might be just normal output and the rest is what is important :(

... maybe show everything or first 25 + ... + last 25 ?

ncalca:
@grosser changed to show first and last 25 lines. How does it look now?

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -123,8 +123,12 @@ def print_logs(client, pod)
             "No logs found"
           end
         end
-        # Don't display hundreds of log lines
-        logs.split("\n").last(50).each { |line| @output.puts "  #{line}" }
+        # Display the first and last lines of the log
+        first_half, second_half = logs.split("\n").in_groups(2, false)
+        logs_to_display = first_half.first(25)
```
grosser: extract 25 to a variable :)

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -123,8 +123,12 @@ def print_logs(client, pod)
...
+        logs_to_display << "  ..." if first_half.size > 25
+        logs_to_display.concat(second_half.last(25))
+        logs_to_display.each { |line| @output.puts "  #{line}" }
       end
```
grosser:
```Ruby
max = 50
lines = logs.split("\n")
lines = lines.first(max / 2) + lines.last(max / 2) if lines.size > max
```

ncalca: uh nicer :)

grosser:
👍

PR #1247, opened 2016-09-01
PAAS-349 add clair security scanner to kubernets pipeline

murat1985:
## Description
As a part of PaaS project we are adding to Samson invoking external scanner functionality. This method implies that CLAIR_EXEC_PATH and CLAIR_SCAN_ENABLED environment variables are supplied. Otherwise it will be disabled to allow using Samson without Clair scans.

## Risks
Low

## JIRA ticket number
https://zendesk.atlassian.net/browse/PAAS-349

## Change Request

https://zendesk.atlassian.net/wiki/pages/viewpage.action?pageId=XXXXXXXXXX

## CC list
@zendesk/operations @zendesk/chef @grosser @jonmoter @ashmckenzie

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -136,3 +136,14 @@ Kubernetes::ReleaseGroup
         |
         -> Kubernetes Pods (how ever many replicas specified)
 ```
+### CLair security scans
+
+To use hyperclair tool (command-line tool for Clair and Docker Registry) for security scans we should supply environment variables:
+
+```
+CLAIR_SCAN_ENABLED = [ true | false ]
+CLAIR_EXEC_SCRIPT = /filesystem/path
+```
+
+CLAIR_EXEC_SCRIPT  - path to shell script that will invoke security scanning. It should be able to return status code. 
+CLAIR_SCAN_EANBLE - should be set to `true` if we want to have clair scans enabled
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -141,9 +141,7 @@ Kubernetes::ReleaseGroup
 To use hyperclair tool (command-line tool for Clair and Docker Registry) for security scans we should supply environment variables:
 
 ```
-CLAIR_SCAN_ENABLED = [ true | false ]
-CLAIR_EXEC_SCRIPT = /filesystem/path
+CLAIR_EXEC_SCRIPT = /filesystem/path/to/hyperclair
 ```
 
 CLAIR_EXEC_SCRIPT  - path to shell script that will invoke security scanning. It should be able to return status code. 
-CLAIR_SCAN_EANBLE - should be set to `true` if we want to have clair scans enabled
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -136,12 +136,13 @@ Kubernetes::ReleaseGroup
         |
         -> Kubernetes Pods (how ever many replicas specified)
 ```
-### CLair security scans
 
-To use hyperclair tool (command-line tool for Clair and Docker Registry) for security scans we should supply environment variables:
+### Clair security scans
+
+To security scan docker images using hyperclair, add:
 
 ```
-CLAIR_EXEC_SCRIPT = /filesystem/path/to/hyperclair
-```
+HYPERCLAIR_PATH=/filesystem/path/to/hyperclair
+``` 
 
-CLAIR_EXEC_SCRIPT  - path to shell script that will invoke security scanning. It should be able to return status code. 
+Must exit with 0 for success and 1 for failure.
--- a/plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
@@ -35,13 +35,13 @@ module Kubernetes
       message = (success ? "completed successfully" : "failed or timed out")
       @output.puts "### Remote build job #{job_name} #{message}"
 
-      if ENV['CLAIR_SCAN_ENABLED'] == 'true'
+      if ENV['CLAIR_EXEC_SCRIPT']
         ### Running security scan
-        result_clair = scan_with_clair(project.permalink.tr('_', '-'), tag)
+        clair_result = scan_with_clair(project.permalink.tr('_', '-'), tag)
         ### Adding clair log to job log and success code
-        if result_clair
-          job_log = job_log << result_clair
-          success &&= false
+        if clair_result
+          job_log << clair_result
+          success = false
         end
       end
 
@@ -101,14 +101,9 @@ module Kubernetes
       k8s_job[:spec][:template][:spec][:containers][0].update(container_params)
     end
 
-    # Stub this, otherwise hard to test
-    def script_exit_status
-      $?
-    end
-
-    # Exctract script execution to a separate method
-    def exec_script(script, project, tag)
-      result = IO.popen([script, project, tag], err: [:child, :out])
+    # External script execution method for Clair security scanner
+    def exec_script(script, registry, project, tag)
+      result = IO.popen([script, registry, project, tag], err: [:child, :out])
       Process.wait(result.pid)
       result.read
     end
@@ -117,7 +112,8 @@ module Kubernetes
     # Hyperclair will pull the image from registry and run scan against Clair scanner
     def scan_with_clair(project, tag)
       clair_result = exec_script(ENV['CLAIR_EXEC_SCRIPT'], @registry[:serveraddress], project, tag)
-      if script_exit_status.success?
+      if $?.success?
+        @output.puts "### Clair scan passed successfully"
         return nil
       else
         @output.puts "### Clair scanner found vulnarabilities, stopping build"
--- a/plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
@@ -116,7 +116,7 @@ module Kubernetes
     # So far we don't have ruby API for clair scanner and using shell script to wrap hyperclair.
     # Hyperclair will pull the image from registry and run scan against Clair scanner
     def scan_with_clair(project, tag)
-      clair_result = exec_script(ENV['CLAIR_EXEC_SCRIPT'], project, tag)
+      clair_result = exec_script(ENV['CLAIR_EXEC_SCRIPT'], @registry[:serveraddress], project, tag)
       if script_exit_status.success?
         return nil
       else
--- a/plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
@@ -35,14 +35,8 @@ module Kubernetes
       message = (success ? "completed successfully" : "failed or timed out")
       @output.puts "### Remote build job #{job_name} #{message}"
 
-      if ENV['CLAIR_EXEC_SCRIPT']
-        ### Running security scan
-        clair_result = scan_with_clair(project.permalink.tr('_', '-'), tag)
-        ### Adding clair log to job log and success code
-        if clair_result
-          job_log << clair_result
-          success = false
-        end
+      if success && clair = ENV['HYPERCLAIR_PATH']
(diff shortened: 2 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -91,6 +94,23 @@ def fill_job_details(k8s_job, build, project, tag:, push: false, tag_as_latest:
       k8s_job[:spec][:template][:spec][:containers][0].update(container_params)
     end
 
+    # So far we don't have ruby API for clair scanner and using shell script to wrap hyperclair.
+    # Hyperclair will pull the image from registry and run scan against Clair scanner
+    def scan_with_clair(project, tag)
+      require 'mixlib/shellout'
+      cmd = Mixlib::ShellOut.new("/usr/local/bin/clair_scan.sh", project, tag, @registry[:serveraddress])
+      cmd.live_stream = STDOUT
+      if File.executable?("/usr/local/bin/clair_scan.sh")
```
grosser: duplicated executable path ... should come from ENV

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -91,6 +94,23 @@ def fill_job_details(k8s_job, build, project, tag:, push: false, tag_as_latest:
...
+        cmd.run_command
+        if cmd.error?
+          @output.puts "### Clair scanner found vulnarabilities, stopping build"
+          raise cmd.error!
```
grosser: this looks strange ... what does it do ?
... should it not show what went wrong in the `@output` ?

grosser:
- won't that take a long time ?
 - should we do this only when a new image was built ? (not reused)
 - should be behind a ENV flag and the flag documented
 - add tests
 - streaming to stdout seems odd ... stream to and StringIO or even @output instead ?
 - use `IO.popen` instead of mixlib ?

grosser:
if the scanner is activated in the ENV but not an executable it should blow up ...

murat1985:
@grosser thank you for comments.

yes we should scan only a new image and it can take a while, but only for the images that never been scanned. Scanning is layer-by-layer. 

Could you help me with pointing out a proper place in code where shall I embed clair method and how to scan only new images  ? :) thank you

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
murat1985: I thought that it will blow the build and show the content of STDERR/STDOUT

murat1985:
Agree to use popen instead

grosser:
nvm, I think you got the position right ... I was thinking it's in the general job-executor :D

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
grosser: it has no connection to `@output`, so I doubt that works ...

jonmoter:
Remember that Samson is an open-source project, so there are people who will be using this without having Clair installed.

We can make this functionality a plugin, or enable/disable it with an ENV variable.

jonmoter:
I think this should be triggered after an image is built, not before it's deployed.  Take a look at app/models/docker_builder_service.rb.

Let's also run the job asynchronously, either in a separate thread, or kick off a Kubernetes job to do the scanning. We don't have to block on the scan being complete before deploying the image.

grosser:
so deploying with a vulnerability is fine ?

jonmoter:
Deploying to master/staging/test environments is fine.  We'd prefer catching it before going to production.

murat1985:
@grosser I wrote unit tests but codeclimate has 4 issues. I don't have access there, could you please tell me how can I get access ? 

Thanks

grosser:
just "Identical code found in 1 other location (mass = 29)" ... so ignore ...

grosser:
can as IT to give you access, not sure how that works

murat1985:
@grosser add ignore in the list. Should be fine now :) good to merge ? :)

plugins/kubernetes/README.md
```diff
@@ -136,3 +136,14 @@ Kubernetes::ReleaseGroup
         |
         -> Kubernetes Pods (how ever many replicas specified)
 ```
+### CLair security scans
+
+To use hyperclair tool (command-line tool for Clair and Docker Registry) for security scans we should supply environment variables:
+
+```
+CLAIR_SCAN_ENABLED = [ true | false ]
```
grosser: how about when the path is set, then it's enabled ?

this could be good enough:

```
To use hyperclair tool (command-line tool for Clair and Docker Registry) for security scans, configure environment variables 
`CLAIR_SCRIPT = /filesystem/path/to/hyperclair`
```

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -35,6 +35,16 @@ def execute!(build, project, tag:, push: false, registry:, tag_as_latest: false)
       message = (success ? "completed successfully" : "failed or timed out")
       @output.puts "### Remote build job #{job_name} #{message}"
 
+      if ENV['CLAIR_SCAN_ENABLED'] == 'true'
+        ### Running security scan
+        result_clair = scan_with_clair(project.permalink.tr('_', '-'), tag)
```
grosser: result_clair -> clair_result ?

... log something if it fails ?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -35,6 +35,16 @@ def execute!(build, project, tag:, push: false, registry:, tag_as_latest: false)
...
+        ### Adding clair log to job log and success code
+        if result_clair
+          job_log = job_log << result_clair
```
grosser: no need to re-assign job_log, `<<` is a modifying operation ?

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -35,6 +35,16 @@ def execute!(build, project, tag:, push: false, registry:, tag_as_latest: false)
...
+          success &&= false
```
grosser: simply set to false ? ... `&&= false` is the same as `= false` ...

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -91,6 +101,31 @@ def fill_job_details(k8s_job, build, project, tag:, push: false, tag_as_latest:
       k8s_job[:spec][:template][:spec][:containers][0].update(container_params)
     end
 
+    # Stub this, otherwise hard to test
```
grosser: set the CLAIR_EXEC_SCRIPT to a tempfile you create, then put either `ls` or bogus in there to test it

grosser:
```Ruby
Tempfile.open('clair'){ |f|  f.write 'ls'; File.chmod 0755, f.path ; with_env(CLAIR_EXEC_SCRIPT : f.path) { ... run test ...} }
```

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -91,6 +101,31 @@ def fill_job_details(k8s_job, build, project, tag:, push: false, tag_as_latest:
...
+    def script_exit_status
+      $?
+    end
+
+    # Exctract script execution to a separate method
```
grosser: this comment is not very helpful

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
```diff
@@ -91,6 +101,31 @@ def fill_job_details(k8s_job, build, project, tag:, push: false, tag_as_latest:
...
+      result = IO.popen([script, project, tag], err: [:child, :out])
+      Process.wait(result.pid)
+      result.read
+    end
+
+    # So far we don't have ruby API for clair scanner and using shell script to wrap hyperclair.
+    # Hyperclair will pull the image from registry and run scan against Clair scanner
+    def scan_with_clair(project, tag)
+      clair_result = exec_script(ENV['CLAIR_EXEC_SCRIPT'], project, tag)
+      if script_exit_status.success?
```
grosser: print the failure here ?

grosser: ... or some kind of "Clair scan passed"

plugins/kubernetes/test/models/kubernetes/build_job_executor_test.rb
```diff
@@ -128,6 +128,44 @@ def execute!(push: false, registry_info: nil)
           assert_empty job_log
         end
 
+        it 'tests clair exit code 0 - success' do
```
grosser: being more descriptive makes diving into this later easy:

`it "prints results when clair scan passes"`
`it "shows errors when clair scan fails"`

plugins/kubernetes/test/models/kubernetes/build_job_executor_test.rb
```diff
@@ -128,6 +128,44 @@ def execute!(push: false, registry_info: nil)
...
+          executor.stubs(:clair_result).returns "test"
+          Process.stubs(:wait).returns true
+          IO.stubs(:popen).returns OpenStruct.new(pid: 123, read: "test", close: "test")
+          with_env(
+            "CLAIR_EXEC_SCRIPT"  => "/invalid/path",
+            "CLAIR_SCAN_ENABLED" => "true"
+          ) do
+            job_api_obj.stubs(:failure?).returns false
+            job_api_obj.stubs(:complete?).returns true
+            success, = execute!(registry_info: registry_info)
```
grosser: `success,` -> `success` ?

plugins/kubernetes/test/models/kubernetes/build_job_executor_test.rb
```diff
@@ -128,6 +128,44 @@ def execute!(push: false, registry_info: nil)
           assert_empty job_log
         end
 
+        it 'tests clair exit code 0 - success' do
+          executor.unstub(:scan_with_clair)
```
grosser: no need to unstub this

plugins/kubernetes/app/models/kubernetes/build_job_executor.rb
murat1985: will do that, it sounds much better for testing purpose

murat1985:
@grosser tried to address your comments, could you please take a look once again, sorry for the mess

jonmoter:
So the way it's currently implemented, the clair scan is only run if you do the `docker build` through a Kubernetes job, but the scan is run locally on the samson host, is that correct?

That seems weird to me.  It seems that whether or not you build the docker image with a Kubernetes job, you should have the option of running it through Clair.

I also don't like that the `DockerBuildExecutor` returns false if the scan fails.  That's not accurate. The docker build completed, but a separate step (the scanning) failed. If we're going to integrate this into Samson in a way that causes builds to fail, I want to do it right, and track the information on Clair builds in a DB table, rather than it just being some text in the build output.

I'd really like to do this asynchronously, rather than the "docker build" step blocking on the security scanning.

murat1985:
@jonmoter that is good idea to extract it on asynchronous job and don't fail a build. But it will require more time to invest. We can just produce warning message with change `success` to `false`, that will partially address your concern. Regarding async job - I can take a look further in it but not sure that would be in time by 13th of September. 

1. Scans are initiated by the script (hyperclair) locally on samson but Clair scanner is a remote instance as well as the registry. 
2. Clair scan by default switched off (if environment variable `CLAIR_EXEC_SCRIPT` is not set).

Shall we proceed with current course of action on temporal basis or come up with more long-term and stable solution ? Pascal told me that we need to fulfil the SOC2 requirement formally without extensive time investment. 

Thanks

cc: @epssy

grosser:
taking this over to clean it up a bit

grosser:
updated:
 - runs independent of job
 - times out after 1 hour
 - amends job log with output
 - only runs when build succeeded
 - reports timeouts/errors
 - kills child on timeout/error/interrupts

grosser:
... do not pass samsons env to the script ... it does not need secrets ?

jonmoter:
👍 on 💚

PR #1249, opened 2016-09-01
now injects the cluster name as an ENV var into all pods

irwaters:
adds the name of the cluster as an ENV var to all pods.

https://zendesk.atlassian.net/browse/PAAS-379

Risk is low.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource_template.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource_template.rb
@@ -178,8 +178,8 @@ module Kubernetes
       end
 
       # name of the cluster
-      [:CLUSTER_NAME].each do |k|
-        env << {name: k, value: DeployGroup.find_by_permalink(metadata[:deploy_group]).kubernetes_cluster.name.to_s}
+      [:KUBE_CLUSTER_NAME].each do |k|
+        env << {name: k, value: DeployGroup.find(metadata[:deploy_group_id]).kubernetes_cluster.name.to_s}
       end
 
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/resource_template.rb
```diff
@@ -177,6 +177,11 @@ def set_env
         env << {name: k, value: template[:spec][:template][:metadata][:labels][k.downcase].to_s}
       end
 
+      # name of the cluster
+      [:CLUSTER_NAME].each do |k|
```
grosser: why the each ?
there are lots of clusters ... call it KUBERNETES_CLUSTER_NAME ?

irwaters: yeah, was already changing that.

plugins/kubernetes/app/models/kubernetes/resource_template.rb
```diff
@@ -177,6 +177,10 @@ def set_env
         env << {name: k, value: template[:spec][:template][:metadata][:labels][k.downcase].to_s}
       end
 
+      # name of the cluster
+      kube_cluster_name = DeployGroup.find(metadata[:deploy_group_id]).kubernetes_cluster.name.to_s
+      env << {name: :KUBE_CLUSTER_NAME, value: kube_cluster_name}
```
grosser: spell out KUBERNETES so it matches the other native KUBERNETES vars ?

grosser:
👍

PR #1248, opened 2016-09-01
prevent 3-state booleans

grosser:
@dragonfax 
/cc @zendesk/samson 

previously caught:

```
Boolean columns missing a default or null: false
 t.boolean  "update_github_pull_requests"
 t.boolean  "comment_on_zendesk_tickets"
 t.boolean  "use_github_deployment_api"
```

PR #1250, opened 2016-09-02
fix migration not working when columns have nil values

grosser:
was hoping it would set the to false automatically ... did not ...

fix for https://github.com/zendesk/samson/pull/1248

@dragonfax

### DIFF
--- a/db/migrate/20160901175936_prevent_3_state_booleans.rb
+++ b/db/migrate/20160901175936_prevent_3_state_booleans.rb
@@ -1,9 +1,19 @@
 # frozen_string_literal: true
 class Prevent3StateBooleans < ActiveRecord::Migration
-  def change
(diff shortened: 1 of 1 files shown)


db/migrate/20160901175936_prevent_3_state_booleans.rb
```diff
@@ -1,9 +1,19 @@
 # frozen_string_literal: true
 class Prevent3StateBooleans < ActiveRecord::Migration
-  def change
-    [:update_github_pull_requests, :comment_on_zendesk_tickets, :use_github_deployment_api].each do |column|
+  COLUMNS = [:update_github_pull_requests, :comment_on_zendesk_tickets, :use_github_deployment_api]
+
+  def up
+    COLUMNS.each do |column|
+      Stage.where(column => nil).update_all(column => false)
```
dragonfax: Whats with the old style fat-arrow. `column: nil` works fine.

dragonfax:
rubocop and 👍

PR #1252, opened 2016-09-03
allow deploying roles with 0 replicas to disable them

grosser:
@jonmoter @irwaters 

https://samsontest.zende.sk/projects/zendesk_voyager/deploys/3095


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -165,19 +164,19 @@ def release_statuses(pods, release_doc)
...
-        [[false, "Missing"]]
-      else
-        pods.map do |pod|
-          if pod.live?
-            if pod.restarted?
-              [false, RESTARTED]
-            else
-              [true, "Live"]
-            end
+      statuses = Array.new(release_doc.desired_pod_count).each_with_index.map do |_, i|
```
grosser: marking pods missing here now then waiting on the outside ... should be more consistent ...

jonmoter:
👍

PR #1253, opened 2016-09-03
fix migration again ... because deleted stages have nil values

grosser:
@dragonfax .... it never ends ...


dragonfax:
Okay that seems like it shouldn't be necessary.
 👍

PR #1255, opened 2016-09-06
dont crash the entire page if the jenkins plugin experiences an error

sandlerr:
Should fix https://zendesk.airbrake.io/projects/95346/groups/1590266375715628337/notices/1768650795732112721

cc @zendesk/samson @clouseauu @rdhanoa @upcomingnewton


grosser:
test ? :)

... otherwise :+1:

PR #1259, opened 2016-09-06
show commit so it is easy to re-deploy an old working master deploy

grosser:
screen shot 2016-09-06 at 4 51 03 pm

@zendesk/samson


sandlerr:
awesome, was just looking for this to use the lotus `?sha=afafaf` query param :+1:

sandlerr:
Although we show quite a bit of metadata in the general commandline output now. Maybe we should pull it out to a table somewhere.

grosser:
tabelizing might be hard since we spit out random bits here and there :)

PR #1262, opened 2016-09-08
PAAS-383 show secrets to any user that is a deployer anywhere

grosser:
### Risks
- Low: secret access breaking
- Low: users that are deployers on 1 project can see visible secrets for all projects

@zendesk/samson


jonmoter:
Good first step.  Makes it more functional.  👍

PR #1263, opened 2016-09-08
cleanup test logic

grosser:
makes the default case success and makes adding new tests less repetitive / makes us only change 1 thing per test


jonmoter:
👍 on 💚

PR #1264, opened 2016-09-09
add a docker-specific release branch to projects

sandlerr:
We need to build docker images automatically for a project, but we don't want releases / git tags created for it.


sandlerr:
@grosser @zendesk/samson can I get a :+1: on this?

grosser:
feels like a redundant feature for strange edge-case ...

grosser:
deploy_with_docker can go away then ?

sandlerr:
relates to https://zendesk.slack.com/archives/deployment/p1472606763000068 @jonmoter

app/models/project.rb
```diff
@@ -71,6 +71,18 @@ def create_releases_for_branch?(branch)
...
 
+  # Whether to create a new docker image when the branch is updated.
+  #
+  # branch - The String name of the branch in question.
+  #
+  # Returns true if new releases should be created, false otherwise.
+  # If the docker_release_branch of the project wasn't set, fallback
+  # to the release_branch for backward compatibility.
+  def build_docker_image_for_branch?(branch)
+    deploy_with_docker? && auto_release_docker_image? &&
```
grosser: `building?` checks if `auto_release` is set ... can we make the naming more consistent ?

app/views/projects/_form.html.erb
```diff
@@ -45,12 +45,23 @@
...
+      </div>
+    </div>
+
+    <div class="form-group">
+      <%= form.label :docker_release_branch, class: "col-lg-2 control-label" %>
+      <div class="col-lg-4">
+        <%= form.text_field :docker_release_branch, class: "form-control" %>
         <div id="docker_release_push">
           <%= form.check_box :auto_release_docker_image  %>
           <%= form.label :auto_release_docker_image, 'Auto push docker images' %>
```
grosser: is there a reason to build images without pushing ?

sandlerr:
> deploy_with_docker can go away then ?

We can do this if the migration also copies the `release_branch` into `docker_release_branch` for all projects that have the `deploy_with_docker` set to true before dropping the column. Do you prefer that

app/views/projects/_form.html.erb
sandlerr: No idea, maybe Jon can answer

grosser:
that sounds good ... at least 1 less condition ... how about simply making deploy_with_docker do exactly what this PR does ... but default to master branch ?

sandlerr:
Actually, from what I can tell `deploy_with_docker` and `auto_release_docker_image` are two guards around the exact same thing, we shouldn't need both of them.

sandlerr:
ok, updated. Is this migration reasonable even though it uses the model?

grosser:
redeclare the model in the migration to avoid any backwards compatibility issues and fun things like soft_deletion

app/controllers/projects_controller.rb
```diff
@@ -82,8 +82,7 @@ def project_params
         :owner,
         :permalink,
         :release_branch,
-        :deploy_with_docker,
-        :auto_release_docker_image,
```
grosser: ❤️

app/models/project.rb
```diff
@@ -71,6 +71,19 @@ def create_releases_for_branch?(branch)
     release_branch == branch
   end
 
+  # Whether to create a new docker image when the branch is updated.
+  #
+  # branch - The String name of the branch in question.
+  #
+  # Returns true if new releases should be created, false otherwise.
```
grosser: kind of excessive docs ... faster to read the method :D

app/models/project.rb
```diff
@@ -71,6 +71,19 @@ def create_releases_for_branch?(branch)
...
+  def build_docker_image_for_branch?(branch)
+    docker_release_branch == branch
+  end
+
+  def deploy_with_docker?
+    docker_release_branch.present?
```
grosser: having the branch does not mean it gets deployed with docker ... but that can be another PR ...

lib/samson/logging.rb
```diff
@@ -3,7 +3,7 @@
 
 if Samson::EnvCheck.set?("RAILS_LOG_TO_STDOUT")
   # heroku and docker: dump everything to stdout
-  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
+  config.logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT))
```
grosser: why ?

test/models/project_test.rb
```diff
@@ -406,6 +406,22 @@ def clone_repository(project)
     end
   end
 
+  describe "#build_docker_image_for_branch?" do
+    context 'when the docker release branch is set' do
```
grosser: `it 'returns false when branch is not set'` ?

lib/samson/logging.rb
sandlerr: sorry, I wrote it in my commit message on https://github.com/zendesk/samson/pull/1264/commits/dab7c9fd5d1cc0efda5b7450a46e582f6a2ec72c

grosser:
👍 looks about right ... nice to see these weird flags go, always confused me :D

lib/samson/logging.rb
grosser: good catch! :D

app/models/project.rb
sandlerr: yeah just copied from `create_releases_for_branch?` method

sandlerr:
rebased, ready to merge now

test/models/project_test.rb
```diff
@@ -406,6 +406,33 @@ def clone_repository(project)
...
+      before do
+        project.docker_release_branch = nil
+        project.release_branch = 'master-of-puppets'
+      end
+      it 'returns false' do
+        refute project.build_docker_image_for_branch?(nil)
+        refute project.build_docker_image_for_branch?('master-of-puppets')
+      end
+    end
+  end
```
grosser: This might be simpler:

```Ruby
describe "#build_docker_image_for_branch?" do
    before do
      project.docker_release_branch = 'master'
    end

    it 'returns true for the right branch' do
      assert project.build_docker_image_for_branch?('master')
    end

    it 'returns false for the wrong branch' do
      refute project.build_docker_image_for_branch?('other')
    end

    it 'returns false when not set' do
      project.docker_release_branch = nil
      refute project.build_docker_image_for_branch?(nil)
      refute project.build_docker_image_for_branch?('other')
    end
  end
```

grosser:
👍

mikehuston:
@sandlerr, @grosser, we used to use the docker binary builder to create JAR files in Samson but were not necessarily interested in releasing a docker image. Is this still possible with these changes?

grosser:
how is binary builder triggered ?
... might need an extra `binary build` flag ?

mikehuston:
It looks like it is triggered here:
https://github.com/zendesk/samson/blob/master/plugins/docker_binary_builder/lib/samson_docker_binary_builder/samson_plugin.rb#L9

and uses `deploy_with_docker?` - whose meaning changed in this PR from checking the (now removed)`deploy_with_docker` column to checking the `docker_release_branch` column
https://github.com/zendesk/samson/blob/master/plugins/docker_binary_builder/app/models/binary_builder.rb#L21

sandlerr:
so you need to run the binary builder on a webhook and manually to run your `pre_build_script.sh`, but not to build or push a docker image? This change involved a migration so it's not so easy to roll it back but I'll restore what you were using once I understand

grosser:
it should check the docker_binary_plugin_enabled ?

PR #1265, opened 2016-09-09
show reason why the pod is pending

grosser:
was wondering if something was broken because a build was in pending and then had no logs, this should clarify the situation a bit 

@jonmoter 

screen shot 2016-09-08 at 7 47 26 pm


irwaters:
I like 👍

PR #1267, opened 2016-09-13
Avoid DataDog spam

asellitt:
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low (imho)

Sending events to DataDog which have an `@` followed by something vaguely email shaped will cause DataDog to send an email to that email account.

On the surface, this isnt so bad, but when you use application specific email addresses you are left in a situation where you are required to create an entirely new DataDog account just to unsubscribe that email address from said spam. 

This, apparently, is a feature.

This change should still allow notifications to be sent, but more importantly, only if it matches a known DataDog account, allowing the individual to manage their spam preferences accordingly.

### DIFF
--- a/plugins/datadog/app/models/datadog_notification.rb
+++ b/plugins/datadog/app/models/datadog_notification.rb
@@ -37,7 +37,7 @@ class DatadogNotification
   private
 
   def body
-    "@#{@deploy.user.email} deployed #{@deploy.short_reference} to #{@stage.name}"
+    "@#{@deploy.user.name} deployed #{@deploy.short_reference} to #{@stage.name}"
   end
 
   def api_key
--- a/plugins/datadog/app/models/datadog_notification.rb
+++ b/plugins/datadog/app/models/datadog_notification.rb
@@ -37,7 +37,7 @@ class DatadogNotification
   private
 
   def body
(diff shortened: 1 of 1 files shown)


grosser:
why not remove the leading @ ?

asellitt:
This is a pretty good question. There are two reasons here.

1. This is a personal preference, but I prefer to be referred to by name (rather than email). Its a bit more personal.
1. One of my colleagues suggested this because it should allow users in datadog (with a matching name) to be able to manage their own preferences.

I dont have a strong preference either way.

grosser:
Using names will make notifications stop working since names are not unique ... also makes it harder to follow up since you might not know how to reach the deployer ... so I'd prefer removing the leading @ since it does not make much sense to notify the deployer about the deploy ... and solves the issue :D

asellitt:
Hey folks, do you have any feedback or anything that i need to do to get this PR ready to go? I suppose I could leave my samson email as `support@datadoghq.com` ( :trollface: ) but i figure this is a much neater solution.

grosser:
thx! :)

asellitt:
Thank you! :D

PR #1268, opened 2016-09-13
github action determines webhook validity

anthonywoo:
We were looking at the wrong `action` key...

The correct one is within `github` to match the payload documented here: https://developer.github.com/v3/activity/events/types/#issuecommentevent

/cc @zendesk/samson @mwerner @alanhogan 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low. Webhook deploys from issue comments may still not work


grosser:
👍

grosser:
maybe try in staging before merging :D

PR #1270, opened 2016-09-13
kill job execution threads that are hanging

grosser:
/cc @zendesk/samson

ran into this when deploying kubernetes and an api call would not complete ... I think we should kill threads if they misbehave instead of letting them hang forever

was unable to test this because of some internal minitest bug that makes before blocks run twice when the thread was killed ...

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -48,7 +48,7 @@ class JobExecution
   end
 
   def wait!
-    @thread.try(:join)
+    @thread.join
   end
 
   def pid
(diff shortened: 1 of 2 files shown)


app/models/job_execution.rb
```diff
@@ -65,14 +65,11 @@ def close
   def stop!
     if @thread
       @executor.stop! 'INT'
-
-      stop_timeout.times do
-        return if @thread.join(1)
-      end
+      return if @thread.join(stop_timeout)
```
grosser: this should do the same and is simpler ...

app/models/job_execution.rb
```diff
@@ -48,7 +48,7 @@ def start!
   end
 
   def wait!
```
grosser: only called in test ... and should not get called before the job is started

irwaters:
lgtm
👍

grosser:
now properly cancels, keep output and kills threads ... I think our cancel issues are over :D

PR #1272, opened 2016-09-13
remove Project#deploy_with_docker? and fix docker binary builder

jonmoter:
The `deploy_with_docker` column on the Project model was intended to say whether docker was "enabled" for a given project. That would cause the "Builds" tab to appear, and enable the user of docker binary builder.  It was intended to be a distinct flag from whether to automatically build a docker image on merge to a branch.

When the column was removed in #1264, that screwed up some of that logic.

This PR changes the functionality to assume that Docker is now just a built-in feature of Samson, rather than enabled on a per-project basis.  This means that if the docker_binary_builder plugin is enabled, and checked for a given stage, it will automatically run as intended.

/cc @zendesk/samson, @sandlerr, @grosser 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low (this is already broken in production)

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -77,11 +77,7 @@ class Project < ActiveRecord::Base
   #
   # Returns true if new releases should be created, false otherwise.
   def build_docker_image_for_branch?(branch)
-    deploy_with_docker? && docker_release_branch == branch
-  end
-
-  def deploy_with_docker?
-    docker_release_branch.present?
+    docker_release_branch == branch
   end
 
   # The user/repo part of the repository URL.
--- a/app/views/shared/_project_tabs.html.erb
+++ b/app/views/shared/_project_tabs.html.erb
@@ -5,7 +5,7 @@
   <li class="<%= 'active' if active == 'releases' %>">
     <%= link_to "Releases", project_releases_path(project) %>
   </li>
-  <% if project.deploy_with_docker? %>
+  <% if ENV['DOCKER_FEATURE'] %>
     <li class="<%= 'active' if active == 'builds' %>">
       <%= link_to "Builds", project_builds_path(project) %>
     </li>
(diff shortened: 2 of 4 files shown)


jvorenkamp:
👍

irwaters:
👍

app/views/shared/_project_tabs.html.erb
```diff
@@ -5,11 +5,9 @@
   <li class="<%= 'active' if active == 'releases' %>">
     <%= link_to "Releases", project_releases_path(project) %>
   </li>
-  <% if project.deploy_with_docker? %>
-    <li class="<%= 'active' if active == 'builds' %>">
-      <%= link_to "Builds", project_builds_path(project) %>
-    </li>
-  <% end %>
+  <li class="<%= 'active' if active == 'builds' %>">
```
grosser: ... gate behind if `ENV['DOCKER_REGISTRY']` ?

app/models/project.rb
```diff
@@ -77,11 +77,7 @@ def create_releases_for_branch?(branch)
   #
   # Returns true if new releases should be created, false otherwise.
   def build_docker_image_for_branch?(branch)
-    deploy_with_docker? && docker_release_branch == branch
-  end
-
-  def deploy_with_docker?
-    docker_release_branch.present?
+    docker_release_branch.present? && docker_release_branch == branch
```
grosser: branch will never be nil/blank ... so do we need the presence check ?

grosser:
👍

app/views/shared/_project_tabs.html.erb
jonmoter: Good call.  Actually, I'll use `ENV['DOCKER_FEATURE']`

sandlerr:
:+1: I wasn't quite sure why two flags existed (this and auto_release_docker_image), should have found out before making changes.

PR #1273, opened 2016-09-13
Add Rails version to migration

tmcinerney:
Attempting to run migrations complained that some were missing the Rails version. This fixes that issue.

/cc @zendesk/samson

### Tasks
 - [x] :+1: from team

### Risks
- Level: Low


grosser:
👍

PR #1275, opened 2016-09-14
fix failing docker builds due to invalid character encoding

jonmoter:
* Create a concern for handling output of docker commands
* TODO: add some encoding magic to catch when there's an encoding mismatch

/cc @grosser

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-280
 - Airbrake: https://zendesk.airbrake.io/projects/95346/groups/1637723283388835045

### Risks
- Level: Low


jonmoter:
As far as I can tell, the strings we're pushing into the `OutputAggregator` have inconsistent encodings, so when `join` gets called in that class, it blows up.  so I think the issue is that we're generating UTF-8 strings from Samson, but we're getting ASCII-8BIT strings back from docker.

When we get strings back from docker, we should encode them into UTF-8, so we have consistent string encodings going into the aggregator.

I created the `HandlesDockerOutput` module so that we have a single function for handling the output we get back from Docker.  And then we won't have as much serialized JSON as output, like this:

```
{"stream":"Step 1 : FROM java:openjdk-8-jdk\n"}
{"stream":" ---\u003e 6c9a006fd38a\n"}
```

grosser:
sending something through JSON.parse already solves the encoding issues, so should be good to go ... even if it does not fix it, it is better then before ... so let's merge and deploy ...

grosser:
... ah need to fix json parse errors still ...

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -62,7 +62,7 @@ def build_file_exist?
 
   def start_build_script
     @output_stream.puts 'Now starting Build container...'
-    @container.tap(&:start).attach { |_stream, chunk| @output_stream.write chunk }
+    @container.tap(&:start).attach { |_stream, chunk| handle_output_chunk(chunk, @output_stream) }
```
jonmoter: Should be `@ouput_stream.write_docker_chunk(chunk)`, right?

grosser: fixed

grosser:
all good now ... try using the new review button :D

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -146,9 +148,13 @@ def image_name
...
     @output_stream.puts 'Now building the build container...'
-    Docker::Image.build_from_dir(
-      @dir, 'dockerfile' => DOCKER_BUILD_FILE, 't' => image_name
-    ) { |chunk| @output_stream.write chunk }
+    build_options = {
+      'dockerfile' => DOCKER_BUILD_FILE,
+      't' => image_name
+    }
+    Docker::Image.build_from_dir(@dir, build_options) do |chunk|
+      handle_output_chunk(chunk, @output_stream)
```
jonmoter: This method got removed, right?

grosser: arg ... fixed ... atm getting segfault when running all tests ... so relying on travis :D

jonmoter:
Looks like the build is failing due to test coverage.

```
Lines missing coverage:
plugins/docker_binary_builder/app/models/binary_builder.rb:69
plugins/docker_binary_builder/app/models/binary_builder.rb:70
plugins/docker_binary_builder/app/models/binary_builder.rb:85
plugins/docker_binary_builder/app/models/binary_builder.rb:87
plugins/docker_binary_builder/app/models/binary_builder.rb:88
plugins/docker_binary_builder/app/models/binary_builder.rb:89
plugins/docker_binary_builder/app/models/binary_builder.rb:90
plugins/docker_binary_builder/app/models/binary_builder.rb:91
plugins/docker_binary_builder/app/models/binary_builder.rb:93
plugins/docker_binary_builder/app/models/binary_builder.rb:94
plugins/docker_binary_builder/app/models/binary_builder.rb:96
plugins/docker_binary_builder/app/models/binary_builder.rb:97
plugins/docker_binary_builder/app/models/binary_builder.rb:98
plugins/docker_binary_builder/app/models/binary_builder.rb:99
plugins/docker_binary_builder/app/models/binary_builder.rb:156
plugins/docker_binary_builder/app/models/binary_builder.rb:168
rake aborted!
```

jonmoter:
Aside from the test coverage, looks good to me. 👍

PR #1278, opened 2016-09-14
show an error when kubernetes is unable to terminate pods instead of …

grosser:
@jonmoter @irwaters 

... happened when deploying translation daemon and it was confusing as to why it could not finish


grosser:
screen shot 2016-09-14 at 4 20 33 pm

jonmoter:
I like descriptive error messages!  👍

PR #1279, opened 2016-09-15
Basic client for ledger (as a plugin)

irwaters:
Provides basic ledger integration for samson.  will post details of a deployment when it starts, and another when that deployment finishes.

https://zendesk.atlassian.net/browse/SAMSON-268

little danger.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -55,6 +55,12 @@ PATH
     samson_pipelines (0.0.0)
 
 PATH
+  remote: plugins/samson_ledger
+  specs:
+    samson_ledger (0.0.1)
+      faraday
+
+PATH
   remote: plugins/slack_app
   specs:
     samson_slack_app (0.0.0)
@@ -570,6 +576,7 @@ DEPENDENCIES
   samson_flowdock!
   samson_jenkins!
   samson_kubernetes!
+  samson_ledger!
   samson_new_relic!
   samson_pipelines!
   samson_slack_app!
--- a/plugins/samson_ledger/README.md
+++ b/plugins/samson_ledger/README.md
@@ -1,0 +1,7 @@
+# Ledger Plugin
+
+Plugin that allows samson to post events to ledger (an internal Zendesk system)
+
+Requires 2 ENV vars to be set in order to work
+LEDGER_BASE_URL="https://ledger.mydomain.com" # base url of the ledger api host (http or https)
+LEDGER_TOKEN="thisismytoken" # api access token for that system                                                                 
--- a/plugins/samson_ledger/lib/samson_ledger/client.rb
+++ b/plugins/samson_ledger/lib/samson_ledger/client.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+#
+# basic ledger event client
+#
+require 'faraday'
+require 'openssl'
+
+module SamsonLedger
+  class Client
+    LEDGER_PATH = "/api/v1/events"
+    class << self
+      def plugin_enabled?
+        if ENV["LEDGER_TOKEN"].nil? || ENV["LEDGER_BASE_URL"].nil?
+          false
+        else
+          true
+        end
+      end
+
+      def post_deployment(deploy)
+        return false unless plugin_enabled?
+        post_event(deploy)
+      end
+
+      private
+
+      def post_event(deploy)
+        results = post(build_event(deploy))
+        if results.status.to_i != 200
+          Rails.logger.error("Ledger Client got a #{results.status} from #{ENV.fetch("LEDGER_BASE_URL")}")
+        end
+        results
+      end
+
+      def build_event(deploy)
+        event = {
+          id:           deploy.id,
+          name:         deploy.project.name,
+          actor:        deploy.user.name,
+          status:       deploy.status,
+          started_at:   deploy.created_at.iso8601,
+          summary:      deploy.summary,
+          environment:  deploy.stage.deploy_groups.map(&:environment).uniq.map(&:permalink).map(&:downcase).join(","),
+          url:          deploy.url,
+          pods:         deploy.stage.deploy_groups.map(&:env_value)
+        }
+        { "events": [event] }
+      end
+
+      def post(data)
+        connection = Faraday.new(url: ENV.fetch("LEDGER_BASE_URL") + LEDGER_PATH)
+        connection.post do |request|
+          request.headers['Content-Type'] = "application/json"
+          request.headers['Authorization'] = "Token token=#{ENV.fetch("LEDGER_TOKEN")}"
+          request.body = data.to_json
+        end
+      end
+    end
+  end
+end
--- a/plugins/samson_ledger/lib/samson_ledger/samson_plugin.rb
+++ b/plugins/samson_ledger/lib/samson_ledger/samson_plugin.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+module SamsonLedger
+  class Engine < Rails::Engine
+  end
+end
+
+Samson::Hooks.callback :before_deploy do |deploy|
+  SamsonLedger::Client.post_deployment(deploy)
+end
+
+Samson::Hooks.callback :after_deploy do |deploy|
+  SamsonLedger::Client.post_deployment(deploy)
+end
--- a/plugins/samson_ledger/samson_ledger.gemspec
+++ b/plugins/samson_ledger/samson_ledger.gemspec
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+Gem::Specification.new 'samson_ledger', '0.0.1' do |s|
+  s.summary = 'client to allow posting events to ledger (internal zendesk system)'
+  s.description = s.summary
+  s.authors = ['ian Waters']
+  s.email = 'iwaters@zendesk.com'
+  s.add_runtime_dependency 'faraday'
+end
--- a/plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
+++ b/plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+require_relative "../../test_helper"
+
+SingleCov.covered!
+
+describe 'SamsonLedger::Client' do
+  let(:deploy) { Deploy.first }
+
+  before do
+    @event_sent = stub_request(:post, "https://foo.bar/api/v1/events")
+  end
+
+  with_env(LEDGER_BASE_URL: 'https://foo.bar', LEDGER_TOKEN: "sometoken")
+
+  describe ".plugin_enabled?" do
+    it "is enabled" do
+      assert SamsonLedger::Client.plugin_enabled?
+    end
+
+    it "is not enabled without token" do
+      ENV.delete("LEDGER_TOKEN")
+      refute SamsonLedger::Client.plugin_enabled?
+    end
+
+    it "is not enabled without base_url" do
+      ENV.delete("LEDGER_BASE_URL")
+      refute SamsonLedger::Client.plugin_enabled?
+    end
+  end
+
+  describe ".post_deployment" do
+    it "posts an event with a valid client" do
+      SamsonLedger::Client.post_deployment(deploy)
+      assert_requested(@event_sent)
+    end
+  end
+
+  describe ".post_deployment" do
+    before do
+      stub_request(:post, "https://foo.bar/api/v1/events").
+        to_return(status: 401)
+    end
+
+    it "rejects our token" do
+      results = SamsonLedger::Client.post_deployment(deploy)
+      results.status.to_i.must_equal(401)
+    end
+  end
+end
--- a/plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
+++ b/plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+require_relative "../../test_helper"
+
+SingleCov.covered!
+
+describe "the samson_ledger plugin callback" do
+  let(:deploy) { Deploy.first }
+
+  before do
+    stub_request(:post, "https://foo.bar/api/v1/events")
+  end
+
+  with_env(LEDGER_TOKEN: "sometoken", LEDGER_BASE_URL: "https://foo.bar")
+
+  let(:deploy) { Deploy.first }
+
+  it "successfully fires the before_deploy" do
+    SamsonLedger::Client.expects(:post_event)
+    Samson::Hooks.fire(:before_deploy, deploy, nil)
+  end
+
+  it "successfully fires the after_deploy" do
+    SamsonLedger::Client.expects(:post_event)
+    Samson::Hooks.fire(:after_deploy, deploy, nil)
+  end
+
+  it "fails to fires the after_deploy" do
+    ENV.delete("LEDGER_TOKEN")
+    SamsonLedger::Client.expects(:post_event).never
+    Samson::Hooks.fire(:after_deploy, deploy, nil)
+  end
+end
--- a/plugins/samson_ledger/test/test_helper.rb
+++ b/plugins/samson_ledger/test/test_helper.rb
@@ -1,0 +1,2 @@
+# frozen_string_literal: true
+require_relative '../../../test/test_helper'


.env.example
```diff
@@ -114,3 +114,7 @@ GITHUB_TOKEN=
 # JENKINS_USERNAME= # user id
 # JENKINS_API_KEY= # API Token from user / Configure page
 # JENKINS_NOTIFY_COMMITTERS=1 # optional: send committers to jenkins as 'emails' params (filtered by GOOGLE_DOMAIN)
+
+## Ledger configs
+#LEDGER_BASE_URL="https://ledger.mydomain.com" # base url of the ledger api host (http or https)
```
grosser: maybe mention something like "Zendesk internal monitoring tool"

Gemfile.lock
```diff
@@ -45,6 +45,11 @@ PATH
     samson_kubernetes (0.0.1)
 
 PATH
+  remote: plugins/ledger
+  specs:
+    ledger (0.0.1)
```
grosser: gem itself should be called samson_ledger to not collide with https://rubygems.org/gems/ledger

app/controllers/deploys_controller.rb
```diff
@@ -99,6 +99,7 @@ def new
   def create
     deploy_service = DeployService.new(current_user)
     @deploy = deploy_service.deploy!(stage, deploy_params)
+    Samson::Hooks.fire(:post_ledger_event, @deploy) if ENV["RAILS_ENV"] != 'test'
```
grosser: hooks should be universal ... reuse before_deploy or after_deploy ?
... and always fire them ...

app/models/job_execution.rb
```diff
@@ -124,6 +124,7 @@ def run!
     params: '{ job_id: id, project: job.project.try(:name), reference: reference }'
 
   def finish
+    Samson::Hooks.fire(:post_ledger_event, @job.deploy) if ENV["RAILS_ENV"] != 'test'
```
grosser: same here

plugins/ledger/README.md
```diff
@@ -0,0 +1,3 @@
+# Ledger Plugin
+
+Little plugin that allows samson to post events to ledger
```
grosser: ... internal zendesk xyz tool

grosser: mention that it needs LEDGER_URL etc ... maybe best to remove that from the global readme since nobody else will need that ...

plugins/ledger/config/initializers/ledger/client.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+#
+# basic ledger event client
+#
+require 'faraday'
```
grosser: plugin should depend on faraday just in case ... `gem.add_runtime_dependency 'faraday'`

plugins/ledger/config/initializers/ledger/client.rb
```diff
@@ -0,0 +1,60 @@
...
+require 'openssl'
+
+LEDGER_URI = "/api/v1/events"
```
grosser: should be inside the class and not global

plugins/ledger/config/initializers/ledger/client.rb
```diff
@@ -0,0 +1,60 @@
...
+
+module Ledger
+  class Client
+    class << self
+      def verify_env
+        raise "LEDGER_TOKEN is not set in ENV" unless ENV.fetch("LEDGER_TOKEN", false)
```
grosser: using fetch with false rarely makes any sense ...

```Ruby
['LEDGER_TOKEN', 'LEDGER_BASE_URL'].each { |k| raise "#{k} is not set in env" unless ENV[k] }
```

plugins/ledger/config/initializers/ledger/client.rb
```diff
@@ -0,0 +1,60 @@
...
+        verify_env
+        deployment(deploy)
+      end
+
+      private
+
+      def create_event(event)
+        events = { "events": [event] }
+        results = post(events)
+        # TODO: raise or display errors if results.status != 200
```
grosser: solve todo ?

plugins/ledger/config/initializers/ledger/client.rb
```diff
@@ -0,0 +1,60 @@
...
+        results
+      end
+
+      def deployment(deploy)
```
grosser: methods that do something should have a verb in their name ... post_deployment or send_deployment

plugins/ledger/config/initializers/ledger/client.rb
```diff
@@ -0,0 +1,60 @@
...
+        event = {
+          id:           deploy.id,
+          name:         deploy.project.name,
+          actor:        deploy.user.name,
+          status:       deploy.status,
+          started_at:   deploy.created_at.iso8601,
+          summary:      deploy.summary,
+          environment:  deploy.stage.deploy_groups.map(&:environment).uniq.map(&:name).join(","),
+          url:          deploy.url,
+          pods:         deploy.stage.deploy_groups.map(&:permalink)
```
grosser: prefer env_value ... should be more stable

plugins/ledger/config/initializers/ledger/client.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      def deployment(deploy)
+        event = {
+          id:           deploy.id,
+          name:         deploy.project.name,
+          actor:        deploy.user.name,
+          status:       deploy.status,
+          started_at:   deploy.created_at.iso8601,
+          summary:      deploy.summary,
+          environment:  deploy.stage.deploy_groups.map(&:environment).uniq.map(&:name).join(","),
```
grosser: might want to use permalink here and/or downcase to avoid `Production` vs `production` ... possibly also not do anything if env name is not in the list of `master, staging, production` since that would just confuse ledger

plugins/ledger/ledger.gemspec
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+Gem::Specification.new 'ledger', '0.0.1' do |s|
+  s.summary = 'Allow deploying projects using Kubernetes'
+  s.description = 'TBD'
+  s.authors = ['ian Waters']
+  s.email = 'iwaters@zendesk.com'
+
+  # Commented out while we use our own branch of kubeclient
```
grosser: don't need that

plugins/ledger/ledger.gemspec
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+Gem::Specification.new 'ledger', '0.0.1' do |s|
+  s.summary = 'Allow deploying projects using Kubernetes'
+  s.description = 'TBD'
```
grosser: add description ?

test/integration/cleanliness_test.rb
```diff
@@ -115,7 +115,8 @@ def check_content(files)
       "plugins/kubernetes/app/decorators/environment_decorator.rb",
       "plugins/kubernetes/app/models/kubernetes/cluster_deploy_group.rb",
       "plugins/kubernetes/app/models/kubernetes/service.rb",
-      "plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb"
+      "plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb",
+      "plugins/ledger/lib/samson_ledger/samson_plugin.rb",
```
grosser: tests :D

plugins/ledger/test/test_helper.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+require_relative '../../../test/test_helper'
+require 'webmock/minitest'
```
grosser: don't need an extra require

plugins/ledger/config/initializers/ledger/client.rb
irwaters: looks like it supports 'other' for ENVs, will use permalink though

app/controllers/deploys_controller.rb
```diff
@@ -99,6 +99,7 @@ def new
   def create
     deploy_service = DeployService.new(current_user)
     @deploy = deploy_service.deploy!(stage, deploy_params)
+    Samson::Hooks.fire(:post_ledger_event, @deploy)
```
grosser: use after or before deploy ... otherwise make a new generic event

plugins/samson_ledger/README.md
```diff
@@ -0,0 +1,8 @@
+# Ledger Plugin
+
+Little plugin that allows samson to post events to ledger (an internal
```
grosser: can remove the redundant `Little plugin that `

plugins/samson_ledger/config/initializers/samson_ledger/client.rb
```diff
@@ -0,0 +1,65 @@
...
+#
+require 'faraday'
+require 'openssl'
+
+module Ledger
+  class Client
+    LEDGER_URI = "/api/v1/events"
+    class << self
+      def verify_env
+        raise "LEDGER_TOKEN is not set in ENV" unless ENV.fetch("LEDGER_TOKEN")
```
grosser: ... already posted the code :(
.fetch will blow up with a KeyError when it is not set, making the raise never fire

grosser: private

grosser: ... also don't need that since plugin_enabled? already takes care of checking ...

plugins/samson_ledger/samson_ledger.gemspec
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+Gem::Specification.new 'samson_ledger', '0.0.1' do |s|
+  s.summary = 'basic client to allow posting events to ledger (internal zendesk system)'
+  s.description = 'small client to post events to the ledger system (an internal zendesk tool)'
+  s.authors = ['ian Waters']
+  s.email = 'iwaters@zendesk.com'
+  s.email = 'iwaters@zendesk.com'
```
grosser: duplicate

plugins/samson_ledger/test/initializers/ledger_test.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+require_relative "../test_helper"
+
+SingleCov.covered! file: 'plugins/samson_ledger/config/initializers/samson_ledger/client.rb'
```
grosser: move test to the right location to avoid setting `file:` ?

plugins/samson_ledger/test/initializers/ledger_test.rb
```diff
@@ -0,0 +1,80 @@
...
+    stub_request(:post, "https://foo.bar/api/v1/events")
+  end
+  after do
+    ENV.delete("LEDGER_BASE_URL")
+    ENV.delete("LEDGER_TOKEN")
+  end
+
+  describe "plugin enabled" do
+    around do |test|
+      ENV.delete("LEDGER_BASE_URL")
```
grosser: using an around as a before ?

... the after outside should already take care of that

plugins/samson_ledger/test/initializers/ledger_test.rb
```diff
@@ -0,0 +1,80 @@
...
+      ENV.delete("LEDGER_TOKEN")
+      test.call
+    end
+    it "is not enabled" do
+      Ledger::Client.plugin_enabled?
+    end
+  end
+
+  describe "cannot create client w/o a token" do
```
grosser: I think `describes` with a single `it` add unnecessary complexity ... also start with valid and the remove single thing

```Ruby
describe ".plugin_enabled?" do
  before do
   ENV['x'] = 1
  end

  it "is enabeld" do
   assert enabled?
  end

  it "is disabled without x" do
   ENV.delete 'x'
   refute enabled?
  end
end
  
end
```

plugins/samson_ledger/test/test_helper.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+require_relative '../../../test/test_helper'
+
+class ActiveSupport::TestCase
```
grosser: this would be a global change ... avoid ...

plugins/samson_ledger/config/initializers/samson_ledger/client.rb
irwaters: so plugin_enabled? is really just used to guard the plugin.  if this was called directly it should blow up.

grosser: but nobody will be calling it directly ... and if then something shady is going on and a .fetch could be good enough of an error message ...

plugins/samson_ledger/test/initializers/ledger_test.rb
irwaters: it's trying to insert a `lib` into the path for some reason.

grosser: ah yes ... it assumes that tested files are in lib or app ... maybe move the code to lib instead ?

plugins/samson_ledger/.gitignore
```diff
@@ -0,0 +1,8 @@
+.bundle/
```
grosser: there is no dummy app here ... and log / .bundle should already be in base samson

grosser: there is no dummy app here ... and log / .bundle should already be in base samson

grosser: there is no dummy app here ... and log / .bundle should already be in base samson

plugins/samson_ledger/config/initializers/samson_ledger/client.rb
```diff
@@ -0,0 +1,65 @@
...
+      def post_deployment(deploy)
+        verify_env
+        post_event(deploy)
+      end
+
+      private
+
+      def verify_env
+        raise "LEDGER_TOKEN is not set in ENV" unless ENV.fetch("LEDGER_TOKEN")
+        raise "LEDGER_BASE_URL is not set in ENV" unless ENV.fetch("LEDGER_BASE_URL")
```
grosser: can simply fetch when used inside the other code since this should never happen because the plugin_enabled guards against it

plugins/samson_ledger/config/initializers/samson_ledger/client.rb
```diff
@@ -0,0 +1,65 @@
...
+          name:         deploy.project.name,
+          actor:        deploy.user.name,
+          status:       deploy.status,
+          started_at:   deploy.created_at.iso8601,
+          summary:      deploy.summary,
+          environment:  deploy.stage.deploy_groups.map(&:environment).uniq.map(&:permalink).map(&:downcase).join(","),
+          url:          deploy.url,
+          pods:         deploy.stage.deploy_groups.map(&:env_value)
+        }
+        create_event(event)
```
grosser: alternative shallower pattern:

```
def post_event
  events = {"events": build_event}
  ...
end
```

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    stub_request(:post, "https://foo.bar/api/v1/events")
+  end
+  after do
+    ENV.delete("LEDGER_BASE_URL")
+    ENV.delete("LEDGER_TOKEN")
+  end
+
+  describe ".plugin_enabled?" do
+    around do |test|
+      ENV["LEDGER_BASE_URL"] = 'https://foo.bar'
```
grosser: can be a simpler before block instead

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      ENV["LEDGER_TOKEN"] = 'token'
+      test.call
+    end
+
+    it "is enabled" do
+      assert Ledger::Client.plugin_enabled?
+    end
+
+    it "is not enabled" do
```
grosser: it "is not enabled without token"
it "is not enabled without url"

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      ENV.delete("LEDGER_TOKEN")
+      refute Ledger::Client.plugin_enabled?
+    end
+  end
+
+  describe ".post_deployment" do
+    around do |test|
+      ENV["LEDGER_BASE_URL"] = 'https://foo.bar'
```
grosser: use simpler before hook

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
grosser: ideally move this to the top instead of repeating it, should be the default for all these tests ...
.. and use with_env

```
describe "foo" do
  with_env(token: 'xyz')
  it "bar" do
  end
end
```

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  describe "leger returns an error" do
+    before do
+      stub_request(:post, "https://foo.bar/api/v1/events").
+        to_return(status: 401)
+    end
+
+    around do |test|
+      ENV["LEDGER_BASE_URL"] = 'https://foo.bar'
```
grosser: same here

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      ENV["LEDGER_TOKEN"] = 'token'
+      test.call
+    end
+    it "posts an event with a valid client" do
+      assert Ledger::Client.post_deployment(deploy)
+    end
+  end
+
+  describe "leger returns an error" do
```
grosser: should be under post_deployment something like

```
it "fails when posting the event fails"
```

plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require_relative "../../test_helper"
+
+SingleCov.covered! file: 'plugins/samson_ledger/lib/samson_ledger/samson_plugin.rb'
```
grosser: both under lib, so no need for `file:` ?

plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
```diff
@@ -0,0 +1,33 @@
...
+
+describe "the plugin callback" do
+  let(:deploy) { Deploy.first }
+
+  before do
+    stub_request(:post, "https://foo.bar/api/v1/events")
+  end
+
+  around do |test|
```
grosser: use with_env

plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require_relative "../../test_helper"
+
+SingleCov.covered! file: 'plugins/samson_ledger/lib/samson_ledger/samson_plugin.rb'
+
+describe "the plugin callback" do
```
grosser: when multiple tests fail this description will not be helpful ... nest under `describe SamsonLedger do`

plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
```diff
@@ -0,0 +1,33 @@
...
+  let(:deploy) { Deploy.first }
+
+  before do
+    stub_request(:post, "https://foo.bar/api/v1/events")
```
grosser: won't fail when the request is not made

```
it "foo" do
  assert_requests :post, "https://foo.bar/api/v1/events" do
    ... fire ...
  end
end
```

plugins/samson_ledger/config/initializers/samson_ledger/client.rb
```diff
@@ -0,0 +1,59 @@
...
+module Ledger
+  class Client
+    LEDGER_URI = "/api/v1/events"
+    class << self
+      def plugin_enabled?
+        ENV["LEDGER_TOKEN"] && ENV["LEDGER_BASE_URL"]
+      end
+
+      def post_deployment(deploy)
+        false unless plugin_enabled?
```
grosser: already checked outside ... also missing the return here ...

plugins/samson_ledger/config/initializers/samson_ledger/client.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+#
+# basic ledger event client
+#
+require 'faraday'
+require 'openssl'
+
+module Ledger
+  class Client
+    LEDGER_URI = "/api/v1/events"
```
grosser: PATH not URI

plugins/samson_ledger/config/initializers/samson_ledger/client.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def post(data)
+        connection.post do |request|
+          request.headers['Content-Type'] = "application/json"
+          request.headers['Authorization'] = "Token token=#{ENV.fetch("LEDGER_TOKEN")}"
+          request.body = data.to_json
+        end
+      end
+
+      def connection
```
grosser: maybe inline this, does not add much value ...

plugins/samson_ledger/samson_ledger.gemspec
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+Gem::Specification.new 'samson_ledger', '0.0.1' do |s|
+  s.summary = 'basic client to allow posting events to ledger (internal zendesk system)'
```
grosser: can do `s.summary = s.description = "`

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,54 @@
...
+  let(:deploy) { Deploy.first }
+
+  before do
+    stub_request(:post, "https://foo.bar/api/v1/events")
+  end
+
+  with_env(LEDGER_BASE_URL: 'https://foo.bar', LEDGER_TOKEN: "sometoken")
+
+  after do
+    ENV.delete("LEDGER_BASE_URL")
```
grosser: not needed, with_env cleans up

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,54 @@
...
+      assert Ledger::Client.plugin_enabled?
+    end
+
+    it "is not enabled without token" do
+      ENV.delete("LEDGER_TOKEN")
+      refute Ledger::Client.plugin_enabled?
+    end
+
+    it "is not enabled without base_url" do
+      ENV["LEDGER_TOKEN"] = 'sometoken'
```
grosser: no need to set that

plugins/samson_ledger/test/config/initializers/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,54 @@
...
+      ENV.delete("LEDGER_BASE_URL")
+      refute Ledger::Client.plugin_enabled?
+    end
+  end
+
+  describe ".post_deployment" do
+    it "posts an event with a valid client" do
+      assert Ledger::Client.post_deployment(deploy)
```
grosser: can do `@event_sent = stub_request` in the before and then `assert_requested @event_sent` here

plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
```diff
@@ -0,0 +1,29 @@
...
+describe "the samson_ledger plugin callback" do
+  let(:deploy) { Deploy.first }
+
+  before do
+    stub_request(:post, "https://foo.bar/api/v1/events")
+  end
+
+  with_env(LEDGER_TOKEN: "sometoken", LEDGER_BASE_URL: "https://foo.bar")
+
+  let(:deploy) { Deploy.first }
```
grosser: .first can be different in different dbs ... prefer `deploys(:succeeded_test)`

plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  it "successfully fires the before_deploy" do
+    assert Samson::Hooks.fire(:before_deploy, deploy, nil)
+  end
+
+  it "successfully fires the after_deploy" do
+    assert Samson::Hooks.fire(:after_deploy, deploy, nil)
+  end
+
+  it "fails to fires the after_deploy" do
```
grosser: other tests do not assert anything, so I could break the hook and this would still pass ... also will fail when any other hook code changes ...

prefer `Foobar.expects(:post_deploy)` & ``Foobar.expects(:post_deploy).never``

plugins/samson_ledger/config/initializers/samson_ledger/client.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
```
grosser: no need for the client to be in the initializer ... can be required by the plugin and live in lib

plugins/samson_ledger/test/lib/samson_ledger/samson_plugin_test.rb
irwaters: all I need is a deploy object here, ordering doesn't matter.

plugins/samson_ledger/lib/samson_ledger/client.rb
```diff
@@ -0,0 +1,56 @@
...
+module Ledger
+  class Client
+    LEDGER_PATH = "/api/v1/events"
+    class << self
+      def plugin_enabled?
+        ENV["LEDGER_TOKEN"] && ENV["LEDGER_BASE_URL"]
+      end
+
+      def post_deployment(deploy)
+        return false unless plugin_enabled?
```
grosser: either check here or in the plugin hooks, but not both

plugins/samson_ledger/lib/samson_ledger/samson_plugin.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+require_relative "client"
```
grosser: can it not be lazy-loaded ?

plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  describe ".post_deployment" do
+    it "posts an event with a valid client" do
+      Ledger::Client.post_deployment(deploy)
+      assert_requested(@event_sent)
+    end
+  end
+
+  describe "fails when posting an event" do
```
grosser: should be under `describe ".post_deployment"`

grosser:
squasch plz, so many commits :)

PR #1282, opened 2016-09-15
notify airbrake on deploy

grosser:
putting our secrets to good use ... mostly meant for kubernetes, but can also replace cap tasks

@zendesk/samson @jonmoter @irwaters 

screen shot 2016-09-15 at 1 59 13 pm

### DIFF
--- a/app/controllers/admin/secrets_controller.rb
+++ b/app/controllers/admin/secrets_controller.rb
@@ -75,7 +75,7 @@ class Admin::SecretsController < ApplicationController
   end
 
   def find_secret
-    @secret = SecretStorage.read(key, include_secret: true)
+    @secret = SecretStorage.read(key, include_value: true)
   end
 
   def find_project_permalinks
--- a/plugins/airbrake/lib/samson_airbrake/samson_plugin.rb
+++ b/plugins/airbrake/lib/samson_airbrake/samson_plugin.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+module SamsonAirbrake
+  class Engine < Rails::Engine
+  end
+
+  class Notification
+    class << self
+      VALID_RAILS_ENV = /^[a-z]+$/
+      SECRET_KEY = 'airbrake_api_key'
+
+      def deliver_for(deploy)
+        return unless DeployGroup.enabled?
+        return unless deploy.stage.notify_airbrake
+        return unless deploy.succeeded?
+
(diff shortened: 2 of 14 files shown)


app/controllers/admin/secrets_controller.rb
```diff
@@ -75,7 +75,7 @@ def failure_response(message)
   end
 
   def find_secret
-    @secret = SecretStorage.read(key, include_secret: true)
+    @secret = SecretStorage.read(key, include_value: true)
```
grosser: renaming that since include_secret + fetch(:value) was confusing ...

grosser:
screen shot 2016-09-15 at 1 05 55 pm

plugins/airbrake/lib/samson_airbrake/samson_plugin.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      private
+
+      def read_secret(project, deploy_groups, key)
+        resolver = Samson::Secrets::KeyResolver.new(project, deploy_groups)
+        full_key = resolver.expand('unused-param', key).map(&:last).first
+        SecretStorage.read_multi([full_key], include_value: true).values.first&.fetch(:value) # read key without raising
+      end
+
+      # not multi-thread safe... this might lead to some samson errors going to other projects / being lost
```
jonmoter: Yeah, the thread-unsafety of this seems dangerous.  Fork?

jonmoter: If you don't fork, then should create a global variable with the original airbrake API key (set in initializer) and always restore to that original key.  Otherwise there could be race conditions of > 1 project clobbering this key.

jonmoter: Or patch the Airbrake library to allow passing in a key.

jonmoter: Here you go.  You can configure multiple notifiers in Airbrake: https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby.rb#L88-L94

grosser:
using api directly now so we don't have to upgrade airbrake and we don't have to know the projects id

grosser:
updated to put the setting on the stage for more granular control

grosser:
@jonmoter good to go ?

PR #1285, opened 2016-09-15
do not fail when error message is too long

grosser:
@sandlerr ... also fixing up test coverage

https://zendesk.airbrake.io/projects/95346/groups/1714212048688589520

### DIFF
--- a/plugins/jenkins/app/models/samson/jenkins.rb
+++ b/plugins/jenkins/app/models/samson/jenkins.rb
@@ -6,8 +6,24 @@ module Samson
     URL = ENV['JENKINS_URL']
     USERNAME = ENV['JENKINS_USERNAME']
     API_KEY = ENV['JENKINS_API_KEY']
+    ERROR_COLUMN_LIMIT = 255
 
     attr_reader :job_name, :deploy
+
+    def self.deployed!(deploy)
+      return unless deploy.succeeded?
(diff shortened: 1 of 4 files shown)


plugins/jenkins/app/models/samson/jenkins.rb
```diff
@@ -6,9 +6,25 @@ class Jenkins
     URL = ENV['JENKINS_URL']
     USERNAME = ENV['JENKINS_USERNAME']
     API_KEY = ENV['JENKINS_API_KEY']
+    ERROR_COLUMN_LIMIT = 255
 
     attr_reader :job_name, :deploy
 
+    def self.deployed!(deploy)
+      return unless deploy.succeeded?
+      deploy.stage.jenkins_job_names.to_s.split(/, ?/).map do |job_name|
```
sandlerr: the old code made sure `jenkins_job_names` was present, should that still happen?

sandlerr:
:+1: on green

plugins/jenkins/app/models/samson/jenkins.rb
grosser: empty string will be split into nothing -> .each will do nothing

sandlerr:
codeclimate and travis were both sad :/

PR #1290, opened 2016-09-19
Validate annotations and catch docker error

jonmoter:
Two things here:

1. Validate that the annotations are a hash, rather than an array
2. If a docker build fails, don't output the entire stream twice

/cc @grosser 

### References
 - Failed deploy: https://samsontest.zende.sk/projects/product_data_warehouse/deploys/3253
 - Airbrake: https://zendesk.airbrake.io/projects/95346/groups/1778521175384075883
 - Failed docker build with extra output: https://samsontest.zende.sk/projects/product_data_warehouse/builds/237#L701

### Risks
- Level: Low

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -75,6 +75,11 @@ class DockerBuilderService
         output.write_docker_chunk(chunk)
       end
     output.puts('### Docker build complete')
+  rescue Docker::Error::UnexpectedResponseError
+    # If the docker library isn't able to find an image id, it returns the
+    # entire output of the "docker build" command, which we've already captured
+    output.puts("Docker build failed (image id not found in response)")
+    nil
   rescue Docker::Error::DockerError => e
     # If a docker error is raised, consider that a "failed" job instead of an "errored" job
     output.puts("Docker build failed: #{e.message}")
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/test/samples/kubernetes_deployment.yml
```diff
@@ -18,6 +18,9 @@ spec:
       labels:
         project: some-project
         role: some-role
+      annotations:
```
grosser: don't need to add that ?

app/models/docker_builder_service.rb
```diff
@@ -75,6 +75,11 @@ def build_image(tmp_dir)
         output.write_docker_chunk(chunk)
       end
     output.puts('### Docker build complete')
+  rescue Docker::Error::UnexpectedResponseError
+    # If the docker library isn't able to find an image id, it returns the
+    # entire output of the "docker build" command, which we've already captured
+    output.puts("Docker build failed (image id not found in response)")
```
grosser: any way to test that ?

plugins/kubernetes/test/samples/kubernetes_deployment.yml
jonmoter: Don't strictly need to, but I wanted to make sure that the verifier passed with a file that did contain annotations.

app/models/docker_builder_service.rb
jonmoter: test added

jonmoter:
Travis tests passing now.

PR #1291, opened 2016-09-20
correctly fetch data when it is an ActionController::Parameters object

anthonywoo:
When we moved to rails 5 - `params` is no longer a hash and is now an `ActionController::Parameters` object. Since it is now this object - when we pass in `params` to `Sawyer` (to get `@data`) it preserves the object type and calling method notation on the Sawyer resource no longer works.

I tried to look into converting `params` to a hash before we create the `Sawyer` resource using `params.permit()` however - that proved to be very difficult when dealing with webhook data

I am open to suggestions if anyone has a better idea on how to convert `ActionController::Parameters` to a hash

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low. Github Pull request Integration may still be broken...

### DIFF
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -84,19 +84,19 @@ class Changeset::PullRequest
   end
 
   def sha
-    @data.head.sha
+    @data['head']['sha']
   end
 
   def branch
-    @data.head.ref
+    @data['head']['ref']
   end
 
   def state
-    @data.state
+    @data['state']
   end
 
   def users
-    users = [@data.user, @data.merged_by]
+    users = [@data['user'], @data['merged_by']]
     users.compact.map { |user| Changeset::GithubUser.new(user) }.uniq
   end
 
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -41,7 +41,7 @@ class Changeset::PullRequest
   end
 
   def self.changeset_from_webhook(project, params = {})
-    data = Sawyer::Resource.new(Octokit.agent, params['pull_request'])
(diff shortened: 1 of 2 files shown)


grosser:
the reason we did not see this was missing tests ... so let's add one that fails without these changes :)

sandlerr:
> I am open to suggestions if anyone has a better idea on how to convert `ActionController::Parameters` to a hash

Would this be suitable? http://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-to_h

anthonywoo:
> Would this be suitable? http://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-to_h

I was looking to use that method - but it only returns permitted keys which I thought might be a hassle to permit all keys we need from the github webhook. Will take a 2nd look at it though as it seems cleaner

anthonywoo:
💇 @grosser @sandlerr added a test - although it would've been difficult to detect this type of change when we upgraded to rails 5 as the `params` object changed from a hash and we assumed it was a hash in our tests

Also rubocop is failing on something not related in this PR - is it ok to ignore it? (Or I can increase the `LineLength` cop 😄 )

grosser:
@dragonfax caused that ... plz fix the line here so we get back our green master :)

grosser:
👍

sandlerr:
There's also a `to_unsafe_h` method that returns an unfiltered version

irwaters:
You can also do a to_unsafe_h, then a to_h on that….

anthonywoo:
Thanks! `.to_unsafe_h` was what I was loooking for!

grosser:
👍

PR #1295, opened 2016-09-20
stop overriding a ActiveRecord method

grosser:
@irwaters removing bad code on the hunt for the evil deploy bug ...
/cc @zendesk/samson


grosser:
@fneves

PR #1296, opened 2016-09-21
do not call clone directory and split clone into real clone and creat…

grosser:
…e_workspace

updated tests got the repo in an unrealistic state that had a full checkout instead of a mirror

@fneves @irwaters @jonmoter

### DIFF
--- a/test/models/git_repository_test.rb
+++ b/test/models/git_repository_test.rb
@@ -22,18 +22,20 @@ describe GitRepository do
     repository.repository_directory.must_equal project.repository_directory
   end
 
-  describe "#clone!" do
+  describe "#create_workspace" do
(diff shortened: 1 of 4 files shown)


test/models/git_repository_test.rb
```diff
@@ -77,48 +79,46 @@
...
-      Dir.chdir(repository.repo_cache_dir) do
-        raise unless system("git remote rm origin")
-      end
+      repository.send(:clone!).must_equal(true)
+      assert system("rm -rf #{repository.repo_cache_dir}/*")
       repository.update_local_cache!.must_equal false
     end
   end
 
   describe "#cheakout!" do
```
irwaters: checkout?

PR #1298, opened 2016-09-21
remove redundant info

grosser:
@dragonfax icon has a hover ... don't need this text on every page ...

screen shot 2016-09-21 at 10 10 00 am


grosser:
@dragonfax

dragonfax:
Agreed :+1:

PR #1299, opened 2016-09-22
blanked add imagePullSecrets

grosser:
@jonmoter @irwaters 

at least until kubernetes 1.3 we need to add imagePullSecrets ... so add a nice way to add them to all projects we deploy so we don't have to change that everywhere

https://github.com/zendesk/chef_zendesk_samson/pull/193/files

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -107,8 +107,6 @@ GITHUB_TOKEN=
 
 ## Kubernetes
 # SECRET_SIDECAR_IMAGE= # optional, path to the docker image that built from the secret_puller service.  service will be disabled w/o it
-# KUBERNETES_IMAGE_PULL_SECRETS="ecr-auth-us-west-1,..." # override imagePullSecrets before deploying to make docker authenticate to your repository
-
 
 ## Jenkins, optional, for triggering Jenkins builds after deployment
 # JENKINS_URL= # server_url of jenkins
--- a/.env.example
+++ b/.env.example
@@ -105,8 +105,9 @@ GITHUB_TOKEN=
 ## Secret storage
 # SECRET_STORAGE_BACKEND= # optional, should be one of: SecretStorage::DbBackend (default) or SecretStorage::HashicorpVault
 
-## Kubernets sidecar
+## Kubernetes
 # SECRET_SIDECAR_IMAGE= # optional, path to the docker image that built from the secret_puller service.  service will be disabled w/o it
+# KUBERNETES_IMAGE_PULL_SECRETS="ecr-auth-us-west-1,..." # override imagePullSecrets before deploying to make docker authenticate to your repository
 
 
 ## Jenkins, optional, for triggering Jenkins builds after deployment
--- a/plugins/kubernetes/app/models/kubernetes/resource_template.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource_template.rb
@@ -202,10 +202,17 @@ module Kubernetes
       env.merge!(Samson::Hooks.fire(:deploy_group_env, project, @doc.deploy_group).inject({}, :merge!))
     end
 
+    # kubernetes needs docker secrets to be able to pull down images from the registry
+    # in kubernetes 1.3 this might work without this workaround
     def set_image_pull_secrets
-      return unless secrets = ENV['KUBERNETES_IMAGE_PULL_SECRETS']
-      template[:spec].fetch(:template, {}).fetch(:spec, {})[:imagePullSecrets] =
-        secrets.split(",").map { |s| {name: s} }
+      secrets = @doc.client.get_secrets(namespace: @doc.namespace)
+      docker_credentials = secrets.
+        select { |secret| secret.type == "kubernetes.io/dockercfg" }.
+        map! { |c| {name: c.metadata.name} }
+
+      return if docker_credentials.empty?
+
+      template[:spec].fetch(:template, {}).fetch(:spec, {})[:imagePullSecrets] = docker_credentials
     end
 
     def needs_secret_sidecar?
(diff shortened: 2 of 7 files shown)


.env.example
```diff
@@ -105,8 +105,9 @@ GITHUB_TOKEN=
 ## Secret storage
 # SECRET_STORAGE_BACKEND= # optional, should be one of: SecretStorage::DbBackend (default) or SecretStorage::HashicorpVault
 
-## Kubernets sidecar
+## Kubernets
```
jonmoter: Spelling fix?

.env.example
```diff
@@ -105,8 +105,9 @@ GITHUB_TOKEN=
...
 # SECRET_SIDECAR_IMAGE= # optional, path to the docker image that built from the secret_puller service.  service will be disabled w/o it
+# KUBERNETES_IMAGE_PULL_SECRETS="ecr-auth-us-west-1,..." # override imagePullSecrets before deploying
```
jonmoter: It's not clear what this ENV var is supposed to be. Could you add a URL to some docs, or explain it?

plugins/kubernetes/app/models/kubernetes/resource_template.rb
```diff
@@ -210,6 +211,12 @@ def set_env
       end
     end
 
+    def set_image_pull_secrets
+      return unless secrets = ENV['KUBERNETES_IMAGE_PULL_SECRETS']
```
jonmoter: Is it a problem if we set this for a 1.3.x release of Kubernetes?  If so, we could ask the Kubernetes API for its version.

```
$ curl -sS http://127.0.0.1:9099/version | jq .
{
  "major": "1",
  "minor": "2",
  "gitVersion": "v1.2.4.3+f4ece03d458602",
  "gitCommit": "f4ece03d458602c440b58ce90d0c396e9547e6c8",
  "gitTreeState": "clean"
}
```

grosser: not sure if it will be solved with 1.3 so I'd rather wait with that until I can verify it works

.env.example
grosser: added

.env.example
grosser: fixed

irwaters:
you can test this in 1.3 locally buy downloading our fork and trying it out.  Follow the instructions at the bottom of this page.
https://zendesk.atlassian.net/wiki/display/PAAS/Building+Kubernetes+from+Source

use the v1.3.4-audit branch

once the building of quick-release is done,  chdir to cluster and ./kube-up.sh to create a local cluster.

grosser:
It does not have any negative effects on 1.3 ... so idk why we should
remove it ... might still be useful for some cases and easy to
ignore/remove when not needed

grosser:
@jonmoter good to go as is ? 
... also need +1 on the chef change after merging this

.env.example
```diff
@@ -105,8 +105,9 @@ GITHUB_TOKEN=
 ## Secret storage
 # SECRET_STORAGE_BACKEND= # optional, should be one of: SecretStorage::DbBackend (default) or SecretStorage::HashicorpVault
 
-## Kubernets sidecar
+## Kubernetes
 # SECRET_SIDECAR_IMAGE= # optional, path to the docker image that built from the secret_puller service.  service will be disabled w/o it
+# KUBERNETES_IMAGE_PULL_SECRETS="ecr-auth-us-west-1,..." # override imagePullSecrets before deploying to make docker authenticate to your repository
```
jonmoter: Suggestion: include a URL for http://kubernetes.io/docs/user-guide/images/#specifying-imagepullsecrets-on-a-pod

Comment that this is a comma-separated list of the names of secrets that are expected to be present in the cluster/namespace that you're deploying to.

jonmoter:
Try loading them from the cluster themselves, rather than needing an env var.

grosser:
imgePullSecrets supports merging multiple secrets, so adding all should be fine

grosser:
only issue could be that you have multiple secrets for the same registry and they override each other ... but hopefully that won't happen ...

grosser:
reworked to look up services instead ... looks good ?

grosser:
works ... https://samsontest.zende.sk/projects/kube_service_watcher/deploys/3310

plugins/kubernetes/app/models/kubernetes/resource_template.rb
```diff
@@ -210,6 +211,20 @@ def set_env
       end
     end
 
+    # kubenretes needs docker secrets to be able to pull down images from the registry
+    # in kubernetes 1.3 this might work without this workaround
+    def set_image_pull_secrets
+      secrets = @doc.deploy_group.kubernetes_cluster.client.
```
jonmoter: Can simplify to use `@doc.client`

plugins/kubernetes/app/models/kubernetes/resource_template.rb
```diff
@@ -210,6 +211,20 @@ def set_env
       end
     end
 
+    # kubenretes needs docker secrets to be able to pull down images from the registry
```
jonmoter: Spelling of "kubernetes"

plugins/kubernetes/app/models/kubernetes/resource_template.rb
```diff
@@ -210,6 +211,20 @@ def set_env
...
+    # in kubernetes 1.3 this might work without this workaround
+    def set_image_pull_secrets
+      secrets = @doc.deploy_group.kubernetes_cluster.client.
+        get_secrets(namespace: @doc.deploy_group.kubernetes_namespace)
```
jonmoter: `ReleaseDoc` also has a `#namespace` method to shorten these method chains

PR #1302, opened 2016-09-22
make env plugin read secrets

grosser:
@henders @jonmoter @rgueldem @andreionut @sbrnunes @appsan @irwaters 

env value would be `secret://foobar` and that would then lookup the secret

enables us to use secrets in env workflow

for https://zendesk.atlassian.net/wiki/display/ENG/Configuring+projects+from+Samson+-+Proposal

downside would be that secrets will be in the env ... so maybe these should only be for "not so secret secrets" ... maybe "visible" secrets ...

### DIFF
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -17,6 +17,9 @@ class EnvironmentVariable < ActiveRecord::Base
       end
 
       resolve_dollar_variables(env)
+      resolve_secrets(project, deploy_group, env)
+
+      env
     end
 
     # also used by private plugin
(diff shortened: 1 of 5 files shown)


grosser:
... could also make this part op-in via an env variable ...

plugins/env/app/models/environment_variable.rb
```diff
@@ -47,6 +50,16 @@ def resolve_dollar_variables(env)
         end
       end
     end
+
+    def resolve_secrets(project, deploy_group, env)
+      resolver = Samson::Secrets::KeyResolver.new(project, Array(deploy_group))
```
henders: If they don't have this plugin enabled, will this throw an exception?

henders: Can't remember if we load all code regardless of which plugins are enabled.

grosser: - secrets is not a plugin
 - we only load code for enabled plugins

henders: Oh, whoops! :)

Anyway, lgtm 👍

PR #1305, opened 2016-09-22
Add more info for ledger

neroleung:
Add pod ids and pull requests. Use `deploy.updated_at` instead of `created_at` for ledger's started_at.

/cc @zendesk/samson, @irwaters, @grosser 

### Risks
- Level: Low

### DIFF
--- a/plugins/samson_ledger/lib/samson_ledger/client.rb
+++ b/plugins/samson_ledger/lib/samson_ledger/client.rb
@@ -34,17 +34,41 @@ module SamsonLedger
 
       def build_event(deploy)
         event = {
-          id:           deploy.id,
-          name:         deploy.project.name,
-          actor:        deploy.user.name,
-          status:       deploy.status,
-          started_at:   deploy.created_at.iso8601,
-          summary:      deploy.summary,
-          environment:  deploy.stage.deploy_groups.map(&:environment).uniq.map(&:permalink).map(&:downcase).join(","),
-          url:          deploy.url,
-          pods:         deploy.stage.deploy_groups.map(&:env_value)
+          id:            deploy.id,
+          name:          deploy.project.name,
+          actor:         deploy.user.name,
+          status:        deploy.status,
+          started_at:    deploy.updated_at.iso8601,
+          summary:       deploy.summary,
+          environment:   deploy.stage.deploy_groups.map(&:environment).uniq.map(&:permalink).map(&:downcase).join(","),
+          url:           deploy.url,
+          pods:          pods(deploy.stage.deploy_groups),
+          pull_requests: pull_requests(deploy.changeset)
         }
-        { "events": [event] }
+        {"events": [event]}
+      end
+
+      def pods(deploy_groups)
+        deploy_groups.map { |dg| dg.env_value[/^(pod|staging|master)(\d+).*/, 2] }.compact.map(&:to_i).sort
+      end
+
+      def pull_requests(changeset)
+        # Note: All HTML is sanitized at rendering time on Ledger.
+        results = changeset.pull_requests.map do |pull_request|
+          github_users = pull_request.users.compact.map do |user|
+            "<a href='#{user.url}'><img src='#{user.avatar_url}' width=20 height=20 /></a>"
+          end
+
+          <<-HTML.strip_heredoc.tr("\n", ' ')
+            <li>
+              #{github_users.join}
+              <strong>##{pull_request.number}</strong>
+              <a href='#{pull_request.url}' target='_blank'>#{pull_request.title}</a>
+            </li>
+          HTML
+        end
+
+        "<ul>#{results.join}</ul>" if results.any?
       end
 
       def post(data)
--- a/plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
+++ b/plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
@@ -4,11 +4,9 @@ require_relative "../../test_helper"
 SingleCov.covered!
 
 describe 'SamsonLedger::Client' do
-  let(:deploy) { Deploy.first }
+  include StructHelper
 
-  before do
-    @event_sent = stub_request(:post, "https://foo.bar/api/v1/events")
-  end
+  let(:deploy) { Deploy.first }
 
   with_env(LEDGER_BASE_URL: 'https://foo.bar', LEDGER_TOKEN: "sometoken")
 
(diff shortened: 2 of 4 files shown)


plugins/samson_ledger/lib/samson_ledger/client.rb
```diff
@@ -34,17 +34,42 @@ def post_event(deploy)
...
+          url:           deploy.url,
+          pods:          pods(deploy.stage.deploy_groups),
+          pull_requests: pull_requests(deploy.changeset)
         }
-        { "events": [event] }
+        {"events": [event]}
+      end
+
+      def pods(deploy_groups)
+        deploy_groups.map { |dg| dg.env_value.match(/^pod(\d+).*/) }.compact.map { |m| m[1].to_i }
```
grosser: ignore things that are not called pod ? ... maybe `use db.env_value[/\d+/]` ?
... in general prefer `[]` to `match` ... `deploy_groups.map { |dg| dg.env_value[/^pod(\d+).*/, 1] }.compact.map(&:to_i)`

plugins/samson_ledger/lib/samson_ledger/client.rb
```diff
@@ -34,17 +34,42 @@ def post_event(deploy)
...
+          result = "<li>"
+          result += github_users.join
+          result += " <strong>##{pull_request.number}</strong> "
+          result += "<a href='#{pull_request.url}' target='_blank'>#{pull_request.title}</a>"
+          result += "</li>"
+
+          result
+        end
+
+        "<ul>#{results.join}</ul>" unless results.empty?
```
grosser: unless empty -> if any

plugins/samson_ledger/lib/samson_ledger/client.rb
```diff
@@ -34,17 +34,42 @@ def post_event(deploy)
...
         }
-        { "events": [event] }
+        {"events": [event]}
+      end
+
+      def pods(deploy_groups)
+        deploy_groups.map { |dg| dg.env_value.match(/^pod(\d+).*/) }.compact.map { |m| m[1].to_i }
+      end
+
+      def pull_requests(changeset)
```
grosser: constructing html in a json api seems like a bad idea ... also great way to get XSS going in ledger ;)

plugins/samson_ledger/lib/samson_ledger/client.rb
```diff
@@ -34,17 +34,42 @@ def post_event(deploy)
...
+        results = changeset.pull_requests.map do |pull_request|
+          github_users = pull_request.users.map do |user|
+            unless user.nil?
```
grosser: unless user.nil? -> if user

grosser: ... or `.compact` above

plugins/samson_ledger/lib/samson_ledger/client.rb
```diff
@@ -34,17 +34,42 @@ def post_event(deploy)
...
+              "<a href='#{user.url}'><img src='#{user.avatar_url}' width=20 height=20 /></a>"
+            end
+          end
+
+          result = "<li>"
+          result += github_users.join
```
grosser: use `<<` to avoid string garbadge ... ideally use heredoc

```Ruby
<<-HTML
  <bla>
    #{value}
  <blob>
HTML
```

plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
```diff
@@ -4,12 +4,23 @@
 SingleCov.covered!
 
 describe 'SamsonLedger::Client' do
-  let(:deploy) { Deploy.first }
+  ComparisonStruct  = Struct.new(:commits)
```
grosser: do not define constants in tests, they will be global ... use let or local variables

plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
```diff
@@ -29,10 +40,71 @@
   end
 
   describe ".post_deployment" do
+    before do
+      @asserts = Proc.new {}
```
grosser: `let(:calls) { [] }` then `calls << event` then `calls.first.must_equal xyz`

plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
```diff
@@ -29,10 +40,71 @@
...
+      @request_lambda = -> (request) do
+        event = JSON.parse(request.body)['events'].first
+        @asserts.call(event)
+        request
+      end
+      @event_sent = stub_request(:post, "https://foo.bar/api/v1/events").with { |request| @request_lambda[request] }
```
grosser: use .call ... more readable then `[]`

grosser: ... or with(& request_lambda) ... also does not need to be an instance variable

plugins/samson_ledger/lib/samson_ledger/client.rb
neroleung: In production, the examples are like `pod2`, `staging101`, `master99`, and `us-west-1`. I don't want `us-west-1` though. I will change the regex to: 
`dg.env_value[/^(pod|staging|master)(\d+).*/, 2]`

plugins/samson_ledger/lib/samson_ledger/client.rb
neroleung: Yeah.. But I think that will make things more flexible though. I use the `sanitize-html` library (https://github.com/punkave/sanitize-html) to sanitize the html before rendering. So it "should" be fine (famous last words). 😅

plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
neroleung: I can't really do `let(:calls) { [] }`. It throws random `[BUG] Segmentation fault at 0x00000000000000` error. Suspecting there might be a bug in the `single_cov` gem because the random error went away after I removed `require 'single_cov'` from `test_helper.rb`.

So instead I do `@results = []` here.

neroleung:
@grosser, all issues should have been addressed. Please review again when you have a moment. Thanks!

plugins/samson_ledger/test/lib/samson_ledger/client_test.rb
grosser: try `let :results` or something else then ... I had a few random segfaults, not sure if that is because of the let or because of singlecov ... but good to know that they are related

neroleung: @grosser, did some more digging into this. It's getting bizarre..

So there are 4 requirements for the random segfaults to happen randomly:
1. `require 'single_cov'`
2. Add my new `#create_singleton_struct` method inside `client_test.rb` instead of `support/struct_helper.rb`
3. Use `let(:results) { [] }` and add item to it inside the `request_lambda` (instance variable `@results = []` does not cause the crash to happen)
4. Do an AR write in the test - `prod_env = environments(:production)`

Take out any of the above will make the random segfaults go away. 

This is more informational at this point. So I now use `let(:results) { [] }` and puts `#create_singleton_struct` in `support/struct_helper.rb`. No crashes.

Github review UI is bugging out like you said. This view is still showing the old code even after I rebased and pushed. Please review when you have a moment. Thanks!

neroleung: Oh. Forgot to mention an interesting side effect when it crashed - the following fixture file were created automatically if and only if it crashes:

```
test/fixtures/slack_identifiers.yml
test/fixtures/kubernetes/cluster_deploy_groups.yml
test/fixtures/kubernetes/clusters.yml
test/fixtures/kubernetes/deploy_group_roles.yml
test/fixtures/kubernetes/release_docs.yml
test/fixtures/kubernetes/releases.yml
test/fixtures/kubernetes/roles.yml
```

grosser: they normally get removed after a run ... but segfault will not run ensure blocks ... so they stay

test/support/struct_helper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+module StructHelper
+  def create_singleton_struct(name, *fields)
```
neroleung: That's because of https://github.com/zendesk/samson/blob/master/app/models/changeset.rb#L83. We need the result of `GITHUB.compare` to be serializable for `Rails.cache`. `Struct.new` without a class name is not serializable. Also referencing https://github.com/zendesk/samson/commit/5b61dc27228e652e15ee9774382ec2fbc02a0bf7. 

Hmm.. Since `GITHUB.compare` returns a `Sawyer::Resource`, I can probably do something like:

```
agent = Sawyer::Agent.new('')
message = Sawyer::Resource.new(agent, message: 'some message')
comparison = Sawyer::Resource.new(agent, commits: [message])
```

Would prefer to do it as a separate refactor PR together with `changeset_test.rb` and `pull_request_test.rb`. Let me know if you want me to fix it in this PR.

The other 2 issues are corrected. Thanks!

plugins/samson_ledger/lib/samson_ledger/client.rb
```diff
@@ -34,17 +34,40 @@ def post_event(deploy)
...
         }
-        { "events": [event] }
+        {"events": [event]}
+      end
+
+      def pods(deploy_groups)
+        deploy_groups.map { |dg| dg.env_value[/^(pod|staging|master)(\d+).*/, 2] }.compact.map(&:to_i).sort
+      end
+
+      def pull_requests(changeset)
```
grosser: a comment like "Html in sanitized in ledger, so no need to be html-safe here"

PR #1306, opened 2016-09-22
dont use cached pull request for issue_comment events

anthonywoo:
Issue comment events should trigger a deploy for the latest commit in the PR. There was an bug where issue comments would find a cached version of the same pull request and deploy an older commit - this happens because the cache key is based on repo and PR number which will be the same for multiple issue comments on the same PR.

Now when finding a pull request associated with an issue comment - it will always send a github request to find the PR data

/cc @zendesk/samson @mwerner @alanhogan 

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/HAR-929

### Risks
- Level: Low

### DIFF
--- a/app/models/changeset/issue_comment.rb
+++ b/app/models/changeset/issue_comment.rb
@@ -34,6 +34,7 @@ class Changeset::IssueComment
   private
 
   def pull_request
-    @pull_request ||= Changeset::PullRequest.find(repo, data['number'])
+    pr_data = GITHUB.pull_request(repo, data['number'])
+    @pull_request ||= Changeset::PullRequest.new(repo, pr_data)
   end
 end
--- a/app/models/changeset/issue_comment.rb
+++ b/app/models/changeset/issue_comment.rb
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 class Changeset::IssueComment
-  attr_reader :repo, :data
+  attr_reader :repo, :data, :comment
   VALID_ACTIONS = ['created'].freeze
 
   def initialize(repo, data)
     @repo = repo
     @body = data['body']
+    @comment = data['comment']
     @data = data['issue']
   end
 
@@ -34,7 +35,9 @@ class Changeset::IssueComment
   private
 
   def pull_request
-    pr_data = GITHUB.pull_request(repo, data['number'])
+    pr_data = Rails.cache.fetch(['IssueComment', repo, comment['id']].join("-")) do
+      GITHUB.pull_request(repo, data['number'])
+    end
     @pull_request ||= Changeset::PullRequest.new(repo, pr_data)
   end
 end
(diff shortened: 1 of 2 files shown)


grosser:
This will cause a lot of overhead for every other usecase ... ideally request an uncached version from inside the controller that receives these updates.

alanhogan:
Setting aside @grosser’s concern for a second (as I’m not sure where else `pull_request` is used and thus can't say when the overhead would come into play), the basic idea here is valid and a welcome fix. Glad to see it. Let someone know if you want help making the change @grosser requested, as this fix is welcome sooner rather than later.

anthonywoo:
What about caching the github response to the `issue_comment` id if it needs to find the associated pull request? I'm not sure about doing it in the controller because most comments aren't trying to trigger a samson deploy and won't need to find its associated pull request (Example - +1 comments won't trigger a deploy and shouldn't make a github request)

test/models/changeset/issue_comment_test.rb
```diff
@@ -29,4 +29,33 @@
...
+    let(:cached_pr_data) do
+      {
+        head: {
+          sha: 'cach123',
+          ref: 'a/test'
+        }
+      }.with_indifferent_access
+    end
+
+    it 'does not read from cached PullRequest' do
```
grosser: giving a reason here would also be helpful for future archeologists

`... does not read from cache to avoid foo bar` ... ideally a higher level test like `shows new comment when user updates pull request` ... so it fails with an error that makes sense / reproduces the actual issue

grosser:
idk much about this codepath / what would be best ... just looking for a fix that does not remove all caching ... otherwise displaying a page with lots of PRs would be super slow

jonmoter:
@grosser I think the `IssueComment` class is only used when a GitHub webhook is triggered. It's a webhook handler, it's not used when displaying comments on a PR. At least as far as I can tell.

So I think this change is safe, and wouldn't impact page rendering.

grosser:
👍  then ...

anthonywoo:
@jonmoter is correct - it shouldn't affect page rendering...
However - I decided to cache the response anyways as the code path is called quite a bit when determining if it should deploy, create a release, build a docker image, etc.

💇 @grosser @mwerner @alanhogan

app/models/changeset/issue_comment.rb
```diff
@@ -34,6 +35,9 @@ def service_type
   private
 
   def pull_request
-    @pull_request ||= Changeset::PullRequest.find(repo, data['number'])
+    pr_data = Rails.cache.fetch(['IssueComment', repo, comment['id']].join("-")) do
```
grosser: I think there is no need to join, rails does that for you

grosser:
👍

mwerner:
Yeah @jonmoter is correct. I originally broke out `IssueComment` to distinguish the three ways a `Changeset` could trigger the webhook `CodePush` `PullRequest` and `IssueComment`. `PullRequest` was the only existing changeset that is being used in other places previous to `[samson review]`. I don't see why any changes on `IssueComment` would affect the performance of `PullRequest` in other places of Samson. +1

PR #1308, opened 2016-09-23
fix bug in creating docker images and add regression test

sandlerr:
Currently if creating a docker image but not a `Release` there is an error due to a missing ivar.

cc @zendesk/samson


grosser:
that testing setup looks a bit fishy ... but 👍

sandlerr:
@grosser I fixed the test setup. The precompile task is crashing Ruby, any idea why? Are we just merging red for now?

grosser:
I'm getting random segfaults so often :(

PR #1309, opened 2016-09-23
make environment variables from env plugin available in kubernetes

grosser:
https://zendesk.atlassian.net/browse/PAAS-415

@jonmoter @irwaters @henders


grosser:
codeclimate is complaining about similar code ...

PR #1310, opened 2016-09-23
Add merge button.

dragonfax:
cc @zendesk/samson @grosser

### Screenshots

samson

blank_skitch_document

### References

* JIRA https://zendesk.atlassian.net/browse/TOOLS-876

### Risks

Low

### DIFF
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -119,6 +119,7 @@ class Admin::DeployGroupsController < ApplicationController
     template_stage.next_stage_ids.delete(stage.id)
     template_stage.save!
 
+    stage.project.stages.reload # need to reload to make verify_not_part_of_pipeline have current data and not fail
     stage.soft_delete!
 
     if !stage.reload.deleted?
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -93,15 +93,7 @@ class Admin::DeployGroupsController < ApplicationController
 
     preexisting_stages.each do |stage|
       template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
-
-      template_stage.deploy_groups += stage.deploy_groups
-      template_stage.next_stage_ids.delete(stage.id)
-      template_stage.save!
-
-      # these interfere with views as they don't respect soft_deletion.
-      DeployGroupsStage.where(stage: stage).delete_all
-
-      stage.soft_delete
+      merge_stage(stage, template_stage)
     end
 
     redirect_to [:admin, deploy_group]
@@ -115,6 +107,22 @@ class Admin::DeployGroupsController < ApplicationController
   end
 
   private
+
+  def merge_stage(stage, template_stage)
+    return unless template_stage
+    return if template_stage.deploy_groups.include?(stage)
+    return if stage.is_template
+    return if stage.deploy_groups.count > 1
+
+    template_stage.deploy_groups += stage.deploy_groups
+    template_stage.next_stage_ids.delete(stage.id)
+    template_stage.save!
+
+    # these don't respect soft_deletion (yet).
+    DeployGroupsStage.where(stage: stage).delete_all
+
+    stage.soft_delete
+  end
 
   def stages_for_creation
     self.class.stages_for_creation(deploy_group)
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -105,7 +105,7 @@ class Admin::DeployGroupsController < ApplicationController
   end
 
   def self.create_all_stages(deploy_group)
-    _, missing_stages = stages_for_creation
+    _, missing_stages = stages_for_creation(deploy_group)
     missing_stages.each do |template_stage|
       create_stage_with_group(template_stage, deploy_group)
     end
@@ -117,40 +117,42 @@ class Admin::DeployGroupsController < ApplicationController
     self.class.stages_for_creation(deploy_group)
   end
 
-  # returns a list of stages already created and list of stages to create (through their template stages)
-  def self.stages_for_creation(deploy_group)
-    environment = deploy_group.environment
-    template_stages = environment.template_stages.all
-    deploy_group_stages = deploy_group.stages.all
+  class << self
+    # returns a list of stages already created and list of stages to create (through their template stages)
+    def stages_for_creation(deploy_group)
+      environment = deploy_group.environment
+      template_stages = environment.template_stages.all
+      deploy_group_stages = deploy_group.stages.all
 
-    preexisting_stages = []
-    missing_stages = []
-    Project.where(include_new_deploy_groups: true).each do |project|
-      template_stage = template_stages.detect { |ts| ts.project_id == project.id }
-      deploy_group_stage = deploy_group_stages.detect { |dgs| dgs.project.id == project.id }
-      if deploy_group_stage
-        preexisting_stages << deploy_group_stage
-      elsif template_stage
-        missing_stages << template_stage
+      preexisting_stages = []
+      missing_stages = []
+      Project.where(include_new_deploy_groups: true).each do |project|
+        template_stage = template_stages.detect { |ts| ts.project_id == project.id }
+        deploy_group_stage = deploy_group_stages.detect { |dgs| dgs.project.id == project.id }
+        if deploy_group_stage
+          preexisting_stages << deploy_group_stage
+        elsif template_stage
+          missing_stages << template_stage
+        end
       end
+
+      [preexisting_stages, missing_stages]
     end
 
-    [preexisting_stages, missing_stages]
-  end
+    def create_stage_with_group(template_stage, deploy_group)
+      stage = Stage.build_clone(template_stage)
+      stage.deploy_groups << deploy_group
+      stage.name = deploy_group.name
+      stage.is_template = false
+      stage.save!
 
-  def self.create_stage_with_group(template_stage, deploy_group)
-    stage = Stage.build_clone(template_stage)
-    stage.deploy_groups << deploy_group
-    stage.name = deploy_group.name
-    stage.is_template = false
-    stage.save!
+      if template_stage.respond_to?(:next_stage_ids) # pipeline plugin was installed
+        template_stage.next_stage_ids << stage.id
+        template_stage.save!
+      end
 
-    if template_stage.respond_to?(:next_stage_ids) # pipeline plugin was installed
-      template_stage.next_stage_ids << stage.id
-      template_stage.save!
+      stage
     end
-
-    stage
   end
 
   def deploy_group_params
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -110,15 +110,16 @@ class Admin::DeployGroupsController < ApplicationController
 
   def merge_stage(stage, template_stage)
     return unless template_stage
-    return if template_stage.deploy_groups.include?(stage)
+    return if template_stage.deploy_groups.include?(stage.deploy_groups.first)
     return if stage.is_template
+    return if stage.deploy_groups.count == 0
     return if stage.deploy_groups.count > 1
 
     template_stage.deploy_groups += stage.deploy_groups
     template_stage.next_stage_ids.delete(stage.id)
     template_stage.save!
 
-    stage.soft_delete
+    stage.soft_delete!
   end
 
   def stages_for_creation
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -98,6 +98,9 @@ class Admin::DeployGroupsController < ApplicationController
       template_stage.next_stage_ids.delete(stage.id)
       template_stage.save!
 
+      # these interfere with views as they don't respect soft_deletion.
+      DeployGroupsStage.where(stage: stage).delete_all
+
       stage.soft_delete
     end
 
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -120,6 +120,10 @@ class Admin::DeployGroupsController < ApplicationController
     template_stage.save!
 
     stage.soft_delete!
+
+    if !stage.reload.deleted?
+      Rails.logger.warn("Soft delete of stage #{stage.id} failed")
+    end
   end
 
   def stages_for_creation
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -118,9 +118,6 @@ class Admin::DeployGroupsController < ApplicationController
     template_stage.next_stage_ids.delete(stage.id)
     template_stage.save!
 
-    # these don't respect soft_deletion (yet).
-    DeployGroupsStage.where(stage: stage).delete_all
-
     stage.soft_delete
   end
 
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -90,6 +90,23 @@ class Admin::DeployGroupsController < ApplicationController
     redirect_to [:admin, deploy_group]
   end
 
+  def merge_all_stages
+    preexisting_stages, _ = stages_for_creation
+    template_stages = deploy_group.environment.template_stages.all
+
+    preexisting_stages.each do |stage|
+      template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
+
+      template_stage.deploy_groups += stage.deploy_groups
+      template_stage.next_stage_ids.delete(stage.id)
+      template_stage.save!
+
+      stage.destroy
+    end
+
+    redirect_to [:admin, deploy_group]
+  end
+
   private
 
   # returns a list of stages already created and list of stages to create (through their template stages)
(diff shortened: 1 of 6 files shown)


app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -90,6 +90,23 @@ def create_all_stages
...
+    template_stages = deploy_group.environment.template_stages.all
+
+    preexisting_stages.each do |stage|
+      template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
+
+      template_stage.deploy_groups += stage.deploy_groups
+      template_stage.next_stage_ids.delete(stage.id)
+      template_stage.save!
+
+      stage.destroy
```
grosser: `soft_delete` ?

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -40,7 +40,16 @@
                     class: "btn btn-default"
         %>
         <%= link_to "Deploy All Projects", deploy_all_admin_deploy_group_path(@deploy_group), class: "btn btn-default", data: { method: "post" } %>
-      <% end %>
+        <%= link_to "Merge All Stages",
```
grosser: sounds super dangerous and is not what is happening ... it finds stages that are only deploying to this deploy group and merges them into the template ...

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -232,5 +232,57 @@
         refute @controller.instance_variable_get(:@preexisting_stages).empty?
       end
     end
+
+    describe "#merge_all_stages" do
+
```
grosser: avoid unnecessary newline

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -232,5 +232,57 @@
...
+        end
+      end
+
+      describe "with a create stage" do
+
+        let(:env) { environments(:staging) }
+        let(:deploy_group) { DeployGroup.create!(name: 'Pod 101', environment: env) }
+        let(:template_stage) { stages(:test_staging) }
+
+        let (:stage) {
```
grosser: multiline = do/end

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -232,5 +232,57 @@
...
+          post :create_all_stages, params: {id: deploy_group}
```
grosser: prefer to crete state instead of calling a different action ... might have weird side-effects and makes debugging hard

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -232,5 +232,57 @@
...
+        it "adds the deploy group to the template stage" do
+          refute template_stage.deploy_groups.include?(deploy_group)
+
+          post :merge_all_stages, params: {id: deploy_group}
+
+          assert template_stage.deploy_groups.include?(deploy_group)
+        end
+
+      end
+
```
grosser: more newlines

test/controllers/admin/deploy_groups_controller_test.rb
dragonfax: Made it a re-usable class-method on the controller.

dragonfax:
@grosser

dragonfax:
- [ ] well this deletes the template stages. hmm

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -82,18 +82,43 @@ def create_all_stages_preview
...
+    self.class.create_all_stages(deploy_group)
+
+    redirect_to [:admin, deploy_group]
+  end
+
+  def merge_all_stages
+    preexisting_stages, = stages_for_creation
+    template_stages = deploy_group.environment.template_stages.all
+
+    preexisting_stages.each do |stage|
```
grosser: this also includes stages that have multiple deploy groups ... that might be an issue that should be addresses in stages_for_creation

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -40,7 +40,16 @@
                     class: "btn btn-default"
         %>
         <%= link_to "Deploy All Projects", deploy_all_admin_deploy_group_path(@deploy_group), class: "btn btn-default", data: { method: "post" } %>
-      <% end %>
+        <%= link_to "Merge These Stages",
+                    merge_all_stages_admin_deploy_group_path(@deploy_group),
+                    class: "btn btn-default",
+                    data: {
+                      method: "post",
+                      confirm: "All stages listed will be deleted, and this Deploy Group will be added to the template stage for each project.\n\n" +
```
grosser: The UI only shows "all stages that use this deploy group" ... and this button only affects "stages that have exactly this deploy group and are on a project that uses include_new_deploy_groups"

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -82,50 +82,88 @@ def create_all_stages_preview
...
+    return if template_stage.deploy_groups.include?(stage)
+    return if stage.is_template
+    return if stage.deploy_groups.count > 1
+
+    template_stage.deploy_groups += stage.deploy_groups
+    template_stage.next_stage_ids.delete(stage.id)
+    template_stage.save!
+
+    # these don't respect soft_deletion (yet).
+    DeployGroupsStage.where(stage: stage).delete_all
```
dragonfax: @grosser I just delete these as we won't need them anymore. The alternative is to add them to soft_delete, which will require, at the least, a database migration. What do you think?

grosser: should be an `after_soft_delete` or `before_soft_delete` hook then to be consistent no matter how things get deleted

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -232,5 +232,99 @@
...
+      it "merges and soft-deletes all non-template stages" do
+        assert_equal 3, deploy_group.stages.count
+
+        post :merge_all_stages, params: {id: deploy_group}
+
+        # should have only template-stages now (all others were soft-deleted)
+        stages = deploy_group.stages.where(is_template: false)
+        Rails.logger.warn("stages after soft delete #{ stages.map { |s| [s.reload.project.name, s.name, s.deleted?] }.inspect }")
+        assert_empty stages
+      end
```
dragonfax: @grosser I've isolated the soft deletion issue with this test. `soft-delete` is definitely called on all 3 stages. Only one stage ever gets soft-delete, all other calls to soft-delete! become NOOP.

This is what the log looks like when soft-delete is called. The last line is my rails log line to verify it was soft deleted properly.

```
   (0.4ms)  SAVEPOINT active_record_1
   (0.5ms)  RELEASE SAVEPOINT active_record_1
  Stage Load (0.8ms)  SELECT  `stages`.* FROM `stages` WHERE `stages`.`id` = 685640165 LIMIT 1
Soft delete of stage 685640165 failed
```

dragonfax: I tried it with and without the callbacks. No difference.

grosser:
found the real issue ... stage.project.stages was old and so verify_not_part_of_pipeline failed ... which is part of a plugin, so it not obvious where it came from ... also a bug in soft_deletion is that it did not blow up when a callback failed in `soft_delete!` which would have show that issue instantly ... fix coming, but for this PR reloading the stage.project.stages is good enouh

dragonfax:
@grosser Tests fixed, rubocop, rebased against master, ready to go.

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -82,50 +82,87 @@ def create_all_stages_preview
...
+    def stages_for_creation(deploy_group)
+      environment = deploy_group.environment
+      template_stages = environment.template_stages.all
+      deploy_group_stages = deploy_group.stages.all
+
+      preexisting_stages = []
+      missing_stages = []
+      Project.where(include_new_deploy_groups: true).each do |project|
+        template_stage = template_stages.detect { |ts| ts.project_id == project.id }
+        deploy_group_stage = deploy_group_stages.detect { |dgs| dgs.project.id == project.id }
```
grosser: prefer `project_id` to avoid 1 lookup

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -82,50 +82,87 @@ def create_all_stages_preview
...
+        end
       end
+
+      [preexisting_stages, missing_stages]
     end
 
-    [preexisting_stages, missing_stages]
-  end
+    def create_stage_with_group(template_stage, deploy_group)
+      stage = Stage.build_clone(template_stage)
```
grosser: maybe cloning a stage should not make it a template to avoid having everything be a template for any new project ? (since ppl will just clone+edit stages)

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -40,7 +40,16 @@
                     class: "btn btn-default"
         %>
         <%= link_to "Deploy All Projects", deploy_all_admin_deploy_group_path(@deploy_group), class: "btn btn-default", data: { method: "post" } %>
-      <% end %>
+        <%= link_to "Merge Elligable Stages",
```
grosser: can we use the copied_from_id here and then do `merge 4 created stages` ? (4 = stages with copied_from_id)

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -232,5 +232,108 @@
...
+          refute DeployGroupsStage.where(stage_id: template_stage.id, deploy_group_id: deploy_group.id).first
+
+          post :merge_all_stages, params: {id: deploy_group}
+
+          assert DeployGroupsStage.where(stage_id: template_stage.id, deploy_group_id: deploy_group.id).first
+        end
+      end
+    end
+
+    describe "with multiple stages to remove" do
```
grosser: needs to be inside the merge_all_stages describe

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -232,5 +232,108 @@
...
+      let(:env) { environments(:staging) }
+      let(:template_deploy_group) { deploy_groups(:pod100) } # needed so the templaste stages have an environment.
+      let(:deploy_group) { DeployGroup.create!(name: 'Pod 101', environment: env) }
+
+      before do
+        Project.any_instance.stubs(:valid_repository_url)
+        # create 2 new template stages, and remember we still have a 3rd from the default fixtures.
+
+        project1 = Project.create!(
```
grosser: we have fixtures for multiple stages ... can we reuse them and not create new projects ?

test/controllers/admin/deploy_groups_controller_test.rb
```diff
@@ -232,5 +232,108 @@
...
+            name: "foo1",
+            include_new_deploy_groups: true,
+            permalink: "foo1",
+            repository_url: "https://github.com/samson-test-org/example-project.git"
+        )
+        Stage.create!(name: "foo1 tstage", project: project1, is_template: true, deploy_groups: [template_deploy_group])
+
+        project2 = Project.create!(
+            name: "foo2",
```
grosser: 2-space indent

app/controllers/admin/deploy_groups_controller.rb
dragonfax: 3 lines down is `stage.is_template = false`.

grosser: What I'm suggesting is that cloning a stage should not clone the is_template value since this means every regular clone will copy it and that most new stages will have `is_template` which means we have lots of templates by accident

dragonfax:
@grosser, all issues addressed.

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -76,56 +76,89 @@ def deploy_all
...
   # Note: you can call this multiple times, and it will create missing stages, but no redundant stages.
   def create_all_stages
-    _, missing_stages = stages_for_creation
-    missing_stages.each do |template_stage|
-      create_stage_with_group(template_stage)
+    stages_created = self.class.create_all_stages(deploy_group)
+
+    flash[:success] = "Created #{stages_created.length} Stages"
+
+    redirect_to [:admin, deploy_group]
```
grosser: prefer rails default notice support: 
 `redirect_to [:admin, deploy_group], notice: "Created #{stages_created.length} Stages"`

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -76,56 +76,89 @@ def deploy_all
...
-        preexisting_stages << deploy_group_stage
-      elsif template_stage
-        missing_stages << template_stage
-      end
-    end
+  def merge_stage(stage)
+    template_stage = stage.template_stage
+
+    return unless template_stage
+    return if template_stage.deploy_groups.include?(stage.deploy_groups.first)
```
grosser: should still be deleted then since it is a duplicate ?

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -76,56 +76,89 @@ def deploy_all
...
+    return if stage.is_template
```
grosser: should still be deleted then ?

grosser: would mean someone manually turned it into a template ... seems like a rare issue ...

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -76,56 +76,89 @@ def deploy_all
...
-      template_stage.save!
+      [preexisting_stages, missing_stages]
     end
 
-    stage
+    def create_stage_with_group(template_stage, deploy_group)
+      stage = Stage.build_clone(template_stage)
+      stage.deploy_groups << deploy_group
+      stage.name = deploy_group.name
+      stage.is_template = false
```
grosser: make build_clone not set template ?

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -40,7 +40,17 @@
                     class: "btn btn-default"
         %>
         <%= link_to "Deploy All Projects", deploy_all_admin_deploy_group_path(@deploy_group), class: "btn btn-default", data: { method: "post" } %>
-      <% end %>
+        <% num_elligable_merge_stages = @deploy_group.stages.where.not(template_stage_id: nil).count %>
+        <%= link_to "Merge #{num_elligable_merge_stages} Elligable Stages",
```
grosser: `Merge copied stages` or `Merge cloned stages` ... also update description below to reflect the logic that checks if it was copied from a template

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -40,7 +40,17 @@
                     class: "btn btn-default"
         %>
         <%= link_to "Deploy All Projects", deploy_all_admin_deploy_group_path(@deploy_group), class: "btn btn-default", data: { method: "post" } %>
-      <% end %>
+        <% num_elligable_merge_stages = @deploy_group.stages.where.not(template_stage_id: nil).count %>
```
grosser: `.where.not(template_stage_id: nil)` was reused twice, maybe turn it into a scope like `not_cloned` ?

grosser:
I guess clone should also not copy the copied_from_template_id ... that could be confusing ...

app/controllers/admin/deploy_groups_controller.rb
dragonfax: This is just a sanity check. But if it has been turned into a template, other stages could reference it, want to merge into it (for some reason).

app/controllers/admin/deploy_groups_controller.rb
dragonfax: `build_clone` is more abstract, not written for pod-o-matic.

grosser:
yes, but it should not build bad stages

dragonfax:
There is nothing bad about those stages.

grosser:
nvm then ... I felt like it's a bit strange to have all stages be templates
...

dragonfax:
@grosser, we've got green. did I miss addressing anything?

app/models/deploy_group.rb
```diff
@@ -36,6 +36,10 @@ def natural_order
     name.split(/(\d+)/).each_with_index.map { |x, i| i.odd? ? x.to_i : x }
   end
 
+  def cloned_stages
```
grosser: better to make this a scope on stage `scope :cloned, -> { where.not(template_stage_id: nil) }` then use group.stages.cloned

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -40,7 +40,16 @@
                     class: "btn btn-default"
         %>
         <%= link_to "Deploy All Projects", deploy_all_admin_deploy_group_path(@deploy_group), class: "btn btn-default", data: { method: "post" } %>
-      <% end %>
+        <%= link_to "Merge #{@deploy_group.cloned_stages.count} Cloned Stages",
+                    merge_all_stages_admin_deploy_group_path(@deploy_group),
+                    class: "btn btn-default",
+                    data: {
+                      method: "post",
+                      confirm: "Stages will be deleted. Before deleting, each stage will have its Deploy Group added to the stage it was copied from. Stages affected are those cloned from a template stage, have exactly this deploy group, no other deploy groups, and are on a project that uses include_new_deploy_groups.\n\n" +
```
grosser: 'have exactly this deploy group, no other deploy groups,' is a bit redundant ...

dragonfax:
@grosser

PR #1312, opened 2016-09-23
PAAS-286 mark pod as unscheduleable when it has abnormal events

grosser:
@jonmoter @irwaters

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -2,8 +2,9 @@
 module Kubernetes
   module Api
     class Pod
-      def initialize(api_pod)
+      def initialize(api_pod, client: nil)
         @pod = api_pod
+        @client = client
       end
 
       def name
@@ -46,6 +47,27 @@ module Kubernetes
 
       def containers
         @pod.spec.containers
+      end
+
+      def logs(container)
+        @client.get_pod_log(name, namespace, previous: restarted?, container: container)
+      rescue KubeException
+        begin
+          @client.get_pod_log(name, namespace, previous: !restarted?, container: container)
+        rescue KubeException
+          nil
+        end
+      end
+
+      def unschedulable?
+        events.any? { |e| e.type != "Normal" }
+      end
+
+      def events
+        @events ||= @client.get_events(
+          namespace: namespace,
+          field_selector: "involvedObject.name=#{name}"
+        )
       end
 
       private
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -88,76 +89,69 @@ def wait_for_resources_to_complete(release, release_docs)
...
+          pod.client = client
+          pod
+        end
+      end
     end
 
-    def show_failure_cause(release)
-      bad_pods(release).each do |pod, client, deploy_group|
+    def show_failure_cause(release, release_docs)
+      pod_statuses(release, release_docs).reject(&:live).select(&:pod).each do |status|
```
grosser: previously this was duplicating the logic from release_statuses

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -321,12 +317,12 @@ def deploy(release_docs)
...
-    def deploy_to_cluster(release, deploys)
-      deploy(deploys)
-      successful = wait_for_resources_to_complete(release, deploys)
+    def deploy_to_cluster(release, release_docs)
+      deploy(release_docs)
+      successful = wait_for_resources_to_complete(release, release_docs)
       unless successful
-        show_failure_cause(release)
-        rollback(deploys)
+        show_failure_cause(release, release_docs)
```
grosser: previously this showed all failed deploys ... which was never a real issue since if something failed the previous deploy would have failed ... , but this should be more correct

grosser:
screen shot 2016-09-23 at 11 40 37 am

plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -48,6 +50,27 @@ def containers
         @pod.spec.containers
       end
 
+      def logs(container)
+        client.get_pod_log(name, namespace, previous: restarted?, container: container)
```
jonmoter: This would crash if you tried calling this without first calling `client=`, right?  If you have methods that expect there to be a valid client object, you should probably add it as a parameter to the constructor.

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -88,76 +89,69 @@ def wait_for_resources_to_complete(release, release_docs)
...
 
+    # user clicked stop button in UI
     def stopped?
       if @stopped
         @output.puts "STOPPED"
         true
       end
     end
 
+    # TODO: cleanup ... a bit ugly with all these arrays
```
jonmoter: If you have 4-tuples of values, we should probably just create an object to encapsulate the data.

PR #1314, opened 2016-09-24
No more constant Struct in tests

neroleung:
Use `Sawyer::Resource` instead.

/cc @grosser 

### Risks
- Level: None


grosser:
:+1: :heart: thx for the followup fix :)

PR #1316, opened 2016-09-27
Save source cloning stage

dragonfax:
### References

* JIRA https://zendesk.atlassian.net/browse/TOOLS-873

### Risks

None

### DIFF
--- a/db/migrate/20160927223306_save_source_template_to_clone_stage.rb
+++ b/db/migrate/20160927223306_save_source_template_to_clone_stage.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 class SaveSourceTemplateToCloneStage < ActiveRecord::Migration[5.0]
   def change
     add_column :stages, :template_stage_id, :integer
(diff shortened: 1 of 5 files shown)


db/migrate/20160927223306_save_source_template_to_clone_stage.rb
```diff
@@ -0,0 +1,5 @@
+class SaveSourceTemplateToCloneStage < ActiveRecord::Migration[5.0]
+  def change
+    add_column :stages, :template_stage_id, :integer
```
grosser: add an index since we query with it ?

grosser:
👍  ... the merge logic can then use that column ?

db/migrate/20160927223306_save_source_template_to_clone_stage.rb
dragonfax: I'm not sure we'll use it, but its harmless to add one.

dragonfax: This really is just to show the OPS person where the clone came from. As a sanity check. We'll show it in a couple of the view templates showing stages.

PR #1318, opened 2016-09-28
PAAS-425 do not override kube-system namespace, it is unique for every cluster

grosser:
@jonmoter


grosser:
need to rework this ... we use namespace in a few places when querying for jobs/services etc ... so the release_doc has to change ... which will mean the release_doc no longer matches the deploy group which is kind of weird ...

grosser:
rebased on the latest cleanup PR ... that should do the trick ...
it will still be a little weird that the namespace does not match, but thats better then having an extra deploy group for all clusters ...

jonmoter:
Could you rebase this after you merge #1337?

grosser:
done

PR #1322, opened 2016-09-29
very few projects need binary builder and having it active for new stages is confusing

grosser:
@zendesk/samson @henders 

every new/existing stage has:

screen shot 2016-09-29 at 9 34 39 am

which every time makes me think "what is this ... do we use/need that" ... lets remove this mental overhead and only enable it when we need it ...

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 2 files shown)


grosser:
@henders

grosser:
@henders can you check this out ?

grosser:
@zendesk/samson I can haz review ?

PR #1324, opened 2016-09-29
rename confirm to more meaningful review before deploying

grosser:
@zendesk/samson 

"Confirm" sounds like useless overhead ... "Review" is closer to what it is used for ... also making the column name more descriptive ... sounds good ?

screen shot 2016-09-29 at 10 11 31 am

screen shot 2016-09-29 at 10 12 01 am


grosser:
I liked "Review" more since the page is the "Review" page where you can see
what you deploy

grosser:
hmm I guess what the page is doing is "Preview" ... since it's not already done ... but that sounds strange too ... maybe I'll just make the column ...

PR #1329, opened 2016-09-30
bump rubocop

grosser:
@zendesk/samson


plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -20,7 +20,7 @@ def live?
       end
 
       def restarted?
-        @pod.status.containerStatuses.try(:any?) { |s| s.restartCount != 0 }
+        @pod.status.containerStatuses.try(:any?) { |s| s.restartCount.nonzero? }
```
sandlerr: this one is probably `positive?` too

grosser: yep ... I can has +1 ?

app/presenters/user_csv_presenter.rb
```diff
@@ -16,7 +16,7 @@ def self.to_csv(
     inherited: false, deleted: false, project_id: nil, user_id: nil, datetime: (Time.now.strftime "%Y%m%d_%H%M")
   )
     inherited = true if project_id || user_id
-    users = (deleted || user_id) ? User.unscoped : User
+    users = (deleted || user_id ? User.unscoped : User)
```
sandlerr: braces probably not needed at all now

plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -113,7 +113,7 @@ def create_container_options
 
     # Mount a cache directory for sharing .m2, .ivy2, .bundler directories between build containers.
     api_version_major, api_version_minor = docker_api_version.scan(/(\d+)\.(\d+)/).flatten.map(&:to_i)
-    if api_version_major == 0 || (api_version_major == 1 && api_version_minor <= 14)
+    if api_version_major.zero? || (api_version_major == 1 && api_version_minor <= 14)
```
sandlerr: I think `< 1` makes sense in this case

plugins/new_relic/test/samson_new_relic/api_test.rb
```diff
@@ -37,7 +37,7 @@ def stub_metric_api(field, value)
 
   describe '.metrics' do
     before do
-      applications = self.applications.map { |a| [a.fetch('name'), SamsonNewRelic::Api::Application.new(a)] }.to_h
+      applications = applications().map { |a| [a.fetch('name'), SamsonNewRelic::Api::Application.new(a)] }.to_h
```
sandlerr: can the local be renamed?

sandlerr:
:+1: on 🍏

app/presenters/user_csv_presenter.rb
grosser: I like these since it makes it obvious that we assign something / that the right part could be extracted into a method

plugins/docker_binary_builder/app/models/binary_builder.rb
grosser: it's a zero check ... cannot just check `.positive?` since it allows 0 and 1 with a special minor version ...

PR #1331, opened 2016-10-03
if a vault instance is down, show an error page

grosser:


app/controllers/admin/secrets_controller.rb
```diff
@@ -13,9 +13,13 @@ class Admin::SecretsController < ApplicationController
...
   def index
-    @secret_keys = SecretStorage.keys
-    if query = params.dig(:search, :query).presence
-      @secret_keys.select! { |s| s.include?(query) }
+    begin
+      @secret_keys = SecretStorage.keys
+      if query = params.dig(:search, :query).presence
+        @secret_keys.select! { |s| s.include?(query) }
+      end
+    rescue RuntimeError => e
```
grosser: the begin/end can be removed since this is the body of the whole method

grosser: ... ideally rescue more specific errors ... but I guess this works ...

app/controllers/admin/secrets_controller.rb
```diff
@@ -74,6 +78,11 @@ def failure_response(message)
     render :edit
   end
 
+  def failure(message)
+    flash[:error] = message
+    render :fail
```
grosser: `render html: "<h1>bla</h1>"` might work too ...

app/views/admin/secrets/fail.html.erb
```diff
@@ -0,0 +1 @@
+<h1>Secret Backend is Down!!</h1>
```
grosser: we don't really know if it's down or if something else weird happened ... `Error reading secrets: <error>`

lib/samson/secrets/hashicorp_vault_backend.rb
```diff
@@ -5,20 +5,26 @@ module Secrets
...
 
-          result = result.to_h
-          result = result.merge(result.delete(:data))
-          result[:value] = result.delete(:vault)
-          result
+            result = result.to_h
+            result = result.merge(result.delete(:data))
+            result[:value] = result.delete(:vault)
+            result
+          rescue Vault::HTTPConnectionError => e
```
grosser: no need to wrap the whole method

grosser: extract this into a method ... `wrap_vault_errors` ... also raise a specific exception that then can have a nice message / can be rescues in the frontend

app/controllers/admin/secrets_controller.rb
irwaters: so the vault client will raise Vault::HTTPConnectionError.  I didn't really wana catch that in the controller in case the backend is DB.

grosser: since you are rescuing RuntimeError ... that would get everything :)

... maybe make it `StandardError` ...

app/controllers/admin/secrets_controller.rb
irwaters: doesn't get us any style.

grosser: sad ... possibly with `render inline: "test"`

irwaters: same :(

app/controllers/admin/secrets_controller.rb
irwaters: using a custom exception now.

grosser:
... `layout: true` ? :D
... if not then just leave ...

irwaters:
that did the business.

app/controllers/admin/secrets_controller.rb
```diff
@@ -17,6 +17,9 @@ def index
     if query = params.dig(:search, :query).presence
       @secret_keys.select! { |s| s.include?(query) }
     end
+    rescue Samson::Secrets::BackendError => e
```
grosser: need an outdent

app/controllers/admin/secrets_controller.rb
```diff
@@ -17,6 +17,9 @@ def index
...
+      #failure e.message
```
grosser: remove comment ?

lib/samson/secrets/hashicorp_vault_backend.rb
```diff
@@ -2,6 +2,12 @@
 
 module Samson
   module Secrets
+    class BackendError < StandardError
+      def initialize(msg)
```
grosser: don't need that initialize ?

lib/samson/secrets/hashicorp_vault_backend.rb
```diff
@@ -18,6 +24,8 @@ def read(key)
           result = result.to_h
           result = result.merge(result.delete(:data))
           result[:value] = result.delete(:vault)
+          rescue Vault::HTTPConnectionError => e
+            wrap_vault_errors(e.message)
           result
```
grosser: outdent and result should be in the happy block

lib/samson/secrets/hashicorp_vault_backend.rb
```diff
@@ -30,25 +38,33 @@ def read_multi(keys)
...
 
         def write(key, data)
-          vault_client.write(
-            vault_path(key),
-            vault: data.fetch(:value),
-            visible: data.fetch(:visible),
-            comment: data.fetch(:comment),
-            creator_id: data.fetch(:user_id)
-          )
+          begin
```
grosser: remove begin/end

lib/samson/secrets/hashicorp_vault_backend.rb
```diff
@@ -30,25 +38,33 @@ def read_multi(keys)
...
           vault_client.delete(vault_path(key))
         end
 
         def keys
-          keys = vault_client.list(VAULT_SECRET_BACKEND + SAMSON_SECRET_NAMESPACE)
-          keys = keys_recursive(keys)
-          keys.uniq! # we read from multiple backends that might have the same keys
-          keys.map! do |secret_path|
-            convert_path(secret_path, :decode) # FIXME: ideally only decode the key(#4) part
+          begin
```
grosser: remove begin/end

lib/samson/secrets/hashicorp_vault_backend.rb
```diff
@@ -59,6 +75,10 @@ def vault_client
           @vault_client ||= VaultClient.new
         end
 
+        def wrap_vault_errors(message)
+          raise Samson::Secrets::BackendError.new("Vault backend is down!! " + message)
```
grosser: maybe something more vague like "Error talking to vault backend: #{message}" so we don't claim something that is wrong and lead ops onto a wild goose chase

app/controllers/admin/secrets_controller.rb
```diff
@@ -17,6 +17,8 @@ def index
     if query = params.dig(:search, :query).presence
       @secret_keys.select! { |s| s.include?(query) }
     end
+  rescue Samson::Secrets::BackendError => e
+    render html: "<div class=\"label-error\">#{e.message}</div>".html_safe, layout: true
```
grosser: reuse flash messages ?

```
flash.now[:error] = "XXX"
render html: "", layout: true
```

grosser:
👍  with flash and without new css

PR #1332, opened 2016-10-04
stop showing stage from in new project, it is confusing

grosser:
@zendesk/samson 

... so many fields ... and getting them wrong means your project does not get saved ...

.. confirmed working in staging


grosser:
screen shot 2016-10-04 at 12 07 30 pm

PR #1338, opened 2016-10-05
make all resources equal

grosser:
... last step in making it trivial to support adding any new resource
... plus some more random cleanup

@jonmoter @irwaters

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -48,7 +48,6 @@ module Kubernetes
         @output.puts "Now deploying other roles ..." if deploys.any?
       end
       if deploys.any?
-        ensure_service(deploys)
         return false unless deploy_to_cluster(release, deploys)
       end
       true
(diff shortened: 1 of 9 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -335,15 +334,6 @@ def deploy_to_cluster(release, release_docs)
       successful
     end
 
-    # Create the service or report it's status
-    def ensure_service(release_docs)
```
grosser: atm we were playing favorites ... but I think we should not touch the template ordering ...

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -24,39 +24,26 @@ def build
...
+      resources.each_with_index do |resource, i|
+        resource.revert(@previous_deploy[i])
       end
     end
 
     # run on unsaved mock ReleaseDoc to test template and secrets before we save or create a build
     def verify_template
-      config = primary_resource(parsed_config_file.elements)
-      template = Kubernetes::ResourceTemplate.new(self, config)
+      primary_config = raw_template.detect { |e| Kubernetes::RoleConfigFile::PRIMARY.include?(e.fetch(:kind)) }
```
grosser: the RoleVerifier already does that
```
SUPPORTED_KINDS = [
      ['Deployment'],
      ['DaemonSet'],
      ['Deployment', 'Service'],
      ['Job']
    ].freeze
```

PR #1340, opened 2016-10-06
support ConfigMap

grosser:
adding new resources is now super easy :D 

@jonmoter @irwaters 

TODO: test deploy ... so many issues :/


grosser:
TODO: do we wnt to make the configmap resourcename configureable ? ... should it reuse the base resource_name ?

grosser:
test deploy finally semi-worked ... one node was borked
 - fixed a bug which made kube-system not fetch bugs
 - show resource errors when it fails
 - do not fail when rollback fails

PR #1343, opened 2016-10-07
how thread counts in logs for debugging

grosser:
@jonmoter 

assuming this is causing the restarts https://zendesk.airbrake.io/projects/95346/groups/1764407700877681738
somehow we are leaking threads ... this should show us what actions do it ...


lib/samson/logging.rb
```diff
@@ -12,7 +12,7 @@
     # show params for every request
     unwanted_keys = %w[format action controller]
     params = event.payload[:params].reject { |key, _| unwanted_keys.include? key }
-    { params: params }
+    { params: params, thread_count: Thread.current.list.size }
```
jonmoter: Should be `thread.list.size`, right?

grosser:
thread count seems to be hovering around 50 ... maybe will jump over time ... will check when we see another restart

PR #1349, opened 2016-10-08
modify service clusterIp so it stays inside the cluster and we can us…

grosser:
@jonmoter @irwaters 

/fyi @zendesk/paas this is how to deploy a service with static ip to all clusters

screen shot 2016-10-07 at 5 18 56 pm

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160929164311) do
+ActiveRecord::Schema.define(version: 20161007231901) do
 
   create_table "builds", force: :cascade do |t|
     t.integer  "project_id",                                       null: false
@@ -183,6 +183,7 @@ ActiveRecord::Schema.define(version: 20160929164311) do
     t.datetime "updated_at"
     t.string   "config_filepath", limit: 255
     t.string   "config_context",  limit: 255
+    t.string   "ip_prefix"
   end
 
   create_table "kubernetes_deploy_group_roles", force: :cascade do |t|
--- a/plugins/kubernetes/app/views/admin/kubernetes/clusters/_form.html.erb
+++ b/plugins/kubernetes/app/views/admin/kubernetes/clusters/_form.html.erb
@@ -2,26 +2,9 @@
   <%= render 'shared/errors', object: @cluster %>
 
   <fieldset>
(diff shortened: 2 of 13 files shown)


plugins/kubernetes/app/views/admin/kubernetes/clusters/_form.html.erb
```diff
@@ -35,6 +35,14 @@
     </div>
 
     <div class="form-group">
+      <%= form.label :ip_prefix, "IP prefix", class: "col-lg-2 control-label" %>
+      <div class="col-lg-4">
+        <%= form.text_field :ip_prefix, class: "form-control" %>
+      </div>
+      <%= additional_info("will replace service clusterIP") %>
```
jonmoter: Could you describe this a little better? Explain that it's the first 1 to 3 sections of an IPv4 address, and it will replace the ClusterIP if specified.

db/schema.rb
```diff
@@ -183,6 +183,7 @@
     t.datetime "updated_at"
     t.string   "config_filepath", limit: 255
     t.string   "config_context",  limit: 255
+    t.string   "ip_prefix"
```
jonmoter: Please add a validation for this in the `Kubernetes::Cluster` model

grosser: do you mean it should be required ? ... or validate format ?

jonmoter: I mean validate format.

grosser: screen shot 2016-10-10 at 4 45 32 pm

grosser:
updated!

plugins/kubernetes/app/views/kubernetes/roles/_form.html.erb
```diff
@@ -24,14 +24,11 @@
...
-    <div class="form-group">
-      <% pattern = Kubernetes::RoleVerifier::VALID_LABEL_JS %>
-      <%= form.label :resource_name, class: "col-lg-2 control-label" %>
-      <div class="col-lg-4">
-        <%= form.text_field :resource_name, class: "form-control", pattern: pattern %>
-      </div>
-      <%= additional_info "Name for the resource this role will create (service/job), must match #{pattern}" %>
-    </div>
+    <%= form.input :resource_name,
+          pattern: Kubernetes::RoleVerifier::VALID_LABEL_JS,
```
jonmoter: Didn't you remove that constant above?

PR #1352, opened 2016-10-09
unify / dry / simplify form elements

grosser:
@zendesk/samson 

WIP, but please take a look and comment on direction / layout ...
planing to re-vamp most forms to use this and cut down on the html noise and make all forms feel the same / have the same usability (clickable checkbox labels / required fields / html validations etc)

future features: 
 - add html limit for column limits / validations
 - add required so we can indicate required fields

before:
screen shot 2016-10-08 at 7 47 15 pm

after:
screen shot 2016-10-08 at 7 43 38 pm
checkbox label and complete empty row are clickable

TODO: tests

### DIFF
--- a/lib/samson/form_builder.rb
+++ b/lib/samson/form_builder.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+module Samson
+  class FormBuilder < ActionView::Helpers::FormBuilder
+    def input(attribute, as: :text_field, help: false, label: false)
+      content_tag :div, class: 'form-group' do
+        content = label(attribute, label, class: "col-lg-2 control-label")
+        content << content_tag(:div, class: 'col-lg-4') do
+          public_send(as, attribute, class: "form-control")
+        end
+        content << @template.additional_info(help) if help
+        content
+      end
+    end
+
+    private
+
+    def content_tag(*args, &block)
+      @template.content_tag(*args, &block)
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


grosser:
yep

lib/samson/form_builder.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+module Samson
+  class FormBuilder < ActionView::Helpers::FormBuilder
+    def input(attribute, as: :text_field, help: false, label: false)
+      content_tag :div, class: 'form-group' do
+        content = label(attribute, label, class: "col-lg-2 control-label")
```
sandlerr: does this render anything if label is false? Does it need a guard?

grosser: when it is 'false' it renders the humanized version of the attribute xy_z -> "Xy Z"

lib/samson/form_builder.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+module Samson
+  class FormBuilder < ActionView::Helpers::FormBuilder
+    def input(attribute, as: :text_field, help: false, label: false)
```
sandlerr: I wonder if there's a more obvious name for this, something like `labelled_input` maybe?

sandlerr: fair enough, all those options do indicate that this more than just an `<input>`

sandlerr:
:+1: on green

PR #1353, opened 2016-10-10
make failed permalink lookups cheaper and silence stage svg 404 error…

grosser:
…s from logs

followup to https://github.com/zendesk/samson/pull/1350

### DIFF
--- a/app/models/concerns/permalinkable.rb
+++ b/app/models/concerns/permalinkable.rb
@@ -9,14 +9,18 @@ module Permalinkable
   end
 
   module ClassMethods
-    def find_by_param!(param)
-      find_by_permalink!(param)
-    rescue ActiveRecord::RecordNotFound
+    # find by permalink or id
+    def find_by_param(param)
+      param = param.to_s
       if param =~ /^\d+$/
-        find_by_id!(param)
+        where("permalink = ? OR id = ?", param, param).first
       else
-        raise
+        where(permalink: param).first
       end
+    end
+
+    def find_by_param!(param)
+      find_by_param(param) || raise(ActiveRecord::RecordNotFound)
     end
   end
 
--- a/app/controllers/stages_controller.rb
+++ b/app/controllers/stages_controller.rb
@@ -98,7 +98,8 @@ class StagesController < ApplicationController
   end
 
   def find_stage
-    @stage = current_project.stages.find_by_param!(params[:id])
+    return if @stage = current_project.stages.find_by_param(params[:id])
+    badge? ? head(:not_found) : raise(ActiveRecord::RecordNotFound)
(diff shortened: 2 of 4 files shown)


grosser:
@jonmoter

app/models/concerns/permalinkable.rb
```diff
@@ -9,14 +9,13 @@ module Permalinkable
   end
 
   module ClassMethods
+    # find by permalink or id
+    def find_by_param(param)
+      where("permalink = ? OR id = ?", param, param).first
```
jonmoter: If you have an `OR` statement like this, does it still take advantage of indices?

jonmoter:
While we're at it, what do you think of adding some caching for the .svg endpoint?  If that endpoint is getting hit over and over again, it seems like a prime candidate for caching.

app/models/concerns/permalinkable.rb
grosser: seems to work fine:

```
=> EXPLAIN for: SELECT `stages`.* FROM `stages` WHERE `stages`.`deleted_at` IS NULL AND (permalink = 'foo' OR id = 'foo') ORDER BY `stages`.`order` ASC
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
| id | select_type | table  | type | possible_keys | key  | key_len | ref  | rows | Extra                       |
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
|  1 | SIMPLE      | stages | ALL  | PRIMARY       | NULL | NULL    | NULL | 1948 | Using where; Using filesort |
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
1 row in set (0.00 sec)

>> Stage.where("permalink = ?", 'foo').explain
=> EXPLAIN for: SELECT `stages`.* FROM `stages` WHERE `stages`.`deleted_at` IS NULL AND (permalink = 'foo') ORDER BY `stages`.`order` ASC
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
| id | select_type | table  | type | possible_keys | key  | key_len | ref  | rows | Extra                       |
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
|  1 | SIMPLE      | stages | ALL  | NULL          | NULL | NULL    | NULL | 1948 | Using where; Using filesort |
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
1 row in set (0.00 sec)
```

grosser: need to do if/else since non-integers don't work in all databases ...

grosser:
caching is for a different PR ... good to go ?

PR #1354, opened 2016-10-11
reuse more forms

grosser:
@sandlerr


sandlerr:
disallow `input_html` if `block_given?` ?

:+1:

grosser:
good idea!

PR #1355, opened 2016-10-11
more form elements

grosser:
bringing checkbox positioning back to it's roots ...

before:
screen shot 2016-10-10 at 10 01 56 pm

after:

screen shot 2016-10-10 at 10 01 27 pm

@sandlerr 
/cc @zendesk/samson

### DIFF
--- a/app/views/stages/_fields.html.erb
+++ b/app/views/stages/_fields.html.erb
@@ -1,92 +1,34 @@
 <fieldset>
-  <div class="form-group">
-    <%= form.label :name, "Name", class: "col-lg-2 control-label" %>
-    <div class="col-lg-4">
-      <%= form.text_field :name, class: "form-control" %>
-    </div>
-  </div>
-
-  <% if form.object.persisted? %>
-    <div class="form-group">
-      <%= form.label :permalink, "Permalink", class: "col-lg-2 control-label" %>
-      <div class="col-lg-4">
-        <%= form.text_field :permalink, class: "form-control" %>
-      </div>
-    </div>
-  <% end %>
-
-  <div class="form-group">
-    <%= form.label :notify_email_address, "Email addresses", class: "col-lg-2 control-label" %>
-    <div class="col-lg-4">
-      <%= form.text_field :notify_email_address, class: "form-control" %>
-      <span class="help-block">Enter the email addresses that should be notified of new deploys, separated by <code>;</code></span>
-    </div>
-  </div>
+  <%= form.input :name %>
+  <%= form.input :permalink if form.object.persisted? %>
+  <%= form.input :notify_email_address,
+        label: "Email addresses",
+        help: "Enter the email addresses that should be notified of new deploys, separated by <code>;</code>".html_safe
+  %>
 
   <% if DeployGroup.enabled? %>
     <%= render 'deploy_groups/deploy_group_select', form: form %>
   <% else %>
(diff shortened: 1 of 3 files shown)


app/views/stages/_fields.html.erb
```diff
@@ -1,92 +1,34 @@
...
-          <%= form.label :production do %>
-            <%= form.check_box :production %>
-            Does this stage impact production servers?
-          <% end %>
-        </div>
-      </div>
-    </div>
+    <%= form.input :production, as: :check_box,
+          label: 'Impacts production',
+          help: "Deploys to production need to pass a buddy check when BuddyCheck is enabled."
```
sandlerr: should this message be conditional on `BuddyCheck.enabled?` ?

app/views/stages/_fields.html.erb
```diff
@@ -126,16 +68,6 @@
...
-      Update Pull Request after deployment
-    <% end %>
-  </div>
-  <div class="col-lg-6 checkbox col-lg-offset-2">
-    <%= form.label :use_github_deployment_api do %>
-      <%= form.check_box :use_github_deployment_api %>
-      Use GitHub Deployment API [beta]
-    <% end %>
-  </div>
+  <%= form.input :update_github_pull_requests, as: :check_box, label: "Update Pull Request after deployment" %>
```
sandlerr: nitpick: Pull Requests (plural)

lib/samson/form_builder.rb
```diff
@@ -1,18 +1,31 @@
...
+        public_send(as, attribute, input_html)
       end
 
       content_tag :div, class: 'form-group' do
-        content = label(attribute, label, class: "col-lg-2 control-label")
-        content << content_tag(:div, class: 'col-lg-4', &block)
-        content << @template.additional_info(help) if help
-        content
+        if as == :check_box
+          content_tag(:div, class: "col-sm-offset-2 col-sm-10 checkbox") do
```
sandlerr: any idea why its `col-sm` here and `col-lg` below?

app/views/stages/_fields.html.erb
grosser: ... I still want to allow people to set this even if they don't have buddy checks ... as documentation / for later use ... leaving it a bit ambiguous so I don't have to add extra logic :)

lib/samson/form_builder.rb
```diff
@@ -1,18 +1,31 @@
...
+            label(attribute) do
+              block.call.dup << SPACER << label << SPACER << help
```
sandlerr: why is it necessary to `dup`? If label and help are empty the spacers are being included for no reason... this is getting more complicated than I expected.

grosser:
label will never be empty ... adding extra logic to help would also work ...
dup is needed because erb blocks return frozen strings ... need a test for
that :/

grosser:
nvm ... the .dup was needed because I was passing a frozen string in the test ... updated that ...

grosser:
... ah no ... inline erb block riggers it too ... keeping the tests frozen to reproduce it ... also less buggy to dup in case someone uses it with a block that returns a constant

grosser:
also updated spacer logic ... either way seems fine to me and there is no visual difference ...

looks good ?

grosser:
... so far loving the cleanliness for experienced users and that adding extra info is super easy for new users

PR #1361, opened 2016-10-12
Adding jobs index on user_id

rikeoka:
* During the audit, the performance of deploys/search was extremely slow and often timed out with a 404 due to the volume of deploys/jobs we have.
* Adding a jobs table index on user_id (deployer) so the search queries would run faster.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: N/A

### Risks
- Level: Low

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161007231901) do
+ActiveRecord::Schema.define(version: 20161012161509) do
 
   create_table "builds", force: :cascade do |t|
     t.integer  "project_id",                                       null: false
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -164,6 +164,7 @@
 
   add_index "jobs", ["project_id"], name: "index_jobs_on_project_id", using: :btree
   add_index "jobs", ["status"], name: "index_jobs_on_status", length: {"status"=>191}, using: :btree
+  add_index "jobs", ["user_id"], name: "index_jobs_on_user_id", using: :btree
```
rikeoka: When I run the migrations the schema.rb file now shows t.index, but it had a lot of changes, so manually updated the file. 

Should I push the new format schema.rb file?

grosser: It seems to always randomly change ... we can try normalizing it again but that should be a separate PR ... we had so many funky issues around that ... so I just stay with hand-picked updates ....

grosser:
👍

PR #1362, opened 2016-10-12
Revert "Adding jobs index on user_id"

rikeoka:
Reverts zendesk/samson#1361

@grosser It showed green and when I hit merge, it changed status.  Probably was delay with the webhook.


grosser:
:+1: or fix instead ;)

PR #1363, opened 2016-10-12
Add job index on user_id

rikeoka:
* During the audit, the performance of deploys/search was extremely slow and often timed out with a 404 due to the volume of deploys/jobs we have.
* Adding a jobs table index on user_id (deployer) so the search queries would run faster.

Added rubocop fix that #1361 needed and travis had a delayed start and odd test results that resulted in reverting in #1362.  

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: N/A

### Risks
- Level: Low


grosser:
👍

PR #1366, opened 2016-10-13
allow slack to only notify on failure

grosser:
@pswadi-zendesk @liulikun @ben 

 - removing send_slack_webhook_notifications since deiver is a noop if there are not hooks
 - ordering check boxes per deploy workflow order
 - adding browser input validation for url
 - renaming tab so it's more obvious which plugin that is

### DIFF
--- a/plugins/slack_webhooks/app/models/slack_webhook.rb
+++ b/plugins/slack_webhooks/app/models/slack_webhook.rb
@@ -6,6 +6,15 @@ class SlackWebhook < ActiveRecord::Base
 
   scope :for_buddy, -> { where(for_buddy: true) }
 
+  def deliver_for?(deploy_phase, deploy)
+    case deploy_phase
+    when :for_buddy then for_buddy?
+    when :before_deploy then before_deploy?
+    when :after_deploy then after_deploy? && (!only_on_failure? || !deploy.succeeded?)
+    else raise
+    end
+  end
+
   private
 
   def validate_url
--- a/plugins/slack_webhooks/test/models/slack_webhook_test.rb
+++ b/plugins/slack_webhooks/test/models/slack_webhook_test.rb
@@ -4,9 +4,9 @@ require_relative '../test_helper'
 SingleCov.covered!
 
 describe SlackWebhook do
-  let(:webhook) { SlackWebhook.new }
+  let(:webhook) { SlackWebhook.new(after_deploy: false) }
 
-  describe '#validate_url' do
+  describe "#validate_url" do
     it "is valid with a valid url" do
       webhook.webhook_url = 'http://example.com'
       assert_valid webhook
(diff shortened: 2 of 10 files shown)


grosser:
screen shot 2016-10-13 at 12 02 08 pm

plugins/slack_webhooks/app/models/slack_webhook.rb
```diff
@@ -6,6 +6,15 @@ class SlackWebhook < ActiveRecord::Base
 
   scope :for_buddy, -> { where(for_buddy: true) }
 
+  def deliver_for?(deploy_phase, deploy)
+    case deploy_phase
+    when :for_buddy then for_buddy?
+    when :before_deploy then before_deploy?
+    when :after_deploy then after_deploy? && (!only_on_failure? || deploy.success?)
```
sandlerr: should it be !deploy.success?

plugins/slack_webhooks/test/models/slack_webhook_test.rb
```diff
@@ -22,4 +22,52 @@
...
+    it "delivers for for_buddy when for_buddy hooks is enabled" do
+      webhook.for_buddy = true
+      assert webhook.deliver_for?(:for_buddy, deploy)
+    end
+
+    it "fails with unknown hook" do
+      assert_raises { webhook.deliver_for?(:foobar, deploy) }
+    end
+
+    describe "with only_on_failure" do
```
sandlerr: maybe I misunderstood but seems all these test cases are inverted

plugins/slack_webhooks/app/models/slack_webhook.rb
grosser: ah yeah ... got that logic backwards :D

plugins/slack_webhooks/test/models/slack_webhook_test.rb
grosser: 2 wrongs make passing tests too :D

grosser:
updated!

plugins/slack_webhooks/app/models/slack_webhook.rb
```diff
@@ -6,6 +6,15 @@ class SlackWebhook < ActiveRecord::Base
 
   scope :for_buddy, -> { where(for_buddy: true) }
 
+  def deliver_for?(deploy_phase, deploy)
+    case deploy_phase
+    when :for_buddy then for_buddy?
+    when :before_deploy then before_deploy?
+    when :after_deploy then after_deploy? && (!only_on_failure? || !deploy.succeeded?)
+    else raise
```
liulikun: Not sure the behaviour after we `raise` here. Will return `false` a better option?

grosser: I want it to blow up something bad happens ... instead of silently ignoring it ...

liulikun:
👍

PR #1369, opened 2016-10-14
fix index length bug

grosser:
the usual index error :/

`Mysql2::Error: Index column size too large. The maximum column size is 767 bytes.: CREATE UNIQUE INDEX `index_vault_servers_on_name`  ON `vault_servers` (`name`)`

@irwaters @jonmoter

### DIFF
--- a/db/migrate/20161011173842_add_vault_servers.rb
+++ b/db/migrate/20161011173842_add_vault_servers.rb
@@ -9,6 +9,6 @@ class AddVaultServers < ActiveRecord::Migration[5.0]
       t.timestamps
     end
 
-    add_index :vault_servers, :name, unique: true
+    add_index :vault_servers, :name, unique: true, length: { name: 191 }
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -502,7 +502,7 @@ ActiveRecord::Schema.define(version: 20161013175858) do
     t.text     "ca_cert",            limit: 65535
     t.datetime "created_at",                                       null: false
     t.datetime "updated_at",                                       null: false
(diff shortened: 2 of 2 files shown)


grosser:
FYI ... tried to build some checker script that makes sure we don't always repeat this bug ... but itgot out of hand and did  not work for all cases like inline `t.index` calls ... 

```
it "does not have string indices without length" do
    schema = File.read("db/schema.rb")

    tables = schema.scan(/create_table "(.*?)"(.*?)^\s+end/m)
    tables.map! do |name, content|
      [name, content.scan(/t.string +"(.*?)"/).flatten]
    end
    tables = tables.to_h

    puts "T #{tables.inspect}"

    indices = schema.scan(/add_index "(.*?)", \[(.*?)\],(.*)/)
    indices.concat schema.scan(/create_table "(.*?)"(.*?)^\s+end/m)
    puts "I #{indices.inspect}"

    bad = indices.map do |table, columns, options|
      strings = tables[table]
      missing_length = strings.select { |s| columns.include?("\"#{s}\"") && !options.include?("\"#{s}\"=>") }
      [table, missing_length] if missing_length.any?
    end.compact

    # know good
    bad.delete(["builds", ["git_sha"]])

    assert(
      bad.empty?,
      "columns that need length on their index:\n#{bad.map { |t, c| "table: #{t} columns: #{c.join(", ")}"}.join("\n")}"
    )
  end
```

sandlerr:
why aren't the migrations failing in travis?

grosser:
they have their charset fixed or something like that ... idk why we never fixed this issue ... just assuming that it was not trivial :/

grosser:
I can haz +1?  ... we can dicuss alternatives after ... don't want to leave this migration in a bad state ...

PR #1370, opened 2016-10-14
simplify admin resource urls/helpers by using show as edit

grosser:
@sandlerr @irwaters


jonmoter:
No, I don't like this. I want to make more of the admin pages viewable by non-admins, but not editable. Right now, unless you're an admin, you can't even see what the Kubernetes clusters are.

PR #1371, opened 2016-10-14
indicate required fields and enforce them via html

grosser:
less round trips for the user

screen shot 2016-10-14 at 9 50 05 am
screen shot 2016-10-14 at 9 49 58 am

### DIFF
--- a/lib/samson/form_builder.rb
+++ b/lib/samson/form_builder.rb
@@ -11,7 +11,7 @@ module Samson
       raise ArgumentError if block && input_html
 
       input_html ||= {}
-      input_html[:pattern] ||= translate_regex_to_js(pattern)
+      input_html[:pattern] ||= translate_regex_to_js(pattern) if pattern
       input_html[:required] ||= required
 
       label ||= attribute.to_s.humanize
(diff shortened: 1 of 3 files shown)


grosser:
could also autodetect this info from validations, but 1 step at a time ...

lib/samson/form_builder.rb
```diff
@@ -12,9 +12,11 @@ def input(
 
       input_html ||= {}
       input_html[:pattern] ||= translate_regex_to_js(pattern)
-      input_html[:required] ||= required # TODO: mark label somehow
+      input_html[:required] ||= required
 
       label ||= attribute.to_s.humanize
+      label = "* " + label if required
```
grosser: putting it left to keep alignment intact

grosser:
@irwaters @sandlerr 
/cc @zendesk/samson

lib/samson/form_builder.rb
```diff
@@ -50,8 +52,12 @@ def actions(delete: false, &block)
 
     # html regexp input ... allows blank values, does not know case insensitive so use a-zA-Z, does not know \A\z
     # FYI: could also read from validations: _validators[attribute].first.options[:with]
+    # js patterns need to match the full inout "/foo" will not match "/foobar" ... so we enforce \A + \z
     def translate_regex_to_js(pattern)
-      pattern.source.sub('\\A', '^').sub('\\z', '$') if pattern
+      pattern = pattern.source
+      raise ArgumentError, "pattern needs \\A" unless pattern.sub!('\\A', '^')
```
sandlerr: does that mean the ^ and $ have no effect?

sandlerr: should it also raise if the regex ends in `/i` ?

sandlerr: looks like rails knows how to do this :D https://github.com/rails/rails/blob/b67043393b5ed6079989513299fe303ec3bc133b/actionpack/lib/action_dispatch/routing/inspector.rb#L42

grosser: that rails method does a bunch of strange things ... `gsub(/\s/ , '')` etc ...
it means that it enforces \A and \z because patterns will not match when they do not match the whole string ... so enforcing \A\z to make users use complete patterns

grosser: should also raise on `\i` ... but that's another PR

sandlerr:
:+1: on green

PR #1374, opened 2016-10-14
Fix: next stage_ids should always be strings

dragonfax:
Lets standardize on strings for Stage ids in the Stage.next_stage_ids field.

Without this, the Stage Edit form can't show the correct list of "next stages" already selected.

### References

 - Jira link: https://zendesk.atlassian.net/browse/TOOLS-961

### Risks
- Level: None

### DIFF
--- a/test/controllers/stages_controller_test.rb
+++ b/test/controllers/stages_controller_test.rb
@@ -207,6 +207,18 @@ describe StagesController do
         end
       end
 
+      it 'checks the appropriate next_stage_ids checkbox' do
+        next_stage = Stage.create!(name: 'food', project: subject.project)
+        subject.next_stage_ids = [next_stage.id.to_s]
+        subject.save!
+
+        get :edit, params: {project_id: subject.project.to_param, id: subject.to_param }
+
+        checkbox = css_select('#stage_next_stage_ids_').
+            detect { |node| node.attribute('value')&.value == next_stage.id.to_s }
+        assert_equal 'checked', checkbox.attribute('checked').value
+      end
+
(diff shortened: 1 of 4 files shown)


grosser:
how did that test work before when the code was wrong ? ... needs a test that fails without this change ?

dragonfax:
The code works fine whether we' use int or string ids. But the Stage Edit form is where it breaks. As the template only checks for strings. 

I figure its better to standardize on one format in all places, rather than modify the edit form to take multiple formats.

dragonfax:
@zendesk/samson @grosser

grosser:
can you add a `serialize :next_stage_ids, Array # array of string ids` or so to clarify ?
... ideally we would not allow non-strings or unify on integers ... but that's a bunch of work/overhead ...

👍

dragonfax:
@grosser I added a test that emit the issue I was correcting. Please review that as well.

test/controllers/stages_controller_test.rb
```diff
@@ -207,6 +207,18 @@
...
       end
 
+      it 'checks the appropriate next_stage_ids checkbox' do
+        next_stage = Stage.create!(name: 'food', project: subject.project)
+        subject.next_stage_ids = [next_stage.id.to_s]
+        subject.save!
+
+        get :edit, params: {project_id: subject.project.to_param, id: subject.to_param }
+
+        checkbox = css_select('#stage_next_stage_ids_').
```
grosser: uuuh fnuky :D

grosser:
👍

PR #1375, opened 2016-10-14
do not try to deploy when daemonset is stuck in a bad desired pod count state

grosser:
@jonmoter @irwaters 

... and a few other random fixes

### DIFF
--- a/plugins/kubernetes/app/views/admin/kubernetes/clusters/_form.html.erb
+++ b/plugins/kubernetes/app/views/admin/kubernetes/clusters/_form.html.erb
@@ -2,10 +2,10 @@
   <%= render 'shared/errors', object: @cluster %>
 
   <fieldset>
-    <%= form.input :name %>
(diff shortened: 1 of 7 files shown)


plugins/kubernetes/app/views/admin/kubernetes/clusters/_form.html.erb
```diff
@@ -2,10 +2,10 @@
   <%= render 'shared/errors', object: @cluster %>
 
   <fieldset>
-    <%= form.input :name %>
+    <%= form.input :name, required: true %>
     <%= form.input :description %>
-    <%= form.input :config_filepath, label: 'Config File path' %>
-    <%= form.input :config_context, label: 'Context' do %>
+    <%= form.input :config_filepath, label: 'Config File path', pattern: /^\/.*/, help: 'Absolute paths only', required: true %>
```
grosser: html pattern needs to match the full values -> need to add `.*`

PR #1377, opened 2016-10-17
Add support for submodules

omgbob:
:snake:

Update to the git repo model to support submodules.

/cc @zendesk/samson @zendesk/taipan 

### Tasks
 - [x] :+1: from team

### Risks
- Low: additional `git submodule` commands should not cause any issues if repo does not contain submodules

### DIFF
--- a/test/support/git_repo_test_support.rb
+++ b/test/support/git_repo_test_support.rb
@@ -49,6 +49,39 @@ module GitRepoTestHelper
     SHELL
   end
 
+  def create_repo_with_submodule
+    create_submodule_repo
+    create_repo_without_tags
+    execute_on_remote_repo <<-SHELL
+      git submodule add #{submodule_repo_temp_dir} submodule
+      git add .gitmodules
+      git add submodule
+      git commit -m "added submodule"
+    SHELL
+  end
+
+  def submodule_repo_temp_dir
+    @submodule_repo_temp_dir ||= Dir.mktmpdir
+  end
+
+  def execute_on_remote_submodule_repo(cmds)
+    result = `exec 2>&1; set -e; cd #{submodule_repo_temp_dir}; #{cmds}`
+    raise "FAIL: #{result}" unless $?.success?
+    result
+  end
+
+  def create_submodule_repo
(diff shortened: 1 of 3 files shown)


grosser:
can we have a cheap check if submodules exist and then not do this ?

test/support/git_repo_test_support.rb
```diff
@@ -49,6 +49,39 @@ def create_repo_with_second_commit
...
+    raise "FAIL: #{result}" unless $?.success?
+    result
+  end
+
+  def create_submodule_repo
+    execute_on_remote_submodule_repo <<-SHELL
+      git init
+      git config user.email "test@example.com"
+      git config user.name "Test User"
+      git config commit.gpgsign false
```
grosser: reuse existing code ?

grosser:
looks alright ... I'm kind of surprised nobody added that already ... I'd guess they added a 'git submodule init' to their build step ... if someone has expenside/big submodules we might need to support making this optional, but should be good

sandlerr:
this was needed to build a docker image that relies on the submodules being present so there's no opportunity to run `git submodule init`

test/support/git_repo_test_support.rb
omgbob: Actually I might just remove these config lines, since we're not testing anything related to them for the submodule repo.

grosser:
👍

catkins:
Nice one @omgbob + @sandlerr

arjun810:
We'd like to make this optional. What's the best way to do so? Where should we add a setting?

johncip:
One case where you need this to be optional is:

* the submodule refers to a private repository
* you want to (continue to) use github deploy keys, which only support one repo each
* a later part of your deploy process handles granting access to and fetching the submodule's repo

... in which case, you don't need samson to grab it initially, and doing so is actually problematic and requires some messy workarounds, like hostname aliases, or deploying as a regular github user.

grosser:
easiest: global ENV flag
most flexible: project db boolean ... pass into git repo from project.rb
... either would be a fine PR!

PR #1378, opened 2016-10-17
Conditional --recursive flag for git submodule sync based on git version

omgbob:
Conditionally adds the `--recursive` flag to the `git submodule sync` command if the git version is  1.8.1 or higher. It was not an option in older versions.

/cc @zendesk/samson

### Tasks
 - [x] :+1: from team

### Risks
- Low, change to a newly added method that fixes a bug due to an old version of git

### DIFF
--- a/config/initializers/git_info.rb
+++ b/config/initializers/git_info.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 class GitInfo
-  @version = `git --version`.scan(/\d+/).join('.')
+  @version = Gem::Version.new(`git --version`.scan(/\d+/).join('.'))
 
   class << self
     attr_reader :version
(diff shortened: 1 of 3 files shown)


sandlerr:
Added in https://github.com/git/git/blob/master/Documentation/RelNotes/1.8.1.txt#L98 in case anyone is interested :+1:

grosser:
.... detect git version and apply selectively ?
... storing that version would be useful for other things like `git worktree` support

config/initializers/git_info.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+class GitInfo
+  @version = `git --version`.scan(/\d+/).join('.')
```
sandlerr: you could use `Gem::Version.new(`git --version`.scan(...))`

PR #1379, opened 2016-10-17
enforce names for all containers to ease transition from single to mu…

grosser:
…lti-container

@jonmoter @irwaters


grosser:
@jonmoter @irwaters

PR #1381, opened 2016-10-18
write out .env before running a docker build

grosser:
the build will ADD the .env and then source it to get credentials

downside / danger would be that the secrets stay in the container ... either as file or as env variables ... which is bad :(

@prudhvi @jonmoter


grosser:
decided against doing that since global config can be put into the Dockerfile and secrets should not go into the image ... if secrets are needed for building use binary builder ... or somehow make build args work

PR #1387, opened 2016-10-19
enable locking of environments

grosser:
https://zendesk.atlassian.net/wiki/display/INFR/Samson+PSA

@zendesk/samson 

trying hard to avoid extra queries ... only extra queries is loading deploy_groups which can be mass-included ... so that should be fine ...

also fixing a few bugs around global locks not being shown and global warnings making stage locks not show

screen shot 2016-10-18 at 9 04 19 pm

screen shot 2016-10-19 at 3 42 39 pm

switching lock from being global or stage to being global or for a resource ... possible to support different things like deploy-groups ... just environments for now though

centralizing all locking logic in lock.rb

user that created a stage/env/global lock can now still deploy, making it consistent with the previous stage logic


app/assets/stylesheets/projects.scss
```diff
@@ -78,24 +78,10 @@
     color: #333;
   }
 
-  .locked {
```
grosser: using generic lock icon now

test/controllers/locks_controller_test.rb
```diff
@@ -4,17 +4,24 @@
 SingleCov.covered!
 
 describe LocksController do
+  def create(resource = nil, options = {})
```
sandlerr: a bit more verbose but maybe `create_lock_for` ? `create stage` doesn't appear to be related to locks

app/models/lock.rb
```diff
@@ -51,9 +58,54 @@ def delete_in=(seconds)
...
+    end
+  end
+
+  # normally there are very few locks, so we grab them all and filter down to avoid lookups
+  # sorted by priority(warning) and specificity(type)
+  def self.for_resource(resource)
+    matching_locks = all_cached.select do |lock|
+      lock.global? ||
+        lock.resource_equal(resource) ||
+        (resource.class.name == "Stage" && resource.environments.any? { |e| lock.resource_equal(e) })
```
sandlerr: `resource.is_a?(Stage)` or `resource_type == 'Stage'` ?

app/models/lock.rb
```diff
@@ -1,24 +1,31 @@
...
+  belongs_to :resource, polymorphic: true
   belongs_to :user
+  belongs_to :environment
 
   validates :user_id, presence: true
   validates :description, presence: true, if: :warning?
+  validates :resource_type, inclusion: RESOURCE_TYPES
   validate :unique_global_lock, on: :create
 
+  after_save :expire_all_cached
```
sandlerr: does this run `after_destroy` etc? it might be worth giving the cache key a 1 minute expiry anyway?

app/models/job.rb
```diff
@@ -121,7 +121,7 @@ def pid
   private
 
   def validate_globally_unlocked
-    errors.add(:project, 'is locked') if Lock.global.exists?
+    errors.add(:project, 'is locked') if Lock.global.first
```
sandlerr: nitpick: could be `.any?`

sandlerr:
Confluence says:

> If the user clicks Deploy for the affected stage, a verification dialog / window needs to be shown where user has to enter the reason and approver's name who authorized the exception. (In future, we should show list of bugs from a project that need to be selected OR a new bug is automatically created in a specific project meant for tracking DF Exceptions.)

wouldn't that be better than allowing one person globally to deploy? Was just speaking with @acottrell , his team is excepted from the upcoming freeze and will need to be deploying regularly as normal :/

grosser:
we don't destroy, but do soft_delete ... which is an update ... not
expiring to be able to find bugs (which would make other random caches
invalid which is harder to find) ... if at some point we give up and cannot
find a certain bug 1min / 5/10 would be totally fine ...

grosser:
good idea, added

grosser:
Looks good to me ... why would it need to check resource_type of the lock ?

grosser:
yes, need that ...
no idea why it had user_id makes no sense ...

sandlerr:
Email responses aren't showing up inline, I'll just wait for more commits to materialise :)

grosser:
done

grosser:
This is meant for RAs ... not freezes ... atm it is planed to use wanrings
and make the warning say what to do for exceptions, so we won't need that
locking logic ... just wanted to have something that makes the most sense
as default

grosser:
pushed 3 fixes

app/models/lock.rb
sandlerr: I'm suggesting reading the `resouce_type` column of the lock rather than immediately loading the resource, since you mentioned it can't be cached. Or if we're loading the resource and checking its class let's do that using `is_a?` as usual.

grosser: hmm yes ... but it would be `if lock.resource_type == "Environment" && resource.is_a?(Stage) && resource.environments { foo }` ... I'll add that :)

sandlerr: ah, got confused because we're in a  class method here. Maybe extract everything in the `select` block to an instance method?

grosser: added that and the instance method ... much better :)

grosser:
thx for the great review 🎉

apanzerj:
👍

PR #1388, opened 2016-10-20
option in stage to send committers email to jenkins

upcomingnewton:
Add an option in edit page of stage to send email of committers to jenkins. Earlier, this option was in .env file, and this changed adds more granularity.

screen shot 2016-10-20 at 3 36 01 pm
screen shot 2016-10-20 at 3 35 56 pm
screen shot 2016-10-20 at 3 35 21 pm
screen shot 2016-10-20 at 3 35 15 pm

When the option is true:
screen shot 2016-10-20 at 3 41 09 pm

When not true:
screen shot 2016-10-20 at 3 41 32 pm

### Tasks
 - [ ] :+1: from team

### References
 - https://zendesk.atlassian.net/browse/SAMSON-292
 - https://zendesk.atlassian.net/browse/VULCAN-197

### Risks
- Level: Low/Med/High

### DIFF
--- a/db/migrate/20161019034559_add_jenkins_email_committers_to_stage.rb
+++ b/db/migrate/20161019034559_add_jenkins_email_committers_to_stage.rb
@@ -1,0 +1,5 @@
+class AddJenkinsEmailCommittersToStage < ActiveRecord::Migration[5.0]
+  def change
+    add_column :stages, :jenkins_email_committers, :boolean
+  end
+end
--- a/db/migrate/20161019034559_add_jenkins_email_committers_to_stage.rb
+++ b/db/migrate/20161019034559_add_jenkins_email_committers_to_stage.rb
@@ -1,5 +1,6 @@
+# frozen_string_literal: true
 class AddJenkinsEmailCommittersToStage < ActiveRecord::Migration[5.0]
   def change
-    add_column :stages, :jenkins_email_committers, :boolean
+    add_column :stages, :jenkins_email_committers, :boolean, default: false, null: false
   end
 end
--- a/plugins/jenkins/app/views/samson_jenkins/_fields.html.erb
+++ b/plugins/jenkins/app/views/samson_jenkins/_fields.html.erb
@@ -10,7 +10,7 @@
   <div class="col-lg-6 checkbox col-lg-offset-2">
     <%= form.label :jenkins_email_committers do %>
         <%= form.check_box :jenkins_email_committers %>
-           Should Jenkins send email to all committers?
+        Jenkins sends email to committers
     <% end %>
    </div>
 </fieldset>
--- a/plugins/jenkins/app/views/samson_jenkins/_fields.html.erb
+++ b/plugins/jenkins/app/views/samson_jenkins/_fields.html.erb
@@ -7,4 +7,10 @@
       <%= form.text_field :jenkins_job_names, class: "form-control", placeholder: "Jenkins jobs" %>
     </div>
   </div>
+  <div class="col-lg-6 checkbox col-lg-offset-2">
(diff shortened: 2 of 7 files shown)


db/migrate/20161019034559_add_jenkins_email_committers_to_stage.rb
```diff
@@ -0,0 +1,5 @@
+class AddJenkinsEmailCommittersToStage < ActiveRecord::Migration[5.0]
+  def change
+    add_column :stages, :jenkins_email_committers, :boolean
```
grosser: this is a 3-state boolean atm with true/false/nil -> add `default: false, null: false`

db/migrate/20161019034559_add_jenkins_email_committers_to_stage.rb
```diff
@@ -0,0 +1,5 @@
+class AddJenkinsEmailCommittersToStage < ActiveRecord::Migration[5.0]
```
grosser: missing frozen string comment ... try pre-commit gem, it warns you :)

plugins/jenkins/app/views/samson_jenkins/_fields.html.erb
```diff
@@ -7,4 +7,10 @@
       <%= form.text_field :jenkins_job_names, class: "form-control", placeholder: "Jenkins jobs" %>
     </div>
   </div>
+  <div class="col-lg-6 checkbox col-lg-offset-2">
+    <%= form.label :jenkins_email_committers do %>
+        <%= form.check_box :jenkins_email_committers %>
+           Should Jenkins send email to all committers?
```
grosser: no indent

plugins/jenkins/test/models/samson/jenkins_test.rb
```diff
@@ -186,7 +185,8 @@ def stub_add_changeset
       end
 
       it "filters emails by GOOGLE_DOMAIN when JENKINS_NOTIFY_COMMITTERS is set" do
```
grosser: update test name

grosser:
looks good!

zehua:
LGTM, apart from the comments and the red builds.

zehua:
Is there a clear policy on whether data belonging to a plugin should be placed in the core tables, `stages` in this case, or in its own table?

When the plugin only needs to introduce one field, it seems convenient to attach it into the core tables. But when we start to expand and add more columns, it will eventually make the core tables look cluttered.

db/migrate/20161019034559_add_jenkins_email_committers_to_stage.rb
upcomingnewton: Corrected.

db/migrate/20161019034559_add_jenkins_email_committers_to_stage.rb
upcomingnewton: Corrected. I thought it will be auto-magically added by rails.

plugins/jenkins/test/models/samson/jenkins_test.rb
```diff
@@ -170,11 +170,10 @@ def stub_add_changeset
       end
 
       it "includes committer emails when JENKINS_NOTIFY_COMMITTERS is set" do
```
upcomingnewton: Corrected

grosser:
👍

Ngo-The-Trung:
:+1:

grosser:
usually we add plugin data if it is 1 column / simple ... but extra tables
if there are new models / more data

plugin migrations are not run when the plugin is not used, so clutter is
limited to the plugins you actually use

PR #1390, opened 2016-10-20
Better documentation and OAuth Testing Controller

apanzerj:
* Creates a test endpoint for the oauth flow so you can play with the API


grosser:
it actually worked :D 
... so amazing ... not having to think through this whole OAuth mess

grosser:
👍

apanzerj:
👍  LGTM

PR #1391, opened 2016-10-21
make api auth jump through less hoops / correct broken auth strategies

grosser:
... finally got this right ... needs a bunch of integration testing ...

fyi @apanzerj 

 - using bearer without /api -> 400
 - able to use api without content_type or .json
 - do not allow to use oauth via param['access_token'] or param['baerer_token']
 - do not redirect post requests (redirected failed doorman posts)
 - actually verify tokens we pass in to doorman
 - generate correct access tokens
 - better warden error debugging
 - fix up other auth methods to not do weird stuff
 - delete session strategy which was a noop
 - move auth tests into individual tests
 - make unauthorized api query always return json / 404
 ... maybe more ...


grosser:
@zendesk/samson need some keen eyes here ... reworking most of the auth flow ...

app/controllers/api/base_controller.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 class Api::BaseController < ApplicationController
-  skip_before_action :verify_authenticity_token
+  skip_before_action :verify_authenticity_token, if: :using_per_request_auth?
```
grosser: @apanzerj FYI another security hole plugged :)

PR #1392, opened 2016-10-21
SAMSON-293 cache failing requests because they take even longer then successful …

grosser:
@zendesk/samson github failures will be cached for 30s now

/cc @pswadi-zendesk


apanzerj:
Seems good 👍

PR #1393, opened 2016-10-21
render environment and stage locks on the deploy/new page

grosser:
@apanzerj 

/fyi @pswadi-zendesk

screen shot 2016-10-21 at 2 19 48 pm


apanzerj:
👍

PR #1394, opened 2016-10-21
show a github icon on the github warning to make it more disinguishable

grosser:
screen shot 2016-10-21 at 2 31 19 pm

/fyi @pswadi-zendesk 

@apanzerj


grosser:
the others are all icons so they get the style from the text itself

PR #1395, opened 2016-10-21
make controller tests sane by rendering unauthorized when warden fail s instead of being a success with a magic flag

grosser:
@apanzerj @bcolferzd 

/cc @zendesk/samson


apanzerj:
👍

PR #1398, opened 2016-10-22
make select beautiful

grosser:
before:
screen shot 2016-10-21 at 5 14 25 pm

after:
screen shot 2016-10-21 at 5 12 46 pm

@irwaters not you fault, but need a review ;)


apanzerj:
👍

PR #1399, opened 2016-10-22
make long webhook content not blow up the page

grosser:
before:

screen shot 2016-10-20 at 9 23 49 am

after:

screen shot 2016-10-21 at 9 29 21 pm

percentage widths sadly don't work on pre-s since they auto-expand to whatever space is available ... need absolute widths ...

@irwaters

### DIFF
--- a/app/assets/stylesheets/webhooks.scss
+++ b/app/assets/stylesheets/webhooks.scss
@@ -1,0 +1,5 @@
+// Gets super bloated wih a long param or tokens
+// and pre auto-expands without limits ... cannot use % widths
+#latest_response pre {
+  max-width: 1000px
+}
--- a/app/assets/stylesheets/webhooks.scss
+++ b/app/assets/stylesheets/webhooks.scss
@@ -1,5 +1,5 @@
 // Gets super bloated wih a long param or tokens
 // and pre auto-expands without limits ... cannot use % widths
 #latest_response pre {
-  max-width: 1000px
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/webhooks.scss
```diff
@@ -0,0 +1,5 @@
+// Gets super bloated wih a long param or tokens
+// and pre auto-expands without limits ... cannot use % widths
+#latest_response pre {
+  max-width: 1000px
```
grosser: updated to use `$container-lg` which is the `desktop page width`

app/controllers/integrations/base_controller.rb
```diff
@@ -56,13 +58,14 @@ def create_new_release
...
   end
 
-  def deploy_to_stages
-    stages = project.webhook_stages_for(branch, service_type, service_name)
+  def deploy_to_stages(stages)
     deploy_service = DeployService.new(user)
 
-    stages.all? do |stage|
+    return unless stages.all? do |stage|
       deploy_service.deploy!(stage, reference: commit).persisted?
```
jonmoter: Out of curiosity, what happens if you try to deploy to 5 stages.  4 of them persist, but one fails. Do the other 4 actually start?

grosser: yep ... and since it uses `.all?` if you have 5 and the 2nd fails the only 1 stage got started ...

apanzerj:
👍

PR #1400, opened 2016-10-22
use icons

grosser:
just an idea ... do you like it @zendesk/samson ?

text:
screen shot 2016-10-21 at 9 54 55 pm

icons:
screen shot 2016-10-21 at 9 54 00 pm


irwaters:
I like it.  takes up less space in the nav bar/is cleaner etc.

grosser:
@sandlerr thoughts ?

apanzerj:
:+1:

PR #1402, opened 2016-10-24
make 404 and other errors render json ... also show that user is unauthorized

grosser:
@apanzerj 

basically forcing json so that any errors also are rendered as json by rails ... this makes the /api detection hack obsolete since the request will fail before auth happens

also making json unauthorized behave like the html part and actually tell the user that auth failed ...

### DIFF
--- a/app/controllers/unauthorized_controller.rb
+++ b/app/controllers/unauthorized_controller.rb
@@ -18,14 +18,14 @@ class UnauthorizedController < ActionController::Metal
   end
 
   def respond
-    if request.path.start_with?('/api/') || request.format == :json
-      render json: {}, status: 404
-    else
-      respond_to do |format|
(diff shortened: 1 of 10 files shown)


app/controllers/unauthorized_controller.rb
```diff
@@ -18,14 +18,14 @@ def self.call(env)
   end
 
   def respond
-    if request.path.start_with?('/api/') || request.format == :json
-      render json: {}, status: 404
```
grosser: the old 404 was also telling users that it was unauthorized since a genuine 404 error looks different ... also routing error would look different ... since we don't have public/private things in the api getting the 401 does not tell an attacker anything ...

apanzerj:
👍

PR #1403, opened 2016-10-24
don't crash if showing alerts when user not logged in

jonmoter:
Getting a crash if showing an alert to a user that's not already logged in.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Airbrake: https://zendesk.airbrake.io/projects/95346/groups/1803890235894343858

### Risks
- Level: Low


grosser:
👍

PR #1404, opened 2016-10-25
silence warnings

grosser:
@apanzerj @irwaters @jonmoter

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -2,7 +2,7 @@
 class Build < ActiveRecord::Base
   SHA1_REGEX = /\A[0-9a-f]{40}\Z/i
   SHA256_REGEX = /\A(sha256:)?[0-9a-f]{64}\Z/i
-  DIGEST_REGEX = /\A[\w._-]+\/[\w\/_-]+@sha256:[0-9a-f]{64}\Z/i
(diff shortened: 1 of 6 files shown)


jonmoter:
Which warnings are these changes silencing?

grosser:
`ruby -v` warnings ... use of uninitialized instance vars and bad regexes

app/models/build.rb
```diff
@@ -2,7 +2,7 @@
 class Build < ActiveRecord::Base
   SHA1_REGEX = /\A[0-9a-f]{40}\Z/i
   SHA256_REGEX = /\A(sha256:)?[0-9a-f]{64}\Z/i
-  DIGEST_REGEX = /\A[\w._-]+\/[\w\/_-]+@sha256:[0-9a-f]{64}\Z/i
+  DIGEST_REGEX = /\A[\w\.-]+\/[\w\/-]+@sha256:[0-9a-f]{64}\Z/i
```
sandlerr: dots are already literal in this context

grosser: ah yeah ... must have been the _ that triggered the warning I saw ... `.` does not cause a warning ... removed ... even though I think `\.` might be less ambiguous

grosser:
I can haz +1 ?

sandlerr:
so this just fixes variable shadowing and `_` already being part of `\w`? :+1: to those changes

grosser:
yep!

PR #1405, opened 2016-10-25
Remove static_render method

bquorning:
The `static_render` method was added (909f74e71b4483b600db843194cad35bbf36e1fa) only to avoid warnings from Brakeman. Instead of working around the problem, we can just use explicit `render` calls just like Brakeman was trying to tell us.

/cc @zendesk/samson

### Tasks
 - [x] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -140,12 +140,6 @@ module ApplicationHelper
     link_to_delete(path, icon_tag('remove') + ' Delete', class: 'btn btn-danger')
   end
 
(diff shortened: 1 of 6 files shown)


grosser:
nice 👍 ❤️

bquorning:
One test job (`BUNDLE_WITHOUT=mysql:sqlite`) has now failed three times, for 3 different reasons. Are the tests usually this flaky?

bquorning:
May the fourth be with ~~you~~ me. The build is finally green.

PR #1406, opened 2016-10-25
test for doorkeeper auth

grosser:
@apanzerj reading doorkeeper docs made me worried we had no auth for applications ... but using the base-controller fixed that ... just documenting / making sure it stays that way ...


apanzerj:
:+1:

PR #1408, opened 2016-10-25
allow merging users that were created by accident when using differen…

grosser:
https://zendesk.atlassian.net/browse/TOOLS-979

We support multiple auth providers, but we cannot trust all of them to properly verify their emails (github for example allows user to enter a bogus email)
So if we had unique users by email then they could either not sign up with a new provider or we would have to merge them with existing users, which would lead to them being able to take over existing users.

So instead of fixing this mess which would be complicated hard to do securely we add a merge button so admins can sort things out manually.

The user needs to be a super-admin since otherwise an admin could merge themselves into a super-admin user.

screen shot 2016-10-25 at 2 55 25 pm

screen shot 2016-10-25 at 3 08 11 pm
screen shot 2016-10-25 at 2 55 21 pm

@jonmoter @stewi2

### DIFF
--- a/app/controllers/admin/user_merges_controller.rb
+++ b/app/controllers/admin/user_merges_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+class Admin::UserMergesController < ApplicationController
+  before_action :authorize_super_admin!
+  before_action :find_user
+
+  def new
+  end
+
+  def create
+    target = User.find(params[:merge_target_id])
+    @user.update_attributes!(external_id: target.external_id)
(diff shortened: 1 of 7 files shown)


jonmoter:
@pswadi-zendesk, do you think we need to run this by security or compliance?

pswadi-zendesk:
No, I don't think so. We have done this in the past. Would be good to have some kind of audit trail though.

grosser:
both users will have an even in their history

rikeoka:
LGTM from a security compliance perspective.

PR #1409, opened 2016-10-25
Remove the stage from next_stage_ids immediately, instead of erroring.

dragonfax:
For pod-o-matic. 

@dawbs was having issues deleting stages. It turns out the pipeline was keeping the stage from deleting. 

In the past, deleting stages was a rare thing. But now, with pod-o-matic, we expect it to happen a lot more often. So lets fix pain point and just auto-delete the stage in question, from the pipelines its in.

/cc @zendesk/samson @grosser @jonmoter 

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-288

### Risks
- Level: None

We don't delete stages we simply soft delete them.

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -195,7 +195,7 @@ class Stage < ActiveRecord::Base
 
   def destroy_stage_pipeline
     project.stages.each do |s|
-      if s.next_stage_ids.delete(id.to_s).present?
+      if s.next_stage_ids.delete(id.to_s)
         s.save!
       end
     end
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -194,9 +194,9 @@ class Stage < ActiveRecord::Base
   end
 
   def destroy_stage_pipeline
-    project.stages.each do |s|
+    (project.stages - [self]).each do |s|
       if s.next_stage_ids.delete(id.to_s)
-        s.save!
+        s.save(validate: false)
       end
     end
   end
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -194,12 +194,8 @@ class Stage < ActiveRecord::Base
   end
 
   def destroy_stage_pipeline
-    puts
-    puts "destroying pipeline"
-    puts
     project.stages.each do |s|
       if s.next_stage_ids.delete(id.to_s).present?
-        puts "found a pipeline to destroy"
         s.save!
       end
     end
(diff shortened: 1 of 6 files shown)


app/models/stage.rb
```diff
@@ -190,4 +192,16 @@ def object_attrs_for_paper_trail(attributes)
   def destroy_deploy_groups_stages
     DeployGroupsStage.where(stage_id: id).delete_all
   end
+
+  def destroy_stage_pipeline
+    puts
```
grosser: ideally some more informative log with ids

app/models/stage.rb
```diff
@@ -190,4 +192,16 @@ def object_attrs_for_paper_trail(attributes)
...
+    puts "destroying pipeline"
+    puts
+    project.stages.each do |s|
+      if s.next_stage_ids.delete(id.to_s).present?
```
grosser: no need to check `.present?`

grosser:
sounds alright ...

app/models/stage.rb
dragonfax: Haha, debugging code. Oops.

dragonfax:
Thanks. I'll get it fixed up.

plugins/pipelines/app/models/concerns/samson_pipelines/stage_concern.rb
```diff
@@ -31,15 +31,4 @@ def valid_pipeline?(origin_id = id)
     end
     true
   end
-
-  # Make sure that this stage isn't referenced by another stage in a pipeline.
-  # This will stop soft_deletion of the stage.
-  def verify_not_part_of_pipeline
```
dragonfax: Nope no test.

dragonfax: And here I am proven to be wrong.

dragonfax:
@grosser

app/models/stage.rb
```diff
@@ -190,4 +192,12 @@ def object_attrs_for_paper_trail(attributes)
   def destroy_deploy_groups_stages
     DeployGroupsStage.where(stage_id: id).delete_all
   end
+
+  def destroy_stage_pipeline
+    project.stages.each do |s|
+      if s.next_stage_ids.delete(id.to_s)
+        s.save!
```
grosser: if the project has an invalid stage then this would fail ... maybe better to do `save validate: false`  ... also avoid (unlikely) loops with `{project.stages - [self]).each`

grosser:
👍

PR #1411, opened 2016-10-26
do not trigger mass versioning when stage is created

grosser:
@jonmoter 

 - fixing original bug
 - cleaning up duplicate versions
 - making future bugs easy to spot

JIRA: https://zendesk.atlassian.net/browse/SAMSON-296

### DIFF
--- a/db/migrate/20161026004331_remove_duplicate_versions.rb
+++ b/db/migrate/20161026004331_remove_duplicate_versions.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+class RemoveDuplicateVersions < ActiveRecord::Migration[5.0]
+  def up
+    # delete all versions that are not unique
+    PaperTrail::Version.connection.execute <<-SQL
+      delete from versions where id not in (
+        SELECT * FROM (select min(id) from versions group by item_id, item_type, object) AS temp_tab
+      )
+    SQL
+  end
+
+  def down
+  end
+end
--- a/app/models/command.rb
+++ b/app/models/command.rb
@@ -9,7 +9,7 @@ class Command < ActiveRecord::Base
 
   validates :command, presence: true
 
-  after_save :trigger_stage_change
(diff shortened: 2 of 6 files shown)


db/migrate/20161026004331_remove_duplicate_versions.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+class RemoveDuplicateVersions < ActiveRecord::Migration[5.0]
+  def up
+    # delete all versions that are not unique
+    PaperTrail::Version.connection.execute <<-SQL
```
grosser: this surprisingly works on mysql/sqlite/postgres

db/migrate/20161026004331_remove_duplicate_versions.rb
```diff
@@ -0,0 +1,14 @@
...
+      delete from versions where id not in (
+        SELECT * FROM (select min(id) from versions group by item_id, item_type, object) AS temp_tab
+      )
```
jonmoter: Have you tried running a variation of this query to find out how many rows would be deleted?

jonmoter:
@pswadi-zendesk, this is a fix for a bug where PaperTrail was creating thousands of records due to an interaction between the `Stage` and `Commands` table.

Since this PR includes a migration that will delete a bunch of data from the `Versions` table in our database, do we need to get some sort of approval before running it?

pswadi-zendesk:
Can we find out what rows will be deleted before hand? Also, lets take a backup of the DB so we have it for audit. I will talk with security to confirm.

grosser:
```
PaperTrail::Version.count_by_sql <<-SQL
  select count(id) from versions where id not in (
    SELECT * FROM (select min(id) from versions group by item_id, item_type, object) AS temp_tab
  )
SQL

=> 451679
```
of ~500k
I can also dump them all into a file and if we see the need for it we can go through that ?
... was assuming we have regular db backup anyway ...

PR #1414, opened 2016-10-26
show diff instead of full dumps

grosser:
@jonmoter 

screen shot 2016-10-26 at 8 29 31 am

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -1,9 +1,36 @@
 # frozen_string_literal: true
 class VersionsController < ApplicationController
   def index
-    @versions = PaperTrail::Version.where(
+    versions = PaperTrail::Version.where(
       item_id: params.require(:item_id),
       item_type: params.require(:item_type)
     ).order('created_at desc')
+    @versions = versions_with_diff(versions)
+  end
+
+  private
+
+  # version holds the previous state of the object
+  # for each version calculate the diff to the previous version
+  # or to the current state for the first version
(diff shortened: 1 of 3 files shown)


app/views/versions/index.html.erb
```diff
@@ -1,12 +1,22 @@
...
   </tr>
-  <% @versions.each do |v| %>
+  <%
+    empty = {}.to_yaml
+    current = YAML.load(@versions[0]&.object || empty)
+    @versions.each_with_index do |v, i|
+      previous = YAML.load(@versions[i + 1]&.object || empty)
+      different = (previous.keys + current.keys).uniq.map { |k| [k, previous[k], current[k]] }.select { |k, p, c| p != c }
+      current = previous
+    %>
```
jonmoter: How about we put this in the controller or a helper?  Seems like a little too much logic to have in a view.

grosser:
updated and now with more intuitive diff to current

screen shot 2016-10-26 at 8 52 44 am

app/controllers/versions_controller.rb
```diff
@@ -1,9 +1,25 @@
...
+
+    begin
+      item = versions[0]&.item
+    rescue NameError
+      nil
+    end
+
+    current = (item ? item.send(:object_attrs_for_paper_trail, item.attributes) : {})
+    @versions = versions.map do |v|
+      previous = YAML.load(v&.object || {}.to_yaml)
```
grosser: nah it's not ... just some old logic needed before refactoring ... gone now

PR #1415, opened 2016-10-26
start all deploys even if one in the beginning fails

grosser:
@jonmoter @sandlerr 

 - make missing token not render as 404, but unauthorized
 - test untested codez
 - show which stages failed to deploy

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -2,6 +2,7 @@
 class Integrations::BaseController < ApplicationController
   skip_around_action :login_user
   skip_before_action :verify_authenticity_token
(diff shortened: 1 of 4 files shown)


grosser:
The current behavior looks like a bug to me, if 1 stage is busted then we still want all the other stages to deploy ... since deploy was done in parallel anyway it would not make much sense if the stages depended on each other ... 1 hook to multiple stages is a rare usecase anyway ... so don't care much about this change ... but let me know what you think would be expected behavior ...

grosser:
leaving the cleanup and reverting to old 'stop everything' workflow ... no good reason to switch, so let's just keep it that way ...

PR #1418, opened 2016-10-26
reuse toggle on webhooks example

grosser:
@adam-hart 

before:

screen shot 2016-10-26 at 12 45 19 pm

after:

screen shot 2016-10-26 at 12 44 10 pm
screen shot 2016-10-26 at 12 44 16 pm


grosser:
@irwaters since adam is not around

PR #1419, opened 2016-10-26
Add 'Delete All' button.

dragonfax:
Finally, lets add the "delete all" button. This lets @dawbs cancel the workflow and cleanup, or restart the whole thing from scratch if needbe.

/cc @zendesk/samson @grosser 

### Screenshot

samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/TOOLS-957

### Risks
- Level: None

### DIFF
--- a/app/controllers/admin/deploy_groups_controller.rb
+++ b/app/controllers/admin/deploy_groups_controller.rb
@@ -90,12 +90,9 @@ class Admin::DeployGroupsController < ApplicationController
   end
 
   def merge_all_stages
-    failures = deploy_group.stages.cloned.map do |stage|
-      reason = merge_stage(stage)
-      [reason, stage]
+    failures = try_each_cloned_stage do |stage|
+      merge_stage(stage)
     end
-
-    failures = failures.select(&:first)
 
     message = failures.map { |reason, stage| "#{stage.project.name} #{stage.name} #{reason}" }.join(", ")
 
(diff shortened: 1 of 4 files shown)


dragonfax:
Apparently I don't have access to code climate anymore.

grosser:
screen shot 2016-10-26 at 2 52 27 pm

grosser:
... we should remove code-climate and add some tool to do it's job ...

app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -132,6 +145,17 @@ def merge_stage(stage)
     nil
   end
 
+  def delete_stage(stage)
+    return "has no template stage" unless stage.template_stage
```
grosser: extract and reuse ?

app/views/admin/deploy_groups/show.html.erb
```diff
@@ -80,7 +80,18 @@
                                "Merge #{@deploy_group.stages.cloned.count} cloned stages now?"
                     }
         %>
-    <% end %>
+        <%= link_to "Delete Cloned Stages",
+                    delete_all_stages_admin_deploy_group_path(@deploy_group),
+                    class: "btn btn-default #{ 'disabled' unless @deploy_group.stages.cloned.exists? }",
```
grosser: doing a count above, so do the count once and then reuse it in these 2 places ?

grosser:
feature looks good!

dragonfax:
@grosser

dragonfax:
That line of code is not highly complex.

app/controllers/admin/deploy_groups_controller.rb
grosser: are these not the same checks as in merge ?

dragonfax:
@grosser

grosser:
👍

dragonfax:
@grosser what do you think of code climates complaint?

PR #1420, opened 2016-10-26
remove unnecessary gem and remove more warnings

grosser:
@apanzerj 

we override valid and authenticate ... and don't use config / `@scope` ... so no need to keep warde-doorkeeper around ... 

other changes are random cleanup ... was removing warnings on the hunt for the segfault ...


apanzerj:
👍

PR #1421, opened 2016-10-26
SAMSON-299 avoid giant commits params we get from github ... they break syslog

grosser:
simple/cheap solution ... ideally compare size and truncate as needed, but maybe this is good enough ?

https://zendesk.atlassian.net/browse/SAMSON-299

@jonmoter @murat1985

### DIFF
--- a/lib/samson/logging.rb
+++ b/lib/samson/logging.rb
@@ -12,6 +12,7 @@ elsif Samson::EnvCheck.set?("RAILS_LOG_TO_SYSLOG")
     # show params for every request
     unwanted_keys = %w[format action controller]
     params = event.payload[:params].reject { |key, _| unwanted_keys.include? key }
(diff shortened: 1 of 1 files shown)


murat1985:
👍

murat1985:
that is quiet strange to have information that available on a github additionally in syslog. Does it really needed for troubleshooting or retrospective analysis ?

jonmoter:
Have you verified this works?

The lograge code you're modifying there says that it should strip out the {{%w[format action controller]}}, but I see those parameters appearing in the bug linked above.  So I'm not confident that's working the way we think it should

grosser:
hmm true ... must be coming from somewhere else :/

jonmoter:
Or maybe the keys to `event.payload[:params]` are symbols instead of strings?

grosser:
ah they appear but not as part of the params, so this should be right

PR #1422, opened 2016-10-27
expire sessions

grosser:
@jonmoter 

works as expected, tried with `20.seconds` ... did you have any specific default in mind ?

... adding ENV override if anyone requests it, until then a sane default might be good enough ...

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -153,8 +153,8 @@ module Samson
 
     config.active_job.queue_adapter = :sucker_punch
     config.samson.export_job = ActiveSupport::OrderedOptions.new
-    config.samson.export_job.downloaded_age = (ENV['EXPORT_JOB_DOWNLOADED_AGE'] || 12.hours).to_i
-    config.samson.export_job.max_age = (ENV['EXPORT_JOB_MAX_AGE'] || 1.day).to_i
+    config.samson.export_job.downloaded_age = Integer(ENV['EXPORT_JOB_DOWNLOADED_AGE'] || 12.hours)
+    config.samson.export_job.max_age = Integer(ENV['EXPORT_JOB_MAX_AGE'] || 1.day)
 
     if !Rails.env.test? && ENV['SERVER_MODE'] && !ENV['PRECOMPILE']
       # flowdock uses routes: run after the routes are loaded
--- a/config/initializers/warden.rb
+++ b/config/initializers/warden.rb
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
+# tested via test/integration/authentication_test.rb
 require 'warden'
 
 Warden::Manager.serialize_into_session(&:id)
 
 Warden::Manager.serialize_from_session do |id|
-  User.find(id)
+  timeout = Integer(ENV['SESSION_EXPIRATION'] || 1.month).seconds.ago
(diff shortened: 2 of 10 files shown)


sandlerr:
is this to free up database space? Wouldn't it be better to do that with a cron?

grosser:
Sessions are on Cookies ... This is to limit Session lifetime so users have
to re login once in a while

jonmoter:
If a user is already logged in when this change goes live, what happens?  Will they be logged out, or do you need to create a new session for this change to take effect?

/cc @zendesk/samson

grosser:
if the user has a old session the session will be replaced with one that expires in 1 month

grosser:
hmmm actually if you keep using samson the session will never expire and always be reset to 1 month from last usage ...

grosser:
default rails session expire_after refreshes when used, so not useful to us
this will also add visibility into how samson is used

Initially all users will be logged out, so we get reliable data and do not
keep old users logged in forever.

grosser:
@zendesk/samson FYI this will log out every user once

db/schema.rb
```diff
@@ -477,22 +477,22 @@
...
     t.boolean  "integration",            default: false,      null: false
     t.boolean  "access_request_pending", default: false,      null: false
-    t.string   "time_format",            limit: 255, default: "relative", null: false
+    t.string   "time_format",            default: "relative", null: false
+    t.datetime "last_login_at"
+    t.index ["external_id", "deleted_at"], name: "index_users_on_external_id_and_deleted_at", length: {"external_id"=>191, "deleted_at"=>nil}, using: :btree
   end
 
-  add_index "users", ["external_id", "deleted_at"], name: "index_users_on_external_id_and_deleted_at", length: {"external_id"=>191, "deleted_at"=>nil}, using: :btree
-
```
jonmoter: Why is this index getting deleted?

config/initializers/warden.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
+# tested via test/integration/authentication_test.rb
 require 'warden'
 
 Warden::Manager.serialize_into_session(&:id)
 
 Warden::Manager.serialize_from_session do |id|
-  User.find(id)
+  User.where('last_login_at > ?', 1.month.ago).find_by_id(id)
```
jonmoter: Let's make this an ENV var that can be configured, with a default to 30 days.

config/application.rb
```diff
@@ -153,8 +153,8 @@ class Application < Rails::Application
 
     config.active_job.queue_adapter = :sucker_punch
     config.samson.export_job = ActiveSupport::OrderedOptions.new
-    config.samson.export_job.downloaded_age = (ENV['EXPORT_JOB_DOWNLOADED_AGE'] || 12.hours).to_i
-    config.samson.export_job.max_age = (ENV['EXPORT_JOB_MAX_AGE'] || 1.day).to_i
+    config.samson.export_job.downloaded_age = Integer(ENV['EXPORT_JOB_DOWNLOADED_AGE'] || 12.hours)
```
grosser: FYI @rikeoka much safer way of loading ... so blank / `1.day` etc fail

PR #1423, opened 2016-10-27
every user should have an external_id

grosser:
@zendesk/samson ... unless anyone know why a user should not have an external_id ...

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -23,6 +23,7 @@ class User < ActiveRecord::Base
 
   before_create :set_token
   validates :time_format, inclusion: { in: TIME_FORMATS }
+  validates :external_id, presence: :true, unless: :integration?
 
   scope :search, ->(query) {
     return self if query.blank?
--- a/db/migrate/20161103221846_remove_invalid_users.rb
+++ b/db/migrate/20161103221846_remove_invalid_users.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+class RemoveInvalidUsers < ActiveRecord::Migration[5.0]
+  class User < ActiveRecord::Base
+  end
+
+  def change
+    User.where(deleted_at: nil, external_id: nil, integration: false).update_all(deleted_at: Time.at(1478211603))
+  end
+end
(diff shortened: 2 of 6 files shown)


app/models/user.rb
```diff
@@ -23,6 +23,7 @@ class User < ActiveRecord::Base
 
   before_create :set_token
   validates :time_format, inclusion: { in: TIME_FORMATS }
+  validates :external_id, presence: :true, unless: :integration?
```
jonmoter: How about we validate on `create` only, so we don't end up with a situation where existing users become unable to save their user objects?

grosser: users without external_id cannot log in ... so they are bogus records that should be invalid ... and if they somehow can log in I'd like to understand why / have them report an issue

jonmoter: If we add a validation like this, we should have a plan on what to do with users that fail the validation.  Not okay to have a bunch of saved records in the database that aren't considered valid anymore.

grosser:
updated to add migration that will mark invalid users as deleted ... will affect 30 users

grosser:
@jonmoter looks good ?

grosser:
@jonmoter review ?

db/migrate/20161103221846_remove_invalid_users.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+class RemoveInvalidUsers < ActiveRecord::Migration[5.0]
+  class User < ActiveRecord::Base
+  end
+
+  def change
+    User.where(deleted_at: nil, external_id: nil, integration: false).update_all(deleted_at: Time.at(1478211603))
```
jonmoter: What the heck is that timestamp?

grosser: just picking a random timestamp so I have something unique to query in case things go wrong :D

grosser: 2016-11-03 15:20:03 -0700

PR #1426, opened 2016-10-27
we are only using 1 connection for all our threads which means everyo…

grosser:
…ne has to wait

seeing big memcached spikes ... which looked like coming from threads just waiting for memcached to be free

we cannot read the puma config during server startup, so hardcoding a value

@zendesk/samson


jonmoter:
@pdeuter, this is what we think the critical PR is to get Samson up and running again.  Okay by you?

pdeuter:
+1 deploy freeze exception approved

Issue #1427, opened 2016-10-27
Move to open source on Code Climate.


dragonfax:
You can setup Samson as an open source project in Code Climate to get it off the Zendesk paid account. Code climate is free for public repos. Then just update the README to use the new code climate url.

I don't have admin on this repo, or I'd do it myself

@grosser

grosser:
Tried to, but don't have access to the token ... might be not possible according to webhook docs

`You're trying to set up a service hook for a repository that lives in the Open Source section of your Dashboard. We don't yet support service hooks for these repositories.`

dragonfax:
Lets forget about this for the moment then. Thanks for giving it a try.

PR #1428, opened 2016-10-27
switch to public code climate

grosser:
@zendesk/samson @dragonfax fixes https://github.com/zendesk/samson/issues/1427


grosser:
hmmm you can view the GPA there, but nothing else useful ...

Cannot add the new token to github either, I'd guess the reason is:

```
You're trying to set up a service hook for a repository that lives in the Open Source section of your Dashboard. We don't yet support service hooks for these repositories.
```

PR #1432, opened 2016-10-28
Fix the regex for the Samson version test

bcolferzd:
* Fix the test that matches on the version of Samson
/cc @zendesk/samson


grosser:
👍

PR #1436, opened 2016-10-31
need assets inside of docker too

grosser:
fixes an issue @bradjones1 found in https://github.com/zendesk/samson/issues/1098#issuecomment-257224036

this adds 4:30 to the build process and will be invalidated on every run since it has to be added after app/plugins

changing .nove-version since that is what nodeenv v1.3.6 wanted ... hope that's ok ...

@zendesk/samson 

/cc @sandlerr in case you have a better idea :(


bradjones1:
Thanks @grosser, this fixes the issue on my end though I'm still mystified why precompiling during image build works but not while running the container. Regardless, I appreciate the help.

grosser:
compiling with a running container won't change the image ...
the only way that would work is compiling and starting the server in 1 command like  `CMD precompile && rails s`

bradjones1:
Yeah I thought I had restarted the server but perhaps that's not as easily done when running with puma. Thanks again.

sandlerr:
if we were somehow able to have the assets from the previous docker build available, would it be fast?

grosser:
yes, it would reuse the parts that did not change afaik ... but means we need some pre-build cache thing ... tricky :/

sandlerr:
can we use a circular dependency?

```
FROM zendesk/samson
...
COPY assets assets
RUN bundle exec rake assets:precompile
```

grosser:
not sure how that is supposed to work :D
... but would definitely break the current simple docker-build flow ...
maybe we can figure out what is slow and then not do that ... most likely
minification I'd think ...

sandlerr:
:+1:

grosser:
removed the compile step and now its down to 1 minute ... which gives us 1mb js ... oh boy ...

```
-rw-r--r--  1 root root 4.1M Nov  3 22:22 application-615892047abbfed61a599a1dc4288d69d01a5d44db24266a57c01dcddb9651b7.js
-rw-r--r--  1 root root 982K Nov  3 22:22 application-615892047abbfed61a599a1dc4288d69d01a5d44db24266a57c01dcddb9651b7.js.gz
```

vs

```
-rw-r--r--  1 root root 1.7M Nov  3 22:22 application-f9ad73a6c692794a981402c36b7fcc9978a16c9faf771cd28de77c07f349cafc.js
-rw-r--r--  1 root root 453K Nov  3 22:22 application-f9ad73a6c692794a981402c36b7fcc9978a16c9faf771cd28de77c07f349cafc.js.gz
```

neerfri:
Just to throw in a thought.
I see you are trying to create a docker image that is good for both production and development.
I've seen other projects that went on a bit different approach, basically they have 2 dockerfiles, one is `Dockerfile` and the other is `Dockerfile.dev`.
Then in development the use `Dockerfile.dev` (sometimes using `docker-compose`)
For production they build from the main `Dockerfile` on which they also run the acceptance tests.

@grosser WDYT?

grosser:
I think this add a little too much extra complexity ... so I'd rather do build args ... but I think for now this is good enough ... but would be glad to see PRs/ideas for this :D

neerfri:
@grosser 
I see.
Can you please explain why the 4 minutes build time is a problem?
In my use case I felt it was better to invest those 4 minutes in build time rather than when I start up the container on the production server.
Are you also using docker for local development?

grosser:
current commit makes it a 1 minute build time, but has bigger assets ... long build time means long wait times to deploy which is always annoying ... 

removing vis/rickshaw gives back 3 minutes and 200kb ...  so making them standalone would help (but that's not easy since we do `require_tree .`) 

we have lots of already minimized assets that we try to re-minimize ... would be great to have easy skipping logic for that

so I guess for now let's go with slow and small ... and then improve on that ... 

compose sounds like it might work, but would need support in samson and our local dev tools ...

PR #1437, opened 2016-11-01
kill processes early if getting the gpid fails, since the UI will not be able to shut them down

grosser:
/cc @zendesk/samson 

hope this helps with runaway processes ...

### Risks
 - deploys failing because they did not generate process-groups


grosser:
if we fail to get a gpid then the only thing we can do to clean up is kill the process ... I'm thinking Term then kill would be nicer ... already had that code somewhere ...

PR #1439, opened 2016-11-01
prepare removal of evil db pool patch

grosser:
basically it's a nasty hack that is easy to replace by just setting up database.yml properly

see https://devcenter.heroku.com/articles/concurrency-and-database-connections where this was originally copied from

needs to set `DB_POOL` env var to silence warnings

/cc @zendesk/samson 

### Risks
- Med: db connection issues if pooling is not set up properly


grosser:
... this will screw up heroku though :(

grosser:
since this is just a warning now it should be good to go without any server changes ...

grosser:
... updated ... now using:
 - RAILS_MAX_THREADS
 - RAILS_MIN_THREADS

PR #1441, opened 2016-11-02
add personal oauth tokens for developers, token UI, usage tracking

grosser:
@zendesk/samson 

@abmartinr @apanzerj 

we previously reused the doorkeeper UI, but it was unusable and only available for admins

screen shot 2016-11-02 at 1 53 31 pm
screen shot 2016-11-02 at 2 06 47 pm
screen shot 2016-11-02 at 1 49 23 pm
screen shot 2016-11-02 at 1 51 33 pm

### DIFF
--- a/lib/warden/strategies/doorkeeper_strategy.rb
+++ b/lib/warden/strategies/doorkeeper_strategy.rb
@@ -24,6 +24,7 @@ class Warden::Strategies::Doorkeeper < ::Warden::Strategies::Base
       elsif !(user = User.find_by_id(token.resource_owner_id))
         halt_json "OAuth token belongs to deleted user #{token.resource_owner_id}"
       else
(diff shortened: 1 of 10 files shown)


lib/warden/strategies/doorkeeper_strategy.rb
```diff
@@ -24,6 +24,7 @@ def authenticate!
       elsif !(user = User.find_by_id(token.resource_owner_id))
         halt_json "OAuth token belongs to deleted user #{token.resource_owner_id}"
       else
+        token.update_column(:last_used_at, Time.now) unless token.last_used_at&.> 1.minutes.ago
```
apanzerj: why is there no space after `&.`?

grosser: ... but I like it more without the space anyway ...

PR #1444, opened 2016-11-02
cleanup access request controller a bit and reuse redirect_to logic

grosser:
- no more session data
 - convert 6 argument method to options hash (keyword args don't seem to work with mailer)

@msufa

/fyi @jonmoter @fneves @sbrnunes


msufa:
👍  (I'm no longer at Zendesk tho, so you might want another pair of eyes on this)

PR #1396, opened 2016-10-21
show lock/warning in environments overview / auto-link urls in warnings

grosser:
@apanzerj 

/fyi @pswadi-zendesk 

screen shot 2016-10-21 at 3 16 26 pm
screen shot 2016-10-21 at 3 10 02 pm

PR #1445, opened 2016-11-03
support safe html in lock messages

grosser:
@pswadi-zendesk @abmartinr 

reverts https://github.com/zendesk/samson/pull/1396

screen shot 2016-11-03 at 10 49 40 am


abmartinr:
:+1:

PR #1447, opened 2016-11-03
allow the same stage to run in parallel

anthonywoo:
We have a use case for lotus where we want a stage to be able to run in parallel with itself. The stage builds some assets and sends it to the servers - but it doesn't switch over. Therefore if branch A is running - I shouldn't have to wait for it to be finished before be able to run branch B.

/cc @zendesk/samson @grosser @bcolferzd 

### Tasks
 - [x] :+1: from team
 - [x] Add admin options
 - [x] Write tests

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-291

### Risks
- Medium. Touches when deploys are executed/queued


app/models/deploy_service.rb
```diff
@@ -31,7 +31,13 @@ def confirm_deploy!(deploy)
       send_after_notifications(deploy)
     end
 
-    JobExecution.start_job(job_execution, key: stage.id)
+    if stage.run_in_parallel
```
anthonywoo: good to know - I'll add the job to the registry with a unique key

grosser:
I though the key is already unique since it is the job id ...

anthonywoo:
The key is based on the `stage.id`. Thats how the job queue sees if there's an active deploy and places deploys in the queue. I need to take another look at the `JobQueue` - but registering by job id should do the trick

anthonywoo:
💇 @zendesk/samson can I get another review?

app/models/deploy_service.rb
```diff
@@ -31,7 +31,13 @@ def confirm_deploy!(deploy)
...
+      # immediately execute job
+      JobExecution.start_job(job_execution, key: "deploy-#{deploy.id}")
```
grosser: avoid copy-paste ...

```
key = (stage.run_in_parallel ? "deploy-#{deploy.id}" : "stage-#{stage.id})
```

test/models/deploy_service_test.rb
```diff
@@ -111,6 +111,34 @@ def create_previous_deploy(ref, stage, successful: true, bypassed: false)
...
+        JobExecution.stubs(:new).returns(job_execution)
+        JobExecution.expects(:start_job).with(job_execution, key: "deploy-#{deploy.id}")
+        deploy.buddy = user
+        service.confirm_deploy!(deploy)
+      end
+    end
+
+    describe "when stage can't run in parallel" do
+      before do
+        stage.stubs(:run_in_parallel).returns false
```
grosser: don't need to stub here since that is the default

grosser:
👍

sandlerr:
@anthonywoo this caused a visual bug because the JobExecution key changed: https://github.com/zendesk/samson/blob/master/app/helpers/deploys_helper.rb#L37

any idea how to fix it?

sandlerr:
in fact, it's probably worth having a look at `JobExecution.find_by_id`, `JobExecution.active?`, `JobExecution.queued?`, `JobExecution.start_job` usages all over the shop, I think it is important to use a consistent id generation function, otherwise we risk similar higher-impact bugs

anthonywoo:
Looking over the usage of `JobExecution.start_job` - it seems like deploys are a special case. All other usages of `start_job` register the job with the job `id`. However - deploys have always not used the job `id`. 

For the immediate fix - I'll add the `deploy_key` method. However - this will be something only used for deploys.

Other usages of `JobExecution.start_job`:
https://github.com/zendesk/samson/blob/master/app/models/docker_builder_service.rb#L60
https://github.com/zendesk/samson/blob/master/app/controllers/jobs_controller.rb#L26
https://github.com/zendesk/samson/blob/master/app/controllers/macros_controller.rb#L46
@grosser - do you know how the above are used and why they are able to register a job using its `job.id`?

grosser:
we execute random jobs ... so the the only thing we have is job_id ... for deploys we wanted to prevent parallelism so I guess we want with stage_id ... don't know much about this code ...

PR #1448, opened 2016-11-04
output more useful information (in JSON) when waiting for jobs

jonmoter:
Gives more useful information, like which project, the process id, and the process group id.

Old:
```
Waiting for jobs: [866444028]
```

New:
```
{"time":"2016-11-04 18:11:36 +0000","message":"waiting for jobs to complete","jobs":[{"job_id":866444028,"pid":84869,"pgid":84869,"project":"Truth Service"}],"locks":{}}
```

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/models/restart_signal_handler.rb
+++ b/app/models/restart_signal_handler.rb
@@ -26,8 +26,6 @@ class RestartSignalHandler
 
     until JobExecution.active.empty? && MultiLock.locks.empty?
       loginfo = {
-        timestamp: Time.now.to_s,
-        message: 'waiting for jobs to complete',
         jobs: JobExecution.active.map do |job_exec|
           {
             job_id: job_exec.id,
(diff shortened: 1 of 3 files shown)


app/models/restart_signal_handler.rb
```diff
@@ -25,15 +25,40 @@ def run
...
+      loginfo = {
+        jobs: JobExecution.active.map do |job_exec|
+          {
+            job_id: job_exec.id,
+            pid: job_exec.pid,
+            pgid: job_exec.pgid,
+            project: job_exec.job.project.name
+          }
+        end,
+        locks: MultiLock.locks
```
grosser: locks.keys ?

app/models/restart_signal_handler.rb
```diff
@@ -25,15 +25,40 @@ def run
...
     Process.kill('SIGUSR2', Process.pid)
   end
+
+  def output(message, data = {})
+    output = {
+      timestamp: Time.now.to_s,
+      message: message
+    }.merge(data).to_json
+
+    if Rails.logger
```
grosser: why would there be no rails .logger ?

app/models/restart_signal_handler.rb
jonmoter: As far as I can tell, both the key and the value of `MultiLock.locks` are strings which contain useful information, so I figure I'd just output them both.

app/models/restart_signal_handler.rb
jonmoter: True, I could remove that line.

grosser:
👍

PR #1449, opened 2016-11-04
DRY up logic for building docker images

jonmoter:
currently Samson doesn't fork when building a docker image. this means that all that work (git etc) happens in the ruby thread, really gumming up the works.

The issue is that when you call Docker::Image.build_from_dir, that creates a tarfile in ruby code, which does a whole mess of disk IO.

So far I'm just refactoring some code before I add the forking logic.

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

@grosser

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -66,7 +66,7 @@ class DockerBuilderService
 
     success, build_log = k8s_job.execute!(build, project,
       tag: docker_ref, push: push,
-      registry: registry_credentials, tag_as_latest: tag_as_latest)
+      registry: DockerBuilderService.registry_credentials, tag_as_latest: tag_as_latest)
 
     build.docker_ref = docker_ref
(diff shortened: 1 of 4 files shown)


app/models/docker_builder_service.rb
```diff
@@ -8,6 +8,28 @@ class DockerBuilderService
...
+      Docker::Image.build_from_dir(dir, docker_options, Docker.connection, registry_credentials) do |chunk|
+        output.write_docker_chunk(chunk)
+      end
+    output.puts('### Docker build complete')
+
+    docker_image
+  end
+
+  def self.registry_credentials
+    return nil unless ENV['DOCKER_REGISTRY'].present?
```
grosser: could do `return unless registry = ENV['DOCKER_REGISTRY'].presence` and then reuse registry below

grosser:
👍

PR #1450, opened 2016-11-04
bump brakeman

grosser:
/cc @zendesk/samson


Rakefile
```diff
@@ -72,12 +72,7 @@ end
 
 desc 'Run brakeman ... use brakewan -I to add new ignores'
 task :brakeman do
-  begin
-    File.symlink('plugins', 'engines')
-    sh "brakeman --exit-on-warn --table-width 500"
-  ensure
-    File.unlink('engines')
-  end
+  sh "brakeman --exit-on-warn --table-width 500 --add-engine-path 'plugins/*'"
```
grosser: ❤️  @presidentbeef this worked out nicely :)

presidentbeef: :+1: You should try the plain text output with `-f plain` then you don't have to worry about `--table-width`

PR #1451, opened 2016-11-04
fix docker output so it can handle multiple lines

jonmoter:
When I was building docker images locally (using Docker for Mac), I'd get a bunch of JSON errors when displaying the Docker output.

I did some digging, and sometimes I would get "chunks" back from docker that contained multiple lines, each line with a separate JSON hash.  So this adds some logic to handle that scenario, and also re-encodes the string first.

/cc @grosser 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -112,8 +112,10 @@ class DockerBuilderService
 
     build.docker_image.push(DockerBuilderService.registry_credentials) do |chunk|
       parsed_chunk = output.write_docker_chunk(chunk)
-      if (status = parsed_chunk['status']) && sha = status[DIGEST_SHA_REGEX, 1]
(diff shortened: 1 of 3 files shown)


app/models/docker_builder_service.rb
```diff
@@ -112,8 +112,10 @@ def push_image(tag, tag_as_latest: false)
 
     build.docker_image.push(DockerBuilderService.registry_credentials) do |chunk|
       parsed_chunk = output.write_docker_chunk(chunk)
-      if (status = parsed_chunk['status']) && sha = status[DIGEST_SHA_REGEX, 1]
-        build.docker_repo_digest = "#{project.docker_repo}@#{sha}"
+      parsed_chunk.each do |output_hash|
```
grosser: `parsed_chunks`

grosser:
👍

PR #1452, opened 2016-11-04
build tarfiles for docker images in a separate process

jonmoter:
Creates a tarfile in a forked process

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-303

### Risks
- Level: Med

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -20,6 +20,11 @@ class DockerBuilderService
     output.puts('### Docker build complete')
 
     docker_image
+  ensure
+    if tarfile
+      tarfile.close
+      FileUtils.rm(tarfile.path, force: true)
+    end
   end
 
   def self.create_docker_tarfile(dir)
@@ -36,7 +41,7 @@ class DockerBuilderService
       end
     end
 
-    if ENV['RAILS_ENV'] == 'test'
+    if Rails.env.test?
       tar_proc.call
     else
       pid = fork(&tar_proc)
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -9,15 +9,41 @@ class DockerBuilderService
   attr_reader :build, :execution
 
   def self.build_docker_image(dir, docker_options, output)
-    output.puts("### Running Docker build")
+    output.puts("### Creating tarfile for Docker build")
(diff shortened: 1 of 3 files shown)


app/models/docker_builder_service.rb
```diff
@@ -9,17 +9,43 @@ class DockerBuilderService
...
         output.write_docker_chunk(chunk)
       end
     output.puts('### Docker build complete')
 
     docker_image
   end
 
+  def self.create_docker_tarfile(dir)
+    dir += '/' unless dir.end_with?('/')
+    tempfile_name = Dir::Tmpname.create('out') {}
```
grosser: normally the dir is cleaned up at the end of the block ?

jonmoter: Good point.  Will clean up.

app/models/docker_builder_service.rb
```diff
@@ -9,17 +9,43 @@ class DockerBuilderService
...
+    # It's possible for the puma process to seize up doing all those syscalls,
+    # especially if the disk is running slow. So we create the tarfile in a
+    # separate process to avoid that.
+    tar_proc = -> do
+      File.open(tempfile_name, 'wb+') do |tempfile|
+        Docker::Util.create_relative_dir_tar(dir, tempfile)
+      end
+    end
+
+    if ENV['RAILS_ENV'] == 'test'
```
grosser: `Rails.env.test?`

grosser:
👍

sandlerr:
The [fork doc](https://ruby-doc.org/core-2.1.2/Process.html#method-c-fork) suggests using `spawn` instead, that seems like it might be worth considering since it allows setting rlimits.

app/models/docker_builder_service.rb
```diff
@@ -9,15 +9,46 @@ class DockerBuilderService
...
 
     docker_image
+  ensure
+    if tarfile
+      tarfile.close
+      FileUtils.rm(tarfile.path, force: true)
+    end
+  end
+
+  def self.create_docker_tarfile(dir)
```
grosser: `private_class_method def self.`

sandlerr:
🎉 thanks Jon!

PR #1454, opened 2016-11-05
page titles for every page

grosser:
@jonmoter 
/cc @zendesk/samson 

 - auto-adding `@project` when available
 - enforced via test


app/views/admin/deploy_groups/edit.html.erb
```diff
@@ -1,4 +1,8 @@
-<h1><%= @deploy_group.new_record? ? 'New' : 'Edit' %> Deploy Group <%= link_to_if @deploy_group.id, @deploy_group.name, [:admin, @deploy_group] %></h1>
+<%= page_title do %>
+  <%= @deploy_group.new_record? ? 'New' : 'Edit' %>
+  Deploy Group
+  <%= link_to_if @deploy_group.id, @deploy_group.name, [:admin, @deploy_group] %>
```
jonmoter: Is it a problem to put an `<a>` tag inside of a title?

grosser: titles get strip_tags

PR #1456, opened 2016-11-05
enforce/remove code duplication with flay

grosser:
@zendesk/samson @mariozaizar


app/controllers/admin/deploy_groups_controller.rb
```diff
@@ -90,20 +90,14 @@ def create_all_stages
...
-    redirect_to [:admin, deploy_group], alert: (failures.empty? ? nil : "Some stages were skipped: #{message}")
+    render_failures try_each_cloned_stage { |stage| merge_stage(stage) }
   end
 
   def delete_all_stages
-    failures = try_each_cloned_stage do |stage|
-      delete_stage(stage)
-    end
+    render_failures try_each_cloned_stage { |stage| delete_stage(stage) }
+  end
```
grosser: fyi @dragonfax

PR #1458, opened 2016-11-08
Keep track of updater when updating vault secret

mikehuston:
@grosser


grosser:
confirmed working on staging:

screen shot 2016-11-07 at 8 59 10 pm

grosser:
👍

PR #1459, opened 2016-11-08
show latest deploy in overview so we see when there is trouble/activity

mikehuston:
@zendesk/samson

screen shot 2016-11-07 at 4 46 35 pm


grosser:
@jonmoter @irwaters

PR #1460, opened 2016-11-08
Splunk Audit Logging

rikeoka:
Setup Audit Logging using Splunk for security monitoring of deploys

/cc @zendesk/samson

### Tasks
 - [x] Setup AUDIT_LOGGER
 - [x] Create AuditLog Plugin
 - [X] Add hook fire :audit_action for relevant controllers for modification of create/delete/modify of Projects, Users, DeployGroups, Stages and Environments
 - [X] Add hook fire :audit_actions for login/logout events success and fails
 - [X] Log hooks from before/after deploy and add hook fire :audit_action for buddy bypass
 - [X] Create AuditPresenter for relevant models
 - [x] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/SECC-2144

### Risks
- Level: Low


rikeoka:
@jonmoter / @grosser Is this on the right track to setup audit logging.  Setting it up so include the AuditLog concern, to allow use of the auditX methods to log to splunk.  Also setup an AuditPresenter to reduce the object to a static hash of what is relevant for logging.

app/presenters/audit/deploy_presenter.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class Audit::DeployPresenter
+  ## Deploy presenter for Audit Logger
+  ## Returns deploy object with only id, stage id and name, reference, deployer and buddy id and name, and time
```
grosser: detailed field description is kind of redundant / will get out of date quickly ...

app/presenters/audit/user_presenter.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class Audit::UserPresenter
+  ## User presenter for Audit Logger
+  ## Returns user object with only id, email and name
+
+  def self.present(user)
+    if user
+      {
+          id: user.id,
```
grosser: 2-space indent

app/presenters/audit_presenter.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class AuditPresenter
```
grosser: feels weird that this is AuditPresenter and then the helpers are nested under Audit and not under AuditPresenter

app/presenters/audit_presenter.rb
```diff
@@ -0,0 +1,30 @@
...
+  ## Presenter for Audit Logger
+
+  AVAILABLE_PRESENTERS = [:user, :deploy]
+
+  def self.present(object)
+    unless (object.nil?)
+      type = object.class.name.downcase.to_sym
```
grosser: prefer `underscore` for `FooBar -> foo_bar instead of foobar`

config/initializers/audit_logger.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+require 'splunk_logger'
+
+token = ENV['SPLUNK_TOKEN']
+url = ENV['SPLUNK_URL']
+interval = ENV['SPLUNK_INTERVAL'].to_i || 1
+ssl_verify = ENV['SPLUNK_DISABLE_VERIFY_SSL'] != "1"
+
+if (token && url)
```
grosser: no need for `()`

config/initializers/audit_logger.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+require 'splunk_logger'
+
```
grosser: maybe better to do:

```Ruby
AUDIT_LOGGER = if url = ENV['SPLUNK_TOKEN']
  url = ENV.fetch('SPLUNK_URL')
else
  nil
end
```

grosser:
could be a plugin ?

app/controllers/concerns/audit_log.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def auditLog(user, action, args)
+    auditLogger(:info, user, action, args)
+  end
+
+  def auditInfo(user, action, args)
+    auditLogger(:info, user, action, args)
+  end
+
+  def auditWarn(user, action, args)
```
grosser: method names should be snake case `audit_warn` ... also should not include a bunch of methods into everything ... prefer Audit.warn or something like that

grosser: ... and might as well do `.log :warn, xyz` to remove boilerplate

app/controllers/concerns/audit_log.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def auditLogger(level, user, action, args)
+    if AUDIT_LOGGER && VALID_METHODS.include?(level)
+      message = {}
+      message.user = AuditPresenter.present(user)
+      message.time = Time.now
+      message.action = action
+      i = 0
+      for arg in args
```
grosser: `args.each_with_index do |arg, i|`

grosser:
if we could do this with hooks, that would be much nicer ...

rikeoka:
@grosser I'm starting to get the hooks setup for logging.  Is this on the right track?  Also where I do this, should I also move the Rails.logger.info into the hook since it's performing the logging into the rails logs?

config/initializers/audit_logger.rb
```diff
@@ -0,0 +1,13 @@
...
+token = ENV['SPLUNK_TOKEN']
+url = ENV['SPLUNK_URL']
+interval = (ENV['SPLUNK_INTERVAL'].presence || 1).to_i
+ssl_verify = ENV['SPLUNK_DISABLE_VERIFY_SSL'] != "1"
+
+AUDIT_LOGGER = if token && url
+  SplunkLogger::Client.new({token: token, url: url, send_interval: interval, verify_ssl: ssl_verify})
```
grosser: no need for the outside `{}`

app/controllers/concerns/audit_log.rb
```diff
@@ -0,0 +1,43 @@
...
+    def self.debug(user, action, *args)
+      logger(:debug, user, action, *args)
+    end
+
+    def self.error(user, action, *args)
+      logger(:error, user, action, *args)
+    end
+
+    def self.logger(level, user, action, *args)
+      if AUDIT_LOGGER && VALID_METHODS.include?(level)
```
grosser: atm a invalid call would silently be ignored ... not good ...
`raise ArgumentError unless VALID_METHODS.include?(level)`

app/controllers/concerns/audit_log.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+module AuditLog
+  extend ActiveSupport::Concern
+
+  private
+
+  class Audit
+    VALID_METHODS = [:info, :warn, :debug, :error]
+
+    def self.log(user, action, *args)
```
grosser: avoid all these methods and only expose `log(level, user, action, *args)` ?

rikeoka:
This is so we can get monitoring inside of Splunk,  PaperTrail is good for forensic purposes, but we can't setup triggers on events that the security team would like to monitor.  One use case for example is deploy events can be correlated with file integrity monitoring so we can detect unexpected changes on servers.

app/controllers/concerns/current_user.rb
```diff
@@ -38,6 +38,7 @@ def warden
 
   def unauthorized!
     Rails.logger.warn "Halted as unauthorized! threw :warden (called from #{caller.first.sub(Rails.root.to_s, '')})"
+    Samson::Hooks.fire(:unauthorized_action, current_user, controller_name, action_name)
```
grosser: might want to call this only if there is a `current_user` ?

grosser: ... to reduce spam / the "initial login" flow ... otherwise you won't be able to see anyone trying to mess with things, just constant spam ...

plugins/audit_log/app/presenters/audit_presenter.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class AuditPresenter
```
grosser: should all be under `module AutitLog` then ... and in audit_log folder

plugins/audit_log/app/presenters/audit_presenter.rb
```diff
@@ -0,0 +1,29 @@
...
+  ## Presenter for Audit Logger
+
+  AVAILABLE_PRESENTERS = [:user, :deploy]
+
+  def self.present(object)
+    return unless object
+    type = object.class.name.underscore.to_sym
+    if AVAILABLE_PRESENTERS.include?(type)
+      self.send(type, object)
```
grosser: no need for `self.`

plugins/audit_log/app/presenters/audit_presenter.rb
```diff
@@ -0,0 +1,29 @@
...
+    else
+      object
+    end
+  end
+
```
grosser: can be private ?

plugins/audit_log/app/presenters/audit_presenter.rb
```diff
@@ -0,0 +1,29 @@
...
+  def self.user(user)
+    AuditPresenter::UserPresenter.present(user)
+  end
+
+  def self.deploy(deploy)
+    AuditPresenter::DeployPresenter.present(deploy)
+  end
+
+  def self.project(project)
```
grosser: not used ?

plugins/audit_log/lib/audit_log/samson_plugin.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+require 'splunk_logger'
+
+module SamsonAuditLog
```
grosser: I think you can simply do `AuditLog`

plugins/audit_log/lib/audit_log/samson_plugin.rb
```diff
@@ -0,0 +1,44 @@
...
+  class Engine < Rails::Engine
+  end
+
+  class Audit
+    class << self
+      VALID_METHODS = [:info, :warn, :debug, :error]
+
+      def plugin_enabled?
+        not (ENV['SPLUNK_TOKEN'].nil? || ENV['SPLUNK_URL'].nil?)
```
grosser: `ENV['SPLUNK_TOKEN'] && ENV['SPLUNK_URL']` ...
but these might be used for any kind of feature, so a seperate flag for this plugin would be nice ...

grosser:
looks about right :)

rikeoka:
@grosser, this should be good for another look now.  

@abujduveanu, this covers the events listed in the PR issue.  If you think anything else needs to be logged let me know.

app/controllers/admin/user_merges_controller.rb
```diff
@@ -10,6 +10,7 @@ def create
     target = User.find(params[:merge_target_id])
     @user.update_attributes!(external_id: target.external_id)
     target.soft_delete!
+    Samson::Hooks.fire(:merged_user, current_user, @user, target)
```
grosser: not audit_action ?

plugins/audit_log/test/samson_audit_log/deploy_group_presenter_test.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.covered! unless defined?(Rake) # rake preloads all plugins
```
grosser: the unless rake is only needed for the actual plugin test, nothing else

app/controllers/admin/user_merges_controller.rb
grosser: hmm actually better for callbacks to do it this way ...

app/controllers/sessions_controller.rb
```diff
@@ -30,20 +30,24 @@ def gitlab
   end
 
   def failure
+    Samson::Hooks.fire(:audit_action, nil, 'login', :failed, auth_hash)
```
grosser: too many args ?

app/controllers/sessions_controller.rb
```diff
@@ -30,20 +30,24 @@ def gitlab
...
 
+    Samson::Hooks.fire(:audit_action, user, 'logout', :success)
     flash[:notice] = "You have been logged out."
     redirect_to root_path
   end
 
   protected
 
   def show_login_restriction
+    Samson::Hooks.fire(:audit_action, nil, 'login', :restricted, auth_hash)
```
grosser: same

app/controllers/admin/user_merges_controller.rb
rikeoka: Did this one separately so it could have a warning log level.

plugins/audit_log/lib/samson_audit_log/samson_plugin.rb
```diff
@@ -0,0 +1,58 @@
...
+
+Samson::Hooks.callback :after_deploy do |deploy|
+  SamsonAuditLog::Audit.log(:info, nil, 'deploy ended', deploy)
+end
+
+Samson::Hooks.callback :before_deploy do |deploy|
+  SamsonAuditLog::Audit.log(:info, nil, 'deploy started', deploy)
+end
+
+Samson::Hooks.callback :audit_action do |current_user, action_text, object|
```
rikeoka: Updated to support *objects, in case there are conditions where more than one subject object is needed.

app/controllers/sessions_controller.rb
rikeoka: I made the audit_action method support more than 3 args for such cases in the future, but this could be simplified to failed login, successful login, etc to eliminate the 3rd arg.

app/controllers/admin/projects_controller.rb
```diff
@@ -17,6 +17,7 @@ def destroy
     if Rails.application.config.samson.project_deleted_email
       ProjectMailer.deleted_email(@current_user, project).deliver_later
     end
+    Samson::Hooks.fire(:audit_action, current_user, 'deleted project', project)
```
grosser: how about instead of adding controller callbacks, which we need to repeat/copy-paste in the api ... we use for example environment_decorator with a new `after_commit :audit` and get the user from Papertrail.whodunnit ?
... would mean 1 module to include in 3-4 models ... possibly could remove hooks too since it could call whatever it needed from inside the plugin like `AuditFoo.record :delete, environment`

roughly:
```Ruby
# app/decorators/environment_decorator.rb

module AuditlogConcern
  def self.included(base)
    base.after_commit -> { |e| AuditLog.record :delete, Papertrail.whodunnit, e, }, on: :destroy
    base.after_commit -> { |e| AuditLog.record :create, Papertrail.whodunnit, e, }, on: :create
    base.after_commit -> { |e| AuditLog.record :update, Papertrail.whodunnit, e, }, on: :update
  end
end

Environment.include AuditlogConcern
```

rikeoka: Had to make it slightly more complex to handle the soft_deletion, but I like it better this way.  Definitely the way to go with  the other app.  

Still need to keep the hooks for certain actions that won't be captured cleanly by the after_commit concern.  IE auth, merge user, deploy start/end, unauthorized actions.

app/models/concerns/audit_log.rb
```diff
@@ -0,0 +1,18 @@
...
+module AuditLog
+  def self.included(base)
+    base.after_create_commit do |object|
+      SamsonAuditLog::Audit.log(:info, PaperTrail.whodunnit, "created #{object.class.name}", object)
+    end
+    base.after_destroy_commit do |object|
+      SamsonAuditLog::Audit.log(:info, PaperTrail.whodunnit, "deleted #{object.class.name}", object)
+    end
+    base.after_update_commit do |object|
+      if object.try(:deleted_at)
```
grosser: nice ... was wondering how to handle that :D

app/models/deploy.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 class Deploy < ActiveRecord::Base
+  include AuditLog
```
grosser: might want to include something like `AuditLog::RecordTracker` instead of the root namespace

grosser:
👍 looks pretty good :)

please squasch before merging or use squasch-merge

app/models/concerns/audit_log.rb
```diff
@@ -2,17 +2,25 @@
 module AuditLog
   def self.included(base)
     base.after_create_commit do |object|
-      SamsonAuditLog::Audit.log(:info, PaperTrail.whodunnit, "created #{object.class.name}", object)
+      SamsonAuditLog::Audit.log(:info, AuditLog::user, "created #{object.class.name}", object)
```
grosser: call static methods with `.`

app/models/deploy.rb
grosser: this will fail when the plugin is not loaded, so needs to be in `plugins/audit/app/decorators/deploy_decorator.rb`

```Ruby
Deploy.include AuditLog
```

config/initializers/paper_trail.rb
```diff
@@ -20,7 +22,11 @@ def with_whodunnit_user(user)
   end
 
   def whodunnit
-    @whodunnit_user&.id
+    @whodunnit_user&.id || default
+  end
+
+  def whodunnit_user
+    @whodunnit_user || default
```
grosser: this should not fall back to default ?

config/initializers/paper_trail.rb
```diff
@@ -1,16 +1,18 @@
...
-  elsif defined?(Rake)
-    "#{`whoami`.strip}: rake"
-  end
 
 class << PaperTrail
   undef_method :whodunnit= # nobody uses this directly and we should never use the default papertrail before_actions
   attr_reader :whodunnit_user
 
+  # record what was done in the console
+  def default
```
grosser: avoid collision by using something more specific like `default_whodunnit`

app/models/concerns/audit_log.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+module AuditLog
```
grosser: move to a nested module to avoid including the whole namespace ?

app/models/concerns/audit_log.rb
```diff
@@ -0,0 +1,18 @@
...
+  def self.included(base)
+    base.after_create_commit do |object|
+      SamsonAuditLog::Audit.log(:info, PaperTrail.whodunnit_user, "created #{object.class.name}", object)
```
grosser: lots of copy-paste ... maybe extract to `audit_log "created"`

```
def audit_log(action)
  SamsonAuditLog::Audit.log(:info, PaperTrail.whodunnit_user, "#{action} #{object.class.name}", object)
end
```

app/models/deploy.rb
grosser: this still

rikeoka:
Is it possible to initialize the client to run asynchronously with the plugin?  
This is synchronous and the callbacks may add to the slowness of Samson.  If there was a way to initialize something like AUDIT_LOG_CLIENT to start an asynchronous client (max 2 threads).

grosser:
ah yes ... just spin up a thread in the plugin initializer and feed it
messages via a queue/array

app/models/concerns/has_role.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module HasRole
   def role
-    Role.find(role_id)
+    Role.where(id: role_id).first
```
rikeoka: This looks like it was intended to fail hard.  However the role lookup occurs once the user/project role already has a committed bad role_id.  This would soft_fail it so it would still allow for logging of the bad id commit, and would also prevent the user reports from failing to generate if a user/project role has a bad role_id.

I can revert this back and add some logic in the presenter to check before calling role if needed.

grosser: roles should never be missing ... but it should behave like other belongs_to ... so this is fine ...

config/initializers/paper_trail.rb
```diff
@@ -20,7 +22,11 @@ def with_whodunnit_user(user)
   end
 
   def whodunnit
-    @whodunnit_user&.id
+    @whodunnit_user&.id || default_whodunnit
+  end
+
+  def whodunnit_user
+    @whodunnit_user || default_whodunnit
```
grosser: no fallback here ... should be a user or nil

plugins/audit_log/lib/samson_audit_log/audit_presenter.rb
```diff
@@ -0,0 +1,53 @@
...
+        :user_project_role
+      ].freeze
+
+      def present(object)
+        return unless object
+        type = object.class.name.underscore.to_sym
+        if AVAILABLE_PRESENTERS.include?(type)
+          send(type, object)
+        else
+          object
```
grosser: this smells like it could be trouble ... maybe do `if object.is_a?(ActiveRecord::Base)` present it, otherwise dump ... so it blows up on unknown objects

plugins/audit_log/lib/samson_audit_log/samson_plugin.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  class Audit
+    VALID_METHODS = Set[:info, :warn, :debug, :error]
+
+    def self.log(level, user, action, *args)
+      return unless AUDIT_LOG_CLIENT
+      raise ArgumentError unless VALID_METHODS.include?(level)
+
+      message = {}
```
grosser: ideally:

```
message = {
  actor: xxx,
  time: sdfsdf
  action: 'sddssd
}
```

grosser:
👍

plugins/audit_log/lib/samson_audit_log/audit_presenter.rb
rikeoka: There are cases where I'm not sending over ActiveRecord objects to the presenter.  It would blow up on non JSON serializable objects.

config/initializers/paper_trail.rb
rikeoka: ok, cleaned it up.

plugins/audit_log/lib/samson_audit_log/samson_plugin.rb
rikeoka: cleaned this up

plugins/audit_log/lib/samson_audit_log/audit_presenter.rb
grosser: it would blow up now too ... since it would go into the else ?

rikeoka: The logger should capture the serialization error silently.  Alternatively I could run a serialization here, because there are a few cases where I send over hash's as the object.

grosser:
I was a little worried that a unknown AR object would be dumped and lead to
big response bodies (potentially dumps all it's loaded dependencies) ... so
rather raise so we find it during testing ... non-AR objects should be find
though

rikeoka:
Ok added in a whitelisted objects Set to list out objects that are fine to be sent without processing, and it raises ArgumentError if its outside a presented or whitelisted object now.  I had a few odd whitelisted objects that were being passed from testing and for failed/restricted logins.

grosser:
still want to do the threaded reporter ?

```Ruby
QUEUE = Queue.new
Thread.new do
  loop do
    item = QUEUE.pop
    LOGGER.log item
  end
end

def log(thing)
  QUEUE.push thing
end

```

rikeoka:
The SplunkLogger gem itself spins up a thread to asynchronously send logs to Splunk, every interval seconds or when batch_size messages are queued.  https://github.com/rikeoka/splunk_logger/blob/master/lib/splunk_logger/client.rb#L35

It sends logs in batches to minimize the number of HTTP requests to Splunk, since Splunk handles multiple log entries sent as separate lines.

grosser:
perfect :)

grosser:
closing until someone revisits this ...

PR #1463, opened 2016-11-09
PAAS-515 do not show the token to all users

grosser:
@irwaters @jonmoter


grosser:
screen shot 2016-11-09 at 2 12 01 pm

PR #1466, opened 2016-11-09
PAAS-514 fail the deploy if not all configs are configured instead of silently…

grosser:
@irwaters 
... I think this might pass when rebased on https://github.com/zendesk/samson/pull/1469 so leaving for now


grosser:
fixed to also cover cases where a single deploy group does not have a role configured ... much nicer error messages now :)

PR #1470, opened 2016-11-10
simplify revision a bit

grosser:
@irwaters 

... was hoping to show the current branch and tag ... but was not easy ... but ended up hopefully simplifying it a bit ...


grosser:
@irwaters

PR #1471, opened 2016-11-10
warning UI improvements

grosser:
@pswadi-zendesk @bhouse 

screen shot 2016-11-10 at 11 20 06 am
screen shot 2016-11-10 at 11 19 08 am
screen shot 2016-11-10 at 11 22 09 am

### DIFF
--- a/app/helpers/locks_helper.rb
+++ b/app/helpers/locks_helper.rb
@@ -2,11 +2,11 @@
 module LocksHelper
   def delete_lock_options
     [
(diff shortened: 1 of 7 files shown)


pswadi-zendesk:
So does it show same text (expire in ..) for both lock and warn? Also, would this allow to deploy when warning is on, and won't allow to deploy when lock is on?

grosser:
yes

pswadi-zendesk:
👍

PR #1473, opened 2016-11-11
kubernetes stages do not use commands, so they should not be shown in…

grosser:
… command list

will make the kubernetes stage not show up on command show page and other stats pages

@irwaters


grosser:
migration will delete the 195 invalid stage-commands in prod

PR #1481, opened 2016-11-14
fix all api post requests needing authenticity token because before_f…

grosser:
…ilter was run before auth

@apanzerj @abmartinr 

controller tests were a bad choice for this ...


abmartinr:
:+1: on 🍏

PR #1482, opened 2016-11-14
Make releases version type agnostic

adam-hart:
We're looking to switch the Classic versioning system to Major-Minor releases to better match our actual release process.  This adds Samson support for more than just Continuous versioning.  

It's rather naive, and will just increment the last digit of any versioning system.  Examples!

`v1` -> `v2`
`v1.0` -> `v1.1`
`v1.5.7` -> `v1.5.8`

Versioning system changes, as well as any non-least-significant release will require a manual release which is already built into Samson.

### Tasks
 - [ ] :+1: from team

### Risks
- Medium:  Could cause errors when tagging releases.

### DIFF
--- a/app/controllers/releases_controller.rb
+++ b/app/controllers/releases_controller.rb
@@ -20,11 +20,13 @@ class ReleasesController < ApplicationController
   end
 
   def create
-    @release = ReleaseService.new(@project).create_release!(release_params)
-    redirect_to [@project, @release]
-  rescue ActiveRecord::RecordInvalid => e
-    flash[:error] = e.message
-    redirect_to action: :new
+    @release = ReleaseService.new(@project).release!(release_params)
+    if @release.persisted?
+      redirect_to [@project, @release]
+    else
+      flash[:error] = @release.errors.full_messages.to_sentence
+      render action: :new
+    end
   end
 
   private
--- a/app/controllers/releases_controller.rb
+++ b/app/controllers/releases_controller.rb
@@ -22,6 +22,9 @@ class ReleasesController < ApplicationController
   def create
     @release = ReleaseService.new(@project).create_release!(release_params)
     redirect_to [@project, @release]
+  rescue ActiveRecord::RecordInvalid => e
+    flash[:error] = e.message
+    redirect_to action: :new
   end
 
   private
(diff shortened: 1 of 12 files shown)


app/models/release.rb
```diff
@@ -49,9 +49,14 @@ def self.find_by_version!(version)
...
 
-    latest_release_number = project.releases.last.try(:number) || 0
-    self.number = latest_release_number + 1
+    latest_release_number = project.releases.last.try(:number) || "0"
+
+    # Split and increment the lowest versioning digit
+    split_release_number = latest_release_number.split(".")
+    split_release_number[-1] = (split_release_number[-1].to_i + 1).to_s
+
+    self.number = split_release_number.join(".")
```
grosser:
```ruby
raise "Unable to auto bump version" unless self.number = latest_release_number.dup.sub!(/\d+$/) { |d| d.to_i + 1}
```

db/migrate/20161114194705_releases_number_to_string.rb
```diff
@@ -0,0 +1,5 @@
+class ReleasesNumberToString < ActiveRecord::Migration[5.0]
+  def change
+    change_column :releases, :number, :string, limit: 255, default: "1", null: false
```
grosser: use 20 instead of 255 to avoid index string length madness ...

test/models/release_test.rb
```diff
@@ -11,31 +11,47 @@
...
     end
 
-    it "increments the release number" do
-      release = project.releases.create!(author: author, commit: "bar")
-      release.update_column(:number, 41)
-      release = project.releases.create!(commit: "foo", author: author)
-      assert_equal 42, release.number
+    describe "when incrementing the release number" do
+      it "correctly increments continuous versions" do
+        release = project.releases.create!(author: author, commit: "bar")
```
grosser: or maybe `"41" => "42"}.each do ... test` ...

grosser:
- form needs to change to text input from number input
 - number needs a "\A\d+(\.\d+)*\z" validation so we don't get weird stuff in there ... then we then cannot bump

app/controllers/releases_controller.rb
```diff
@@ -22,6 +22,9 @@ def new
   def create
     @release = ReleaseService.new(@project).create_release!(release_params)
```
grosser: ideally should be

```Ruby
@release = ReleaseService.new(@project).create_release(release_params)
if @release.persisted?
  ...
else
  render :new
end
```

grosser:
👍

PR #1483, opened 2016-11-14
make deploy group permalink editable

grosser:
@irwaters


grosser:
screen shot 2016-11-14 at 12 56 24 pm

grosser:
Permalinkable adds validate_unique_permalink

PR #1485, opened 2016-11-15
use actual user for whodunnit

grosser:
it's silly that we only have the id available ... might as well keep the full object to do all kinds of reporting

/cc @zendesk/samson

will make https://github.com/zendesk/samson/pull/1460 less hacky


grosser:
screen shot 2016-11-15 at 9 02 30 am

PR #1487, opened 2016-11-16
do not create pending builds for every release

grosser:
/cc @zendesk/samson

@sandlerr @jonmoter 

14k of our 18k builds are bogus builds that never ran ... so cleaning them up too

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -12,14 +12,12 @@ class Integrations::BaseController < ApplicationController
       return head(:ok)
     end
 
-    release = project.create_releases_for_branch?(branch)
-    record_log :info, "Branch #{branch} is release branch: #{release}"
-    if release
-      create_release_and_build_record
-    end
+    create_release = project.create_releases_for_branch?(branch)
+    record_log :info, "Branch #{branch} is release branch: #{create_release}"
+    release = find_or_create_release if create_release
 
     if project.build_docker_image_for_branch?(branch)
-      create_docker_image
+      create_docker_image(release)
     end
 
     stages = project.webhook_stages_for(branch, service_type, service_name)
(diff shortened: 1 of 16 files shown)


grosser:
also various cleanups for madness I ran into when trying to test it locally :/

grosser:
@jonmoter @sandlerr review would be great ;)

app/controllers/integrations/base_controller.rb
```diff
@@ -105,12 +102,6 @@ def service_name
     @service_name ||= self.class.name.demodulize.sub('Controller', '').downcase
   end
 
-  def create_release_and_build_record
-    release = create_new_release
-    release = latest_release unless release && release.persisted?
-    create_build(release.version, [release])
```
sandlerr: is there a way to keep this information in the case we create a build after creating a release?

grosser: yes, good point ... the build would not be connected to the release ... which we don't need since the build should have the same sha as the release ... will fix for now and add a comment about it being unnecessary

grosser:
it's amazing how much bad code there can be in so little space ... should be better now ...

app/controllers/integrations/base_controller.rb
```diff
@@ -53,11 +51,10 @@ def release_params
     { commit: commit, author: user }
   end
 
-  def create_new_release
-    unless project.last_release_contains_commit?(commit)
-      release_service = ReleaseService.new(project)
-      release_service.release!(release_params)
-    end
+  def find_or_create_release
+    latest_release = project.releases.order(:id).last # FIXME: ordering by number would be good here
```
grosser: @adam-hart here it's more necessary to sort by number :(

grosser: (but this is a old bug ... so leaving as FIXME)

grosser:
checked if we have many releases that do not have shas ... about 170 ... so backfilling them from their builds ... next PR will make releases always have commits, since keeping something like `master` as `commit` is useless at best ... dangerous at worst

app/controllers/integrations/base_controller.rb
grosser: nvm ... numbers should always be increasing ... so need a new validation for that ... next PR

PR #1490, opened 2016-11-17
make sure all release commits are proper shas and not refs that can c…

grosser:
…hange later

followup to https://github.com/zendesk/samson/pull/1487

the 99% case is creating releases via integration controller ... and they already use shas, so making the manual creation do the same so we can do cheap compares like we do in `contains_commit?` and know what exactly was deployed even if `master` later changes

@henders @jonmoter

### DIFF
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -5,9 +5,10 @@ class Release < ActiveRecord::Base
   belongs_to :build # direct association is not necessary since the release commit is the same as the build sha
 
   before_validation :assign_release_number
+  before_validation :covert_ref_to_sha
 
   validates :number, format: { with: /\A\d+(.\d+)*\z/, message: "may only contain numbers and decimals." }
-  # TODO: commit must be a sha
+  validates :commit, format: { with: Build::SHA1_REGEX, message: "can only be a full sha"}, on: :create
 
   # DEFAULT_RELEASE_NUMBER is the default value assigned to release#number by the database.
   # This constant is here for convenience - the value that the database uses is in db/schema.rb.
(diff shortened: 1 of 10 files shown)


grosser:
there are a few more loose ends here like `Deploy` not knowing it's sha1 ... and `Job#commit` being a ref and getting replaced after ... but 1 issue at a time :)

jonmoter:
Looks like the tests are failing, or timing out

grosser:
cloning was not subbed, so it was fast locally and slow on travis ...
so good that the timeouts caught that :)

grosser:
all good now!

app/models/release.rb
```diff
@@ -61,4 +62,14 @@ def contains_commit?(other_commit)
     Airbrake.notify(e, parameters: { github_repo: project.github_repo, commit: commit, other_commit: other_commit})
     false # Err on side of caution and cause a new release to be created.
   end
+
+  private
+
+  def covert_ref_to_sha
+    return if commit =~ Build::SHA1_REGEX
```
jonmoter: If `commit` is blank, not worth updating the local cache below. I'd change this to

```
return if commit.nil? || commit =~ Build::SHA1_REGEX
```

app/models/release.rb
```diff
@@ -5,9 +5,10 @@ class Release < ActiveRecord::Base
   belongs_to :build # direct association is not necessary since the release commit is the same as the build sha
 
   before_validation :assign_release_number
+  before_validation :covert_ref_to_sha
 
   validates :number, format: { with: /\A\d+(.\d+)*\z/, message: "may only contain numbers and decimals." }
-  # TODO: commit must be a sha
+  validates :commit, format: { with: Build::SHA1_REGEX, message: "can only be a full sha"}
```
jonmoter: Let's do this `on_create`

PR #1491, opened 2016-11-17
always start builds

grosser:
@jonmoter @henders 

already talked through this once and kept it since releases had these weird unbuild-builds ... but no longer need that ... I hope
(removing the json replies since we don't need that ... )

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -20,25 +20,10 @@ class BuildsController < ApplicationController
   end
 
   def create
-    @build = create_build
+    @build = new_or_modified_build
     @build.creator = current_user
-    @build.save
-
-    start_docker_build if @build.persisted? && params[:build_image].present?
-
-    respond_to do |format|
-      format.html do
-        if @build.persisted?
-          redirect_to [current_project, @build]
-        else
-          render :new, status: :unprocessable_entity
-        end
-      end
-
-      format.json do
-        render json: {}, status: (@build.persisted? ? :created : :unprocessable_entity)
-      end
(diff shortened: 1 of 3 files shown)


app/controllers/builds_controller.rb
```diff
@@ -8,36 +8,21 @@ class BuildsController < ApplicationController
...
     @build = current_project.builds.build
   end
 
   def create
-    @build = create_build
+    @build = new_or_modified_build
     @build.creator = current_user
-    @build.save
-
-    start_docker_build if @build.persisted? && params[:build_image].present?
```
grosser: also fixing that previously failed updates where ignored since they were 'persisted?'

app/controllers/builds_controller.rb
```diff
@@ -8,36 +8,21 @@ class BuildsController < ApplicationController
 
   def index
     @builds = current_project.builds.order('id desc').page(params[:page])
-
-    respond_to do |format|
-      format.html
-      format.json { render json: @builds }
-    end
```
grosser: because it is not needed ?

jonmoter:
@grosser please don't randomly remove functionality, like the JSON replies. People outside of Zendesk use Samson, and might be using that endpoint.  Does it do any harm to have it in the controller?

grosser:
we have an api now, so ideally should remove regular json ... and ideally all unused code ... but yeah hard to know if it is actually unused ... I'll revert that part ...

grosser:
reverted, good to go ?

grosser:
flay found duplication :D

PR #1495, opened 2016-11-18
warn when deploying a previous release

grosser:
/cc @zendesk/samson 
@jonmoter @stubotnik 

screen shot 2016-11-18 at 9 55 35 am

needs https://github.com/zendesk/samson/pull/1496

### DIFF
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -98,7 +98,8 @@ $(function () {
     $ref_problem_list.empty();
     $.each(status_list, function(idx, status) {
       if (status.state != "success") {
-        var item = $ref_problem_list.append($("<li>"));
+        var item = $("<li>");
+        $ref_problem_list.append(item);
         item.text(status.state + ": " + status.description);
         if(status.target_url) {
           item.append(' <a href="' + status.target_url + '">details</a>');
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 # Used to display all warnings/failures before user actually deploys
 class CommitStatus
-  def initialize(user_repo_part, reference)
-    @repo = user_repo_part
+  def initialize(stage, reference)
+    @stage = stage
     @reference = reference
   end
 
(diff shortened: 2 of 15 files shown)


app/assets/javascripts/deploys.js
```diff
@@ -98,7 +98,8 @@ $(function () {
     $ref_problem_list.empty();
     $.each(status_list, function(idx, status) {
       if (status.state != "success") {
-        var item = $ref_problem_list.append($("<li>"));
+        var item = $("<li>");
+        $ref_problem_list.append(item);
```
grosser: previously only showed the last warning since .append returns the ul and then we put text into the ul ...

app/models/commit_status.rb
```diff
@@ -17,13 +17,45 @@ def status_list
...
+
+  # checks if other stages that deploy to the same hosts as this stage have deployed a newer release
+  # @return [nil, error-state]
+  def release_status
+    return unless DeployGroup.enabled?
+    return unless number = @reference[Release::VERSION_REGEX, 1]
+
+    deploy_groups = @stage.deploy_groups.pluck(:deploy_group_id)
+    influencing_stages = DeployGroupsStage.where(deploy_group_id: deploy_groups).map(&:stage).uniq
+    last_deploys = influencing_stages.map(&:last_successful_deploy).compact # N+1 ... cannot order and then group
```
grosser: this is a bit shitty since it is a n+1 that also does a join ...

grosser:
optimized sql to be a bit faster ... this takes about 0.3s compared to 0.1s before (0.1s = github lookup)

grosser:
another big win with max(id) grouping and adding reorder(nil) ... now it's dirt cheap at 0.01s :D

app/models/commit_status.rb
```diff
@@ -17,13 +17,47 @@ def status_list
...
+  # @return [nil, error-state]
+  def release_status
+    return unless DeployGroup.enabled?
+    return unless number = @reference[Release::VERSION_REGEX, 1]
+
+    deploy_groups = @stage.deploy_groups.reorder(nil).pluck(:deploy_group_id)
+    influencing_stages = DeployGroupsStage.reorder(nil).where(deploy_group_id: deploy_groups).pluck(:stage_id).uniq
+    last_deployed = Deploy.reorder(nil).where(stage_id: influencing_stages).group(:stage_id).pluck('max(id)')
+    last_deployed_references = Deploy.reorder(nil).where(id: last_deployed).pluck(:reference)
+    last_deployed_numbers = last_deployed_references.map { |r| r[Release::VERSION_REGEX, 1] }.compact
```
jonmoter: Seems like this logic could get refactored into a class method of something like `DeployGroup`, and then have its own tests.

jonmoter:
Could you make the error message more specific?  Something like "v2 was already deployed to pod100 by the `foobar` stage"?

app/models/commit_status.rb
grosser: extracted it into a method ... looks alright ?
 ... otherwise could extract it to stage ... but that feels weird ... (not fetching complete deploy groups for speed)

grosser:
error message now also has the other stage ... re-fetching to keep the reference queries fast ... once we found an error we can spend a few more sql round-trips

grosser:
extracted influencing_stage_ids method that I will reuse in the auto-deploy PR

app/models/release.rb
```diff
@@ -1,13 +1,17 @@
 # frozen_string_literal: true
 class Release < ActiveRecord::Base
+  NUMBER = '\d+(:?.\d+)*'
```
jonmoter: Don't you need a backslash before the `.` in there, so it's treated as a literal period?

PR #1498, opened 2016-11-18
add project_id to deploys to avoid expensive joins whenever we want t…

grosser:
…o know the project of a deploy

@zendesk/samson

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -13,14 +13,14 @@ class Project < ActiveRecord::Base
   after_soft_delete :clean_repository
   before_soft_delete :destroy_user_project_roles
 
-  has_many :builds
-  has_many :releases
+  has_many :builds, dependent: :destroy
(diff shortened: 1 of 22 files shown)


app/models/project.rb
```diff
@@ -13,14 +13,14 @@ class Project < ActiveRecord::Base
...
-  has_many :jobs, -> { order(created_at: :desc) }
-  has_many :webhooks
-  has_many :outbound_webhooks
-  has_many :commands
+  has_many :deploys
+  has_many :jobs, -> { order(id: :desc) }
+  has_many :webhooks, dependent: :destroy
+  has_many :outbound_webhooks, dependent: :destroy
+  has_many :commands, dependent: :destroy
   has_many :macros, dependent: :destroy
```
grosser: adding missing cleanup calls ... and using faster sort for jobs

PR #1500, opened 2016-11-22
TOOLS-1024 automated deploys

grosser:
@ubootda @jonmoter @bcolferzd 

https://zendesk.atlassian.net/browse/TOOLS-1024

a rough draft of what I think it should look like ... just wanted to type it all out to not forget the corner cases ... please let me know what you think / if thats the right place to put it

```
curl -X POST -H 'Authorization: Bearer TOKEN-GOES-HERE' 'http://localhost:3000/api/projects/kube_translation_daemon/automated_deploys.json?deploy_group=pod1&env[host]=app1.foo.bar'
{"id":12,"stage_id":5,"job_id":13,"reference":"master","created_at":"2016-11-22T00:26:44.000Z","updated_at":"2016-11-22T00:26:44.000Z","buddy_id":null,"started_at":null,"deleted_at":null,"build_id":null,"release":true,"kubernetes":true}
```

```
curl -X POST -H 'Authorization: Bearer INSERT_TOKEN_HERE' 'https://samsontest.zende.sk/api/projects/rosetta/automated_deploys.json?deploy_group=pod99&env%5BHOST%5D=app1.foo.bar'
{"id":4002,"stage_id":240,"job_id":4521,"reference":"master","created_at":"2016-12-01T20:32:24.000Z","updated_at":"2016-12-01T20:32:24.000Z","buddy_id":null,"started_at":null,"deleted_at":null,"build_id":null,"release":true,"kubernetes":false,"project_id":95}

...

» cd /data/samson/tmp/samson-rosetta-452120161201-2417-137es6p
» export PARAM_HOST=app1.foo.bar
» export DEPLOY_GROUPS=master99
```

this will not be able to be re-deployed ... but fail nicely when it is done

TODO:
 - [x] tests
 - [ ] check with compliance
 - [x] try on staging

### DIFF
--- a/app/controllers/api/automated_deploys_controller.rb
+++ b/app/controllers/api/automated_deploys_controller.rb
@@ -9,7 +9,7 @@ class Api::AutomatedDeploysController < Api::BaseController
 
   def create
     deploy_service = DeployService.new(current_user)
-    env = params.require(:env).to_unsafe_hash.map { |k, v| "export PARAM_#{k}=#{v.shellescape}" }
+    env = params.require(:env).to_unsafe_hash.map { |k, v| "export PARAM_#{k}=#{v.gsub("\n", "\\n").shellescape}" }
     env << "export DEPLOY_GROUPS=#{@deploy_group.env_value}"
 
     deploy = deploy_service.deploy!(
--- a/app/controllers/api/automated_deploys_controller.rb
+++ b/app/controllers/api/automated_deploys_controller.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+# deploys that automatically get triggered when new hosts come up or are restarted
+class Api::AutomatedDeploysController < Api::BaseController
+  STAGE_NAME = 'Automated Deploys'
+
+  before_action :find_or_create_stage
+  before_action :find_deploy_group
+  before_action :find_last_deploy
+
+  def create
+    deploy_service = DeployService.new(current_user)
+    env = params.require(:env).to_unsafe_hash.map { |k, v| "export PARAM_#{k}=#{v.shellescape}" }
+    env << "export DEPLOY_GROUPS=#{@deploy_group.env_value}"
+
+    deploy = deploy_service.deploy!(
+      @stage,
+      reference: @last_deploy.reference,
+      buddy_id: @last_deploy.buddy_id,
+      before_command: env.join("\n") << "\n",
+      skip_deploy_group_validation: true
+    )
+
+    if deploy.persisted? # TODO: also check that the deploy is running and not waiting for buddy
+      render json: deploy.to_json, status: :created, location: [@stage.project, deploy]
+    else
+      failed! "Unable to start deploy: #{deploy.errors.full_messages}"
+    end
+  end
+
+  private
+
+  def find_or_create_stage
+    project = Project.find_by_permalink!(params.require(:project_id))
+    @stage = project.stages.where(name: STAGE_NAME).first || begin
+      unless template = project.stages.where(is_template: true).first
+        return failed! "Unable to find template for #{project.name}"
+      end
+
+      @stage = Stage.build_clone(template)
+      @stage.name = STAGE_NAME
+      @stage.dashboard = "Automatically created stage from Api::AutomatedDeploysController<br>" \
+        "that will deploy to individual deploy groups or hosts when called via api."
+      unless @stage.save
+        failed!("Unable to save stage: #{@stage.errors.full_messages}")
+      end
+      @stage
+    end
+  end
+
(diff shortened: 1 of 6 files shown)


epssy:
Is it possible to require all teams to have a template stage that can generically deploy to any `deploy_group`? The guest starting will be aware of it's environment and can pass this, though I'm wondering if there's any cases (except for maybe classic?) where this is really required.  

Are we to pass `project_id` through as well? It doesn't appear in your example.

grosser:
project_id is the project permalink, in the example that is kube_translation_daemon

briancolfer:
Should we accomdate any type of stage?  Is there a property related to the stage entity that identifies it as being a production level stage?

grosser:
template stages should be good for any kind of stage ... and usually they are production ... since we use them for creating new pods too

grosser:
for the mandatory template thing I think we'll just go through the projects that we want to deploy and set it, should not be that many

app/controllers/api/automated_deploys_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  before_action :find_or_create_stage
+  before_action :find_deploy_group
+  before_action :find_last_deploy
+
+  def create
+    deploy_service = DeployService.new(current_user)
+    deploy = deploy_service.deploy!(
+      @stage,
+      reference: @last_deploy.reference,
+      buddy_id: @last_deploy.buddy_id,
```
jonmoter: We should check with compliance on whether these deploys should be tagged as having a buddy. Since nobody actually approved this particular deploy, I think that column should be blank.

Though it would be nice if this deploy had a reference to the `@last_deploy` that it copied from, or otherwise tag this as an auto-deploy.

jonmoter: Or maybe we have a special "auto-deploy-buddy" user that can act as a stand-in for a real buddy.

app/models/stage.rb
```diff
@@ -83,8 +83,10 @@ def current_release?(release)
   end
 
   def create_deploy(user, attributes = {})
+    before_command = attributes.delete(:before_command)
     deploys.create(attributes.merge(release: !no_code_deployed)) do |deploy|
-      deploy.build_job(project: project, user: user, command: script, commit: deploy.reference)
+      commands = before_command.to_s.dup << script
+      deploy.build_job(project: project, user: user, command: commands, commit: deploy.reference)
```
jonmoter: I'm still coming back to the idea of the Deploy having a hash of ENV vars that we store as a column in the DB.

That way, the hash can have `{ 'HOSTS' => ['app1.pod12...', 'app2.pod12...'] }`.  And if we wait for 5 minutes to start, we can add hosts to that hash.  That way, we also have a record of which hosts that was run, rather than needing to parse the first line of the commands that are run.

app/controllers/api/automated_deploys_controller.rb
grosser: we use the same copy-the-buddy approach when deploying an already approved version to other stages, so I don't think we should just reuse that approach ... if we want to do something else that can be a separate PR and compliance discussion for what they'd want to see

grosser:
made this more flexible by allowing any kind of env var to be set ... as long as it starts with PARAM_ ... so nobody can override PATH or other important things

```
params.require(:env).map { |k, v| "export PARAM_#{k}=#{v.shellescape}" }.join("\n")
```

grosser:
deploying from staging:

https://samsontest.zende.sk/projects/rosetta/deploys/4007

```
curl -X POST -H 'Authorization: Bearer 7f428ff3540411f9e70856b349a7eae490b' 'https://samsontest.zende.sk/api/projects/rosetta/automated_deploys.json?deploy_group=pod99&env%5BHOSTS_FILE%5D=app1.pod99.aws1.zdsystest.com%3A%5Cn++deploy_projects%3A%5Cn++-+zendesk_rosetta%5Cn++environment%3A+master%5Cn++hostgroup%3A+rosetta%5Cn++pod%3A+99%5Cn++tags%3A+%5B%5D%5Cn'
```

with HOSTS_FILTER param coming from this, not the double-escaped newlines

```
require 'cgi'
host = <<-HOST.gsub("\n", "\\n")
app1.pod99.aws1.zdsystest.com:
  deploy_projects:
  - zendesk_rosetta
  environment: master
  hostgroup: rosetta
  pod: 99
  tags: []
HOST
CGI.escape host
```

... and TADA!

```
app1.pod99.aws1.zdsystest.com - [~] ➔ ls /data/zendesk_rosetta/current
app		config		 Dockerfile	 lib		   public     script	 vendor
```

PR #1501, opened 2016-11-23
add ecr seed button to kubernetes cluster

grosser:
@irwaters @jonmoter 

screen shot 2016-11-22 at 6 27 44 pm
screen shot 2016-11-22 at 7 16 06 pm

adding created to annotations or otherwise it breaks the request for not being base64 encoded
tested on sandbox cluster ... master/staging did not work ...

### DIFF
--- a/plugins/kubernetes/app/controllers/admin/kubernetes/clusters_controller.rb
+++ b/plugins/kubernetes/app/controllers/admin/kubernetes/clusters_controller.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 class Admin::Kubernetes::ClustersController < ApplicationController
   before_action :authorize_admin!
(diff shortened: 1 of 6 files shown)


plugins/kubernetes/app/controllers/admin/kubernetes/clusters_controller.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 class Admin::Kubernetes::ClustersController < ApplicationController
   before_action :authorize_admin!
-  before_action :authorize_super_admin!, only: [:create, :new, :update, :edit]
+  before_action :authorize_super_admin!, except: [:index, :show, :seed_ecr]
```
grosser: it does not have negative side-effects, so should be fine for almost anyone to do

PR #1502, opened 2016-11-23
simplify plugin activation logic

grosser:
saw this yesterday and wanted to clean that up a bit

@irwaters @jonmoter


test/lib/samson/hooks_test.rb
```diff
@@ -5,67 +5,56 @@
 
 describe Samson::Hooks do
   let(:number_of_plugins) { Dir['plugins/*'].size }
-  let(:plugins) { 'nope' }
+  let(:plugins) { nil }
 
   describe '.plugins' do
-    before do
-      Rails.stubs(:env).returns(env.inquiry)
-      ENV['PLUGINS'] = plugins
```
grosser: also no longer leaking plugin state here

PR #1503, opened 2016-11-23
do not modify ENV inside of tests via normal setters, it is easy to f…

grosser:
@irwaters


grosser:
failure looks like:

```
test/controllers/access_requests_controller_test.rb: use with_env to setup ENV variables during test
test/controllers/integrations/travis_controller_test.rb: use with_env to setup ENV variables during test
test/helpers/access_request_helper_test.rb: use with_env to setup ENV variables during test
test/models/changeset/pull_request_test.rb: use with_env to setup ENV variables during test
plugins/slack_app/test/helpers/slack_app_helper_test.rb: use with_env to setup ENV variables during test

```

PR #1506, opened 2016-11-24
add docker-compose

grosser:
now all a user needs to do is `docker-compose up` ... and they have samson running ...

fixes https://github.com/zendesk/samson/issues/1098


grosser:
@sandlerr @kasi1999999

jonmoter:
How about we make a mention of it in the Readme?

grosser:
added it to getting-started

Issue #409, opened 2015-05-13
Show latest tags when creating a new deploy


ncalca:
A common use case when creating a new deploy is to deploy one of the tags that was recently created.

It would be nice to show the latest n (= 5 ?) tags in the following form so that the user can just click on one of them to create a new deploy.

![screen shot 2015-05-13 at 11 20 16](https://cloud.githubusercontent.com/assets/2805296/7607510/36e048f4-f962-11e4-8601-1f041f1d1ea9.png)

mblaschke:
Or just set the default tag/branch as option in deployment stage settings

:+1:

PR #1507, opened 2016-11-27
Show recent releases on new deploy page

ibrahima:
* Lists 5 recent releases on the new deploy page.

show_recent_releases

* Remember to add unit tests
* Addresses #409
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -176,6 +176,7 @@ $(function () {
     }
   }
 
+  // When user clicks a release label, fill the deploy reference field with that version
   $("#recent-releases .release-label").on('click', function(event){
     event.preventDefault();
     // Get version number from link href
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -176,6 +176,13 @@ $(function () {
     }
   }
 
+  $("#recent-releases .release-label").on('click', function(event){
+    event.preventDefault();
+    // Get version number from link href
+    var version = event.target.href.substring(event.target.href.lastIndexOf('/') + 1);
+    $("#deploy_reference").val(version);
+  });
+
   $form.submit(function(event) {
     var $this = $(this);
 
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -94,7 +94,6 @@ class DeploysController < ApplicationController
 
   def new
     @deploy = current_project.deploys.build(params.except(:project_id).permit(:stage_id, :reference))
-    @recent_releases = current_project.releases.last(5)
   end
 
   def create
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -94,7 +94,7 @@ class DeploysController < ApplicationController
 
   def new
     @deploy = current_project.deploys.build(params.except(:project_id).permit(:stage_id, :reference))
-    @recent_releases = current_project.releases.order(id: :desc).first(5)
+    @recent_releases = current_project.releases.last(5)
   end
 
   def create
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -94,7 +94,7 @@ class DeploysController < ApplicationController
 
   def new
     @deploy = current_project.deploys.build(params.except(:project_id).permit(:stage_id, :reference))
-    @releases = current_project.releases.order(id: :desc).last(5)
+    @recent_releases = current_project.releases.order(id: :desc).last(5)
   end
 
   def create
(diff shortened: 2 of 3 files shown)


ibrahima:
Wanted to gauge whether I was on the right track before working on this further (I suppose I need to add unit tests). Would this be something that people are interested in? It would certainly help me save a little time. The release labels are clickable, and clicking on them inserts the version tag into the input field above.

app/assets/javascripts/deploys.js
```diff
@@ -176,6 +176,13 @@ $(function () {
     }
   }
 
+  $("#recent-releases .release-label").on('click', function(event){
+    event.preventDefault();
+    // Get version number from link href
+    var version = event.target.href.substring(event.target.href.lastIndexOf('/') + 1);
```
grosser: how about: `parts = split('/'); parts[parts.size - 1]`

app/assets/javascripts/deploys.js
```diff
@@ -176,6 +176,13 @@ $(function () {
     }
   }
 
+  $("#recent-releases .release-label").on('click', function(event){
```
grosser: some description would be helpful for future archeologists:

```
// when user clicks a release label then fill the deploy reference field with that version
```

grosser:
looks about right!

app/assets/javascripts/deploys.js
ibrahima: I dunno, I prefer this way because there's no need to actually split it into all the parts.

app/assets/javascripts/deploys.js
ibrahima: Good idea!

app/controllers/deploys_controller.rb
```diff
@@ -94,6 +94,7 @@ def search
 
   def new
     @deploy = current_project.deploys.build(params.except(:project_id).permit(:stage_id, :reference))
+    @releases = current_project.releases.order(id: :desc).last(5)
```
ibrahima: Agreed!

ibrahima:
@grosser what further do I need to do for this branch to be merge-ready? Not sure what kind of tests I should add for this feature, there don't seem to be any page tests?

app/views/deploys/new.html.erb
```diff
@@ -24,6 +24,17 @@
 
       <%= Samson::Hooks.render_views(:deploy_form, self, project: @project, form: form) %>
 
+      <% if @recent_releases %>
```
grosser: don't need an if

app/controllers/deploys_controller.rb
```diff
@@ -94,6 +94,7 @@ def search
 
   def new
     @deploy = current_project.deploys.build(params.except(:project_id).permit(:stage_id, :reference))
+    @recent_releases = current_project.releases.order(id: :desc).last(5)
```
ibrahima: oh oops, should be `.first(5)`. I just wanted to show the most recent release first.

app/views/deploys/new.html.erb
ibrahima: I think you do, because if you just create a new project with no releases, it'll crash (can't call each on nil). The specs were crashing before I added this check.

app/controllers/deploys_controller.rb
```diff
@@ -94,6 +94,7 @@ def search
 
   def new
     @deploy = current_project.deploys.build(params.except(:project_id).permit(:stage_id, :reference))
+    @recent_releases = current_project.releases.order(id: :desc).first(5)
```
grosser: just do `releases.last(5)` ... no need to change ordering around ...

app/views/deploys/new.html.erb
grosser: `.first(5)` will always return something

Release.where('1=2').first(5) => []

app/controllers/deploys_controller.rb
ibrahima: sure thing!

app/views/deploys/new.html.erb
ibrahima: Ah okay, I don't know why that didn't seem to work earlier. I'll revert the commit and see if CI passes again!

ibrahima: Well, it's also kind of weird to show a label for "Recent Releases" and then show nothing if there are no recent releases. What do you think about that?

grosser:
then should be:
if @recent_releases.any?

ibrahima:
Oh I see, the CI failure is for when create fails, because it's rendering the new form. Should I just add the instance variable to the create action as well?

grosser:
can make the view pull the data ... problem avoided ;)

ibrahima:
That's what I was thinking... feels awful, but less awkward than putting the same query in two places.

ibrahima:
Not sure if you were serious about pulling data in the view, but I did that if that seems reasonable to you.

PR #1511, opened 2016-12-01
try to reduce beachball on firefox

sandlerr:
cc @zendesk/samson randomly paused the debugger in here a couple of times, but maybe all the message handlers need to be throttled.


grosser:
:+1: ❤️

sandlerr:
@grosser tried to fix a flaky test, does this look reasonable?

test/models/terminal_executor_test.rb
```diff
@@ -65,13 +65,13 @@
 
     it "terminates the program on getpgid failures so we fail fast" do
       Process.expects(:getpgid).raises(Errno::ESRCH)
-      subject.execute!('echo "hi"').must_equal(nil) # status? of a killed process is 9
+      subject.execute!('sleep 1').must_equal(nil) # status? of a killed process is 9
     end
 
-    it "terminates the program on getpgid + kill failures when the program was already dead" do
+    it "handles the case when getpgid + kill both fail" do
```
grosser: I like the original better `handles` is always weird :D 
`it "kills the program when xyz"`

grosser:
👍 prefer `sleep 0.1` ... we have enough slow tests already :D

test/models/terminal_executor_test.rb
sandlerr: I used `handles` to mean `doesn't crash`

grosser: 👍

sandlerr:
ok, I'm done now even though the tests flaked again it looks like it was mysql's fault. We should probably not talk to the db so much in tests. Can I get one more +1?

grosser:
👍

PR #1515, opened 2016-12-05
move job execution key logic from deploy service to deploy

anthonywoo:
Previously the job execution key for deploys logic was in the deploy service. However - the same key needs to be available to check if the deploy is running, queued, etc.

/cc @zendesk/samson @sandlerr 

### Tasks
 - [x] :+1: from team

### References
 - Jira link:

### Risks
- Level: Med. Touches deploys and could break deploy notifications


grosser:
👍

PR #1516, opened 2016-12-06
PAAS-443 add cron for deleting dead nodes

grosser:
@zendesk/paas


grosser:
looks like we can achieve this by setting cloud provider aws

PR #1520, opened 2016-12-06
make buddy check simpler/faster and make deleting a command also expi…

grosser:
…re buddy checks

... also fix nasty bug that did not record when adding or removing commands

@irwaters @jonmoter

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -145,11 +145,9 @@ class Stage < ActiveRecord::Base
     emails.uniq.presence
   end
 
-  def command_updated_at
-    [
-      command_associations.maximum(:updated_at),
-      commands.maximum(:updated_at)
-    ].compact.max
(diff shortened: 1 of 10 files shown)


grosser:
FYI somehow travis databases lag a few seconds behind and that breaks test cases ... so need the space things by minutes to properly test timing sensitive things.

app/models/stage.rb
```diff
@@ -177,6 +175,11 @@ def environments
     DeployGroup.enabled? ? deploy_groups.map(&:environment).uniq : []
   end
 
+  def command_ids=(value)
+    @command_ids_changed = (command_ids != value.map(&:to_i))
```
irwaters: does this need any ordering?

grosser: if the ordering changed then the script will be modified, so we should record a new version

PR #1522, opened 2016-12-07
keep keys cache for longer and make modifying or deleting a secret si…

grosser:
fetching all keys from all vault servers takes forever ... the current cache is nice, but only holds 10 minutes ... also every change expires everything which makes checking the index page painful ... so let's keep our cache longer ...

@jonmoter @irwaters 

... a little bit racy ... but hopefully not that bad / won't change that often ...


grosser:
I'd prefer to avoid background jobs ... because then we need to run them for everyone and somehow monitor them too :D 
... if this works out it should be good enough and only work when you actually use this feature

PR #1523, opened 2016-12-08
add cmd2json for nice cron logging

grosser:
atm our cron logging is non-existent ... tries to pipe some rake output to syslog ... and if it fails we get nothing ... not nice ... so switch to https://github.com/grosser/cmd2json

```
cmd2json -a 'foo=bar' echo hello | logger -t test
tail -1 /var/log/user.log
2016-12-08T00:34:17+00:00 app1.hostfoo.com test: {"timestamp":"2016-12-08 00:34:17 +0000", "message":"hello", "exit":"0", "foo":"bar"}
```

@murat1985 @irwaters @jonmoter 
/cc @zendesk/samson @mfischer-zd


murat1985:
using this tool is great development 👍

PR #1524, opened 2016-12-08
lock description is too small for html

grosser:
https://zendesk.airbrake.io/projects/95346/groups/1836367035064855865?tab=occurrences

@abmartinr


abmartinr:
:+1: on 🍏

PR #1528, opened 2016-12-08
injecting the vault server address into all containers if we have one associated with the deploy group

irwaters:
if a vault server is associated with a deploy group, then all containers get it in their ENV

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -64,14 +64,13 @@ module Kubernetes
     # /vaultauth is a secrets volume in the cluster
     # /secretkeys are where the annotations from the config are mounted
     def set_secret_sidecar
-
       containers.push(
         image: SIDECAR_IMAGE,
         name: 'secret-sidecar',
         volumeMounts: [
           { mountPath: "/vault-auth", name: "vaultauth" },
           { mountPath: "/secretkeys", name: "secretkeys" }
-        ],
+        ]
       )
 
       # share secrets volume between all containers
@@ -196,13 +195,12 @@ module Kubernetes
       env.merge!(Samson::Hooks.fire(:deploy_group_env, project, @doc.deploy_group).inject({}, :merge!))
     end
 
-
     def set_vault_env
       if needs_vault?
         containers.each do |container|
           env = (container[:env] ||= [])
           vault_client = Samson::Secrets::VaultClient.client.client(@doc.deploy_group)
-          env << {name: "VAULT_ADDR", value:vault_client.options.fetch(:address)}
+          env << {name: "VAULT_ADDR", value: vault_client.options.fetch(:address)}
           env << {name: "VAULT_SSL_VERIFY", value: vault_client.options.fetch(:ssl_verify).to_s}
         end
       end
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -63,9 +63,6 @@ module Kubernetes
     # /vaultauth is a secrets volume in the cluster
     # /secretkeys are where the annotations from the config are mounted
     def set_secret_sidecar
-      unless vault_client = Samson::Secrets::VaultClient.client.client(@doc.deploy_group)
-        raise "Could not find Vault config for #{@doc.deploy_group.permalink}"
-      end
 
       containers.push(
         image: SIDECAR_IMAGE,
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -160,26 +153,28 @@ def project
...
-        POD_NAMESPACE: 'metadata.namespace',
-        POD_IP: 'status.podIP'
-      }.each do |k, v|
-        env << {
-         name: k.to_s,
-         valueFrom: {fieldRef: {fieldPath: v}}
-       }
-      end
+      containers.each do |cont|
+        env = (cont[:env] ||= [])
```
grosser: added env is always the same ... we can prepare that outside the look and then only do add+unique inside ?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -198,6 +193,14 @@ def static_env
       kube_cluster_name = DeployGroup.find(metadata[:deploy_group_id]).kubernetes_cluster.name.to_s
       env[:KUBERNETES_CLUSTER_NAME] = kube_cluster_name
 
+      # the vault server to be used if we are using vault and there's an instance associated
+      # with the deploy group
+      if @doc.deploy_group.vault_server_id
```
grosser: might be better to fail fast here and not after the deploy is underway

-> require there to be a vault server if secrets are used and vault backend is used ?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -171,9 +164,9 @@ def set_env
         POD_IP: 'status.podIP'
       }.each do |k, v|
         env << {
-         name: k.to_s,
-         valueFrom: {fieldRef: {fieldPath: v}}
-       }
+        name: k.to_s,
+        valueFrom: {fieldRef: {fieldPath: v}}
+      }
```
grosser: wrong indent ?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -202,6 +195,17 @@ def static_env
       env.merge!(Samson::Hooks.fire(:deploy_group_env, project, @doc.deploy_group).inject({}, :merge!))
     end
 
+    def set_vault_env
+      if needs_vault?
+        containers.each do |container|
+          env = (container[:env] ||= [])
+          vault_client = Samson::Secrets::VaultClient.client.client(@doc.deploy_group)
+          env << {name: "VAULT_ADDR", value: vault_client.options.fetch(:address)}
```
grosser: @doc.deploy_group.vault_sever.address / .ssl_verify

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -187,9 +181,25 @@
       end
 
       it "fails when vault is not configured" do
-        Samson::Secrets::VaultClient.client.expects(:client).returns(nil)
-        e = assert_raises { template.to_hash }
-        e.message.must_equal "Could not find Vault config for pod1"
+        with_env('SECRET_STORAGE_BACKEND': "SecretStorage::HashicorpVault") do
```
grosser: can use regular symbols `SECRET_STORAGE_BACKEND:`

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -187,9 +181,25 @@
...
+          Samson::Secrets::VaultClient.client.expects(:client).raises("Could not find Vault config for pod1")
+          e = assert_raises { template.to_hash }
+          e.message.must_equal "Could not find Vault config for pod1"
+        end
+      end
+
+      it "get the VAULT_ADDR in it's env" do
```
grosser: when test description gives background information instead of repeating the test code it is easier to understand for newcomers and can reveal logical errors in the test code itself:
`it "adds then vault server address to the containers env"`

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -187,9 +181,25 @@
...
+        with_env('SECRET_STORAGE_BACKEND': "SecretStorage::HashicorpVault") do
+          assert template.to_hash[:spec][:template][:spec][:containers].
+            first[:env].any? { |env| env.any? { |_k, v| v == "VAULT_ADDR" } }
+        end
+      end
+
+      it "does not get the VAULT_ADDR in it's env" do
+        with_env('SECRET_STORAGE_BACKEND': "foobar") do
+          refute template.to_hash[:spec][:template][:spec][:containers].
```
grosser: reused logic ... extract into a let ... otherwise could lead to false positives when someone only updates one of them

grosser:
feels a bit weird ... but I guess reusing the logic is nice ...

PR #1532, opened 2016-12-10
simplify and dry up job execution logic and queing

grosser:
this started as a simple refactor ... but I got drawn deeper and deeper into the rabbit hole ...

 - clearer names
 - JobQueue is responsible for queuing, others do not know about internals
 - fix memory leak in JobQueue
 - reuse 'waiting for restart' logic
 - fix that executing but inactive jobs were not shown
 - fix new jobs during restarts being stored in registry and never getting removed
 - full test coverage for JobQueue

@sandlerr @jonmoter @anthonywoo

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -27,9 +27,9 @@ class Deploy < ActiveRecord::Base
     [super, commit]
   end
 
-  def job_execution_key
-    # if stage can run in parallel - set key to a unique id so it can immediately execute
-    stage.run_in_parallel ? "deploy-#{id}" : "stage-#{stage.id}"
+  # queue deploys on stages that cannot execute in parallel
+  def job_execution_queue_name
+    "stage-#{stage.id}" unless stage.run_in_parallel
   end
 
   def summary
--- a/app/models/job_queue.rb
+++ b/app/models/job_queue.rb
@@ -1,55 +1,49 @@
 # frozen_string_literal: true
+# make jobs with the same queue run in serial and track their status
 class JobQueue
   LOCK = Mutex.new
 
   def initialize
-    @queue = Hash.new { |h, k| h[k] = [] }
+    @queue = Hash.new { |h, q| h[q] = [] }
     @active = {}
-    @registry = {}
   end
 
   def active
     @active.values
   end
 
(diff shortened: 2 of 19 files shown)


app/models/deploy.rb
```diff
@@ -27,9 +27,10 @@ def cache_key
     [super, commit]
   end
 
-  def job_execution_key
-    # if stage can run in parallel - set key to a unique id so it can immediately execute
-    stage.run_in_parallel ? "deploy-#{id}" : "stage-#{stage.id}"
+  # use a unique queue for stages that can execute in parallel so they start immediately
+  # job_id is what JobExecution also internally uses as queue for 'normal' jobs
+  def job_execution_queue
```
sandlerr: `job_execution_queue_name` since it doesn't return a `JobQueue` instance

app/models/job_queue.rb
```diff
@@ -1,55 +1,48 @@
...
-  def active?(key, id)
-    @active[key].try(:id) == id
+  def active?(id)
+    active.detect { |je| je.id == id }
   end
 
-  def queued?(key, id)
-    @queue.dup[key].any? { |je| je.id == id }
+  def queued?(id)
+    @queue.values.detect { |jes| jes.detect { |je| return je if je.id == id } }
```
sandlerr: wasn't `any?` better here and in `active?` I'd expect this to return boolean

grosser:
Hehe yeah needed the return value for find-by-id :)

grosser:
green locally and green in docker ... damn it :/

app/models/job_queue.rb
grosser: It's either having a `queued(id)` and `queued?(id)` method or this non-boolean `?` method ... you can pick ... I don't care much :D

grosser:
found it ... multilock was left open and test order changed ... good fun ...all green now :)

app/models/job_queue.rb
sandlerr: I'd go for having both methods to be more conventional, not a big deal.

grosser:
hope these lock / job fixes make our build more stable too ...

PR #1533, opened 2016-12-11
do not print all tests to stdout when running tests in docker

grosser:
otherwise seeing lots of this:
```
-------------------------------------------------
.------------------------------------------------------------------
JobExecution::#stop!: test_0001_stops the execution with interrupt
------------------------------------------------------------------
.-------------------------------------------------------------------------------------------------
JobExecution::#stop!: test_0002_stops the execution with kill if job has already been interrupted
-------------------------------------------------------------------------------------------------
.

```

@sandlerr


sandlerr:
do you still get green dots? It's caused by this module: https://github.com/rails/rails/blob/master/activesupport/lib/active_support/testing/tagged_logging.rb wasn't sure how to get rid of it

sandlerr:
I think the verbose output that prints the name of each test is generated by ActiveSupport::Testing::TaggedLogging, is there a proper way to turn it off? Green dots is great.

grosser:
it is off ... only seeing green dots atm ...

sandlerr:
:+1: but setting the Rails log level to `warn` seems better to me - isn't it possible that this loses other logs that someone needs?

grosser:
this is only in test env ... and it makes the logs go to the log as they would in a regular rails setup ... the log_to_stdout is only meant for server mode anyway ... so I think this is a good step forward ... they ideal logic would be `when in server mode` ... but for now this is 1 step better :D

PR #1542, opened 2016-12-13
Add routing support for releases with dots

adam-hart:
After the changes from https://github.com/zendesk/samson/pull/1482, we allow releases with dots (ex v1.0).  The routing didn't support this, and it interpreted the `v1` as the `id`, and the `0` as the `format`.  This adds an explicit regex for the `id`, and restricts the format to json.

### Tasks
 - [ ] :+1: from team

### References
 - https://github.com/zendesk/samson/pull/1482

### Risks
- Level: Low:  Could cause release failures.

### DIFF
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -3,6 +3,7 @@ class Release < ActiveRecord::Base
   NUMBER = '\d+(:?\.\d+)*'
   NUMBER_REGEX = /\A#{NUMBER}\z/
   VERSION_REGEX = /\Av(#{NUMBER})\z/
+  ROUTE_REGEX = /v(#{NUMBER})/
 
   belongs_to :project, touch: true
   belongs_to :author, polymorphic: true
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -56,7 +56,7 @@ Samson::Application.routes.draw do
       end
     end
 
-    resources :releases, only: [:show, :index, :new, :create], id: /v(\d+(:?\.\d+)*)/, format: /json/
+    resources :releases, only: [:show, :index, :new, :create], id: Release::ROUTE_REGEX
 
     resources :stages do
       collection do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -56,7 +56,7 @@ Samson::Application.routes.draw do
       end
     end
 
-    resources :releases, only: [:show, :index, :new, :create]
+    resources :releases, only: [:show, :index, :new, :create], id: /v(\d+(:?\.\d+)*)/, format: /json/
 
     resources :stages do
       collection do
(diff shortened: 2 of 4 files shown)


test/controllers/releases_controller_test.rb
```diff
@@ -6,17 +6,23 @@
 describe ReleasesController do
   let(:project) { projects(:test) }
   let(:release) { releases(:test) }
+  let(:release_dot) { releases(:test_dot) }
```
grosser: prefer modifying an existing release instead of adding a new fixture every test has to load

config/routes.rb
```diff
@@ -56,7 +56,7 @@
       end
     end
 
-    resources :releases, only: [:show, :index, :new, :create]
+    resources :releases, only: [:show, :index, :new, :create], id: /v(\d+(:?\.\d+)*)/, format: /json/
```
grosser: `Release::VERSION_REGEX`

grosser: format looks like it only allows json ... but we need `no format` too ?

jcheatham:
👍  LGTM

config/routes.rb
adam-hart: I can't use the `VERSION_REGEX` because routes don't allow string anchors, but I can create a new one without them.  Also do we need `no format`?  It defaults to JSON, right?

grosser: ^/$ also does not work ? ... last resort would to surround with `\b`

config/routes.rb
```diff
@@ -56,7 +56,7 @@
       end
     end
 
-    resources :releases, only: [:show, :index, :new, :create]
+    resources :releases, only: [:show, :index, :new, :create], id: Release::ROUTE_REGEX, format: /json|/
```
grosser: do you need to give a format regex ?

test/controllers/releases_controller_test.rb
```diff
@@ -12,7 +12,14 @@
...
     describe "#show" do
-      it "renders" do
+      it "renders continuous versions" do
+        get :show, params: {project_id: project.to_param, id: release.version}
+        assert_response :success
+      end
+
+      it "renders major-minor versions" do
+        release.number = "12.3"
+        release.save!
```
grosser: `.update_column :number, '12.3'`

config/routes.rb
adam-hart: Nope!  I guess not.

app/models/release.rb
```diff
@@ -3,6 +3,7 @@ class Release < ActiveRecord::Base
   NUMBER = '\d+(:?\.\d+)*'
   NUMBER_REGEX = /\A#{NUMBER}\z/
   VERSION_REGEX = /\Av(#{NUMBER})\z/
+  ROUTE_REGEX = /v(#{NUMBER})/
```
grosser: surround with `\b` ?

pdeuter:
:chicken:

PR #1546, opened 2016-12-13
document that we need write permissons

grosser:
@jonmoter

### DIFF
--- a/test/models/release_service_test.rb
+++ b/test/models/release_service_test.rb
@@ -4,57 +4,77 @@ require_relative '../test_helper'
 SingleCov.covered!
 
 describe ReleaseService do
+  include StubGithubAPI
+
   let(:project) { projects(:test) }
-  let(:author) { users(:deployer) }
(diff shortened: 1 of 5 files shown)


jonmoter:
How about when you specify a release branch, we check whether we have enough permissions, and give you a warning if we don't?

grosser:
updated to check on save when release_branch gets set:

screen shot 2016-12-14 at 11 35 44 am

test/models/release_service_test.rb
```diff
@@ -4,57 +4,77 @@
...
   let(:project) { projects(:test) }
-  let(:author) { users(:deployer) }
   let(:service) { ReleaseService.new(project) }
-  let(:commit) { "abcd" * 10 }
-  let(:release_params_used) { [] }
 
-  before do
-    GITHUB.stubs(:create_release).capture(release_params_used)
-  end
+  describe "#release!" do
```
grosser: diff is moving all the existing tests into `descibe "release!"`

PR #1549, opened 2016-12-14
Improve Heroku Deploy

chitsaou:
* Added `template` parameter to Heroku Deploy URL so it can load `app.json` automatically
* Mark some parameters as not required
* Workaround for post-deploy script by adding `DISABLE_DATABASE_ENVIRONMENT_CHECK=1`
  * Not sure if this is the right way to solve the problem, though

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app.json
+++ b/app.json
@@ -63,7 +63,7 @@
     }
   },
   "scripts": {
-    "postdeploy": "bundle exec rake db:schema:load"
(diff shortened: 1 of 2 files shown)


grosser:
❤️

PR #1551, opened 2016-12-14
Adds a 'hand' mouse pointer to the Warning labels

mariozaizar:
/cc @zendesk/samson

### Description

Adds `?` mouse pointer to any label (with a tooltip) to help indicate there is a descriptive message when the mouse is hover.  

*Before:*
screen shot 2016-12-14 at 12 42 19 pm

*After:*
screen shot 2016-12-14 at 12 39 07 pm

### Tasks
 - [x] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/assets/stylesheets/projects.scss
+++ b/app/assets/stylesheets/projects.scss
@@ -169,3 +169,6 @@ pre.pre-inline {
     margin-right: 10px;
   }
 }
+
+// Labels with tooltips.
+span.label[title] { cursor: help; }


grosser:
hmm Ideally have it for `span[title]` or so ?

mariozaizar:
You're right! Ok, new code works with other labels (like `.label-info`) and only if the `span` contains a `title` attribute in it.

PR #1552, opened 2016-12-14
use with_connection to avoid "Cannot lease connection"

grosser:
@sandlerr @jonmoter 

After adding clair this old friend popped up again:

https://zendesk.airbrake.io/projects/95346/groups/1794866641583417835/notices/1840523327485950109
Cannot lease connection, it is already in use by a different thread

per discussion in https://github.com/rails/rails/issues/25585
ActiveRecord::Base.connection_pool.with_connection is preferred and seems cleaner to me too
so let's give that a try ...

### DIFF
--- a/lib/samson/clair.rb
+++ b/lib/samson/clair.rb
@@ -12,6 +12,7 @@ module Samson
       def append_job_with_scan(job, docker_tag)
         return unless clair = ENV['HYPERCLAIR_PATH']
 
+        ActiveRecord::Base.clear_active_connections!
         Thread.new do
           sleep 0.1 if Rails.env.test? # in test we reuse the same connection, so we cannot use it at the same time
           success, output, time = scan(clair, job.project, docker_tag)


grosser:
failed on travis ... reverting to ActiveRecord::Base.clear_active_connections!

lib/samson/clair.rb
```diff
@@ -12,6 +12,7 @@ class << self
       def append_job_with_scan(job, docker_tag)
         return unless clair = ENV['HYPERCLAIR_PATH']
 
+        ActiveRecord::Base.clear_active_connections!
```
sandlerr: could this go inside the thread? that way if the bug wasn't triggered it will not actually release anything?

grosser: this is to avoid the bug, at this point we don't know if there is a bug ... same code as https://github.com/zendesk/samson/blob/master/app/models/job_execution.rb#L43-L45
afaik have to do this before using db in a new thread

sandlerr: right, I think this should be the first line inside the Thread.new block though

grosser:
docs say: `Returns any connections in use by the current thread back to the pool, and also returns connections to the pool cached by threads that are no longer alive.` 

so calling it in the new thread would do nothing ...

grosser:
more details on https://bibwild.wordpress.com/2014/07/17/activerecord-concurrency-in-rails4-avoid-leaked-connections/ ... basically everybody saying we should use with_connection :(
... good enough for now ... we can revisit if it ever pops up again :)

sandlerr:
hm, that blog post suggests that we need to release the connection back to the pool manually at the end of the thread

grosser:
yeah ... I'll stay with using what worked before :D

PR #1562, opened 2016-12-17
Move interesting code into TerminalExecutor#execute! instead of rando…

grosser:
@irwaters

### DIFF
--- a/app/models/terminal_executor.rb
+++ b/app/models/terminal_executor.rb
@@ -26,44 +26,8 @@ class TerminalExecutor
 
   def execute!(*commands)
     return false if @stopped
-    if @verbose
-      commands.map! { |c| "echo » #{c.shellescape}\n#{resolve_secrets(c)}" }
-    else
-      commands.map! { |c| resolve_secrets(c) }
-    end
-    commands.unshift("set -e")
-
-    execute_command!(commands.join("\n"))
-  end
-
-  def stop!(signal)
-    @stopped = true
-    system('kill', "-#{signal}", "-#{pgid}") if pgid
-  end
-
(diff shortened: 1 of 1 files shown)


grosser:
@sandlerr while you on a roll ;)

sandlerr:
while you're refactoring this, is there any way to make the stdout not report as a tty? That would clean up our `npm install` logs quite a lot

was experimenting passing in different `in` and `out` values to `PTY.spawn` but couldn't figure out how to make it not block endlessly

```ruby
irb(main):001:0> require 'pty'
=> true
irb(main):002:0> output, input, pid = PTY.spawn("ruby -e 'puts STDOUT.tty?'")
=> [#<File:/dev/ttys003>, #<File:/dev/ttys003>, 61860]
irb(main):003:0> Process.wait(pid, Process::WNOHANG)
=> nil
irb(main):004:0> output.read
=> "true\r\n"
irb(main):005:0> output, input, pid = PTY.spawn("ruby -e 'puts STDOUT.tty?' | cat")
=> [#<File:/dev/ttys005>, #<File:/dev/ttys005>, 61866]
irb(main):006:0> Process.wait(pid, Process::WNOHANG)
=> nil
irb(main):007:0> output.read
=> "false\r\n"
```

sandlerr:
I'd like to avoid commands sending carriage returns to the deploy logs

grosser:
we used PTY.spawn to get this 'feature' ... we want colors ... :/
... otherwise switch to IO.popen like lib/samson/command_executor.rb

grosser:
... also we want to see git progress ...

PR #1563, opened 2016-12-17
make stopping an execution call hooks once

grosser:
- only call finish from either ensure or stop!
 - avoid db updates from possibly killed thread that might have a borked db connection
 - do not call finish twice when stopping multiple times
 - do not mark job as cancelled when it finished regularly

This should solve the bugs we saw with active threads being unable to be found
(see JobQueue#delete_and_enqueue_next)

@sandlerr @jonmoter @irwaters

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 class Job < ActiveRecord::Base
   belongs_to :project
-  belongs_to :user, -> { unscope(where: 'deleted_at') }
+  belongs_to :user, -> { unscope(where: :deleted_at) }
 
   has_one :deploy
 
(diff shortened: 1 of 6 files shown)


app/models/job.rb
```diff
@@ -20,7 +20,7 @@ def self.valid_status?(status)
   end
 
   def self.non_deploy
-    includes(:deploy).where(deploys: { id: nil })
+    joins('left join deploys on deploys.job_id = jobs.id').where(deploys: { id: nil })
```
grosser: making this cheaper ... the include triggered join fetches all the columns, which we don't need

app/models/job_execution.rb
```diff
@@ -57,13 +60,13 @@ def close
   end
 
   def stop!
-    if @thread
-      @executor.stop! 'INT'
-      return if @thread.join(stop_timeout)
+    return unless @thread&.alive?
```
grosser: stopping when it is not running should never happen ... could also raise here ... basically trying to avoid that a failed/succeeded job gets changed to cancelled

sandlerr: maybe add a manual airbrake?

grosser:
afaik this is the user clicking stop like crazy ... so not airbrake worthy
:)

PR #1566, opened 2016-12-19
PAAS-533 convert secret sidecar into secret init container

grosser:
@irwaters @jonmoter


grosser:
confirmed working on staging with truth service and manually added `SECRET_PULLER_IMAGE="zendesk/samson_secret_puller:latest"`
```
cat /secrets/TEST
master
```

PR #1567, opened 2016-12-20
only fail if readiness probe repeatedly fails

grosser:
@zenvdeluca @jonmoter @irwaters 

truth-service failed the deploy because the probe sends a warning

```
"reason": "Unhealthy",
            "message": "Readiness probe failed: Get http://1.2.3.4:4242/: dial tcp 1.2.3.4:4242: getsockopt: connection refused",
```

also made it ignore Liveliness probe just in case ...
I think this gives us 10s time to get ready since the default interval is 1s
... looking deeper into the docs I think we can just set failureThreshold: 100 ... I'll try that next before merging ...

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -59,8 +59,12 @@ module Kubernetes
         end
       end
 
-      def abnormal_events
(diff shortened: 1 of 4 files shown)


grosser:
... failureThreshold does not help :/
... but I guess I can read the config and mark it as failed only after failureThreshold is exceeded ...

grosser:
done ... looks good ?

PR #1573, opened 2016-12-21
bump rails to 5.0.1

grosser:
@irwaters @jonmoter


grosser:
boot time is about the same, so the rake patch did not help much ...

found 1 minor bug I could report: https://github.com/rails/rails/issues/27426
and 1 wonky thing that was really hard to reproduce ... something about first_or_create in combination with maximum ... just rewrote it in simpler terms and that works ... can re-check on the next upgrade

grosser:
test deploy worked and deploy from staging worked too ... looks good ...

sandlerr:
@grosser this should have fixed the issue in https://github.com/zendesk/samson/pull/1552

> Ensure concurrent invocations of the connection reaper cannot allocate the same connection to two threads.
Fixes #25585.
Matthew Draper

https://github.com/rails/rails/blob/v5.0.1/activerecord/CHANGELOG.md

PR #1575, opened 2016-12-21
record history for environment variables

grosser:
@irwaters @jonmoter 

screen shot 2016-12-21 at 10 27 55 am
screen shot 2016-12-21 at 10 28 02 am


plugins/env/app/models/environment_variable.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 class EnvironmentVariable < ActiveRecord::Base
+  has_paper_trail
```
irwaters: super useful thing that.

PR #1578, opened 2016-12-22
add ability to stop builds

grosser:
screen shot 2016-12-21 at 4 38 35 pm
screen shot 2016-12-21 at 4 46 54 pm

 - make jobs stoppable by whoever started them
 - unify deploy/build/job 'executing' logic
 - add stop button that directs user back to the build page

@jonmoter @irwaters @sandlerr

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -4,7 +4,8 @@ class JobsController < ApplicationController
 
   skip_before_action :require_project, only: [:enabled]
 
-  before_action :authorize_project_admin!, only: [:new, :create, :destroy]
+  before_action :authorize_project_admin!, only: [:new, :create]
+  before_action :authorize_project_deployer!, only: [:new, :create, :destroy]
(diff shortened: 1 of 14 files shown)


app/views/builds/show.html.erb
```diff
@@ -45,24 +45,24 @@
 
   <hr/>
 
-  <% job = @build.docker_build_job %>
-  <% if job %>
+  <% if job = @build.docker_build_job %>
```
sandlerr: I'm guessing rubocop misses this on a technicality since its erb?

app/controllers/jobs_controller.rb
```diff
@@ -51,15 +52,14 @@ def enabled
   end
 
   def destroy
```
sandlerr: is this action called `destroy` elsewhere? we aren't marking a record as deleted, not sure if DELETE is the right http verb.

grosser:
we never want to delete a job/deploy ... so abusing this :D
same in deploys controller :)

PR #1580, opened 2016-12-22
PAAS-633 push to multiple images

grosser:
for now only supporting pushing without or with same credentials

 - pushing to extra registries is done without namespacing ... not sure if that is what we want (given that it won't work for multiple ECR registries), but it is what would make the transition easiest ...
 - ECR auth will only perform for the first registry ... this is tricky to implement for all since we atm store credentials in the ENV ... so will need to rework if we ever want multiple ECR registries

@jonmoter @irwaters 

confirmed working with https://samsontest.zende.sk/projects/truth_service/builds/470

FYI we do not push our automated-build-triggered-via-deploy-4131 tags

simplified testing and made it always list all pushes/tags
fixed bugs:
 - where tagging something as latest would not force-push that tag.
 - when pushing failed but digest was created it did not stop fail build


grosser:
all working now :D

catkins:
:+1: nice one

PR #1581, opened 2016-12-22
Make release sources configurable

adam-hart:
The Samson releases feature by default uses any webhook for a project to create a release.  This keeps the default behavior, but adds an option to restrict the release to a single webhook source (Travis for example).  This means that you could wait for a green build from CI before tagging a new release.

### Risks
- Medium:  Could cause release feature to break.

### DIFF
--- a/app/helpers/webhooks_helper.rb
+++ b/app/helpers/webhooks_helper.rb
@@ -1,11 +1,20 @@
 # frozen_string_literal: true
 module WebhooksHelper
-  def webhook_sources(sources)
-    [
-      ['Any CI', 'any_ci'],
-      ['Any code push', 'any_code'],
-      ['Any Pull Request', 'any_pull_request'],
-      ['Any', 'any']
-    ] + sources.map { |source| [source.titleize, source] }.to_a
+  GENERIC_SOURCES = [
+    ['Any CI', 'any_ci'],
+    ['Any code push', 'any_code'],
+    ['Any Pull Request', 'any_pull_request'],
+    ['Any', 'any']
+  ].freeze
+
+  NO_SOURCE = [['None', 'none']].freeze
+
+  def webhook_sources_select(sources, none: false)
+    default_sources = none ? GENERIC_SOURCES + NO_SOURCE : GENERIC_SOURCES
+    default_sources + sources_for_select(sources)
+  end
+
+  def sources_for_select(sources)
+    sources.map { |source| [source.titleize, source] }.to_a
   end
 end
--- a/app/helpers/webhooks_helper.rb
+++ b/app/helpers/webhooks_helper.rb
@@ -9,7 +9,7 @@ module WebhooksHelper
 
   NO_SOURCE = [['None', 'none']].freeze
 
-  def webhook_sources_select(sources, none: false)
+  def webhook_sources_for_select(sources, none: false)
     default_sources = none ? GENERIC_SOURCES + NO_SOURCE : GENERIC_SOURCES
     default_sources + sources_for_select(sources)
   end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -58,8 +58,12 @@ class Project < ActiveRecord::Base
   # branch - The String name of the branch in question.
   #
   # Returns true if new releases should be created, false otherwise.
-  def create_releases_for_branch?(branch)
-    release_branch == branch
+  def create_releases_for_branch?(branch, service_type, service_name)
+    release_branch == branch && release_source_included?(service_type, service_name)
+  end
+
+  def release_source_included?(service_type, service_name)
+    release_source == 'any' || release_source == "any_#{service_type}" || release_source == service_name
   end
 
   # Whether to create a new docker image when the branch is updated.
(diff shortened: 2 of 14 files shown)


grosser:
remove the github webhook ... problem solved ?

adam-hart:
The Github webhook is currently being used for things in samson other than releases.  I think we should be able to have more than one webhook and still maintain functionality.

grosser:
what do we use hooks for other than releases/deploys ?

grosser:
how about we only show this field if you have 2+ webhooks to keep the UI simple ?

app/models/project.rb
```diff
@@ -55,8 +55,12 @@ def docker_repo
   # branch - The String name of the branch in question.
   #
   # Returns true if new releases should be created, false otherwise.
-  def create_releases_for_branch?(branch)
-    release_branch == branch
+  def create_releases_for_branch?(branch, service_type, service_name)
```
grosser: might be better to rename this to `create_release?(branch, service_type, service_name)`

app/models/project.rb
```diff
@@ -55,8 +55,12 @@ def docker_repo
...
+    release_branch == branch && release_source_included?(service_type, service_name)
+  end
+
+  def release_source_included?(service_type, service_name)
```
grosser: I bet we already have this logic somewhere ... reuse ?

app/views/projects/_form.html.erb
```diff
@@ -8,6 +8,12 @@
     <%= form.input :repository_url, required: true %>
     <%= form.input :permalink, required: true if project.persisted? %>
     <%= form.input :release_branch, help: 'New commits on this branch will cause a release when a webhook arrives.' %>
+    <div class="form-group">
```
grosser: best to add some help text like 'select which type of webhook will trigger a release'

app/views/projects/_form.html.erb
```diff
@@ -8,6 +8,12 @@
...
+      <%= form.label "Release source", class: "col-lg-2 control-label" %>
+      <div class="col-lg-4">
+        <%= form.collection_select :release_source, webhook_sources(Samson::Integration::SOURCES), :second, :first, {}, class: "form-control" %>
```
grosser: `webhook_sources`  -> `webhook_sources_select_option` or so ...

test/controllers/integrations/base_controller_test.rb
```diff
@@ -42,6 +42,31 @@ class BaseTestController < Integrations::BaseController
       project.releases.count.must_equal 1
     end
 
+    describe 'when the release branch source is defined' do
```
grosser: it's always defined ... so maybe defined -> specific

app/views/projects/_form.html.erb
grosser: also none should not be an option ...

app/models/project.rb
adam-hart: This was the only place I could find it, and it's just a query.

https://github.com/zendesk/samson/blob/master/app/models/webhook.rb#L18

grosser:
then leave the logic in Webhook ... maybe as static method ... so it's all
together at least / easy to fin when someone tries to change

app/helpers/webhooks_helper.rb
```diff
@@ -1,11 +1,23 @@
...
-      ['Any', 'any']
-    ] + sources.map { |source| [source.titleize, source] }.to_a
+    titleize_sources(sources) + GENERIC_SOURCES
+  end
+
+  def webhook_sources_select_option(sources)
+    titleize_sources(sources) + GENERIC_SOURCES << NO_SOURCE
+  end
+
+  def titleize_sources(sources)
```
grosser: `sources_for_select` ?

grosser: hmmm idk this is getting weird ... only have a single webhook_sources_for_select method that takes a `none: true|false` option ?

app/models/webhook.rb
```diff
@@ -17,4 +17,8 @@ def self.for_branch(branch)
   def self.for_source(service_type, service_name)
     where(source: ['any', "any_#{service_type}", service_name])
   end
+
+  def self.source_matches?(release_source, service_type, service_name)
```
grosser: need unit test in webhook_test.rb

app/helpers/webhooks_helper.rb
grosser: also needs unit test

app/views/projects/_form.html.erb
grosser: use `input` with `help` options and `do` to generate custom fields

test/models/webhook_test.rb
```diff
@@ -90,4 +90,24 @@
       assert_equal Webhook.for_branch('master').pluck(:branch), ['', 'master']
     end
   end
+
+  describe '.for_source' do
+    Samson::Integration::SOURCES.each do |release_source|
+      it 'is true when the source does match' do
```
grosser: `is true when the source is the same`

PR #1585, opened 2016-12-24
alert when we use AR in a thread without switching connections

grosser:
@zendesk/samson 

for example [this build](https://travis-ci.org/zendesk/samson/jobs/186449157) looks like we might be messing with our fixtures ... so let's try to alert when using threads without switching connections ... which is a dangerous thing to do anyway ...


grosser:
this sadly did not find anything new, but it blows up nicely if anyone uses AR inside a thread without with_connection ... so I guess more future proofing ...

PR #1588, opened 2016-12-27
dont update a release with build if the release doesnt exist

craig-day:
Projects might not be using the releases feature, but have a docker build branch set. This causes an error to happen when a build is run without a release being created. This will allow the build to run without blowing up on the missing release. 

An alternative option would be to not run the build if the release isn't created, but I think this is simpler. 

Example error:
```
{"method":"POST","path":"/integrations/travis/[TOKEN]","format":"*/*","controller":"integrations/travis","action":"create","status":500,"error":"NoMethodError: undefined method `update_attribute' for nil:NilClass"}
```

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Low. Docker builds will proceed without a release.

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -104,7 +104,7 @@ class Integrations::BaseController < ApplicationController
 
   def create_docker_image(release)
     build = find_or_create_build(branch)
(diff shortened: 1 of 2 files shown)


grosser:
❤️

PR #1590, opened 2016-12-27
improve permisson and plugin docs

grosser:
/cc @zendesk/samson 

permission was talking about user and system-level ... which are the same, so unifying them.

plugins was out of date and should start with a simple list of available plugins to get people hooked.


docs/plugins.md
```diff
@@ -1,40 +1,48 @@
...
-To enable all plugins, use a value of "all":
+To enable all plugins, use "all":
 
 `PLUGINS="all"`
 
-To selectively disable plugins, use "all", and then a comma-separated list of
-plugins, with a minus sign in front of each:
+To disable selected plugins, use "all", and a comma-separated list of plugins, with a minus sign in front of each:
 
 `PLUGINS="all,-flowdock,-slack"`
```
jonmoter: Change to `slack_webhooks` as above.

grosser: done

PR #1591, opened 2016-12-28
rename previous_deploy to better reflect what it does

grosser:
@prudhvi @irwaters 

also fix rollback always rolling back


prudhvi:
👍  after 🍏

PR #1605, opened 2017-01-02
Add flag to reuse previous build so we can debug faster in non-prod

grosser:
@jonmoter @prudhvi @irwaters 

screen shot 2017-01-02 at 2 26 34 pm

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -233,22 +233,15 @@ module Kubernetes
     end
 
     def find_or_create_build
-      return unless build = (Build.find_by_git_sha(@job.commit) || create_build)
+      return unless build = (find_build || create_build)
       wait_for_build(build)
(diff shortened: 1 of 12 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -233,22 +233,15 @@ def print_statuses(status_groups)
...
-        loop do
-          break if @stopped
-          sleep TICK
-          break if build.docker_build_job.reload.finished?
-        end
-      end
-      build.reload
+    def find_build
+      Build.find_by_git_sha(@job.commit) ||
+        (@job.deploy.kubernetes_reuse_build && @job.deploy.previous_deploy&.kubernetes_release&.build)
```
prudhvi: Ah ruby 2.3 `try` equivalent .. seeing usage first time here 😄

prudhvi:
Don't know codebase much to understand the behaviour when both flags are on( is that a valid thing)
Otherwise looks good to me 👍 Thanks again @grosser , makes debugging life so much faster.

PR #1607, opened 2017-01-03
do not send binary builder errors to airbrake

grosser:
users fault ... so let them fix it ...

https://zendesk.airbrake.io/projects/95346/groups/1774910190152787640/notices/1854650037090094484

@henders @apanzerj @jonmoter 

also:
 - remove stubs and test real execution
 - test homegrown untar
 - use a temp dir that we clean up instead of polluting /tmp

### DIFF
--- a/plugins/docker_binary_builder/app/models/binary_builder.rb
+++ b/plugins/docker_binary_builder/app/models/binary_builder.rb
@@ -74,6 +74,7 @@ class BinaryBuilder
     untar(File.join(@dir, ARTIFACTS_FILE))
   end
 
+  # FIXME: does not set permissons ... reuse some library instead of re-inventing
   def untar(file_path)
     @output.puts "About to untar: #{file_path}"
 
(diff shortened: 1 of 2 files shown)


plugins/docker_binary_builder/app/models/binary_builder.rb
```diff
@@ -23,55 +23,47 @@ def build
...
   def start_build_script
-    @output_stream.puts 'Now starting Build container...'
-    @container.tap(&:start).attach { |_stream, chunk| @output_stream.write_docker_chunk(chunk) }
-  rescue => ex
-    @output_stream.puts "Failed to run the build script '#{BUILD_SCRIPT}' inside container."
-    raise ex
+    @output.puts 'Now starting Build container...'
+    @container.tap(&:start).attach { |_stream, chunk| @output.write_docker_chunk(chunk) }
+  rescue
+    raise Samson::Hooks::UserError, "Failed to run the build script '#{BUILD_SCRIPT}' inside container."
```
henders: Do we want to log the original exception to get a backtrace for debugging?

henders:
Approved with just a suggestion for logging.

grosser:
added logging and original error to the output that the user will see

```
    Rails.logger.error("Binary builder error:\n#{$!}\n#{$!.backtrace.join("\n")}")
    raise Samson::Hooks::UserError, "Failed to run the build script '#{BUILD_SCRIPT}' inside container.\n#{$!}"
```

PR #1608, opened 2017-01-03
PAAS-633 always namespace images

grosser:
@jonmoter 

I though we used no namespace ... but turns out we used a different namespace ... so better to unify and then update the image names everywhere

confirmed working on staging: https://samsontest.zende.sk/projects/truth_service/builds/470#L98


grosser:
@jonmoter updated to support N registries with their own credentials and namespaces

jonmoter:
How much work would it be to allow an upgrade path?  Anyone who's currently running Samson specifying `DOCKER_REGISTRY` will have it suddenly break, 'cause it's now using a different ENV var.

grosser:
updated to support deprecated env vars and warn

catkins:
@jonmoter @grosser has this change gone out to prod samon / the config updated?

I'm noticing that our docker builds are still going to the un-namespaced path

https://samson.zende.sk/projects/people/builds/20558#L391

grosser:
going out today

catkins:
Ah cool, thanks

grosser:
live now

PR #1609, opened 2017-01-03
make logs visible when debugging from integration service

grosser:
atm the response in githubs debug view is not very helpful ... and users need to dig into samson to find details ... this should make things simpler

@henders @jonmoter @irwaters 

### Risks: 
 - medium: a few integrations might expect an empty response, we'll have to add exceptions for these, but I'd start by assuming they don't care ...


henders:
Nice! More info is definitely better especially for people who don't have access to the samson logs

👍

PR #1613, opened 2017-01-03
make admin/users filterable by role

grosser:
@jonmoter 

screen shot 2017-01-03 at 2 33 37 pm

filtering projects role is more tricky and ugly ... this is a nice UI win and simple ... so let's do that for now ...

project roles on /admin/users/1 need a rewrite so that they only show the existing roles + a from to create a new role ... but that is another PR

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -5,22 +5,7 @@
 <section id="user-project-roles" class="tabs">
   <h2 class="section-subtitle">Project Level Roles</h2>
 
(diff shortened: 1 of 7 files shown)


app/views/users/index.html.erb
```diff
@@ -37,7 +22,7 @@
     </tr>
     </thead>
     <tbody>
-    <% if @users.empty? %>
+    <% if @users.length == 0 %>
```
grosser: small hack to avoid doing a count + find or the same page

PR #1619, opened 2017-01-05
re-dump schema to reduce diff for future migrations

grosser:
rails 5 bump changed the dumped schema a lot, so this should make things simpler when migrating


sandlerr:
did you run that locally or in production / staging? Dumping the prod schema would probably be safest?

grosser:
this is re-dumped dev seeds ... not trying to change/update them ... that can be a separate PR

PR #1625, opened 2017-01-05
Use Github tags if present

adam-hart:
If a release is about to be created, and it has a tag already defined on Github, use that one for a release instead of creating a new one in Samson.

### Risks
- Medium:  Could cause release numbers to go all sorts of wrong.

### DIFF
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -56,9 +56,15 @@ class Release < ActiveRecord::Base
     # release-number-from-ci plugin.
     return if number != DEFAULT_RELEASE_NUMBER && number.present?
 
-    latest_release_number = project.releases.last.try(:number) || "0"
+    raise "Unable to auto bump version" unless self.number = next_release_number
+  end
 
-    raise "Unable to auto bump version" unless self.number = latest_release_number.dup.sub!(/\d+$/) { |d| d.to_i + 1 }
+  def next_release_number
+    # If Github has a version tagged for this commit, use that version instead of ours
+    latest_samson_version = Gem::Version.new(project.releases.last.try(:number) || "0").to_s
+    latest_github_version = Gem::Version.new(project.repository.tag_from_ref(commit)).to_s
+
+    latest_github_version.present? ? latest_github_version : latest_samson_version.to_s.dup.sub!(/\d+$/) { |d| d.to_i + 1 }
   end
 
   def contains_commit?(other_commit)
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -59,18 +59,6 @@ class Release < ActiveRecord::Base
     errors.add :number, "Unable to auto bump version" unless self.number = next_release_number
   end
 
-  def next_release_number
-    # If Github has a version tagged for this commit, use that version instead of ours
-    latest_samson_version = Gem::Version.new(project.releases.last&.number || "0")
-    latest_github_version = Gem::Version.new(project.repository.tag_from_ref(commit))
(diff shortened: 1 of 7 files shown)


app/models/release.rb
```diff
@@ -56,9 +56,15 @@ def assign_release_number
     # release-number-from-ci plugin.
     return if number != DEFAULT_RELEASE_NUMBER && number.present?
 
-    latest_release_number = project.releases.last.try(:number) || "0"
+    raise "Unable to auto bump version" unless self.number = next_release_number
+  end
+
+  def next_release_number
+    # If Github has a version tagged for this commit, use that version instead of ours
+    latest_samson_version = Gem::Version.new(project.releases.last.try(:number) || "0").to_s
```
grosser: can use fancy new `. last&.number`

app/models/release.rb
```diff
@@ -56,9 +56,15 @@ def assign_release_number
     # release-number-from-ci plugin.
     return if number != DEFAULT_RELEASE_NUMBER && number.present?
 
-    latest_release_number = project.releases.last.try(:number) || "0"
+    raise "Unable to auto bump version" unless self.number = next_release_number
```
grosser: no need for self.

grosser: add an error instead of raising ?

app/models/release.rb
```diff
@@ -56,9 +56,15 @@ def assign_release_number
...
+  end
+
+  def next_release_number
+    # If Github has a version tagged for this commit, use that version instead of ours
+    latest_samson_version = Gem::Version.new(project.releases.last.try(:number) || "0").to_s
+    latest_github_version = Gem::Version.new(project.repository.tag_from_ref(commit)).to_s
 
-    raise "Unable to auto bump version" unless self.number = latest_release_number.dup.sub!(/\d+$/) { |d| d.to_i + 1 }
+    latest_github_version.present? ? latest_github_version : latest_samson_version.to_s.dup.sub!(/\d+$/) { |d| d.to_i + 1 }
```
grosser: what about that "and is greater then last" ?

app/models/release.rb
sandlerr: shouldn't redundant `self` be a rubocop ci error?

grosser:
yep

app/models/release.rb
adam-hart: Something is funny here.  I'm not sure if it needs an `attr_writer`, but if I remove the `self`, it doesn't stick.

app/models/release.rb
```diff
@@ -56,9 +56,19 @@ def assign_release_number
     # release-number-from-ci plugin.
     return if number != DEFAULT_RELEASE_NUMBER && number.present?
 
-    latest_release_number = project.releases.last.try(:number) || "0"
+    errors.add :number, "Unable to auto bump version" unless self.number = next_release_number
+  end
+
+  def next_release_number
```
grosser: private ?

test/models/release_test.rb
```diff
@@ -10,6 +10,10 @@
   let(:commit) { "abcde" * 8 }
 
   describe "create" do
+    before do
+      GitRepository.any_instance.stubs(:tag_from_ref).returns("")
```
grosser: why do we need to stub this ? ... too slow ?

adam-hart: Yeah, this request seems to fail left and right on Travis.

PR #1626, opened 2017-01-05
callback improvements

grosser:
saw these weird bugs in airbake about being unable to send emails ...
https://zendesk.airbrake.io/projects/95346/groups/1748045765307475064/notices/1856265870884964594?tab=notice-detail

turns out users put crazy stuff into the email field ...
also noticed that atm when 1 hook fails we stop all hooks which is bad

 - consolidate email sending logic
 - do not stop all after deploy hooks when one of them fails
 - validate that emails are not crazy

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -100,10 +100,6 @@ class Stage < ActiveRecord::Base
     notify_email_address.split(";").map(&:strip)
   end
 
-  def send_email_notifications?
-    notify_email_address.present?
-  end
-
   def global_name
     "#{name} - #{project.name}"
   end
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -29,6 +29,10 @@ class Stage < ActiveRecord::Base
 
   validates :name, presence: true, uniqueness: { scope: [:project, :deleted_at] }
 
+  # n emails separated by ;
+  email = '([^\s;]+@[^\s;]+)'
+  validates :notify_email_address, format: /\A#{email}((\s*;\s*)?#{email}?)*\z/, allow_blank: true
+
   before_create :ensure_ordering
   after_destroy :destroy_deploy_groups_stages
   after_destroy :destroy_stage_pipeline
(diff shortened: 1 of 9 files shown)


app/models/stage.rb
```diff
@@ -28,6 +28,7 @@ class Stage < ActiveRecord::Base
   default_scope { order(:order) }
 
   validates :name, presence: true, uniqueness: { scope: [:project, :deleted_at] }
+  validates :notify_email_address, format: /\A([^;]+@[^;]+)(; ?([^;]+@[^;]+)?)*\z/, allow_blank: true
```
sandlerr: wow, is this format documented? I know its normal for emails but nothing else is separated by semicolons. Also, could replace ` ?` by `\W*` if I understand the intention, and maybe replace `[^;]` by `[^\W;]` (or maybe we shouldn't do this by regex?)

grosser:
I'm trying to say 'email;email' ... don't know why I would want \W in there
... I'll update with some docs ... it sure looks kind of funky

grosser:
UI already has `Enter the email addresses that should be notified of new deploys, separated by <code>;</code>`

sandlerr:
I'm suggesting two things: disallow whitespace within an email, allow arbitrary whitespace between emails

grosser:
updated

app/models/stage.rb
```diff
@@ -29,6 +29,10 @@ class Stage < ActiveRecord::Base
 
   validates :name, presence: true, uniqueness: { scope: [:project, :deleted_at] }
 
+  # n emails separated by ;
+  email = '([^\s;]+@[^\s;]+)'
+  validates :notify_email_address, format: /\A#{email}((\s*;\s*)?#{email}?)*\z/, allow_blank: true
```
sandlerr: nitpick: put the `?` directly on `;`, then you don't need a capture group

grosser: that would allow `email   email`

PR #1630, opened 2017-01-05
remove whitelisting of env vars while creating .env

grosser:
This got us into trouble when trying to deploy samson itself with a .env file that was leftover ... so I'd rather get rid of this magic / crazy feature until we figure out a better way to do it ... maybe .env.required or something like that ...

@jonmoter @irwaters


grosser:
any preference on how to do required keys ? ...
I felt it kind of weird that `.env` suddenly means required keys ... not very intuitive ...

irwaters:
don't have an real good thoughts about this other than having a LIST of them somewhere that we verify are in ENV.

PR #1633, opened 2017-01-06
parallelize secret listing for faster index page on cold cache

grosser:


lib/samson/secrets/vault_client.rb
```diff
@@ -20,9 +20,9 @@ def read(key)
       # different servers have different keys so combine all
       def list_recursive(path)
         path = wrap_key(path)
-        all = clients.each_value.flat_map do |vault|
-          with_retries { vault.logical.list_recursive(path) }
-        end
+        all = clients.each_value.map do |vault|
+          Thread.new { with_retries { vault.logical.list_recursive(path) } }
+        end.map(&:value).flatten(1)
```
jonmoter: If you call `Thread.new { ... }`, does that block until the thread completes?  Don't you have to make some sort of `join` call to wait for all the threads to complete?  I think you want something more like:

```
values = Concurrent::Array.new
threads = clients.each_value.map do |vault|
  Thread.new { with_retries { values += vault.logical.list_recursive(path) } }
end
threads.each { |thr| thr.join }
```

grosser:
.new does not block
.value is .join + 'give me your result'

jonmoter:
Ahhh, didn't know that's how `value` worked.  TIL.

PR #1635, opened 2017-01-06
show why there might be an error without completely ignoring that CI …

grosser:
…might be taking its time

@mfischer-zd @jonmoter 

before:

screen shot 2017-01-06 at 9 53 33 am

after:
screen shot 2017-01-06 at 10 09 57 am

### DIFF
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -11,7 +11,16 @@ class CommitStatus
   end
 
   def status_list
-    combined_status.fetch(:statuses).map(&:to_h)
+    list = combined_status.fetch(:statuses).map(&:to_h)
+    if list.empty?
+      list << {
+        state: 'pending',
+        description:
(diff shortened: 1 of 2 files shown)


app/models/commit_status.rb
```diff
@@ -11,7 +11,11 @@ def status
   end
 
   def status_list
-    combined_status.fetch(:statuses).map(&:to_h)
+    list = combined_status.fetch(:statuses).map(&:to_h)
+    if list.empty?
+      list << {state: 'pending', description: 'Waiting for reply from CI, ignore if none is set up'}
```
mfischer-zd: Suggested verbiage:  "CI system has not reported a test result for this version."

grosser: I like the original more ... it's shorter and it tells a user that this might be a false positive

grosser:
screen shot 2017-01-17 at 10 01 40 am

Updated, good to go ?

PR #1636, opened 2017-01-06
use forking test runner to get rid of random test errors

grosser:
/cc @zendesk/samson @adam-hart 

 - more granular coverage reporting since now only 1 file runs at a time
 - isolated pollution / reproducible errors
 - can see where warnings are printed

runtime is almost identical (4:50 -> 5:00), so basically free features :)


grosser:
fixed a bunch of errors ... still some random errors left, but much better now ...

PR #1639, opened 2017-01-07
Add and use exact tags

adam-hart:
The command I used to pull tags in https://github.com/zendesk/samson/pull/1625 allows for a "fuzzy" tag return.  If no explicit tag is set, it takes the latest tag and appends the shortened commit SHA.  This caused an error when converting it to a gem version for comparison.  This adds a new command which should check for explicitly defined tags.

It also adds an early return if no commit is defined (which broke the form initializer).

### Risks
- Medium:  Could cause auto tagging to break more.

### DIFF
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -75,12 +75,16 @@ class Release < ActiveRecord::Base
   def next_release_number
     # If Github has a version tagged for this commit, use that version instead of ours
     latest_samson_version = Gem::Version.new(project.releases.last&.number || "0")
-    latest_github_version = Gem::Version.new(project.repository.tag_from_ref(commit))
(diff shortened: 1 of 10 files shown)


app/models/release.rb
```diff
@@ -75,12 +75,16 @@ def contains_commit?(other_commit)
   def next_release_number
     # If Github has a version tagged for this commit, use that version instead of ours
     latest_samson_version = Gem::Version.new(project.releases.last&.number || "0")
-    latest_github_version = Gem::Version.new(project.repository.tag_from_ref(commit))
+    next_samson_version = latest_samson_version.to_s.dup.sub!(/\d+$/) { |d| d.to_i + 1 }
+
+    return next_samson_version unless commit
```
grosser: missing test for this logic

PR #1640, opened 2017-01-07
do not send our ENV to airbrake on rake errors, it includes tokens/secrets

grosser:
/cc @zendesk/samson
@cericksen 

fix is for airbrake v4 https://github.com/airbrake/airbrake/blob/v4.3.8/lib/airbrake/rake_handler.rb#L30-L34

this seems to still exist in latest airbrake:
https://github.com/airbrake/airbrake/blob/05e66f4941d4cedcb690124703b63ec0b498c293/spec/unit/rake/tasks_spec.rb#L21-L28

so most likely we want to do this for all our apps :(

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -10,6 +10,9 @@ if defined?(Airbrake)
     # this will be blacklist_params in v5 ... does not support the full rails syntax
     config.params_filters = Rails.application.config.filter_parameters
 
+    # do not send our environment (secrets etc) to airbrake
+    config.rake_environment_filters.concat ENV.keys
+
     # report in development:
     # - uncomment
     # - add development in application.rb


jonmoter:
To be clear, this is the scenario when you're letting Airbrake know a deploy has happened via Capistrano.  This isn't the scenario of an application generating an error.

So this means that any ENV vars that are used during a deployment are getting sent to Airbrake, not the application runtime secrets, right?

grosser:
No, this is the scenario where an error happened.

jonmoter:
Okay, whenever a project runs `rake <task>` with Airbrake configured, and if that rake task generates an error, all the `ENV` vars get passed along to Airbrake.

Just clarifying that this isn't happening for every error that's sent to Airbrake when `puma` or `unicorn` are running.

grosser:
correct

PR #1642, opened 2017-01-09
rewrite CommandExecutor without threading to prevent errors without b…

grosser:
…acktraces

previously used thread to get around weird IO.open with block behavior,
finally found a good explanation and fix, so simpler now and exceptions raised inside will behave nicely

@adam-hart

### DIFF
--- a/lib/samson/command_executor.rb
+++ b/lib/samson/command_executor.rb
@@ -4,30 +4,29 @@ module Samson
   module CommandExecutor
     class << self
       # timeout could be done more reliably with timeout(1) from gnu coreutils ... but that would add another dependency
+      # popen vs timeout http://stackoverflow.com/questions/17237743/timeout-within-a-popen-works-but-popen-inside-a-timeout-doesnt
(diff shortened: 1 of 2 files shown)


lib/samson/command_executor.rb
```diff
@@ -4,30 +4,29 @@ module Samson
...
+            pio = IO.popen(env, command.map(&:to_s), unsetenv_others: true, err: err)
+            output = pio.read
+            pio.close
+            [$?.success?, output]
           rescue Errno::ENOENT
-            output = "No such file or directory - #{command.first}"
-            false
+            [false, "No such file or directory - #{command.first}"]
+          ensure
+            if pio && !pio.closed?
```
sandlerr: shouldn't this ensure block be after rescuing the timeout? what if the timeout gets raised while we are trying to kill the process?

grosser: good catch https://github.com/zendesk/samson/pull/1647

PR #1650, opened 2017-01-11
Update slack webhook plugin content with PR info

mariovisic:
Add pull request information to the slack notifications for pending and
running deploys. Notifications for other states like
error/failed/succeeded won't output this information as there's no need
to print it twice.

It looks like this:

example notification

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/plugins/slack_webhooks/test/models/slack_webhook_notification_renderer_test.rb
+++ b/plugins/slack_webhooks/test/models/slack_webhook_notification_renderer_test.rb
@@ -5,11 +5,17 @@ SingleCov.covered!
 
 describe SlackWebhookNotificationRenderer do
   let(:subject) { 'Deploy starting' }
+  let(:pull_requests) do
+    [
+      stub("PR one", number: 42, title: 'Fix bug', url: 'http://pr1.url/', users: [stub(login: 'author1')]),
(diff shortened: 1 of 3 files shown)


plugins/slack_webhooks/test/models/slack_webhook_notification_renderer_test.rb
```diff
@@ -96,4 +107,36 @@
...
+        changeset: changeset,
+        pending?: false,
+        running?: false,
+        errored?: false,
+        failed?: false,
+        succeeded?: true,
+        url: "http://sams.on/url")
+
+      result = SlackWebhookNotificationRenderer.render(deploy, subject)
+      result.wont_match /Pull Requests/
```
grosser: FYI prefer `wont_include` as it will have a more readable error when it fails

mariovisic: Ah easy, i just copied the matcher further up the file. Want me to put a PR with a change, or did you want to quickly push a commit with it?

grosser:
does not matter, just FYI ;)

mariovisic:
👍

PR #1652, opened 2017-01-11
cannot sort by a string or weird things happen

grosser:
fallout from https://github.com/zendesk/samson/pull/1482

/cc @yatishb

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -109,7 +109,7 @@ class Project < ActiveRecord::Base
   end
 
   def release_prior_to(release)
-    releases.where('number < ?', release.number).order(:number).last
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -109,7 +109,7 @@ def gitlab?
   end
 
   def release_prior_to(release)
-    releases.where('number < ?', release.number).order(:number).last
+    releases.where('id < ?', release.id).order(:id).last
```
zehua: Assuming we still support only one release branch, this should work.

grosser: yeah, we use this logic elsewhere / you cannot create a `lower` release

PR #1658, opened 2017-01-12
notify airbrake on restart errors and make messages easier to grep in…

grosser:
… logs

aslo making sure that we keep samson running in an error event, so deploys do not get killed

/cc @jonmoter

### DIFF
--- a/app/models/restart_signal_handler.rb
+++ b/app/models/restart_signal_handler.rb
@@ -31,8 +31,11 @@ class RestartSignalHandler
     JobExecution.clear_queue
 
     output "Passing #{PASSED_SIGNAL} on."
-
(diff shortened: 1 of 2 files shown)


app/models/restart_signal_handler.rb
```diff
@@ -31,8 +31,11 @@ def run
     JobExecution.clear_queue
 
     output "Passing #{PASSED_SIGNAL} on."
-
     Process.kill(PASSED_SIGNAL, Process.pid) # shut down underlying server
+  rescue
+    output "Failed #{$!.message} ... restart manually when all deploys have finished"
```
sandlerr: does "restart manually" here mean SIGKILL or something else?

grosser:
basically `sv samson restart`

PR #1664, opened 2017-01-12
Update before_deploy hook location

mariovisic:
When the feature to enqueue jobs as 'pending' was added, it meant that
the before_deploy hooks now fire before deploy are enqueued, not before
they're deployed.

To fix this we can create a hook for job execution for `on_start`
similar to how `on_complete` works currently :)

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -23,7 +23,8 @@ class JobExecution
     @executor = TerminalExecutor.new(@output, verbose: true, deploy: job.deploy)
     @viewers = JobViewers.new(@output)
 
-    @subscribers = []
+    @start_subscribers = []
+    @complete_subscribers = []
     @env = env
     @job = job
     @reference = reference
@@ -72,8 +73,12 @@ class JobExecution
     finish
   end
 
+  def on_start(&block)
+    @start_subscribers << JobExecutionSubscriber.new(job, &block)
+  end
+
   def on_complete(&block)
-    @subscribers << JobExecutionSubscriber.new(job, &block)
+    @complete_subscribers << JobExecutionSubscriber.new(job, &block)
   end
 
   def descriptor
@@ -94,6 +99,8 @@ class JobExecution
   end
 
   def run!
+    @start_subscribers.each(&:call)
+
     @output.write('', :started)
 
     @job.run!
@@ -127,7 +134,7 @@ class JobExecution
   def finish
     return if @finished
     @finished = true
-    @subscribers.each(&:call)
+    @complete_subscribers.each(&:call)
   end
 
   def execute!(dir)
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -99,7 +99,13 @@ class JobExecution
   end
 
   def run!
-    @start_callbacks.each(&:call)
+    begin
+      @start_callbacks.each(&:call)
+    rescue => e
+      @output.write(e.message)
+      finish
+      return
+    end
 
     @output.write('', :started)
 
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -23,8 +23,8 @@ class JobExecution
     @executor = TerminalExecutor.new(@output, verbose: true, deploy: job.deploy)
     @viewers = JobViewers.new(@output)
 
-    @start_subscribers = []
-    @complete_subscribers = []
+    @start_callbacks = []
+    @complete_callbacks = []
     @env = env
     @job = job
     @reference = reference
@@ -74,11 +74,11 @@ class JobExecution
   end
 
   def on_start(&block)
-    @start_subscribers << JobExecutionSubscriber.new(job, &block)
+    @start_callbacks << block
   end
 
   def on_complete(&block)
-    @complete_subscribers << JobExecutionSubscriber.new(job, &block)
+    @complete_callbacks << JobExecutionSubscriber.new(job, &block)
   end
 
   def descriptor
@@ -99,7 +99,7 @@ class JobExecution
   end
 
   def run!
-    @start_subscribers.each(&:call)
+    @start_callbacks.each(&:call)
 
     @output.write('', :started)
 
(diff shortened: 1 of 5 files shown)


mariovisic:
@grosser no worries, just trying to figure out a nice way to do this. I think I might just be able to redirect the out/err of those blocks to the `@output` buffer, that way the symmetry of the start/complete callback style blocks will persist :)

mariovisic:
oh wait.... yeah this is actually easy because exceptions will get caught already and passed to the `error!` method anyways.

Updated in: 763ec513e899c478113fdd354591252efd02412b

mariovisic:
ok, change is in, tests are green, @grosser want to once over it again for me please?

test/lib/buddy_check_test.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 4
+SingleCov.covered! uncovered: 5
```
grosser: missing a test ?

grosser:
would be nice to have a test for `fails when on start callback fails`, to make sure nobody changes this around to Subscribers .. otherwise looks good :)

the stubbing going on is only getting worse ... threaded execution is not fun :(

test/lib/buddy_check_test.rb
mariovisic: It's due to this line:

> JobQueue.any_instance.stubs(:delete_and_enqueue_next) # we do not properly add the job, so removal fails

As `delete_and_enqueue_next` no longer runs it won't call the `stop_expired_deploys` method.

mariovisic:
@grosser I just realised that the `error!` hook will run for these on start callbacks, which will actually notify airbrake. I'm guessing we dont' want to do that? In that case will need to add a seperate rescue and early return.

mariovisic:
ok great, the handling of failed before hooks + test are added now in 51e0b68edd2b9e09f43015d5912d069f9258e4d7

app/models/job_execution.rb
```diff
@@ -94,6 +99,14 @@ def error!(exception)
...
   def run!
+    begin
+      @start_callbacks.each(&:call)
+    rescue => e
+      @output.write(e.message)
+      finish
+      return
+    end
+
     @output.write('', :started)
```
grosser: should this come after the `:started` write since that starts the streaming ... might not be sent to the user if it comes before ...

app/models/job_execution.rb
```diff
@@ -94,6 +99,14 @@ def error!(exception)
   end
 
   def run!
+    begin
+      @start_callbacks.each(&:call)
+    rescue => e
+      @output.write(e.message)
+      finish
+      return
```
grosser: finish is already rescues below ... why do we need a separate one ?

mariovisic: We have to return in the rescue, otherwise the body of the job will run, so we can't really re-use that, we could set a flag instead and then skip things if the flag is already set, but that might get a bit hanky...

mariovisic: actually.... it could simply be something like:

```
  rescue => e
    error!(e) if !@stopped && @error_in_start
```

mariovisic: in the original rescue.

grosser:
why not just let it raise ?

mariovisic:
@grosser I don't think i'm doing a very good job of explaining the changes to the `run!` method. Let me try again.

Essentially it's currently something like:

```ruby
def run!
# job body here
rescue
  # report errors from job body to airbrake
ensure
  # after callbacks
```


If we change the method to now perform the before callbacks, so:

```ruby
def run!
# before callbacks here
# job body here
rescue
  # report errors from job body to airbrake
ensure
  # after callbacks
```

The issue is now that airbrake errors are reported for before callbacks, which we don't want, but the ensure is still run which we do want.

So instead i've changed the code to be:

```ruby
def run!
begin
  # before callbacks
ensure
  # print error and run after callbacks

# job body

....
```

Does that make more sense? If we re-use the existing rescue and ensures then I can't see an easy way to not run the code that reports to airbrake. So we need a seperate set up top with an early return to ensure the job body + everything else does not run.

mariovisic:
> why not just let it raise ?


well it'll fall to the rescue below which will notify airbrake, which we do not want.

grosser:
we can raise `Samson::Hooks::UserError` (for now don't change anything ...
we'll catch these errors as we go) ... that will not spam airbrake ... and
then the behavior is roughly as before with the added benefit of actually
logging these errors to the output :)

mariovisic:
That sounds tidy. I'll update the PR on Monday. Have a nice weekend!

grosser:
thx for your work here, this is one of these ugly things that are tricky to
get right :D

mariovisic:
Good morning @grosser 

I've addresses those last changes in 557831cc6a61d3c6e872a8e6b14a62fc63ca89cb please take a look when you have a spare moment :)

grosser:
Don't be silly, it's late afternoon ;)
👍 looks good!

test/models/job_execution_test.rb
```diff
@@ -185,12 +189,25 @@ def last_line_of_output
...
-    execute_job { called_subscriber = true }
+    execute_job('master', on_start: -> { called_subscriber = true })
     assert_equal true, called_subscriber
   end
 
+  it 'fails when on start callback fails' do
+    execute_job('master', on_start: -> { raise(Samson::Hooks::UserError, 'failure') })
+
+    job.output.wont_include 'monkey'
+    assert_equal 'errored', job.status
```
grosser: assert that the output includes `failure` ?

mariovisic: Ta, fixed in f8c047dc

mariovisic:
@grosser did you want me to squash this all down to a single commit with a useful title/description? Or are you going to do that when you merge ?

grosser:
You can just merge with the squash setting 😉

mariovisic:
@grosser oh, I don't work here, would you like to do the honours?

grosser:
ah, it says 'contributor' ... I though it was 'collaborator' :D

mariovisic:
@grosser thanks for all your help getting this in :)

grosser:
Thx for fixing all this up :D

PR #1666, opened 2017-01-15
sort clusters by natural order name

grosser:
before:
screen shot 2017-01-14 at 5 42 47 pm

after:
screen shot 2017-01-14 at 5 42 55 pm


grosser:
so the trick is to not mention you ;)

PR #1669, opened 2017-01-16
replace different periodical execution with background job interface

grosser:
@zendesk/samson

Unify periodical tasks ... get rid of crons ...
So far we had a bad experience with crons / have to manage them manually / reporting sucks / uncovered airbrake security holes ...
This should make things more consistent and easier on simple platforms like heroku.


grosser:
This removes all our current crons, the only thing left would be 'general' things like tmp dir cleanup, but these have to run on every system anyway.

grosser:
no negative feedback from sandler/moter either, confirmed working in staging, so here goes nothing :)

PR #1677, opened 2017-01-18
track when projects finish so we can show build duration

grosser:
screen shot 2017-01-18 at 10 00 05 am

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -47,6 +47,7 @@ class DockerBuilderService
   def run!(push: false, tag_as_latest: false)
     build.docker_build_job.try(:destroy) # if there's an old build job, delete it
     build.docker_tag = build.label.try(:parameterize).presence || 'latest'
+    build.started_at = Time.now
 
     job = build.create_docker_job
     build.save!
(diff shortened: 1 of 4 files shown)


jonmoter:
Do we distinguish between when a deploy was created, and when it started?  For example, if you create a deploy at 10am, but you don't get a buddy approval 'til 10:15am, do we save 10:15am in some column?

grosser:
the build will be created when he deploy starts, so it will not depend on
buddy approval

atm we don't do reporting on it, so I'm not indexing it ... but yeah might
be interesting later on to build some graphs with this ...

jonmoter:
Ah yes, this is just for `Build` objects.

Should we also add a `finished_at` column for `Deploy` and/or `Job`?

grosser:
for deploy we use started_at and updated_at and already display duration
 ... for builds we need a finished at because it can be updated

jonmoter:
Okay.  Using `updated_at` to mean finished at feels hacky to me.  Like if we ever run a script that updates a bunch of old Deploys, we'll lose the information about how long they took to run.

grosser:
Yeah true ... but I'll make that it's own PR

grosser:
confirmed working on staging

PR #1680, opened 2017-01-19
Correct exact tags

adam-hart:
https://github.com/zendesk/samson/pull/1639 introduced the `exact_tag_from_ref` method, which was relying on `git tag --points-at`, which wasn't introduced into git until v2.7.  Doesn't look like we're running that in production, so a pivot is probably easiest.

This regexes the version number through our version regex to filter out made up version numbers like `v1.0-1-gbcf85da`.

### Tasks
 - [ ] :+1: from team

### Risks
- Low:  `exact_tag_from_ref` is already broken.

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -47,14 +47,15 @@ class GitRepository
 
   # @return [nil, tag-sha or tag]
   def fuzzy_tag_from_ref(git_reference)
-    return unless ensure_local_cache!
+    return unless update_local_cache!
     capture_stdout 'git', 'describe', '--tags', git_reference
   end
 
   # @return [nil, tag]
   def exact_tag_from_ref(git_reference)
-    return unless ensure_local_cache!
-    capture_stdout 'git', 'tag', '--points-at', git_reference
+    fuzzy_tag = fuzzy_tag_from_ref(git_reference)
+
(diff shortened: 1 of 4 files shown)


app/models/git_repository.rb
```diff
@@ -47,14 +47,15 @@ def commit_from_ref(git_reference)
 
   # @return [nil, tag-sha or tag]
   def fuzzy_tag_from_ref(git_reference)
-    return unless ensure_local_cache!
+    return unless update_local_cache!
     capture_stdout 'git', 'describe', '--tags', git_reference
   end
 
   # @return [nil, tag]
```
grosser: some docs on `fuzzy tags returns v76 or v76-20-gedd45c2 and we only want it in the first case, prefer to use git tag --points-at but it is only available in git 2.x+`

app/models/git_repository.rb
```diff
@@ -47,14 +47,15 @@ def commit_from_ref(git_reference)
...
   def exact_tag_from_ref(git_reference)
-    return unless ensure_local_cache!
-    capture_stdout 'git', 'tag', '--points-at', git_reference
+    fuzzy_tag = fuzzy_tag_from_ref(git_reference)
+
+    fuzzy_tag && fuzzy_tag[Release::VERSION_REGEX, 0]
```
grosser: should return the full tag and not only the version part ...

PR #1681, opened 2017-01-19
add finished_at to deploys so we do not lose duration when deploy is …

grosser:
…updated

fix a few bugs around showing a unstarted deploy as started and deploys without buddy check never getting started_at

TODO:
 - backfill started_at where not started_at and finished

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -5,9 +5,6 @@ class Job < ActiveRecord::Base
 
   has_one :deploy
 
-  # Used by status_panel
-  alias_attribute :start_time, :created_at
-
   after_update { deploy&.touch }
 
   validate :validate_globally_unlocked
@@ -79,6 +76,7 @@ class Job < ActiveRecord::Base
 
   def run!
     status!("running")
+    update_attribute(:started_at, Time.now)
   end
 
   def success!
@@ -103,6 +101,10 @@ class Job < ActiveRecord::Base
 
   def finished?
     !ACTIVE_STATUSES.include?(status)
+  end
+
+  def finished_at
+    updated_at if finished?
   end
 
   def queued?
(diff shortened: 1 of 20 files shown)


grosser:
@jonmoter modified to reply on job attributes instead ... otherwise would have to copy data around and backfill a ton of deploys

db/migrate/20170119044208_add_finished_at_to_deploys.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+class AddFinishedAtToDeploys < ActiveRecord::Migration[5.0]
```
jonmoter: Should rename the migration, 'cause now it's removing instead of adding.

db/migrate/20170119044208_add_finished_at_to_deploys.rb
```diff
@@ -0,0 +1,7 @@
...
+  def change
+    remove_column :deploys, :finished_at, :datetime
```
jonmoter: Delete this line.

app/models/job.rb
```diff
@@ -6,7 +6,8 @@ class Job < ActiveRecord::Base
   has_one :deploy
 
   # Used by status_panel
-  alias_attribute :start_time, :created_at
+  alias_attribute :started_at, :created_at
+  alias_attribute :finished_at, :updated_at
```
jonmoter: Sure you don't want to add a `finished_at` column to Job?  'Cause right now, if you called `job.finished_at` while the Job is still running, you'll get a non-nil value back.

Should be relatively easy to add a column and do a backfill.

grosser:
- goldiloader will take care of N+1
 - we already rely on job to hold the status ... and we display status more
often then start time
 - finished_at could be a method that only returns updated_at when state is
finished ?

grosser:
closing this mess out ... will open a new PR with the salvageable improvements and no db changes later

PR #1682, opened 2017-01-19
do not allw infitite ram/cpu limits

grosser:
` {\"message\":\"Minimum memory limit allowed is 4MB\"}"`


plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
```diff
@@ -6,6 +6,7 @@ class DeployGroupRole < ActiveRecord::Base
     belongs_to :deploy_group
     belongs_to :kubernetes_role, class_name: 'Kubernetes::Role'
     validates :ram, :cpu, :replicas, presence: true
+    validates :ram, :cpu, numericality: { greater_than_or_equal_to: 4 }
```
jonmoter: Bad idea to have a minimum of 4 CPU cores.

grosser:
fixed up

PR #1684, opened 2017-01-20
stream logs from live/dead pods to fight incomplete logs without havi…

grosser:
…ng to sleep

We had many deploys where asking for the logs even 2-10s after the pod died
did not result in the complete logs.

FYI using no container argument was often more reliable at getting logs.

With this approach we can increase the timeout without affecting good deploys
instead of always sleeping which would be ineffective.
We also get notified when the logs did not finish so we know to inspect manually.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -49,11 +49,12 @@ module Kubernetes
         @pod.spec.containers.map(&:to_h)
       end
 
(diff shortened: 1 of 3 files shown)


grosser:
confirmed working in staging https://samsontest.zende.sk/projects/kube_service_watcher/deploys/4228

grosser:
confirmed working in prod ... https://samson.zende.sk/projects/aws_rds_mysql_consul_tagger/deploys/241294 ... just needs some newlines ...

PR #1685, opened 2017-01-21
logs need newlines for readability

grosser:
getting weird logs atm ... but at least they are complete now ...


plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -92,11 +92,11 @@ def fetch_logs(container, previous:)
           begin
             timeout_logs do
               @client.watch_pod_log(name, namespace, container: container).each do |log|
-                result << log
+                result << log << "\n"
```
jonmoter: Maybe do a `result << log.rstrip << "\n"`?

grosser: for now I'll stay with the 'exact' version ... we can revisit if there is trouble ...

PR #1687, opened 2017-01-24
Change status message for successful deploy

jacobat:
The current wording:

Alice (with Bob) deployed v11105 to Production deploy 8 minutes ago, it took 5
minutes 36 seconds.

makes it hard to tell if the deploy was started or finished 8 minutes ago. This
commit changes the wording to be:

Alice (with Bob) started deploying v11105 to Production deploy 8 minutes ago,
it took 5 minutes 36 seconds.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Low, purely a string change.

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -28,7 +28,7 @@ class Deploy < ActiveRecord::Base
     "running"    => "is deploying",
     "cancelling" => "is cancelling a deploy",
     "cancelled"  => "cancelled a deploy",
-    "succeeded"  => "deployed",
+    "succeeded"  => "started deploying",
     "failed"     => "failed to deploy",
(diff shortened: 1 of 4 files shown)


grosser:
Yeah the current wording is confusing ... but the 'started deploying' makes it sounds like it is still underway.

How about "Alice (with Bob) deployed v11105 to Production. Started 8 minutes ago,
took 5 minutes 36 seconds."

jacobat:
That's fine as well. I tried to make a PR but the current way the strings are constructed makes it a bit more work than expected. @grosser would you be willing to make the change?

jacobat:
Going with #1735 instead.

PR #1688, opened 2017-01-24
cache into a directory that is automatically cleaned up

grosser:
@craig-day 
/fyi @bquorning


grosser:
FYI this reduced built time 1:39 vs 1:18 ... but that is still way faster then our slowest build and makes all other builds faster (less download) .... so overall still a win I think

PR #1690, opened 2017-01-25
Slack notification threads

ben:
This adds the ability for the Slack webhook notifications to collapse notifications into a thread so as not to clutter up the chat flow.

### Lo-fi mockup

A successful deploy would look like this in the main flow:
image

Full thread contents:
image

Failure and error messages would get dropped into the main channel as well:
image

/cc @zendesk/samson

### Tasks
- [x] New DB table
- [ ] Alter logic to use [thread timestamps and reply_broadcast](https://api.slack.com/docs/message-threading#posting_replies)
- [ ] Tests
- [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/plugins/slack_webhooks/db/migrate/20170125010044_add_slack_timestamp_for_threading.rb
+++ b/plugins/slack_webhooks/db/migrate/20170125010044_add_slack_timestamp_for_threading.rb
@@ -1,0 +1,11 @@
+class AddSlackTimestampForThreading < ActiveRecord::Migration[5.0]
+  def change
+    create_table :slack_webhook_threads do |t|
+      t.integer :deploy_id, null: false
+      t.string :slack_ts, null: false
+      t.timestamps
+    end
+
+    add_index :slack_webhook_threads, :deploy_id
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170118172040) do
+ActiveRecord::Schema.define(version: 20170125010044) do
 
   create_table "builds", force: :cascade do |t|
     t.integer  "project_id",                                       null: false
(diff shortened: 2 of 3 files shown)


plugins/slack_webhooks/db/migrate/20170125010044_add_slack_timestamp_for_threading.rb
```diff
@@ -0,0 +1,11 @@
+class AddSlackTimestampForThreading < ActiveRecord::Migration[5.0]
+  def change
+    create_table :slack_webhook_threads do |t|
+      t.integer :deploy_id, null: false
+      t.string :slack_ts, null: false
+      t.timestamps
+    end
+
+    add_index :slack_webhook_threads, :deploy_id
```
grosser: FYI you can do this inside the create_table now

grosser: should this be unique ?

plugins/slack_webhooks/db/migrate/20170125010044_add_slack_timestamp_for_threading.rb
```diff
@@ -0,0 +1,11 @@
+class AddSlackTimestampForThreading < ActiveRecord::Migration[5.0]
+  def change
+    create_table :slack_webhook_threads do |t|
+      t.integer :deploy_id, null: false
+      t.string :slack_ts, null: false
```
grosser: ts is what ... pick a longer/clearer name ?

grosser:
I like the idea :)

plugins/slack_webhooks/db/migrate/20170125010044_add_slack_timestamp_for_threading.rb
ben: This is a convention in the Slack API, it means "timestamp". I'll fix.

ben:
Well 💩. Looks like you can't do threads with webhooks, since the only response the outside service gets is `"ok"`; there's no timestamp to use for threading. This will have to be added to the `slack_app` plugin instead.

PR #1693, opened 2017-01-27
increase log timeout ... 10s was not enough

grosser:
still had a partial log ... better wait longer :/


jonmoter:
Wanna set this as an ENV var (with a default) rather than hard-coding it?

grosser:
done

PR #1696, opened 2017-01-30
move k8s labels that Samson adds to annotations

jonmoter:
A lot of the `id` fields that Samson adds as labels aren't things we ever use to filter. Move them to annotations instead.

Things like `project`, `role`, and `deploy_group` are left as labels, 'cause I'm more likely to select or filter based on those.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


grosser:
is there a downside to having lots of labels ?

PR #1698, opened 2017-01-31
Add 'Cancel scheduled builds' option

mariovisic:
Add option for stages to be able to cancel pending deploys for a user
when a new pending deploy is queued, this allows projects with frequent
deploys to skip deploys and save this, this is useful for non critial
applications, a staging environment is a good example.

#### Options with tooltip:

___

screen shot 2017-01-31 at 12 08 51 pm

___

#### Deployed cancelled as more are queued:

___
queued_jobs

___

/cc @zendesk/samson

### Tasks
 - [x] Write some tests
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -15,8 +15,9 @@ class DeployService
 
       if stage.cancel_queued_deploys?
         stage.deploys.pending.prior_to(deploy).for_user(user).each do |queued_deploy|
-          JobExecution.dequeue(queued_deploy.job.id)
-          queued_deploy.job.cancelled!
+          if JobExecution.dequeue(queued_deploy.job.id)
+            queued_deploy.job.cancelled!
+          end
         end
       end
 
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -13,6 +13,13 @@ class DeployService
     if deploy.persisted?
       send_sse_deploy_update('new', deploy)
 
+      if stage.cancel_queued_deploys?
+        stage.deploys.pending.prior_to(deploy).for_user(user).each do |queued_deploy|
+          JobExecution.dequeue(queued_deploy.job.id)
+          queued_deploy.job.cancelled!
+        end
+      end
+
       if !deploy.waiting_for_buddy? || copy_approval_from_last_deploy(deploy)
         confirm_deploy!(deploy)
       end
--- a/app/views/stages/_fields.html.erb
+++ b/app/views/stages/_fields.html.erb
@@ -20,7 +20,7 @@
 
   <%= form.input :run_in_parallel, as: :check_box, label: "Can run in parallel", help: "Deploys are not queued. Executed immediately" %>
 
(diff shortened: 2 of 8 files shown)


mariovisic:
I still need to write tests, but wouldn't mind some feedback now to make any changes that are required.

grosser:
maybe `Cancel duplicate queued deploys` ... to make it more obvious

mariovisic:
I don't think duplicate is the word we want to use as the deploys would usually be for different revisions, although I think you're right, the text could be more expressive.

mariovisic:
"Multiple" maybe? hmmmmm

mariovisic:
"Prevent multiple queued deploys"
"Cancel multiple queued deploys"
"Only allow a single queued deploy"

...

grosser:
`Allow 1 concurrent deploy from webhook.`

mariovisic:
@grosser that works but is misleading as it is per user, not neccessarily per webhook. The same behavior would occur if the same user manually created deploys.

mariovisic:
I don't actually think the naming is too bad now, the tooltip explains exactly what it does in good detail.

mariovisic:
@grosser anyways, this is ready to go, tests added, just need to settle on the name 👍

app/models/deploy_service.rb
```diff
@@ -13,6 +13,13 @@ def deploy!(stage, attributes)
     if deploy.persisted?
       send_sse_deploy_update('new', deploy)
 
+      if stage.cancel_queued_deploys?
+        stage.deploys.pending.prior_to(deploy).for_user(user).each do |queued_deploy|
+          JobExecution.dequeue(queued_deploy.job.id)
```
grosser: only do the cancel if it is successfully dequeued ... the job you are looking at might already be running now

`if JobExecution.dequeue ... `

grosser:
`Allow 1 queued deploy per user` then ?
... maybe add `this most useful when trying to not deploy every push for frequently updated projects.` to the help text to make the usecase a bit clearer.

mariovisic:
@grosser easy, updated in 993d3666

app/views/stages/_fields.html.erb
```diff
@@ -20,6 +20,8 @@
 
   <%= form.input :run_in_parallel, as: :check_box, label: "Can run in parallel", help: "Deploys are not queued. Executed immediately" %>
 
+  <%= form.input :cancel_queued_deploys, as: :check_box, label: "Allow 1 queued deploy per user", help: "When a new deploy is created for a user. Any queued deploys for that user are cancelled. This most useful when trying to not deploy every push for frequently updated projects." %>
```
grosser: ... hmmm shorter / better ? `Max 1 queued deploy per user`

mariovisic: Yes, I think that is an improvement :)

mariovisic: Updated in caddc787

PR #1699, opened 2017-01-31
added samson/before_docker_build

staugaard:
This adds a hook for running a script before building docker containers. All you have to do is have a `samson/before_docker_build` executable in your repository, and it will get run just before starting the docker builds.

@jonmoter and I agreed that this was a good addition.

/cc @jonmoter @zendesk/samson

### Tasks
 - [x] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -133,7 +133,7 @@ class DockerBuilderService
     build.save!
   end
 
-  def before_docker_build(tmp_dir)
+  def execute_before_docker_build_script(tmp_dir)
     before_docker_build_file = File.join(tmp_dir, BEFORE_DOCKER_BUILD)
     if File.file?(before_docker_build_file)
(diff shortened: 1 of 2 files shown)


app/models/docker_builder_service.rb
```diff
@@ -132,8 +133,22 @@ def run_build_image_job(local_job, push: false, tag_as_latest: false)
     build.save!
   end
 
-  def build_image(tmp_dir)
+  def before_docker_build(tmp_dir)
+    before_docker_build_file = File.join(tmp_dir, BEFORE_DOCKER_BUILD)
```
grosser: this block could be it's own 'execute_before_docker_build_script' method

grosser: ideally just be another hook ... but afaik there is no nice way of setting this up without creating a super tiny plugin :/

PR #1700, opened 2017-01-31
better perf debugging

grosser:
- `GITHUB get (0.185264894s) https://api.github.com/repos/zendesk/foo/pulls/12641`
 - automatic logging to stdout
 - disable_cache helper

/cc @zendesk/samson


sandlerr:
can octokit be hooked with newrelic / statsd easily?

grosser:
newrelic already reports it as http/external, this is mostly for making it visible in console like the sql queries are

PR #1701, opened 2017-01-31
SAMSON-320 make caching last longer so pages like releases#index do not get slow…

grosser:
… ... make connections use ENV var too

/cc @zendesk/samson

this might make some caching issues pop up ... we should either fix these or make them have a local 'expires_in'

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -67,9 +67,9 @@ module Samson
       options = {
         value_max_bytes: 3000000,
         compress: true,
-        expires_in: 1.day,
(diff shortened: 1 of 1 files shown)


grosser:
default is to never expire ... instantly expiring would make any cache useless :D

irwaters:
correct, but do we want to start with something < infinity > 1.day? :)

grosser:
hmm ok ... 7 days then to purge everything eventually ... ideally keep memcached as full as possible :D

PR #1703, opened 2017-01-31
make email label clearer

grosser:
screen shot 2017-01-31 at 1 52 45 pm


grosser:
yeah ... was trying to come up with an exact but not too long name too :(

PR #1704, opened 2017-01-31
execute the contents of samson/before_docker_build

staugaard:
Execute the contents of `samson/before_docker_build` instead of the file itself.
Why? Because this gives us secret resolution in this script.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -138,7 +138,9 @@ class DockerBuilderService
     if File.file?(before_docker_build_file)
       output.puts "Running #{BEFORE_DOCKER_BUILD} ..."
 
-      unless execution.executor.execute!(before_docker_build_file)
+      commands = File.readlines(before_docker_build_file)
+
+      unless execution.executor.execute!(*commands)
         raise Samson::Hooks::UserError, "Error running #{BEFORE_DOCKER_BUILD}"
       end
     end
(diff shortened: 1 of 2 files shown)


grosser:
yeah ... but that's kind of sketchy since nobody will know which secrets you use ... prefer setting env vars and then use them in the script for better transparancy

staugaard:
That is the thing. There is no way for me to set environment variables. This is for docker builds, which are not part of a stage, so I can not configure it.

staugaard:
Oh well this does not work anyway because there is no deploy involved, so the TerminalExecutor can not resolve the project. So it only has access to global secrets.

jonmoter:
Probably the best option is to have a UI in Samson allowing you to specify a command you run before the docker build happens.  So your pre-build command could be something like:

```
export S3_API_KEY="secret://aws_api_key"
./samson/before_docker_build
```

Then Samson would execute that (using the normal secret-replacement logic), and then execute the docker build.

The other week, @mfischer-zd asked me if there was a way in Samson to pass `--build-arg` params to docker build.  Currently there isn't.  Perhaps we could also fold that functionality into this.

staugaard:
That sounds like a big project @jonmoter. When can you have that ready 😉 ?

jonmoter:
Mick and I talked about this.  We're going to try a different approach like the one I suggested above.

PR #1705, opened 2017-02-01
clean up UI when adding project roles for a user

jonmoter:
Currently, the UI when adding a role to a user is a pain, since the projects aren't sorted, and it includes projects you already have a role for.

fullscreen_1_31_17__4_14_pm

This sorts and filters the projects in the dropdown.

/cc @zendesk/samson, @grosser 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -23,9 +23,10 @@ class Admin::UsersController < ApplicationController
     if role_id = params[:role_id].presence
       @projects = @projects.where("user_project_roles.role_id >= ?", role_id)
     end
-    @projects = @projects.select('projects.*, user_project_roles.role_id AS user_project_role_id').sort_by(&:natural_order)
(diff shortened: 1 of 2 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -23,7 +23,9 @@ def show
     if role_id = params[:role_id].presence
       @projects = @projects.where("user_project_roles.role_id >= ?", role_id)
     end
-    @projects = @projects.select('projects.*, user_project_roles.role_id AS user_project_role_id')
+    @projects = @projects.select('projects.*, user_project_roles.role_id AS user_project_role_id').sort_by(&:natural_order)
```
grosser: do we even need `@projects` now ?

grosser: ... can then get rid of the extra ordering

jonmoter: Yes, `@projects` is still used in the view.

PR #1706, opened 2017-02-01
use a clean cache instead of disabling cache so we get a more realist…

grosser:
…ic caching experience ... the same cache can be used multiple times on a single page

followup to https://github.com/zendesk/samson/pull/1700

I realized that this noop cache was giving me impossibly slow response times ... this should be more realistic

### DIFF
--- a/lib/samson/console_extensions.rb
+++ b/lib/samson/console_extensions.rb
@@ -20,14 +20,10 @@ module Samson
       end
     end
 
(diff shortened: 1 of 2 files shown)


sandlerr:
would it be better to pass this as a startup option, then it can be handled by environment.rb files?

grosser:
we don't have startup options for our console helpers ... and the server
should never be run with that

PR #1708, opened 2017-02-01
display previous stages in the pipeline as well as the next ones

jonmoter:
Added a `Stage#previous_stages` method, and updated the view so it shows the previous stages as well as the next stages.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/plugins/pipelines/app/decorators/stage_decorator.rb
+++ b/plugins/pipelines/app/decorators/stage_decorator.rb
@@ -21,7 +21,7 @@ Stage.class_eval do
 
   def previous_stages
     @previous_stages ||= project.stages.to_a.select do |stage|
-      stage.next_stage_ids.include? id
+      stage.next_stage_ids.map(&:to_i).include? id
     end
   end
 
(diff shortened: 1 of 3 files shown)


grosser:
next PR: generate a graph and render it ? :D

jonmoter:
@grosser could you try running this locally and confirming it looks okay?  When I try running Samson on my laptop, the ruby interpreter segfaults.

grosser:
pushed a fix:

screen shot 2017-02-01 at 1 18 37 pm

grosser:
try the grosser/24 branch ... does that help ?

plugins/pipelines/app/decorators/stage_decorator.rb
```diff
@@ -19,6 +19,12 @@ def next_stages
     Stage.where(id: next_stage_ids).to_a
   end
 
+  def previous_stages
+    @previous_stages ||= project.stages.to_a.select do |stage|
+      stage.next_stage_ids.map(&:to_i).include? id
```
jonmoter: @grosser is there some way to force `next_stage_ids` to be an array of ints?  It's pretty unintuitive that the method returns an array of strings.

grosser:
yeah, this is messy ... ran into this a few times already :D
let's finish this PR and then do a cleanup PR

PR #1711, opened 2017-02-01
pass DOCKER_HOST env var to commands

jonmoter:
`DOCKER_HOST` is the env var that the docker CLI uses

```
$ host_value=tcp://redacted.docker-server.example.com:2375

$ DOCKER_URL=$host_value docker ps | head -1
Cannot connect to the Docker daemon. Is the docker daemon running on this host?

$ DOCKER_HOST=$host_value docker ps | head -1
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
```

Pass this environment variable to the scripts, if they want to connect to the docker daemon.

/cc @zendesk/samson, @staugaard 

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -72,7 +72,7 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60
 ## Docker
 # DOCKER_FEATURE=1 # optional, docker support
 # DOCKER_REGISTRIES=https://user:pass@my.registry/some-namespace # required, where to push/pull your docker images
-# DOCKER_URL= # optional, use this url instead of connecting to unix socket
+# DOCKER_HOST= # optional, use this url instead of connecting to unix socket
 # DOCKER_KEEP_BUILT_IMGS # optional. Set to 1 to keep built images for cache. Fills the disk so some cleanup machanism is needed
 
 # FLOWDOCK_API_TOKEN= # optional. only required for the flowdock integration user mention autocomplete in the buddy approval request form (when BUDDY_CHECK_FEATURE=1). Buddy approval notification would still work without this
--- a/.env.example
+++ b/.env.example
@@ -72,7 +72,7 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60
 ## Docker
 # DOCKER_FEATURE=1 # optional, docker support
 # DOCKER_REGISTRIES=https://user:pass@my.registry/some-namespace # required, where to push/pull your docker images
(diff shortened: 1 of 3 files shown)


app/models/terminal_executor.rb
```diff
@@ -106,10 +106,13 @@ def whitelisted_env
...
       'PATH', 'HOME', 'TMPDIR', 'CACHE_DIR', 'TERM', 'SHELL', # general
       'RBENV_ROOT', 'RBENV_HOOK_PATH', 'RBENV_DIR', # ruby
-      'DOCKER_URL', 'DOCKER_REGISTRY' # docker
+      'DOCKER_HOST', 'DOCKER_URL', 'DOCKER_REGISTRY' # docker
     ] + ENV['ENV_WHITELIST'].to_s.split(/, ?/)
     env = ENV.to_h.slice(*whitelist)
     env['DOCKER_REGISTRY'] ||= DockerRegistry.first&.host # backwards compatibility
+
+    # DOCKER_HOST is the env var that the docker CLI uses to connect to a remote host
+    env['DOCKER_HOST'] ||= env['DOCKER_URL'] if env['DOCKER_URL']
```
grosser: not a big fan of this ... no need for custom logic like that ... can we use _HOST or _URL consistently ?

jonmoter: We could also change over the ENV var that Samson uses to `DOCKER_HOST`, but then we'd need to add backward-compatibility logic so it doesn't break with users who had `DOCKER_URL`. Have a preference?

grosser:
prefer that ... easy to add a bit more logic and deprecation ...

jonmoter:
Updated to use `DOCKER_HOST` instead of `DOCKER_URL`, but I'm whitelisting both for now.

.env.example
```diff
@@ -72,7 +72,7 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60
 ## Docker
 # DOCKER_FEATURE=1 # optional, docker support
 # DOCKER_REGISTRIES=https://user:pass@my.registry/some-namespace # required, where to push/pull your docker images
-# DOCKER_URL= # optional, use this url instead of connecting to unix socket
+# DOCKER_HOST= # optional, use this url instead of connecting to unix socket
```
grosser: `use this host` ?

PR #1712, opened 2017-02-02
make rendering index faster when user has no stars

grosser:
/cc @jonmoter @sandlerr 

atm it's crazy slow since we render so much stuff for all our projects ... no need for that ...

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -92,7 +92,7 @@ class ProjectsController < ApplicationController
     if ids = current_user.starred_project_ids.presence
       Project.where(id: ids)
     else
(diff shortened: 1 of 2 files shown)


grosser:
last will keep things interesting ... first would just mean same old stale repos :D 
... maybe simplest to do "Star projects to see them here" ...

sandlerr:
or auto-star a randomised subset of projects for new users + backfill :D

grosser:
nah ... fake data is always trouble :D

grosser:
changed from 'Dashboard' to
screen shot 2017-02-01 at 5 21 07 pm

... this should things easier to understand ...

PR #1714, opened 2017-02-02
Add note about execute permissions for build.sh

jsdnxx:
Just a small docs update to clarify instructions for the docker binary builder plugin

/cc @zendesk/samson

### Risks
- none


grosser:
thx!

PR #1715, opened 2017-02-02
show haw many mysql connections are waiting ... should always be 0 or…

grosser:
… we are in big trouble

@jonmoter @sandlerr

### DIFF
--- a/lib/samson/logging.rb
+++ b/lib/samson/logging.rb
@@ -18,7 +18,8 @@ elsif Samson::EnvCheck.set?("RAILS_LOG_TO_SYSLOG")
       params: params,
       user_id: request.env['warden']&.user&.id,
       ip: request.remote_ip,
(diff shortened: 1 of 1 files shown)


sandlerr:
do we want to do it for connections that belong to the current thread?

grosser:
This is more of an overall health indicator ... There should be only 1
connection per thread and when the request is finished it will not be
waiting

sandlerr:
but won't this give false positives? if there is a connection waiting from another thread it'll still show up? anyway :+1: let's give it a shot

grosser:
yes, it's not an indicator for why this thread was slow ... more like an overall system health thing like 'thread_count' ... I guess we should find a better place for this, but it's kind of nice to have that in the logs for later analysis ... maybe should be part of the progname or so ...

PR #1716, opened 2017-02-02
init fixes

grosser:
screen shot 2017-02-01 at 5 21 07 pm

/cc @jonmoter @sandlerr

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -13,6 +13,9 @@ if defined?(Airbrake)
     # do not send our environment (secrets etc) to airbrake
     config.rake_environment_filters.concat ENV.keys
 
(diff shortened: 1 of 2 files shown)


config/initializers/airbrake.rb
```diff
@@ -13,6 +13,9 @@
     # do not send our environment (secrets etc) to airbrake
     config.rake_environment_filters.concat ENV.keys
 
+    # send correct errors even when something blows up during initialization
+    config.environment_name = Rails.env
```
sandlerr: is this papering over some other problem?

grosser:
That we had errors during initialization ... Which the other commit takes
care of ... But there might be more in the future ... And knowing what
environment they come from is good 🤗

PR #1717, opened 2017-02-02
fix bugs introduced in PR #1696

jonmoter:
In #1696, I accidentally introduced a bug that broke Kubernetes deployments, 'cause `DeployExecutor` was looking for labels that no longer existed in the pods. Whoops.

This fixes the problem, and adds a test to ensure it doesn't happen again.

/cc @grosser, @craig-day 

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -46,8 +46,7 @@ module Kubernetes
 
     def pod_selector(deploy_group)
       {
-        release_id: id,
-        deploy_group_id: deploy_group.id,
+        release_id: id
       }
     end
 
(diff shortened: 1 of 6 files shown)


plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -46,8 +46,7 @@ def clients
 
     def pod_selector(deploy_group)
       {
-        release_id: id,
-        deploy_group_id: deploy_group.id,
```
grosser: that looks suspicious ...

grosser: should either not pass in deploy_group and make sure all callers work with this ... or still pass it in ...

jonmoter:
This is turning into a mess.  Closing in favor of #1720.

PR #1732, opened 2017-02-06
send more info to compliance

grosser:
@laurenmbade


grosser:
this is used to trigger tickets for compliance to talk to the owners of new projects, so sending to splunk would mean they have to alert on that or something like that ... would be a more complicated workflow

PR #1736, opened 2017-02-07
remove weird/dangerous job UI

grosser:
screen shot 2017-02-06 at 3 57 54 pm

@zendesk/samson  users either do a deploy or macro or build ... don't need yet another UI to execute random commands ?


jonmoter:
Have you looked through our projects to see if any teams are using Jobs right now?

henders:
Yea, the only use I have is checking the output of previous macro executions.

grosser:
Craig mentioned that too ... will have to somehow make macros show their
executions ... which would be good anyway ... or kill macros :D

jonmoter:
In general, I'm wary of removing functionality that other people may be using.  I'd be upset if I updated to a newer version of an app, and some functionality that I relied upon was just gone.

henders:
I do like the macros as they're the best way to run restarts without redeploying, so I'd shed a tear if they got removed.

grosser:
idk of any usecase for jobs ... we can find out and then revert ... maybe
ask more zendeskians ...

we can convert macros into stages with 'no code deployed' ... that would be
fast then too ...

maybe we start with just removing the execute mechanic and see if anyone
complains ?

grosser:
the dangerous bit is gone in https://github.com/zendesk/samson/pull/1761 ... so only thing left is linking macros to their jobs ... new PR will be simpler

PR #1737, opened 2017-02-07
looks like we can use prepend now

grosser:
fixed in https://github.com/makandra/query_diet/pull/21

seeing lots of segfaults ... so upgraded to ruby 2.3.3


grosser:
nvm ... still happens ... but only in some tests ...

grosser:
maybe coming here https://github.com/makandra/query_diet/pull/21 ... but who knows

PR #1738, opened 2017-02-07
synchroize our test transactions so all threads can share one connect…

grosser:
…ion and we can test more realistically

this is a bit funky, but allows us to use transactional fixtures in multiple threads ... so we can test job execution and other crazyness realistically ...

afik only alternative is to not use transactional fixtures and database cleaner which adds time/complexity overhead


henders:
Yea, a bit funky to use alright. Not sure if someone would know that it was available for use when writing new tests. Also not very intuitive to use with `ArMultiThreadedTransactionalTests.activate`. Would it make sense to always have it activated since we don't run tests concurrently?

grosser:
always active would work too, was just trying to apply it surgically
afaik we only need it for things that test jobs (threaded execution)
so explicitly having it there calls attention to it

grosser:
@henders want to give it a try ?

henders:
Yup. why not. Wondering if it's worth calling out in a section in the README to help future testing?

👍

grosser:
it's only used in 2 specs ... so not really something we should call out in general docs ... also there is no readme for testing/development 😞

PR #1742, opened 2017-02-08
[SAMSON-337] fix bug where edit stage page crashes on pipeline plugin

jonmoter:
Was crashing when editing a stage.  Apparently we have a bunch of stages with an order of nil.  This patches the crash, but we should probably fix the records missing.

/cc @zendesk/samson, @staugaard 

### Tasks
 - [ ] :+1: from team

### References
 - JIRA: https://zendesk.atlassian.net/browse/SAMSON-337
 - Airbrake: https://zendesk.airbrake.io/projects/95346/groups/1881455414135434331

### Risks
- Level: Low


grosser:
closing in favor of https://github.com/zendesk/samson/pull/1743

PR #1746, opened 2017-02-09
json api should return json for errors

grosser:
@apanzerj @henders

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -6,6 +6,25 @@ class Api::BaseController < ApplicationController
   prepend_before_action :store_requested_oauth_scope
   before_action :require_project
 
+  # default error has very little information
+  # http://stackoverflow.com/questions/33704640/how-to-render-correct-json-format-with-raised-error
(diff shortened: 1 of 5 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -6,6 +6,25 @@ class Api::BaseController < ApplicationController
...
+    render json: {error: exception.record.errors}, status: 422
+  end
+
+  # default error has very little information
+  # https://github.com/rails/strong_parameters/issues/157
+  rescue_from ActionController::ParameterMissing do |exception|
+    render json: {error: {exception.param => ["is required"]}}, status: :bad_request
+  end
+
+  # otherwise renders a 500 and goes to airbrake
```
henders: Comment doesn't reflect the `:bad_request` return type. Did you want to return `internal_server_error`?

henders:
👍  after changing return type to match comment.

app/controllers/api/base_controller.rb
grosser: 'otherwise' -> if we do not use this hack

PR #1747, opened 2017-02-09
Delete jenkins cache configuraton after posting

upcomingnewton:
1. we do not delete cache after we have posted a job configuration. So if someone changes the job configuration, and we, after that, a new stage is added or new parameters, then we end up overwriting those changes to jenkins jobs.
2. Fetch the job configuration and add build params and description to it to ensure we are always updating the latest configuration.

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/VULCAN-296

### Risks
- Level: Low/Med/High


plugins/jenkins/app/models/samson/jenkins.rb
```diff
@@ -62,21 +66,31 @@ def jenkins_job_config
...
+        changes["build_params"] = missing_build_params
+      end
+      unless description_exists?(conf)
+        changes['job_description'] = true
+      end
+      changes
+    end
+
+    def build_job_config(changes)
+      Rails.cache.delete(job_name + JENKINS_JOB_CACHE_SUFFIX)
```
grosser: these 2 could usages could get out of sync ... prefer a cache_key method ... don't need that constant then either

plugins/jenkins/app/models/samson/jenkins.rb
```diff
@@ -183,8 +198,9 @@ def build
         emails: notify_emails
       }
       if deploy.stage.jenkins_build_params
-        post_config, new_config = check_and_build_job_config
-        if post_config
+        changes = check_job_config
+        if changes.keys.any?
```
grosser: don't need to do `.keys`

grosser:
looks good, getting rid of `post_config` variable is nice

plugins/jenkins/app/models/samson/jenkins.rb
upcomingnewton: corrected.

plugins/jenkins/app/models/samson/jenkins.rb
upcomingnewton: corrected.

PR #1754, opened 2017-02-11
only run binary builder if build.sh exists too

grosser:


prudhvi:
May be better to fix the setting check?

grosser:
more complicated ... and I want to delete it anyway ... so not adding a bunch of extra work :D

prudhvi:
👍

PR #1759, opened 2017-02-14
PAAS-743 do not allow deploy groups with different environments to sh…

grosser:
…are a vault server

This will avoid accidental misconfiguration that might leak production secrets to staging
since a changed secret will be written to all vault servers in that environment

screen shot 2017-02-14 at 1 50 51 pm

### DIFF
--- a/app/models/deploy_group.rb
+++ b/app/models/deploy_group.rb
@@ -6,7 +6,7 @@ class DeployGroup < ActiveRecord::Base
   include Permalinkable
 
   belongs_to :environment
-  belongs_to :vault_server
+  belongs_to :vault_server, class_name: 'Samson::Secrets::VaultServer'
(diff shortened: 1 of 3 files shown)


app/models/deploy_group.rb
```diff
@@ -50,4 +51,9 @@ def initialize_env_value
   def destroy_deploy_groups_stages
     DeployGroupsStage.where(deploy_group_id: id).delete_all
   end
+
+  def validate_vault_server_has_same_environment
+    return if !vault_server || [[], [environment_id]].include?(vault_server.deploy_groups.pluck(:environment_id).uniq)
+    errors.add :vault_server_id, "vault server #{vault_server.name} is already in use an a different environment"
```
grosser: no, we got 2 different ones

PR #1761, opened 2017-02-16
remove direct job execution

grosser:
@jonmoter @henders 

good to remove that ... if anyone complains we can roll back

split out from https://github.com/zendesk/samson/pull/1736


jbarreneche:
Hi! We were using the option to run custom jobs because macros don't support parameters and some times we need to run some task changing some configuration. 
Any chance this feature could be restored or allow custom parameters/environment for macros?

grosser:
Removed it to prevent this 1-off command running since it let's anyone do anything to all servers without much of a audit-trail / verification.

Custom environment sounds useable / simple ... basically have multiple `Name: [   ] Value: [    ]` lines per execution ... macros don't really have a 'configure' page and are pretty wonky anyway, so I'd prefer to build that for stages instead which already has a "Confirm" page where we can easily add a few new fields.

For starters this can be free-form ... later we could add some configuration so that it prompts the user to fill in these fields ... could even offer a select of acceptable values.

grosser:
how about https://github.com/zendesk/samson/pull/1803 ?

jbarreneche:
@grosser yes, allowing to set environment variables is enough for our use case. And I agree it's better for auditing purposes :)

grosser:
Got any idea how to pre-fill this with the required env var names ?
... I think I'll have to persist them to db anyway since otherwise the
"re-deploy" button won't work.

PR #1770, opened 2017-02-18
use real repo to run proper setup and not crash

grosser:
now that we have proper mutithreaded testing we can use less stubs and more real callbacks

previously all jobs crashed in `setup` and the tests just ignored that ... also found a bug in the thread monitoring code :)

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -263,7 +263,8 @@ class JobExecution
     Dir.mktmpdir("samson-#{@job.project.permalink}-#{@job.id}") do |dir|
       result = yield dir
     end
-  rescue Errno::ENOTEMPTY
+  rescue Errno::ENOTEMPTY, Errno::ENOENT
+    Airbrake.notify("Notify: make_tempdir error #{$!}")
     result # tempdir ensure sometimes fails ... not sure why ... return normally
   end
 
--- a/test/models/job_execution_test.rb
+++ b/test/models/job_execution_test.rb
@@ -30,12 +30,11 @@ describe JobExecution do
   let(:execution) { JobExecution.new('master', job) }
   let(:deploy) { Deploy.create!(stage: stage, job: job, reference: 'master', project: project) }
 
-  with_job_execution
-  around { |t| ArMultiThreadedTransactionalTests.activate &t }
+  with_full_job_execution
 
   before do
+    execute_on_remote_repo "git tag v1"
     Project.any_instance.stubs(:valid_repository_url).returns(true)
-    create_repo_with_tags('v1')
     user.name = 'John Doe'
     user.email = 'jdoe@test.com'
     project.repository.update_local_cache!
@@ -43,7 +42,6 @@ describe JobExecution do
   end
 
   after do
-    FileUtils.rm_rf(repo_temp_dir)
     FileUtils.rm_rf(repo_dir)
     project.repository.clean!
   end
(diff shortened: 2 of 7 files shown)


app/models/job_execution.rb
```diff
@@ -263,7 +263,7 @@ def make_tempdir
     Dir.mktmpdir("samson-#{@job.project.permalink}-#{@job.id}") do |dir|
       result = yield dir
     end
-  rescue Errno::ENOTEMPTY
+  rescue Errno::ENOTEMPTY, Errno::ENOENT
     result # tempdir ensure sometimes fails ... not sure why ... return normally
```
sandlerr: can we report this to airbrake so we know if it stops happening?

test/models/job_execution_test.rb
```diff
@@ -424,16 +422,22 @@ def with_hidden_errors
 
   describe "#pid" do
     it "returns current pid" do
+      job.command = 'sleep 0.5'
       execution = JobExecution.new('master', job)
-      execution.executor.stubs(pid: 2)
-      execution.pid.must_equal 2
+      JobExecution.start_job(execution)
+      sleep 0.5
```
sandlerr: how likely is this to be a race?

app/models/job_execution.rb
grosser: yes, but I'll put it on ignore so I don't get spammed :)

test/models/job_execution_test.rb
grosser: not very likely since close etc takes 0.1s or so, but having 0.4 would make things clearer ... I'll update

test/models/job_test.rb
```diff
@@ -253,14 +254,19 @@ def create
     end
 
     it "stops a queued job" do
-      active = stub(start!: true, on_complete: true, id: 123)
+      active_job = project.jobs.create!(command: 'cat foo', user: user, project: project, commit: 'master')
```
sandlerr: nitpick: is it better for this testcase to not hit the db?

grosser: why ?

grosser: I'd rather have things more realistic

grosser:
updated

test/models/job_execution_test.rb
grosser: should not happen since git clone takes a bit, but better safe ... changed to 0.4

grosser:
got a few more PRs if you fancy a weekend read ;)
https://github.com/zendesk/samson/pulls/grosser

test/models/job_test.rb
sandlerr: I thought a best practice was to use `new` over `create` unless its quite inconvenient so that the tests run faster. In this test as far as I can see the object is only used in-memory. Trying to read it from the db again could be considered a performance bug. The main reason is so the tests run faster though.

sandlerr:
Lot of people in the office for a weekend :)

test/models/job_execution_test.rb
```diff
@@ -424,16 +422,24 @@ def with_hidden_errors
...
+      execution.wait!
     end
   end
 
   describe "#make_tempdir" do
+    # the actual issue we saw was Errno::ENOTEMPTY ... but that is harder to reproduce
     it "does not fail when directory cannot be removed" do
-      Dir.expects(:rmdir).at_least_once.raises(Errno::ENOTEMPTY)
-      execution.send(:make_tempdir) { 111 }.must_equal 111
+      Airbrake.expects(:notify).with { |e| e.must_include "Notify: make_tempdir error No such" }
```
sandlerr: TIL

test/models/job_test.rb
grosser: true, updated

test/models/job_execution_test.rb
grosser: a bit weird/hacky ... but otherwise I'd have to do `capture` ... 2-for-1 ;)

grosser: ... I guess `.with(matching(xxx))` or something like that could also work ... but this gives a nice error message when it fails

sandlerr: :+1:

PR #1771, opened 2017-02-18
expire projects cache when locks change

grosser:
@craig-day


sandlerr:
hm, could this bug exist in other places? Is there global state other than Lock that affects responses?

grosser:
locks is the only global state I know of that is on many pages

grosser:
I expect more of them to crop up now that we have longer caching

PR #1772, opened 2017-02-18
reset output height to previous hight instead of hardcoded value

grosser:
- when scrolling around output got resized to 550px
 - move following logic into a sane order and name the ids after the button labels
 - do not scroll the user when in expanded mode
 - do not follow/unfollow when in expanded mode

### DIFF
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -13,7 +13,8 @@ $(function () {
       $placeholderPanes = $container.find(".changeset-placeholder"),
       $form = $("#new_deploy"),
       $submit = $form.find('input[type=submit]'),
-      $messages = $("#messages");
+      $messages = $("#messages"),
+      old_height = $messages.css('max-height');
 
   $("#deploy-tabs a[data-type=github]").click(function (e) {
       e.preventDefault();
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/deploys.js
```diff
@@ -126,43 +129,47 @@ $(function () {
...
-    }
-  });
+  $("#output-no-follow").click(function() {
+    following = false;
 
-  $("#output-grow").click(function() {
-    growOutput();
+    shrinkOutput();
 
     $("#output-options > button").removeClass("active");
```
sandlerr: this is a bug right? select "Don't Follow" shouldn't deactivate "Expand"?

sandlerr: or if they're mutually exclusive then we should have this line in the other handler too?

grosser:
it's a 3-way modal "Follow", "Don't Follow", "Expand"

app/assets/javascripts/deploys.js
grosser: the other 3 already have this line, but it was messy ... refactored into a single 'activate' method

grosser:
forgot to push ... updated now

grosser:
I like that

grosser:
... this needs more rework ... merging for now ... can start a new PR from that

PR #1774, opened 2017-02-20
stop sharing global following state

grosser:
@henders slowly getting better :D


grosser:
extracted that whole mess into it's own concerns ... so much better :D

henders:
Nice refactor 👍

PR #1781, opened 2017-02-23
allow exporting of projects to a CSV file

jonmoter:
I wanted to download a CSV file of all the projects in Samson, along with their owners, so we could see which projects belonged to which teams, and so we can easily export the set of projects into a spreadsheet.

/cc @zendesk/samson, @pswadi-zendesk, @JLandersZen, @kazraza 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -24,7 +24,7 @@ class ProjectsController < ApplicationController
       end
 
       format.csv do
-        datetime = Time.now.strftime "%Y%m%d_%H%M"
+        datetime = Time.now.strftime "%Y-%m-%d_%H-%M"
         send_data as_csv, type: :csv, filename: "Projects_#{datetime}.csv"
       end
     end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require 'csv'
+
 class ProjectsController < ApplicationController
   include CurrentProject
 
@@ -19,6 +21,11 @@ class ProjectsController < ApplicationController
 
       format.json do
         render json: Project.ordered_for_user(current_user).all
+      end
+
+      format.csv do
+        datetime = Time.now.strftime "%Y%m%d_%H%M"
+        send_data as_csv, type: :csv, filename: "Projects_#{datetime}.csv"
       end
     end
   end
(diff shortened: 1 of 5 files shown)


irwaters:
looks like a rubocop nag.

app/serializers/project_serializer.rb
```diff
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
 class ProjectSerializer < ActiveModel::Serializer
-  attributes :id, :name, :url, :permalink, :repository_url
+  attributes :id, :name, :url, :permalink, :repository_url, :owner, :created_at
+
+  def self.csv_header
```
grosser: this relies on the ordering of attributes and csv_line to be always the same, which is not tested ... how about 
```
PROPERTIES = [:id, :foo]
attributes *PROPERTIES

def csv_header
  PROPERTIES.map { |p| p.to_s.humanize }
end

def csv_line
  PROPERTIES.each { |p| public_send p }
end
```

app/controllers/projects_controller.rb
```diff
@@ -20,6 +22,11 @@ def index
       format.json do
         render json: Project.ordered_for_user(current_user).all
       end
+
+      format.csv do
+        datetime = Time.now.strftime "%Y%m%d_%H%M"
```
grosser: how about 2017-02-01_13-34

app/controllers/projects_controller.rb
```diff
@@ -100,4 +107,16 @@ def projects_for_user
   def require_project
     @project = (Project.find_by_param!(params[:id]) if params[:id])
   end
+
+  def as_csv
+    projects = Project.order(:id).all
```
grosser: `:id` is the default anyway ?

app/controllers/projects_controller.rb
jonmoter: Okay.  Copy-pasted from the other place we're exporting a CSV.

app/serializers/project_serializer.rb
jonmoter: `csv_line` should be `PROPERTIES.map`, right?

app/controllers/projects_controller.rb
jonmoter: Being explicit about it, since there's a controller test that assumes it's ordered by ID.

jonmoter:
Updated based on feedback.

PR #1786, opened 2017-02-25
Add documentation for GITHUB_HOOK_SECRET

ibrahima:
* Documents how to set up the GIthub webhook
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/docs/ci.md
+++ b/docs/ci.md
@@ -40,7 +40,7 @@ You can combine those 2 above as well.
     * You can add any value to the 'Token' field, as it is not used
 * GitHub
     * You may add a webhook for push events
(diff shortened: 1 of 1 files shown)


grosser:
thx for the PR, but that is not correct ... 
"To verify Github hooks, add a secret value as `GITHUB_HOOK_SECRET` to the environment and the same secret value in the *Secret* field in the Github webhook."

ibrahima:
Ah okay, thanks! I guess I set up GITHUB_HOOK_SECRET and then forgot about it, so I assumed that it had to be set for the webhooks to work at all.

grosser:
It used to be this way, but we changed that a few months ago in
https://github.com/zendesk/samson/pull/1611

PR #1789, opened 2017-02-27
do not blow up on new cluster page when ecr is enabled

grosser:
https://zendesk.airbrake.io/projects/95346/groups/1895106658683005459?tab=comments

@jonmoter

### DIFF
--- a/plugins/kubernetes/app/views/admin/kubernetes/clusters/edit.html.erb
+++ b/plugins/kubernetes/app/views/admin/kubernetes/clusters/edit.html.erb
@@ -22,7 +22,7 @@
       %>
 
       <%= form.actions do %>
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/controllers/admin/kubernetes/clusters_controller.rb
```diff
@@ -68,7 +68,7 @@ def load_default_config_file
       last_cluster.config_filepath
     end
 
-    @context_options = Kubeclient::Config.read(@config_file).contexts if @config_file
+    # @context_options = Kubeclient::Config.read(@config_file).contexts if @config_file
```
jonmoter: Is this instance variable not being used?

grosser: hehe ... forgot while playing locally

Issue #869, opened 2016-04-22
Run db migrations before kubernetes deployment


rata:
Hi!

We need a way to run commands, to at least run the db migrations, before the deployment. We want this to be run in only one place and have logs and no auto-retry if it fails, etc. and in kubernetes ML it was recommended to do it as a another stage, just before the actual deploy.

Can I ask what do you do today to deploy apps on kubernetes that need db migrations? How do you handle that?

It will be nice to have the possibility to run arbitrary commands (maybe inside the docker image, to run the migrations too) but with a way to run the db migrations, it's fine for us for now :)

jonmoter:
That's on my "todo" list as something to figure out.

My thinking was to use a [Kubernetes Job](http://kubernetes.io/docs/user-guide/jobs/), using the same `Build` that you're going to deploy, but running a command like `rake db:migrate` instead.

I haven't yet thought of a good UI or workflow to make that happen. I'm open to ideas.

Another thing I haven't figured out is how an end user of Kubernetes would do the equivalent of a `rails console` for an app running in Kubernetes.

rata:
@jonmoter: A job is an interesting think to be able to schedule in a stage, but I haven't use them. I'm not sure if it can be configured to not-restart if it fails (with a failed migration, you don't want it to restart probably). Do you know? I will try early next week if not :)

Maybe a YAML with jobs, just like the deployment job, and in the UI they can be associated in order with different "stages" ? (I know they are not stages now, but something like that will be needed I think)

jonmoter:
Yup, you can set the [restartPolicy](http://kubernetes.io/docs/user-guide/jobs/#handling-pod-and-container-failures) of a job to `OnFailure` or `Never`.

rata:
On Sun, Apr 24, 2016 at 10:30:05AM -0700, Jon Moter wrote:
> Yup, you can set the
> [restartPolicy](http://kubernetes.io/docs/user-guide/jobs/#handling-pod-and-container-failures)
> of a job to `OnFailure` or `Never`.

Seems like a nice way to do it! It should be useful besides this.

Perhaps when init containers are a reality (not sure if they will be implemented
in k8s or not, yet) schema can be done with init-containers, but this way seems
quite useful meanwhile =)

jbarreneche:
@jonmoter Hi, I'd like to know if you've been working in the posibility of triggering kubernetes jobs for running macros (which would at least allow to have a workaround for running database migrations manually before deploying).

[...]

PR #1796, opened 2017-03-03
PAAS-802 we need to wait until jobs are completed (migrations etc) ..…

grosser:
…. not until they are running

fixes https://github.com/zendesk/samson/issues/869

@jonmoter @irwaters


grosser:
@jbarreneche

jbarreneche:
Seems it should fix the problem 👍 I'll test it this week and confirm.
Thanks!

PR #1803, opened 2017-03-08
allow defining env variables per deploy

grosser:
WIP for https://github.com/zendesk/samson/pull/1761

screen shot 2017-03-08 at 9 40 56 am

TODO:
 - prefill names ?
 - make copy-row not copy the label
 - persist so re-deploy works ... possibly using env plugin
 - highlight prefixed section with a comment + newline after


grosser:
might actually move this into the env plugin so I can reuse it's models ... that could be fun and also make this work for kubernetes ...

grosser:
bit swamped atm, so don't expect much progress ... feel free to fork/work on this in the meantime ... just let me know so I don't duplicate work :)

PR #1807, opened 2017-03-09
namespace pods so it is always clear where they live no matter what t…

grosser:
…hey report to

should be useful for any kind of error / newrelic etc reporting where we often only see the hostname

@zendesk/paas 

... tried this and seems to not work ... need to dig deeper ... alternatively change name of deployment, but that would be sad ...

these don't work either:
```
        pod.beta.kubernetes.io/hostname: my-pod-name
        pod.beta.kubernetes.io/subdomain: foo
```


grosser:
seems impossible without changing the name of the deployment ... closing for now

PR #1810, opened 2017-03-10
feat: add manage projects link under projects menu

tobico:
As a new user to Samson, when I want to add a new project, I first look under the "Projects" menu. However, to add a new project, you actually need to go to the "Manage" menu, and then choose "Projects".

This PR adds a menu item on the "Projects" menu for "Manage Projects", so this common action can be accomplished in a more straightforward and logical manner.

#### Before:

screen shot 2017-03-10 at 14 21 42

#### After:

screen shot 2017-03-10 at 14 21 23

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Medium, adds new item to projects menu

### DIFF
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -24,6 +24,10 @@
                 <% end %>
               </li>
             <% end %>
+            <% if current_user.try(:admin?) %>
+              <li class="divider"></li>
+              <li><%= link_to "Manage Projects", admin_projects_path %></li>
+            <% end %>
           </ul>
         </li>
         <% if DeployGroup.enabled? %>
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -24,7 +24,7 @@
                 <% end %>
               </li>
             <% end %>
-            <% if current_user.try(:admin?) %>
+            <% if current_user&.admin? %>
               <li class="divider"></li>
               <li><%= link_to "Manage Projects", admin_projects_path %></li>
             <% end %>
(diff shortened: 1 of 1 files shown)


tobico:
💇  @sandlerr

app/views/shared/_header.html.erb
```diff
@@ -24,6 +24,10 @@
                 <% end %>
               </li>
             <% end %>
+            <% if current_user&.admin? %>
+              <li class="divider"></li>
+              <li><%= link_to "Manage Projects", admin_projects_path %></li>
```
grosser: can just be `Manage` ... and remove the duplicate link from admin ?

tobico: I wouldn't want to remove the existing link right away, that would cause problems for people that are already used to its current position.

Having the full text "Manage Projects" should make the link easier to parse at a glance, as you get a clear description of the action, instead of having to infer what it does from the context of its parent menu.

grosser:
We already have a few "Manage" links, so I'm not really worried about it.

I doubt anyone will clean up that duplicate link ... but it does not hurt I
guess  ...

app/views/shared/_header.html.erb
```diff
@@ -24,6 +24,10 @@
                 <% end %>
               </li>
             <% end %>
+            <% if current_user&.admin? %>
```
grosser: logged out users don't see this header, so no need for `&.`

tobico:
💇  @grosser

PR #1811, opened 2017-03-10
get rid of distracting active checkboxes on empty stage page

grosser:
I always do a double-check on these ... would be easier to read if they were off by default
also adding an error when leaving all options blank for the slack webhook

screen shot 2017-03-09 at 7 34 20 pm
screen shot 2017-03-09 at 7 34 28 pm
screen shot 2017-03-09 at 8 37 24 pm

### DIFF
--- a/plugins/flowdock/db/migrate/20170310033621_remove_enabled_flag.rb
+++ b/plugins/flowdock/db/migrate/20170310033621_remove_enabled_flag.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+class RemoveEnabledFlag < ActiveRecord::Migration[5.0]
+  class FlowdockFlow < ActiveRecord::Base
+  end
+
+  def up
+    FlowdockFlow.where(enabled: false).each do |flow|
+      puts "Deleting flow from stage #{flow.stage_id}: #{flow.name} #{flow.token}"
+      flow.destroy
+    end
+    remove_column :flowdock_flows, :enabled
+  end
+
+  def down
+    add_column :flowdock_flows, :enabled, default: true, null: false
+  end
+end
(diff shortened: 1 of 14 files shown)


grosser:
@sandlerr

plugins/slack_webhooks/app/models/slack_webhook.rb
```diff
@@ -11,19 +12,23 @@ def deliver_for?(deploy_phase, deploy)
...
       URI.parse(webhook_url)
     rescue URI::InvalidURIError
       false
     end
 
     errors.add(:webhook_url, "is invalid") unless valid
   end
+
+  def validate_used
+    errors.add :base, "Pick a deliery option" if !for_buddy && !before_deploy && !after_deploy
```
sandlerr: delivery missing v

sandlerr: maybe easier to understand "select at least one delivery time" or something?

plugins/flowdock/db/migrate/20170310033621_remove_enabled_flag.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+class RemoveEnabledFlag < ActiveRecord::Migration[5.0]
+  class FlowdockFlow < ActiveRecord::Base
+  end
+
+  def up
+    FlowdockFlow.where(enabled: false).each do |flow|
+      puts "Deleting flow from stage #{flow.stage_id}: #{flow.name} #{flow.token}"
```
sandlerr: if this was being done by samson, it'd leak the tokens to whoever could see the deploy log :)

grosser: they were previously visible in plain sight anyway ... so not too worried ... for more sensitive things we could write them to cache and tell the user where they are ...

plugins/slack_webhooks/app/models/slack_webhook.rb
grosser: fixed

PR #1812, opened 2017-03-10
fix: gracefully handle error when github user is unauthorized

tobico:
When GitHub user is not correctly configured and Samson attempts to create a release, this causes a confusing exception to occur. In this fix, we instead catch the error and pass back `false` to the validation, causing it to display the usual nice error message explaining that write access is required.

#### Before:

screen shot 2017-03-10 at 14 47 14

#### After:

screen shot 2017-03-10 at 14 46 58

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low - just adds nicer error handling

### DIFF
--- a/app/models/release_service.rb
+++ b/app/models/release_service.rb
@@ -15,7 +15,7 @@ class ReleaseService
 
   def can_release?
     GITHUB.repo(@project.user_repo_part).permissions[:push]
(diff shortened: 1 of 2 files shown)


grosser:
❤️ the PRs ... keep em coming :D

PR #1813, opened 2017-03-10
fold empty stage sections

grosser:
we have so many plugins we don't need / don't regularly use ... maybe it's time to fold the sections that are not filled out ...

screen shot 2017-03-09 at 8 59 42 pm

@zendesk/samson


jonmoter:
Rebase and merge?

grosser:
yeah need to look at that again, was not super happy with how it looked ... but overall nice clutter-removal ...

PR #1814, opened 2017-03-10
Restyle project search box

tobico:
Restyled the search box in the "Projects" drop-down menu to make it look a bit more modern.

Also, fixed an issue where an error alert box shows when the "enter" key is pressed while the project search box is focused.

#### Before:

screen shot 2017-03-10 at 16 02 11

#### After:

screen shot 2017-03-10 at 16 09 35

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low - style change

### DIFF
--- a/app/assets/javascripts/directives/handle_dropdown_selection_directive.js
+++ b/app/assets/javascripts/directives/handle_dropdown_selection_directive.js
@@ -16,7 +16,10 @@ samson.directive('handleDropdownSelection', function($timeout) {
           listItems.eq((index + (e.keyCode - 39)) % listItems.length).addClass('selected');
         } else if (e.keyCode == 13) { // enter
           e.preventDefault();
-          selected.find('a').get(0).click();
+          var a = selected.find('a').get(0);
+          if (a) {
+            a.click();
+          }
         } else {
           $timeout(function() {
             if(!selected.is(':visible')) {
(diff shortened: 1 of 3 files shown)


grosser:
idk ... the old one is super simple and intuitive ... also won't break on any browser ... and making 1 element sexy with a bunch of css just leads to more css everywhere else to keep in sync ...

so I'm more on the 👎 ... it looks a bit nicer, but usability is the same or worse ... and imagining that it was this way and someone did the opposite "simplify search box" PR I would merge that ...

tobico:
why do you think usability would be worse? It's the exact same user interaction

also, it's really not much CSS, probably less than the styles the previous design was inheriting through bootstrap.

tobico:
don't underestimate the usability importance of making stuff look nice, either. beautiful things put people into a happy emotional state, making it easier for them to make decisions calmly and carefully

grosser:
The previous UI was a textfield ... doubt it had much css ... and even if
it did, we did not write/maintain it and it was pretty much unbreakable.

I'm thinking usability might be slightly worse since a textfield is obvious
how to use ... a text without any highlighting is a bit novel ... and the
search icon also distracts / might be mistaken for a button.

I don't have anything against it, but I'd also be ok with making the
opposite ('simplify') PR, so I'd rather not merge it ... but go crazy if
you find someone else that wants it :)

PR #1816, opened 2017-03-11
make stars do an ajax request only

grosser:
reloading the whole page is a bit silly and makes it feel sluggish to star projects

@henders @msufa @sbrnunes 

/fyi code partially written by @pswadi-zendesk

### DIFF
--- a/app/controllers/stars_controller.rb
+++ b/app/controllers/stars_controller.rb
@@ -1,23 +1,15 @@
 # frozen_string_literal: true
 class StarsController < ApplicationController
-  before_action :require_project
+  include CurrentProject
 
+  # toggles star by creating/destroying it
   def create
-    current_user.stars.create!(project: @project)
+    if star = current_user.stars.find_by_project_id(current_project.id)
+      star&.destroy
+    else
+      current_user.stars.create!(project: current_project)
+    end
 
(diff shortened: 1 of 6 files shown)


app/controllers/stars_controller.rb
```diff
@@ -1,23 +1,14 @@
...
-  before_action :require_project
+  include CurrentProject
 
   def create
-    current_user.stars.create!(project: @project)
+    if star = current_user.stars.find_by_project_id(current_project.id)
+      star&.destroy
+    else
+      current_user.stars.create!(project: current_project)
+    end
```
jonmoter: Add a comment to the `#create` method saying it's actually doing a create/destroy toggle. Otherwise I'd be pretty confused looking at the implementation.

app/helpers/projects_helper.rb
```diff
@@ -1,24 +1,16 @@
...
-        }
-      else
-        path = stars_path(id: project)
-        options = {
-          class: 'glyphicon glyphicon-star-empty',
-          method: :post,
-          title: 'Star this project'
-        }
-      end
+    stared = current_user.starred_project?(project)
```
jonmoter: Should be "starred"

PR #1824, opened 2017-03-15
remove hotfix feature

grosser:
Saw that we do a lot of github checks to find out if a release or deploy is "hotfix" (include any commit that starts with "HOTFIX")

I never saw this being used, so I'd opt to delete that feature to avoid all these github lookups
on deploy and when displaying release labels

@zendesk/samson
/cc @henders @jonmoter @irwaters

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -36,11 +36,6 @@
   margin-right: 15px;
 }
 
-.label-hotfix {
-  background-color: #f0ad4e;
-  text-transform: uppercase;
-}
-
 a.no-hover:hover {
   text-decoration: none;
 }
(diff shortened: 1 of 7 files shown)


jonmoter:
Yeah, I've been thinking of getting rid of that for a while.  You might want to check with the Sustaining team to see if they use the feature.  Or if people complain, we could add it back as a plugin.

grosser:
@zendesk/sustaining speak up or it goes ... 24h notice ;)

jcheatham:
never used it, don't think anybody in our team has used it or uses it?
I have seen it in use from some other teams, thought it was just a team-style thing, hadn't realized it was some secret menu functionality flag.

jonmoter:
Also, good to know if the @zendesk/qa team uses this feature in Samson.

pdeuter:
I have never used it.  @smedberg ... do you use it?

smedberg:
@pdeuter I don't think I was aware this feature existed- if I've used it, it was probably an accident.

PR #1825, opened 2017-03-16
Add timestamps to output log

adam-hart:
We've had some cases of builds becoming inexplicably long for no apparent reason.  Debugging and general readability might be better with some timestamps.

### Tasks
 - [ ] :+1: from team

### Risks
- Low:  Could cause render errors on execution log.


test/test_helper.rb
```diff
@@ -42,12 +42,14 @@ def create_default_stubs
     SseRailsEngine.stubs(:send_event).returns(true)
     Project.any_instance.stubs(:clone_repository).returns(true)
     Project.any_instance.stubs(:clean_repository).returns(true)
+    DateTime.stubs(:now).returns(DateTime.new(2001,2,3,4,5,6))
```
grosser: not a fan of this ... prefer doing it in the affected tests ... maybe add a global `freeze_time '2001-02-01 03:04:05'` kind of helper

grosser:
@jonmoter you wanted this too ?

sandlerr:
probably doesn't matter much but should it be coerced to utc?

grosser:
only alternative I could think of is showing time since last command ...
ideally don't just dump it into the output but have it as metadata like travis does ... but I guess this is simpler ... 
I doubt this helps much since all our deploys are executed from a single command ... so you'd see that timestamp and then nothing ...

sandlerr:
certainly better than the hack in https://github.com/zendesk/samson/pull/1243 :)

craig-day:
This is nice, but don't we really want this in `zendesk_deployment`? Otherwise this will only give a single timestamp for the `bundle exec cap deploy` command which is usually the last one anywho. 

On that note, I'm currently looking at improving logging in zendesk_deployment

jonmoter:
I don't think this hurts, but like @craig-day and @grosser, it won't help much either, since we generally execute a single `bundle exec cap deploy` command that runs for however many minutes.

If you wanted to inject timestamps into the output like Travis does, you'd have to add logic in `TerminalExecutor#execute` that splits the output on `"\n"`, and injects a timestamp in there somehow. Which might be interesting, or possibly terrible.

pdeuter:
:chicken:

adam-hart:
Alright, the last commit should inject the timestamp into shell lines as well.

sandlerr:
does it affect the terminal replace functionality here? https://github.com/zendesk/samson/blob/4eacc6678df3e03edb7a8e3135f88fc78ce17863/app/models/terminal_output_scanner.rb#L39

Although IMO it would probably work better if we stopped trying to mimic support for this type of stuff and just told apps they are running in a non-interactive / non-tty shell

app/models/terminal_executor.rb
```diff
@@ -31,7 +31,7 @@ def execute!(*commands)
     output, input, pid = PTY.spawn(whitelisted_env, script(commands), options)
     record_pid(pid) do
       begin
-        output.each(256) { |line| @output.write line }
+        output.each(256) { |line| @output.write "[#{DateTime.now.utc.strftime("%T")}] #{line}" }
```
jonmoter: Hmm, does this actually work?  'cause the way I read this, that block is chunking output every 256 characters, not every newline.  So this would inject a timestamp every 256 characters, wherever that would be.

`chunk` would be a better name for this variable than `line`.  Unless I'm missing something?

adam-hart:
What's everyone here's opinion on adding a margin to the output log with fewer timestamps? Like so:

grosser:
margin looks weird to me ... maybe because the other output wraps into it ...

brianburnszd:
Is it possible to only include the timestamp when a command is executed (detect `»` maybe, or is that too brittle?). The main thing is the timestamps are present!

adam-hart:
It's possible, but we want timestamps on every line even when it's not an explicit terminal command.  Otherwise we'll lose timestamps on scripts, like `bundle exec cap deploy`.

adam-hart:
Alright, back to one stamp per line.

jcheatham:
could throw in milliseconds if you wanted to spice it up :D

grosser:
meh ... seconds is fine ...

app/models/output_buffer.rb
```diff
@@ -86,6 +89,18 @@ def each(&block)
 
   private
 
+  def inject_timestamp(data)
+    if data.is_a?(String)
+      stamped_data = ""
+      data.each_line do |line|
+        stamped_data += "[#{Time.now.utc.strftime("%T")}] #{line}"
```
grosser: prefer `<<` to avoid garbadge

app/models/output_buffer.rb
```diff
@@ -35,6 +37,7 @@ def puts(line = "")
   end
 
   def write(data, event = :message)
+    data = inject_timestamp(data)
```
grosser: maybe `data = inject_timestamp(data) if data.is_a?(String)`

app/models/terminal_executor.rb
```diff
@@ -31,7 +31,7 @@ def execute!(*commands)
     output, input, pid = PTY.spawn(whitelisted_env, script(commands), options)
     record_pid(pid) do
       begin
-        output.each(256) { |line| @output.write line }
+        output.each(256) { |chunk| @output.write chunk }
```
grosser: ❤️

test/models/docker_builder_service_test.rb
```diff
@@ -151,11 +151,15 @@ def execute_job
     describe "with build_command" do
       let(:command) { Command.create!(command: "echo foo\r\necho bar") }
 
-      before { project.update_column(:build_command_id, command.id) }
+      before do
+        project.update_column(:build_command_id, command.id)
+        freeze_time
```
grosser: make sure we unfreeze time in test-helper

test/test_helper.rb
```diff
@@ -78,6 +78,10 @@ def refute_valid(record, error_keys = nil)
     end
   end
 
+  def freeze_time
```
grosser: a bit misleading ... would have expected this to only do a `Timecop.freeze` ... but I guess this is ok

PR #1827, opened 2017-03-16
use combo boxes for env selection

grosser:
screen shot 2017-03-15 at 11 30 03 pm
screen shot 2017-03-15 at 11 29 43 pm


grosser:
add row does not work with these ... reverting :/

PR #1828, opened 2017-03-16
PAAS-864 enforce required environment variables

grosser:
verify all `env` with `value: filled-by-samson` are filled out

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -11,7 +11,7 @@ pending.**
 
 ## Overview
 
-The plugin works by communicating with one or more Kubernetes clusters via 
+The plugin works by communicating with one or more Kubernetes clusters via
 their APIs. It uses a ruby gem called [kubeclient](https://github.com/abonas/kubeclient).
 It is possible for Samson to communicate with multiple clusters, e.g. one
 cluster running locally on your laptop, a second cluster running in an AWS
@@ -20,7 +20,7 @@ configuring clusters.
 
 Once you have connected to one or more clusters, you can configure a project
 to deploy it to Kubernetes. That involves:
- 
+
 1. Set up your project so you can create Builds with Docker images
 2. Configuring one or more "Roles" for the project
 3. Creating one or more Kubernetes configuration files in the project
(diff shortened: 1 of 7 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -409,7 +409,7 @@ def deploy_group_for_pod(pod, release)
     # verify with a temp release so we can verify everything before creating a real release
     # and having to wait for docker build to finish
     def verify_kubernetes_templates!
-      release = Kubernetes::Release.new(project: @job.project, git_sha: @job.commit)
+      release = Kubernetes::Release.new(project: @job.project, git_sha: @job.commit, git_ref: 'master')
```
jonmoter: Why was this change made?

grosser:
it is written to an env var and setting env vars fails without it

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
jonmoter: You're adding `git_ref` as a param to this method, but I don't see it used in the method body. Is there some ruby magic going on here I don't understand?

jonmoter:
So the idea is that in your K8S YAML files, if you have environment variables with `filled-by-samson` as a value, we'll confirm they're replaced?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -171,10 +179,16 @@ def project
       @project ||= @doc.kubernetes_release.project
     end
 
+    def env
+      (container[:env] ||= [])
```
jonmoter: Should we do the checking and replacing on all the containers, rather than just the first one?

grosser:
that's another PR / issue ... atm we support 1 :)

grosser:
yes

plugins/kubernetes/README.md
```diff
@@ -97,16 +97,17 @@ requires:
...
+repository. That file is expected to contain the definition of a Deployment/Daemonset/Service/Job/etc.
 Samson will read in that file from the project repository, make some
 changes (like updating the Docker image and adding labels), and send that
 to the Kubernetes API.
 
 Each role is expected to have its own configuration file. The contents of
 the files will likely be similar for each role, though it will likely
-have a different command or liveness probe.
+have a different command or liveness probe
+([kucodiff](https://github.com/grosser/kucodiff can be used to make sure they stay in sync).
```
jonmoter: I think you're missing a closing parenthesis in your link markup.

PR #1834, opened 2017-03-17
add convenient option to find only bypassed deploys

grosser:
screen shot 2017-03-17 at 4 14 29 pm
screen shot 2017-03-17 at 4 21 11 pm

@dragonfax @JLandersZen


dragonfax:
👍

dragonfax:
Ah , when we spoke I didn't realize you meant combining the 2 checkboxes into one checkbox. That is a much better solution.

PR #1836, opened 2017-03-18
clen up buddy check mess

grosser:
- test all code
 - fail on invalid numbers
 - make grace_period and time_limit both return seconds
 - deprecate BYPASS_JIRA_EMAIL in favor of comma separated BYPASS_EMAIL
 - remove tests that are already covered by other classes

### DIFF
--- a/lib/buddy_check.rb
+++ b/lib/buddy_check.rb
@@ -1,29 +1,26 @@
 # frozen_string_literal: true
 module BuddyCheck
-  module_function
+  class << self
+    def enabled?
+      "1" == ENV["BUDDY_CHECK_FEATURE"]
+    end
 
-  def enabled?
-    "1" == ENV["BUDDY_CHECK_FEATURE"]
-  end
+    # how long can the same commit be deployed ?
+    def grace_period
+      Integer(ENV["BUDDY_CHECK_GRACE_PERIOD"].presence || "4").hours
+    end
 
-  def bypass_email_address
-    ENV["BYPASS_EMAIL"]
-  end
+    def time_limit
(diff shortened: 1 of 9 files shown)


lib/buddy_check.rb
```diff
@@ -1,29 +1,26 @@
...
-  def grace_period
-    (ENV["BUDDY_CHECK_GRACE_PERIOD"].presence || "4").to_i.hours
-  end
-
-  def bypass_jira_email_address
-    ENV["BYPASS_JIRA_EMAIL"]
-  end
+    # how long can the same commit be deployed ?
+    def grace_period
+      Integer(ENV["BUDDY_CHECK_GRACE_PERIOD"].presence || "4").hours
```
sandlerr: why Integer over to_i?

test/lib/buddy_check_test.rb
```diff
@@ -1,130 +1,90 @@
...
+    it "is enabled when 1" do
+      with_env BUDDY_CHECK_FEATURE: "1" do
+        assert BuddyCheck.enabled?
+      end
+    end
 
-    stage.stubs(:production?).returns(true)
-    service.expects(:confirm_deploy!).once
+    it "is disabled when not 1" do
+      with_env BUDDY_CHECK_FEATURE: "true" do
```
sandlerr: might confuse people, maybe crash instead?

sandlerr:
somehow deleting a lot of tests and increasing coverage, awesome :D

grosser:
Blows up when the value is not a number ... See tests

grosser:
Hmm true ... Future pr ... For now keeping old behavior

grosser:
Deleting duplicate tests ;)

PR #1837, opened 2017-03-18
make required env less vorbose and less error prone

grosser:
atm if a user has a typo or does not use samson weird default values will be set,
also the syntax is very verbose.

so switch to a simpler configuration style that uses an annotation key

supported formats:
```
required_env: FOO,BAR
required_env: >
  FOO
  BAR
required_env: |
  FOO
  BAR
```


grosser:
@jonmoter @irwaters need feedback here

jonmoter:
Small nitpick: would be nice to name the annotation something like `samson/required_env`, to make it more obvious it's an annotation that Samson reads.

For example, the [annotations that prometheus uses](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml#L82-L90) are of the form `prometheus.io/scrape`, `prometheus.io/scheme`, etc.

PR #1840, opened 2017-03-20
add useages to group edit view

grosser:
hide delete when it is not useable

screen shot 2017-03-20 at 2 25 36 pm
screen shot 2017-03-20 at 2 24 07 pm


plugins/env/app/views/admin/environment_variable_groups/form.html.erb
```diff
@@ -16,6 +16,13 @@
       <%= form.input :comment, as: :text_area, input_html: {size: "80x4"} %>
     </fieldset>
 
+    <% if @group.projects.length > 0 %>
```
grosser: using length here so we don't do a count and then a fetch right after

PR #1841, opened 2017-03-21
cleanup abandomed environment variables

grosser:
we are leaving join tables and environment vars behind ... not nice ...

TODO:
 - [x] run queries in production and make sure they are sane


grosser:
not too many rows and all queries completed fast ... also found only things that were broken ... so looks fine database wise, only risk is that it's not getting all I'd say

PR #1850, opened 2017-03-27
Feature/macros with custom env

jbarreneche:
Since the ability to run custom [jobs was removed](https://github.com/zendesk/samson/pull/1761), we discussed the possibility of setting environment variables. I know macros are a bit flunky, but stages would mean having superflous stages (unrelated to tasks that need to be run once in a while) and this implementation is enough  for our use case.

image

image

### TODO
- [ ] unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


jbarreneche:
Hi! Please let me know what you think. I'll add the tests later if the implementation/workflow it's ok for you.

grosser:
This is a fine temp hack for your usecase. 
Minor issues: the extra button is a bit distracting ... and it's hard to remember which env vars are allowed ...

I want to get rid of macros since they are weird/duplicating functionality ... 

So I'm thinking I'd like to take this more in the direction of:
Add a checkbox to stage to make it 'no branch selection' ... so they can be 'pretend macros' ... even live in that tab if they want ... or I guess better underneath the regular stages ... and their deploy buttons change to 'Execute' ... 

Then have a "Configureable env" section in stage config where you can say 'pick FOO_BAR and BAR_FOO' ... and only then we show this extra step ... maybe part of the `env` plugin ... are you using the env plugin ?

jbarreneche:
Regarding the stages as macros, I wouldn't mind selecting the branch/revision from where to run the task. When we are doing a rework on a task we some times want to run it from a branch in a previous environment (instead of having duplicated tasks per environment).

We are not using the env plugin, as we don't normally use `.env` files (and from the README I don't see any other feature in it). Anyway, the environment variables should be set in samson and not on the servers (these tweak our capistrano task).

I think I get more or less the idea you have, but it's much more work for very low payoff for us (and we need this as soon as possible). We really want to have our samson version as similar as the original, but if this doesn't works for you (or it needs big changes) we can keep it in our fork.

Thanks!

grosser:
closing this since macros are dead 💀

PR #1852, opened 2017-03-28
when reloading rails in development periodical state is destroyed and…

grosser:
… we fail when calling overdue?

@dbuchi 

```
rails c
reload!
Samson::Periodical -> global state should still be there
```


dbuchi:
+1

PR #1853, opened 2017-03-28
WIP tag finder

grosser:
@dbuchi


grosser:
decided to do this by (ab)using deployments api

PR #1861, opened 2017-03-31
load absolute path to avoid duplicate loading via eager-load

grosser:


grosser:
prints warnings ... looks like a bug ... so better fix :)

PR #1872, opened 2017-04-04
jobs were running in before-deploy but checked for stability

grosser:
that lead to the second run of wait_for_resources_to_complete to start at the stability test and
then crash instantly

also adding sanity check, so if this ever happens in the future we will get a nice error

this will lead to weird situations when someone is not using prerequisite, but we have to migrate that at some point anyway ... so why not now ...


grosser:
made this even better by removing the duplicate logic in pod and resource ... now only resource knows if it is a prerequisite or not

PR #1876, opened 2017-04-05
OP-27926 let viewers see the stage so they can see what it does and see all de…

grosser:
…ploys

@jonmoter @irwaters @zen-aglasman @dtatarsky


jonmoter:
Agreed.  Seems bizarre that someone would have the ability to do a deploy, but not see what it does.

grosser:
This change is not giving show access to deployers, but giving show access to viewers ... so they can see what the stage does and what was deployed.

ashmckenzie:
YIS, thanks @grosser, you are unstoppable!

Issue #1880, opened 2017-04-06
Circle CI cancelled deployment


mirago:
Hi guys, first of all thank you for samson, we started using it a few days ago and it already makes our life easier. This is rather a question than an issue, so sorry if this is not the place for questions.

We use samson with circle ci, after the build is done there is a webhook to samson and it then deploys the docker image to kubernetes. We do not use kubernetes plugin however. Yesterday I found one deployment, and it says 'Circle ci cancelled' it. It took about 8 minutes, which is unusual, as usually it takes about 3 minutes to deploy. Do you happen to know any reasons why this cancel happened? Does it timeout after certain amount of time? Is there any way to debug this?

grosser:
Webhooks don't have a `cancel` callback, so that means something other happened.
Looking at the code we never write down who actually cancelled the deploy ... so the message is misleading ... should be something like "Circle CIs deploy was cancelled" ... https://github.com/zendesk/samson/pull/1881

Actual reason could be:
 - someone cancelling from the UI ... would find a DELETE request to `#destroy` in samsons logs
 - some kind of automatic failure
 - someone canceling it from the commandline

mirago:
I see. Thanks for this, will have a look if there are any logs. Happy to close this.

grosser:
also storing canceller with https://github.com/zendesk/samson/pull/1882

PR #1881, opened 2017-04-06
do not claim the deployer has cancelled/is cancelling the deploy

grosser:
... and other wording improvements

fixes https://github.com/zendesk/samson/issues/1880


app/models/deploy.rb
```diff
@@ -42,16 +40,23 @@ def job_execution_queue_name
     "stage-#{stage.id}" unless stage.run_in_parallel
   end
 
-  def summary
-    "#{job.user.name} #{deploy_buddy} #{summary_action} #{short_reference} to #{stage.name}"
+  # job has almost identical code, keep it in sync
+  def summary(global: false)
```
jonmoter: I don't understand what `global` means in this context.  Better name for variable?

app/models/deploy.rb
```diff
@@ -42,16 +40,23 @@ def job_execution_queue_name
...
+    project_name = " #{project&.name}" if global
+    if ["cancelled", "cancelling"].include?(status)
+      "#{job.user.name}s deploy#{deploy_buddy} of #{short_reference} to#{project_name} #{stage&.name} is #{status}"
```
jonmoter: Should have an apostrophe before the first "s"

grosser:
we use `global` in a few places to mean `not scoped to the project` ...
when we want to show a deploy but it's not on the project page
... could change to `project_scoped: false` ... is that clearer ?

jonmoter:
Okay, I get it.  So you're displaying the summary in a context that isn't specific to the project.

How about just `show_project_name: true`?

grosser:
`show_project` then ... I'll update the other usages too ...

PR #1888, opened 2017-04-08
Add more information to the github deployments payload

hkjorgensen:
Extend the Github plugin, so it provides more information; "user_email" (?string), "buddy" (?string), "buddy_email" (?string) and "production" (bool)

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team @zendesk/samson

### DIFF
--- a/plugins/github/app/models/github_deployment.rb
+++ b/plugins/github/app/models/github_deployment.rb
@@ -26,9 +26,19 @@ class GithubDeployment
     {
       accept: DEPLOYMENTS_PREVIEW_MEDIA_TYPE,
       force: true,
-      payload: {deployer: @deploy.user.name}.to_json,
+      payload: payload.to_json,
       environment: @stage.name,
       description: @deploy.summary
+    }
+  end
+
+  def payload
+    {
+      deployer: @deploy.user.name,
+      deployer_email: @deploy.user.email,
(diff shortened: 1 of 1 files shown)


plugins/github/app/models/github_deployment.rb
```diff
@@ -26,12 +26,22 @@ def deployment_options
...
     }
   end
 
+  def payload
+    {
+      deployer: @deploy.user.name,
+      deployer_email: @deploy.user.email,
+      buddy: @deploy.buddy_name,
+      buddy_email: @deploy.buddy_email,
+      production: @stage.production?
```
grosser: maybe better to send `environment` ?

hkjorgensen: @grosser I could. I'm just interested in `production?` in our use case. How would the environment look like?

grosser:
hmmm I guess that's not as easy since this might be going to multiple
environments ... if we just want the first it would be
stage.deploy_groups.first.environment.name

hkjorgensen:
Ok thanks, sticking with production for now.

grosser:
are you reading that back via the api or is it visible in github UI ?

grosser:
ok if I change this to ?

```
buddy: {
  name: foo
  email: bar
}
```

PR #1889, opened 2017-04-08
remove binary builder

grosser:
/cc @zendesk/samson @sandlerr @jsdnxx 

looks like we don't use it anymore ... whoever still wants it can extract it into a gem ... ideally use the new pre-build command foobar ... searched our repos for Dockerfile.build and did not find anything ...

also added a conditional down migration that will print the active stages in case we have to revert

found a few projects still using it with https://github.com/grosser/github-grab ... will follow up with them before merging ... 18 projects to go ...

... all done now :)


grosser:
validating via

```ruby
bad = Stage.where(docker_binary_plugin_enabled: true).pluck(Arel.sql('distinct project_id'))
used = Project.where(id: bad).select do |project|
  puts project.name
  [BinaryBuilder::DOCKER_BUILD_FILE, 'build.sh'].all? do |file|
    out = `cd #{project.repository.repo_cache_dir} && git ls-tree master #{file}`
    out != ""
  end
end
```

grosser:
all good now and ready to go @zendesk/bre

app/models/image_builder.rb
```diff
@@ -24,6 +24,28 @@ def build_image(dir, build, executor, tag_as_latest:, **args)
       end
     end
 
+    # store logins in a temp file and make it not accidentally added via `ADD .`
+    def local_docker_login
+      Dir.mktmpdir 'samson-tmp-docker-config' do |docker_config_folder|
+        # copy existing credentials
+        regular_config = File.join(ENV["DOCKER_CONFIG"] || File.expand_path("~/.docker"), "config.json")
```
ragurney: Could maybe add `DOCKER_CONFIG` to `env.example`?

db/migrate/20181111003546_remove_binary_builder.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
```
ragurney: Should this be in the plugin's `db/migrate`?

app/models/image_builder.rb
ragurney: Could add `DOCKER_CONFIG` to `.env.example`?

db/migrate/20181111003546_remove_binary_builder.rb
grosser: can't really be there since I'm deleting that plugin :D 
... but I guess it needs to not blow up when column is not there

ragurney: Ah 🤦‍♂️ makes sense

app/models/image_builder.rb
grosser: done

grosser:
had to add back the docker-api since it was still used ... but should be easy to remove later ... but that's another PR :)

db/migrate/20181111003546_remove_binary_builder.rb
```diff
@@ -0,0 +1,8 @@
...
+class RemoveBinaryBuilder < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :stages, :docker_binary_plugin_enabled
+  rescue StandardError # rubocop:disable Lint/HandleExceptions
```
esselius: This `rescue` doesn't seem to work and broke migrations for us :(

PR #1893, opened 2017-04-10
hashify use payload for simple future additions

grosser:


hkjorgensen:
Will the buddy data be applied later - when the users approves?

grosser:
no, this is before_deploy, so it should be after buddy approval ... but not
all deploys have a buddy

PR #1898, opened 2017-04-12
add deprecation notice to binary builder since it will take a few wee…

grosser:
…ks to get everything ready for https://github.com/zendesk/samson/pull/1889

screen shot 2017-04-11 at 5 09 38 pm

### DIFF
--- a/docs/pre_build_command_example.md
+++ b/docs/pre_build_command_example.md
@@ -1,16 +1,32 @@
-Example `build.sh` that uses a secondary `Dockerfile.build` to generate artifacts the main `Dockerfile`
+Example `build_binary.sh` that uses a secondary `Dockerfile.build` to generate artifacts the main `Dockerfile`
 cannot produce without adding secrets or installing dev-tools that increase the image size.
 
-```Bash
-# This script prepares the target folder with compiled classes that the main Dockerfile needs to build the main image
-# This can be used both locally and on samson.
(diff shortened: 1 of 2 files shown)


docs/pre_build_command_example.md
```diff
@@ -1,4 +1,4 @@
-Example `build.sh` that uses a secondary `Dockerfile.build` to generate artifacts the main `Dockerfile`
+Example `build_binary.sh` that uses a secondary `Dockerfile.build` to generate artifacts the main `Dockerfile`
```
henders: I think this was right and used `build.sh` to build inside the docker container.

henders: https://github.com/zendesk/samson/blob/6df74907d9c1892e7aa2914dd686d75b65103290/plugins/docker_binary_builder/app/models/binary_builder.rb#L4

henders: I guess this is a doc for doing things the new way? Though it's not clear about why you would name it `build_binary.sh`, and where you would call it from. We can probably just delete this doc now that we can build scala/java objects on the Samson host itself.

henders: No need for Dockerfile.build anymore at all right?

grosser: building directly on samson also works and would avoid this whole docker mess, but it's more brittle ...
I named it `build_binary.sh` so existing stages will opt-out of binary builder autoatically
the call to build_binary.sh should be in a command ... I'll add that ...

grosser: looks better ?

henders:
👍

PR #1899, opened 2017-04-12
reuse commandexecutor and add timeout for deploys

grosser:
- remove duplication of code execution and all the error handling that comes with it
 - allow setting timeouts for deploys

Current issues:
 - <del>`script` messes up the log stream and the test output ... </del>
 - <del>cannot use `script` with `pgroup: true` or IO.read just hangs... have to verify if kill still works cleanly</del>

@henders


grosser:
@zendesk/samson FYI might break in some unexpected ways ...

grosser:
seems to not work on travis ... works fine in docker and locally

using `script` is most likely to blame for that ... overall not as nice as I'd hoped ... closing ...

PR #1904, opened 2017-04-14
return the last successful deploy of each project in JSON/CSV

jonmoter:
When dumping Samson data into a spreadsheet, it's handy to know the last time a project was successfully deployed, so you can figure out whether the project is still being used.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -21,14 +21,13 @@ class ProjectsController < ApplicationController
       end
 
       format.json do
-        @projects = Project.ordered_for_user(current_user).all
-        render json: { projects: as_json }
+        projects = Project.ordered_for_user(current_user).all
+        render json: { projects: projects_as_json(projects) }
       end
 
       format.csv do
-        @projects = Project.order(:id).all
         datetime = Time.now.strftime "%Y-%m-%d_%H-%M"
-        send_data as_csv, type: :csv, filename: "Projects_#{datetime}.csv"
+        send_data projects_as_csv(Project.all), type: :csv, filename: "Projects_#{datetime}.csv"
       end
     end
   end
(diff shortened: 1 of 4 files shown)


app/controllers/projects_controller.rb
```diff
@@ -21,10 +21,12 @@ def index
       end
 
       format.json do
-        render json: Project.ordered_for_user(current_user).all
+        @projects = Project.ordered_for_user(current_user).all
+        render json: { projects: as_json }
```
grosser: as_json -> projects_as_json ?

grosser: maybe pass the projects to that method

app/controllers/projects_controller.rb
```diff
@@ -21,10 +21,12 @@ def index
...
       end
 
       format.csv do
+        @projects = Project.order(:id).all
```
grosser: `.reorder` or this will do nothing

grosser: ... does not have a default scope ... so no need

app/controllers/projects_controller.rb
```diff
@@ -110,14 +112,36 @@ def require_project
     @project = (Project.find_by_param!(params[:id]) if params[:id])
   end
 
-  def as_csv
-    projects = Project.order(:id).all
+  # Determine the timestamp of the successful deploy for each project, and store
+  # the Datetime in a hash that is indexed by project id
+  def project_last_deployed_at
```
grosser: projects_last_deployed_at

grosser: better as a model method ... easier to test too

app/controllers/projects_controller.rb
```diff
@@ -110,14 +112,36 @@ def require_project
...
+    @project_last_deployed_at ||= begin
+      query = Deploy.successful.select('max(created_at) as last_deploy_at').group(:project_id)
```
grosser: work with pluck to reduce overhead ?

grosser: query -> deploy_times ?

app/controllers/projects_controller.rb
```diff
@@ -110,14 +112,36 @@ def require_project
...
+    end
+  end
 
-      projects.each do |project|
-        csv << ProjectSerializer.new(project).csv_line
+  def as_json
+    @projects.map do |project|
+      project.as_json.merge(
+        'last_deployed_at' => project_last_deployed_at[project.id]
+      )
```
grosser: prefer faster assign over merge

app/controllers/projects_controller.rb
```diff
@@ -110,14 +112,36 @@ def require_project
...
+    end
+  end
+
+  def as_csv
```
grosser: some method-level docs on what hackery this does

test/controllers/projects_controller_test.rb
```diff
@@ -50,7 +50,16 @@
...
+        projects.length.must_equal 1
+        projects.first['name'].must_equal 'foo'
+      end
+
+      it "returns the last deploy date in JSON" do
+        get :index, params: {format: 'json'}
+        result = JSON.parse(response.body)
+        project = result['projects'].first
+        project['last_deployed_at'].must_start_with '2014-01'
       end
```
grosser: test for csv too

app/controllers/projects_controller.rb
jonmoter: If it were a model method, it would do n+1 database queries.  Doing it this way allows me to do a single query that will load them all at once.

grosser:
class level model method

jonmoter:
Updated based on feedback.  Just need to write a test for `Deploy.last_deploys_for_projects`

PR #1918, opened 2017-04-18
include deployer and URL of last deploy in projects csv/json

jonmoter:
Follow-on from #1904.  Also return the email of the person who did the last deploy, and the URL of the last deploy.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -112,7 +112,11 @@ class ProjectsController < ApplicationController
   end
 
   def last_deploy_for(project)
-    @projects_last_deployed_at ||= Deploy.successful.last_deploys_for_projects
+    @projects_last_deployed_at ||= Deploy.successful.
+      last_deploys_for_projects.
+      includes(:project, job: :user).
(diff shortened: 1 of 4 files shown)


app/controllers/projects_controller.rb
```diff
@@ -112,7 +112,11 @@ def require_project
   end
 
   def last_deploy_for(project)
-    @projects_last_deployed_at ||= Deploy.successful.last_deploys_for_projects
+    @projects_last_deployed_at ||= Deploy.successful.
+      last_deploys_for_projects.
+      includes(:project, job: :user).
```
grosser: goldiloader should take care of that ... did it not ?

jonmoter: Haven't actually tried. I'm just accustomed to be explicit about that sort of thing to avoid N+1 queries.

grosser:
I prefer letting goldiloader handle this ... so I never forget and never
add unnecessary things :)

PR #1923, opened 2017-04-20
fix deploy group check boxes not working for names with spaces

grosser:
- use permalink since that is unique and a good selector
 - remove id logic ... only rely on data-target and class
 - use jquery properly to avoid a few iterations


henders:
lgtm 👍

grosser:
found even more to clean up after the initial round ... looking good now :D

PR #1927, opened 2017-04-20
allow seeding with pods

grosser:
we had job/deploy hardcoded ... but any `primary` resource should be able to see ... 
pods are a bit special since they don't have the `spec.template` indentation, but that's easy to work around


henders:
LGTM 👍

PR #1928, opened 2017-04-20
log/notify on failure to report to slack as we do with connection errors

grosser:
got all our slack hooks silently failing with a 404 without anyone noticing ... should report that ...

### DIFF
--- a/plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
+++ b/plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
@@ -36,10 +36,13 @@ module SamsonSlackWebhooks
       payload[:channel] = webhook.channel unless webhook.channel.blank?
       payload[:attachments] = attachments if attachments.present?
 
-      Faraday.post(webhook.webhook_url, payload: payload.to_json)
-    rescue Faraday::ClientError => e
-      Airbrake.notify(e)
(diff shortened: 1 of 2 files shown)


plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
```diff
@@ -36,10 +36,13 @@ def deliver_message_via_webhook(webhook:, message:, attachments:)
       payload[:channel] = webhook.channel unless webhook.channel.blank?
       payload[:attachments] = attachments if attachments.present?
 
-      Faraday.post(webhook.webhook_url, payload: payload.to_json)
-    rescue Faraday::ClientError => e
-      Airbrake.notify(e)
-      Rails.logger.error("Could not deliver slack message to webhook #{webhook.webhook_url}: #{e.message}")
+      begin
```
henders: why bother with this scoped `begin`? Seems unlikely an exception would be thrown in the above 3 lines?

henders:
👍  regardless of comment

plugins/slack_webhooks/lib/samson_slack_webhooks/slack_webhooks_service.rb
henders: And even if it was, probably still want to know about it

grosser: RuntimeError could be coming from anywhere above

PR #1929, opened 2017-04-20
global slack notification on buddy requests

grosser:
will use this for internal auditing

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -97,6 +97,7 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # SLACK_CLIENT_ID= # optional, see plugins/slack_app/README.md
 # SLACK_CLIENT_SECRET= # optional, see plugins/slack_app/README.md
 # SLACK_VERIFICATION_TOKEN= # optional, see plugins/slack_app/README.md
+# SLACK_GLOBAL_BUDDY_REQUEST="<webhook_url>#<channel>" # send notifications for all buddy requests to this channel
 
 ## Export job cleanup
 ## EXPORT_JOB_DOWNLOADED_AGE determines how long a csv export job and file should
--- a/plugins/slack_webhooks/test/samson_slack_webhooks/samson_plugin_test.rb
+++ b/plugins/slack_webhooks/test/samson_slack_webhooks/samson_plugin_test.rb
@@ -14,10 +14,23 @@ describe SamsonSlackWebhooks do
       Samson::Hooks.fire(:buddy_request, deploy)
     end
 
-    it "does not send notifications when disabled" do
-      webhook.buddy_request = false
-      SlackWebhookNotification.any_instance.expects(:deliver).never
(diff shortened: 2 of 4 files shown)


.env.example
```diff
@@ -97,6 +97,7 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # SLACK_CLIENT_ID= # optional, see plugins/slack_app/README.md
 # SLACK_CLIENT_SECRET= # optional, see plugins/slack_app/README.md
 # SLACK_VERIFICATION_TOKEN= # optional, see plugins/slack_app/README.md
+# SLACK_GLOBAL_BUDDY_REQUEST="<webhook_url>#<channel>" # send notifications for all buddy requests to this channel
```
henders: Using `#` as a delimiter might confuse things, since the above line has 3 of them :)

And if they forget to quote the string, they'll only get the first one.

henders: Maybe use the `?` character as that would be a natural URI break? I guess that would confuse things as well if you ever passed query params to the slack URL.

plugins/slack_webhooks/test/samson_slack_webhooks/samson_plugin_test.rb
```diff
@@ -14,10 +14,21 @@
...
+      it "does not send notifications when disabled" do
+        SlackWebhookNotification.any_instance.expects(:deliver).never
+        Samson::Hooks.fire(:buddy_request, deploy)
+      end
+
+      # using full http test here to make sure our hackery works
+      it "sends global hook when configured" do
+        with_env SLACK_GLOBAL_BUDDY_REQUEST: 'http://foo.com/baz#bar' do
+          stub_request(:post, "http://foo.com/baz")
+          Samson::Hooks.fire(:buddy_request, deploy)
```
henders: Maybe also verify that `SlackWebhook.any_instance.expects(:new).with(webhook_url: url, channel: channel)`? Just to verify parsing.

.env.example
grosser: yeah I know ... make a better suggestion :D 
`:` is already in `http://` ... `;` is weird ... `@` might collide with basic auth ... `,` is for multiple entries

plugins/slack_webhooks/test/samson_slack_webhooks/samson_plugin_test.rb
henders: Warning: Above mocha line was completely made up as I didn't look up the syntax! :)

grosser: I'd rather verify the payload then ... but not a bad idea

.env.example
henders: `|`  :)

henders: The old classic

grosser:
pipe and # are the same kind of evil for bash ... and # makes more sense
since it's the channel prefix anyway ?

henders:
Fair enough! 👍

PR #1930, opened 2017-04-20
fix rack blowing up when trying to soft-restart puma because the pid …

grosser:
…still exists

needs https://github.com/rack/rack/issues/1159 or https://github.com/puma/puma/issues/1060 to get resolved


config/environments/development.rb
```diff
@@ -5,10 +5,10 @@
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  config.cache_classes = true
```
henders: These 2 changes will screw up hot-reloading in dev mode right?

grosser:
ops, need to revert ... added while testing :D

grosser:
fixed

henders:
Cool, 👍

PR #1932, opened 2017-04-21
allow search for if code was deployed or not

grosser:
screen shot 2017-04-20 at 5 00 11 pm

also makes json search for true/false work

travis is random error


henders:
Cool.... LGTM 👍  on 🍏

PR #1933, opened 2017-04-21
avoid hanging tabs on error

grosser:
regular error pages are good enough and better_errors looks fancy,
but takes forever to render and makes tabs crash regularly


henders:
😿 
👍

PR #1936, opened 2017-04-22
cache templates to avoid disk IO

grosser:
tested with `sudo chmod -R -r app/views` 

made local rendering (with caching/eager load etc enabled) go from 150ms -> 80s and not touch a single file, that should be a good speed bump for prod when it's under high IO load

### DIFF
--- a/config/initializers/cached_templates.rb
+++ b/config/initializers/cached_templates.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+# Samson often gets slow because deploys are IO heavy
+# avoid loading 10+ templates per page render from disk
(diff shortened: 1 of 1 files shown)


jonmoter:
I assume you mean 150ms -> 80ms, and not 80s?

grosser:
yep :D

grosser:
Rolled this out April 24th and it got rid of a lot of the pikes, so definitely helped:

screen shot 2017-05-02 at 8 39 24 am

mikhailov:
@grosser that's amazing!!! is it worth to consider reusing the same technique in Classic and other Rails apps?

grosser:
No, only if you app is crazily disk-io bound, otherwise the OS does a good
enough job at caching file reads in memory.

PR #1947, opened 2017-04-27
Use batch/v1 instead of extensions/v1beta1 for jobs

jbarreneche:
Using `jobs` is failing in kubernetes 1.6 because the `extensions/v1beta1` support for jobs has been removed in favor of the `batch/v1`. This causes launching jobs to fail silently in samson when the target cluster is kubernetes 1.6. This happens because the `extension_client` (initialized with `extensions/v1beta1`) doesn't have the method `get_job` because it isn't supported for this api anymore.
Luckily the `batch/v1` was already supported in kubernetes 1.2, so, this change should be backward compatible.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/cluster.rb
+++ b/plugins/kubernetes/app/models/kubernetes/cluster.rb
@@ -24,6 +24,10 @@ module Kubernetes
       @extension_client ||= build_client 'extensions/v1beta1'
     end
 
+    def batch_client
+      @batch_client ||= build_client 'batch/v1'
+    end
+
     def context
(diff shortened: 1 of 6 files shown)


grosser:
we are not currently using jobs and are only planing to use them from cron once we reached 1.6 ... so if it works for you that should be good enough :)

just need to fix the tests ...

jbarreneche:
@grosser Yeah, sorry about that! Was going to run them before making the PR but had a couple of meetings 🙃 .
All fixed now!

PR #1958, opened 2017-05-02
rails 5.1

grosser:
@zendesk/samson 
@zendesk/zendesk-rails-upgraders


bquorning:
There’s a substantial risk in upgrading ~100 gems at the same time. I’d prefer to see this PR become as little as possible, and reduce the risk of Samson going down tomorrow and someone having to hunt around those 100 gems to see what could possibly cause the downtime.

Samson may not be a customer-facing application, but it’s almost as important.

grosser:
good point ... I'll make a separate PR that bumps as much as possible while leaving rails at 5.0 then we can see if that causes any issues

PR #1960, opened 2017-05-02
remove foreign keys

grosser:
we only used 1 so pretty inconsistent / surprising ... also breaks rails 5.1 migrations see
https://github.com/rails/rails/pull/26266#issuecomment-298678182


grosser:
@adam-hart added in df3ee462

grosser:
... super weird ... part of https://github.com/zendesk/samson/pull/1581 but the PR diff does not show it ...

PR #1962, opened 2017-05-03
User ::Rails in case Airbrake::Rails exists

jbarreneche:
Hi! We aren't using airbrake, but, because our Dockerfile is based on your image, the gem 'airbrake' gets installed and **loaded** by bundler. 
Loading the Airbrake gem defines `Airbrake::Rails`, which makes the patch in `config/initializers/airbrake.rb` that redefines `Airbrake.notify` fail, because `Airbrake::Rails` doesn't have the method `.env`.
This changes disambiguates the costant lookup to explicitly use `::Rails`.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -36,10 +36,10 @@ if defined?(Airbrake) && key = ENV['AIRBRAKE_API_KEY']
 else
   module Airbrake
     def self.notify(ex, *_args)
-      if Rails.env.test?
+      if ::Rails.env.test?
         raise ex # tests have to use Airbrake.expects(:notify) to not hide unintented errors
       else
-        Rails.logger.error "AIRBRAKE: #{ex.class} - #{ex.message} - #{ex.backtrace[0..5].join("\n")}"
+        ::Rails.logger.error "AIRBRAKE: #{ex.class} - #{ex.message} - #{ex.backtrace[0..5].join("\n")}"
         nil
       end
     end


grosser:
thx!

grosser:
are you using sth else for error reporting ?
airbrake has a free plan so does not hurt ... also makes prod errors show what exactly went wrong

jbarreneche:
Hi, yes, we use bugsnag for reporting exceptions (because we use it in all of our projects). We've integrated using a plugin which includes bugsnag and other specific tools that we use.

Thanks!

grosser:
would be nice to have a generic interface for error reporting to make this simpler ...
could just be a hook I guess ... I'll whip something up ...

grosser:
... turns out it's not straight forward :(
 - airbrake captures controller/rake errors that we'd otherwise could not report on
 - airbrake includes filtering logic for exceptions that we don't want to report to statsd

mind open-sourcing your samson-bugsnag plugin ?

jbarreneche:
The current plugin depends on an internal library that setups multiple defaults from our platform (like logging, metrics, and exception notification). In the case of bugsnag it only loads it and sets the API_KEY (then bugsnag hooks up to rails, rack, rake, etc.), so there isn't much to share currently.
We will probably add some similar configuration like the ones you have for airbrake. But currently we didn't have the need. If we do, we surely can do that configuration on separate plugin and share it 👍

grosser:
if you find a nice way to use more hooks for this, let me know :)
could even make airbrake it's own plugin to make it a template for other
error reporting tools ...

PR #1963, opened 2017-05-03
reset bundler ENV vars before restarting so new server boots with cor…

grosser:
…rect Gemfile

had fun errors like: https://zendesk.airbrake.io/projects/95346/groups/1942179845832648326?tab=notice-detail

`NameError: uninitialized constant PaperTrail::RecordTrail` which came from new gems not being available after restart ... finally tracked that down ...

proper fix will be here https://github.com/puma/puma/pull/1282

### DIFF
--- a/config/boot.rb
+++ b/config/boot.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
 require 'bundler/setup'
+raise "ENV clearing failed" if File.expand_path(ENV.fetch("BUNDLE_GEMFILE")) != File.expand_path("Gemfile")
 require_relative 'dotenv'
 require 'bootscale/setup' if ['development', 'test'].include?(ENV['RAILS_ENV'] || ENV['RACK_ENV'] || 'development')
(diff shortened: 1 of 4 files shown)


grosser:
more testing showed that this still did not work since ORIGINAL_ENV also still included the BUNDLE_GEMFILE ... switching to clean_env and adding a sanity check in boot.rb :)

grosser:
now on restart this causes

```
* Restarting...
/opt/rbenv/versions/2.3.3/lib/ruby/2.3.0/rubygems.rb:241:in `bin_path': can't find gem puma (>= 0.a) (Gem::GemNotFoundException)
	from /data/samson/releases/20170503-201257-1fae041/vendor/bundle/ruby/2.3.0/bin/puma:22:in `<main>'
```

... since the gem is no longer in the load path and then puma tries to restart with just `puma` ... so needs some ops change to make the runner have `--restart-command 'bundle exec --keep-file-descriptors puma'` ... 

it will still work until then, just have the semi-bad behavior of crashing after each restart ... 
... I think we should just get rid of this thing altogether and simply rely on our service monitor to do the restarting ...

PR #1964, opened 2017-05-03
simplify restarting by discarding the ENV completely

grosser:
- reusing the ENV is a terrible idea
 - cleaning the ENV is hard
 - restarting an app with the exact same args is hard
... let's get out of that business ...

@jonmoter 
/cc @zendesk/samson 

for more details see:
 - https://github.com/zendesk/samson/pull/1963
 - puma restart handling code
 - bundler ENV cleaning code
 - the many ways this can go wrong and is hard to reason about / test

### DIFF
--- a/app/models/restart_signal_handler.rb
+++ b/app/models/restart_signal_handler.rb
@@ -3,7 +3,7 @@
 # Self-pipe is also best practice, since signal handlers can themselves be interrupted
 class RestartSignalHandler
   LISTEN_SIGNAL = 'SIGUSR1'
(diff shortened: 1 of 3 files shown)


jonmoter:
So if you send a `SIGTERM` to puma instead of a `SIGUSR2`, what effect does that have?  Will we have some downtime? Any other side effects?

grosser:
downtime while server reboots ... which means will drop  requests ... ah damn ... I though that's what it did anyway ... but that's the reason we have all this restart magic ... I'll use the other PR then ... but need to first prepare the runner with `--restart-args` :/

PR #1967, opened 2017-05-04
rails 5.1 --conservative

grosser:
same as https://github.com/zendesk/samson/pull/1958 but with less gems changed ...
dev works ... test-deployed too


grosser:
@zendesk/zendesk-rails-upgraders

PR #1968, opened 2017-05-04
allow multiline commands for better readability

grosser:
screen shot 2017-05-03 at 10 08 59 pm

### DIFF
--- a/test/models/job_test.rb
+++ b/test/models/job_test.rb
@@ -76,6 +76,11 @@ describe Job do
       job.command = "a\rb\r\nc\nd"
       job.commands.must_equal ["a", "b", "c", "d"]
     end
+
+    it "does not split multiline commands" do
+      job.command = "a\\\nb\nc\\\r\nd\re\\f"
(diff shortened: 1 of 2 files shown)


jonmoter:
So in the Command, there can be multi-line commands (ending with `\`), but they'll be joined together when executed?  Seems reasonable.

test/models/job_test.rb
```diff
@@ -76,6 +76,11 @@
       job.command = "a\rb\r\nc\nd"
       job.commands.must_equal ["a", "b", "c", "d"]
     end
+
+    it "does not split multiline commands" do
+      job.command = "a\\\nb\nc\\\r\nd\re\\f"
```
jonmoter: Would be easier to read this if you have something like:

```
job.command = <<-EO_CMD
a\
b
c\
d
e\f
EO_CMD
```

Though it does make it several lines longer.  Hmm.

grosser:
yep, mostly to avoid having endless lines for long commands

PR #1970, opened 2017-05-05
Fix booting in development

sandlerr:
Followup to https://github.com/zendesk/samson/pull/1933 cc @grosser @henders


grosser:
thx for following up, I totally forgot about that :D

PR #1973, opened 2017-05-05
redirect to where the user came from after login

grosser:
origin param made no sense ... was also unsafe since anyone could enter anything

```
http://localhost:3000/projects/consul2dogstats/stages/staging ->
http://localhost:3000/login?redirect_to=%2Fprojects%2Fconsul2dogstats%2Fstages%2Fstaging ->
http://localhost:3000/projects/consul2dogstats/stages/staging
```

### DIFF
--- a/app/controllers/unauthorized_controller.rb
+++ b/app/controllers/unauthorized_controller.rb
@@ -25,7 +25,8 @@ class UnauthorizedController < ActionController::Metal
       end
       format.html do
         flash[:authorization_error] = message
-        redirect_to login_path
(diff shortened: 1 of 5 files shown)


app/controllers/unauthorized_controller.rb
```diff
@@ -25,7 +25,8 @@ def respond
       end
       format.html do
         flash[:authorization_error] = message
-        redirect_to login_path
+        attempted_path = "/#{url_for(params).split("/", 4).last}"
```
grosser: cannot uue request.fullpath here since that is `/unauthenticated`

PR #1974, opened 2017-05-09
re-dump schema with rails 5.1 to avoid future diffs

grosser:
this is pretty ugly, but it's what we have in prod ... and that should be our dev starting point ...

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,527 +12,529 @@
 
 ActiveRecord::Schema.define(version: 20170505201145) do
 
-  create_table "builds", force: :cascade do |t|
-    t.integer  "project_id",                                       null: false
-    t.integer  "number"
-    t.string   "git_sha",                                          null: false
-    t.string   "git_ref",                                          null: false
-    t.string   "docker_image_id"
-    t.string   "docker_tag"
-    t.string   "docker_repo_digest"
-    t.integer  "docker_build_job_id"
-    t.string   "label"
-    t.string   "description",         limit: 1024
-    t.integer  "created_by"
+  create_table "builds", id: :integer, force: :cascade do |t|
+    t.integer "project_id", null: false
+    t.integer "number"
+    t.string "git_sha", null: false
+    t.string "git_ref", null: false
+    t.string "docker_image_id"
+    t.string "docker_tag"
+    t.string "docker_repo_digest"
+    t.integer "docker_build_job_id"
+    t.string "label"
+    t.string "description", limit: 1024
+    t.integer "created_by"
     t.datetime "created_at"
     t.datetime "updated_at"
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -12,527 +12,529 @@
...
+    t.string "source_url"
+    t.index ["created_by"], name: "index_builds_on_created_by"
+    t.index ["git_sha"], name: "index_builds_on_git_sha", unique: true
+    t.index ["project_id"], name: "index_builds_on_project_id"
   end
 
-  create_table "commands", force: :cascade do |t|
-    t.text     "command",    limit: 10485760
+  create_table "commands", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
+    t.text "command", limit: 16777215
```
sandlerr: does that match your dev database or our prod one?

sandlerr:
`DEFAULT CHARSET=utf8` seems wrong, don't we want `utf8mb4` everywhere in mysql? is there a way to get this working in postgresql and sqlite too?

db/schema.rb
grosser: prod

grosser:
I'll see what I need to change to get this loadable on postgres/sqlite ...

db/schema.rb
```diff
@@ -12,527 +12,529 @@
...
     t.datetime "deleted_at"
-    t.string   "source",     null: false
-    t.index ["project_id", "branch"], name: "index_webhooks_on_project_id_and_branch", length: { branch: 191 }, using: :btree
-    t.index ["stage_id", "branch"], name: "index_webhooks_on_stage_id_and_branch", length: { branch: 191 }, using: :btree
+    t.string "source", null: false
+    t.index ["project_id", "branch"], name: "index_webhooks_on_project_id_and_branch"
+    t.index ["stage_id", "branch"], name: "index_webhooks_on_stage_id_and_branch"
   end
 
+  add_foreign_key "deploy_groups", "environments"
```
sandlerr: any idea why these foreign keys were missing? hope we don't have any constraint violations in our database

grosser: idk ... will add a new migration to drop them ... already tried a week ago or so and that did not work out :/

sandlerr:
shame about `id: :integer` noise, would have thought that is the default

grosser:
they changed it to bigint in 5.1 ... so all the old ones are integer ... we
could migrate em ... but don't see a reason why ... we won't hit 32bit
anytime ...

grosser:
manually updated `output` and `object` column to what was specified in the migrations ... prod mysql multiplies them by 4 because we use multibyte ...

grosser:
added Rakefile fixes so load/dump will not produce diff anymore

PR #1977, opened 2017-05-12
builds and deploys compete for the same build

grosser:
the builds do the actual tagging, so we make the deploys wait if builds are set up
otherwise we might end up with builds not getting tagged randomly

should fix these:
`JobExecution failed: Couldn't find Job with 'id'=304440`
https://samson.zende.sk/projects/zendesk_inbox/deploys/288147#L16

and the kubernetes builds just sitting with seemingly no output

@staugaard


staugaard:
Thanks @grosser

PR #1979, opened 2017-05-13
fix local cache bug

grosser:
https://github.com/rails/rails/issues/29067


jonmoter:
How does this manifest in actual Samson bugs?

grosser:
https://zendesk.airbrake.io/projects/95346/groups/1948925214310476079/notices/1948925213736631219?tab=comments

PR #1986, opened 2017-05-17
no primary elements can be a valid deploy ... and load roles from cur…

grosser:
…rent branch

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role.rb
@@ -91,7 +91,7 @@ module Kubernetes
     end
 
     def defaults
-      return unless resource = role_config_file&.primary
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -121,8 +121,8 @@ def defaults
       }
     end
 
-    def role_config_file
-      return unless raw_template = project.repository.file_content(config_file, 'HEAD', pull: false)
+    def role_config_file(reference)
```
craig-day: default the argument maybe? `reference = 'HEAD'`

grosser: nope, only called in 2 locations ... makes it more obvious that it should not be randomly picked

PR #1993, opened 2017-05-23
do not auto-deploy to Automated deploys stages on new releases

grosser:


staugaard:
Do we need a backfill to fix the existing stages?

grosser:
doing a manual backfill

PR #1997, opened 2017-05-24
PAAS-992 allow pasting .env files for quick filling env vars

grosser:
screen shot 2017-05-24 at 12 57 58 pm
screen shot 2017-05-24 at 12 58 02 pm
screen shot 2017-05-24 at 12 58 35 pm

@jurisgalang 
/cc @craig-day @jonmoter @irwaters

### DIFF
--- a/plugins/env/app/assets/javascripts/env/application.js
+++ b/plugins/env/app/assets/javascripts/env/application.js
@@ -1,21 +1,84 @@
-$(document).on("click", ".duplicate_previous_row", function(e){
-  e.preventDefault();
+(function(){
+  // copied from https://github.com/motdotla/dotenv/blob/master/lib/main.js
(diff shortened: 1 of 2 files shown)


jonmoter:
It's not entirely obvious what the behavior is from the "Paste here" prompt.  Like:

* What format are you expecting to paste as?  I assume `NAME=VALUE`, one on each line?
* Will it immediately submit the values, or just populate fields on the page?
* What scope will be attached to the values you specify?
* Will it replace existing values?

I'm sure a user could figure it out with some trial and error, but I'd be nervous to use that UI if I didn't know what would happen.

grosser:
updated to "Paste .env formatted variables here. Fills the form but does not submit. Uses last selected scope"

jurisgalang:
:+1:

PR #1998, opened 2017-05-24
fix deploys for deleted stages messing up the current deployed

grosser:
https://zendesk.airbrake.io/projects/95346/groups/1957551004975572643/notices/1957550994995428628?tab=notice-detail


app/views/deploys/_currently_deployed.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if !ActiveRecord::Base.connection.class.name.downcase.include?('postgres') && deployed = @project.deploys.group(:stage_id).successful.sort_by { |d| d.stage.order }.presence %>
+<% if !ActiveRecord::Base.connection.class.name.downcase.include?('postgres') && deployed = @project.deploys.group(:stage_id).successful.select(&:stage).sort_by { |d| d.stage.order }.presence %>
```
jonmoter: Don't need a `compact` before the `sort_by`?

grosser:
no deploy is nil, this just removes deploys without stages

PR #1999, opened 2017-05-24
condense table so we can display long env vars without breaking

grosser:
before:
screen shot 2017-05-24 at 4 02 10 pm

after:
screen shot 2017-05-24 at 4 03 13 pm


jurisgalang:
That works :-) 👍

PR #2000, opened 2017-05-24
PAAS-978 show cluster capacity

grosser:
screen shot 2017-05-24 at 6 01 58 pm

@jonmoter @irwaters @JLandersZen

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/cluster.rb
+++ b/plugins/kubernetes/app/models/kubernetes/cluster.rb
@@ -46,6 +46,12 @@ module Kubernetes
       @kubeconfig ||= Kubeclient::Config.read(config_filepath)
     end
 
+    def schedulable_nodes
+      client.get_nodes.reject { |n| n.dig(:spec, :unschedulable) }
+    rescue
+      []
+    end
+
     private
 
     def connection_valid?
--- a/plugins/kubernetes/app/views/admin/kubernetes/clusters/_capacity.html.erb
+++ b/plugins/kubernetes/app/views/admin/kubernetes/clusters/_capacity.html.erb
@@ -1,0 +1,31 @@
+<%
+  @deploy_group_usage ||= Kubernetes::DeployGroupRole.usage # avoiding N+1 for every cluster
+  nodes = cluster.schedulable_nodes
+  available_cpu = nodes.sum { |n| n.dig(:status, :capacity, :cpu).to_f }
+  available_memory = nodes.sum { |n| n.dig(:status, :capacity, :memory).to_i / 1024 / 1024 }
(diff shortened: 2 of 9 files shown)


grosser:
this shows we only use 15% of our cluster ... seems suspiciously low ...

jonmoter:
You should probably filter out the API hosts (the ones with `node-type: node`, or where scheduling is disabled), since those nodes aren't available to deploy on.

plugins/kubernetes/app/views/admin/kubernetes/clusters/_capacity.html.erb
```diff
@@ -0,0 +1,33 @@
...
+<%
+  @used_cpu ||= Kubernetes::DeployGroupRole.group(:deploy_group_id).sum("requests_cpu * replicas")
+  @used_memory ||= Kubernetes::DeployGroupRole.group(:deploy_group_id).sum("requests_memory * replicas")
+  nodes = cluster.client.get_nodes rescue []
+  available_cpu = nodes.sum { |n| n.dig(:status, :capacity, :cpu).to_i }
+  available_memory = nodes.sum { |n| n.dig(:status, :capacity, :memory).to_i / 1024 / 1024 }
+  used_cpu = cluster.deploy_groups.sum { |dg| @used_cpu[dg.id] || 0 }
+  used_memory = cluster.deploy_groups.sum { |dg| @used_memory[dg.id] || 0 } / 1024
+  used_cpu_percent = (available_cpu.zero? ? 0 : (used_cpu * 100.0 / available_cpu).round)
+  used_memory_percent = (available_memory.zero? ? 0 : (used_memory * 100.0 / available_memory).round)
```
jonmoter: Seems like enough logic here that it would be best putting it into model code, rather than a view.

grosser:
extracted the biggest chunk of logic into the models ... rest is still ugly, but extracting that would mean having to keep big hashes in sync ... and the logic is pretty straight forward ... looks ok ?

plugins/kubernetes/app/models/kubernetes/cluster.rb
```diff
@@ -46,6 +46,14 @@ def kubeconfig
       @kubeconfig ||= Kubeclient::Config.read(config_filepath)
     end
 
+    def schedulable_nodes
+      client.get_nodes.
+        reject { |n| n.dig(:spec, :unschedulable) }.
+        select { |n| n.dig(:metadata, :labels, :"node-type") == "node" }
```
jonmoter: We probably only need the first one, since the API hosts are always unscheduable, and `node-type: node` is something Zendesk-specific.

PR #2008, opened 2017-05-31
limit replicas too

grosser:


grosser:
nvm ... this is silly ... need to just enforce that `selected values * replicas` is less than the limit ...

PR #2011, opened 2017-05-31
reduce max size of current deploys to not block view for projects wit…

grosser:
…h lots of stages

(screenshot from 100px ... so 3x that space)
screen shot 2017-05-31 at 3 46 37 pm

@jacobat


jacobat:
Thanks :+1:

PR #2020, opened 2017-06-05
make current deploy labels clickable

grosser:
screen shot 2017-06-05 at 9 31 44 am

... the 3rd PR for that thng ... hope it finally works :D

### DIFF
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -67,13 +67,11 @@ $(function () {
     }
   }
 
-  // When user clicks a release label, fill the deploy reference field with that version
+  // When user clicks a release or deploy label, fill the deploy reference field with that version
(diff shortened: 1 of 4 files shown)


ibrahima:
Hi guys, I dunno if this is happening on your instances, but for me, it looks like this javascript is no longer triggered, and the release labels are still acting like links, and the current deploy tags just do nothing when clicked. Interestingly if I paste in the event handler into my browser console it does work, which is leads me to suspect that the javascript just isn't running at the right time. I am on the latest release too.

grosser:
hmm all working for me ... check js console for errors ... otherwise sprinkle some alerts in the js to see where it stops working

ibrahima:
Oh duh, I'm an idiot. Forgot to run assets:precompile after upgrading last time, everything works now!

PR #2029, opened 2017-06-06
Risks no longer required to be last section

ibrahima:
* The risks section of the PR can now be followed by other sections
* Added unit tests for this case

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -126,7 +126,7 @@ class Changeset::PullRequest
   private
 
   def parse_risks(body)
-    body.to_s.split(RISKS_SECTION, 2)[1].to_s.strip.split(SECTION_HEADING).first.presence
+    body.to_s.split(RISKS_SECTION, 2)[1].to_s.strip.split(SECTION_HEADING).first.to_s.strip.presence
   end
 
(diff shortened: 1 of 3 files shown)


ibrahima:
As a side note, I noticed the use of Rails' strip_heredoc in some of the specs. As of Ruby 2.3 you can now use squiggly heredocs to avoid needing that: https://infinum.co/the-capsized-eight/multiline-strings-ruby-2-3-0-the-squiggly-heredoc

grosser:
ah yes plz ... want to make another PR ? :)

grosser:
ideally the risk regexp would be smart enough to stop at headers ... but that might be tricky ... good enough and we have the tests if we ever want to refactor ...

just merge ahead on green ...

ibrahima:
Cool cool. Yeah, I thought about making that regexp do more, but I think I'd personally prefer more readable code than a concise but hard to read regex.

ibrahima:
Oh hold on a sec, I'll update the docs too.

ibrahima:
Should be good to go now!

grosser:
🎉

PR #2031, opened 2017-06-06
Replace strip_heredoc with squiggly

ibrahima:
* Just replaces usages of Rails' strip_heredoc with Ruby 2.3's squiggly heredoc: https://infinum.co/the-capsized-eight/multiline-strings-ruby-2-3-0-the-squiggly-heredoc

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/controllers/oauth_test_controller.rb
+++ b/app/controllers/oauth_test_controller.rb
@@ -19,7 +19,7 @@ class OauthTestController < ActionController::Base
     # alternatively we could call whatever the oauth controller does internally directly
     access_token = oauth_client.auth_code.get_token(params[:code], redirect_uri: token_url).token
 
-    render plain: <<-TEXT.strip_heredoc
+    render plain: <<~TEXT
       Your access token is: #{access_token}
 
       You can use this to make requests:
@@ -47,7 +47,7 @@ class OauthTestController < ActionController::Base
 
   def ensure_application
     return if application
-    message = <<-TEXT.strip_heredoc
+    message = <<~TEXT
       Add an OAuth application at
 
       #{new_oauth_application_url}
--- a/app/mailers/project_mailer.rb
+++ b/app/mailers/project_mailer.rb
@@ -12,7 +12,7 @@ class ProjectMailer < ApplicationMailer
 
   def build_mail(user, project, address, action)
     subject = "Samson Project #{action.titleize}: #{project.name}"
-    body = <<-TEXT.strip_heredoc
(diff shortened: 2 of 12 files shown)


app/models/changeset/pull_request.rb
```diff
@@ -9,6 +9,9 @@ class Changeset::PullRequest
   # Matches a markdown section heading named "Risks".
   RISKS_SECTION = /^\s*#*\s*Risks?\s*#*\s*\n(?:\s*[-=]*\s*\n)?/i
 
+  # Matches a markdown section heading
+  SECTION_HEADING = /^\s*#*\s*\w+.*\n(?:\s*[-=]*\s*\n)?/i
```
grosser: this does not belong here

grosser:
rebase ... otherwise looks good

ibrahima:
Oh, I just based it on my other PR since that one adds heredocs, but I can just change the ones in that PR there and then have this one for the existing strings. Rebased and removed those commits!

ibrahima:
Not sure if that was meant for me, but I don't have write access, I'm not from Zendesk - I've just submitted a few PRs before. But thanks for accepting my changes!

grosser:
oh, it said `Contributor` ... but we are hiring anyways ... so you can always change that ;)

ibrahima:
Looks like Brakeman doesn't understand this change somehow? https://github.com/zendesk/samson/pull/2031/files#diff-b17d53435fe1389ba1c27dd996b4c93e

Does it need upgrading to support Ruby 2.3?

grosser:
needs https://github.com/seattlerb/ruby_parser/issues/218 ... so I guess this will sit here for a while :(

... could do some gsub hacks in brakeman ... but feels dirty ... don't need it that bad ...

ibrahima:
I could also just revert that one change, and you'd have the new syntax elsewhere? Or will Brakeman break on other files too? I guess it's not specific to that file, but some combination of syntaxes that causes the parser error.

It definitely handles some form of squiggly heredocs because our codebase has them and we also have Brakeman enabled.

grosser:
I'd thin kit ignores them because the syntax is still somehow valid ... but
most likely broken ... I'll revert the ones that got merged manually so we
don't skip any brakeman warnings by accident

grosser:
ah these were in tests ... so that's ok since brakeman does not check them

PR #2036, opened 2017-06-07
paste fixes

grosser:
@jurisgalang 

screen shot 2017-06-06 at 8 21 17 pm


jurisgalang:
:+1:

Didn't know you could insert commands for values (eg: `{bundle exec rake secret}`) that's nice.

grosser:
... this is just a placeholder so users know where to get it :)

PR #2038, opened 2017-06-08
cut down boottime more ... 6s -> 4.5s

grosser:
@irwaters 

/cc @zendesk/zendesk-rails-upgraders


grosser:
there is something weird going on with coverage detection and bootsnap ...
even basic Bootsnap.setup make the tests fail in `rake test` ... but not anywhere else ... 

```
plugins/airbrake/lib/samson_airbrake/samson_plugin.rb was expected to be covered, but was already loaded before tests started, which makes it uncoverable.
```

so must be loading/activating some settings ... will dig more tomorrow ...

config/boot.rb
```diff
@@ -2,4 +2,16 @@
...
+rails_env = ENV['RAILS_ENV'] || ENV['RACK_ENV'] || 'development'
+if ['development', 'test'].include?(rails_env)
+  require 'bootsnap'
+  Bootsnap.setup(
+    cache_dir:            'tmp/bootsnap', # Path to your cache
+    development_mode:     rails_env == "development",
+    load_path_cache:      true, # Should we optimize the LOAD_PATH with a cache?
+    autoload_paths_cache: true, # Should we optimize ActiveSupport autoloads with cache?
+    disable_trace:        false, # Sets `RubyVM::InstructionSequence.compile_option = { trace_instruction: false }`
+    compile_cache_iseq:   true, # Should compile Ruby code into ISeq cache?
```
sandlerr: disabling this in the test environment might fix coverage problems

grosser:
still breaks running rake in dev so I'd like to ideally fix this and not
hack around ...

config/boot.rb
grosser: derp ... that was it ... tried commenting out and was still failing ... the default is true :D

PR #2039, opened 2017-06-08
Fix broken outbound_webhooks migration

hyperair:
When `default: nil` is set on `deleted_at`, Rails attempts to create a `timestamp
DEFAULT NULL` field, which causes an error when run against a MySQL database.

This commit removes the default value so that the field is generated as
`timestamp NULL DEFAULT NULL`.

### Tasks
 - [ ] :+1: from team

### References
 - JIRA link: https://zendesk.atlassian.net/browse/VULCAN-374

### Risks
- Level: Low


grosser:
I guess that was valid in some older days ... definitely ran this on mysql before :D

PR #2043, opened 2017-06-08
refactor secret forms

grosser:
use form_for so we can input and dry up all the html structure
also made the controller safer by not even including the secret value

@swatikri


swatikri:
👍

PR #2044, opened 2017-06-08
request global secrets to be shared instead of sharing by default

grosser:
- User sets new `SECRET_STORAGE_SHARING_GRANTS=true`
 - Global secrets are no longer auto-resolved
 - On secret page a link appears
screen shot 2017-06-08 at 3 30 10 pm
 - admin following the link creates a new grant (non-admin see a instruction)
screen shot 2017-06-08 at 3 30 16 pm
 - ... and redirects back
screen shot 2017-06-08 at 3 46 06 pm
 - comes with overview page
screen shot 2017-06-08 at 3 46 32 pm
 - history
screen shot 2017-06-08 at 3 46 18 pm
 - linked on secrets index page
screen shot 2017-06-08 at 3 46 25 pm

@swatikri @jonmoter @irwaters

### DIFF
--- a/app/controllers/secret_sharing_grants_controller.rb
+++ b/app/controllers/secret_sharing_grants_controller.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+class SecretSharingGrantsController < ApplicationController
+  before_action :authorize_admin!, except: [:index, :show]
+  before_action :find_grant, only: [:show, :destroy]
+
+  def index
+    @secret_sharing_grants = SecretSharingGrant.page(page).per(25)
+  end
+
+  def new
+    @secret_sharing_grant = SecretSharingGrant.new(key: params.dig(:secret_sharing_grant, :key))
+  end
+
+  def create
+    attributes = params.require(:secret_sharing_grant).permit(:key, :project_id)
+    @secret_sharing_grant = SecretSharingGrant.create(attributes)
+    if @secret_sharing_grant.persisted?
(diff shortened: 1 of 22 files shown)


app/controllers/secret_sharing_grants_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+class SecretSharingGrantsController < ApplicationController
+  before_action :authorize_admin!, except: [:index, :show]
+  before_action :find_grant, only: [:show, :destroy]
+
+  def index
+    @secret_sharing_grants = SecretSharingGrant.all
```
irwaters: is this gonna require pagination?

grosser: yeah ... most likely soon

app/models/secret_sharing_grant.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class SecretSharingGrant < ActiveRecord::Base
+  has_paper_trail skip: [:created_at]
+  belongs_to :project
+  validates :key, uniqueness: true
```
davidreuss: I'm no rails/active-record expert, but this seems like possible cause of tripping up on uniqueness constraints, causing projects not being able to add secrets if one is already added for a key (no matter project) ?

grosser:
yes, fixed in https://github.com/zendesk/samson/pull/2060 if you would +1
that :)

PR #2052, opened 2017-06-09
move evil delete to the right to avoid fat-finger issues

grosser:
needs more fixes, but as general idea we could have it on the right ... otherwise turn it 

screen shot 2017-06-09 at 3 01 30 pm

@irwaters @mikehuston


mikehuston:
👍

libo:
👍

PR #2055, opened 2017-06-10
record request params too in case the payload is not in the body

grosser:
also fixing json payload to correctly finding it's action since it is not nested in github as
params are due to rails wrap_parameters

@flochaz 

so rails wraps json post params into a controller prefix, in this case `github` but when payload is sent in the wrapping step does not take place, so all the code that tried to access `github` in pull_request never worked on json payload or when fetching data directly via the api

### DIFF
--- a/app/controllers/integrations/circleci_controller.rb
+++ b/app/controllers/integrations/circleci_controller.rb
@@ -3,7 +3,7 @@ class Integrations::CircleciController < Integrations::BaseController
   protected
 
   def payload
-    @payload ||= params.fetch('payload', '{}')
+    @payload ||= params.fetch('payload', {})
   end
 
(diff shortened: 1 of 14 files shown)


app/controllers/integrations/circleci_controller.rb
```diff
@@ -3,7 +3,7 @@ class Integrations::CircleciController < Integrations::BaseController
   protected
 
   def payload
-    @payload ||= params.fetch('payload', '{}')
+    @payload ||= params.fetch('payload', {})
```
grosser: payload is used as a hash in the controller, so should not be a string

grosser:
... and don't log params that should be filtered

PR #2062, opened 2017-06-13
Fix uniqueness to be scoped on project

davidreuss:
This key is already added to the database, but the app needs to know
that this uniqueness is scoped as well.

I'm no AR expert, but from gleaning docs this is what i came up
with.

/cc @zendesk/samson


grosser:
fixed in https://github.com/zendesk/samson/pull/2060

davidreuss:
Duh.

PR #2066, opened 2017-06-13
reverse search secrets by value

grosser:
- super slow
 - can speed up by pre-filtering keys
 - hiding query value in the view by treating it as password field
 - `value` is not logged
 - only unsafe component is the query URL possibly getting stored in nginx logs or browser history ... could encrypt on the browser or so ... maybe that's overkill ...
 - opened a vault topic to see if they might add this as a feature, but doubt they'll respond https://groups.google.com/forum/#!msg/vault-tool/Jo6nw7X5Lhc/OS1K_HB2AgAJ

reworked search to has specific fields for all:
screen shot 2017-06-13 at 1 50 35 pm

then added value:
screen shot 2017-06-13 at 3 46 09 pm

@zendesk/paas


grosser:
to make this faster:
 - parallel read_multi https://github.com/zendesk/samson/pull/2071
 - cached group/env lookups https://github.com/zendesk/samson/pull/2070

PR #2068, opened 2017-06-14
avoid all caching round-trips and allow caching arbitrary data per model

grosser:
will use this to cache commen lookups like all deploy-group/env permalinks
and vault_server_ids to avoid memcache and db requests for simple data
without having to add cache-expire callbacks to all places that want to
cache that data


grosser:
did not solve the issue and still has race-conditions ...

PR #2072, opened 2017-06-14
PAAS-860 revert vault writes if they fail to stay in a consistent state

grosser:
- when secrets get partially created we end up with undeletedable or unfindable secrets (since we read from the first server)
 - when secrets get partially updated we might live in a split-brain world
... let's avoid that ...


grosser:
it could fail, but hopefully it fails on the same servers the original request failed on making it a noop

PR #2074, opened 2017-06-15
do not take down the server in staging when an exception happens in a…

grosser:
… thread

also make test/dev fail softly and at an expected place

had staging restart on me because a secret could not be read :/


grosser:
fixed 2 more threading bugs ... that should make the tests less flaky :D

PR #2076, opened 2017-06-16
paper_trail -> audited

grosser:
reduces complexity a lot and cuts down on the amount of data we store ... from everything to only changed attributes

will leave old tables around for a year or so in case anything goes wrong ...

@jonmoter @irwaters 

TODO:
 - [x] test all audited classes can be linked to
 - [X] migrate versions
 - [x] check if having auditing always-on makes tests a lot slower or not ...
 - [x] rake / rails c auditing

before:
screen shot 2017-06-16 at 2 37 07 pm

after:
screen shot 2017-06-16 at 3 08 05 pm


grosser:
test speed slowdown seems minimal:

```
benchmark 5 ruby test/models/stage_test.rb

runs: 5
total: 89.06
average: 17.81
range: 16.55..20.53

vs

runs: 5
total: 88.3
average: 17.66
range: 15.68..20.02
```

grosser:
tried it: takes 1s/1000versions and 1% of audits is removed because they are empty updates

grosser:
fixed another bug: we were creating audits on `last_seen_at` for users

grosser:
tested in staging, dry-ran in prod ... looks good :)

grosser:
walked it through with @cvgw too and fixed a few more bugs

PR #2085, opened 2017-06-19
re-raise errors from forks so users can see them

grosser:
especially `Docker build failed: Cannot locate specified Dockerfile: Dockerfile`
which happens because the tarfile was not written

@benhass @jonmoter @irwaters @rgueldem


benhass:
👍 on 📗

PR #2086, opened 2017-06-19
show limit id for easier changes / followup

grosser:
@craig-day 

screen shot 2017-06-19 at 3 09 42 pm


craig-day:
Idk if the id number is worth it, I like the rewording to **kubernetes usage limit** though

grosser:
I like the number in case it's unclear which limit applied ... direct link
would be nice, but that is trickier

PR #2087, opened 2017-06-20
make the first user be an admin, even if seeded

yizhang-zen:
@irwaters @grosser

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -74,12 +74,12 @@ class User < ActiveRecord::Base
     user = User.where(external_id: hash[:external_id].to_s).first || User.new
 
     # attributes are always a string hash
-    attributes = user.attributes.merge(hash.stringify_keys) do |key, old, new|
-      if key == 'role_id'
-        if !User.exists? # first user will be the super admin
-          Role::SUPER_ADMIN.id
+    attributes = user.attributes.merge(hash.stringify_keys) do |attribute, old, new|
+      if attribute == 'role_id'
+        if !User.where.not(email: 'seed@example.com').exists?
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -93,6 +93,11 @@ def self.create_or_update_from_hash(hash)
     user
   end
 
+  # first user will be the super admin
+  private_class_method def self.initial_user?
+    !User.exists? || User.limit(2).pluck(:email) == ['seed@example.com']
```
grosser: does to db queries atm, but we can reduce that to 1

grosser:
rubocop failed, should be easy to fix

grosser:
the 2 failed tests look like random failures ...

irwaters:
looks like rubocop fix is required, so 👍 on green

app/models/user.rb
```diff
@@ -74,12 +74,12 @@ def self.create_or_update_from_hash(hash)
     user = User.where(external_id: hash[:external_id].to_s).first || User.new
 
     # attributes are always a string hash
-    attributes = user.attributes.merge(hash.stringify_keys) do |key, old, new|
-      if key == 'role_id'
-        if !User.exists? # first user will be the super admin
-          Role::SUPER_ADMIN.id
+    attributes = user.attributes.merge(hash.stringify_keys) do |attribute, old, new|
+      if attribute == 'role_id'
+        if !User.where.not(email: 'seed@example.com').exists?
```
grosser: FYI @yizhang-zen updated this to be even more efficient ... only a single query and not actually reading any data from the db

PR #2095, opened 2017-06-21
Fix appending app init-containers to samson injected init-containers

davidreuss:
The init container key is stored as a symbol on the docs, and was not
being picked up properly, and thus "ignored" as samson was overwriting
the existing key.

/cc @zendesk/samson


grosser:
also a generic test for `should not include string keys` so we don't run into the same issue again ..

grosser:
test-deployed and seems to work

PR #2104, opened 2017-06-22
introduce can checks

grosser:
... not super happy with how it works, a bit clumsy to have to sent the project over all the time, but it's not too bad ... thoughts ?


grosser:
refactored a bit more ... a bit nicer now ... also added permissions for projects ... will merge on green and then make another PR for more controllers, getting a bit big already ...

PR #2111, opened 2017-06-28
PAAS-1045 fix blocking dev restarts and infititly busy connections

grosser:
@irwaters 
/fyi @pschambacher 

This was fun ... kept a log for future debugging/wtf-is-this-code diggers ...

Starting with simple printer:

```Ruby
Thread.new do
  loop do
    sleep 1
    puts "AR #{ActiveRecord::Base.connection_pool.stat}"
  end
end
```

after the initial start 3 connections are in busy state ... that seems odd .. they get un-busy on reload

we can see during a reload (changed controller) that we end up waiting even though we only use 5/10 connections:
`AR {:size=>10, :connections=>5, :busy=>5, :dead=>0, :idle=>0, :waiting=>1, :checkout_timeout=>5}`

adding to active_record/connection_adapters/abstract/connection_pool.rb to get more info:

```
in_flight: @now_connecting,
blocking: @threads_blocking_new_connections
```

... now we can see that blocking causes the wait ... so onto finding out what triggers blocking ...

adding some caller debugging to find out ...

```Ruby
puts ">>>>>>>>#{__FILE__}:#{__LINE__}>>>>>>>>"
puts caller.first(5)
puts "<<<<<<<<#{__FILE__}:#{__LINE__-2}<<<<<<<<"
```

`clear_reloadable_connections` seems guilty ... always hits the timeout and then unlocks ...
so someone is checking a connection out and never in

putting a caller in `Connection#checkout` shows 2 places that get a connection during bootup ...
periodical and Job.running in application.rb ...
removing them bot removes the blocking issue and reduces busy connection to 0

then putting a check in so this never creeps in again ... needs to be after after_initialize so
refactoring the previous checks into class methods to call them all from our after-after_initialize

### TODO:
 - [ ] tests


irwaters:
agreed, super interesting and also thanks for the explanation! :)

PR #2112, opened 2017-06-29
SAMSON-358 do not fail to list keys if some vault servers are unresponsive

grosser:
previously if a single vault server was down/unresponsive .keys would fail
and everything would stop working, now we still send the error to Airbrake but
continue executing

https://zendesk.airbrake.io/projects/95346/groups/1958152296001382773


grosser:
yep, everything that is specific to a single group should fail/block ...
otherwise we get into weird scenarios ... this already is kinda shady since
it will look like some keys are missing ... but it beats making everything
fail :(

PR #2126, opened 2017-07-08
replace docker image building via gem with shellout

grosser:
also make all credentials available during build phase

@jonmoter @irwaters POC ... need some sanity check ...

I guess switching home is not an option ... so need to find something better for temp credentials ...
guess we cold just use `docker login` so the deploy user stays logged in ...

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'docker'
+require 'shellwords'
 
 class DockerBuilderService
   DIGEST_SHA_REGEX = /Digest:.*(sha256:[0-9a-f]+)/i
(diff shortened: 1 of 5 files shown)


app/models/docker_builder_service.rb
```diff
@@ -8,21 +9,19 @@ class DockerBuilderService
...
-
-      docker_image =
-        Docker::Image.build_from_tar(tarfile, docker_options, Docker.connection, credentials_to_pull) do |chunk|
-          output.write_docker_chunk(chunk)
-        end
-      output.puts('### Docker build complete')
+  def self.build_docker_image(dir, output)
+    TerminalExecutor.new(output).execute!(
+      "cd #{dir.shellescape}",
+      *local_docker_login,
```
jonmoter: Might need to be careful about echo'ing out registry credentials, which are secrets.

app/models/docker_builder_service.rb
```diff
@@ -8,21 +9,19 @@ class DockerBuilderService
...
+      "docker build ."
+    ) && output.to_s[/Successfully built (\S+)/, 1]
```
jonmoter: Previously, `build_docker_image` returned a docker image (or maybe the id of the docker image). So we'll need to add some logic to watch stdout and try to extract the id of the image.

app/models/docker_builder_service.rb
```diff
@@ -51,11 +50,15 @@ def run!(push: false, tag_as_latest: false)
...
-      elsif build_image(tmp_dir)
-        ret = true
-        ret = push_image(tag_as_latest: tag_as_latest) if push
-        build.docker_image.remove(force: true) unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
-        ret
+      else
+        if build_image(tmp_dir) # rubocop:disable Style/IfInsideElse
+          ret = true
+          ret = push_image(tag_as_latest: tag_as_latest) if push
+          build.docker_image.remove(force: true) unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
```
jonmoter: Are you thinking of continuing to use the docker gem to handle pushing the image and deleting the image?

grosser:
maybe covert the other parts later too ... for now just trying to get the
actual issue resolved :)

grosser:
thoughts on just using `docker login` to store credentials in home of
current user ?
... hard to do anything else since docker insists on that location :/

jonmoter:
According to the [docker CLI docs](https://docs.docker.com/engine/reference/commandline/cli/) you can specify the directory where the config file exists, either by saying `docker --config /some/dir` or setting the `DOCKER_CONFIG` env var.

Then when you run [docker login](https://docs.docker.com/engine/reference/commandline/login/), it appears that it saves the credentials in `$DOCKER_CONFIG/config.json`.  Though I haven't tried it to be sure.

So if an admin of Samson wants those to be stored in a directory other than the home directory of the user, then it appears that's possible by setting env vars. But we'd need to test to be sure.

grosser:
sweet, did not see DOCKER_CONFIG ... --config did work for pull but not
push :)

grosser:
worked :)
screen shot 2017-07-09 at 12 13 53 pm

also copying the existing config if it exists just in case ... will play with it on staging then rewrite most of the tests most likely ...

grosser:
thx, will take it out when I got a few consecutive hours of online time

PR #2127, opened 2017-07-11
small UI tweaks for instances with lots projects or env

craig-day:
This modifies two places, the home page which lists projects, and the environment variable groups page. It puts a limit on the max number of stages display on the projects page and max number of env vars or projects displayed in each env var group row.

#### Projects page with a show more link
screen shot 2017-07-10 at 7 18 33 pm

#### Env var groups page with show more links
screen shot 2017-07-10 at 7 11 14 pm

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low. Only updating the UI index pages.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -266,12 +266,10 @@ module ApplicationHelper
     result
   end
 
-  def list_with_show_more(items, display_limit, show_more_tag = nil, ul_options = {}, &block)
+  def list_with_show_more(items, display_limit, show_more_tag, ul_options = {}, &block)
     li_tags = items.first(display_limit).map { |i| capture(i, &block) }
     li_tags << show_more_tag if items.size > display_limit
 
-    content_tag(:ul, ul_options) do
-      li_tags.join.html_safe
-    end
+    content_tag(:ul, li_tags.join.html_safe, ul_options)
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -265,4 +265,13 @@ module ApplicationHelper
     result << " #{objects.total_count} records" if objects.total_pages > 1
     result
   end
+
+  def list_with_show_more(items, display_limit, show_more_tag = nil, ul_options = {}, &block)
+    li_tags = items.first(display_limit).map { |i| capture(i, &block) }
+    li_tags << show_more_tag if items.size > display_limit
+
+    content_tag(:ul, ul_options) do
+      li_tags.join.html_safe
+    end
+  end
 end
--- a/app/views/projects/_project.html.erb
+++ b/app/views/projects/_project.html.erb
@@ -1,4 +1,5 @@
 <%# Is expired on deploy because job touches project (job->deploy->stage->project) %>
+<% max_display_projects = 5 %>
 <% cache [project, Lock.cache_key] do %>
   <div class="col-sm-4 project-tile">
     <h2>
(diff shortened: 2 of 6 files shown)


app/views/projects/_project.html.erb
```diff
@@ -31,6 +31,11 @@
           </li>
         <% end %>
       <% end %>
+      <% if project.stages.count > 8 %>
```
grosser: would be good to extract the `magic number` 8

grosser: also this is doing 3 queries now ... select + count + count
cheapest would be to do `.first(max+1).size == max` to detect if to display a `More` link or not (without indicator)

plugins/env/app/views/environment_variable_groups/index.html.erb
```diff
@@ -19,16 +23,20 @@
     </tr>
     </thead>
     <% @groups.each do |group| %>
+      <% sorted_vars = group.environment_variables.map(&:name).uniq.sort %>
```
grosser: maybe do a `.sort_by(&:id)` for clarity (I know it's old code :D )

plugins/env/app/views/environment_variable_groups/index.html.erb
```diff
@@ -19,16 +23,20 @@
...
       <tr>
         <td>
           <%= link_to group.name, group %>
           <ul>
-            <% group.environment_variables.map(&:name).uniq.sort.each do |name| %>
+            <% sorted_vars.first(5).each do |name| %>
```
grosser: same magic number

grosser:
maybe:
```Ruby
<%= list_with_more items, 5, content_tag(:li, link_to("More", foo_path)) %>
```

craig-day:
@grosser helper written

app/helpers/application_helper.rb
```diff
@@ -265,4 +265,13 @@ def paginate(objects, *)
     result << " #{objects.total_count} records" if objects.total_pages > 1
     result
   end
+
+  def list_with_show_more(items, display_limit, show_more_tag = nil, ul_options = {}, &block)
+    li_tags = items.first(display_limit).map { |i| capture(i, &block) }
+    li_tags << show_more_tag if items.size > display_limit
```
grosser: that `= nil` in the arguments would be weird here ... do we need to allow it to be nil ?

app/helpers/application_helper.rb
```diff
@@ -265,4 +265,13 @@ def paginate(objects, *)
...
+
+    content_tag(:ul, ul_options) do
```
grosser: can use argument instead of block

app/helpers/application_helper.rb
```diff
@@ -265,4 +265,13 @@ def paginate(objects, *)
     result << " #{objects.total_count} records" if objects.total_pages > 1
     result
   end
+
+  def list_with_show_more(items, display_limit, show_more_tag = nil, ul_options = {}, &block)
+    li_tags = items.first(display_limit).map { |i| capture(i, &block) }
```
grosser: start with `"".html_safe` and use `<<` to avoid having to call `.html_safe` at then end on an array that might have included other things

grosser:
could just do `More` instead of running another `.count` query ...
looks alright though 👍

craig-day:
@grosser since we aren't doing any scoping fubar I just materialized the query so I can call `count` or `size` without going back to the db. Plan to merge this once its green unless you have any objections.

PR #2128, opened 2017-07-11
Rollbar plugin

hkjorgensen:
Send deploy notifications to [Rollbar](https://www.rollbar.com) after deploy. It's configurable to N number of webhooks if users have multiple Rollbar instances around the globe. 

Screenshot:
image

/cc @zendesk/samson @craig-day 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170622010111) do
+ActiveRecord::Schema.define(version: 20170711174532) do
 
   create_table "audits", force: :cascade do |t|
     t.integer "auditable_id", null: false
@@ -385,6 +385,16 @@ ActiveRecord::Schema.define(version: 20170622010111) do
     t.integer "build_id"
     t.index ["build_id"], name: "index_releases_on_build_id"
     t.index ["project_id", "number"], name: "index_releases_on_project_id_and_number", unique: true
+  end
+
+  create_table "rollbar_webhooks", force: :cascade do |t|
+    t.text "webhook_url", null: false
+    t.string "access_token", null: false
+    t.string "environment", null: false
+    t.integer "stage_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["stage_id"], name: "index_rollbar_webhooks_on_stage_id"
   end
 
   create_table "secret_sharing_grants", force: :cascade do |t|
--- a/plugins/rollbar/README.md
+++ b/plugins/rollbar/README.md
@@ -1,0 +1,8 @@
+# Rollbar Plugin
+
+This plugin adds deploy tracking to [Rollbar](https://rollbar.com/)
+
+1. The webhook URL is `https://api.rollbar.com/api/1/deploy/` unless it's a self-hosted instance.
+2. Sign in to Rollbar and go to the "Deploys" area.
+3. c/p the `access_token` to Samson.
+4. The `environment` name must match one from Rollbar "Settings > General > Environments".
--- a/plugins/rollbar/README.md
+++ b/plugins/rollbar/README.md
@@ -2,7 +2,7 @@
 
 This plugin adds deploy tracking to [Rollbar](https://rollbar.com/)
 
-1. The webhook URL is `https://api.rollbar.com/api/1/deploy/` unless it's a self-hosted instance.
+1. Add `https://api.rollbar.com/api/1/deploy/` to the the webhook URL, unless you're using a self-hosted instance of Rollbar.
 2. Sign in to Rollbar and go to the "Deploys" area.
 3. c/p the `access_token` to Samson.
 4. The `environment` name must match one from Rollbar "Settings > General > Environments".
--- a/plugins/rollbar/app/models/rollbar_notification.rb
+++ b/plugins/rollbar/app/models/rollbar_notification.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+require 'faraday'
+
+class RollbarNotification
+  def initialize(webhook_url:, access_token:, environment:, revision:)
+    @webhook_url = webhook_url
+    @access_token = access_token
+    @environment = environment
+    @revision = revision
+  end
+
+  def deliver
+    Rails.logger.info "Sending Rollbar notification..."
+    response = Faraday.post(
+      @webhook_url,
+      access_token: @access_token,
+      environment: @environment,
+      revision: @revision
+    )
+
+    if response.success?
+      Rails.logger.info "Sent Rollbar notification"
+    else
+      Rails.logger.info "Failed to send Rollbar notification. HTTP #{response.status}: #{response.body}"
+    end
+  end
+end
--- a/plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
+++ b/plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
@@ -8,15 +8,15 @@
       <%= form.fields_for :rollbar_webhooks do |rollbar_field| %>
         <div class="form-group">
           <div class="col-lg-4">
-            <%= rollbar_field.url_field :webhook_url, class: "form-control", placeholder: "Webhook URL" %>
+            <%= rollbar_field.url_field :webhook_url, class: "form-control", placeholder: "Webhook URL", required: true  %>
           </div>
 
           <div class="col-lg-3">
-            <%= rollbar_field.text_field :access_token, class: "form-control", placeholder: "Access token" %>
+            <%= rollbar_field.text_field :access_token, class: "form-control", placeholder: "Access token", required: true %>
           </div>
 
           <div class="col-lg-3">
-            <%= rollbar_field.text_field :environment, class: "form-control", placeholder: "Environment" %>
+            <%= rollbar_field.text_field :environment, class: "form-control", placeholder: "Environment", required: true %>
           </div>
 
           <%= delete_checkbox rollbar_field %>
--- a/plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
+++ b/plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
@@ -1,0 +1,27 @@
+<% if DeployGroup.enabled? %>
+  <fieldset>
+    <legend>Rollbar</legend>
+    <% stage = form.object %>
+    <% stage.rollbar_webhooks.build %>
+
+    <div class="col-lg-offset-2">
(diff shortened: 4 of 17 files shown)


plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <fieldset>
+    <legend>Rollbar</legend>
+    <% stage = form.object %>
+    <% stage.rollbar_webhooks.build %>
+
+    <div class="col-lg-offset-2">
+      <%= form.fields_for :rollbar_webhooks do |rollbar_field| %>
+        <div class="form-group">
+          <div class="col-lg-4">
+            <%= rollbar_field.url_field :webhook_url, class: "form-control", placeholder: "Webhook URL" %>
```
craig-day: maybe default the value to rollbar.com in addition to the placeholder?

db/schema.rb
```diff
@@ -125,7 +125,7 @@
   create_table "environment_variable_groups", id: :integer, force: :cascade do |t|
     t.string "name", null: false
     t.text "comment"
-    t.index ["name"], name: "index_environment_variable_groups_on_name", unique: true
+    t.index ["name"], name: "index_environment_variable_groups_on_name", unique: true, length: { name: 191 }
```
craig-day: I would undo all of the changes in here unrelated to your table, this file is a little hard to manage because samson supports three db types

plugins/rollbar/README.md
```diff
@@ -0,0 +1,8 @@
+# Rollbar Plugin
+
+This plugin adds deploy tracking to [Rollbar](https://rollbar.com/)
+
+1. The webhook URL is `https://api.rollbar.com/api/1/deploy/` unless it's a self-hosted instance.
```
craig-day: Does the code default to this, or are you telling people what it normally is?

db/schema.rb
hkjorgensen: Ok will change.

plugins/rollbar/README.md
hkjorgensen: This is the default url that must be put into Samson, will rephrase it.

plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
hkjorgensen: Then it will add an entry on every save. Added it to the README.md - because that's probably the first thing you have to look up :-)

plugins/rollbar/app/models/rollbar_notification.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class RollbarNotification
+  def initialize(webhook_url:, access_token:, environment:, revision:)
+    @webhook_url = webhook_url
+    @access_token = access_token
+    @environment = environment
+    @revision = revision
+  end
+
+  def self.deliver(webhook_url:, access_token:, environment:, revision:)
```
grosser: can we handle delivery right here instead of 8+ lines of extra passing things around to the instance ?

plugins/rollbar/app/models/rollbar_notification.rb
```diff
@@ -0,0 +1,34 @@
...
+    Rails.logger.info "Sending Rollbar notification..."
+    response = Faraday.post(@webhook_url,
+      access_token: @access_token,
+      environment: @environment,
+      revision: @revision)
+
+    if response.success?
+      Rails.logger.info "Sent Rollbar notification"
+    else
+      Rails.logger.info "Failed to send Rollbar notification: #{response.body}"
```
grosser: `status_code` would also be interesting

plugins/rollbar/app/models/rollbar_notification.rb
```diff
@@ -0,0 +1,34 @@
...
+      webhook_url: webhook_url,
+      access_token: access_token,
+      environment: environment,
+      revision: revision
+    ).deliver
+  end
+
+  def deliver
+    Rails.logger.info "Sending Rollbar notification..."
+    response = Faraday.post(@webhook_url,
```
grosser: prefer consistent alignment:

```
foo(
  bar,
  baz: 1
)
```

plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
grosser: `placeholder: 'rollbar.com'`

plugins/rollbar/db/migrate/20170711174532_create_rollbar_webhooks.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+class CreateRollbarWebhooks < ActiveRecord::Migration[5.1]
+  def change
+    create_table :rollbar_webhooks do |t|
+      t.text :webhook_url, null: false
+      t.string :access_token, null: false
+      t.string :environment, null: false
+      t.integer :stage_id, null: false
+
+      t.timestamps
```
grosser: FYI can use `t.index` here

plugins/rollbar/test/decorators/stage_decorator_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  let(:attributes) do
+    {
+      webhook_url: "http://rollbar.com/abcdef",
+      access_token: "test",
+      environment: 'staging',
+      stage_id: stage.id
+    }
+  end
+
+  describe "#rollbar" do
```
grosser: `rollbar` is not a method ... so use `describe "assigning rollbar attributes"` or so

plugins/rollbar/test/decorators/stage_decorator_test.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+describe Stage do
+  let(:stage) { stages(:test_staging) }
+  let(:attributes) do
```
grosser: not used ?

plugins/rollbar/test/models/rollbar_notification_test.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.covered! uncovered: 2
```
grosser: cover em all!

hkjorgensen:
@grosser thank you for your feedback, I'll address some of it.
It _is_ just another webhook (like most of Samsons plugins), but the general webhooks does not allow defining custom variables and mapping deploy data to the webhook payload - that would be awesome a lot of the Samson plugins.

In this case, Rollbar requires the following data; `access_token`, `environment`, `revision`.

hkjorgensen:
@grosser can you give it another spin? I'll squash the last commits after an approval.

plugins/rollbar/app/models/rollbar_notification.rb
hkjorgensen: I've updated it to match the suggested style, Rubocop should be updated to enforce this :-)

plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
hkjorgensen: The other plugins that define a placeholder uses. "Webhook URL" was aiming for consistency. `README.md` explains how to get things up and running.

grosser:
Only thing that would need a mapping would be `revision` everything else can be url-encoded ... so could add this as `mappings` field to outbound-webhooks ... empty by default and can add `revision=%{revision}` ?

grosser:
would still like to get rid of the `.deliver -> initialize -> #deliver` indirection ... but otherwise looks alright.

grosser:
... unifying more of these things into outbound webhooks would be nice since we can share UI improvements / error handling etc ... but maybe for others mapping is not that easy as in this case ... still most of them should be straight forward like `revision` here ...

hkjorgensen:
> ... unifying more of these things into outbound webhooks would be nice since we can share UI improvements / error handling etc ... but maybe for others mapping is not that easy as in this case ... still most of them should be straight forward like revision here ...

Upgrading the general webhook plugin could turn it into a powerful tool! The UI is the tricky part. Besides a 1:1 mapping and it could use simple string transformation DSL; maybe regexp based search/replace.

grosser:
maybe no UI at all ... just add %{revision} into the url and we replace it ?

hkjorgensen:
> would still like to get rid of the .deliver -> initialize -> #deliver indirection ... but otherwise looks alright.
@grosser addressed - approved?

plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
```diff
@@ -0,0 +1,27 @@
...
+          </div>
+
+          <div class="col-lg-3">
+            <%= rollbar_field.text_field :access_token, class: "form-control", placeholder: "Access token" %>
+          </div>
+
+          <div class="col-lg-3">
+            <%= rollbar_field.text_field :environment, class: "form-control", placeholder: "Environment" %>
```
grosser: `required: true` for all of them ?

plugins/rollbar/test/models/rollbar_notification_test.rb
grosser: still 2 uncovered lines ?

grosser:
use a password field for access_token ?
... also might want to ignore that from the logs ... see 00_filter_parameter_logging.rb initializer

hkjorgensen:
@grosser - thanks, applied all of them. 👍  ?

plugins/rollbar/app/views/samson_rollbar/_stage_form.html.erb
```diff
@@ -0,0 +1,27 @@
+<% if DeployGroup.enabled? %>
+  <fieldset>
+    <legend>Rollbar</legend>
+    <% stage = form.object %>
+    <% stage.rollbar_webhooks.build %>
```
grosser: FYI `@stage` works too

plugins/rollbar/test/models/rollbar_notification_test.rb
```diff
@@ -0,0 +1,41 @@
...
+          environment: 'test',
+          revision: 'v1'
+        }
+      )
+
+      notification.deliver
+      assert_requested request
+    end
+
+    it 'doesnt work' do
```
grosser: better with `it 'shows error messages'` or so ... but not important

grosser:
maybe have some kind of test callback when saving ... not sure if that is possible without a deploy ... also check that all rollbar instances are even reachable from samson ... from console or curl from host

PR #2133, opened 2017-07-18
Rollbar plugin. Do not make fields required

hkjorgensen:
This is regression after introducing the Rollbar plugin. https://github.com/zendesk/samson/pull/2128

Having `required: true` disables the page and users can not save without entering Rollbar information.


grosser:
need fields to be required then or partially filled out will fail with sql errors ?

PR #2137, opened 2017-07-20
wait for pod to be delete before trying to re-create

grosser:
`KubeException: object is being deleted: pods "foo-migrate" already exists`

https://zendesk.airbrake.io/projects/95346/groups/1998210054914253528/notices/1998210053150968889?tab=notice-detail

previously we sent a delete and then kept going ... which worked pretty well but was a race condition ... especially if any pod takes a second to shut down ... and for other services/daemonsets it would run 2 in parallel ... which is pretty scary too ... so making sure delete properly succeeds before we continue now

also making delete consistent across all resource types to always ignore if the resource does not currently exist

I think the retry logic is unnecessary for everything except Pod, but it is cheap and keeps things consistent

/cc @jonmoter 
@irwaters 

... test-deployed in staging and works ...

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -38,9 +38,18 @@ module Kubernetes
         end
       end
 
+      # wait for delete to finish before doing further work so we don't run into duplication errors
+      # - first wait is 0 since the request itself already took a few ms
+      # - sum of waits should be ~30s which is the default delete timeout
       def delete
-        request(:delete, name, namespace)
-        expire_cache
+        return true unless running?
+        request_delete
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -38,9 +38,17 @@ def revert(previous)
...
       end
 
+      # wait for delete to finish before doing further work so we don't run into duplication errors
+      # sum of waits should be ~30s which is the default delete timeout
       def delete
-        request(:delete, name, namespace)
-        expire_cache
+        return true unless running?
+        request_delete
+        [0.0, 0.1, 0.2, 0.5, 1, 2, 4, 8, 16].each do |wait|
```
grosser: first wait is 0 since the request itself already takes a few ms and might be long enough

PR #2138, opened 2017-07-20
dont use list_with_show_more if display_limit is nil

craig-day:
Small fix from a bug introduced in https://github.com/zendesk/samson/pull/2127

For environment variable groups, the index page can be compact but the show page should list all of the projects using a group. This will not use the `list_with_show_more` helper if the display_limit is nil.

/cc @zendesk/samson

### References
 - https://zendesk.airbrake.io/projects/95346/groups/1998865137687537194?tab=overview

### Risks
- Level: Low. The Environment Variable Groups page is broken as-is.


grosser:
:+1: dammit, just tracked that down to your PR :D

PR #2139, opened 2017-07-21
also detect squasch merge commits if they include the pr number as pr…

grosser:
…oposed by github UI

copied from https://github.com/zendesk/changes_since/pull/25

not super reliable but hopefully better then not supporting it at all ...
asked github support if we can get proper api

screen shot 2017-07-21 at 11 29 34 am

@said-abidi


grosser:
github support says use https://help.github.com/articles/searching-issues/#search-by-the-commit-shas-within-a-pull-request ... so I'll give that a try next ...

PR #2143, opened 2017-07-25
Remove release validation that build project is same as release project

mikehuston:
Allow build to be deployed in project that it was not created in so that
we can deploy the same build with different environment variables in
different projects.

Other solutions considered:
1. making builds unique on project AND sha but this meant having to
build the same sha twice which seemed kind of wasteful
2. use multiple stages, and leave a role unset for a stage. this does
not work because roles are configured per deploy group

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-1059

### Risks
- Level: Med - I have low familiarity with this codebase and critical review is appreciated

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -12,7 +12,6 @@ module Kubernetes
 
     validates :project, :git_sha, :git_ref, presence: true
     validate :validate_docker_image_in_registry, on: :create
-    validate :validate_project_ids_are_in_sync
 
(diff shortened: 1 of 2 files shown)


grosser:
was added in https://github.com/zendesk/samson/pull/912 so adding /cc @jonmoter 

I'd say risk is that some callbacks/notifications might be called on the wrong project ... but that will only affect the projects that do this weird build sharing ... so hopefully not that bad if it happens ...

PR #2148, opened 2017-07-26
allow deprecating keys so we stop resolving them but do not break run…

grosser:
…ning pods

Replacing used secrets is tricky atm: they will be instantly deleted and running pods can no longer restart,
so all projects that use the secret need to be re-deployed which is dangerous and brittle.

This PR will make new deploys fail during resolve phase and old deploys able to restart
once a id is deprecated a while it should be able to be deleted without side-effects

@irwaters @jonmoter @craig-day

### DIFF
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -1,28 +1,29 @@
 # frozen_string_literal: true
 class SecretsController < ApplicationController
   ADD_MORE = 'Save and add another'
+  UPDATEDABLE_ATTRIBUTES = [:value, :visible, :deprecated_at, :comment].freeze
 
   include CurrentProject
 
   before_action :find_project_permalinks
   before_action :find_secret, only: [:update, :show]
 
-  before_action :convert_visible_to_boolean, only: [:update, :create, :new]
+  before_action :normalize_params_for_backend, only: [:update, :create, :new]
   before_action :authorize_resource!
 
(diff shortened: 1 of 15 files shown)


.env.example
```diff
@@ -129,6 +129,7 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 ## Secret storage
 # SECRET_STORAGE_BACKEND= # optional, should be one of: SecretStorage::DbBackend (default) or SecretStorage::HashicorpVault
 # SECRET_STORAGE_SHARING_GRANTS=true # optional, instead of sharing global secrets by default, access has to be granted
+# SECRET_STORAGE_DEPRECATED_IDS=a/b/c/d # optional, ids that should longer be resolved, to not use a id on a new deploy but keep the old ones working
```
irwaters: this seems like a strange place to have this as it'll require a re-deploy if someone wants to deprecate a key.  Can't really think of a better place though, unless we store it in db or something?

grosser:
reworked now to use actual columns ... and reuse the lookup_cache to show visible/deprecated on the index page:

screen shot 2017-07-27 at 5 20 42 pm
screen shot 2017-07-27 at 5 25 57 pm
screen shot 2017-07-27 at 5 26 09 pm
screen shot 2017-07-27 at 5 20 30 pm

PR #2150, opened 2017-07-26
speed up secret search by value

grosser:
instead of caching just the ids switch to caching selected metadata ... being careful not to keep secret values in memory

 - prepares duplication check on save ... since now it can be fast
 - prepared deprecated secrets ... since now we can store `deprecated` it in this cache
 - prepares showing more metadata in index page (visible/deprecated/has-comment etc)

### DIFF
--- a/app/models/secret_storage.rb
+++ b/app/models/secret_storage.rb
@@ -1,16 +1,18 @@
 # frozen_string_literal: true
+require 'digest/sha2'
+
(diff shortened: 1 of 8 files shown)


grosser:
threw in a mutex for the race condition ... I'll test it on staging and then prod ... so we can revert if it somehow fails ... before I start building more features on this :)

PR #2159, opened 2017-07-29
Add API user search endpoint

ashmckenzie:
This PR adds a new `/api/user_search` endpoint that allows one to search for a user given their email address.

/cc @zendesk/samson @grosser 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -3,7 +3,13 @@ class Api::UsersController < Api::BaseController
   before_action :authorize_resource!
 
   def destroy
-    User.find(params.require(:id)).soft_delete!
-    head :ok
+    user = User.where(id: params.require(:id)).first
+
+    if user
+      user.soft_delete!
+      head :ok
+    else
+      head :not_found
+    end
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/api/users_controller.rb
```diff
@@ -3,7 +3,13 @@ class Api::UsersController < Api::BaseController
   before_action :authorize_resource!
 
   def destroy
-    User.find(params.require(:id)).soft_delete!
```
grosser: old code already returned 404 when not found ... via `.find` failure

grosser:
search endpoint is `index` for most controllers ... how about adding a email filter there ?

app/controllers/api/users_controller.rb
ashmckenzie: it also returns a boatload of HTML junk as well :)  this address that

ashmckenzie:
will change to `#index` and user query string

ashmckenzie:
green :)

app/controllers/api/users_controller.rb
grosser: then the solution should not be a 1-off fix ... if api 404 renders as html then we need a general fix ...

grosser:
... I'll make a PR to fix the not-found and add the search ... brb

grosser:
404 works just fine :)
https://github.com/zendesk/samson/pull/2160

grosser:
... and here is the search by email api https://github.com/zendesk/samson/pull/2161

ashmckenzie:
Closing in favour of https://github.com/zendesk/samson/pull/2162

PR #2184, opened 2017-08-14
do not add email flag on newer docker clients

grosser:
`[22:06:28] Flag --email has been deprecated, will be removed in 17.06.`


grosser:
single-cov is having none of it either ... `app/models/docker_builder_service.rb new uncovered lines introduced (1 current vs 0 configured)
`

PR #2186, opened 2017-08-15
Prevent editing and destroying build commands which are being used elsewhere.

hsume2:
Recently, I noticed that it is possible for a build command to be used as a command in a stage. When editing the stage, that change will propagate to the build command and vice versa. This is potentially dangerous when the link between the two is unclear. This PR makes it clear where else the build command may be used.

Alternatively, it is possible to make it such that build commands cannot be re-used. Though this change is outside the scope of this PR and would require some discussion as well.

### Steps to Reproduce the Issue

1. Navigate to http://samson.zd-dev.com/projects/example-project/build_command and save a build command
2. [Edit the stage of any project](http://samson.zd-dev.com/projects/example-project/stages/production/edit) and check the command from step 1 and click Save.
3. Navigate to http://samson.zd-dev.com/projects/example-project/build_command, select all the text in the command textarea, delete it, and click Save.
4. Edit the stage from step 2 and notice that the command is no longer there. In fact it was destroyed despite its usages.

With this PR this is what step 3 looks like:

screen shot 2017-08-14 at 8 06 01 pm

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Low: Editing pre build command may not work

### DIFF
--- a/app/controllers/build_commands_controller.rb
+++ b/app/controllers/build_commands_controller.rb
@@ -11,7 +11,11 @@ class BuildCommandsController < ApplicationController
   def update
     command = params[:command][:command]
     if command.blank?
-      @command.destroy!
+      Command.transaction do
+        current_project.update_attributes(build_command: nil)
+        @command.projects.reload
+        @command.destroy!
+      end
     else
       @command.command = command
       @command.project = @project
--- a/app/controllers/build_commands_controller.rb
+++ b/app/controllers/build_commands_controller.rb
@@ -11,7 +11,7 @@ class BuildCommandsController < ApplicationController
   def update
     command = params[:command][:command]
     if command.blank?
-      @command.destroy
+      @command.destroy!
     else
       @command.command = command
       @command.project = @project
--- a/app/controllers/build_commands_controller.rb
+++ b/app/controllers/build_commands_controller.rb
@@ -11,11 +11,9 @@ class BuildCommandsController < ApplicationController
   def update
     command = params[:command][:command]
     if command.blank?
-      Command.transaction do
-        current_project.update_attributes(build_command: nil)
-        @command.projects.reload
-        @command.destroy!
-      end
+      current_project.update_attributes(build_command: nil)
+      @command.projects.reload
+      @command.destroy!
     else
       @command.command = command
       @command.project = @project
--- a/app/controllers/commands_controller.rb
+++ b/app/controllers/commands_controller.rb
@@ -8,7 +8,7 @@ class CommandsController < ApplicationController
   before_action :authorize_custom_project_admin!, except: PUBLIC
 
   def index
-    @commands = Command.order(:project_id).page(page).includes(:stages, :projects)
+    @commands = Command.order(:project_id).page(page)
     if search = params[:search]
       if query = search[:query].presence
         query = ActiveRecord::Base.send(:sanitize_sql_like, query)
--- a/app/controllers/commands_controller.rb
+++ b/app/controllers/commands_controller.rb
@@ -8,7 +8,7 @@ class CommandsController < ApplicationController
   before_action :authorize_custom_project_admin!, except: PUBLIC
 
   def index
-    @commands = Command.order(:project_id).page(page)
+    @commands = Command.order(:project_id).page(page).includes(:stages, :projects)
     if search = params[:search]
       if query = search[:query].presence
         query = ActiveRecord::Base.send(:sanitize_sql_like, query)
@@ -52,8 +52,14 @@ class CommandsController < ApplicationController
   end
 
   def destroy
-    @command.destroy
-    successful_response('Command removed.')
+    if @command.destroy
+      successful_response('Command removed.')
+    else
+      respond_to do |format|
+        format.html { render :show }
+        format.js { render json: {}, status: :unprocessable_entity }
+      end
+    end
   end
 
   private
--- a/app/models/command.rb
+++ b/app/models/command.rb
@@ -13,6 +13,8 @@ class Command < ActiveRecord::Base
 
   validates :command, presence: true
 
+  before_destroy :ensure_unused
+
   def self.global
     where(project_id: nil)
   end
@@ -43,5 +45,27 @@ class Command < ActiveRecord::Base
   def self.usage_ids
     StageCommand.pluck(:command_id) +
     Project.pluck(:build_command_id)
+  end
+
+  # We should not destroy the command if it is associated with any stages.
+  # Also, we should not destroy the command if it is the build commands for multiple projects.
+  #
+  # It isn't possible through the UI to associate a build command with multiple projects.
+  # However, it is possible through the API and the console to do this. So in the case that
+  # this does happen, we should prevent such commands from being destroyed.
+  #
+  # When the build command is associated with only one project, it can be destroyed via this code path:
+  # https://github.com/zendesk/samson/blob/master/app/controllers/build_commands_controller.rb#L13
+  def unused?
+    stages.empty? && projects.count <= 1
+  end
+
+  private
+
+  def ensure_unused
+    unless unused?
+      errors.add(:base, 'Can only delete unused commands.')
+      throw(:abort)
+    end
   end
 end
--- a/app/models/command.rb
+++ b/app/models/command.rb
@@ -47,23 +47,10 @@ class Command < ActiveRecord::Base
     Project.pluck(:build_command_id)
   end
 
-  # We should not destroy the command if it is associated with any stages.
-  # Also, we should not destroy the command if it is the build commands for multiple projects.
-  #
-  # It isn't possible through the UI to associate a build command with multiple projects.
-  # However, it is possible through the API and the console to do this. So in the case that
-  # this does happen, we should prevent such commands from being destroyed.
-  #
-  # When the build command is associated with only one project, it can be destroyed via this code path:
-  # https://github.com/zendesk/samson/blob/master/app/controllers/build_commands_controller.rb#L13
-  def unused?
-    stages.empty? && projects.count <= 1
-  end
-
   private
 
   def ensure_unused
-    unless unused?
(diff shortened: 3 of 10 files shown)


grosser:
Good catch ...
I'd prefer not duplicating all that logic ...
either use a partial or ideally just disable edit and say `Command is used multiple times, edit on command page <link>`

hsume2:
@grosser I like the second option. I'll make that change. Also, it seems that we don't guard against deleting commands with usages at the controller level. What do you think about:

1. Adding a guard here: https://github.com/zendesk/samson/blob/master/app/controllers/commands_controller.rb#L54
2. Adding a guard here as well to prevent destroying if there are usages: https://github.com/zendesk/samson/blob/master/app/controllers/build_commands_controller.rb#L13

grosser:
I think that should be add the model level ... and then then controller
should use `if destroy` logic ... before_destroy and errors.add should do
the trick

hsume2:
@grosser turns out there's a bit more nuance with destroying commands. For a command that's only used as a stage command, it is easy to go to each stage and remove usage of that command. For build commands, there is no way to unlink a build command from a project. Hence why the build command controller destroys the command if it's blank.

However, that means that we lose the command history at that point. Given that stages keep track of their history when a command is deleted I propose the following things:

1. Don't allow destroying build commands. I think it's important to have the history there for auditing purposes. Which means there are two options for clearing your build command:
2. Allow commands to be blank. The user clears the command, clicks save, and we keep the history.
3. Don't allow commands to be blank and the user must put in a null command. For example: "# do nothing".

grosser:
I meant for commands to simply fail their destruction ... they are not supposed to auto-clean-up

app/controllers/build_commands_controller.rb
```diff
@@ -10,14 +10,10 @@ def show
 
   def update
     command = params[:command][:command]
-    if command.blank?
```
grosser: do a `if blank && unused` check instead ... would love to not leave weird commands hanging around

app/controllers/commands_controller.rb
```diff
@@ -52,8 +52,14 @@ def update
   end
 
   def destroy
-    @command.destroy
-    successful_response('Command removed.')
+    if @command.destroy
+      successful_response('Command removed.')
+    else
+      respond_to do |format|
+        format.html { render :show }
```
grosser: does the user see anything when this happens ? ... set a flash ?

app/models/command.rb
```diff
@@ -11,7 +11,14 @@ class Command < ActiveRecord::Base
 
   belongs_to :project, optional: true
 
-  validates :command, presence: true
+  before_destroy :check_usages
+
+  def check_usages
```
grosser: private

app/models/command.rb
```diff
@@ -11,7 +11,14 @@ class Command < ActiveRecord::Base
 
   belongs_to :project, optional: true
 
-  validates :command, presence: true
+  before_destroy :check_usages
```
grosser: `ensure_unused`

app/models/command.rb
```diff
@@ -11,7 +11,14 @@ class Command < ActiveRecord::Base
...
+
+  def check_usages
+    if usages.count.positive?
```
grosser: `usages.any?`

app/controllers/build_commands_controller.rb
grosser: ... just do `destroy!` and let it blow up ... user should have seen the warning ... also don't need js response ...

hsume2:
@grosser sorry, I should clarify. I am describing what the user would do in the situation that they want to clean up usage of the command so they can delete it.

app/views/build_commands/show.html.erb
```diff
@@ -13,9 +13,15 @@
...
-    <%= form.text_area :command, cols: 155, rows: [form.object.command.to_s.count("\n") + 1, 10].max %>
+    <% if used_elsewhere %>
+      <p><b>This command is used multiple times, go <%= link_to "here", command_path(@command) %> to edit.</b></p>
+    <% end %>
+
+    <%= form.text_area :command, cols: 155, rows: [form.object.command.to_s.count("\n") + 1, 10].max, disabled: used_elsewhere %>
 
-    <%= form.submit "Save", class: "btn btn-primary" %> |
+    <% unless used_elsewhere %>
+      <%= form.submit "Save", class: "btn btn-primary" %> |
```
grosser: can do a 1-liner too ... might look neater

grosser: nvm ... trailing `|`

config/environments/development.rb
```diff
@@ -11,7 +11,7 @@
   config.eager_load = false
 
   # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
+  config.consider_all_requests_local       = false
```
grosser: undo

app/controllers/build_commands_controller.rb
hsume2: @grosser If we just do `destroy!` and let it blow up then the only way the user can have the project not do anything before building the docker image is to save a null command. For example, `/* do nothing */`. See point 3 of https://github.com/zendesk/samson/pull/2186#issuecomment-322545884.

I'm fine with any of those options. I just want to make it clear how we intend users to use Samson. What do you think?

grosser: so a user would not be able to unlink a command unless it's the last usage ... 
but we cannot allow editing a potentially reused/global command ... that might be bad ...
users could work around it by deleting it from other usage locations ...

how about: do the actual bugfix first ... make it not editable/destroyable if it's reused ... then we worry about usability later ...

app/controllers/commands_controller.rb
hsume2: I think it will be good to keep this just in case. Good for API consumers as well.

hsume2:
@grosser 💇‍♂️ I've made the minimal changes necessary for the bug fix.

app/controllers/commands_controller.rb
```diff
@@ -8,7 +8,7 @@ class CommandsController < ApplicationController
   before_action :authorize_custom_project_admin!, except: PUBLIC
 
   def index
-    @commands = Command.order(:project_id).page(page)
+    @commands = Command.order(:project_id).page(page).includes(:stages, :projects)
```
grosser: no includes ... we use goldiloader

app/models/command.rb
```diff
@@ -44,4 +46,26 @@ def self.usage_ids
...
+  # Also, we should not destroy the command if it is the build commands for multiple projects.
+  #
+  # It isn't possible through the UI to associate a build command with multiple projects.
+  # However, it is possible through the API and the console to do this. So in the case that
+  # this does happen, we should prevent such commands from being destroyed.
+  #
+  # When the build command is associated with only one project, it can be destroyed via this code path:
+  # https://github.com/zendesk/samson/blob/master/app/controllers/build_commands_controller.rb#L13
+  def unused?
+    stages.empty? && projects.count <= 1
```
grosser: there is a usages method for that ... better to use that then re-implement usage counting here

app/views/commands/index.html.erb
```diff
@@ -26,8 +26,8 @@
           <td>
             <%= link_to "Edit", command %>
             |
-            <% used = usage_ids.count(command.id) %>
-            <% if used > 0 %>
+            <% unless command.unused? %>
+              <% used = usage_ids.count(command.id) %>
```
grosser: should not modify this line ... already checked via usage_ids

app/models/command.rb
hsume2: @grosser If `stages.empty?` and `projects.count == 1` then we should allow destroying. The `usages` method doesn't account for that logic.

grosser:
it's kind of the same as saying 'if it has 1 usage' since it will always
the project usage it's associated to

app/models/command.rb
hsume2: @grosser if it has 1 usage and the usage is on a stage, then should we allow destroying?

grosser: maybe make the workflow be `unset on project ... try to destroy` ... this way the usages will be 0 ?

hsume2:
@grosser how's it look now?

app/controllers/build_commands_controller.rb
```diff
@@ -11,7 +11,11 @@ def show
   def update
     command = params[:command][:command]
     if command.blank?
-      @command.destroy
+      Command.transaction do
+        current_project.update_attributes(build_command: nil)
+        @command.projects.reload
+        @command.destroy!
```
grosser: could remove the transaction and do a destroy here ... that would then fail ... but we don't care ... and the build command would still be removed

test/controllers/build_commands_controller_test.rb
```diff
@@ -48,6 +49,33 @@
           project.reload.build_command.must_equal nil
         end
       end
+
+      # It isn't possible through the UI to associate a build command with multiple projects.
+      # However, it is possible through the API and the console to do this. So in the case that
+      # this does happen, we should prevent such commands from being destroyed.
```
grosser: I think this can happen when creating a build command and then using it in a stage ... that should be possible ...

hsume2: @grosser That would be through `:stage_commands`. I don't think it's possible through the UI to make more than one Project have the same `build_command_id`.

```
class Command
  has_many :projects, foreign_key: :build_command_id
end
```

app/controllers/build_commands_controller.rb
hsume2: 👍

grosser:
🎉  thx for the PR nice safe-guard :)

hsume2:
@grosser thanks for the extensive review, glad to help!

PR #2189, opened 2017-08-15
simplify job methods

grosser:
/cc @zendesk/samson

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -60,7 +60,7 @@ class Job < ActiveRecord::Base
     updated_at - created_at
   end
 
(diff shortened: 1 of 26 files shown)


app/models/job.rb
```diff
@@ -74,12 +74,12 @@ def commands
     commands
   end
 
-  def stop!(canceller)
+  def cancel(canceller)
```
craig-day: why not a `!` method anymore?

grosser: part of my next change ... de-bang all this foobar

pschambacher: @grosser, de-banger of foobars

PR #2199, opened 2017-08-17
bump audit changes size so we can audit big env variable setups

grosser:
https://zendesk.airbrake.io/projects/95346/groups/2019018614023597976


db/migrate/20170817213634_bump_audit_size_for_big_envs.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+class BumpAuditSizeForBigEnvs < ActiveRecord::Migration[5.1]
+  def up
+    change_column :audits, :audited_changes, :text, limit: 1.gigabyte / 4 - 1
```
pschambacher: I know that `/` takes precedence over `-` but I would have added brackets for human clarity

grosser: updated

PR #2201, opened 2017-08-17
replace datatables with dedicated js for fixed column header scrollin…

grosser:
…g to reduce size/gems

screen shot 2017-08-17 at 4 42 52 pm


grosser:
ran into jshint trouble .. so made a vendor/assets folder for all our copy-pasted assets ...

PR #2205, opened 2017-08-21
TOOLS-1816 add custom message when kubernetes limits are reached

grosser:
screen shot 2017-08-21 at 4 14 53 pm

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -137,6 +137,7 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # KUBERNETES_LOG_TIMEOUT=20 # how long to wait for logs to appear in seconds
 # KUBERNETES_LOG_LINES=50 # how many lines of logs to show when a deploy fails
 # KUBERNETES_ALLOWED_VOLUME_HOST_PATHS=/data/ # prevent containers from mounting dangerous directories
+# KUBERNETES_USAGE_LIMIT_WARNING="If you need more, ask Steve!" # help message to display when user reaches usage limit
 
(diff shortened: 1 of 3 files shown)


.env.example
```diff
@@ -137,6 +137,7 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # KUBERNETES_LOG_TIMEOUT=20 # how long to wait for logs to appear in seconds
 # KUBERNETES_LOG_LINES=50 # how many lines of logs to show when a deploy fails
 # KUBERNETES_ALLOWED_VOLUME_HOST_PATHS=/data/ # prevent containers from mounting dangerous directories
+# KUBERNETES_USAGE_LIMIT_WARNING="If you need more, ask Steve!" # help message to display when user reaches usage limit
```
pschambacher: Steve approves this message!

plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
```diff
@@ -92,16 +92,18 @@ def requests_below_limits
     def requests_below_usage_limits
       return unless limit = UsageLimit.most_specific(project, deploy_group)
       message = "must be less than or equal to kubernetes usage limit"
+      usage_limit_warning = ENV["KUBERNETES_USAGE_LIMIT_WARNING"]
```
pschambacher: Should this be a constant so it's not necessary to read it every time?

grosser: makes testing easier ... and not really a big issue ...

PR #2208, opened 2017-08-22
TOOLS-1701 tag docker images after deploy with stage and deploy group…

grosser:
… permalinks

If it fails it will report to airbrake and not interrupt the deployment flow

@jonmoter @bcolferzd

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -40,6 +40,11 @@ PATH
       flowdock (~> 0.6.0)
 
 PATH
+  remote: plugins/gcloud_image_tagger
+  specs:
+    gcloud_image_tagger (0.0.0)
+
+PATH
   remote: plugins/github
   specs:
     samson_github (0.0.0)
@@ -521,6 +526,7 @@ DEPENDENCIES
   faraday-http-cache
   flay
   forking_test_runner
+  gcloud_image_tagger!
   github-markdown
   goldiloader
   kaminari
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -137,6 +137,11 @@ class Job < ActiveRecord::Base
     execution.try :pid
   end
 
+  def append_output!(more)
+    reload
+    update_columns(output: "#{output}#{more}", updated_at: Time.now)
+  end
+
   private
 
   def validate_globally_unlocked
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -16,6 +16,7 @@ Available plugins:
  - [ENV var management](https://github.com/zendesk/samson/tree/master/plugins/env)
  - [Flowdock notification](https://github.com/zendesk/samson/tree/master/plugins/flowdock)
  - [Github](https://github.com/zendesk/samson/tree/master/plugins/github)
+ - [Gcloud image tagger](https://github.com/zendesk/samson/tree/master/plugins/gcloud_image_tagger)
  - [Jenkins jobs management](https://github.com/zendesk/samson/tree/master/plugins/jenkins)
  - [Hipchat notification](https://github.com/listia/samson_hipchat)
  - [Hyperclair security scanner for docker images](https://github.com/zendesk/samson/tree/master/plugins/hyperclair)
--- a/plugins/gcloud_image_tagger/README.md
+++ b/plugins/gcloud_image_tagger/README.md
@@ -1,0 +1,11 @@
+# Gcloud image tagger Plugin
+
+Tag gcloud images with the stage permalink they deployed to, so developers can pull down the "producion" image.
+
+## Setup
+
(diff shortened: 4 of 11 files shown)


app/models/deploy_service.rb
```diff
@@ -122,4 +123,25 @@ def notify_outbound_webhooks(deploy)
   def send_sse_deploy_update(type, deploy)
     SseRailsEngine.send_event('deploys', type: type, deploy: DeploySerializer.new(deploy, root: nil))
   end
+
+  # TODO: make this work with remote images too ... and images where the local image no longer exists
```
jonmoter: Here's a blog post explaining how you can add a tag to a docker image in a remote registry, without (necessarily) needing to do a `docker pull` and then a `docker push`: https://wheleph.gitlab.io/2016/07/13/tagging-using-docker-registry-http-api-v2/

app/models/deploy_service.rb
```diff
@@ -122,4 +123,25 @@ def notify_outbound_webhooks(deploy)
...
+  def tag_docker_images(deploy)
+    return unless ENV["DOCKER_FEATURE"]
+    return unless deploy.succeeded?
+    return unless builds = deploy.project.builds.
+      where(git_sha: deploy.job.commit).where.not(docker_image_id: nil).presence
```
jonmoter: Might be a good idea to extract this into a method of `deploy`, and then have a test or two on that method.

app/models/docker_builder_service.rb
```diff
@@ -102,6 +102,11 @@ def run(push: false, tag_as_latest: false)
     JobExecution.start_job(@execution)
   end
 
+  def tag(tag)
```
jonmoter: Since this causes an immediate mutation of the remote registry, this might be worth calling `tag!`

grosser:
works when DOCKER_KEEP_BUILT_IMGS=1 is set ... otherwise fails with `Failed to find local docker image` ... the images seem to be pushed with different digests ... not sure why, maybe a docker thing ... but can pull locally after via ` docker pull foo.com/project-name:sandbox`

app/models/docker_builder_service.rb
grosser: renamed to push_tag for clarity ...

grosser:
https://success.docker.com/Cloud/Solve/How_do_I_authenticate_with_the_V2_API%3F seemed like a good way to auth via the api but did not work ... tried via faraday too and with docker-api gem ...

would want to use api v2 to add tags via manifest endpoint https://wheleph.gitlab.io/2016/07/13/tagging-using-docker-registry-http-api-v2/

jonmoter:
For the record, authenticating with an artifactory docker registry has its own special fu: https://www.jfrog.com/knowledge-base/how-to-use-docker-registry-api-with-artifactory-docker-repository-when-not-using-docker-client/

grosser:
Jfrog has their own foobar ... could not get it to work and asked their support ... https://www.jfrog.com/knowledge-base/how-to-use-docker-registry-api-with-artifactory-docker-repository-when-not-using-docker-client/

grosser:
Needs to enable "Google Cloud Container Builder API" api  https://console.cloud.google.com/apis/api/cloudbuild.googleapis.com/overview for this to work

grosser:
done now I hope ... review plz @jonmoter @irwaters

grosser:
- .json file needs to be readable by deploy group
 - possibly: login command was not executed as deploy user but as root ?
 - service-account is not allowed to add tags ... or the beta api is dead ...

```
`gcloud beta container images add-tag gcr.io/docker-images-180022/apps/gcr_watcher@sha256:803e79b637868afda4c9ceb95745147979f6ec469906b07aaf50d9205f9f74df gcr.io/docker-images-180022/apps/gcr_watcher:testing --quiet`
ERROR: gcloud crashed (BadStateException): Bad status during token exchange: 403
{"errors":[{"code":"DENIED","message":"Unable to access the repository; please check that you have permission to access it."}]}
```

jonmoter:
I might need to add IAM permissions for the Samson service account to modify the registry. Lemme try to figure out what the permissions need to be...

jonmoter:
Okay, I've added checking out [this page](https://cloud.google.com/container-registry/docs/access-control), I added the `roles/storage.admin` IAM role to the Samson service account.

grosser:
Same error ... also tried a re-login via gcloud auth
activate-service-account --key-file /data/samson/config/gcloud_auth.json
... but no luck ...

grosser:
works https://samsontest.zende.sk/projects/kube_service_watcher/deploys/4786
... good to go ? @jonmoter

jonmoter:
We'll have to figure out how to handle images that are also copied to other regions. Not sure if we should have Samson do that, or in another component.  Either way, can do that in another PR.

PR #2211, opened 2017-08-23
allow super admins to delete clusters

grosser:
screen shot 2017-08-23 at 4 16 36 pm
screen shot 2017-08-23 at 4 16 47 pm
screen shot 2017-08-23 at 4 34 08 pm
screen shot 2017-08-23 at 5 35 25 pm

### DIFF
--- a/app/models/command.rb
+++ b/app/models/command.rb
@@ -51,7 +51,7 @@ class Command < ActiveRecord::Base
 
   def ensure_unused
     if usages.any?
-      errors.add(:base, 'Can only delete unused commands.')
+      errors.add(:base, 'Can only delete when unused.')
       throw(:abort)
     end
   end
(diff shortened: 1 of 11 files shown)


jonmoter:
Is this a hard-delete or soft delete?  What happens if there are DeployGroups associated with the cluster?

grosser:
hard-delete ... their association is cleaned up
I'd prefer to keep it hard ... but add auditing ... need that for
updates/creates too ...
Could also show a usage count in the index UI ... maybe not allow deleting
used ... so many fancy features :D

jonmoter:
How about we prevent deleting unless you remove all associations to Deploy Groups? I want to make sure that I (or someone else) don't screw up Samson by accidentally deleting a legit cluster.

grosser:
sounds good ... also makes it obvious when a cluster is unused ... so might as well add that to the UI like commands does

grosser:
now with all the bells and whistles ...

PR #2212, opened 2017-08-25
remove redundant direct association in preparation for multi builds PR

grosser:
don't want to add a has_many for builds ... so just remove them ... we already have a association by `same git sha`

@irwaters

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -4,7 +4,6 @@ module Kubernetes
     self.table_name = 'kubernetes_releases'
 
     belongs_to :user
-    belongs_to :build, optional: true
     belongs_to :project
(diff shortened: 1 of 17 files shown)


plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -54,6 +53,10 @@ def url
       Rails.application.routes.url_helpers.project_kubernetes_release_url(project, self)
     end
 
+    def builds
+      Build.where(git_sha: git_sha)
```
jonmoter: It's possible for more than one `Project` to point to the same repository. So I'd scope this by the project, just to be safe.

grosser: builds are unique on the git-sha so that should be fine ... but still changed just to be safe ...

PR #2213, opened 2017-08-25
support builds for multiple dockerfiles

grosser:
support builds for multiple dockerfiles (just 2nd commit, first commit is from https://github.com/zendesk/samson/pull/2212)

@jonmoter @zenhao @irwaters 

in the container add `samson/dockerfile: Dockerfile.foobar`
tested this in master and it does not break deployment, so we should be fine to use that ...

for non-kubernetes the project gets a `dockerfiles` attribute that can be set to what dockerfiles should be built ... builds will then get queued for each dockerfile ... and have a the dockerfile suffix appended to their repo ... so foobar + Dockerfile.new -> foobar-new

the current workflow is a bit shitty ... user needs to set `dockerfiles` in the project to make everything work ... but that should only be an issue for those needing multiple dockerfiles ... so good enough until there is more trouble ...

first multi-stage deploy: https://samsontest.zende.sk/projects/truth_service/deploys/4624

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -22,7 +22,7 @@ class Integrations::BaseController < ApplicationController
     end
 
     if project.build_docker_image_for_branch?(branch)
-      create_docker_image
+      create_docker_images
     end
 
     stages = project.webhook_stages_for(branch, service_type, service_name)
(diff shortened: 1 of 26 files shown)


app/controllers/integrations/base_controller.rb
```diff
@@ -111,17 +111,20 @@ def service_name
...
+      options = {git_sha: commit, dockerfile: dockerfile}
+      next if scope.where(options).first
+
+      build = project.builds.create!(options.merge(
+        git_ref: branch,
+        description: message,
+        creator: user,
+        label: branch
+      ))
+      DockerBuilderService.new(build).run(push: true, tag_as_latest: true)
```
jonmoter: Remind me, does calling `run` block, or return almost immediately?  We might want to create all the builds first, and then build them in a separate loop.

grosser: it backgrounds the actual work ... so this should work just fine

app/models/project.rb
```diff
@@ -47,8 +47,16 @@ class Project < ActiveRecord::Base
 
   scope :search, ->(name) { where("name like ?", "%#{name}%") }
 
-  def docker_repo(registry)
-    File.join(registry.base, permalink_base)
+  def docker_repo(registry, dockerfile)
+    repo = File.join(registry.base, permalink_base)
+    if suffix = dockerfile.sub(/^Dockerfile\.?/, '').presence
+      repo << "-#{suffix.parameterize}"
+    end
```
jonmoter: Maybe add logic for handling the case where `dockerfile` looks like `nginx/Dockerfile` rather the `Dockerfile.nginx`

grosser: yep, good catch, fixed

devstuff:
@grosser Have you looked into supporting Docker's native multi-builds? https://docs.docker.com/engine/userguide/eng-image/multistage-build/

grosser:
different usecase.
 - will be supported once we upgrade docker
 - currently supported via pre-build-commands

jonmoter:
What does the Build#index page look like with multiple images?

PR #2215, opened 2017-08-25
rename builds label to name

grosser:


grosser:
@jonmoter  / @bcolferzd

PR #2218, opened 2017-08-28
tell user when we're deleting a docker image

jonmoter:
If `DOCKER_KEEP_BUILT_IMGS` env var isn't set, tell the end user we're deleting the image.

/cc @zendesk/samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/SAMSON-359

### Risks
- Level: Low

### DIFF
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -82,7 +82,10 @@ class DockerBuilderService
         if build_image(tmp_dir) # rubocop:disable Style/IfInsideElse
           ret = true
           ret = push_image(tag_as_latest: tag_as_latest) if push
-          build.docker_image.remove(force: true) unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
+          unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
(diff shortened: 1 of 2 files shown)


app/models/docker_builder_service.rb
```diff
@@ -82,7 +82,10 @@ def run(push: false, tag_as_latest: false)
         if build_image(tmp_dir) # rubocop:disable Style/IfInsideElse
           ret = true
           ret = push_image(tag_as_latest: tag_as_latest) if push
-          build.docker_image.remove(force: true) unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
+          unless ENV["DOCKER_KEEP_BUILT_IMGS"] == "1"
+            output.puts("Deleting docker image #{build.docker_image.id}")
```
grosser: this returns the `sha256:41d6fa09472fdf8ed663cb36d0dae276b632729c7df6a3693f2e86266acfbab0` ... which is a bit ugly ... should return the short-form id instead ? (same as shown in `docker images`)

```
id = build.docker_image.id.to_s
id = id[/sha256:(.{12})/, 1] || id
```

jonmoter:
Formatting the docker image id was too much of a hassle.

PR #2225, opened 2017-09-02
Export commit range in job execution

elblivion:
* exports the deploy changeset's commit range when executing jobs, so we can work out which files have been changed / added / deleted since the previous commit.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -207,6 +207,10 @@ class JobExecution
       TAG: (@job.tag || @job.commit)
     }.merge(@env)
 
+    if deploy = @job.deploy
+      env[:COMMIT_RANGE] = deploy.changeset.commit_range
+    end
+
     env.merge!(Hash[*Samson::Hooks.fire(:job_additional_vars, @job)])
 
     base_commands(dir, env) + @job.commands
--- a/test/models/job_execution_test.rb
+++ b/test/models/job_execution_test.rb
@@ -156,6 +156,7 @@ describe JobExecution do
     lines.must_include "[04:05:06] DEPLOYER_NAME=John Doe"
     lines.must_include "[04:05:06] REFERENCE=master"
     lines.must_include "[04:05:06] REVISION=#{job.commit}"
+    lines.must_include "[04:05:06] COMMIT_RANGE=#{job.commit}...#{job.commit}"
     lines.must_include "[04:05:06] TAG=v1"
     lines.must_include "[04:05:06] FOO=bar"
   end


test/models/job_execution_test.rb
```diff
@@ -139,6 +139,7 @@
     lines.must_include "DEPLOYER_NAME=John Doe"
     lines.must_include "REFERENCE=master"
     lines.must_include "REVISION=#{job.commit}"
+    lines.must_include "COMMIT_RANGE=#{deploy.changeset.commit_range}"
```
grosser: better spell it out so we can see what it will actually look like

app/models/job_execution.rb
```diff
@@ -202,6 +202,7 @@ def commands(dir)
       PROJECT_REPOSITORY: @job.project.repository_url,
       REFERENCE: @reference,
       REVISION: @job.commit,
+      COMMIT_RANGE: @job.deploy.changeset.commit_range,
```
grosser: not every job has a deploy ... so only set it if it has a deploy

grosser: also double-check that it's a string ...

test/models/job_execution_test.rb
elblivion: Not sure I follow... the actual commits depend on the test repo so they cannot be hardcoded, right?

Since there is only one deploy in the test I've changed it to expect a commit range beginning and ending with `job.commit`, how does that sound?

app/models/job_execution.rb
elblivion: Updated, let me know what you think. Apologies, I'm an infra engineer with not much dev background - learning as I go along!

app/models/job_execution.rb
```diff
@@ -207,6 +207,11 @@ def commands(dir)
       TAG: (@job.tag || @job.commit)
     }.merge(@env)
 
+    deploy = @job.deploy
+    if deploy && deploy.changeset.commit_range.is_a?(String)
```
grosser: `if deploy = @job.deploy` will be good enough ... I though commit-range is some kind of array :D

test/models/job_execution_test.rb
```diff
@@ -156,6 +156,7 @@ def last_line_of_output
     lines.must_include "[04:05:06] DEPLOYER_NAME=John Doe"
     lines.must_include "[04:05:06] REFERENCE=master"
     lines.must_include "[04:05:06] REVISION=#{job.commit}"
+    lines.must_include "[04:05:06] COMMIT_RANGE=#{job.commit}...#{job.commit}"
```
grosser: makes sense ... was thinking the commits might be static :)

elblivion:
Build is 💚,  please let me know if anything else is required.

BTW we intend to use this to trigger `kubectl apply` on Kubernetes specs from our infrastructure repo, updating the specs that have changed. I've seen that you've implemented Kubernetes support but in a particular (opinionated?) way, it would be awesome to know if there's a blog post or talk describing your set-up - googling anything related to Zendesk is hard because so many people use your service (including Contentful where I work)!

In any case many thanks for an awesome tool. ❤️

grosser:
check out the kubernetes plugin ... let me know what is unclear / needs more docs :)

grosser:
@elblivion are you still using this ... just came through that it's kinda low usage / needs an extra remote call which is kinda sad ...

elblivion:
No, we stopped using this because reasons... go ahead and remove it if you want. Thanks!

grosser:
thx!

Issue #2176, opened 2017-08-08
Pipelining deployments randomly queue for no reason


jacobbednarz:
Hello!

We recently swapped to a pipelined deployment where we have two stages (canary and production). Canary and production stages are similar however the canary stage only deploys to a tiny subset of our infrastructure whereas production will deploy to the remainder of the infrastructure. Since this swap, we've been seeing random queuing of deployments that never run. Here are the configurations for each stage:

`canary`

![](https://jacobbednarz.s3.amazonaws.com/shares/jxpzg.jpg)

`production`

![](https://jacobbednarz.s3.amazonaws.com/shares/jeykh.jpg)

I've also tried setting the max deployments to one per user but to no avail. Have you got any idea on what it could be or how to debug this further? I'm guessing it's something to do with the cleanup of locks or managing of locks since it's thinking something is deploying but there is nothing there. Appending `debug=1` to the deploys page doesn't show anything.

grosser:
- these 'random' deploys are for the canary stage ?
 - these 'random' deploys are duplicates (deploying the same sha/branch as another deploy) ?

I'd guess there is a webhook that triggers the canary deploy ... then also creates a release ... which triggers another canary deploy ... does removing the `Automatically deploy new releases` help ?

jacobbednarz:
> these 'random' deploys are for the canary stage ?

Sorry, I failed to mention it's only ever `production` that gets held up.

> these 'random' deploys are duplicates (deploying the same sha/branch as another deploy) ?

Nope, usually different commits/branches.

> I'd guess there is a webhook that triggers the canary deploy ... then also creates a release ... which triggers another canary deploy ... does removing the Automatically deploy new releases help ?

Yep, buildkite triggers a release  + deployment. If we turn that off, deployments won't automatically happen.

grosser:
Try turning off the `Automatically deploy new releases`

grosser:
if that solves it then we just need some new logic in the integration controller to never do duplicate deplosy

jacobbednarz:
@grosser I'm slightly confused here with removing that from the canary stage when it's the production stage that is randomly queuing. The canary stage never ends up in a weird state, just the second step of the pipeline.

grosser:
Did you set up a deploy pipeline between canary and production ? ... or are
you manually doing that ?

If there is nothing connecting the 2 stages and no other webhook then
that's really weird :(

jacobbednarz:
[...]

PR #2226, opened 2017-09-04
tell users to cancel bad deploys so they do not wait forever

grosser:
for https://github.com/zendesk/samson/issues/2176#issuecomment-326464560

better than telling users everything is ok and they wait forever ... maybe later either airbrake or auto-cancel ... not sure ... 

@jacobbednarz @irwaters

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -102,7 +102,7 @@ class Job < ActiveRecord::Base
   end
 
   def queued?
-    pending? && JobExecution.queued?(id)
+    JobExecution.queued?(id)
   end
 
   def active?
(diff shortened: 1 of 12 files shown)


jacobbednarz:
Code looks good to me. Is there anything I can provide to help track down where this bug is actually happening?

grosser:
hmmm could use the same check in the deploy-service and then immediately blow up ... 

can you try adding it in app/models/deploy_service.rb:45 ?

```
raise "Bad job" if deploy.job.pending? && !deploy.job.queued?
```

grosser:
just saw the message for 1s after doing a new deploy locally ... which happens because the job will be enqueued in a different thread ... saw more weird code ... so doing some refactoring in this PR next :/

PR #2228, opened 2017-09-06
extract build logic into a new class for clarity

grosser:
was getting too crazy and I'm about to add even more logic ...


grosser:
now in it's own file and dedicated tests ... much nicer :D

PR #2230, opened 2017-09-07
display & in titles nicely

grosser:
before:
screen shot 2017-09-06 at 5 05 44 pm

after:
screen shot 2017-09-06 at 5 04 41 pm

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -221,7 +221,10 @@ module ApplicationHelper
 
   def page_title(content = nil, in_tab: false, &block)
     content ||= capture(&block)
-    content_for :page_title, strip_tags(content) + (@project ? " - #{@project.name}" : "")
+    title_content = content
+    title_content += " - #{@project.name}" if @project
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -221,7 +221,9 @@ def additional_info(text)
 
   def page_title(content = nil, in_tab: false, &block)
     content ||= capture(&block)
-    content_for :page_title, strip_tags(content) + (@project ? " - #{@project.name}" : "")
+    title_content = strip_tags(content).gsub('&amp;', '&').html_safe
+    title_content << " - #{@project.name}" if @project
```
jonmoter: Is it possible for the project name to have a character like `&` in it?

grosser: stages can have that

grosser: ah I see ... yes we should do that after :D

grosser: fixed

PR #2231, opened 2017-09-07
Update Samson to support StatefulSets

libo:
Samson doesn't whitelist StatefulSet as a supported object in Kubernetes. Now that we're on K8S 1.5, we should allow that.

Not sure this is enough @grosser let me know

JIRA: https://zendesk.atlassian.net/browse/TOOLS-1865

cc: @jonmoter

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/cluster.rb
+++ b/plugins/kubernetes/app/models/kubernetes/cluster.rb
@@ -27,6 +27,10 @@ module Kubernetes
       @extension_client ||= build_client 'extensions/v1beta1'
     end
 
+    def apps_client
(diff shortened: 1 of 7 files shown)


libo:
cc: @dasch @jacobat

grosser:
It's not that easy ;)

I'll take over the PR ... need to add new `resource.rb` subclass and then check how these things are configured/deployed ... hopefully just the same as a deployment ...

libo:
Yes take over please! I am actually pretty curious to see. Let's try to see if "same as deployment" works :-) Thanks!

libo:
@irwaters here is a green deploy
https://samson.zende.sk/projects/help_center-memcached-k8s/deploys/344042

PR #2232, opened 2017-09-07
TOOLS-1855 allow multiple services per resource

grosser:
@sandlerr @jonmoter @irwaters 

```
kube-service-watcher            None             <nodes>       80:30653/TCP                                                  1m
kube-service-watcher2           None             <nodes>       81:30291/TCP                                                  1m
```

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -69,14 +69,7 @@ module Kubernetes
 
         case resource[:kind]
         when 'Service'
-          name = kubernetes_role.service_name
-          if name.to_s.include?(Kubernetes::Role::GENERATED)
-            raise(
-              Samson::Hooks::UserError,
-              "Service name for role #{kubernetes_role.name} was generated and needs to be changed before deploying."
-            )
-          end
-          resource[:metadata][:name] = name.presence || resource[:metadata][:name]
+          resource[:metadata][:name] = generate_service_name(resource[:metadata][:name])
 
           prefix_service_cluster_ip(resource)
 
@@ -90,6 +83,25 @@ module Kubernetes
           resource
         end
       end
+    end
+
+    def generate_service_name(config_name)
+      return config_name unless name = kubernetes_role.service_name.presence
+      if name.include?(Kubernetes::Role::GENERATED)
+        raise(
+          Samson::Hooks::UserError,
+          "Service name for role #{kubernetes_role.name} was generated and needs to be changed before deploying."
+        )
+      end
+
+      # users can only enter a single service-name so for each additional service we make up a name
+      # unless the given name already fits the pattern ... slight chance that it might end up being not unique
+      return config_name if config_name.start_with?(name)
+
+      @service_names_generated ||= 0
+      @service_names_generated += 1
+      name += "-#{@service_names_generated}" if @service_names_generated >= 2
+      name
     end
 
     # no ipv6 support
--- a/plugins/kubernetes/app/models/kubernetes/role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role.rb
@@ -58,7 +58,7 @@ module Kubernetes
         name = resource.fetch(:metadata).fetch(:labels).fetch(:role)
 
         # service
-        if service = config_file.service
+        if service = config_file.services.first
           service_name = service[:metadata][:name]
           if where(service_name: service_name).exists?
             service_name << "#{GENERATED}#{rand(9999999)}"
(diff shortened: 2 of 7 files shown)


plugins/kubernetes/app/models/kubernetes/role_verifier.rb
```diff
@@ -62,8 +62,15 @@ def verify_namespace
     end
 
     def verify_kinds
-      kinds = map_attributes([:kind]).sort_by(&:to_s)
-      return if SUPPORTED_KINDS.include?(kinds - ['ConfigMap'])
+      kinds = map_attributes([:kind])
+      kinds -= ['ConfigMap']
+      if kinds.count('Service') > 1
+        kinds -= ['Service']
+        kinds += ['Service']
```
henders: This is confusing :)

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -92,6 +85,23 @@ def store_resource_template
...
+          Samson::Hooks::UserError,
+          "Service name for role #{kubernetes_role.name} was generated and needs to be changed before deploying."
+        )
+      end
+
+      return config_name if config_name.start_with?(name)
+
+      @service_names_generated ||= 0
+      @service_names_generated += 1
+      name += "-#{@service_names_generated}" if @service_names_generated >= 2
```
henders: nit: `if @service_names_generated > 1` seems a bit more intuitive. But either way.

grosser:
hehe ... if there is more then 1 service consider it 1 service ... for
comparison with supported configurations

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -58,7 +58,7 @@ def self.seed!(project, git_ref)
         name = resource.fetch(:metadata).fetch(:labels).fetch(:role)
 
         # service
-        if service = config_file.service
+        if service = config_file.services.first
```
henders: I'm assuming it doesn't matter that we don't show all the services in the role UI.

grosser:
that's a bit weird ... since we only have 1 field for service-name I'm
using it and then adding a suffix ... kinda shady ...

plugins/kubernetes/app/models/kubernetes/role_verifier.rb
henders: Oh, you're just trying to remove duplicates?

henders:
Cool, I think it's grand anyway if the UI lags a bit behind for this, I don't know who uses the UI to look at the service name anyway.

👍

grosser:
yeah, but only for Service ... so it's a bit ugly :/

grosser:
added more docs for the weird parts

klaus993:
Hi @grosser 

I have an issue with some of this code (though I'm using `v1938`, this code is the same but in template_filler). I'm trying to deploy only one role, its yaml defines a deployment with three containers and only one service:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: app-name
  labels:
    project: app-name
    role: web
spec:
  type: NodePort
  ports:
  - name: http
    port: 80
    protocol: TCP
  selector:
    project: app-name
    role: web
```

When deploying, after the Docker build is successful, this is the error in the Samson console:

```
[04:29:43] JobExecution failed: Service name for role web was generated and needs to be changed before deploying.
[04:29:43] plugins/kubernetes/app/models/kubernetes/template_filler.rb:103:in `generate_service_name'
[04:29:43] plugins/kubernetes/app/models/kubernetes/template_filler.rb:91:in `set_service_name'
[04:29:43] plugins/kubernetes/app/models/kubernetes/template_filler.rb:25:in `to_hash'
[04:29:43] plugins/kubernetes/app/models/kubernetes/release_doc.rb:77:in `block in store_resource_template'
[04:29:43] plugins/kubernetes/app/models/kubernetes/release_doc.rb:75:in `map'
[04:29:43] plugins/kubernetes/app/models/kubernetes/release_doc.rb:75:in `store_resource_template'
[04:29:43] plugins/kubernetes/app/models/kubernetes/release.rb:80:in `block (2 levels) in create_release_docs'
[04:29:43] plugins/kubernetes/app/models/kubernetes/release.rb:79:in `each'
[04:29:43] plugins/kubernetes/app/models/kubernetes/release.rb:79:in `block in create_release_docs'
[04:29:43] plugins/kubernetes/app/models/kubernetes/release.rb:78:in `each'
```

I don't understand, what does `Service name for role web was generated and needs to be changed before deploying` mean? This is deployed with a previous version of Samson and there is no problem.

Thank you in advance :)

grosser:
check the kubernetes roles UI ... the name there will have `generated-` ... replace that with something real like `my-app-server`

klaus993:
Thank you! I had not seen that, sorry. So it's only a matter of names :)

PR #2233, opened 2017-09-08
make access token a bit more obvious and prevent users from clicking …

grosser:
…on the oauth application

old view was misleading users to think the app is the link to the token ... hopefully better now ...
screen shot 2017-09-08 at 11 28 13 am

screen shot 2017-09-08 at 11 34 49 am

making the token bold so it stands out more and is not overlooked as yet another flash:
screen shot 2017-09-08 at 11 32 05 am


bcolferzd:
+1

A much clearer identification of what to do with the token.

PR #2241, opened 2017-09-13
update kubernetes statefulset OnDelete with patch magic

grosser:
this does not allow updating annotations etc, but for everything else like env/limits/requests/replicase/image it works fine ...

@zendesk/paas
/cc @jacobat @bquorning @libo 

... actually worked :D 
https://samsontest.zende.sk/projects/inbox-memcached-k8s/deploys/4685


grosser:
nope there is no "fixing" ... but we can use RollingUpdate on 1.7 :)

grosser:
dried up the weird pod-selector logic too ... much better now :)

PR #2242, opened 2017-09-13
let users pick a updateStrategy for statefulsets since they default i…

grosser:
…s bad

OnDelete might be supported once I get https://github.com/zendesk/samson/pull/2241 working, but it's kinda wonky ... so nudge them into RollingUpdate

@jonmoter @irwaters 

/cc @jacobat  @libo

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_verifier.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_verifier.rb
@@ -30,6 +30,7 @@ module Kubernetes
       verify_numeric_limits
       verify_project_and_role_consistent
       verify_stateful_set_service_consistent
+      verify_stateful_set_restart_policy
       verify_annotations || verify_prerequisites
       verify_env_values
       verify_host_volume_paths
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/role_verifier.rb
```diff
@@ -146,11 +147,22 @@ def verify_project_and_role_consistent
...
       return unless service = @elements.detect { |t| t[:kind] == "Service" }
-      return unless set = @elements.detect { |t| t[:kind] == "StatefulSet" }
+      return unless set = find_stateful_set
       return if set.dig(:spec, :serviceName) == service.dig(:metadata, :name)
       @errors << "Service metadata.name and StatefulSet spec.serviceName must be consistent"
     end
 
+    def verify_stateful_set_restart_policy
+      return unless set = find_stateful_set
+      return if set.dig(:spec, :updateStrategy)
```
libo: I am not sure this matches our yml file structure. https://github.com/zendesk/help-center-memcached-k8s/blob/a0790e62a4cbda6fc52172f1ae79154d6d385679/kubernetes/memcached.yml#L20

Honestly, I rather have @jacobat and @dasch opinion here. Not sure if we use `OnDelete` or `RollingUpdate `

dasch:
This only seems to be supported in k8s 1.7. But yeah, RollingUpdate seems to be the one we want.

plugins/kubernetes/app/models/kubernetes/role_verifier.rb
dasch: It looks :+1: to me.

grosser: the default is OnDelete, that's why I'm calling it out explicitly

PR #2250, opened 2017-09-16
allow selecting no dockerfile to deploy containers with pre-selected …

grosser:
…images

@zenhao @jonmoter
/cc @irwaters

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -162,6 +162,7 @@ samson will then override the `project` labels and keep deployments/services uni
 
  - Set the projects `dockerfiles` attribute to all the files that need to be built.
  - Add `samson/dockerfile: Dockerfile.foobar` to the container configuration (same level as `image`) to use a different Dockerfile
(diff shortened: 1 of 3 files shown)


grosser:
We already allow this if the project has no dockerfile ... But maybe we
should only allow for whitelisted repos or still build bogus images that
only have a from ... Not sure ...

irwaters:
@jonmoter what are your thoughts on this?  Pretty sure that sec-ops is going to want everything scanned, but that's just like my opinion man.

jonmoter:
Yeah, now that I think about it some more, we should move to a pattern where we only deploy docker images from our registry (ECR, GCR, Artifactory, whatever) that Samson is aware of. So let's not merge this change.

grosser:
I think this is still a good usecase for others / is consistent with the current "no Dockerfile present logic" ... so we should still merge this but also add a "Only allow images built by samson" or "only allow builds from these registries" feature.

So preferably merge this and build the whitelist feature after ... alternatively keep this open and build the whitelist feature then merge this.

PR #2251, opened 2017-09-18
try running Datadog's APM for performance monitoring

jonmoter:
Add the gem to `Gemfile`, but only load the gem if `ENABLE_DATADOG_APM` env var is present.

/cc @zendesk/samson

### Risks
- Level: Low


.env.example
```diff
@@ -80,6 +80,9 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # NEW_RELIC_APP_NAME: Samson
 # NEW_RELIC_LOG_FILE_PATH: STDOUT
 
+## Datadog APM
+# ENABLE_DATADOG_APM=true
```
grosser: would be nice to keep all related settings together when printing the ENV ... so DATADOG_APM_ENABLED=true

Gemfile
```diff
@@ -45,6 +45,7 @@ gem 'concurrent-ruby'
 gem 'vault'
 gem 'docker-api'
 gem 'newrelic_rpm'
+gem 'ddtrace', require: false # datadog APM
```
grosser: no need for`require: false`, samson only loads the `preload` group

PR #2256, opened 2017-09-22
need more room for longer env var values

grosser:
got some public keys in there now and 255 is not cutting it :(


jonmoter:
If someone tries to paste something longer than the string limit, do they get an error in the UI?

grosser:
worked on localhost but not the servers ... working on that in parallel ... something fishy about strict mode

jonmoter:
As long as we're at it, want to give a little more room, like 2048?

Using a `TEXT` field is probably overkill...

grosser:
updated

PR #2260, opened 2017-09-24
Add sign in with bitbucket

wafcio:
User can sign in with Bitbucket. This feature will allow to add full bitbucket support (#1)
screen shot 2017-09-24 at 18 38 01


grosser:
❤️

PR #2263, opened 2017-09-25
better names for Datadog logging

jonmoter:
By default Datadog creates names like "mysql", "rails-cache", and "rails-controller". Give them names specific to Samson.

service_list__env_production____datadog

/cc @zendesk/samson, @craig-day 

### Risks
- Level: Low

### DIFF
--- a/config/initializers/datadog_apm.rb
+++ b/config/initializers/datadog_apm.rb
@@ -4,9 +4,17 @@
 if ENV.fetch('DATADOG_ENABLE_APM', 'false') != 'false'
   require 'ddtrace'
 
+  Rails.logger.info("Enabling Datadog APM tracer")
+
   Rails.configuration.datadog_trace = {
     auto_instrument: true,
     default_service: 'samson',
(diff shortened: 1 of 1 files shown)


config/initializers/datadog_apm.rb
```diff
@@ -4,9 +4,17 @@
...
+  Rails.logger.info("Enabling Datadog APM tracer")
+
   Rails.configuration.datadog_trace = {
     auto_instrument: true,
     default_service: 'samson',
-    trace_agent_hostname: ENV.fetch('STATSD_HOST', 'localhost')
+    default_controller_service: 'samson-rails-controller',
+    default_cache_service: 'samson-cache',
+    default_database_service: 'samson-mysql',
+    trace_agent_hostname: ENV.fetch('STATSD_HOST', 'localhost'),
```
grosser: seems a bit odd that there is no `trace_agent_port`

jonmoter: There is, I'm just not overriding it. I could add that line for completeness.

PR #2264, opened 2017-09-25
show cancel button for deploys that are queued or invalid status

grosser:
saw a few cases where threads just died and then there was no way of cancelling the job
 - docker builds
 - deploy timing out and getting killed while being hung


vanchi-zendesk:
👍 Hey nice.

PR #2268, opened 2017-09-28
TOOLS-1907 expose project.repo_path to api users so they can find mat…

grosser:
…ching projects without having to parse urls

 - unifying the api to not expose project tokens or deleted_at
 - unifying user_repo_path + github_repo + gitlab_repo
 - naming repository_url and repository_path similarly

@jonmoter @irwaters 

### Risks
 - Low: breaking usage for people fetching project tokens via show endpoint ... that was never intended


jonmoter:
For this repo, would `repository_path` be `zendesk/samson`?

So presumably if an end-user wanted to distinguish Github org from repo name, they could just split on forward-slash?

grosser:
yes

PR #2269, opened 2017-09-28
TOOLS-1934 set preStop hook to prevent 502s during reboots because se…

grosser:
…rvers are stale for a few seconds

@jonmoter @irwaters @ilkkao @smo921 @libo

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -34,6 +34,7 @@ module Kubernetes
           set_replica_target unless ['DaemonSet', 'Pod'].include?(kind)
 
           make_stateful_set_match_service if kind == 'StatefulSet'
+          set_pre_stop if kind == 'Deployment'
 
           set_name
           set_deployer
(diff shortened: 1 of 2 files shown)


libo:
Given that these 3 seconds do not solve the problem, we still see 502 during deploys, I am not sure if it's worth probably to wait to merge...

grosser:
@libo is anyone investigating this issue ? / when can I expect an update ?

libo:
We are seeing 502 also outside deployments so we would like to see that solved first to better understand what this is..

libo:
I added a 10 sec today and this reduced a bit:
screen shot 2017-10-04 at 17 55 31

But this didn't removed completely. We could try 30sec but I am wondering if we had a better way then waiting? 

I also notice a lot of `kube_stats.pod_unhealthy` during deployments. I don't understand if it's supposed to be like this or symptomatic of something wrong @jonmoter ?
screen shot 2017-10-04 at 17 57 27

grosser:
I think pod_unhealthy is expected during restart since the health checks of
the new pods would fail for a while

grosser:
do 25s just to see if it is completely gone, but I doubt it helps ... any
theory on why it keeps happening ?

PR #2271, opened 2017-09-29
Improve project search

cakejelly:
Currently, searching for projects behaves inconsistently between mysql and
postgres based deployments because the SQL query is hardcoded to use `LIKE`.
This results in search being case sensitive for postgres but case insensitive
for mysql.

This PR:
- makes the project search scope case insensitive for postgres by extending it
  to use the appropriate SQL syntax, based on the DB being used. E.g. `ILIKE`
  for postgres, otherwise `LIKE`.
- makes the project search scope fuzzier by splitting the query argument into
  multiple strings and adding a separate expression to the SQL query for each

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -107,8 +107,7 @@ class ProjectsController < ApplicationController
     per_page = 9 # 3 or 1 column layout depending on size
     scope = Project.alphabetical
     if query = params.dig(:search, :query).presence
-      query = ActiveRecord::Base.send(:sanitize_sql_like, query)
-      scope = scope.where('name like ?', "%#{query}%")
+      scope = scope.search(query)
     elsif ids = current_user.starred_project_ids.presence
       # fake association sorting since order by array is hard to support in mysql+postgres+sqlite
       array = scope.all.sort_by { |p| ids.include?(p.id) ? 0 : 1 }
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -45,7 +45,13 @@ class Project < ActiveRecord::Base
       alphabetical
   }
 
-  scope :search, ->(name) { where("name like ?", "%#{name}%") }
+  scope :search, ->(query) {
+    scope = self
+    sanitize_sql_like(query).split(' ').each do |word|
+      scope = scope.where(Project.arel_table[:name].matches("%#{word}%"))
+    end
+    scope
+  }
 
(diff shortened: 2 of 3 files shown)


app/controllers/projects_controller.rb
```diff
@@ -108,7 +108,7 @@ def projects_for_user
     scope = Project.alphabetical
     if query = params.dig(:search, :query).presence
       query = ActiveRecord::Base.send(:sanitize_sql_like, query)
-      scope = scope.where('name like ?', "%#{query}%")
+      scope = scope.search(query)
```
grosser: would prefer the `.search` to handle sanitize_sql_like

app/models/project.rb
```diff
@@ -45,7 +45,15 @@ class Project < ActiveRecord::Base
...
   }
 
-  scope :search, ->(name) { where("name like ?", "%#{name}%") }
+  scope :search, ->(query) {
+    if query.present?
+      query.split(' ').reduce(current_scope) do |scope, word|
+        scope.where(Project.arel_table[:name].matches("%#{word}%"))
+      end
+    else
+      current_scope.none
```
grosser: ... empty query should find all

grosser: would do:

```Ruby
scope = self # maybe current_scope ... not sure what the difference is
query.split(' ').each { |work| scope = scope.where(Project.arel_table[:name].matches("%#{word}%")) }
scope
```

grosser:
... ideally add an `search_ilike` method to `ActiveRecord::Base` and use that throughout the app ... so much duplication atm ... should handle escaping / splitting / case-sensitivity

app/models/project.rb
cakejelly: Fixed :)

cakejelly:
> ideally add an search_ilike method to ActiveRecord::Base

I noticed you already have a `Searchable` concern included in a couple of models. Maybe it would make sense to add the `search_ilike` method there? It would probably need to be a bit more generic too & also accept an array of columns to search against since User already searches against `name` and `email`. What do you think? :)

Also, do you see it being required for this PR to be merged?

grosser:
hmm could also do `foo bar` -> `%foo%bar%` so make the ordering matter ... not sure

PR #2277, opened 2017-10-03
[AHA-1249] Change .env files permission to 0600

ePak:
`.env` files could potentially contain secrets but the default file permission is `0644` which allows the world to read its contents. Changing the default `.env` file permission to `0600`.

/cc @zendesk/samson @zendesk/bunyip-numbat 

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/AHA-1249

### Risks
- Level: Med - Could be a problem if application expects the `.env` file to be readable by other than file owner

### DIFF
--- a/plugins/env/lib/samson_env/samson_plugin.rb
+++ b/plugins/env/lib/samson_env/samson_plugin.rb
@@ -36,6 +36,7 @@ module SamsonEnv
       groups.each do |suffix, data|
         generated_file = "#{base_file}#{suffix}"
         File.write(generated_file, generate_dotenv(data))
+        File.chmod(0600, generated_file)
       end
     end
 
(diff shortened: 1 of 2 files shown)


plugins/env/lib/samson_env/samson_plugin.rb
```diff
@@ -36,6 +36,7 @@ def write_dotenv(base_file, groups)
       groups.each do |suffix, data|
         generated_file = "#{base_file}#{suffix}"
         File.write(generated_file, generate_dotenv(data))
+        File.chmod(0600, generated_file)
```
grosser: this would leave the file readable for a short time, prefer using atomic: `File.write generated_file, generate_dotenv(data), 0, perm: 0600`

grosser:
good catch, should be safe to change

ePak:
💇‍♂️ @grosser

PR #2280, opened 2017-10-05
fix sync with lots of elements blowing up

grosser:
`NoMethodError: undefined method `[]' for #<Enumerator:`

https://zendesk.airbrake.io/projects/95346/groups/2049537181500252652

### DIFF
--- a/lib/samson/parallelizer.rb
+++ b/lib/samson/parallelizer.rb
@@ -3,9 +3,11 @@ module Samson
   class Parallelizer
     class << self
       def map(elements, db: false)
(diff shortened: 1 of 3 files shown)


lib/samson/parallelizer.rb
```diff
@@ -3,9 +3,11 @@ module Samson
   class Parallelizer
     class << self
       def map(elements, db: false)
+        raise ArgumentError, "argument must be arrayish" unless elements.respond_to?(:[])
+
         max = elements.size
         return [] if max.zero?
-        return [yield(elements.first)] if max == 1
+        return [yield(elements[0])] if max == 1
```
grosser: using consistent api

PR #2282, opened 2017-10-05
[WIP] Experiment some more with Datadog's APM

jonmoter:
Don't want to merge quite yet. This is a based on some pairing and experimentation I did with the Datadog APM team.

/cc @craig-day, @adam-hart, @grosser 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


config/initializers/datadog_apm.rb
```diff
@@ -12,9 +18,27 @@
...
+  pin = Datadog::Pin.get_from(::Dalli)
+  pin.service = 'samson-memcached'
+
+  # Faraday instrumentation
+  Datadog::Monkey.patch_module(:faraday)
+  pin = Datadog::Pin.get_from(::Faraday)
+  pin.service = 'samson'
+
+  # SuckerPunch instrumentation
+  require 'sucker_punch'
```
grosser: ... I'll get rid of sucker-punch ... should never have been added :(

grosser: ... or maybe our job-execution should use sucker-punch ... idk ... but having 2 systems is weird :D

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -116,7 +116,21 @@ def delete_pods(pods)
       end
 
       def request(method, *args)
-        client.send("#{method}_#{@template.fetch(:kind).underscore}", *args)
+        resource_name = @template.fetch(:kind).underscore
+        client_method = "#{method}_#{resource_name}"
+
+        if DatadogAPM.enabled?
```
grosser: use a helper to remove duplication and make it more readable ?
```Ruby
def datadog_trace(key)
  if DatadogAPM.enabled?
    ...
    yield
  else
    yield
  end
end
```

jonmoter:
Not moving forward with this for now. Closing.

jonmoter:
@dragonfax, @ragurney, since we're dropping New Relic and using Datadog APM, we may want to re-open this PR, or recreate it in a different PR.

PR #2286, opened 2017-10-09
Revert "[AHA-1249] Change .env files permission to 0600 (#2277)"

ePak:
This reverts commit d39bf328174cfa63ccddbf13a45036a476e75b66.

*Only valid if we couldn't resolve the other issue*
The previous PR on changing the default `.env` file permission is causing problem with other tools. Roll back that change.

/cc @zendesk/samson  @zendesk/bunyip-numbat 

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/AHA-1215

### Risks
- Level: Low - reverting previous change


ePak:
Nevermind, we should change the permission to `640`. Closing this.

PR #2288, opened 2017-10-09
consolidate on "M" k8s resource units

craig-day:
The UI asks for memory in "MB", which means Megabytes (1000^2). The template filler would then take that value and apply it in "Mi" which means Mebibytes (1024^2). This means that each app was actually being given just a little more memory than it had asked for.

We should make the template filler consistent with the UI and reflect how people normally thing about units of digital memory. 

/cc @zendesk/samson

### Risks
- Level: Low. There will be a slight change to the k8s release docs.


craig-day:
@grosser I pulled in some other updates because so many of these things are tied together. Can you take another look?

PR #2301, opened 2017-10-19
Sort stages list alphabetically

guillaumelecerf:
* Sort stages list alphabetically

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/views/projects/_project.html.erb
+++ b/app/views/projects/_project.html.erb
@@ -7,7 +7,7 @@
       <%= repository_web_link(project) %>
     </h2>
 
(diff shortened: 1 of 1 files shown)


jonmoter:
We don't necessarily want this. You can order the stages in whatever way you want by clicking the "Manage" link on the project home page.

Some teams want to display the stages in the order they occur, which is not necessarily alphabetically.

guillaumelecerf:
I didn't realize that :man_facepalming: 
Thanks.

PR #2302, opened 2017-10-19
POC always use as_json to serialize json responses

grosser:
turns out the funky issue from https://github.com/zendesk/samson/pull/2284
is due to active_model_serializers kicking in when doing `render json: object` but not `render json: {foo: object}`

 - changes outbound webhook api to have a json date format updated_at and a few new columns
 - removes unused deploy objects from sse replies

... guess I'll split this into multiple PRs to make it less scary / more reviewable ...
... could also keep the serializers and just remove associations, but I prefer the "include all" approach of as_json since it means less work to get new columns out ...

@jonmoter 
/cc @zendesk/samson


jonmoter:
So am I correct that this PR doesn't change the format of any of the response JSON objects? It's just a refactoring of code and removing the dependency on `active_model_serializers`?

I want to make sure we're not breaking contract here on any of our endpoints.

grosser:
this changes a bunch of apis slightly ... I'll smaller PRs with detailed breakdowns ... this is more of an overview of everything that needs to get moved to make the transition work ...

atm we use as_json and serializer ... I want to unify that since it's confusing ... 
I'm proposing we go with as_json which would also means new columns get automatically added to the api and whatever way `render` is called it will work out.

alternative would be to use serializers for everything, which means new columns need to be explicitly added to the api and we need to neuter as_json so it does not get used by accident as it has been before

let me know what you prefer ...

grosser:
almost all done now via individual PRs ... time to close 🎉

PR #2330, opened 2017-10-28
wait for external builds to arrive

grosser:
before declared failure instantly ... error was `was created but never ran, run it manually`

@dndungu @dragonfax 
/cc @jonmoter

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/build_finder.rb
+++ b/plugins/kubernetes/app/models/kubernetes/build_finder.rb
@@ -137,24 +137,23 @@ module Kubernetes
     end
 
     def wait_for_build(build)
-      if !build.docker_repo_digest && build.docker_build_job&.active?
-        @output.puts("Waiting for Build #{build.url} to finish.")
-        loop do
-          break if @cancelled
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/build_finder.rb
```diff
@@ -137,24 +137,23 @@ def create_build(dockerfile)
...
 
     def ensure_build_is_successful(build)
       if build.docker_repo_digest
         @output.puts("Build #{build.url} is looking good!")
       elsif build_job = build.docker_build_job
-        raise Samson::Hooks::UserError, "Build #{build.url} is #{build_job.status}, rerun it manually."
+        raise Samson::Hooks::UserError, "Build #{build.url} is #{build_job.status}, rerun it."
       else
-        raise Samson::Hooks::UserError, "Build #{build.url} was created but never ran, run it manually."
+        raise Samson::Hooks::UserError, "Build #{build.url} was created but never ran, run it."
```
grosser: removing 'manually' since that is not true for external builds

PR #2333, opened 2017-10-29
cleanup JobQueue

grosser:
There can be only 1 queue and it is used by multiple jobs,
so time to better represent that.

@dragonfax


app/models/job_execution.rb
```diff
@@ -4,12 +4,6 @@
 class JobExecution
```
dragonfax: Is this class even used? You seem to replace all references with JobQueue

grosser:
it sure is :)
... only replacing all usages that were redirected to JobQueue

Issue #2321, opened 2017-10-26
Migration fails due to build_id column missing


Ngo-The-Trung:
```
Mysql2::Error: Can't DROP 'build_id'; check that column/key exists: ALTER TABLE `releases` DROP `build_id`
/usr/local/bundle/gems/mysql2-0.4.5/lib/mysql2/client.rb:120:in `_query'
/usr/local/bundle/gems/mysql2-0.4.5/lib/mysql2/client.rb:120:in `block in query'
/usr/local/bundle/gems/mysql2-0.4.5/lib/mysql2/client.rb:119:in `handle_interrupt'
/usr/local/bundle/gems/mysql2-0.4.5/lib/mysql2/client.rb:119:in `query'
/usr/local/bundle/gems/activerecord-5.1.0/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:214:in `block (2 levels) in execute'
```

This seems to be from `20170824174718_remove_build_from_releases.rb` but I can't find the migration script where the column was first created.

@trungnn @hyperair @grosser

grosser:
was added in db/migrate/20150508142111_add_build_references.rb

you can run migrations manually from the console to fix weird bugs like that too
 - paste migration
 - `AddBuildReferences.migrate(:up)`

Ngo-The-Trung:
Hi, I'll try to reproduce the correct error for that issue later, although there are other issues with the migration:

https://github.com/zendesk/samson/blob/master/db/migrate/20170824174718_remove_build_from_releases.rb
This touches the kubernetes_release table which should belong to the plugin.

20170922174850_expand_env_vars.rb
This touches the environment_variables table which should belong to the plugin 'env'.

grosser:
Good find I'll fix em up :)

Ngo-The-Trung:
Thanks!

grosser:
fixed by https://github.com/zendesk/samson/pull/2322

grosser:
(when it gets merged ... a few hours ... so maybe use the branch for now or just wait)

PR #2336, opened 2017-10-30
Add travis test with no plugins

hyperair:
This should help to catch issues where migrations with plugin-specific
tables[1][2] end up inside the main migrations folder.

[1] https://github.com/zendesk/samson/pull/2129
[2] https://github.com/zendesk/samson/issues/2321

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


grosser:
👏 ❤️ so smart :D

PR #2340, opened 2017-10-30
remove build validation since not all builds are used for a deploy

grosser:
@jonmoter


jonmoter:
related to https://zendesk.atlassian.net/browse/TOOLS-2145

jonmoter:
Should we add a different validation in this place?  We need to ensure that before we deploy, the Builds that will be used were all successfully built and pushed.

grosser:
if a build has no digest the deploy will blow up since image will be `nil` and the prior checks will also fail ... should all be good

PR #2345, opened 2017-10-31
make external_id truly unique to avoid any kind of issues or hacks

grosser:
we use it for auth so better be carful ...

```
== 20171031104602 MakeExternalIdUnique: migrating =============================
Updating user 3 external_id
-- remove_index(:users, [:external_id, :deleted_at])
   -> 0.0507s
-- add_index(:users, [:external_id, :deleted_at], {:unique=>true})
   -> 0.0297s
== 20171031104602 MakeExternalIdUnique: migrated (0.1576s) ====================
```

@dragonfax

### DIFF
--- a/db/migrate/20171031104602_make_external_id_unique.rb
+++ b/db/migrate/20171031104602_make_external_id_unique.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+class MakeExternalIdUnique < ActiveRecord::Migration[5.1]
+  class User < ActiveRecord::Base
+  end
+
+  def up
+    # update the external_id of all uses that would violate the new uniqueness
+    scope = User.where(deleted_at: nil)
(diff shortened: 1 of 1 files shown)


db/migrate/20171031104602_make_external_id_unique.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+class MakeExternalIdUnique < ActiveRecord::Migration[5.1]
+  class User < ActiveRecord::Base
```
dragonfax: Whats this about again?

Is this to skip all the validation? Or are we not loading the rails models for some reason?

grosser: loading rails models in migrations is dangerous since migrations will run at a random future version of the codebase ... so the code might have changed to use different sorting/scopes or depend on certain columns to already be there

dragonfax:
Gotcha, thanks for the explanation.

PR #2351, opened 2017-10-31
[TOOLS-2157] fix thread contention issue during docker builds

jonmoter:
A thread in the Kubernetes::BuildFinder class was calling #reload on an
instance of build that was being used by the DockerBuilderService class,
and the state of that object was being discarded. This fix safely reloads
the data from the DB without mutating the object, at least while it's
in an active state.

The bug was introduced in #2330. In particular, the changes to the `#wait_for_build` method.

/cc @zendesk/samson, @grosser

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/TOOLS-2157

### Risks
- Level: Med

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/build_finder.rb
+++ b/plugins/kubernetes/app/models/kubernetes/build_finder.rb
@@ -137,14 +137,21 @@ module Kubernetes
     end
 
     def wait_for_build(build)
-      return unless build.reload.active?
+      # We reload the build object from the DB each time we check this, because we want
+      # to know if our instance is stale. But we don't want to keeping calling
+      # build.reload, because that may mutate the object that's being used in another
+      # thread.
+      return unless Build.find(build.id).active?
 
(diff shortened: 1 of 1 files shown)


jonmoter:
If anyone has a brilliant idea on how to add a regression test for this issue, I'm all ears.

jonmoter:
For the record, the code being executed in another thread was the [DockerBuilderService#push_images_to_registries](https://github.com/zendesk/samson/blob/c3fb22ed7b70a9aee06330869c881f02fc14e80c/app/models/docker_builder_service.rb#L187-L222) method.

Inside the loop in that method, I saw examples where the first time it was executed, `build.docker_image.tag` worked, but on the next execution, `build.docker_image` was `nil`. I'm pretty sure that's because that `build` object had `reload` called on it in the other thread.

jonmoter:
Closed in favor of #2352.

PR #2354, opened 2017-11-01
move api/deploys/api_count to regular controller

grosser:
@dragonfax


grosser:
I like the simplicity of doing `/deploys` -> `/deploys.json` ... also makes usage easy and intuitive ... duplicating controller logic was not fun either ... generally like how cerebro truned out and trying to replicate that :)

dragonfax:
makes sense. Thanks for the answer.

PR #2355, opened 2017-11-01
use Arel#matches to get consistent ilike behavior on all databases

grosser:
followup to https://github.com/zendesk/samson/pull/2271

@dragonfax 
/cc @zendesk/samson @cakejelly


grosser:
it's not about style ... I like the pure sql too ... but a `like` is postgres is case-sensitive ... and mysql does not have `ilike` :(

dragonfax:
Makes sense.

PR #2359, opened 2017-11-03
move api/deploys/:id to regular controller

grosser:
this will change the api to no longer automatically nest user/project/stage
they have to be requested via `?includes=` now

@dragonfax 
/cc @zendesk/samson

### DIFF
--- a/app/controllers/concerns/json_renderer.rb
+++ b/app/controllers/concerns/json_renderer.rb
@@ -6,7 +6,6 @@ module JsonRenderer
   class << self
     # make sure we are not including sensitive/unsupported associations
     def validate_includes(requested, allowed)
-      return if allowed == :all
       forbidden = (requested - allowed.map(&:to_s))
       return if forbidden.empty?
       raise(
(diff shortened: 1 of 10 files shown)


app/controllers/concerns/json_renderer.rb
```diff
@@ -6,7 +6,6 @@ class ForbiddenIncludesError < RuntimeError
   class << self
     # make sure we are not including sensitive/unsupported associations
     def validate_includes(requested, allowed)
-      return if allowed == :all
```
grosser: deploy.user is no association ... so we cannot allow `all` includes to be called ... they could be things like `delete_all`

app/controllers/deploys_controller.rb
```diff
@@ -66,8 +66,11 @@ def create
       end
 
       format.json do
-        status = (@deploy.persisted? ? :created : :unprocessable_entity)
-        render json: @deploy.to_json, status: status, location: [current_project, @deploy]
+        if @deploy.persisted?
+          render json: @deploy.to_json, status: :created, location: [current_project, @deploy]
+        else
+          render_json_error 422, @deploy.errors
```
grosser: rendering a `new` deploy does not help the requesting user much, breaks url/status/summary rendering and it does not show the errors either ... so changing to make it the same as the `rescue_from ActiveRecord::RecordInvalid`

app/models/deploy.rb
```diff
@@ -201,6 +204,13 @@ def csv_line
     ]
   end
 
+  def as_json(options = {})
+    raise unless options.empty?
+    hash = super(methods: [:status, :url, :production])
+    hash["summary"] = summary_for_timeline
+    hash
+  end
```
grosser: same as the `DeploySerializer` it replaces ... DeploySerializer will be removed later

PR #2367, opened 2017-11-07
Make SMTP configurable by environment vars

### LINKED ISSUE #2327: Cannot configure SMTP settings in Docker deployment
Hi Samson team,

We are deploying Samson in Kubernetes using the official image. I have just realised that in this scenario Samson is unable to send emails, e.g. an attempt to Bypass a Buddy Check results in:

```
[06:40:29] JobExecution failed: Connection refused - connect(2) for "localhost" port 25
```

I've looked but there seem to be no environment vars that would allow us to configure the SMTP service without modifying code. What would you recommend in this situation?

We could use a sidecar SMTP container as a workaround, but that's extra hassle. I'm not too proficient with Rails, but presumably tweaking https://github.com/zendesk/samson/blob/master/config/initializers/smtp.rb to make some additional `smtp_settings` configurable by ENV, would that be the preferred way?

Thanks,
Anthony

elblivion:
Closes #2327.

* Adds an `SMTP_URL` with a default of `smtp://localhost:25`
* Credentials can be specified in URL or via `SMTP_USER` & `SMTP_PASSWORD` to enable populating from secret storage.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/config/initializers/smtp.rb
+++ b/config/initializers/smtp.rb
@@ -1,5 +1,16 @@
 # frozen_string_literal: true
+require 'uri'
+
+smtp_url = URI.parse(ENV["SMTP_URL"] || "smtp://localhost")
+smtp_url.port ||= 25
+smtp_url.user ||= ENV["SMTP_USER"]
+smtp_url.password ||= ENV["SMTP_PASSWORD"]
+
 ActionMailer::Base.smtp_settings = {
+  port:                 smtp_url.port,
+  address:              smtp_url.host,
+  user_name:            smtp_url.user,
+  password:             smtp_url.password,
   authentication:       'plain',
   enable_starttls_auto: false,
   openssl_verify_mode:  'none'
--- a/.env.example
+++ b/.env.example
@@ -47,6 +47,11 @@ GITHUB_TOKEN=
 # BITBUCKET_KEY=
 # BITBUCKET_SECRET=
 
+## SMTP settings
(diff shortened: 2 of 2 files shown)


elblivion:
Should any tests be added for this, if so where?

config/initializers/smtp.rb
```diff
@@ -1,6 +1,22 @@
...
+
+smtp_settings = {
+  port:                 smtp_url.port,
+  address:              smtp_url.host,
   authentication:       'plain',
   enable_starttls_auto: false,
   openssl_verify_mode:  'none'
 }
+
+if smtp_url.user && smtp_url.password
```
elblivion: I added way more `if`s than I'd have liked, mostly because I don't want to make any assumptions on how ActionMailer will treat `nil` credentials.

grosser: nil works jus fine, tried it just now ... so ifs can go away!

grosser: and can directly assign then too

elblivion:
Build seems to be failing due to long-running tests not necessarily related... will push again later.

grosser:
I'll just merge, got to fix that anyway :D

grosser:
thx!

elblivion:
🙇

PR #2371, opened 2017-11-08
TOOLS-2166 use docker cache from argument to speed up builds

grosser:
needs docker v1.13+

seems to work:

```
[02:42:37] » docker build -f Dockerfile . --cache-from foobar.com/samson_non_prod/kube_service_watcher@sha256:e6a1e740f2b6af9fdc6442d744a75c6c7f717b50c4a184a34b1b43be2d195d78
```

@jonmoter

### DIFF
--- a/config/initializers/docker.rb
+++ b/config/initializers/docker.rb
@@ -17,4 +17,19 @@ if !Rails.env.test? && !ENV['PRECOMPILE'] && ENV['DOCKER_FEATURE']
     warn "Unable to connect to docker!"
     Airbrake.notify($!)
   end
+
+  # ensure that --cache-from is supported (v13+)
+  min_version = 13
+  begin
+    Timeout.timeout(1) do
+      local = Integer(`docker -v`[/Docker version (\d+)/, 1])
+      server = Integer(Docker.version.fetch("Version")[/\d+/, 0])
+      if local < min_version || server < min_version
+        raise Docker::Error::VersionError, "Expected docker version to be >= #{min_version}"
+      end
+    end
+  rescue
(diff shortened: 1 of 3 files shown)


config/initializers/docker.rb
```diff
@@ -17,4 +17,16 @@
     warn "Unable to connect to docker!"
     Airbrake.notify($!)
   end
+
+  # ensure that --cache-from is supported (v13+)
+  begin
+    Timeout.timeout(1) do
+      if Integer(`docker -v`[/Docker version (\d+)/, 1]) < 13
```
jonmoter: I think that `docker -v` just returns the version of the client binary.  Which isn't necessarily the same as the docker daemon you're connecting to.

If you run `docker version`, you get info about both the client and the server.

```
jmoter@deploy1:~$ docker version
Client:
 Version:      17.04.0-ce
 API version:  1.28
 Go version:   go1.7.5
 Git commit:   4845c56
 Built:        Mon Apr  3 17:54:19 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.04.0-ce
 API version:  1.28 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   4845c56
 Built:        Mon Apr  3 18:01:50 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

jonmoter: Looks like the `Docker.version` method of the docker-api gem returns the version of the server.

https://github.com/swipely/docker-api#global-calls

grosser:
it will fail if the client is too old too ... so we should validate both

PR #2386, opened 2017-11-20
store secrets in memory and not on disk

grosser:
@zendesk/paas 

same as https://github.com/zendesk/samson_secret_puller/pull/52


grosser:
confirmed working

PR #2396, opened 2017-11-30
avoid having secrets inside of commands readable by ps

grosser:
@jacobat @jonmoter @dragonfax

### DIFF
--- a/test/models/terminal_executor_test.rb
+++ b/test/models/terminal_executor_test.rb
@@ -86,15 +86,19 @@ describe TerminalExecutor do
       refute subject.pid
       refute subject.pgid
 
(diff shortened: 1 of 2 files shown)


test/models/terminal_executor_test.rb
```diff
@@ -86,13 +86,13 @@
...
 
-      Thread.new { subject.execute('sleep 0.1') }
+      t = Thread.new { subject.execute('sleep 0.1') }
 
       sleep 0.05 # wait for execution to start
       assert subject.pid
       assert subject.pgid
 
-      sleep 0.1 # wait for execution to finish
+      t.join # wait for execution to finish
```
grosser: randomly failed even before this change ... that should fix it for good

PR #2397, opened 2017-11-30
do not send emails to empty strings, it makes the mailer sad

grosser:
@dragonfax

### DIFF
--- a/test/models/stage_test.rb
+++ b/test/models/stage_test.rb
@@ -217,6 +217,11 @@ describe Stage do
       stage = Stage.new(notify_email_address: "a@foo.com;b@foo.com ; c@foo.com; ")
       stage.notify_email_addresses.must_equal ["a@foo.com", "b@foo.com", "c@foo.com"]
     end
+
+    it "ignores whitespace" do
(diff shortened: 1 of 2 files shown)


test/models/stage_test.rb
```diff
@@ -217,6 +217,11 @@
       stage = Stage.new(notify_email_address: "a@foo.com;b@foo.com ; c@foo.com; ")
       stage.notify_email_addresses.must_equal ["a@foo.com", "b@foo.com", "c@foo.com"]
     end
+
+    it "ignroes whitespace" do
```
jonmoter: Spelling error.

test/models/stage_test.rb
```diff
@@ -217,6 +217,11 @@
...
+      stage = Stage.new(notify_email_address: "  ")
+      stage.notify_email_addresses.must_equal []
+    end
```
jonmoter: What happens if you try with ";"?  Does it end up with a pair of empty strings?

jonmoter: Nevermind, tested it and it works.

PR #2400, opened 2017-12-01
support --cache-from on GCR

grosser:
first some refactoring ... then adding it ...
on gcr we need to check if the image exists since it might be:
 - deleted
 - some previous provider (internal or ECR) that the use switched to

this is the last todo and GCR should not fully work 🎉 

@jonmoter @dragonfax


grosser:
@jonmoter @dragonfax

Issue #2393, opened 2017-11-25
Global Concurrent Deploys Limit


mpchadwick:
It would be useful to be able to define a global limit for the number of concurrent deploys that Samson will run at any given time. For example, if I define a global limit of 3 and a user attempts to initiate a new deployment while 3 are in progress, Samson would add that deployment to a queue and start it when the next one finishes. Is this something that would be considered for Samson or better handled via a plugin?

grosser:
Sounds useful in general, make a PR so the number is configureable via ENV var and default is unlimited. Might be best to implement this as "max concurrent jobs" in job_queue.rb

Implementation will be a little tricky since atm active jobs enqueue the next jobs when they are done, so you'll need to implement something like "if job is done and there are no other jobs in it's queue then grab a job from another queue that is not currently executing and run that" ... also it should be fair, so no finishing of a single queue but some fifo queue ... gotta be fun :D

PR #2402, opened 2017-12-02
Max Concurrent Jobs

mpchadwick:
An implementation of the feature described in https://github.com/zendesk/samson/issues/2393

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/models/job_queue.rb
+++ b/app/models/job_queue.rb
@@ -55,7 +55,7 @@ class JobQueue
 
     if JobQueue.enabled
       @lock.synchronize do
-        if should_queue_job?(queue)
+        if full?(queue)
           @queue.push('queue': queue, 'job_execution': job_execution)
           false
         else
@@ -123,7 +123,7 @@ class JobQueue
     end
   end
 
-  def should_queue_job?(queue)
+  def full?(queue)
     return true if @executing[queue]
 
     return false if JobQueue.concurrency == 0
--- a/app/models/job_queue.rb
+++ b/app/models/job_queue.rb
@@ -43,7 +43,7 @@ class JobQueue
     if JobQueue.enabled
       @lock.synchronize do
         if full?(queue)
-          @queue.push(queue: queue, 'job_execution': job_execution)
+          @queue.push(queue: queue, job_execution: job_execution)
           false
         else
           @executing[queue] = job_execution
--- a/app/models/job_queue.rb
+++ b/app/models/job_queue.rb
@@ -23,11 +23,24 @@ class JobQueue
   end
 
   def queued?(id)
-    @queue.values.detect { |jes| jes.detect { |je| return je if je.id == id } }
+    @queue.each do |i|
+      if i[:job_execution].id == id
+        return i[:job_execution]
+      end
+    end
+
+    nil
   end
 
   def dequeue(id)
-    !!@queue.values.detect { |jes| jes.reject! { |je| je.id == id } }
+    @queue.each do |i|
+      if i[:job_execution].id == id
+        @queue.delete(i)
+        return true
+      end
+    end
+
+    false
   end
 
   def find_by_id(id)
@@ -41,8 +54,8 @@ class JobQueue
 
     if JobQueue.enabled
       @lock.synchronize do
-        if @executing[queue]
-          @queue[queue] << job_execution
+        if should_queue_job?(queue)
+          @queue.push('queue': queue, 'job_execution': job_execution)
           false
         else
           @executing[queue] = job_execution
@@ -79,7 +92,7 @@ class JobQueue
   private
 
   def initialize
-    @queue = Hash.new { |h, q| h[q] = [] }
+    @queue = []
     @lock = Mutex.new
     @executing = {}
     @threads = {}
@@ -99,6 +112,15 @@ class JobQueue
     end
   end
 
+  def should_queue_job?(queue)
+    return true if @executing[queue]
+
+    limit = ENV['MAX_CONCURRENT_JOBS'].to_i
+    return false if limit == 0
+
+    executing.length >= limit
+  end
+
   def delete_and_enqueue_next(job_execution, queue)
     @lock.synchronize do
       previous = @executing.delete(queue)
@@ -106,12 +128,17 @@ class JobQueue
         raise "Unexpected executing job found in queue #{queue}: expected #{job_execution&.id} got #{previous&.id}"
       end
 
-      if JobQueue.enabled && (next_execution = @queue[queue].shift)
-        @executing[queue] = next_execution
-        perform_job(next_execution, queue)
+      if JobQueue.enabled && !@queue.empty?
+        @queue.each do |i|
+          if @executing[i[:queue]]
+            next
+          end
+          @queue.delete(i)
+          @executing[i[:queue]] = i[:job_execution]
+          perform_job(i[:job_execution], i[:queue])
+          break
+        end
       end
-
-      @queue.delete(queue) if @queue[queue].empty? # save memory, and time when iterating all queues
     end
 
     instrument
@@ -121,7 +148,7 @@ class JobQueue
     ActiveSupport::Notifications.instrument(
       "job_queue.samson",
       threads: @executing.length,
-      queued: @queue.values.sum(&:count)
+      queued: @queue.length
     )
   end
 end
--- a/app/models/job_queue.rb
+++ b/app/models/job_queue.rb
@@ -6,7 +6,7 @@ class JobQueue
   # Whether or not execution is enabled. This allows completely disabling job
   # execution for testing purposes and when restarting samson.
   class << self
-    attr_accessor :enabled, :concurrency
+    attr_accessor :enabled
   end
 
   class << self
@@ -93,7 +93,6 @@ class JobQueue
     @lock = Mutex.new
     @executing = {}
     @threads = {}
-    JobQueue.concurrency = ENV['MAX_CONCURRENT_JOBS'].to_i
   end
 
   # assign the thread first so we do not get into a state where the execution is findable but has no thread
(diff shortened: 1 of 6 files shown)


app/models/job_queue.rb
```diff
@@ -23,11 +24,24 @@ def executing?(id)
   end
 
   def queued?(id)
-    @queue.values.detect { |jes| jes.detect { |je| return je if je.id == id } }
+    @queue.each do |i|
```
grosser: can do `.detect` to avoid the `nil` below

app/models/job_queue.rb
```diff
@@ -41,8 +55,8 @@ def perform_later(job_execution, queue: nil)
 
     if JobQueue.enabled
       @lock.synchronize do
-        if @executing[queue]
-          @queue[queue] << job_execution
+        if should_queue_job?(queue)
+          @queue.push('queue': queue, 'job_execution': job_execution)
```
grosser: use `queue:` and not `'queue':`

app/models/job_queue.rb
```diff
@@ -99,19 +123,32 @@ def perform_job(job_execution, queue)
     end
   end
 
+  def should_queue_job?(queue)
```
grosser: `should_` is kinda redundant since the method is already `?` -> `def queue_job?`

app/models/job_queue.rb
```diff
@@ -76,13 +90,23 @@ def kill(id)
...
+
   private
 
   def initialize
-    @queue = Hash.new { |h, q| h[q] = [] }
+    @queue = []
     @lock = Mutex.new
     @executing = {}
     @threads = {}
+    JobQueue.concurrency = ENV['MAX_CONCURRENT_JOBS'].to_i
```
grosser: this looks wrong ... add a delegate and then `self.concurrency = ` ?

app/models/job_queue.rb
```diff
@@ -99,19 +123,32 @@ def perform_job(job_execution, queue)
...
     @lock.synchronize do
       previous = @executing.delete(queue)
       unless job_execution == previous
         raise "Unexpected executing job found in queue #{queue}: expected #{job_execution&.id} got #{previous&.id}"
       end
 
-      if JobQueue.enabled && (next_execution = @queue[queue].shift)
-        @executing[queue] = next_execution
-        perform_job(next_execution, queue)
+      if JobQueue.enabled && !@queue.empty?
```
grosser: !empty? -> any?

app/views/deploys/active.erb
```diff
@@ -12,6 +12,6 @@
     <%= render 'job_executions', job_executions: active %>
 
     <h2>Queued</h2>
-    <%= render 'job_executions', job_executions: queued %>
+    <%= render 'job_executions', job_executions: JobQueue.grouped_queue %>
```
grosser: this looks suspicious ... should this assign `queue = JobQueue.grouped_queue` somewhere ?

test/models/job_queue_test.rb
```diff
@@ -94,6 +94,23 @@ def locked
       end
     end
 
+    it 'queues a job if max concurrent jobs is hit' do
+      with_job_execution do
+        locked do
+          JobQueue.concurrency = 1
```
grosser: need to unset that in the `ensure`

app/models/job_queue.rb
grosser: how about putting `queue, job_execution` in the queue so these each/detect get simpler ?

grosser: something like `@queue.each do |queue, job_execution|`

app/models/job_queue.rb
```diff
@@ -23,11 +24,24 @@ def executing?(id)
...
+        return i[:job_execution]
+      end
+    end
+
+    nil
   end
 
   def dequeue(id)
-    !!@queue.values.detect { |jes| jes.reject! { |je| je.id == id } }
+    @queue.each do |i|
```
grosser: could be a `!!@queue.reject! { |i| } ` so on delete it would return true

app/models/job_queue.rb
```diff
@@ -41,8 +55,8 @@ def perform_later(job_execution, queue: nil)
 
     if JobQueue.enabled
       @lock.synchronize do
-        if @executing[queue]
-          @queue[queue] << job_execution
+        if should_queue_job?(queue)
```
grosser: could read nicer with `if full?(queue)`

grosser:
logic looks about right ... just a bit rough around the edges :)

I think `@queue` should get a rename ... it's getting confusing :D

app/views/deploys/active.erb
mpchadwick: What do you think about debug accepting a `grouped` argument (default to false) and handle the grouping in `debug`. Then we can call it here by passing true?

mpchadwick:
Still working through these. `@queue` seems like a logical name for me. What would you suggest as a replacement?

app/models/job_queue.rb
mpchadwick: Not really understanding these comments. This change would make `@queue` an array of hashes. I'm interpreting this comment as a suggestion to make it a hash with pairs of queue => job_execution which wouldn't work because there could be multiple entries for each queue. I'm probably misunderstanding something.

app/models/job_queue.rb
mpchadwick: Sorry, but I'm confused by this comment. I did some quick research and it looks like delegate in Ruby is a way to have a method in one class reference a method from another class. In this case I don't see understand what what what would be delegated to who?

app/models/job_queue.rb
mpchadwick: It seems like the easiest way to make this more succinct is to do...

```
@queue.detect  { |i| i[:job_execution].id == id}
```

Then the places that consume `@queued` can be updated to expect the hash with `queue: 'foo', 'job_execution': bar` and extract `job_execution` as needed.

mpchadwick: Actually it seems like it can just be this...

```
@queue.detect  { |i| return i[:job_execution] if i[:job_execution] == id}
```

app/models/job_queue.rb
```diff
@@ -41,8 +42,8 @@ def perform_later(job_execution, queue: nil)
 
     if JobQueue.enabled
       @lock.synchronize do
-        if @executing[queue]
-          @queue[queue] << job_execution
+        if full?(queue)
+          @queue.push(queue: queue, 'job_execution': job_execution)
```
grosser: `'job_execution':` is the same as `job_execution:`

app/models/job_queue.rb
grosser: what is strange here is that initializing a class sets global state

I think we can simplify this to:

```
def concurrency
  ENV['MAX_CONCURRENT_JOBS'].to_i
end
```

and then in the test you can either do `job_queue.expects(:concurrency).returns(1)` or 

```
with_env MAX_CONCURRENT_JOBS: '1' do
 ... tests ...
end
```

app/models/job_queue.rb
```diff
@@ -99,19 +110,32 @@ def perform_job(job_execution, queue)
...
     @lock.synchronize do
       previous = @executing.delete(queue)
       unless job_execution == previous
         raise "Unexpected executing job found in queue #{queue}: expected #{job_execution&.id} got #{previous&.id}"
       end
 
-      if JobQueue.enabled && (next_execution = @queue[queue].shift)
-        @executing[queue] = next_execution
-        perform_job(next_execution, queue)
+      if JobQueue.enabled && @queue.any?
```
grosser: don't need the `@queue.any` since the `.each` will be a noop if it is empty

app/views/deploys/active.erb
grosser: how about debug keeps returning what it returned before: a grouped queue ... then less tests have to change ?

test/support/job_queue_support.rb
```diff
@@ -7,6 +7,7 @@ def with_job_execution
     JobQueue.enabled = true
     yield
   ensure
+    JobQueue.concurrency = 0
```
grosser: don't need a global reset when it is only used in one place

test/models/job_queue_test.rb
grosser: something like:

```
begin
  test
ensure
  ... unset ...
end
```

but as discussed above if you can do this via an instance method then we don't need to change global state at all

mpchadwick:
@grosser I think all the comments have been addressed. Now I just have one test that's failing

I've been pulling my hair over this for the past 30 minutes, but am really struggling understand what is going on. If I'm interpreting the backtrace correctly the test is hanging on the code that executes starting from `with_a_queued_job` on line 180. What doesn't make sense to me is that `with_a_queued_job` is used without any issue in all the other tests starting in the `it 'has a queued job'` block.

Do you have any thoughts about what is happening?

```
------ >>> test/models/job_queue_test.rb 
Run options: --seed 11288
# Running:
...............E
Error:
JobQueue::#perform_later::with queued job#test_0003_does not perform the next job when job execution is disabled:
Maxitest::Timeout::TestCaseTimeout: Test took too long to finish, aborting. To use a debugger, set Maxitest.timeout = false at the top of the test file.
    test/models/job_queue_test.rb:24:in `sleep'
    test/models/job_queue_test.rb:24:in `wait_for_jobs_to_finish'
    test/models/job_queue_test.rb:60:in `locked'
    test/models/job_queue_test.rb:44:in `block in with_a_queued_job'
    test/support/job_queue_support.rb:8:in `with_job_execution'
    test/models/job_queue_test.rb:43:in `with_a_queued_job'
    test/models/job_queue_test.rb:181:in `block (4 levels) in <top (required)>'
bin/rails test test/models/job_queue_test.rb:180
```

mpchadwick:
Actually I figured out the problem with the test. The clear wasn't doing what it needed to do because debug no longer directly exposed the queue. Pushing up a fix. The stack trace was a bit misleading...

grosser:
good job figuring all this stuff out ... the job-queue is a fun place to loose countless hours of productivity :D

mpchadwick:
@grosser thanks for you're review. My company (Something Digital) is a happy user of Samson for our deployments, but it sometimes gets killed by oom-killer with too many jobs running at once (we've found Gulp to particularly be a memory hog). Hope this feature comes in useful for others as well.

grosser:
we `nice` all our jobs, but we still run into issues from time to time, but mostly from a single super-slow job ... that's why we added gcb-builds so all that can happen on gcloud :D

PR #2404, opened 2017-12-05
Show expanded path for secrets when previewing

davidreuss:
Allows easy grok'ing where secrets are actually found for a deployment.

Also, added JSON formatting.

### DIFF
--- a/plugins/env/app/controllers/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/environment_variable_groups_controller.rb
@@ -50,6 +50,11 @@ class EnvironmentVariableGroupsController < ApplicationController
         @project = Project.find(params[:project_id])
         SamsonEnv.env_groups(@project, DeployGroup.all, preview: true)
       end
+
+    respond_to do |format|
+      format.html
+      format.json { render json: { groups: @groups} }
+    end
   end
 
   private
--- a/plugins/env/app/controllers/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/environment_variable_groups_controller.rb
@@ -53,7 +53,7 @@ class EnvironmentVariableGroupsController < ApplicationController
 
     respond_to do |format|
       format.html
-      format.json { render json: { groups: @groups} }
+      format.json { render json: { groups: @groups || [] } }
     end
   end
 
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -57,8 +57,8 @@ class EnvironmentVariable < ActiveRecord::Base
           resolved =
             if preview
               path = resolver.expand_key(secret_key)
-              expanded_path = !path.nil? ? path : "n/a"
-              found ? "#{TerminalExecutor::SECRET_PREFIX}#{expanded_path} ✓" : "#{value} X"
+              expanded_path = !path.nil? ? path : ""
+              "#{TerminalExecutor::SECRET_PREFIX}#{expanded_path}"
             else
               found.to_s
             end
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -56,9 +56,8 @@ class EnvironmentVariable < ActiveRecord::Base
           found = resolver.read(secret_key)
           resolved =
             if preview
-              expanded_secrets = {}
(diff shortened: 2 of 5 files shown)


davidreuss:
@grosser not entirely sure about the code here, you probably have some ideas for cleaning it a bit, end result would be that you get the fully resolved path for a secret in the preview view.

Should ease debugging when secrets aren't correct, as it's pretty quick to determine where a secret comes from for non k8s deploys as well.

wdyt?

plugins/env/app/models/environment_variable.rb
```diff
@@ -56,7 +56,10 @@ def resolve_secrets(project, deploy_group, env, preview:)
           found = resolver.read(secret_key)
```
grosser: we don't need the `value` ... so can make a new `expand_key` method that returns the full path if the key is found ... then show the full path

davidreuss: i opted for seeing how far i could get with existing API’s, but yep, that’d be preferable..

Should that be another PR? otherwise i can add it tomorrow to this one.

davidreuss: .. unless you have time to provide something

grosser:
just add it to this one so it's 1 clean change
glad you got this far, it's nice having some fresh eyes on that code ...
let me know if you find anything confusing

davidreuss:
PTAL :)

plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -50,6 +50,11 @@ def preview
         @project = Project.find(params[:project_id])
         SamsonEnv.env_groups(@project, DeployGroup.all, preview: true)
       end
+
+    respond_to do |format|
+      format.html
+      format.json { render json: { groups: @groups} }
```
grosser: needs a test

plugins/env/app/models/environment_variable.rb
```diff
@@ -56,7 +56,9 @@ def resolve_secrets(project, deploy_group, env, preview:)
...
           resolved =
             if preview
-              found ? "#{value} ✓" : "#{value} X"
+              path = resolver.expand_key(secret_key)
+              expanded_path = !path.nil? ? path : ""
```
grosser: this would be simpler ?
`path = resolver.expand_key(secret_key) || ""`

grosser: also `secret://` seems like an odd response ...

grosser: maybe something like
```Ruby
path = resolver.expand_key(secret_key)
value = path ? "secret://#{path} " : "#{value} X"
```

davidreuss: i'd say a response with arbitrarily picked characters is a worse response than something that tries to resolve the path, and it's nowhere to be found.

A consumer of said response needs to know the prefix `secret://`, and then pick up the path to actually look in vault by taking the rest of that string, which will be empty.

For this model, you'd need special logic where "if the string ends with ` X`, we know there's no secret to be resolved at the other end.

I do know what i prefer, and what i care about is data models/structures -- and not the actual code that munges them, but in this case i kinda need to.

davidreuss:
duh, i’ll fix the failing tests tomorrow

grosser:
updated

davidreuss:
the travis failure at least makes kogical sense to me - why is it nil in local testing?

davidreuss:
and not really sure how to write the test to verify data/structure is as expected

grosser:
I did not have any nils ... what test is failing locally ?

grosser:
```
EnvironmentVariableGroupsController::as a viewer::a json GET to #preview#test_0001_succeeds 
[/home/travis/build/zendesk/samson/plugins/env/test/controllers/environment_variable_groups_controller_test.rb:105]:
Expected [[".pod2", {"X"=>"Y", "Y"=>"Z"}], [".pod-100", {"X"=>"Y", "Y"=>"Z"}], [".pod1", {"X"=>"Y", "Y"=>"Z"}]] to be nil.
```

?

grosser:
`env_group.update_column :id, 1 # need static id` <- ... most likely the reason ...

grosser:
better now ...

davidreuss:
ah, i see

travis is still grumpy though :joy:

PR #2408, opened 2017-12-07
allow searching by what other secrets have this value

grosser:
screen shot 2017-12-07 at 10 59 26 am
screen shot 2017-12-07 at 10 59 39 am

@dragonfax @jonmoter


grosser:
otherwise there is no way to know if it has the same value as another secret when trying to do cleanup

dragonfax:
gotcha, that sheds light onit.

PR #2409, opened 2017-12-08
Augment k8s stages with kubernetes matrix in json responses

davidreuss:


app/controllers/stages_controller.rb
```diff
@@ -22,7 +22,13 @@ def show
       format.html do
         @deploys = @stage.deploys.page(page)
       end
-      format.json { render json: {stage: @stage} }
+      format.json do
+        data = { stage: @stage }
+        if @stage.kubernetes
+          data["kubernetes"] = Kubernetes::DeployGroupRole.matrix(@stage)
```
grosser: should this be part of the stage and not standalone ?
... 'kubernetes' is already a part of the stage, so would need a different name
also this approach won't work for index ...

since nobody else needs this I'd prefer to keep this behind an `?include=kubernetes`

davidreuss: i don't know how to make it part of the stage, but yes - that would be preferable.

I'm also fine with keeping it behind a query param, that i should be able to figure out how to do :)

grosser:
stage = @stage.as_json
if params[:include].to_s.split(",").include?("kubernetes_matrix")
  stage[:kubernetes_matrix] = foo
end
render json: {stage: stage}

davidreuss:
ah, `as_json` was key here, i thought i had to pass baked objects, but i see what's going on now .. thanks. 👍

PR #2410, opened 2017-12-08
Allow filtering env groups by deploy_group

davidreuss:


plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -42,13 +42,14 @@ def destroy
   end
 
   def preview
+    @deploy_groups = !params[:scope].empty? ? DeployGroup.where(permalink: params[:scope]) : DeployGroup.all
```
grosser: prefer a step-by-step approach

```Ruby
@deploy_groups = DeployGroup.all
@deploy_groups = @deploy_groups.where(permalink: params[:scope]) if params[:scope].present?
```

prefer calling it deploy_group so it's easier to add environment or other things for filtering

plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -42,13 +42,14 @@ def destroy
...
     @groups =
       if params[:group_id]
         @group = EnvironmentVariableGroup.find(params[:group_id])
-        SamsonEnv.env_groups(Project.new(environment_variable_groups: [@group]), DeployGroup.all, preview: true)
+        SamsonEnv.env_groups(Project.new(environment_variable_groups: [@group]), @deploy_groups, preview: true)
       else
         @project = Project.find(params[:project_id])
-        SamsonEnv.env_groups(@project, DeployGroup.all, preview: true)
+        SamsonEnv.env_groups(@project, @deploy_groups, preview: true)
```
grosser: also kinda copy-pasty here ... would be nice to extract project to a standalone if/else and then just have 1 `SamsonEnv.env_groups(@project, @deploy_groups, preview: true)`

davidreuss:
.. restarted the failing travis test, that failed because it ran for too long.

yay, for the cloud the \o/

davidreuss:
no idea about failure now, seems unrelated to this change..

grosser:
some comment on what you are using this for would also be nice since atm nothing links to it ...

plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -42,14 +42,17 @@ def destroy
...
-    @groups =
-      if params[:group_id]
-        @group = EnvironmentVariableGroup.find(params[:group_id])
-        SamsonEnv.env_groups(Project.new(environment_variable_groups: [@group]), DeployGroup.all, preview: true)
-      else
-        @project = Project.find(params[:project_id])
-        SamsonEnv.env_groups(@project, DeployGroup.all, preview: true)
-      end
+    @deploy_group = DeployGroup.all
+    @deploy_group = @deploy_group.where(permalink: params[:scope]) if params[:scope].present?
```
grosser: don't name it scope ... and add a test :)

grosser:
errors seem random ... restarted

davidreuss:
i'll be using this as part of the effort for trying to understand which env variables are being used across our projects.. more PR's for that will likely be incoming for that effort, if i can get away with it :)

plugins/env/app/controllers/environment_variable_groups_controller.rb
davidreuss: what should it be named then? it says scope in the UI? and permalink seems anything but descriptive, for something that's not really a link.

grosser:
don't you need some UI for that too ?

grosser:
deploy_group ? ... deploy_group_permalink ?

davidreuss:
No, this is all done in ad-hoc scripts for now, so no need for that -- yet.

`deploy_group` seems fine i'll update the pr with that :)

grosser:
fixed up, waiting for green

PR #2415, opened 2017-12-08
do not pick up random digests from the build output

grosser:
- possibly picking up cache-from
 - possibly picking up digest from earlier build
 - possibly picking up random other output
 - introudced when converting from chunk reading to reading all output in https://github.com/zendesk/samson/pull/2403

@staugaard @dragonfax


grosser:
worst case it will fail the build ... testing it now in staging ... but better then picking up a random digest :D

PR #2416, opened 2017-12-11
Allow expanding k8s template from deploy group role

davidreuss:


plugins/kubernetes/app/controllers/kubernetes/roles_controller.rb
```diff
@@ -43,6 +43,10 @@ def create
   end
 
   def show
+    respond_to do |format|
+      format.html
+      format.json { render json: {role: @role} }
```
grosser: test

plugins/kubernetes/app/controllers/kubernetes/deploy_group_roles_controller.rb
```diff
@@ -28,9 +30,40 @@ def index
...
+    respond_to do |format|
+      format.html
+      format.json do
+        deploy_group_role = @deploy_group_role.as_json
+        if params[:include].present? && params[:include].to_s.split(',').include?("template")
+
+          release = Kubernetes::Release.create(
+            git_sha: params['git_sha'] || DEFAULT_BRANCH,
+            git_ref: params['git_ref'] || DEFAULT_BRANCH,
+            project: Project.find_by_id(@deploy_group_role.project_id),
```
grosser: this looks dangerous ... if I set a specific sha then the ref should not get set to master ?

grosser: `.find` is fine ... blow up when there is no project

plugins/kubernetes/app/controllers/kubernetes/deploy_group_roles_controller.rb
```diff
@@ -28,9 +30,40 @@ def index
...
+      format.json { render json: {deploy_group_roles: @deploy_group_roles} }
+    end
   end
 
   def show
+    respond_to do |format|
+      format.html
+      format.json do
+        deploy_group_role = @deploy_group_role.as_json
+        if params[:include].present? && params[:include].to_s.split(',').include?("template")
```
grosser: don't need `params[:include].present? &&`

plugins/kubernetes/app/controllers/kubernetes/deploy_group_roles_controller.rb
```diff
@@ -28,9 +30,40 @@ def index
...
+            user: current_user,
+            deploy_groups: [DeployGroup.find_by_id(@deploy_group_role.deploy_group)]
+          )
+          release_doc = Kubernetes::ReleaseDoc.new(
+            kubernetes_release: release,
+            kubernetes_role: Kubernetes::Role.find_by_id(@deploy_group_role.kubernetes_role_id),
+            deploy_group: DeployGroup.find_by_id(@deploy_group_role.deploy_group)
+          )
+          deploy_group_role[:template] = release_doc.send(:verification_template)
+        end
```
grosser: that whole chunk can go into a `generate_template` method

grosser: deploy_executor does something similar ... maybe extract as a method if it's almost the same

plugins/kubernetes/app/controllers/kubernetes/deploy_group_roles_controller.rb
```diff
@@ -28,9 +30,40 @@ def index
...
   end
 
   def show
+    respond_to do |format|
+      format.html
+      format.json do
+        deploy_group_role = @deploy_group_role.as_json
+        if params[:include].present? && params[:include].to_s.split(',').include?("template")
+
+          release = Kubernetes::Release.create(
```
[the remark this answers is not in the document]
davidreuss: couldn’t figure out how to do it by just instantiating a new object, if that’s possible i’d be happy with some pointers.

There were some logic around persisting the release, and calling methods with bang, so not sure what that meant.

davidreuss:
thanks for review - i’ll get these addressed, but pointers on building release without saving any objects to storage/db would be appreciated

grosser:
see deploy_executer.rb it does kind of the same logic

davidreuss:
hm, must study it even closer then must’ve missed some bits on first pass of that

davidreuss:
okay, i tried looking into refactoring some of the stuff here, but i didn't find any good way of doing it, i don't see value in trying to cram too much into single methods, when the code here, basically just uses the API, and has no real duplicated logic in it.

There's likely potential for reuse though, but i can't see it, and it'll take me too long to actually understand the relationships between things fully, to get any useful progress on it.

Especially the intersection between `Kubernetes::DeployExecutor.deploy_group_configs()`, and `Kubernetes::DeployGroupRole.matrix()` .. likely has some potential for refactoring, but i probably won't have the time to soak into it.

plugins/kubernetes/app/controllers/kubernetes/roles_controller.rb
davidreuss: 👍

plugins/kubernetes/app/controllers/kubernetes/deploy_group_roles_controller.rb
davidreuss: blow up how? what do i raise? 

didn't find any good example of errors/messages for json responses..

If i just give a bogus id locally, i get an `ActiveRecord::NotFound` exception, but no idea if that is wrapped by something if it's configured for production?

grosser:
notfound is fine, it will be a nice json response

davidreuss:
oh, deploy_group is already an object here

grosser:
the 'template' is a bit misleading since it is only the "partially filled out template for the primary resource" ... so I changed it to 'verification_template' to make it clear that this is special/weird

davidreuss:
i’m cool with that :+1:

PR #2419, opened 2017-12-11
do not print cursor movements, it looks weird

grosser:
@jonmoter @dragonfax 

before:
screen shot 2017-12-11 at 2 40 15 pm

after:
screen shot 2017-12-11 at 2 46 27 pm

http://ascii-table.com/ansi-escape-sequences.php
there is potential for more cleanup of cursor sequences, but this was the ugliest one we saw ...


davidreuss:
this would be nice 👍

grosser:
need a real +1 :D

PR #2421, opened 2017-12-12
 prevent slow tests by blocking git-clone

grosser:
@dragonfax 

local tests timed out a lot because of it hangs on git-clone


grosser:
turns out this is just my laptop being shitty ... process start randomly takes 4s ... idk why ... saw that in another project too ...

grosser:
yeah ... don't want any clones to happen here either way ... but not a nice global fix ...

PR #2424, opened 2017-12-13
Limit depth while git fetch to reduce traffic

mikhailov:
Everytime we fetch git repo, we get the full history while we need only recent commits, that PR helps to reduce traffic and speed up deployments.

Idea has taken from Travis, every Travis build is limiting git history by 50.

/cc @zendesk/samson @DinoRosic 

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: High - can break deployment

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -131,7 +131,7 @@ class GitRepository
   add_method_tracer :create_workspace
 
   def update!
-    executor.execute("cd #{repo_cache_dir}", 'git fetch -p')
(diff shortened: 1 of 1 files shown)


grosser:
Don't think that's an issue since we keep the repo around ... so we'll only fetch new things.
Also travis works on recent commits whereas samson might also work with old commits.

mikhailov:
@grosser alright, got it, closing now!

PR #2425, opened 2017-12-13
use deploy-timeout for gcr builds to have same timeout as local builds

grosser:
- makes the iamge-builder use the executor of the docker build,
so it will actually stop when the build is cancelled
 - uses shelljoin to avoid any kind of command injection or accidental bugs

@jonmoter @dragonfax 
/cc @zenhao


zenhao:
Great! Thanks for the quick turnaround

PR #2427, opened 2017-12-13
avoid builds with nil projects

grosser:
saw a build triggered by a kubernetes deploy that had a nil project,
but when fetching it from DB it looked normal ... so the clone might be to blame

https://zendesk.airbrake.io/projects/95346/groups/2073278695669703679/notices/2103996377796181626?dur=90d%2F1d&tab=notice-detail

@dragonfax 
/cc @dongxiaohe


dongxiaohe:
Thanks for fixing it :)

PR #2428, opened 2017-12-14
TOOLS-2228 link deploys to builds ... also make job commits unique

grosser:
had to bump bootsnap to avoid a weird 'unsupported platform' bug
adding a "search by git sha" UI to deploys to show results and maybe useful in general

@jonmoter @dragonfax 

screen shot 2017-12-13 at 3 41 43 pm
screen shot 2017-12-13 at 4 10 17 pm


app/views/deploys/index.html.erb
```diff
@@ -9,12 +9,17 @@
...
 
-      <% unless @project %>
-        <div class="col-sm-2">
+      <% if @project %>
+        <div class="col-sm-1">
+          <%= label_tag :git_sha %>
+          <%= text_field_tag 'search[git_sha]', params.dig(:search, :git_sha), class: "form-control" %>
+        </div>
+      <% else %>
+        <div class="col-sm-1">
```
grosser: sm-2 was breaking the line ... so had to make that smaller ... maybe we need to have a smarter way for cases with too many options ...

jonmoter:
Does this show all the deploys that involved with a Build?  Or just the live ones?

On the Builds#index page, it would be nice to have a list of which builds are live, like we do on the Releases#index page.

grosser:
it shows all deploys

PR #2439, opened 2017-12-16
 make builds available in environment

grosser:
@dragonfax @jonmoter @davidreuss @craig-day 

screen shot 2017-12-15 at 4 14 19 pm

### DIFF
--- a/app/controllers/stages_controller.rb
+++ b/app/controllers/stages_controller.rb
@@ -133,6 +133,7 @@ class StagesController < ApplicationController
       :cancel_queued_deploys,
       :periodical_deploy,
       :no_reference_selection,
+      :builds_in_environment,
       {
         deploy_group_ids: [],
         command_ids: []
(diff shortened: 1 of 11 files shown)


davidreuss:
I’m slightly confused. What’s the pupose here? Can you describe a usecase that’s using this? 

Lacking context.

davidreuss:
Oh, this is so the build you’ve configured to happen in docker will be available to pre-deploy/build hooks? .. to ease moving parts of your build to be containerized, but still deploying the app to vm’s with artifacts from your containerized build?

jonmoter:
Right.  So for teams that have really expensive `sbt` builds, they do all their expensive build/compile steps in GCB. The docker image contains their compiled artifacts, and Samson is notified about the docker image.

So then their Samson deploy stage could look something like:

```
container=$(docker create $BUILD_FROM_Dockerfile)
docker cp $container:/app/whatever.java .
docker rm $container

cap deploy $DEPLOY_TARGET
```

grosser:
better not leave the container hanging when something goes wrong:
```
docker run --rm --entrypoint cat $BUILD_FROM_Dockerfile  /app/foo.jar > foo.jar
```

grosser:
updated to download builds too and tested on staging ... https://samsontest.zende.sk/projects/kube_service_watcher/deploys/4843

good to go ?

PR #2440, opened 2017-12-17
Alter bin/setup to conditionally load custom config and seed data

jgoerz:
The goal of this commit is to make it so we can spin up from scratch using external configuration  and data files using only bin/setup.

Files modified:
.gitignore
bin/setup
db/seeds.rb

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@
 # Ignore config
 /config/database.yml
 
+# Ignore custom setup files (see bin/setup)
+/custom
+
 /node_modules
 
 /public/assets
--- a/bin/setup
+++ b/bin/setup
@@ -47,9 +47,9 @@ try "Bundling" "bundle check > /dev/null || bundle --local --quiet "
 
 echo
 bold "Copying example files"
-try "Creating .env" "test -e .env || cp .env.bootstrap .env"
-try "Creating config/database.yml" "test -e config/database.yml || cp config/database.mysql.yml.example config/database.yml"
-try "Creating databases" "bundle exec rake db:setup > /dev/null"
+try "Creating .env" "test -e .env || (test -e custom/.env.bootstrap && cp custom/.env.bootstrap .env) || cp .env.bootstrap .env"
+try "Creating config/database.yml" "test -e config/database.yml || (test -e custom/database.yml && cp custom/database.yml config/database.yml) || cp config/database.mysql.yml.example config/database.yml"
+try "Creating databases" "(test -e custom/seeds.rb && cp -f custom/seeds.rb db/seeds.rb) && false || bundle exec rake db:setup > /dev/null"
 
 echo
 bold "Success!"


.gitignore
```diff
@@ -20,6 +20,11 @@
 # Ignore config
 /config/database.yml
 
+# Ignore SETUP_USE_CUSTOM files
+/config/database.mysql.yml.custom
+/.env.bootstrap.custom
+/db/seeds.custom.yml
```
grosser: not a fan of custom ignores ... how about putting it into `tmp/` or something like `custom/` and then ignore that ?

grosser:
how about a `modify - setup - restore` workflow ... then there would be no diff left ?

.gitignore
jgoerz: Yes, either one will work fine. tmp/ or custom/.

jgoerz:
> how about a modify - setup - restore workflow ... then there would be no diff left ?

That would be even better.  That's really what I'm trying to do, have a way to spin up from scratch that doesn't impact the existing "spin the docker container up and see a working example".

.gitignore
jgoerz: I removed the custom gitignore entries.

grosser:
so why not override config/database.yml.example instead of making a new one ?
... same for seeds ... override the db/seeds.rb ?

grosser:
Loading seeds from yml files is kinda interesting (maybe using ActiveRecord::Fixtures would make it more flexible), but it also limits the amount of hackery that users can do, so I'm not that big of a fan

jgoerz:
Let me start off with saying I'm really new to this workflow, so if I dork something up, it's probably ignorance.  I don't mind being corrected, even if your blunt.  I'd actually prefer that.

> so why not override config/database.yml.example instead of making a new one ?
> ... same for seeds ... override the db/seeds.rb ?

We cloned your repo and are actually managing our copy in our on premise Gitlab installation.  So the idea is we will keep syncing with your master and try to make any changes we make as modular as possible so we can hopefully get you to merge them mainline to make our management easier (selfish) and share any improvements (hopefully generous).  With that in mind, I don't want to deal with conflicting files while we are doing sync merges with upstream.  So that was why I was _overriding_ the file by using a completely different one.  If there is a better way, or maybe more correct way, I'd be happy to try and implement that.  I'm not seeing how overriding those files directly would avoid conflicts.  Am I missing something?

jgoerz:
> Loading seeds from yml files is kinda interesting (maybe using ActiveRecord::Fixtures would make it more flexible), but it also limits the amount of hackery that users can do, so I'm not that big of a fan

I can't tell by the way you worded that if you're for or against ActiveRecord::Fixtures.  Would you like me to explore a solution with ActiveRecord::Fixtures?

grosser:
Fixtures was just a suggestion in case you want to do yml loading but be more open to all kinds of fields/models ... not sure it would work out perfectly since linking is kinda hard, but could be a fun project to try.

how about:
 - for .env have a `custom` folder and a .env.bootstrap in there and copy that if it exists during bin/setup
 - for seeds have a `custom` folder and a seeds.rb in there and load that if it exists from db/seeds.rb
 - add the `custom` folder to .gitignore (here or in your repo)

jgoerz:
> for seeds have a custom folder and a seeds.rb in there and load that if it exists from db/seeds.rb

Hmm.  I didn't pay close enough attention to that.  I overwrote the seeds.rb directly.  I can change that.  Let me know if you think I should change the current overwrite implementation to what you specified.

jgoerz:
If I'm ready that failure right, it wants the latest brakeman (4.1.1 vs. 4.1.0).  So this is a CI problem, not a real "code" failure.  fyi, I did a local update of brakeman and it passed.

grosser:
looks good! :)

grosser:
our own setup is that we combine public samson with our internal plugins and then deploy, but either way should work

PR #2441, opened 2017-12-18
make osx command execution fast again

grosser:
osx has a fun startup delay for unknown executables ... we tried a bunch of things but could not track it down, so avoiding it now by having a static executable.

This makes local testing faster and stops breaking any testcase that uses a command.

caused by https://github.com/zendesk/samson/pull/2411

@dragonfax @jonmoter 

/cc @zendesk/samson @craig-day @jcheatham 

before:

```
Benchmark.realtime { TerminalExecutor.new(STDOUT).execute "echo #{rand}" }
=> 3.752056999997876
```

```
Benchmark.realtime { TerminalExecutor.new(STDOUT).execute "echo #{rand}" }
=> 0.012225999998918269
```

### DIFF
--- a/app/models/terminal_executor.rb
+++ b/app/models/terminal_executor.rb
@@ -29,8 +29,8 @@ class TerminalExecutor
   def execute(*commands, timeout: @timeout)
     return false if @cancelled
     options = {in: '/dev/null', unsetenv_others: true}
-    script_as_executable(script(commands)) do |path|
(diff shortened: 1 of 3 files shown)


grosser:
I'd rather use this workaround then tell everyone on osx to do some weird SIP workaround ...

jcheatham:
haha, yeah, don't think we _could_ advise people to intentionally disable SIP here or something, mostly curiosity / matter of diagnosis.

PR #2443, opened 2017-12-19
make path available to avoid python not found erors

grosser:
```
  File "/opt/google-cloud-sdk/lib/googlecloudsdk/core/execution_utils.py", line 60, in GetPythonExecutable
    raise ValueError('Could not find Python executable.')
ValueError: Could not find Python executable.
```

https://github.com/GoogleCloudPlatform/gsutil/issues/402


grosser:
confirmed working in staging

PR #2445, opened 2017-12-19
having a docker release branch but builds disabled makes no sense

grosser:
screen shot 2017-12-19 at 3 53 44 pm

@dragonfax @jonmoter


jonmoter:
Maybe use a dropdown or radio buttons instead of 2 check boxes, if the two are mutually exclusive?

grosser:
it's a text field and a checkobx :D

PR #2447, opened 2017-12-20
disable deploy on release for production stages

grosser:
screen shot 2017-12-19 at 5 29 21 pm

@jonmoter @dragonfax


jonmoter:
Are we going to have a separate check for webhooks that trigger a production stage?

grosser:
yeah will need that too ...

PR #2450, opened 2017-12-21
images get deleted all the time, we do not need to store them

grosser:
@dragonfax @jonmoter

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -32,8 +32,6 @@ class Build < ActiveRecord::Base
 
   before_create :assign_number
 
-  attr_accessor :docker_image_id
-
   def nice_name
     name.presence || "Build #{id}"
   end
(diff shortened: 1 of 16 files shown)


app/models/build.rb
```diff
@@ -26,13 +26,14 @@ class Build < ActiveRecord::Base
...
     )
   end
-  validates :docker_image_id, format: SHA256_REGEX, allow_nil: true
   validates :docker_repo_digest, format: DIGEST_REGEX, allow_nil: true
   validates :external_url, format: /\Ahttps?:\/\/\S+\z/, allow_nil: true
   validates :external_status, inclusion: Job::VALID_STATUSES, allow_nil: true
 
   before_create :assign_number
 
+  attr_accessor :docker_image_id
```
jonmoter: Are you sure we still want to have this as an accessor, if it's not clear what's responsible for setting the value?

grosser: it's not perfect, but a bit better than before ... next refactoring should get rid of it

grosser:
should be done now ... both image-builders have the same interface and DockerBuilderService is only responsible for preparing the build and the storing the digest

PR #2453, opened 2017-12-22
make service update work like any other resource but allow persistent…

grosser:
without these hacks we get:

```
Service "truth-service" is invalid: [metadata.resourceVersion: Invalid value: "": must be specified for an update, spec.clusterIP: Invalid value: "": field is immutable]
```

for example `metadata.labels.proxy,metadata.labels.proxyEnabled`

confirmed working on staging

@jonmoter @dragonfax

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -92,8 +92,13 @@ module Kubernetes
 
       # FYI: do not use result of update call, see https://github.com/abonas/kubeclient/issues/196
       def update
-        request(:update, @template)
+        request(:update, template_for_update)
         expire_cache
+      end
+
+      # hook for subclasses to override
+      def template_for_update
+        @template
       end
 
       def fetch_resource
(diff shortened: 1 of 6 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -152,16 +157,20 @@ class ConfigMap < Base
...
-      def deploy
-        return if running?
-        create
-      end
+      private
 
-      def revert(previous)
-        delete unless previous
+      def template_for_update
+        new = @template.deep_dup
```
jonmoter: generally avoid using `new` as a local variable

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -152,16 +157,20 @@ class ConfigMap < Base
...
+        [
+          "metadata.resourceVersion",
+          "spec.clusterIP",
+          *ENV["KUBERNETES_SERVICE_PERSISTENT_FIELDS"].to_s.split(",")
+        ].each do |keep|
+          path = keep.split(".").map!(&:to_sym)
+          old_value = resource.dig(*path)
+          new.dig_set path, old_value unless old_value.nil? # boolean fields are kept, but nothing is nil in kubernetes
+        end
+        new
```
jonmoter: Would be a good idea to add a comment in here explaining why this is necessary for Service objects, and why the behavior is different than something like a Deployment.

jonmoter:
FYI, if you're interested, I saw a talk at KubeCon about how `kubectl apply` works.

https://www.youtube.com/watch?v=CW3ZuQy_YZw

If you do `kubectl apply -f resource.yml`, it stores that object definition in an annotation of the object itself. So the next time you do an apply, it compares 3 things:

1. The current state of the object
2. The object definition from the last time you did `apply`
3. The new YAML file that you specified

In theory we could do something similar. But it makes the logic pretty complicated.

PR #2454, opened 2017-12-23
update-many kubernetes resources

grosser:
flow to update multiple deploy-group-roles

screen shot 2017-12-28 at 4 54 32 pm

screen shot 2017-12-28 at 4 54 50 pm

@jonmoter @dragonfax 

Alternatives:
 - add a "apply changes to all siblings" box ... that is less work/code, but displaying errors is harder ... 
 - make deploy-group-roles per environment so we don't have that much repetition

### DIFF
--- a/plugins/kubernetes/app/controllers/kubernetes/deploy_group_roles_controller.rb
+++ b/plugins/kubernetes/app/controllers/kubernetes/deploy_group_roles_controller.rb
@@ -1,13 +1,14 @@
 # frozen_string_literal: true
 class Kubernetes::DeployGroupRolesController < ApplicationController
   before_action :find_role, only: [:show, :edit, :update, :destroy]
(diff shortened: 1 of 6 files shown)


jonmoter:
So the "Edit All" link is on the `Stage#show` page?

How about if we created a "Resources" sub-tab inside the Kubernetes tab for a project? That page could show all the DeployGroupRole definitions for the project, and the resources allocated to them. I think it would be nice to have a view of all the resources, not limited by stage.

Then on that page we could have the "Edit All" button on that page, which would display everything.

If we were really fancy, we could also display the calculated Kubernetes Usage Limits on that page, so it's clear what the limits are before they try filling something out.

grosser:
Yeah I like that :)

grosser:
updated, looking good ?

PR #2455, opened 2017-12-26
add autoscaled setting to roles to disable replica updates

grosser:
@craig-day @catkins 
@dragonfax @jonmoter 

screen shot 2017-12-26 at 10 57 36 am

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -7,9 +7,10 @@ module Kubernetes
   # and see what it does internally ... simple create/update/delete requests or special magic ?
   module Resource
     class Base
-      def initialize(template, deploy_group)
+      def initialize(template, deploy_group, autoscaled:)
         @template = template
         @deploy_group = deploy_group
+        @autoscaled = autoscaled
       end
 
       def name
@@ -35,7 +36,7 @@ module Kubernetes
 
       def revert(previous)
         if previous
-          self.class.new(previous, @deploy_group).deploy
+          self.class.new(previous, @deploy_group, autoscaled: @autoscaled).deploy
         else
           delete
         end
@@ -66,7 +67,16 @@ module Kubernetes
         resource&.dig_fetch(:metadata, :uid)
       end
 
+      def desired_pod_count
+        replica_source.dig_fetch(:spec, :replicas)
+      end
+
       private
+
+      # when autoscaling we expect as many pods as we currently have
+      def replica_source
+        (@autoscaled && resource) || @template
+      end
 
       def backoff_wait(backoff, reason)
         backoff.each do |wait|
@@ -96,9 +106,15 @@ module Kubernetes
         expire_cache
       end
 
-      # hook for subclasses to override
       def template_for_update
-        @template
+        copy = @template.deep_dup
+
+        # when autoscaling on a resource with replicas we should keep replicas constant
+        # (not setting replicas will make it use the default of 1)
+        path = [:spec, :replicas]
+        copy.dig_set(path, replica_source.dig(*path)) if @template.dig(*path)
+
+        copy
       end
 
       def fetch_resource
@@ -163,7 +179,7 @@ module Kubernetes
       # we also keep whitelisted fields that are manually changed for load-balancing
       # (meant for labels, but other fields could work too)
       def template_for_update
-        copy = @template.deep_dup
+        copy = super
         [
           "metadata.resourceVersion",
           "spec.clusterIP",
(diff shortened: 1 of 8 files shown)


jonmoter:
So in this implementation, I see it's a property of the DeployGroupRole. Which means that we could use autoscaling for some deploy groups, but not others?

What's the behavior if someone clicks the "autoscaling" box but does not implement an autoscaler?

Another way we could implement it is if there's a `HorizontalPodAutoscaler` object in the role file, then Samson automatically treats that role as using autoscaling.

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -31,8 +31,7 @@ def to_hash
             set_history_limit
           end
 
-          set_replica_target unless ['DaemonSet', 'Pod'].include?(kind)
-
+          set_replica_target unless ['DaemonSet', 'Pod'].include?(template[:kind])
```
craig-day: shouldn't this be checking the autoscale thing and not do it if the role is autoscaled?

grosser: the "set replicas" logic is in resource.rb where the actual update is being done, the template does not know if it is being used for a create or update and I wanted the create to still have the user-defined number of replicas

grosser:
If a user sets autoscale but does not implement one then replicase will never be changed.

Having the autoscaled on the deploy-group-role is nice for a smooth rollout, but could lead to inconsistent state ... so how about we move it to the role ? ... so it's consistent ...

The autoscaler could be external, so I don't want make it required to be in the role file.

grosser:
updated to be configured from role ... looks good ?

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -96,9 +97,12 @@ def update
         expire_cache
       end
 
-      # hook for subclasses to override
       def template_for_update
-        @template
+        copy = @template.deep_dup
+        if @autoscaled && spec = copy[:spec]
+          spec.delete(:replicas)
```
craig-day: should this be `copy`?

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -7,9 +7,10 @@ module Kubernetes
   # and see what it does internally ... simple create/update/delete requests or special magic ?
   module Resource
     class Base
-      def initialize(template, deploy_group)
+      def initialize(template, deploy_group, autoscaled:)
```
craig-day: any backwards compatibility concerns? Should it default to false?

plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -84,6 +85,17 @@ def assert_pod_deletion
       assert_requested get, times: 2
     end
 
+    it "does not set replicase when autoscaled, to not revert autoscaler changes" do
+      resource = Kubernetes::Resource.build(template, deploy_group, autoscaled: true)
+      get = stub_request(:get, url).to_return(body: '{}')
+      update = stub_request(:put, url).with {  |x| x.body.wont_include 'replica'; true }.to_return(body: "{}")
```
craig-day: extra space before the loop argument

plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -74,7 +75,7 @@ def assert_pod_deletion
     it "updates existing" do
       get = stub_request(:get, url).to_return(body: '{}')
 
-      update = stub_request(:put, url).to_return(body: "{}")
+      update = stub_request(:put, url).with {  |x| x.body.must_include 'replica'; true }.to_return(body: "{}")
```
craig-day: extra space here too

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: nope, I want to force every callsite to know about this

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: no, it's spec.replicas

craig-day: oh, I thought this read `spec == copy[:spec]`, but nvm...

grosser:
how about now ?

grosser:
... looks like not setting replicas is setting it to 1 :/

grosser:
added an icon for it on the stage view so it's less confusing ... 
screen shot 2017-12-27 at 3 14 06 pm

PR #2457, opened 2017-12-29
upgrade rubocop to get parallel option (2m -> 18s)

grosser:
@dragonfax


dragonfax:
This seems wrong. Your upgrading rubocop, not changing the code guidelines for the project.
Why are all these changes necessary to get parallel rubocop?

grosser:
updated rubocop includes new cops that are on by default and also changes to existing cops

grosser:
@dragonfax good to go ?

app/controllers/integrations/base_controller.rb
```diff
@@ -132,7 +132,7 @@ def create_docker_images
   end
 
   def record_log(level, message)
-    (@recorded_log ||= "".dup) << "#{level.upcase}: #{message}\n"
+    (@recorded_log ||= +"") << "#{level.upcase}: #{message}\n"
```
dragonfax: This is weird. Is this string concatenation or some other thing?

app/controllers/dashboards_controller.rb
```diff
@@ -6,7 +6,7 @@ def show
     @before = Time.parse(params[:before] || Time.now.to_s(:db))
     @deploy_groups = @environment.deploy_groups.sort_by(&:natural_order)
     @projects = Project.all
-    @failed_deploys = params[:failed_deploys] == "true" ? true : false
+    @failed_deploys = (params[:failed_deploys] == "true")
```
dragonfax: Fair enough. that is a better way to write it.

app/controllers/concerns/current_user.rb
```diff
@@ -31,7 +31,8 @@ def warden
   end
 
   def unauthorized!
-    Rails.logger.warn "Halted as unauthorized! threw :warden (called from #{caller.first.sub(Rails.root.to_s, '')})"
+    called_from = caller(1..1).first
+    Rails.logger.warn "Halted as unauthorized! threw :warden (called from #{called_from.sub(Rails.root.to_s, '')})"
```
dragonfax: What was wrong with the original?

dragonfax:
Are we deciding our own styleguide and using rubocop to enforce it? Or are we just going with whatever arbitrary style-guide the author of rubocop thought everyone should use by default?

app/controllers/integrations/base_controller.rb
grosser: rubocop rule is to prefer `+""` over `"".dup`

app/controllers/concerns/current_user.rb
grosser: `caller(1..1)` is cheaper than `caller`

app/models/csv_export.rb
```diff
@@ -17,7 +17,7 @@ class CsvExport < ActiveRecord::Base
   }
 
   def status?(state)
-    'ready' == state.to_s ? ['downloaded', 'finished'].include?(status) : status == state.to_s
+    state.to_s == 'ready' ? ['downloaded', 'finished'].include?(status) : status == state.to_s
```
dragonfax: What was wrong here? why does the order matter?

app/models/user.rb
```diff
@@ -25,7 +25,7 @@ class User < ActiveRecord::Base
   before_create :set_token
   validates :time_format, inclusion: { in: TIME_FORMATS }
   validates :external_id,
-    uniqueness: {scope: :deleted_at}, presence: :true, unless: :integration?, if: :external_id_changed?
+    uniqueness: {scope: :deleted_at}, presence: true, unless: :integration?, if: :external_id_changed?
```
dragonfax: Thats actually a good catch.

config/application.rb
```diff
@@ -161,7 +161,7 @@ class Application < Rails::Application
       config.after_initialize do
         # Token used to request badges
         config.samson.badge_token = \
-          Digest::MD5.hexdigest('badge_token'.dup + Samson::Application.config.secret_key_base)
+          Digest::MD5.hexdigest('badge_token' + Samson::Application.config.secret_key_base)
```
dragonfax: What was with the dups on hard-coded strings? (I'm asking for myself, not criticizing). Was this some old ruby performance thing that is no longer necessary?

db/migrate/20160413171549_unique_roles.rb
```diff
@@ -5,7 +5,7 @@ class UserProjectRole < ActiveRecord::Base
 
   def up
     # delete the lowest permission duplicate roles
-    UserProjectRole.all.group_by { |role| [role.user_id, role.project_id] }.each do |_, roles|
+    UserProjectRole.all.group_by { |role| [role.user_id, role.project_id] }.each_value do |roles|
```
dragonfax: That's a good catch too, more explicit.

app/models/csv_export.rb
grosser: basic code style of having the variable on the right `a == 1` and not `1 == a`

app/models/csv_export.rb
```diff
@@ -32,7 +32,7 @@ def filters
     filter = super.clone
     if filter['deploys.created_at']
       dates = filter['deploys.created_at'].scan(/-?\d+-\d+-\d+/)
-      filter['deploys.created_at'] = (DateTime.parse(dates[0] + "T00:00:00Z")..DateTime.parse(dates[1] + "T23:59:59Z"))
+      filter['deploys.created_at'] = (Time.parse(dates[0] + "T00:00:00Z")..Time.parse(dates[1] + "T23:59:59Z"))
```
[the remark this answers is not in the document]
grosser: it said no more datatime ... tests seems to still pass, so I think it is working

grosser:
this PR was updating rubocop and applying all style-guides that added by default (I disable things I don't agree with or would be too much work)

test/helpers/projects_helper_test.rb
```diff
@@ -30,7 +29,8 @@
     it 'returns the deployment alert data' do
       job = project.jobs.create!(command: 'cat foo', user: users(:deployer), status: 'failed')
       deploy = stage.deploys.create!(reference: 'master', job: job, project: project)
-      expected_title = "#{deploy.updated_at.strftime('%Y/%m/%d %H:%M:%S')} Last deployment failed! #{deploy.user.name} failed to deploy '#{deploy.reference}'"
+      expected_title = "#{deploy.updated_at.strftime('%Y/%m/%d %H:%M:%S')} Last deployment failed! " \
+        "#{deploy.user.name} failed to deploy '#{deploy.reference}'"
```
dragonfax: That works? 2 strings next to each other with no + in there? I would personally prefer the +. It looks odd otherwise.

grosser:
the .dup here was unnecessary, for the others +"" is more efficient
according to
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Performance/UnfreezeString

dragonfax:
Most of these changes do seem like a good idea.

app/controllers/integrations/base_controller.rb
dragonfax: Yeah, seems faster too. http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Performance/UnfreezeString

Okay. I buy it.

app/controllers/concerns/current_user.rb
dragonfax: Fair enough.

app/models/csv_export.rb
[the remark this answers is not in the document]
dragonfax: Yeah, looks like Time is compatible as long as your not doing some weird calendar stuff, which no one does. https://github.com/bbatsov/rubocop/issues/1952

test/helpers/projects_helper_test.rb
dragonfax: Okay. I learned alot, so in the end, this is the only issue I have with the rubocop default style changes. What do you think about sticking a plus in there?

grosser: works but rubocop disagrees :)
```
test/helpers/projects_helper_test.rb:32:102: C: Style/LineEndConcatenation: Use \ instead of + or << to concatenate those strings.
      expected_title = "#{deploy.updated_at.strftime('%Y/%m/%d %H:%M:%S')} Last deployment failed! " +
```

grosser:
merging to not get more conflicts ... we can revisit disabling/changing style in a separate PR

dragonfax:
yeah, I expected more changes from that. no worries. I'll approve it unless something truly new pops up.

PR #2458, opened 2017-12-29
use assert_request helper to make sure each stub is actually used

grosser:
@jonmoter @dragonfax

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/resource_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/resource_test.rb
@@ -5,12 +5,13 @@ SingleCov.covered!
 
 describe Kubernetes::Resource do
   def assert_pod_deletion
-    delete_pod = stub_request(:delete, "http://foobar.server/api/v1/namespaces/name1/pods/pod1").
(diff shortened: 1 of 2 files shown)


bblimke:
@grosser Thank you for the suggestion.

WebMock seeps global stub and request registries, therefore the solution you have suggestion only works in some cases.

The stub_request invocation inside the block is a global declaration and the stub is available after the block as well.
At the same time the request registry is global as well, therefore any requests executed before the block will exist in that registry as well.

grosser:
... I'm not super worried about that since my usecase is mostly "requests were actually made", but undoing the stubs would be a nice cleanup ... any api I can use to de-register the stubs ?

requests executed before the block will fail, so I'm not worried about that

bblimke:
There is no API to unstub. It won't be hard to add a method for that, by going through WebMock::StubRegistry and removing all matching stubs.

oliakremmyda:
@bblimke can't `remove_request_stub` api function be used for this case?

grosser:
works, thx! :)
https://github.com/zendesk/samson/pull/2530

grosser:
FYI made the helper behave more like rspec/mocha assertions in https://github.com/zendesk/samson/pull/2473 ... would love to see that in webmock itself :)

PR #2433, opened 2017-12-15
show 'running on' on builds

grosser:
screen shot 2017-12-14 at 7 47 28 pm
screen shot 2017-12-14 at 7 41 59 pm
screen shot 2017-12-14 at 8 23 26 pm
screen shot 2017-12-14 at 8 22 29 pm

a lot of cleanup too ... and removing the duplicate show on releases/index when expanding a release

@jonmoter

PR #2466, opened 2018-01-05
show active stages in different color as before

grosser:
broken in https://github.com/zendesk/samson/pull/2433
... removing cloud-icon since color is enough

@dragonfax @bquorning 
  
screen shot 2018-01-05 at 9 45 36 am

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -303,11 +303,11 @@ module ApplicationHelper
     stages.each do |stage|
       next unless deploy = stage.deployed_or_running_deploy
       next unless deploy.references?(reference)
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -303,11 +303,11 @@ def deployed_or_running_list(stages, reference)
     stages.each do |stage|
       next unless deploy = stage.deployed_or_running_deploy
       next unless deploy.references?(reference)
+      label = (deploy.active? ? "label-warning" : "label-success")
```
bquorning: Why ‘warning’ and not e.g. ‘active’?

grosser: warning is what we used before, it makes it yellow

PR #2469, opened 2018-01-08
delete orphaned pods when we delete statefulset

grosser:
not only do they keep hanging around, they also break future deploys
this caused memcached deploy foobar on Friday

@dragonfax @jonmoter 
/cc @libo


libo:
giphy-tumblr

PR #2472, opened 2018-01-10
dont fail a deploy when a autoscale down event happens during a deploy

craig-day:
This adds a check to the deploy executor for an autoscaled role. If a role is autoscaled, sort the pods for that role by "liveliness". This will order pods following this order: live, pending, stopped. After sorting, only use the most alive pod to check status because there might be a scale event during a deploy which destroys pods. We don't want that scale down event to fail a deploy if there is still at least some valid pods running.

I also added a note to the deploy output when you are only going to see one pod when there might by many replicas.

```
Creating for Pod 100 role resque-worker
Creating for Pod 100 role app-server
Waiting for pods to be created
Deploy status after 0 seconds:
  Pod 100 resque-worker: Live (autoscaled role, only showing one pod)
  Pod 100 app-server: Live
READY, starting stability test
```

/cc @zendesk/samson

### Risks
- Level: Low. Kubernetes deploys might fail to properly display errors is a role is marked as autoscaled.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -218,7 +218,7 @@ module Kubernetes
       statuses = Array.new(release_doc.desired_pod_count).each_with_index.map do |_, i|
         pod = pods[i]
 
-        if !pod
+        s = if !pod
           {live: false, details: "Missing", pod: pod}
         elsif pod.restarted?
           {live: false, stop: true, details: "Restarted", pod: pod}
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -218,7 +218,7 @@ def release_statuses(pods, release_doc)
       statuses = Array.new(release_doc.desired_pod_count).each_with_index.map do |_, i|
         pod = pods[i]
 
-        if !pod
+        s = if !pod
```
grosser: `status` ?

PR #2473, opened 2018-01-11
Kubernetes Blue/Green Deployment

zenoaken:
Currently Kubernetes supports two deployment types: Recreate and Rolling Update. While these types fit nicely for most use cases, some applications might have specific needs i.e. not run version N and N+1 in parallel and have (near) zero downtime while deploying. In this case, a blue green deployment is the most suitable choice.

This PR adds a blue/green deployment feature for Kubernetes deployments.

* New `Blue Green Deployment` flag on Kubernetes Roles
screen shot 2018-01-19 at 15 43 11

* Display `color` in Kubernetes Releases
screen shot 2018-01-11 at 15 53 44

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - https://zendesk.atlassian.net/browse/PAAS-1185
 - https://zendesk.atlassian.net/browse/INFR-4854

### Risks
- Level: Med


db/schema.rb
```diff
@@ -21,7 +21,7 @@
     t.string "user_type"
     t.string "username"
     t.string "action", null: false
-    t.text "audited_changes", limit: 1073741823
+    t.text "audited_changes", limit: 4294967295
```
grosser: revert that :D

db/schema.rb
```diff
@@ -246,6 +246,7 @@
     t.integer "deploy_id"
     t.string "git_sha", limit: 40, null: false
     t.string "git_ref"
+    t.string "color", limit: 8
```
grosser: better something like "blue_green_phase" ... maybe even a boolean

plugins/kubernetes/app/views/samson_kubernetes/_stage_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <fieldset>
   <legend>Kubernetes</legend>
   <%= form.input :kubernetes, as: :check_box, label: "Deploy via kubernetes / ignore commands" %>
+  <%= form.input :blue_green, as: :check_box, label: "Blue/Green Deployment" %>
```
grosser: maybe add `help: "explain what blue-green is and how it works under the hood"`

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -252,17 +265,24 @@ def release_doc_metadata
...
           revision: release.git_sha,
           tag: release.git_ref.parameterize.tr('_', '-')
         )
+
+        meta[:blue_green] = blue_green_color if blue_green_color
+        meta
       end
     end
 
+    def set_spec_match_labels
```
grosser: `def mark_blue_green_phase` ?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -81,16 +83,25 @@ def set_project_labels
...
+          template.dig_set([:spec, :template, :metadata, :labels, :blue_green], blue_green_color)
+        end
       end
     end
 
     def set_service_name
       template[:metadata][:name] = generate_service_name(template[:metadata][:name])
     end
 
+    def set_service_selector
```
grosser: would also put something about blue-green in the method name here

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -81,16 +83,25 @@ def set_project_labels
...
+      template.dig_set([:spec, :selector, :blue_green], blue_green_color) if blue_green_color
```
grosser: this looks like it should belong close to the other blue_green setters above

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -215,7 +216,8 @@ def request_delete
         # Wait for there to be zero pods
         loop do
           loop_sleep
-          break if fetch_resource.dig_fetch(:status, :replicas).zero?
```
grosser: `break if fetch_resource.dig(:status, :replicas).to_i.zero?`

... also some docs on why something would have no replica setting

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -60,6 +68,11 @@ def builds
       Build.where(git_sha: git_sha)
     end
 
+    def previous_successful_release
+      previous_successful_deploy = deploy.previous_successful_deploy
```
grosser: `deploy.previous_successful_deploy&.kubernetes_release`

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -19,6 +19,14 @@ def user
     def self.create_release(params)
       Kubernetes::Release.transaction do
         release = create(params.except(:deploy_groups))
+
+        if release.deploy.stage.blue_green
+          previous = release.previous_successful_release
+          color = 'green' if previous && previous.color == 'blue'
```
grosser: `previous&.color == 'blue'`

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -19,6 +19,14 @@ def user
...
+          color ||= 'blue'
+          update(release.id, :color => color)
```
grosser: `color: color`

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -343,6 +343,8 @@ def deploy(release_docs)
     end
 
     def deploy_and_watch(release, release_docs)
+      return deploy_and_watch_blue_green(release, release_docs) if @job.deploy.stage.blue_green
```
grosser: better put the if/else to where deploy_and_watch is called

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -355,6 +357,54 @@ def deploy_and_watch(release, release_docs)
       end
     end
 
+    def deploy_and_watch_blue_green(release, release_docs)
```
grosser: could this live in it's own BlueGreenDeployer class ?

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -355,6 +357,54 @@ def deploy_and_watch(release, release_docs)
...
+      deploy_blue_green_resources(release_docs)
+
+      # wait for resources to be ready
+      result = wait_for_resources_to_complete(release, release_docs)
+      if result == true
+        deploy_blue_green_service(release_docs)
```
grosser: more like `switch_blue_green_service` ?

grosser:
that looks about right logic-wise ... I was hoping for some logic on when doing the switch was ok, but I guess this is a good first step to get the strategy

let me know when you are happy with the state it's in and I'll give this a bit of refactoring/testing etc goodness

zenoaken:
Thanks for reviewing this @grosser ! I just pushed new code addressing the comments.

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -23,6 +23,11 @@ def initialize(stop: false, live:, details:, pod:, role:, group:)
       end
     end
 
+    def self.build(output, job:, reference:)
```
grosser: `*args` and then pass that on

grosser: arg ... nah we need the job ... `output, job:. **args` maybe ... but not worth the trouble

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -19,11 +19,25 @@ def user
     def self.create_release(params)
       Kubernetes::Release.transaction do
         release = create(params.except(:deploy_groups))
+
+        update(release.id, blue_phase: !release.previous_successful_release&.blue_phase) if release.blue_green?
```
grosser: `release.update_column : blue_phase, !release.previous_successful_release&.blue_phase`
... maybe even a dedicated `flip_blue_green!` method

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -32,6 +32,14 @@ def revert
       end
     end
 
+    def blue_green?
```
grosser: `delegate :blue_green?, : blue_green_color, to: : kubernetes_release `

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -45,6 +53,16 @@ def resource_template
       @resource_template ||= Array.wrap(super).map(&:deep_symbolize_keys)
     end
 
+    def non_service_resources
+      @non_service_resources ||= resources.select { |r| r != service_resource}
```
grosser: maybe: select != -> reject ==

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -45,6 +53,16 @@ def resource_template
...
+    end
+
+    def service_resource
```
grosser: cache won't work with nil

```
return @service_resource if defined?(@service_resource)
@service_resource = ...
```

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -215,7 +216,13 @@ def request_delete
...
+          r = fetch_resource
+          # prevent cases when status.replicas are missing
+          # e.g. running locally on Minikube, after scale replicas to zero
+          # $ kubectl scale deployment {DEPLOYMENT_NAME} --replicas 0
+          # "replicas" key is actually removed from "status" map
+          # $ {"status":{"conditions":[...],"observedGeneration":2}}
```
grosser: thx, good to know :)

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -215,7 +216,13 @@ def request_delete
...
+          break if !r[:status].key?(:replicas) || r.dig_fetch(:status, :replicas).zero?
```
grosser: `break if fetch_resource.dig(:status, :replicas).to_i.zero?`

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -74,6 +76,16 @@ def images
 
     private
 
+    def set_service_blue_green
+      template.dig_set([:spec, :selector, :blue_green], blue_green_color)
+    end
+
+    def set_blue_green
```
grosser: `set_resource_blue_green` to be symmetric

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -83,7 +95,6 @@ def set_project_labels
         template.dig_set([:spec, :selector, :project], project_label)
       elsif kind != "Pod"
         template.dig_set([:spec, :selector, :matchLabels, :project], project_label)
-        template.dig_set([:spec, :template, :metadata, :labels, :project], project_label)
```
grosser: accident ?

plugins/kubernetes/app/views/samson_kubernetes/_stage_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <fieldset>
   <legend>Kubernetes</legend>
   <%= form.input :kubernetes, as: :check_box, label: "Deploy via kubernetes / ignore commands" %>
+  <%= form.input :blue_green, as: :check_box, label: "Blue/Green Deployment", help: "Starts a new isolated deployment shifting between blue and green sufixes, switching service selectors if successfully deployed and deleting previous resources" %>
```
grosser: great help text!

app/models/job_execution.rb
```diff
@@ -163,8 +163,8 @@ def execute(dir)
 
     ActiveSupport::Notifications.instrument("execute_job.samson", payload) do
       payload[:success] =
-        if defined?(Kubernetes::DeployExecutor) && stage&.kubernetes
-          @executor = Kubernetes::DeployExecutor.new(@output, job: @job, reference: @reference)
+        if stage&.kubernetes
+          @executor = Kubernetes::DeployExecutor.build(@output, job: @job, reference: @reference)
```
grosser: how about `elsif kubernetes && blugreen` here and avoid the `build` method ?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
zenoaken: oops 🙈

zenoaken:
@grosser thanks for the comments. Whenever you're available to give that bit of refactoring/testing, it would be very welcome.

grosser:
done except for the deploy executor ... coming tomorrow ...

I switched the blue/green logic in release.rb since it looked wrong to me, please double-check

grosser:
you just pushed some commit, right ? ... had to rebase

zenoaken:
Sorry for that. I synced with master and started to fix some test issues. Thanks for the commits, I'll start to double-check right now

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -18,23 +18,16 @@ def user
     # Creates a new Kubernetes Release and corresponding ReleaseDocs
     def self.create_release(params)
       Kubernetes::Release.transaction do
-        release = create(params.except(:deploy_groups))
-        release.flip_blue_green!
+        release = create(params.except(:deploy_groups)) do |release|
+          release.blue_phase = release.previous_successful_release&.blue_phase if release.blue_green_color
```
zenoaken: `!release.previous_successful_release&.blue_phase`

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -556,6 +555,24 @@ def add_init_container(container)
         template.to_hash.dig_fetch(:spec, :scaleTargetRef, :name).must_equal("test-app-server")
       end
     end
+
+    describe "blue-green" do
+      before { doc.kubernetes_release.deploy.stage.blue_green = true }
+
+      it "modifies the service" do
+        raw_template[:kind] = 'Service'
+        template.to_hash.dig_fetch(:spec, :selector, :blue_green).must_equal 'green'
```
zenoaken: The very first deploy should be 'blue'. Maybe it's green now because we're not flipping `blue_phase` in https://github.com/zendesk/samson/pull/2473/commits/0bfb637a67b9559c1a9867376d10ae83a68f0a58#diff-aa5b7e4c40a65c9a9a4b56d6b66e9d5aR22

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -18,12 +18,18 @@ def user
     # Creates a new Kubernetes Release and corresponding ReleaseDocs
     def self.create_release(params)
       Kubernetes::Release.transaction do
-        release = create(params.except(:deploy_groups))
+        release = create(params.except(:deploy_groups)) do |release|
+          release.blue_phase = release.previous_successful_release&.blue_phase if release.blue_green_color
```
zenoaken: `!release.previous_successful_release&.blue_phase`

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
zenoaken: The very first deploy should be 'blue'. Maybe it's green now because we're not flipping `blue_phase` in https://github.com/zendesk/samson/pull/2473/commits/0bfb637a67b9559c1a9867376d10ae83a68f0a58#diff-aa5b7e4c40a65c9a9a4b56d6b66e9d5aR22

grosser: adding tests for this atm and it's green ;)

plugins/kubernetes/app/models/kubernetes/release.rb
grosser: good catch!

grosser:
more fixes and tests, hope it's green now ...
I added a sanity check so we don't deploy and delete the same color ... just in case a deployment got cancelled in a weird state

grosser:
please give it  another test-run and then it should be good to go ...

@jonmoter @dragonfax for more eyes

jonmoter:
A few thoughts:

1. "Use blue/green deployment" should be an attribute of a role, rather than an attribute of a stage. You might want a scenario where you want to use a blue/green strategy on the app server, but not the workers.  Also, is there a scenario where we'd want some stages to use blue/green, and some stages where we don't?
2. How this works should be documented in the plugin README or some other doc file.
3. If you deploy with blue/green enabled, and then deploy again with it disabled, does it handle that case properly?

plugins/kubernetes/app/models/kubernetes/blue_green_deploy_executor.rb
```diff
@@ -0,0 +1,60 @@
...
+        other_resources(release_doc).each(&:deploy)
+      end
+    end
+
+    def switch_service(release_docs)
+      release_docs.each do |release_doc|
+        next unless service = service(release_doc)
+        @output.puts "Switching service for #{release_doc.deploy_group.name} " \
+          "role #{release_doc.kubernetes_role.name} to #{release_doc.blue_green_color.upcase}"
+        service.deploy
```
jonmoter: What happens if this fails?

grosser: keeps pointing at the old resources and deploy fails

zenoaken: Maybe rescue exceptions and call `delete_resources(release_docs)`?

grosser:
converting that to role is fun ... also shows a bunch of other bugs: console was not supported and disabling rollbacks was not either ... so rework incoming

grosser:
ok, moved to role ... but now everything is intertwined in deploy_executor again :D

grosser:
now also supports mixing roles (console + app-server) and having prerequisites (pod/job)

zenoaken:
Good thinking @jonmoter ! It really makes sense control `blue-green` in the roles.

There was one Rubocop offense breaking CI. I just pushed a fix. I'll also give another round of run/tests locally  🤞

zenoaken:
We're almost there. It worked fine for one role using blue/green, handling successful and failure deployments 🎉 

I might have found a bug though. When deploying two roles: one blue/green and another "normal". Both are being considered as blue/green, both release docs are being overwritten with blue/green properties, etc.

Probably the root cause is that we're controlling the flow using `if release_doc.blue_green_color`, which delegates to `release`.

`release.blue_green_color` is configured if [any](https://github.com/zendesk/samson/pull/2473/files#diff-aa5b7e4c40a65c9a9a4b56d6b66e9d5aR22) role is blue/green. So, check for `release_doc.blue_green_color` would return the color even if the current doc isn't blue/green, but some other is. We should take decisions based on `doc.kubernetes_role.blue_green` instead.

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -18,7 +18,13 @@ def user
     # Creates a new Kubernetes Release and corresponding ReleaseDocs
     def self.create_release(params)
       Kubernetes::Release.transaction do
-        release = create(params.except(:deploy_groups))
+        release = create(params.except(:deploy_groups)) do |release|
+          if params.fetch(:deploy_groups).any? { |dg| dg.fetch(:roles).any? { |role| role.fetch(:role).blue_green? } }
+            release.blue_green_color = begin
+              release.previous_successful_release&.blue_green_color == "blue" ? "green" : "blue"
```
zenoaken: very minor: as a user, I'd expect the first deploy to be `blue` (to be consistent with the strategy name). What if we invert it comparing previous color to green?

grosser:
will fix these up shortly

On Fri, Jan 19, 2018 at 8:18 AM, Luiz Carvalho Jr. <notifications@github.com

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -426,5 +442,37 @@ def temp_release_docs
...
+    def finish_blue_green_deployment(release, release_docs)
+      switch_blue_green_service(release_docs)
+      previous = release.previous_successful_release
+      if previous && previous.blue_green_color != release.blue_green_color
+        previous.release_docs.each { |d| delete_blue_green_resources(d) }
+      end
+    end
+
+    def switch_blue_green_service(release_docs)
+      release_docs.each do |release_doc|
```
zenoaken: We should select only `release_docs` having `kubernetes_role.blue_green` to perform this switch. All non-blue/green services were already created in [release_doc.deploy](https://github.com/zendesk/samson/pull/2473/files#diff-026ab5a569538b3fed9f467d2b8acbf3R372)

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -426,5 +442,37 @@ def temp_release_docs
         end
       end
     end
+
+    def finish_blue_green_deployment(release, release_docs)
+      switch_blue_green_service(release_docs)
+      previous = release.previous_successful_release
+      if previous && previous.blue_green_color != release.blue_green_color
+        previous.release_docs.each { |d| delete_blue_green_resources(d) }
```
zenoaken: We should select only `release_docs` having `kubernetes_role.blue_green` to perform the deletion. Maybe a `blue_green_release_docs` method to help?

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -426,5 +442,37 @@ def temp_release_docs
...
+    end
+
+    # used for blue_green deployment
+    def non_service_resources(release_doc)
+      release_doc.resources - Array(service_resource(release_doc))
+    end
+
+    # used for blue_green deployment
+    def service_resource(release_doc)
+      release_doc.resources.detect { |r| r.is_a?(Kubernetes::Resource::Service) }
```
zenoaken: Looking again to it I'm wondering if there would be some case with more than one service in the same release_doc 🤔 If so, maybe we could use `select` instead of `detect` and apply the switch for all services in then list?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -359,6 +375,11 @@ def static_env
       env.merge!(Samson::Hooks.fire(:deploy_group_env, project, @doc.deploy_group).inject({}, :merge!))
     end
 
+    def blue_green_color
+      return @blue_green_color if defined?(@blue_green_color)
```
zenoaken: `return if !@doc.kubernetes_role.blue_green`

zenoaken:
ps: I sent a comment regarding `as a user, I'd expect the first deploy to be blue ... What if we invert it comparing previous color to green`. I already deleted it, please ignore. I was in need of a coffee when I wrote it 😆

plugins/kubernetes/app/models/kubernetes/template_filler.rb
grosser: added that logic into release_doc itself

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
grosser: yeah, let's do that ... was thinking about that too

grosser:
done, anything else ?

zenoaken:
Sweet! It's handling nicely mixed strategies now:

screen shot 2018-01-19 at 17 29 25

Release docs are also consistent. 👍 LGTM

PR #2475, opened 2018-01-11
help text as popover not just html title

craig-day:
This changes how the `additional_info` renders help text. Instead of just adding the html `title` attribute, we add a bootstrap popover for two reasons:

1. The help text now shows immediately on hover instead of waiting
1. We can render actual HTML in the helper for a nicer display or examples

**before**

**after**

/cc @zendesk/samson

### Risks
- Level: Low. The help icon may fail to show if there is a javascript problem.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -29,3 +29,8 @@
 //= require underscore_mixins
 //= require jstimezonedetect
 //= require_tree .
+
+// initialize popovers after all of our other javascript is done
+$(function () {
+  $('[data-toggle="popover"]').popover()
+})
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -29,8 +29,3 @@
 //= require underscore_mixins
 //= require jstimezonedetect
 //= require_tree .
-
-// initialize popovers after all of our other javascript is done
-$(function () {
-  $('[data-toggle="popover"]').popover()
-})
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/application.js
```diff
@@ -29,3 +29,8 @@
 //= require underscore_mixins
 //= require jstimezonedetect
 //= require_tree .
+
+// initialize popovers after all of our other javascript is done
+$(function () {
+  $('[data-toggle="popover"]').popover()
+})
```
grosser: maybe not start the junk-drawer ... separate file ?

app/helpers/application_helper.rb
```diff
@@ -216,7 +217,13 @@ def link_to_history(resource, counter: true)
...
 
   def additional_info(text)
-    content_tag :i, '', class: "glyphicon glyphicon-info-sign", title: text
+    html_attrs = if text.html_safe?
+      { 'data-content': h(h(text).to_str), 'data-html': true }
+    else
+      { 'data-content': text }
+    end.merge(BOOTSTRAP_TOOLTIP_PROPS)
+
+    content_tag :i, '', class: "glyphicon glyphicon-info-sign", **html_attrs
```
grosser: content_tag -> tag

app/assets/javascripts/application.js
craig-day: the file will have to have a dumb name, because `require_tree .` needs to be the last directive so we need to make sure its the last thing in the tree.

PR #2477, opened 2018-01-11
dont fail a deploy when a hpa warns about metrics or failed scales

craig-day:
Deploys can be marked as failure if a pod event includes warnings from a horizontal pod autoscaler. 

Example:
```
FailedRescale: New size: 1; reason: All metrics below target; error: Operation cannot be fulfilled on deployments.extensions "some-app-server": the object has been modified; please apply your changes to the latest version and try again x3
```

This isn't a hard failure, just a warning, so lets not fail a deploy because of it. 

/cc @zendesk/samson

### Risks
- Level: Low. Deploys will succeed even if HPA might fail due to warnings that we ignore.

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/api/pod_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/api/pod_test.rb
@@ -305,6 +305,24 @@ describe Kubernetes::Api::Pod do
         assert events_indicate_failure?
       end
     end
+
+    describe "HorizontalPodAutoscaler with failures we don't care about" do
+      before do
+        event.merge!(kind: 'HorizontalPodAutoscaler', type: 'Warning')
+      end
+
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/test/models/kubernetes/api/pod_test.rb
```diff
@@ -305,6 +305,24 @@ def events_indicate_failure?
         assert events_indicate_failure?
       end
     end
+
+    describe "HorizontalPodAutoscaler with failures we don't care about" do
+      before do
+        event.merge!(kind: 'HorizontalPodAutoscaler', type: 'Warning')
+      end
+
+      it "ignores failing to get metrics" do
```
grosser: negative test missing ?

PR #2478, opened 2018-01-11
allow any users to create their personal access token

grosser:
create always submits resource_owner_id

@adam-hart


grosser:
@dragonfax

grosser:
yep, fails without the fix and code is now more general by always going through the same flow

grosser:
need a real `review`

PR #2482, opened 2018-01-15
remove heartbeat hack

grosser:
tested locally and seems to behave exactly the same as before ... 

https://github.com/puma/puma/issues/389 suggested maybe using force_shutdown_after ...

we'll see if any user complains and then use that option or revert and add better docs on what exactly this does/prevent


grosser:
@dragonfax

grosser:
@dragonfax

PR #2483, opened 2018-01-16
allow plugins to hook into access-control

grosser:


grosser:
@dragonfax 
/cc @jonmoter

grosser:
@dragonfax

PR #2484, opened 2018-01-19
hide cluttered mass-rollout UI

grosser:
very few people use this, so better hide it ...

before:
screen shot 2018-01-19 at 1 37 25 pm

after:

screen shot 2018-01-19 at 1 36 59 pm

screen shot 2018-01-19 at 1 37 04 pm


grosser:
@dragonfax

PR #2485, opened 2018-01-19
extract mass-rollout controller

grosser:
@dragonfax


grosser:
also adding
screen shot 2018-01-19 at 2 31 58 pm

because I did not figure out how to use it :)

PR #2488, opened 2018-01-22
make default seeds include kubernetes

grosser:
- allow plugins to seed
 - make kubernetes work out of the box with docker-for-mac (needs edge atm)
 - fix kubernetes deploys not working in development mode since classes are not defined


grosser:
@jonmoter @dragonfax @zenoaken

PR #2492, opened 2018-01-24
Optional Feature: require admin for prod (un)lock.

jgoerz:
- When set, will require admin role for (un)locking stages with production attribute set.
- All other locking is unchanged.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -173,3 +173,6 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # GCLOUD_OPTIONS=--verbose    # optional, options
 # GCLOUD_PROJECT=foo-123
 # GCLOUD_ACCOUNT=my-account
+
+## Feature: Only admins can (un)lock stages which affect production.
+# PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN=1
--- a/app/controllers/concerns/current_stage.rb
+++ b/app/controllers/concerns/current_stage.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+module CurrentStage
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :require_stage
+    helper_method :current_stage
+  end
+
+  def current_stage
+    @stage
+  end
+
+  protected
+
+  def require_stage
+    if current_project
+      query = Stage.where(project_id: current_project.id)
+      stage_param = params[:stage_id] || params[:id]
+      @stage = query.find_by_param!(stage_param) if stage_param
+    end
+  end
+end
--- a/app/controllers/concerns/current_stage.rb
+++ b/app/controllers/concerns/current_stage.rb
@@ -14,10 +14,8 @@ module CurrentStage
   protected
 
   def require_stage
-    if current_project
-      query = Stage.where(project_id: current_project.id)
-      stage_param = params[:stage_id] || params[:id]
-      @stage = query.find_by_param!(stage_param) if stage_param
-    end
+    return unless current_project
+    return unless stage_param = params[:stage_id] || params[:id]
+    @stage = Stage.where(project_id: current_project.id).find_by_param!(stage_param)
   end
 end
--- a/app/controllers/concerns/current_user.rb
+++ b/app/controllers/concerns/current_user.rb
@@ -66,7 +66,11 @@ module CurrentUser
   end
 
   def can?(action, resource_namespace, scope = nil)
-    scope ||= current_project if respond_to?(:current_project)
+    scope ||= if resource_namespace == :locks
+      try(:current_stage)
+    else
+      try(:current_project)
+    end
     AccessControl.can?(current_user, action, resource_namespace, scope)
   end
 end
--- a/app/controllers/concerns/current_user.rb
+++ b/app/controllers/concerns/current_user.rb
@@ -67,9 +67,9 @@ module CurrentUser
 
   def can?(action, resource_namespace, scope = nil)
     scope ||= if resource_namespace == :locks
-      try(:current_stage)
+      current_stage if respond_to?(:current_stage)
     else
-      try(:current_project)
+      current_project if respond_to?(:current_project)
     end
     AccessControl.can?(current_user, action, resource_namespace, scope)
   end
--- a/app/controllers/concerns/current_user.rb
+++ b/app/controllers/concerns/current_user.rb
@@ -66,10 +66,10 @@ module CurrentUser
   end
 
   def can?(action, resource_namespace, scope = nil)
-    scope ||= if resource_namespace == :locks
-      current_stage if respond_to?(:current_stage)
-    else
-      current_project if respond_to?(:current_project)
+    scope ||= if resource_namespace == :locks && respond_to?(:current_stage)
+      current_stage
+    elsif respond_to?(:current_project)
+      current_project
     end
     AccessControl.can?(current_user, action, resource_namespace, scope)
   end
--- a/app/controllers/locks_controller.rb
+++ b/app/controllers/locks_controller.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 class LocksController < ApplicationController
   include CurrentProject
+  include CurrentStage
 
-  before_action :require_project, if: :for_stage_lock?
+  before_action :require_stage, if: :for_stage_lock?
   before_action :authorize_resource!
 
   def index
@@ -56,13 +57,13 @@ class LocksController < ApplicationController
     @lock ||= Lock.find(params[:id])
   end
 
-  # Overrides CurrentProject#require_project
-  def require_project
+  # Overrides CurrentStage#require_stage
+  def require_stage
     case action_name
     when 'create' then
-      @project = Stage.find(params[:lock][:resource_id]).project
+      @stage = Stage.find(params[:lock][:resource_id])
     when 'destroy' then
-      @project = lock.resource.project
+      @stage = lock.resource
     else
       raise 'Unsupported action'
     end
--- a/app/models/access_control.rb
+++ b/app/models/access_control.rb
@@ -18,15 +18,12 @@ class AccessControl
         case action
         when :read then true
         when :write
-
-          if scope.nil? # global locks
+          if scope.is_a?(Stage) # stage locks
+              return user.admin_for?(scope.project) if ENV['PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN'] && scope.production
+              user.deployer_for?(scope.project)
+          else # global locks
             user.admin?
-          elsif ENV['PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN'] && scope.production
-            user.admin_for?(scope.project) # stage locks
-          else
-            user.deployer_for?(scope.project) # stage locks
           end
-
         else raise ArgumentError, "Unsupported action #{action}"
         end
       when :projects, :build_commands, :stages, :user_project_roles
--- a/app/models/access_control.rb
+++ b/app/models/access_control.rb
@@ -18,11 +18,15 @@ class AccessControl
         case action
         when :read then true
         when :write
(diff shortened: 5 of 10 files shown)


jgoerz:
Build failure is unrelated to the changes for this pull request.  Fails on slow test (didn't see this when I ran it locally)

>  Error:
> JobExecution::builds_in_environment#test_0001_makes builds available via env:
> Maxitest::Timeout::TestCaseTimeout: Test took too long to finish, aborting. To use a debugger, set Maxitest.timeout = false at the top of the test file.

And coverage: 
> app/models/event_streamer.rb has less uncovered lines (2 current vs 3 configured), decrement configured uncovered?

grosser:
hmmm I kinda like the idea of only letting admins unlock ... do you think the same restriction is necessary for locking ? ... I'd rather have anyone able to lock when something goes wrong ...

.env.example
```diff
@@ -173,3 +173,6 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # GCLOUD_OPTIONS=--verbose    # optional, options
 # GCLOUD_PROJECT=foo-123
 # GCLOUD_ACCOUNT=my-account
+
+## Feature: Only admins can (un)lock stages which affect production.
+# FEATURE_ONLY_ADMINS_CAN_LOCK_UNLOCK_PRODUCTION_STAGE=1
```
grosser: `ONLY_ADMINS_CAN_LOCK_UNLOCK_PRODUCTION_STAGE` should be long enough :D

grosser: ... maybe `PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN`

app/controllers/locks_controller.rb
```diff
@@ -60,11 +60,25 @@ def lock
   def require_project
     case action_name
     when 'create' then
-      @project = Stage.find(params[:lock][:resource_id]).project
+      stage = Stage.find(params[:lock][:resource_id])
```
grosser: this should only be handles by `access_control.rb` and not copied here ...

app/controllers/locks_controller.rb
```diff
@@ -60,11 +60,25 @@ def lock
...
+
+      if ENV['FEATURE_ONLY_ADMINS_CAN_LOCK_UNLOCK_PRODUCTION_STAGE'] && stage.production
+        @project.define_singleton_method(:production) { true }
+      else
+        @project.define_singleton_method(:production) { false }
+      end
     when 'destroy' then
       @project = lock.resource.project
+      if ENV['FEATURE_ONLY_ADMINS_CAN_LOCK_UNLOCK_PRODUCTION_STAGE'] && lock.resource.production
+        @project.define_singleton_method(:production) { true }
```
grosser: this looks very wrong :D

.env.example
jgoerz: yeah, naming things is always hard ;-)  I like PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN.

app/controllers/locks_controller.rb
jgoerz: I agree with this, but it seems that only the project object is passed into access control to make the decision for the locks.  I struggled with my debugger configuration, so maybe I overlooked something... but this is the reason for the hacking singleton method below.  I was basically trying to attach the stage attribute information to the project.  I'll do some more investigating...

app/controllers/locks_controller.rb
jgoerz: Yeah, see my other comment above.

app/controllers/locks_controller.rb
jgoerz: So I have a question about this:
```ruby 
# app/controllers/locks_controller.rb line 6
 before_action :require_project, if: :for_stage_lock?
```
I was going to originally change this to 
```ruby 
 before_action :require_stage, if: :for_stage_lock?
```
Is there some reason that could not be done?  Are there other use cases that use a `Project` to determine lockability?  It seems the locks apply to 2 things, global, and `Stages`.  Why is it that we pass a `Project` in for the access check?

grosser:
I think it's just because admin_for etc need a project, should be alright
to pass the stage in

jgoerz:
The above is going to fail on the "locks" current_user test (line 245: "renders when authorized via the project").  I'm a little confused by this test.  It seems to be testing if a project can be viewed by checking project level authorization... but a lock is only accessed via the project so I don't believe this is a valid test for the locks controller.  Let me know what you think.

jgoerz:
>  grosser commented:
> hmmm I kinda like the idea of only letting admins unlock ... do you think the same restriction is necessary for locking ? ... I'd rather have anyone able to lock when something goes wrong ...

For our use case, we just want to prevent accidental production deploys.  We plan on locking production stages with non-expiring locks.  So letting anyone lock wouldn't be a problem for us.  I could see how someone would want to strictly control both (un)lock.  The only thing I could see a problem with, is that after someone locked it, because of the role/perms, the Unlock button would be grayed out or not displayed at all... which might be confusing if someone was just pushing buttons (exploring/learning).

If it is implemented... maybe split the lock actions into 
1. :read
2. :create_or_update (or maybe just :lock)
3. :destroy or :unlock 

If I had a choice, I'd choose to leave it as is (less work for me :-P ).  But I'm flexible.

app/controllers/concerns/current_stage.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def current_stage
+    @stage
+  end
+
+  protected
+
+  def require_stage
+    @stage = (Stage.find_by_param!(params[:stage_id]) if params[:stage_id])
```
grosser: this won't work since stage permalinks are not globally unique, need to find base on project

app/controllers/concerns/current_user.rb
```diff
@@ -66,7 +66,12 @@ def resource_action
   end
 
   def can?(action, resource_namespace, scope = nil)
-    scope ||= current_project if respond_to?(:current_project)
+    case resource_namespace
```
grosser: this should be `scope ||=` to avoid duplication and to avoid the `respond_to?` calls

app/models/access_control.rb
```diff
@@ -18,11 +18,15 @@ def can?(user, action, resource_namespace, scope = nil)
         case action
         when :read then true
         when :write
-          if scope
-            user.deployer_for?(scope) # stage locks
-          else
+
+          if scope.nil?
             user.admin? # global locks
+          elsif ENV['PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN'] && scope.production
```
grosser: might be easier to read as:
```
else # stage locks
  if PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN
  else
  end
end

File: test/controllers/concerns/current_user_test.rb [hunk 7]
```diff
@@ -242,8 +243,11 @@ def perform_get(add = {})
         assert_response :unauthorized
       end
 
-      it "renders when authorized via the project" do
-        perform_get(project_id: projects(:test).id)
+      # This still authorizes based on Project, but for the lock controller, it needs a Stage object
+      # passed to access control since locks are per Stage.  See app/models/access_control.rb lines
+      # 22 - 28.
```
grosser: lines change ... prefer names :)

jgoerz:
Ready for review/feedback.

app/controllers/concerns/current_user.rb
```diff
@@ -66,7 +66,11 @@ def resource_action
   end
 
   def can?(action, resource_namespace, scope = nil)
-    scope ||= current_project if respond_to?(:current_project)
+    scope ||= if resource_namespace == :locks
+      try(:current_stage)
```
grosser: `try` does not make sense here, try is for nil prevention like `nil.try(:foo?)`

app/controllers/concerns/current_stage.rb
```diff
@@ -0,0 +1,23 @@
...
+    helper_method :current_stage
+  end
+
+  def current_stage
+    @stage
+  end
+
+  protected
+
+  def require_stage
```
grosser: bit simpler like this and also avoids building the `query` when not needed
added some caching on top too since we don't want to fetch the stage multiple times

```
def require_stage
  return unless current_project
  return unless stage_param = params[:stage_id] || params[:id]
  @stage ||= Stage.where(project_id: current_project.id).find_by_param!(stage_param)
end
```

grosser: updated, don't need the caching

app/models/access_control.rb
```diff
@@ -18,11 +18,15 @@ def can?(user, action, resource_namespace, scope = nil)
         case action
         when :read then true
         when :write
-          if scope
-            user.deployer_for?(scope) # stage locks
+
+          if scope.nil? # global locks
+            user.admin?
+          elsif ENV['PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN'] && scope.production
```
grosser: prefer to guard this in a `elsif scope.is_a?(Stage)`so future additions don't require a refactor

test/controllers/locks_controller_test.rb
```diff
@@ -118,6 +119,33 @@ def create_lock(resource = nil, options = {})
       end
     end
 
+    describe '#create with PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN' do
```
grosser: if it's not too complicated I'd prefer this under `#create` as `describe "with PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN"`

test/controllers/locks_controller_test.rb
```diff
@@ -159,6 +212,48 @@ def create_lock(resource = nil, options = {})
...
```
grosser: too much copy-paste for my taste ... the tests above should be sufficient ...

test/controllers/locks_controller_test.rb
```diff
@@ -170,6 +265,19 @@ def create_lock(resource = nil, options = {})
       end
     end
 
+    describe '#destroy with PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN' do
```
grosser: same here

test/controllers/locks_controller_test.rb
```diff
@@ -136,6 +164,31 @@ def create_lock(resource = nil, options = {})
...
+        delete :destroy, params: {id: lock.id}
+
+        assert_redirected_to '/back'
+        assert flash[:notice]
+
+        Lock.count.must_equal 0
+      end
+
+      it 'cannot destroy a stage production lock' do
+        stage.production = true
```
grosser: `stage.update_column(:production, true)`

test/controllers/locks_controller_test.rb
```diff
@@ -136,6 +164,31 @@ def create_lock(resource = nil, options = {})
...
+        stage.save
+        delete :destroy, params: {id: lock.id}
+
+        assert_response :unauthorized
```
grosser: this might be enough and drop the later assertions ?

test/controllers/locks_controller_test.rb
```diff
@@ -118,6 +119,33 @@ def create_lock(resource = nil, options = {})
...
 
+    describe '#create with PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN' do
+      with_env 'PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN' => 'true'
+      before { travel_to Time.now }
+      after { travel_back }
+
+      it 'cannot create a stage lock for a production stage' do
+        create_lock prod_stage, delete_in: 3600
+
+        assert_response :unauthorized
```
grosser: can drop the later assertions ... same in the test below

jgoerz:
Ready for review/feedback.

PR #2495, opened 2018-01-25
PAAS-1266 allow GCR rebuild

grosser:
@dndungu @jonmoter


app/controllers/builds_controller.rb
```diff
@@ -34,8 +34,8 @@ def new
   end
 
   def create
-    external_id = params.dig(:build, :external_id).presence
-    if external_id && @build = Build.where(external_id: external_id).first
+    if registering_external_build? && @build = find_external_build
+      return head :unprocessable_entity if @build.docker_repo_digest
```
ragurney: This can be removed.

grosser: nah, actually need this since edit is calling it with `validate: false` so it won't blow up but get silently ignored ... blowing up makes debugging faster :)

PR #2499, opened 2018-01-26
 Prevent cache overflow from secrets lookup

ragurney:
@grosser


config/application.rb
```diff
@@ -49,7 +49,7 @@ class Application < Rails::Application
     config.eager_load_paths << "#{config.root}/lib"
 
     if Rails.env.test?
-      config.cache_store = :memory_store
+      config.cache_store = :memory_store, { max_size: 1_000_000 }
```
grosser: revert that

Gemfile.lock
```diff
@@ -257,6 +257,7 @@ GEM
       http (= 0.9.8)
       recursive-open-struct (= 1.0.0)
       rest-client
+    large_object_store (1.3.3)
```
grosser: make sure to bump before merge

grosser:
FYI current `Marshal.dump(SecretStorage.lookup_cache).size` was 123253 ... but still good to not have a time-bomb in the codebase 💣

PR #2502, opened 2018-01-28
refactor deploy group role passing

grosser:


plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
```diff
@@ -161,6 +161,12 @@ def worker_is_unstable
       out.wont_include "Deploy status after"
     end
 
+    it "does limited amounts of queries" do
+      assert_sql_queries(37) do
```
grosser: before we had 38, so a little bit better even :)

PR #2503, opened 2018-01-28
remove deprecated DOCKER_REGISTRY

grosser:


app/models/docker_registry.rb
```diff
@@ -3,38 +3,22 @@
 class DockerRegistry
   class << self
     def check_config!
-      if ENV['DOCKER_REGISTRIES'].blank? && ENV['DOCKER_REGISTRY'].blank?
+      return if ENV['DOCKER_REGISTRIES'].present?
+      if ENV['DOCKER_REGISTRY']
+        abort '*** rename DOCKER_REGISTRY env var to DOCKER_REGISTRIES ***'
```
grosser: ... and include user/password

grosser: fixed

PR #2508, opened 2018-01-30
allow fetching repo digest from gcloud when sync failed

grosser:
@jonmoter @ragurney 

screen shot 2018-01-29 at 6 46 53 pm
screen shot 2018-01-29 at 6 36 52 pm
screen shot 2018-01-29 at 6 36 05 pm

### DIFF
--- a/plugins/gcloud/app/controllers/gcloud_controller.rb
+++ b/plugins/gcloud/app/controllers/gcloud_controller.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+class GcloudController < ApplicationController
+  def sync_build
+    build = Build.find(params[:id])
+    path = [build.project, build]
+
+    if error = update_build(build)
+      redirect_to path, alert: "Failed to sync: #{error}"
+    else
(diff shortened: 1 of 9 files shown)


plugins/gcloud/app/controllers/gcloud_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    success, output = Samson::CommandExecutor.execute(*command, timeout: 10, whitelist_env: ["PATH"])
+    return "Failed to execute gcloud command: #{output}" unless success
+
+    JSON.parse(output).dig_fetch("results", "images").each do |image|
+      name = image.fetch("name")
+      if name.end_with?("/#{build.image_name}")
+        digest = image.fetch("digest")
+        build.update_attributes!(
+          docker_repo_digest: "#{name}@#{digest}",
+          external_status: "succeeded"
```
jonmoter: What about the scenario where Samson thinks the build is still running, but it actually failed?  Would be good to update the status in that case too.

PR #2509, opened 2018-01-30
do not ignore build updates when they come in in rapid succession

grosser:
@jonmoter @ragurney @dndungu 

should take care of most gcr-watcher sisues ...
and if not we got https://github.com/zendesk/samson/pull/2508

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -34,15 +34,20 @@ class BuildsController < ApplicationController
   end
 
   def create
-    if registering_external_build? && @build = find_external_build
-      return head :unprocessable_entity if @build.docker_repo_digest
-      @build.attributes = edit_build_params(validate: false)
-    else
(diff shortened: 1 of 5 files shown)


jonmoter:
So the issue is that when two POSTs from gcr_watcher came in rapid succession, and they were being processed at the same time?

I thought the way gcr_watcher worked, it would wait for one POST to Samson to return before sending the next one.

grosser:
does not seem to do that ... saw the same request except with a different
status coming in right after each other and the second one failed with
uniquness
I'll debug this a bit more tomorrow, still need to verify rebuild work in
prod anyway

grosser:
definitely not working, requests came in 1s apart and the first request took 4s to finish

2018-01-29T23:14:09.887Z ... duration":3819.98
2018-01-29T23:14:10.941Z

grosser:
I think this is a good pattern in general since we had the same issue already when `release branch` and `deploy on release` tried to create builds at the same time

PR #2516, opened 2018-02-01
DEVEX-54 Allow all users to see projects#edit page

ragurney:
screen shot 2018-01-31 at 5 14 11 pm

### DIFF
--- a/app/views/projects/_form.html.erb
+++ b/app/views/projects/_form.html.erb
@@ -5,7 +5,7 @@
 
   <%= render 'shared/errors', object: @project %>
 
-  <fieldset>
+  <fieldset <%= can?(:write, :projects, @project) ? '' : 'disabled' %>>
     <%= form.input :name, required: true %>
     <%= form.input :description, as: :text_area, help: 'How would you explain the project to someone new?' %>
     <%= form.input :owner, help: 'An email or group to contact with questions.' %>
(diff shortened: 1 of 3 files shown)


app/views/projects/_form.html.erb
```diff
@@ -5,7 +5,7 @@
 
   <%= render 'shared/errors', object: @project %>
 
-  <fieldset>
+  <fieldset <%= can?(:write, :projects, @project) ? '' : 'disabled' %>>
```
grosser: prefer screenshot when doing UI changes

grosser: hide the save button ?

test/controllers/projects_controller_test.rb
```diff
@@ -139,9 +139,18 @@
   end
 
   as_a_deployer do
-    unauthorized :get, :edit, id: :foo
```
grosser: never delete these, just move them up

grosser: ... hmm guess viewers can see them too so no real use anymore ...

test/controllers/projects_controller_test.rb
```diff
@@ -152,6 +161,9 @@
       it "renders" do
         get :edit, params: {id: project.to_param}
         assert_template :edit
+        assert_select 'fieldset' do |fs|
```
grosser: `def fields_disabled?` and then `assert fields_disabled?` + `refute fields_disabled?` for symmetry and so they never get out of sync by accident

app/views/projects/_form.html.erb
grosser: I like the fieldset disable, better that js hacks :D

PR #2520, opened 2018-02-02
DEVEX-79 Have mass stage creation set kubernetes replicas/limits/requests

ragurney:


plugins/kubernetes/lib/samson_kubernetes/samson_plugin.rb
```diff
@@ -63,3 +63,18 @@ def self.retry_on_connection_errors(&block)
...
 
 Samson::Hooks.callback(:deploy_group_includes) { :kubernetes_cluster }
+
+Samson::Hooks.callback(:stage_clone) do |old_stage, new_stage|
+  roles_to_copy = Kubernetes::DeployGroupRole.where(
+    project: old_stage.project,
+    deploy_group: old_stage.deploy_groups.last
+  )
+
+  roles_to_copy.each do |dgr|
```
grosser: clone does not always have to have 1 deploy-group, so should be `new_stage.deploy_groups.each do`

PR #2522, opened 2018-02-02
explain that build disabled and build with GCB cannot be both enabled

grosser:
screen shot 2018-02-01 at 8 06 42 pm

### DIFF
--- a/plugins/gcloud/app/decorators/project_decorator.rb
+++ b/plugins/gcloud/app/decorators/project_decorator.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+Project.class_eval do
+  validate :validate_not_using_gcb_and_external
+
+  private
+
+  def validate_not_using_gcb_and_external
+    return unless build_with_gcb && docker_image_building_disabled
+    errors.add(:build_with_gcb, "cannot be enabled when Docker images are built externally")
+  end
+end
--- a/plugins/gcloud/test/decorators/project_decorator_test.rb
+++ b/plugins/gcloud/test/decorators/project_decorator_test.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+describe Project do
+  let(:project) { projects(:test) }
+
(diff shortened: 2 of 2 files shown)


jonmoter:
How about we make a dropdown UI element instead of check boxes?

grosser:
tried that here, but it's a lot more code/indirection and there is no nice way to put hint text

https://github.com/zendesk/samson/pull/2523

I think this is already a good step forward, doubt anyone will be confused/complain, can but a real refactor to a single column into the backlog, but doubt it will ever be the highest priority item ...

jonmoter:
> doubt anyone will be confused/complain

I've already spoken to several people who are confused by this.

"I'm using GCB to build my images, so I should check the box saying Build with GCB, right?"
"No, not that checkbox..."

plugins/gcloud/app/decorators/project_decorator.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+Project.class_eval do
+  validate :validate_not_using_gcb_and_external
+
+  private
+
+  def validate_not_using_gcb_and_external
+    return unless build_with_gcb && docker_image_building_disabled
+    errors.add(:build_with_gcb, "cannot be enabled when building is disabled")
```
jonmoter: None of the checkboxes say anything about building being disabled.

The text of the checkboxes is:
* Docker images built externally 
* Include in New Deploy Groups 
* Build docker on GCB 

The second checkbox doesn't have anything to do with the first two, which is confusing enough.

So at the very least, this error message should be "cannot be enabled when Docker images are built externally."

jonmoter:
How about on the project settings page, we create a separate section for the docker-related settings.  That would include:

* Docker release branch
* Dockerfiles
* Docker images built externally
* Build docker on GCB

And put some text there explaining how that section works.  Also, if you check "Docker images built externally", that should clear/disable the "Dockerfiles" and "Docker release branch" inputs.

jonmoter:
Can you change the error message to "cannot be enabled when Docker images are built externally."?

grosser:
updated message, added "Reorganize docker section" task DEVEX-92 to make this better

PR #2525, opened 2018-02-02
Sync periodical with restarts

ragurney:

### DIFF
--- a/app/models/restart_signal_handler.rb
+++ b/app/models/restart_signal_handler.rb
@@ -50,7 +50,8 @@ class RestartSignalHandler
     output 'preparing restart'
 
     JobQueue.enabled = false # Disable new job execution
-    wait_for_active_jobs_to_finish
+    Samson::Periodical.enabled = false
+    wait_for_active_jobs_to_stop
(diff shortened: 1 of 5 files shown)


lib/samson/periodical.rb
```diff
@@ -46,8 +48,10 @@ def run
           next unless config.fetch(:active)
           with_consistent_start_time(config) do
             Concurrent::TimerTask.new(config) do
-              ActiveRecord::Base.connection_pool.with_connection do
-                execute_block(config)
+              with_running_task do
```
grosser: maybe better as `track_running_count do`

app/models/restart_signal_handler.rb
```diff
@@ -50,6 +50,7 @@ def run
     output 'preparing restart'
 
     JobQueue.enabled = false # Disable new job execution
+    Samson::Periodical.enabled = false
     wait_for_active_jobs_to_finish
```
grosser: that method name is a bit misleading now too ... `wait_for_activities_to_stop`

PR #2527, opened 2018-02-02
Don't clone the pipeline

dragonfax:
Otherwise, we get duplicate builds.

I.e. after the clone, Production(stage) finish and starts up Pod 15(stage) and Pod17(stage). But then after Pod17 (stage) is done, it starts up Pod 15 (stage) __again__.

/cc @zendesk/samson

@grosser  I'm thinking maybe this should go deeper, i.e. inside `build_clone`. So that other cloned stages don't get the pipeline clones either. What do you think?

### References
 - Jira link: https://zendesk.atlassian.net/browse/DEVEX-82

### Risks
- Level: Low

### DIFF
--- a/app/controllers/mass_rollouts_controller.rb
+++ b/app/controllers/mass_rollouts_controller.rb
@@ -134,6 +134,7 @@ class MassRolloutsController < ApplicationController
       is_template: false,
       next_stage_ids: []
     )
+
     stage.save!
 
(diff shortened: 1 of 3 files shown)


grosser:
failing test would be nice, also how about moving this into `build_clone` to avoid that issue for regular clone workflow ?

dragonfax:
Thats what I suggested. Thanks.

dragonfax:
@grosser

PR #2532, opened 2018-02-05
No need for a github token during db:migrate

dragonfax:
There should be no need to have github token set just to run `rake db:migrate`

I'm __certain__ there is a better way to write this.

@zendesk/samson

### DIFF
--- a/config/initializers/octokit.rb
+++ b/config/initializers/octokit.rb
@@ -30,7 +30,7 @@ Octokit.connection_options[:request] = { open_timeout: 2 }
 
 token = ENV['GITHUB_TOKEN']
 
-unless Rails.env.test? || ENV['PRECOMPILE']
+unless Rails.env.test? || ENV['PRECOMPILE'] || ( defined?(Rake) && Rake.application.top_level_tasks.include?("db:migrate") )
   raise "No GitHub token available" if token.blank?
 end
 


dragonfax:
Rake gives an error when I try to startup the application. Its not defined.
Only when your running a rake task.

If a migration is accessing github thats a major problem. Migrations are
database only, by design. Right?

dragonfax:
I mean, Migrations are supposed to be deterministic. You can't trust a
migration if its getting some state from the internet.

grosser:
I can imagine a scenario where it might be needed, so I'd rather not allow
it to be blank unless there is a good reason.

Also the current code would allow it to be blank for all rake tasks.

Why is having a github-token an issue ? The default env seed include it ?

dragonfax:
What kind of scenario do you have in mind? I can't imagine a reasonable
scenario where I would approve a database migration that needed to check
github.

I'm going to close this PR. I wasn't using the seed env file, which I found
out when I tried to get tests to run locally.

grosser:
I'm mostly thinking of usecases where some git_sha / ref / kubernetes_role
is wrong and would need new data from github.

Try bin/setup, that should make everything work

dragonfax:
yeah, my fault. I did zdi bootstrap and went from there.

PR #2535, opened 2018-02-05
Add a little documentation for setup using zdi

dragonfax:
@zendesk/samson 

@grosser  Hows this?


grosser:
zdi does not belong into a public repo ... make bootstrap "just work" instead ?

dragonfax:
Found a place on our internal docs to put this.

PR #2542, opened 2018-02-06
Feature: Use LDAP_UID as user.external_id

jgoerz:
- Use the value of the LDAP_UID not just for the query but also as the user's
  external_id when using LDAP.

* Add description here
* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -90,11 +90,12 @@ class SessionsController < ApplicationController
   end
 
   def login(options = {})
-    uid = auth_hash.uid
-    if ENV['AUTH_LDAP'] && ENV['USE_LDAP_UID_AS_EXTERNAL_ID']
+    if auth_hash.provider == 'ldap' && ENV['AUTH_LDAP'] && ENV['USE_LDAP_UID_AS_EXTERNAL_ID']
       uid_field = Rails.application.config.samson.ldap.uid
       uid = auth_hash.extra.raw_info.send(uid_field).presence || raise
       uid = Array(uid).first
+    else
+      uid = auth_hash.uid
     end
 
     user = User.create_or_update_from_hash(options.merge(
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -91,7 +91,7 @@ class SessionsController < ApplicationController
 
   def login(options = {})
     uid = auth_hash.uid
-    if ENV['USE_LDAP_UID_AS_EXTERNAL_ID']
+    if ENV['AUTH_LDAP'] && ENV['USE_LDAP_UID_AS_EXTERNAL_ID']
       uid_field = Rails.application.config.samson.ldap.uid
       uid = auth_hash.extra.raw_info.send(uid_field).presence || auth_hash.uid
       uid = uid.is_a?(Array) ? uid.first : uid
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -93,8 +93,8 @@ class SessionsController < ApplicationController
     uid = auth_hash.uid
     if ENV['AUTH_LDAP'] && ENV['USE_LDAP_UID_AS_EXTERNAL_ID']
       uid_field = Rails.application.config.samson.ldap.uid
-      uid = auth_hash.extra.raw_info.send(uid_field).presence || auth_hash.uid
-      uid = uid.is_a?(Array) ? uid.first : uid
+      uid = auth_hash.extra.raw_info.send(uid_field).presence || raise
+      uid = Array(uid).first
     end
 
     user = User.create_or_update_from_hash(options.merge(
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -90,8 +90,15 @@ class SessionsController < ApplicationController
   end
 
   def login(options = {})
+    uid = auth_hash.uid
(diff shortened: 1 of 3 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -90,8 +90,15 @@ def redirect_to_origin_or_default
   end
 
   def login(options = {})
+    uid = auth_hash.uid
+    if ENV['USE_LDAP_UID_AS_EXTERNAL_ID']
+      uid_field = Rails.application.config.samson.ldap.uid
+      uid = auth_hash.extra.raw_info.send(uid_field).presence || auth_hash.uid
+      uid = uid.is_a?(Array) ? uid.first : uid
```
grosser: can simplify to `Array(uid).first`

app/controllers/sessions_controller.rb
```diff
@@ -90,8 +90,15 @@ def redirect_to_origin_or_default
   end
 
   def login(options = {})
+    uid = auth_hash.uid
+    if ENV['USE_LDAP_UID_AS_EXTERNAL_ID']
+      uid_field = Rails.application.config.samson.ldap.uid
+      uid = auth_hash.extra.raw_info.send(uid_field).presence || auth_hash.uid
```
grosser: if you already have this feature turned on, should it maybe fail if it's not set since it would not be working as intended ?

grosser: `auth_hash.extra.raw_info.send(uid_field).presence || raise`

test/controllers/sessions_controller_test.rb
```diff
@@ -257,6 +262,15 @@
         @controller.send(:current_user).must_equal(user)
       end
 
+      it 'logs the user in with USE_LDAP_UID_AS_EXTERNAL_ID' do
+        Rails.application.config.samson.ldap.stub(:uid, 'sAMAccountName') do
+          with_env 'USE_LDAP_UID_AS_EXTERNAL_ID' => 'true' do
```
grosser: can use `USE_LDAP_UID_AS_EXTERNAL_ID : 'true'`

test/controllers/sessions_controller_test.rb
```diff
@@ -257,6 +262,15 @@
...
+            post :ldap
+            @controller.send(:current_user).must_equal(user)
```
grosser: this would still pass if the new code was removed, right ?
... so add something to make sure it does not use the old code

grosser:
isn't that just a bug atm that LDAP_UID is ignored ?

grosser:
so we should use it if it's set without the extra flag/docs ?

jgoerz:
> isn't that just a bug atm that LDAP_UID is ignored ?
Well, technically, LDAP_UID is not being ignored, it's being used by the bind query.  But what is weird, is the returned "auth_hash" puts the distinguished name into the auth_hash.uid attribute.  
https://github.com/omniauth/omniauth-ldap/blob/master/lib/omniauth/strategies/ldap.rb#L17
This is not just unfortunate, it's confusing.  Because if you look here:
https://github.com/omniauth/omniauth-ldap/blob/master/lib/omniauth/strategies/ldap.rb#L66
You'll see that auth_hash.extra.uid, actually gives you the uid.  With the _extra_ data, you can also get a bunch of things that _make sense_, like gid, global_uid (raw hex, needs post processing), samaccountname, userprincipalname, etc.  It's kinda of weird that you use the _ldap_uid_ attribute to search and identify the user, and then they don't return that as the unique identifier for the user... nuts.

jgoerz:
Ready for review/feedback.

app/controllers/sessions_controller.rb
```diff
@@ -90,8 +90,15 @@ def redirect_to_origin_or_default
   end
 
   def login(options = {})
+    uid = auth_hash.uid
```
grosser: this should live in the `else` ?

test/controllers/sessions_controller_test.rb
```diff
@@ -288,6 +293,21 @@
         request.flash[:error].wont_be_nil
       end
     end
+
+    describe 'with LDAP_UID as external_id' do
+      it 'logs the user in' do
+        Rails.application.config.samson.ldap.stub(:uid, 'sAMAccountName') do
+          with_env 'AUTH_LDAP': 'true' do
```
grosser: can combine these 2 `with_env`

grosser: also symbols work: `with_env AUTH_LDAP: 'true'`

app/controllers/sessions_controller.rb
```diff
@@ -90,8 +90,15 @@ def redirect_to_origin_or_default
...
+    if ENV['AUTH_LDAP'] && ENV['USE_LDAP_UID_AS_EXTERNAL_ID']
```
grosser: I think this might fail if a user allows multiple auth methods (edge-case) and this ... so for example a github login comes in and then ldap blows up ... but I guess nobody uses that, so not really an issue

jgoerz: This is a good catch.  I can fix that by checking the auth_hash.provider:
https://github.com/omniauth/omniauth/wiki/Auth-Hash-Schema#schema-10-and-later

grosser:
thx! 🎉

PR #2547, opened 2018-02-07
Wait for scan results when they would block the deploy

ragurney:


grosser:
works!

```
[19:38:44] Waiting for Build /builds/1356 to finish.
[19:39:48] Waiting for GCR scan to finish...
[19:40:38] No vulnerabilities found
```

PR #2550, opened 2018-02-07
PAAS-1277 allow creating GKE clusters via the UI

grosser:
screen shot 2018-02-07 at 1 46 05 pm
screen shot 2018-02-07 at 2 23 52 pm
screen shot 2018-02-07 at 1 45 49 pm

@jonmoter @ragurney 

need to set `CLOUDSDK_CONTAINER_USE_CLIENT_CERTIFICATE=True` or the certs are unuseable


plugins/gcloud/app/controllers/gke_clusters_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  def new
+  end
+
+  def create
+    project = params.require(:gke_cluster).require(:gcp_project)
+    cluster = params.require(:gke_cluster).require(:cluster_name)
+    zone = params.require(:gke_cluster).require(:zone)
+
+    # prepare the file gcloud will write into
+    folder = ENV.fetch("GCLOUD_GKE_CLUSTERS_FOLDER")
```
jonmoter: Might be nice to show an error on the `new` page if this ENV var isn't set, rather than firing off an Airbrake error when you try to create.

jonmoter: Ah, nevermind. I see that you do that check below in the view that makes the button appear.

grosser:
confirmed working in staging :D 

error `get-credentials requires edit permission on foo` was fixed by adding `clusters.getCredentials` to the serviceaccount

PR #2555, opened 2018-02-09
Add Assertible plugin.

zacharyw:
Adds a plugin for notifying [Assertible](https://assertible.com/) of successful deploys.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


.env.example
```diff
@@ -79,6 +79,9 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 
 # AIRBRAKE_API_KEY= # optional, report errors to airbrake
 
+# ASSERTIBLE_SERVICE_KEY= # optional, report deploys to Assertible
```
grosser: maybe move the comment ontop

```
# optional, report deploys to Assertible
...
```

db/migrate/20180209012126_add_notify_assertible_to_stages.rb
```diff
@@ -0,0 +1,5 @@
+class AddNotifyAssertibleToStages < ActiveRecord::Migration[5.1]
+  def change
+    add_column :stages, :notify_assertible, :boolean, default: false, null: false
```
grosser: move to plugins/assertible/db/migrate

plugins/assertible/lib/samson_assertible/samson_plugin.rb
```diff
@@ -0,0 +1,52 @@
...
+module SamsonAssertible
+  class Engine < Rails::Engine
+  end
+
+  class Notification
+    class << self
+      def deliver(deploy)
+        return unless service_key.present?
+        return unless deploy_token.present?
+        return unless deploy.stage.notify_assertible?
```
grosser: fail-fast would be nice here ... if the stage is set to `notify_assertible ` then blow up if key/token is not set ?
`ENV.fetch('ASSERTIBLE_DEPLOY_TOKEN')`

plugins/assertible/test/samson_assertible/samson_plugin_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    subject { SamsonAssertible::Notification.deliver(deploy) }
+
+    let(:url_helpers) { Rails.application.routes.url_helpers }
+    let(:deploy) { deploys(:succeeded_test) }
+
+    before do
+      deploy.stage.update_column(:notify_assertible, true)
+    end
+
+    with_env(ASSERTIBLE_SERVICE_KEY: 'test_token')
```
grosser: can fit both into a single `with_env`

plugins/assertible/test/samson_assertible/samson_plugin_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    context 'When no service key' do
+      with_env(ASSERTIBLE_SERVICE_KEY: nil)
+
+      it 'Does not send notification' do
+        Faraday.expects(:post).never
+        subject
+      end
+    end
```
grosser: can simplify this to:

```Ruby
it "does not send notifications without service key" do
  with_env(ASSERTIBLE_SERVICE_KEY: nil) { subject } 
end
```

(no negative assertion needed since it would blow up if a request was made)

plugins/assertible/test/samson_assertible/samson_plugin_test.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+end
+
+describe :after_deploy do
+  subject { Samson::Hooks.fire :after_deploy, deploy, users(:admin) }
+
+  let(:deploy) { deploys(:succeeded_test) }
+
+  it 'Triggers delivery' do
```
grosser: for all the it/context prefer lowercase `it "triggers deploy"`

zacharyw:
@grosser Thank you for the feedback. I've implemented your suggestions and I think it's ready for another look.

There is one failing test in the build, but it seems to be unrelated.

PR #2561, opened 2018-02-11
Fix Jira issues tab and refactor changeset tabs

StaphSynth:
/cc @zendesk/samson 
/cc @zendesk/wombat 

While looking into adding a new nav tab to the changeset tabs rendered during new deploy creation, I noticed the existing Jira issues one was not working. I fixed this and re-factored the code for generating the changeset nav tabs a little to make it easier for me to add new functionality later.

The `Changeset` model now has a list of what I've called `ATTRIBUTE_TABS` for programmatically generating the tabs. I've modified the changeset partials so they all accept a `Changeset` instance for easier programmatic re-use, as well as rationalizing their names.

screen shot 2018-02-11 at 9 15 35 pm

### Tasks
 - [x] :+1: from team

### References
 - Jira link: Nil

### Risks
- MED: breaks changeset rendering blinding deploys.


app/views/changeset/_risks.html.erb
```diff
@@ -1,9 +1,9 @@
-<% if pull_requests.nil? || pull_requests.empty? %>
+<% if changeset.risky_pull_requests.nil? || changeset.risky_pull_requests.empty? %>
```
ragurney: Would be nice if `changeset.risky_pull_requests` would always return an empty array, or use `blank?`.

ragurney: Since `changeset.risky_pull_requests` is also used in the else, we could DRY this up to use 
```
if pull_requests = changeset.risky_pull_requests.presence
...
```

app/views/deploy_mailer/_commit_info.html.erb
```diff
@@ -1,10 +1,6 @@
...
 
-<h3>Risk</h3>
-<%= render 'changeset/risks', pull_requests: @changeset.risky_pull_requests, type: "deploy" %>
-
-<h3>Commits</h3>
-<%= render 'changeset/commits', changeset: @changeset %>
-
-<h3>Pull Requests</h3>
-<%= render 'changeset/pulls', changeset: @changeset %>
+<% %w(risks commits pull_requests).each do |attribute| %>
```
ragurney: Could we either use the existing constant or even better split the attributes out into a new constant such as `EAGER_ATTRIBUTES` and `LAZY_ATTRIBUTES`.

ragurney: Looking further into this can the commit have `jira_issues` or even `risks`?

app/views/deploys/changeset.html.erb
```diff
@@ -1,21 +1,7 @@
...
-
-<div class="tab-pane<%= @changeset.risks? ? " active" : "" %>" id="risks">
-  <%= render 'changeset/risks', pull_requests: @changeset.risky_pull_requests, type: "deploy" %>
-</div>
-
-<div class="tab-pane" id="jira-issues">
-  <%= render 'changeset/jira_issues', jira_issues: @changeset.jira_issues, type: "deploy" %>
-</div>
+<% Changeset::ATTRIBUTE_TABS.each do |tab| %>
+  <div class="tab-pane<%= (tab == 'risks' && @changeset.risks?) ? ' active' : '' %>" id="<%= tab %>">
```
ragurney: The active tab logic is duplicated in `tab_list`. It would be nice to extract that to a helper method `active_changeset_tab`.

app/views/changeset/_risks.html.erb
StaphSynth: 👍

grosser:
yeah, can be added later, already an improvement

app/views/changeset/_risks.html.erb
StaphSynth: Putting an assignment in an `if` statement obfuscates the meaning here I think - especially since I want the array, not the boolean return value (which would look something like this: `if (risky_prs = changeset.risky_pull_requests).blank?` ). I'm happy to use `blank?` but I don't see the gain in obfuscating the meaning of the code, just to avoid one more attribute method call on `changeset`.

grosser: `if pull_requests = changeset.risky_pull_requests.presence` would return array nor nil

app/views/deploys/changeset.html.erb
```diff
@@ -1,21 +1,7 @@
...
-
-<div class="tab-pane<%= @changeset.risks? ? " active" : "" %>" id="risks">
-  <%= render 'changeset/risks', pull_requests: @changeset.risky_pull_requests, type: "deploy" %>
-</div>
-
-<div class="tab-pane" id="jira-issues">
-  <%= render 'changeset/jira_issues', jira_issues: @changeset.jira_issues, type: "deploy" %>
-</div>
+<% Changeset::ATTRIBUTE_TABS.each do |tab| %>
+  <div class="tab-pane <%= active_changeset_tab(tab, 'risks') %>" id="<%= tab %>">
```
grosser: this used to only be active when there were risks ... now it's always active ?

app/views/changeset/_tab_list.html.erb
```diff
@@ -0,0 +1,7 @@
+<ul class="nav nav-tabs">
+  <% Changeset::ATTRIBUTE_TABS.each do |tab| %>
+    <li class="<%= active_changeset_tab(tab, active) %>">
```
grosser: risks is hardcoded in the other partial and not here ... looks like one of them is wrong ?

app/views/deploys/changeset.html.erb
StaphSynth: Nice catch

app/views/changeset/_tab_list.html.erb
StaphSynth: After a little playing around, I'm thinking of removing this, since there's already tab-switching logic (including default 'active' class) in `deploys.js`. What do you think?

grosser: sounds good

app/views/changeset/_jira_issues.html.erb
```diff
@@ -1,7 +1,7 @@
-<% if jira_issues.any? %>
+<% if changeset.jira_issues.any? %>
```
grosser: this assumes `jira_issues` is cached ... could avoid that assumption and cache-logic by using `if jira_issues = changeset.jira_issues.presence`

app/views/changeset/_risks.html.erb
```diff
@@ -1,9 +1,9 @@
-<% if pull_requests.nil? || pull_requests.empty? %>
+<% if changeset.risky_pull_requests.blank? %>
```
grosser: same here, prefer assign with `.presence`

PR #2562, opened 2018-02-12
Add styles for additions/deletions in files/PRs

ellimist:
Right now, the number of additions and deletions in every pull request
is invisible, as it has the same color as the background, white.
This change makes those numbers visible again.

Result:

screen shot 2018-02-12 at 12 56 33

### Tasks
 - [ ] :+1: from team

### Risks
- Level: None

### DIFF
--- a/app/assets/stylesheets/global.scss
+++ b/app/assets/stylesheets/global.scss
@@ -180,6 +180,24 @@ h3 {
 .label {
   padding: .3em 0.65em .35em;
   border-radius: .2em;
+
+  &.additions {
+    color: green;
+
+    &::before {
+      content: "+";
+      padding: 0 2px;
+    }
+  }
+
+  &.deletions {
+    color: red;
+
+    &::before {
+      content: "-";
+      padding: 0 2px;
+    }
+  }
 }
 
 .label-success {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -52,3 +52,23 @@ a.btn.disabled[title] {
...
+  }
+
+  &.deletions {
+    color: red;
+
+    &::before {
+      content: "-";
+      padding: 0 2px;
+    }
+  }
```
brianburnszd: If it's specific to deploys (seems to be) it could be placed in `deploys.scss`

app/assets/stylesheets/application.scss
```diff
@@ -52,3 +52,23 @@ a.btn.disabled[title] {
...
   font-family: monospace;
 }
+
+table td span {
+  &.additions {
+    color: green;
+
+    &::before {
+      content: "+";
+      padding: 0 2px;
```
brianburnszd: Nice style!

app/assets/stylesheets/application.scss
ellimist: The code is actually shared between both `deploys` and `releases`. I've moved it in the global.css which already has a `.label` definition

pdeuter:
:chicken:

PR #2569, opened 2018-02-13
Avoid potential problems with chrdir

ragurney:


grosser:
I'm not getting the notification emails ... just saw it in jira :/

grosser:
found it .. all notifications were disabled :D

test/lib/samson/command_executor_test.rb
```diff
@@ -17,6 +17,12 @@
       Samson::CommandExecutor.execute("sh", "-c", "echo hello 1>&2", err: '/dev/null', timeout: 1).must_equal [true, ""]
     end
 
+    it "runs in specified directory" do
+      Dir.mktmpdir("foobar") do |dir|
+        Samson::CommandExecutor.execute("pwd", timeout: 1, dir: dir).second.must_match /^.*#{dir}$/
```
grosser: '^.*' <- matches everything
so just do `must_include dir` ?

lib/samson/command_executor.rb
```diff
@@ -6,14 +6,19 @@ class << self
...
-      def execute(*command, timeout:, whitelist_env: [], env: {}, err: [:child, :out])
+      def execute(*command, timeout:, whitelist_env: [], env: {}, err: [:child, :out], dir: nil)
         raise ArgumentError, "Positive timeout required" if timeout <= 0
         env = ENV.to_h.slice(*whitelist_env).merge(env)
         pio = nil
+        popen_options = {
+          unsetenv_others: true,
+          err: err,
+          chdir: dir
+        }.reject { |_k, v| v.nil? }
```
grosser: prefer to set things if needed instead of building and then rejecting
```Ruby
popen_options = {err: err, unsetenv_others: true}
popen_options[:chdir] = dir if dir
```

grosser:
```
>> puts Dir.pwd; Thread.new { IO.popen 'sleep 10', chdir: '/' }; sleep 1; puts Dir.pwd
/Users/mgrosser
/Users/mgrosser
```

seems to work nicely and not mess with the directory :)

the original chdir is also thread-exclusive, so idk why it ever warned ... still good to fix it ...

```
puts Dir.pwd; Thread.new { Dir.chdir("/") { IO.popen 'sleep 10' } }; sleep 1; puts Dir.pwd
/Users/mgrosser
/Users/mgrosser

>> puts Dir.pwd; Thread.new { Dir.chdir("/") { IO.popen 'sleep 10' } }; Dir.chdir("/tmp") { sleep 1; puts Dir.pwd }
/Users/mgrosser
(irb):2: warning: conflicting chdir during another chdir block
/private/tmp
```

grosser:
... ideally add a cleanliness test that checks Dir.chdir is not used in code too so this won't pop up again

PR #2573, opened 2018-02-13
unify build finding logic

grosser:
@jonmoter @ragurney 

WIP ... but where I think we need to go

### DIFF
--- a/lib/samson/build_finder.rb
+++ b/lib/samson/build_finder.rb
@@ -12,12 +12,12 @@ module Samson
   class BuildFinder
     TICK = 2.seconds
 
-    def initialize(output, job, reference, images: nil)
+    def initialize(output, job, reference, build_selectors: nil)
       @output = output
       @job = job
       @reference = reference
       @cancelled = false
-      @images = images
+      @build_selectors = build_selectors
     end
 
     # deploy was cancelled, so finish up as fast as possible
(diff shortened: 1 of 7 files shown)


lib/samson/build_finder.rb
```diff
@@ -26,51 +26,55 @@ def cancelled!
...
+      needed = (@images ? @images : @job.project.dockerfile_list.map { |d| [d, @job.project.docker_image(dockerfile)] })
+      build_disabled = @job.project.docker_image_building_disabled?
+      all = []
+
+      wait_for_build_creation do |last_try|
+        needed.delete_if do |selector|
+          found = detect_build_by_selector!(possible_builds, *selector, fail: (last_try && build_disabled))
+          if found
+            all << found
+          elsif last_try && !build_disabled
```
jonmoter: Kindof duplicated

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -69,14 +69,26 @@ def set_secrets
...
 
     private
 
+    def build_selector_for_container(container)
+      if project.docker_image_building_disabled?
+        # TODO: weird that there is no opt-out here
+        [nil, container.fetch(:image)]
+      else
+        selected = container[:"samson/dockerfile"] || 'Dockerfile'
+        next if selected == 'none'
```
jonmoter: This is causing an error. Can't use next when not in a loop.

grosser: just need your time for the review, I can debug travis ;)

lib/samson/build_finder.rb
```diff
@@ -26,51 +26,56 @@ def cancelled!
...
+            all << create_build(dockerfile)
+          end
+        end
+        needed.empty? # stop the waiting when we got everything
+      end
+
+      all
+    end
+
+    def self.detect_build_by_selector!(builds, dockerfile, image, fail:)
```
ragurney: Would this be more clear as:
```ruby
    class << self
      def detect_build_by_selector!(builds, dockerfile, image, fail:)
        image_name = image.split('/').last.split(/[:@]/, 2).first if image

        if (matching_build = find_matching_build(builds, dockerfile, image_name))
          matching_build
        elsif fail
          raise(
            Samson::Hooks::UserError,
            "Did not find build for dockerfile #{dockerfile.inspect} or image_name #{image_name.inspect}.\n" \
            "Found builds: #{builds.map { |b| [b.dockerfile, b.image_name] }.uniq.inspect}."
          )
        end
      end

      private

      def find_matching_build(builds, dockerfile, image_name)
        builds.detect do |b|
          (image_name && b.image_name == image_name) || (dockerfile && b.dockerfile == dockerfile)
        end
      end
    end
```
?

grosser:
kinda similar ... maybe flattening it our with a early return/raise helps,
I'll play with this a bit

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -69,14 +69,25 @@ def set_secrets
...
+        [selected, nil]
```
ragurney: Would this break if someone has their dockerfile as `Dockerfile` and specifies an image name on their build? Then they would have a build with a [nil docker file](https://github.com/zendesk/samson/blob/master/app/models/build.rb#L76) and we wouldn't be able to find a matching build.

jonmoter:
I'd rather focus on getting the fix out and making it look cleaner later. There are devs who are blocked by this bug, and unable to deploy.

grosser:
oh yeah ... all green without having breaking any tests, I got high hopes for this one :D

plugins/kubernetes/app/models/kubernetes/template_filler.rb
grosser: yes, builds with nil `dockerfile` column cannot be found when not using `external images` feature, but that's a separate discussion/feature ... trying to keep this refactoring only :)

PR #2589, opened 2018-02-17
a build with failed external status is not active

grosser:
had a deploy hanging forever even after we killed unresponsive builds ...


test/models/build_test.rb
```diff
@@ -229,8 +234,6 @@ def valid_build(attributes = {})
     end
 
     it 'cancels builds that have been running for too long' do
-      puts build.created_at
```
ragurney: 😓

grosser: thought that was mine, did not even blame it :D

PR #2592, opened 2018-02-21
Refactor project docker settings to make them more clear

ragurney:
@zendesk/devex

### DIFF
--- a/app/helpers/projects_helper.rb
+++ b/app/helpers/projects_helper.rb
@@ -43,4 +43,13 @@ module ProjectsHelper
       ""
     end
   end
+
+  def docker_build_methods_help_text
+    help_texts = Project::DOCKER_BUILD_METHODS.map do |method|
+      if method[:help_text].present?
+        content_tag(:b, "#{method[:label]}: ") << method[:help_text]
+      end
+    end.compact
+    safe_join(help_texts, "<br/><br/>".html_safe)
+  end
 end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -1,5 +1,20 @@
 # frozen_string_literal: true
 class Project < ActiveRecord::Base
+  # Used as part of temporary abstract attribute for docker build radio buttons
+  DEFAULT_DOCKER_BUILD_METHOD = 'samson'
+  DOCKER_BUILD_METHODS = [
+    {
+      label: 'Docker images build externally',
+      method: 'docker_image_building_disabled',
+      help_text: 'Disable local building of docker images, they must be added via api.'
+    },
+    {
+      label: 'Samson manages docker images',
+      method: DEFAULT_DOCKER_BUILD_METHOD,
+      help_text: ''
+    }
+  ] + Samson::Hooks.fire(:project_docker_build_method_options).flatten(1)
+
   has_soft_deletion default_scope: true unless self < SoftDeletion::Core
   audited
 
@@ -55,6 +70,22 @@ class Project < ActiveRecord::Base
     end
     scope
   }
+
+  # Forms use abstract "docker_build_method " attribute
+  def docker_build_method
+    active = DOCKER_BUILD_METHODS.map { |h| h.fetch(:method) }.detect do |build_method|
+      build_method != DEFAULT_DOCKER_BUILD_METHOD && send("#{build_method}?")
+    end
+
+    active || DEFAULT_DOCKER_BUILD_METHOD
+  end
+
+  def docker_build_method=(selected_method)
+    DOCKER_BUILD_METHODS.each do |method|
+      send("#{method[:method]}=", false) unless method[:method] == DEFAULT_DOCKER_BUILD_METHOD
+    end
+    send("#{selected_method}=", true) unless selected_method == DEFAULT_DOCKER_BUILD_METHOD
+  end
 
   def docker_repo(registry, dockerfile)
     File.join(registry.base, docker_image(dockerfile))
--- a/app/views/projects/_form.html.erb
+++ b/app/views/projects/_form.html.erb
@@ -15,15 +15,15 @@
     <%= form.input :release_source, help: 'Selects which type of webhook will create a release' do %>
       <%= form.collection_select :release_source, webhook_sources_for_select(Samson::Integration::SOURCES, none: true), :second, :first, {}, class: "form-control" %>
     <% end %>
-    <% if ENV['DOCKER_FEATURE'] %>
-      <%= form.input :docker_release_branch, help: 'New commits on this branch will cause a docker image to be built when a webhook arrives.' %>
-      <%= form.input :dockerfiles, help: 'Space separated list of dockerfiles to build, default: Dockerfile' %>
-      <%= form.input :docker_image_building_disabled, label: "Docker images built externally", as: :check_box, help: 'Disable local building of docker images, they must be added via api.' %>
-    <% end %>
     <% if DeployGroup.enabled? %>
       <%= form.input :include_new_deploy_groups, as: :check_box, label: 'Include in New Deploy Groups', help: 'The "Create all stages" button for a deploy group will create a stage for this project.' %>
     <% end %>
     <%= Samson::Hooks.render_views(:project_form_checkbox, self, form: form) %>
+
+    <% if ENV['DOCKER_FEATURE'] %>
+      <%= render 'docker_fields', form: form %>
+    <% end %>
+
     <%= render 'shared/dashboard_field', form: form %>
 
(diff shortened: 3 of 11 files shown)


app/helpers/projects_helper.rb
```diff
@@ -43,4 +43,10 @@ def repository_web_link(project)
       ""
     end
   end
+
+  def docker_build_methods_help_text
+    Project::DOCKER_BUILD_METHODS.map do |method|
+      ["<b>#{method[:label]}</b>: #{method[:help_text]}"] unless method[:help_text].blank?
```
grosser: unless blank -> if present

app/helpers/projects_helper.rb
```diff
@@ -43,4 +43,10 @@ def repository_web_link(project)
...
+    end.compact.join("<br/><br/>").html_safe
```
grosser: XSS! :)

```
["<hacks>"].join("<br>").html_safe
=> "<hacks>"
```

app/models/project.rb
```diff
@@ -56,6 +69,24 @@ class Project < ActiveRecord::Base
     scope
   }
 
+  # Methods used to negotiate temporary abstract attribute `docker_build_method`
```
grosser: `# forms use abstract "docker_build_method " attribute`

app/models/project.rb
```diff
@@ -56,6 +69,24 @@ class Project < ActiveRecord::Base
...
+  def docker_build_method
+    result = DOCKER_BUILD_METHODS.detect do |build_method|
+      build_method[:method] != 'samson' && send("#{build_method[:method]}?")
+    end
+
+    result&.[](:method) || 'samson'
+  end
+
+  def docker_build_method=(selected_method)
+    deselected_methods = DOCKER_BUILD_METHODS.reject { |bm| bm.include?(selected_method) }
```
grosser: how is this include supposed to work on a hash ? ... missing test ?

app/models/project.rb
```diff
@@ -56,6 +69,24 @@ class Project < ActiveRecord::Base
...
   }
 
+  # Methods used to negotiate temporary abstract attribute `docker_build_method`
+  def docker_build_method
+    result = DOCKER_BUILD_METHODS.detect do |build_method|
+      build_method[:method] != 'samson' && send("#{build_method[:method]}?")
+    end
+
+    result&.[](:method) || 'samson'
+  end
```
grosser: can map to method early to avoid hash access:

```Ruby
default = "samson"
DOCKER_BUILD_METHODS.map { |h| h.fetch(:method) }.detect { |m| m != default && send("#{build_method[:method]}?") } || default
```

app/models/project.rb
```diff
@@ -56,6 +69,24 @@ class Project < ActiveRecord::Base
...
+
+  def docker_build_method=(selected_method)
+    deselected_methods = DOCKER_BUILD_METHODS.reject { |bm| bm.include?(selected_method) }
+
+    deselected_methods.each do |deselected_method|
+      send("#{deselected_method[:method]}=", false) unless deselected_method[:method] == 'samson'
+    end
+    send("#{selected_method}=", true) unless selected_method == 'samson'
+  end
```
grosser: too much 'samson' flying around ... time for a constant ?

app/views/projects/_form.html.erb
```diff
@@ -15,15 +15,15 @@
...
-      <%= form.input :dockerfiles, help: 'Space separated list of dockerfiles to build, default: Dockerfile' %>
-      <%= form.input :docker_image_building_disabled, label: "Docker images built externally", as: :check_box, help: 'Disable local building of docker images, they must be added via api.' %>
-    <% end %>
     <% if DeployGroup.enabled? %>
       <%= form.input :include_new_deploy_groups, as: :check_box, label: 'Include in New Deploy Groups', help: 'The "Create all stages" button for a deploy group will create a stage for this project.' %>
     <% end %>
     <%= Samson::Hooks.render_views(:project_form_checkbox, self, form: form) %>
+
+    <% if ENV['DOCKER_FEATURE'] %>
+      <%= render 'docker_fields', form: form %>
```
grosser: can put the if inside the  `<%=` too

test/models/project_test.rb
```diff
@@ -358,6 +358,37 @@ def clone_repository(project)
...
+      project.update_column(:docker_image_building_disabled, true)
+      project.docker_build_method.must_equal 'docker_image_building_disabled'
+
+      project.update_columns(docker_image_building_disabled: false, build_with_gcb: true)
+      project.docker_build_method.must_equal 'build_with_gcb'
+    end
+  end
+
+  describe '#docker_build_method=' do
+    it 'updates attributes based on docker_build_method' do
```
grosser: same here

app/helpers/projects_helper.rb
ragurney: Since both the `label` and `help_text` aren't user supplied, is this still a concern?

test/models/project_test.rb
```diff
@@ -358,6 +358,37 @@ def clone_repository(project)
     end
   end
 
+  describe '#docker_build_method' do
+    it 'gets the correct docker_build_method' do
```
grosser: prefer 1 block per test

```Ruby
it "is samson when nothing is selected"
it "is disabled when external was selected"
it "is gcb when gcb was selected"


```

test/helpers/projects_helper_test.rb
```diff
@@ -107,4 +107,13 @@ def config_mock
...
     end
   end
+
+  describe "#docker_build_methods_help_text" do
+    it 'returns help text html for build methods' do
+      result = docker_build_methods_help_text
+      result.must_include '<b>Docker images build externally</b>'
+      result.must_include '<b>Build docker with GCB CLI</b>'
+      result.wont_include '<b>Samson manages docker images</b>'
+    end
```
grosser: check for html_safe too

app/models/project.rb
```diff
@@ -1,5 +1,19 @@
...
+      label: 'Docker images build externally',
+      method: 'docker_image_building_disabled',
+      help_text: 'Disable local building of docker images, they must be added via api.'
+    },
+    {
+      label: 'Samson manages docker images',
+      method: 'samson',
+      help_text: ''
+    }
+  ] + Samson::Hooks.fire(:project_docker_build_method_options).flatten
```
grosser: `.flatten` is usually a bad idea since it's expensive ... `.flatten(1)` is what you want here

`[[[1]]].flatten(1) -> [[1]]`

app/helpers/projects_helper.rb
grosser:
```Ruby
html = "".html_safe
foo.each do
```

grosser:
train wifi is messing up my comments :D

app/models/project.rb
ragurney: Tests passed because it would just set all methods to false, then set the one `selected_method` to true. I don't think we really need `deselected_methods` since it only saves one assignment.

app/helpers/projects_helper.rb
grosser: not really a concern here, but bad practice and might end up getting copy-pasted or learned by other devs ... so avoid :)

https://apidock.com/rails/ActionView/Helpers/OutputSafetyHelper/safe_join

app/helpers/projects_helper.rb
```diff
@@ -43,4 +43,13 @@ def repository_web_link(project)
       ""
     end
   end
+
+  def docker_build_methods_help_text
+    help_texts = Project::DOCKER_BUILD_METHODS.map do |method|
+      if method[:help_text].present?
+        "<b>".html_safe << "#{method[:label]}:" << '</b>'.html_safe << " #{method[:help_text]}"
```
grosser: can do `content_tag(:b, "#{method[:label]}: ") << method[:help_text]`

app/models/project.rb
```diff
@@ -33,7 +48,6 @@ class Project < ActiveRecord::Base
 
   belongs_to :build_command, class_name: 'Command', optional: true
 
-  # For permission checks on callbacks. Currently used in private plugins.
```
grosser: add back the comment ?

app/models/project.rb
```diff
@@ -56,6 +70,22 @@ class Project < ActiveRecord::Base
     scope
   }
 
+  # Forms use abstract "docker_build_method " attribute
+  def docker_build_method
+    result = DOCKER_BUILD_METHODS.map { |h| h.fetch(:method) }.detect do |build_method|
```
grosser: `result` -> `active`

PR #2593, opened 2018-02-21
Add a new plugin

yifeiwu:
* Add description here
* Remember to add unit tests

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High


grosser:
maybe some details on what this does ? :)

yifeiwu:
Ahh sorry, work in progress. I'm trying to get this to fetch the jenkin's stagingstatus view and boil it down to a yea or nay on production deployments. Unfortunately, I don't have a place to test it out yet :\

grosser:
staging might work ?
or test only the api calls from `rails console` to see what they would return

yifeiwu:
@grosser, I'd love to get some input on changes to make :)

grosser:
This would fit very nicely into "Commit status" ... which shows status checks when entering a commit.

grosser:
The name canideploy is meaningless, users should be able to quickly tell if they want/can use a plugin, maybe something like "jenkins status checker" ... we already have a jenkins plugin that does something very similar (display status after deploy) ... so adding to that would also make a lot of sense to me.

plugins/canideploy/app/models/canideploy.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+require 'jenkins_api_client'
+require 'singleton'
+# checks jenkins staging status and publishes unstable builds checklist
+class Canideploy
+  include Singleton
+
+  URL = ENV['JENKINS_URL'] # 'https://jenkins.zendesk.com'
```
grosser: `.yourcompany.com`

plugins/canideploy/app/models/canideploy.rb
```diff
@@ -0,0 +1,50 @@
...
+  USERNAME = ENV['JENKINS_USERNAME']
+  API_KEY = ENV['JENKINS_API_KEY']
+  JENKINS_CHECKLIST = ENV['JENKINS_STAGING_VIEW'] # '/view/StagingStatus'
+
+  def initialize
+    @client ||= \
```
grosser: it's a singleton, so why does it need `@client ||=` ?

plugins/canideploy/app/models/canideploy.rb
```diff
@@ -0,0 +1,50 @@
...
+      JenkinsApi::Client.new(
+        server_url: URL,
+        username: USERNAME,
+        password: API_KEY
+      )
+  end
+
+  def checklist
```
grosser: some pass/fail boolean would be nice so we can not show this when it passes (reduce clutter)

plugins/canideploy/app/models/canideploy.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    items
+  end
+
+  private
+
+  def check_jenkins
+    @client.api_get_request(JENKINS_CHECKLIST)
+  rescue Timeout::Error => e
+    Rails.logger.warn "Timeout while fetching checklist.  #{e.class} #{e}"
```
grosser: `while fetching jenkins checklist` <- so it's more clear where this comes from ... we got lots of logs :)

plugins/canideploy/app/models/canideploy.rb
```diff
@@ -0,0 +1,50 @@
...
+    return nil
+  rescue JenkinsApi::Exceptions::ApiException => e
```
grosser: same as above ... `rescue Timeout::Error, JenkinsApi::Exceptions::ApiException`

plugins/canideploy/app/views/samson_canideploy/_deploys_new.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if stage.production? && ENV['CANIDEPLOY_ENABLED'] %>
```
grosser: check the cheap one first

plugins/canideploy/test/models/canideploy_test.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+# rubocop:disable Metrics/LineLength
+require_relative '../../test/test_helper'
+
+SingleCov.covered! uncovered: 0
```
grosser: can leave out the `uncovered: 0`

plugins/canideploy/test/models/canideploy_test.rb
```diff
@@ -0,0 +1,67 @@
...
+
```
grosser: avoid newlines between lets and empty before blocks

plugins/canideploy/test/models/canideploy_test.rb
```diff
@@ -0,0 +1,67 @@
...
+  before do
+  end
+
+  describe "initialized" do
```
grosser: prefer method names as describes for easy matching between code and tests: `describe ".instance"`

plugins/canideploy/test/models/canideploy_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    it "has an available singleton instance" do
+      assert_instance_of Canideploy, Canideploy.instance
+    end
+  end
+
+  describe "checklist" do
+    before do
+      stub_request(:get, "http://www.test-url.com/api/json").
+        with(headers: {'Accept' => '*/*', 'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization' => 'Basic dXNlckB0ZXN0LmNvbTpqYXBpa2V5', 'User-Agent' => 'Ruby'}).
```
grosser: don't need that `with`

plugins/canideploy/test/models/canideploy_test.rb
```diff
@@ -0,0 +1,67 @@
...
+        to_return(status: 200, body: json_response)
```
grosser: `status: 200` is the default, can leave out

plugins/canideploy/test/models/canideploy_test.rb
grosser: lets can be nested in the checlist test, nothing else uses them ?

plugins/canideploy/test/models/canideploy_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
```
grosser: prefer test setup with less indirection:

```
it "does something good" do
  some_helper_method(good_response)
  assert foo
end

it "does something bad" do
  some_helper_method(bad_response)
  refute foo
end

```

grosser:
checking jenkins status sounds like a good idea, but it's basically backwards from the expected/regular flow ... jenkins should set a github commit status so devs have visibility and then samsons commit-status feature would already show them ... so might be better to invest time in getting that working then more samson features

yifeiwu:
Thanks for the review and I'll take a look at how the commit status can be changed

plugins/canideploy/app/models/canideploy.rb
yifeiwu: I think we still need to memoize the client connection?

grosser:
rebase to get doorkeeper update from master

grosser:
... will take a few mins, need to fix up the PR

app/views/deploys/new.html.erb
```diff
@@ -40,6 +40,8 @@
         </div>
       </div>
 
+      <%= Samson::Hooks.render_views(:deploy_new, self, project: @project, form: form, stage: @stage) %>
+
       <%= Samson::Hooks.render_views(:deploy_form, self, project: @project, form: form) %>
```
grosser: why not put this into the deploy_form ?
... can add the `stage` to that hook, it's just an options hash

lib/samson/hooks.rb
```diff
@@ -13,6 +13,7 @@ class UserError < StandardError
       :build_button,
       :build_new,
       :build_show,
+      :deploy_new,
```
grosser: then we don't need this anymore

plugins/jenkins_status_checker/README.md
```diff
@@ -0,0 +1 @@
+This plugin allows you to see a list of failing jenkins jobs that may block a deploy prior to deploying. This only shows for production, same rules as the buddy check. It shares the jenkins credentials of the jenkins plugin.
```
grosser: "Shows failing jenkins jobs when users try to create a new deploy."

grosser: mention the settings it needs (including JENKINS_STATUS_CHECKER) and mention JENKINS_STATUS_CHECKER in .env.example

plugins/jenkins_status_checker/app/models/jenkins_status_checker.rb
```diff
@@ -0,0 +1,49 @@
...
+# checks jenkins staging status and publishes unstable builds checklist
+class JenkinsStatusChecker
+  include Singleton
+
+  URL = ENV['JENKINS_URL'] # 'https://jenkins.yourcompany.com'
+  USERNAME = ENV['JENKINS_USERNAME']
+  API_KEY = ENV['JENKINS_API_KEY']
+  JENKINS_CHECKLIST = ENV['JENKINS_STAGING_VIEW'] # '/view/StagingStatus'
+
+  def checklist
```
grosser: should fail when one of the required settings are not set ... atm it fails in strange ways with nil errors

for example:
```
username: ENV.fetch('JENKINS_USERNAME'),
```

plugins/jenkins_status_checker/app/views/samson_jenkins_status_checker/_deploys_new.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if ENV['JENKINS_STATUS_CHECKER'] && stage.production? %>
+  <div class="form-group">
+    <div class="alert col-lg-5 col-lg-offset-2">
+      <p>Potential Deploy Blockers:</p>
+      <% JenkinsStatusChecker.instance.checklist.each do |item| %>
+        <ul>
```
grosser: how about one ul with multiple li ?

.env.example
```diff
@@ -168,6 +168,13 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # JENKINS_USERNAME= # user id
 # JENKINS_API_KEY= # API Token from user / Configure page
 
+## Jenkins Status Checker Plugin
+# JENKINS_STATUS_CHECKER=true
+# JENKINS_URL=https://jenkins.yourcompany.com
+# JENKINS_USERNAME=someguy@yourcompany.com
+# JENKINS_API_KEY=somekey
+# JENKINS_STAGING_VIEW=/view/StagingStatus
```
grosser: does not need to be staging ... maybe do `JENKINS_STATUS_CHECKER=/view/StagingStatus` to solve both with a single flag ?

plugins/jenkins_status_checker/README.md
```diff
@@ -0,0 +1,12 @@
...
+This plugin shows failing jenkins jobs when users try to create a new deploy. This only shows for production, same rules as the buddy check. It shares the jenkins credentials of the jenkins plugin.
+
+### Setup
+
+Required Environment Variables (see .env.example)
+JENKINS_STATUS_CHECKER # set to true for projects that require this check
+JENKINS_URL            # 'https://jenkins.yourcompany.com', should already be set by jenkins plugin
+JENKINS_USERNAME	   # should already be set by jenkins plugin
+JENKINS_API_KEY		   # should already be set by jenkins plugin
+JENKINS_STAGING_VIEW   # '/view/StagingStatus'
```
grosser: line up the `#`

grosser: `should already be set by` -> `is shared with` ... we don't know that everyone uses both plugins

grosser:
does every production deploy for every project need to show this ? ... sounds like a lot of noise ... also decent chance it will break/slow everything if jenkins goes down.
maybe make it opt-in per project with a boolean flag ?

yifeiwu:
> does every production deploy for every project need to show this ? ... sounds like a lot of noise ... also decent chance it will break/slow everything if jenkins goes down.
maybe make it opt-in per project with a boolean flag ?

https://github.com/zendesk/samson/pull/2593/files#diff-8b591f6ef77959ee3f527a89a28c26caR1
I thought that if the env var wasn't defined in the project, this would not run.

grosser:
ENV vars are global to samson ... not per project
you'd need a new boolean column for the projects table to make it opt-in per project

yifeiwu:
I've added a column to projects `show_jenkins_status_checker` and added a checkbox to the builds page for enabling it for particular projects.

plugins/jenkins_status_checker/app/models/jenkins_status_checker.rb
```diff
@@ -0,0 +1,47 @@
...
+        items << "#{job["name"]} is not stable" if job["color"] != "blue"
+      end
+    end
+    items << "All projects stable!" if items.empty?
+    items
+  end
+
+  private
+
+  def check_jenkins
```
grosser: lots of code/tests here that has very little usability impact, would be better to show the errors to the user

```
def check_jenkins
  client.get
rescue
  "Error: #{$!}"
end

...

if result.is_a?(String) # error
  items << error
else
```

plugins/jenkins_status_checker/app/views/samson_jenkins_status_checker/_deploys_form.html.erb
```diff
@@ -0,0 +1,13 @@
...
+      <p>Jenkins Status Checker:</p>
+      <ul>
+        <% JenkinsStatusChecker.instance.checklist.each do |item| %>
+          <li><%= item %></li>
+        <% end %>
+      </ul>
+    </div>
+  </div>
+<% end %>
+
```
grosser: neat view ❤️

plugins/jenkins_status_checker/lib/samson_jenkins_status_checker/samson_plugin.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+module SamsonJenkinsStatusChecker
+  class Engine < Rails::Engine
+  end
+end
+
+Samson::Hooks.view :deploy_form, "samson_jenkins_status_checker/deploys_form"
```
grosser: deploys_form -> deploy_form

plugins/jenkins_status_checker/app/views/samson_jenkins_status_checker/_deploys_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if ENV['JENKINS_STATUS_CHECKER'] && stage.production? && project.show_jenkins_status_checker %>
```
grosser: can it use `@stage` here and then remove the `stage: @stage` passing logic in app/views/deploys/new.html.erb ?

yifeiwu:
@grosser, I've updated the error handling.

plugins/jenkins_status_checker/app/views/samson_jenkins_status_checker/_project_form_checkbox.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if ENV['JENKINS_STATUS_CHECKER'] %>
+  <%= form.input :jenkins_status_checker, label: 'Enable Jenkins Status Checker', as: :check_box %>
```
grosser: Could be "Show Jenkins Status on new deploy page"

PR #2601, opened 2018-02-22
Refactor error reporting to use common interface and create Airbrake backend plugin

ragurney:
TODO: change monitors from airbrake_error_sent -> errors.count


lib/error_notifier.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module ErrorNotifier
+  class << self
+    def notify(exception, options = {})
+      debug_info = Samson::Hooks.fire(:exception, exception, options).compact.first
```
grosser: might be better to do `.detect { |m| m.is_a?(String) }` to avoid caching things like the `false` from statsd.rb or whatever error handler does not support this feature

config/initializers/statsd.rb
```diff
@@ -52,9 +52,8 @@ class << Samson
 end
 
 # test: enable local airbrake, see airbrake.rb
-# will not report for ignored errors from airbrake.rb
-Airbrake.add_filter do |notice|
-  notice[:errors].each do |error|
-    Samson.statsd.increment "airbrake_error_sent", tags: ["class:#{error[:type]}"]
-  end
+# will not report for ignored errors
```
grosser: this is not true anymore :(
... also should say `from  ignored_errors.rb`

PR #2615, opened 2018-03-01
show duration for external builds too

grosser:


app/models/build.rb
```diff
@@ -67,6 +67,14 @@ def active?
     end
   end
 
+  def duration
+    if external?
+      updated_at - created_at if created_at.to_i != updated_at.to_i
```
ragurney: `unless created_at.to_i == updated_at.to_i`?

PR #2616, opened 2018-03-02
Support new kubernetes initContainer syntax

ragurney:

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/cluster.rb
+++ b/plugins/kubernetes/app/models/kubernetes/cluster.rb
@@ -59,6 +59,13 @@ module Kubernetes
       []
     end
 
+    def server_version
+      version = Rails.cache.fetch(cache_key, expires_in: 1.hour) do
+        JSON.parse(client.create_rest_client('version').get.body).fetch('gitVersion')[1..-1]
+      end
+      Gem::Version.new(version)
+    end
+
     private
 
     def connection_valid?
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -47,11 +47,9 @@ module Kubernetes
           set_secrets
           set_image_pull_secrets
           set_resource_blue_green if blue_green_color
+          set_init_containers
         end
-
-        hash = template
-        Rails.logger.info "Created Kubernetes hash: #{hash.to_json}"
-        hash
+        template
       end
     end
 
@@ -69,8 +67,7 @@ module Kubernetes
     end
 
     def build_selectors
-      all = containers
-      modify_init_container { |containers| all += containers }
+      all = containers + init_containers
       all.map { |c| build_selector_for_container(c) }.compact
     end
 
@@ -203,20 +200,18 @@ module Kubernetes
     # /vaultauth is a secrets volume in the cluster
     # /secretkeys are where the annotations from the config are mounted
     def set_secret_puller
-      secret_vol = { mountPath: "/secrets", name: "secrets-volume" }
-      modify_init_container do |containers|
-        containers.unshift(
-          image: SECRET_PULLER_IMAGE,
-          imagePullPolicy: 'IfNotPresent',
-          name: 'secret-puller',
-          volumeMounts: [
-            { mountPath: "/vault-auth", name: "vaultauth" },
-            { mountPath: "/secretkeys", name: "secretkeys" },
-            secret_vol
-          ],
-          env: vault_env
-        )
-      end
+      secret_vol = {mountPath: "/secrets", name: "secrets-volume"}
+      init_containers.unshift(
+        image: SECRET_PULLER_IMAGE,
+        imagePullPolicy: 'IfNotPresent',
+        name: 'secret-puller',
+        volumeMounts: [
+          {mountPath: "/vault-auth", name: "vaultauth"},
+          {mountPath: "/secretkeys", name: "secretkeys"},
+          secret_vol
+        ],
+        env: vault_env
+      )
 
       # share secrets volume between all containers
       containers.each do |container|
@@ -238,11 +233,20 @@ module Kubernetes
 
     # Init containers are stored as a json annotation
     # see http://kubernetes.io/docs/user-guide/production-pods/#handling-initialization
-    def modify_init_container
+    def set_init_containers
+      return if init_containers.empty?
       key = Kubernetes::Api::Pod::INIT_CONTAINER_KEY
-      init_containers = JSON.parse(annotations[key] || '[]', symbolize_names: true)
-      yield init_containers
-      annotations[key] = JSON.pretty_generate(init_containers) if init_containers.any?
+      if init_containers_in_beta?
+        pod_template.dig_set([:spec, :initContainers], init_containers)
+        annotations.delete(key)
+      else
+        annotations[key] = JSON.pretty_generate(init_containers)
+        pod_template[:spec].delete(:initContainers)
+      end
+    end
+
+    def init_containers_in_beta?
+      @doc.deploy_group.kubernetes_cluster.server_version >= Gem::Version.new('1.6.0')
     end
 
     # This key replaces the default kubernetes key: 'deployment.kubernetes.io/podTemplateHash'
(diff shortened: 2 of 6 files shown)


grosser:
a bit more specific would be nice like "support new kubernetes initContainer syntax"

grosser:
parse/generate the json 3 or so times, also asks for version 3 times ... seems kin da inefficient :(

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -22,37 +22,43 @@ def to_hash(verification: false)
...
-          set_resource_blue_green if blue_green_color
         end
-
-        hash = template
-        Rails.logger.info "Created Kubernetes hash: #{hash.to_json}"
-        hash
       end
+
+      hash = template
+      Rails.logger.info "Created Kubernetes hash: #{hash.to_json}"
```
grosser: don't need these big hashes in logs ... they are stored in the release-doc anyway

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative "../../test_helper"
 
-SingleCov.covered! uncovered: 3
+SingleCov.covered! uncovered: 4
```
grosser: this number is supposed to go down :D

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -357,6 +357,46 @@ def add_init_container(container)
         end
       end
 
+      describe '#modify_init_container' do
+        def add_init_contnainer_new_syntax(container)
+          raw_template[:spec][:template][:spec][:initContainers] = [container].to_json
+        end
+
+        let(:init_containers) do
```
grosser: `let :spec_init_containers`

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -357,6 +357,46 @@ def add_init_container(container)
...
+
+        it 'sets init containers in spec if using >= 1.6.0 k8s server version' do
+          stub_request(:get, 'http://foobar.server/version').to_return(body: '{"gitVersion": "v1.6.0"}')
+
+          add_init_contnainer_new_syntax("samson/dockerfile": 'Dockerfile')
+          spec_init_containers[0].must_equal("samson/dockerfile" => "Dockerfile", "image" => image)
+        end
+
+        it 'sets init containers using old syntax to new syntax if using >= 1.6.0 k8s server version' do
+          stub_request(:get, 'http://foobar.server/version').to_return(body: '{"gitVersion": "v1.6.0"}')
```
grosser: copy-pasted ... use a describe

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -423,6 +439,11 @@ def set_pre_stop
       end
     end
 
+    def init_containers
```
grosser: I think we have to delete them from annotations and spec to avoid errors from unexpected keys being used

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -240,9 +243,25 @@ def set_secret_puller
...
+        annotations.delete(key)
+      else
+        annotations[key] = JSON.pretty_generate(@init_containers) if @init_containers.any?
+        pod_template[:spec].delete(:initContainers)
+      end
+    end
+
+    def init_containers_in_beta?
+      version = JSON.parse(
+        @doc.deploy_group.kubernetes_cluster.client.create_rest_client('version').get.body
```
jonmoter: Let's cache this value for each kubernetes cluster, at least for a minute or two.

Otherwise we'll end up calling out to the cluster each time we render a template that uses secrets.

jonmoter: Also extract to a method, so you could call `@doc.deploy_group.kubernetes_cluster.server_version`, which would have the caching logic, and return a `Gem::Version` object.

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -423,6 +439,11 @@ def set_pre_stop
       end
     end
 
+    def init_containers
+      (JSON.parse(annotations[Kubernetes::Api::Pod::INIT_CONTAINER_KEY] || '[]', symbolize_names: true) +
+        JSON.parse((pod_template.dig(:spec, :initContainers) || '[]'))) || []
```
jonmoter: I don't think you'd need to do a `JSON.parse` on `spec.initContainers`.  Only in the annotations do you have a JSON string in the middle of your YAML file.  For `spec.initContainers`, it's normal YAML structure.

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
ragurney: Thought it was a bug like we saw before but I found the branch that wasn't covered.

plugins/kubernetes/app/models/kubernetes/template_filler.rb
ragurney: You mean something like what I'm doing [here](https://github.com/zendesk/samson/blob/1469bbbd9281b713955e8155352076648b21c243/plugins/kubernetes/app/models/kubernetes/template_filler.rb#L246) and [here](https://github.com/zendesk/samson/blob/1469bbbd9281b713955e8155352076648b21c243/plugins/kubernetes/app/models/kubernetes/template_filler.rb#L249)?

plugins/kubernetes/app/models/kubernetes/cluster.rb
```diff
@@ -59,6 +59,13 @@ def schedulable_nodes
       []
     end
 
+    def server_version
+      version = Rails.cache.fetch(cache_key, expires_in: 2.minutes) do
```
grosser: can cache that for an hour or so :)

plugins/kubernetes/test/models/kubernetes/cluster_test.rb
```diff
@@ -128,6 +128,20 @@ def stub_response
     end
   end
 
+  describe '#server_version' do
+    before do
+      stub_request(:get, 'http://foobar.server/version').to_return(body: '{"gitVersion": "v1.6.0"}')
+    end
+
+    it 'caches the clusters server version and returns Gem::Version object' do
```
grosser: adding `and` to a test is always a good sign for "this should be 2 tests"

testing caching: usually more stable to call it twice and make sure the request was done only once

prefer asserting that the request was made to not get blindsided by caching

```
assert_request :get, foo, to_return: {body: 'xxx'} do
```

plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
```diff
@@ -20,6 +20,8 @@
   before do
     stage.update_column :kubernetes, true
     deploy.update_column :kubernetes, true
+
+    stub_request(:get, 'http://foobar.server/version').to_return(body: '{"gitVersion": "v1.5.0"}')
```
grosser: this should go to the global stubs actually ... see kubernetes/test/test_helper.rb

plugins/kubernetes/app/models/kubernetes/cluster.rb
```diff
@@ -59,6 +59,13 @@ def schedulable_nodes
       []
     end
 
+    def server_version
+      version = Rails.cache.fetch(cache_key, expires_in: 1.hour) do
+        JSON.parse(client.create_rest_client('version').get.body).fetch('gitVersion')[1..-1]
```
grosser: good call on not caching the `Gem::Version` ... custom ruby objects are just trouble :)

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -238,11 +233,21 @@ def set_secret_puller
 
     # Init containers are stored as a json annotation
     # see http://kubernetes.io/docs/user-guide/production-pods/#handling-initialization
-    def modify_init_container
-      key = Kubernetes::Api::Pod::INIT_CONTAINER_KEY
-      init_containers = JSON.parse(annotations[key] || '[]', symbolize_names: true)
-      yield init_containers
-      annotations[key] = JSON.pretty_generate(init_containers) if init_containers.any?
+    def set_init_containers
+      if init_containers.any?
```
grosser: FYI `return if init_containers.empty?` can be nicer

PR #2617, opened 2018-03-02
Make build create endpoint idempotent

ragurney:
@sathishavm

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -36,21 +36,27 @@ class BuildsController < ApplicationController
   def create
     new = false
     saved = false
+    external_build_has_digest = false
 
     Samson::Retry.retry_when_not_unique do
       if registering_external_build? && @build = find_external_build
-        return head :unprocessable_entity if @build.docker_repo_digest
+        external_build_has_digest = @build.docker_repo_digest.present?
         @build.attributes = edit_build_params(validate: false)
       else
         @build = scope.new(new_build_params.merge(creator: current_user))
       end
 
       new = @build.new_record?
-      saved = @build.save
+      changed = @build.changed?
+
+      return head :unprocessable_entity if external_build_has_digest && changed
+      saved = !changed || @build.save # nothing has changed or save result
     end
 
     start_docker_build if saved && !registering_external_build?
-    respond_to_save saved, (new ? :created : :ok), :new
+
+    status = new ? :created : :ok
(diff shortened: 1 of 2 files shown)


grosser:
can you describe what this changes / how it was broken before ?

grosser:
ahh ... so before pot that did not change anything also get unprocessable ...

app/controllers/builds_controller.rb
```diff
@@ -36,21 +36,35 @@ def new
...
-        return head :unprocessable_entity if @build.docker_repo_digest
+        external_build_has_digest = @build.docker_repo_digest.present?
         @build.attributes = edit_build_params(validate: false)
       else
         @build = scope.new(new_build_params.merge(creator: current_user))
       end
 
       new = @build.new_record?
-      saved = @build.save
+      updated = @build.changed?
```
grosser: ... more like `changed` since it's not updated at this point
or `will_update`

app/controllers/builds_controller.rb
```diff
@@ -36,21 +36,35 @@ def new
...
+      status =
+        if new
+          :created
+        elsif updated
+          :ok
+        else
+          :no_content
```
grosser: not sure if we even need that ... :ok would work too ?

app/controllers/builds_controller.rb
```diff
@@ -36,21 +36,35 @@ def new
...
+        end
+
+      return head :unprocessable_entity if external_build_has_digest && updated
+      saved = !updated || @build.save # nothing has changed or save result
     end
 
     start_docker_build if saved && !registering_external_build?
-    respond_to_save saved, (new ? :created : :ok), :new
+
+    respond_to_save saved, status, :new
```
grosser: this would send `:no_content` with a body which I'd think is invalid

app/controllers/builds_controller.rb
ragurney: I think you're right. I was under the impression that the client would need unique status codes depending on what happened (update vs no change) but it sounds like that's not the case anymore so just a :ok would work.

test/controllers/builds_controller_test.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 2
+SingleCov.covered! uncovered: 1
```
grosser: 🎉

PR #2622, opened 2018-03-05
lower uncovered now that single_cov is fixed

grosser:
@ragurney

### DIFF
--- a/test/models/job_execution_test.rb
+++ b/test/models/job_execution_test.rb
@@ -2,7 +2,7 @@
 require_relative '../test_helper'
 require 'ar_multi_threaded_transactional_tests'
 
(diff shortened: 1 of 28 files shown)


test/models/job_execution_test.rb
```diff
@@ -2,7 +2,7 @@
 require_relative '../test_helper'
 require 'ar_multi_threaded_transactional_tests'
 
-SingleCov.covered! uncovered: 11 # randomly says it only has 10 ... keep at 11
```
ragurney: 🤔

grosser: I'm hoping that bug is gone now ... might have to re-introduce that :D

PR #2627, opened 2018-03-06
PAAS-1360 explains why updating matchLabels is a bad idea

grosser:
@jonmoter

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -118,13 +118,17 @@ module Kubernetes
       end
 
       def ensure_not_updating_match_labels
+        # blue-green deply is allowed to do this, see template_filler.rb + deploy_executor.rb
+        return if @template.dig(:spec, :selector, :matchLabels, :blue_green)
+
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -111,10 +111,23 @@ def create
...
       def update
+        ensure_not_updating_match_labels
         request(:update, template_for_update)
         expire_cache
       end
 
+      def ensure_not_updating_match_labels
+        static = [:spec, :selector, :matchLabels]
+        old = @resource.dig(*static)
+        new = @template.dig(*static)
```
jonmoter: Generally avoid using `new` as a local variable.

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -111,10 +111,23 @@ def create
...
+        if old != new
+          raise(
+            Samson::Hooks::UserError,
+            "Updating #{static.join(".")} from #{old.inspect} to #{new.inspect} can only be done with a delete"
```
jonmoter: "can only be done by deleting and redeploying"

PR #2629, opened 2018-03-07
consider deploys stable after a certain time has passed to make big d…

grosser:
…eploys not take longer

@ragurney @jonmoter


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -107,7 +105,7 @@ def wait_for_resources_to_complete(release_docs)
             return success
           else
             @output.puts "READY, starting stability test"
-            @testing_for_stability = 0
+            @testing_for_stability = Time.now
```
jonmoter: How about `@stability_test_end = Time.now + CHECK_STABLE`, and then in the above comment, you could say:

```puts "Stable... testing for #{(@stability_test_end - Time.now).to_i} more seconds"```

PR #2632, opened 2018-03-08
run dev in pura puma so we can reproduce restart issues better

grosser:
without this sending signals to puma process does not reach RestartSignalHandler but goes straight to puma itself

checked that dev reload still works ... we'll see if we run into more issues ...

@jonmoter @ragurney


grosser:
/fyi @zendesk/devex using this in cerebro too ... might be good pattern for dev/prod parity ...

PR #2634, opened 2018-03-08
Change favicon color depending on the deploy status

ragurney:


app/views/deploys/show.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for(:favicon) { favicon_link_tag(favicon, type: 'image/png', id: 'favicon') } %>
```
grosser: id and type are copy-paste ... maybe only set `@favicon_path`

config/application.rb
```diff
@@ -175,6 +175,9 @@ class Application < Rails::Application
   end
 
   RELEASE_NUMBER = '\d+(:?\.\d+)*'
+  SUCCEEDED_FAVICON = '32x32_green.png'
```
grosser: hash might be nice and then `.fetch` from it

PR #2637, opened 2018-03-09
stop puma from sending INT on SIGHUP which we use for logrotate

grosser:
https://github.com/rails/rails/issues/32211

/cc @ragurney @jonmoter

### DIFF
--- a/lib/samson/logging.rb
+++ b/lib/samson/logging.rb
@@ -26,4 +26,26 @@ elsif Samson::EnvCheck.set?("RAILS_LOG_TO_SYSLOG")
   require 'syslog/logger'
   config.logger = Syslog::Logger.new('samson')
   config.lograge.formatter = Lograge::Formatters::Logstash.new
+elsif ENV["SERVER_MODE"] # regular file logger that needs rotating
+  # Reopen logfile when we receive a SIGHUP for log-rotation
+  # puma normally kills itself with INT if it receives a SIGHUP
+  # see https://github.com/puma/puma/pull/1527 and https://github.com/puma/puma/blob/master/docs/signals.md
+  # We need to use self-pipe since we cannot reopen logs in an interrupt.
+  # This will break stdout_redirect feature of puma since we cannot support both
+  read, write = IO.pipe
+  Signal.trap(:SIGHUP) { write.puts }
+  Thread.new do
+    loop do
+      read.gets # blocking wait for trap
+      Rails.logger.info "Received SIGHUP ... reopening logs"
+      # Rails.logger does not know what file it opened, so we help it https://github.com/rails/rails/issues/32211
+      dev = Rails.logger.instance_variable_get(:@logdev).dev
+      Rails.logger.reopen(dev.path)
+    end
+  end
+end
+
+# workaround for our server ... always gets SIGHUP even though it does not do file logging
+if ENV["IGNORE_SIGHUP"]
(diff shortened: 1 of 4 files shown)


lib/samson/logging.rb
```diff
@@ -27,3 +27,23 @@
   config.logger = Syslog::Logger.new('samson')
   config.lograge.formatter = Lograge::Formatters::Logstash.new
 end
+
+# puma kills itself with INT when it receives a SIGHUP see https://github.com/puma/puma/blob/master/docs/signals.md
+# If we have a file based logger then we need to reopen that file to log to the new logfile after log rotation.
+# The rails logger does not know what file it opened, so we have to help it https://github.com/rails/rails/issues/32211
+# If we do not have a file-based logger we still do not want to kill the process, but log that we got the signal.
+# We need to use self-pipe since we cannot reopen logs in an interrupt.
+# This will break stdout_redirect feature of puma, which I hope nobody uses.
```
jonmoter: Nice comment!  <3

lib/samson/logging.rb
```diff
@@ -27,3 +27,23 @@
...
+if ENV["SERVER_MODE"]
+  read, write = IO.pipe
+  Signal.trap(:SIGHUP) { write.puts }
```
jonmoter: Doesn't this swallow SIGHUP entirely?  Don't we need to propagate the signal somehow?

lib/samson/logging.rb
```diff
@@ -27,3 +27,23 @@
...
+  Thread.new do
+    loop do
+      read.gets
+      Rails.logger.info "Received SIGHUP ... reopening logs"
+      dev = Rails.logger.instance_variable_get(:@logdev)&.dev
+      path = dev.path if dev&.is_a?(File)
+      Rails.logger.reopen(path)
```
jonmoter: We want to call `Rails.logger.reopen(nil)` if `dev` isn't a file?

lib/samson/logging.rb
grosser: swallowing is exactly what we want since puma translates HUP -> INT

lib/samson/logging.rb
grosser: Calling with nil is a noop for all the other loggers, so not calling it is fine too ... maybe more obvious what is going on by not calling it

grosser:
reworked to make our hack more obvious since we don't need all this file reopen logic anyway ...

looks good ?

lib/samson/logging.rb
```diff
@@ -26,4 +26,26 @@
...
+  # This will break stdout_redirect feature of puma since we cannot support both
+  read, write = IO.pipe
+  Signal.trap(:SIGHUP) { write.puts }
+  Thread.new do
+    loop do
+      read.gets
+      Rails.logger.info "Received SIGHUP ... reopening logs"
+      # Rails.logger does not know what file it opened, so we help it https://github.com/rails/rails/issues/32211
+      dev = Rails.logger.instance_variable_get(:@logdev)&.dev
+      Rails.logger.reopen(dev.path)
```
jonmoter: Isn't it possible for `dev` to be `nil` here, so you'd be calling `nil.path`?

grosser: no since this must be a file logger at that point ... I'll remove the &. too that was ;eftover from previous approach

PR #2640, opened 2018-03-09
cloning a template should not make the other a template

grosser:
- leads to confusion since there are multiple templates
 - leads to bugs if someone clones and then modifies a stage
 - rewriting tests to make them helpful when things break

@dragonfax @ragurney


dragonfax:
Probably don't need this line anymore then, eh?
https://github.com/zendesk/samson/blob/c1f9af186fcb3a7431b1cb401b04adbce31a9c44/app/controllers/mass_rollouts_controller.rb#L139

grosser:
yes, removed that too :)

PR #2642, opened 2018-03-09
Create a Rollbar project on Samson project creation

ragurney:


plugins/rollbar/lib/samson_rollbar/samson_plugin.rb
```diff
@@ -8,6 +8,20 @@ class Engine < Rails::Engine
   end
 end
 
+Samson::Hooks.callback :project_created do |name|
```
grosser: should be optional on ROLLBAR_ACCOUNT_TOKEN being set

ragurney:
This feature will be moved to Cerebro in the form of a general tool to be used for current and new projects.

PR #2645, opened 2018-03-10
prefer previous builds when flag is used

grosser:
currently what often happens is that I deploy with reuse, but a new active build for the current sha is already building ... but I still want the old build so I can deploy right now and not wait for the new build to finish

@ragurney

### DIFF
--- a/lib/samson/build_finder.rb
+++ b/lib/samson/build_finder.rb
@@ -45,9 +45,10 @@ module Samson
       all = []
 
       wait_for_build_creation do |last_try|
+        possible = possible_builds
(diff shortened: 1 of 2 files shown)


lib/samson/build_finder.rb
```diff
@@ -45,9 +45,10 @@ def find_or_create_builds
       all = []
 
       wait_for_build_creation do |last_try|
+        possible = possible_builds
```
grosser: some general efficiency, no need to fetch these multiple times when we look for multiple builds

PR #2651, opened 2018-03-14
Tag Production Images as 'Production'

ragurney:
Tag images considered to be production as 'production', otherwise us the stage permalink.

### DIFF
--- a/plugins/gcloud/README.md
+++ b/plugins/gcloud/README.md
@@ -2,7 +2,8 @@
 
 ## Image tagging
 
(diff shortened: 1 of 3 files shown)


grosser:
update readme for gcloud plugin to point this out

grosser:
... maybe it should tag production AND permalink ... but not sure ... this is fine too ... we can wait for someone to complain

ragurney:
I was thinking about that... it would be nice to keep the stage permalink...

PR #2653, opened 2018-03-16
bump maxitest to cache rogue threads

grosser:
@ragurney


test/controllers/streams_controller_test.rb
```diff
@@ -10,7 +10,7 @@
   let(:stage) { stages(:test_staging) }
   let(:job) { jobs(:running_test) }
 
-  after { kill_extra_threads } # SSE heartbeat never finishes
+  after { maxitest_kill_extra_threads } # SSE heartbeat never finishes
```
ragurney: Could this be moved to `test_helper.rb`? Since we aren't testing for dangling threads anymore, why not just kill all extra threads each time?

grosser:
killing them each time means hiding potential bugs where code opens threads
and never closes them, also can leave code in a weird state, so I'd prefer
to avoid that

Issue #1212, opened 2016-08-17
[Docker + MySQL] Unable to save environemnts


machei:
I tried to save enviroments into specific project, it takes long time then i got error 404. I don't know.

Docker + Mysql External

log:

```
Started PATCH "/projects/example" for 222.222.222.21 at 2016-08-17 02:30:43 +0000
Processing by ProjectsController#update as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"XXX", "project"=>{"name"=>"Example", "description"=>"", "owner"=>"", "repository_url"=>"git@git.example.pl:example/example.git", "permalink"=>"example", "release_branch"=>"master", "auto_release_docker_image"=>"0", "environment_variables_attributes"=>{"0"=>{"name"=>"aa", "value"=>"aaa"}}, "environment_variable_group_ids"=>[""]}, "id"=>"xxx"}
```

grosser:
- try reproducing from the console, maybe that shows more details
 - missing 404 / end of logs ... enable request-id tagged logging to see where the request ends / all it's messages
 - does it work when saving non-environment settings ?
 - can you reproduce in development environment ?

machei:
Now all works.
What I have done:
- copied ssh key into instance based on docker ex. `COPY ssh/id_rsa /root/.ssh/`
- replaced Repository URL into valid with prefix : `ssh://`

PR #2656, opened 2018-03-19
Make 'favorited' star color consistent

### LINKED ISSUE #2655: Favorite stars are inconsistent
### Description
I believe this PR introduced some inconsistencies with favorited stars which caused me a small amount of confusion when favoriting a project recently.

> make starred and non-starred easy to see
> https://github.com/zendesk/samson/pull/1896

In the dropdown, a black star means favorited. Here is a snippet of that
image

Then here is the heading for the project page, where a black star means not favorited
image

and a gold star for favorited
image

### What should be changed?
The coloring here should be consistent. Either we use a gold star in the drop down as well, or introduce a new _non-favorited_ star for the project page - I saw it was originally a non-filled in star but it could be anything.

E.g.
image

ragurney:
fixes #2655

### DIFF
--- a/app/assets/javascripts/projects.js
+++ b/app/assets/javascripts/projects.js
@@ -1,7 +1,7 @@
 $(function() {
   // switch icons when user stars or unstars a project.
   // keep in sync with app/helpers/projects_helper.rb
-  $('.star a').bind('ajax:success', function() {
+  $('.star.project-star').bind('ajax:success', function() {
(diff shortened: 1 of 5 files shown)


grosser:
if you make the description of the PR `fixes #1212` then it will link it and auto-close on merge

ragurney:
screen shot 2018-03-18 at 8 54 43 pm

PR #2660, opened 2018-03-21
Rollbar Views for Project and Deploy

ragurney:
@zendesk/devex 
screen shot 2018-03-21 at 12 28 25 pm
screen shot 2018-03-21 at 12 28 05 pm
screen shot 2018-03-21 at 12 42 23 pm
screen shot 2018-03-21 at 12 41 40 pm
screen shot 2018-03-21 at 12 40 56 pm


plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,29 @@
+<%
+  occurrence_term = 'occurrences_count'
```
grosser: makes more sense inside the if

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  if deploy.succeeded?
+%>
+  <% if deploy.project.rollbar_dashboards_settings.empty? %>
+    <%= render partial: 'rollbar_dashboards/no_dashboards', locals: { project: deploy.project } %>
```
grosser: only projects that opt-in should get any UI change

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <%
+    else
+      @data = deploy.project.rollbar_dashboards_settings.map do |dashboard|
```
grosser: maybe class name `RollbarDashboard` makes more sense ?

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        params = {
+          date_from: deploy.created_at.in_time_zone(dashboard.time_zone).strftime('%m-%d-%Y %H:%M:%S'),
+          date_to: deploy.next_successful_deploy&.created_at&.in_time_zone(dashboard.time_zone)&.strftime('%m-%d-%Y %H:%M:%S') || '',
+        }
+
+        items = RollbarDashboards::Client.new(dashboard).items(params)
+        [dashboard, items&.sort_by { |i| i[occurrence_term] }&.reverse&.first(SamsonRollbarDashboards::Engine::NUM_ITEMS)]
```
grosser: prefer .fetch

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  occurrence_term = 'occurrences_count'
+  if deploy.succeeded?
+%>
+  <% if deploy.project.rollbar_dashboards_settings.empty? %>
+    <%= render partial: 'rollbar_dashboards/no_dashboards', locals: { project: deploy.project } %>
+  <%
+    else
+      @data = deploy.project.rollbar_dashboards_settings.map do |dashboard|
+        params = {
+          date_from: deploy.created_at.in_time_zone(dashboard.time_zone).strftime('%m-%d-%Y %H:%M:%S'),
```
grosser: would prefer if only users have a time zone and dashboards have not

grosser: ... we got time display helpers for that flying around somewhere

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
grosser: can sort by `-` to avoid the reverse ... or use `.last`

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        items = RollbarDashboards::Client.new(dashboard).items(params)
+        [dashboard, items&.sort_by { |i| i[occurrence_term] }&.reverse&.first(SamsonRollbarDashboards::Engine::NUM_ITEMS)]
+      end
+  %>
+    <%=
+      render partial: 'rollbar_dashboards/dashboard',
+        collection: @data,
+        as: :dashboard_data,
+        locals: {
+          title: "Top #{SamsonRollbarDashboards::Engine::NUM_ITEMS} Items That Occurred This Deploy",
```
grosser: items -> errors ?

"Top 5 Errors During This Deploy"

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_project.html.erb
```diff
@@ -0,0 +1,21 @@
+<% if project.rollbar_dashboards_settings.empty? %>
+  <%= render partial: 'rollbar_dashboards/no_dashboards', locals: { project: project } %>
```
grosser: same here, no UI change if users did not opt in

plugins/rollbar_dashboards/app/views/samson_rollbar_dashboards/_fields.html.erb
```diff
@@ -0,0 +1,37 @@
+<fieldset>
+  <legend>
+    Rollbar Dashboards
+  </legend>
+  <% project = form.object %>
```
grosser: `@project` is also available

plugins/rollbar_dashboards/config/routes.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
```
grosser: where is that used ?

plugins/rollbar_dashboards/app/models/rollbar_dashboards/setting.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module RollbarDashboards
+  class Setting < ActiveRecord::Base
+    self.table_name = 'rollbar_dashboards_settings'
+
+    belongs_to :project
+
+    validates :base_url, :read_token, :time_zone, presence: true
+
+    def resolved_read_token
```
grosser: if this is used on every project show/deploy-show then we might want to cache that ... also not blow up and make the project undeployable in cas vault has a hickup

plugins/rollbar_dashboards/lib/rollbar_dashboards/client.rb
```diff
@@ -0,0 +1,40 @@
...
+        hours: hours,
+        environments: environments.join(',')
+      }.to_query
+
+      response = Faraday.get("#{@base}/reports/top_active_items?#{args}")
+      if response.success?
+        data = ::JSON.parse(response.body)
+        data['result'].map { |item_hash| item_hash['item'] }
+      end
+    rescue ::JSON::ParserError
```
grosser: need to rescue more here ... send to ErrorNotifier and return nil ... we can't blow up peoples project/deploy pages
... and needs a timeout ideally at the http level since `Timeout.timeout` is dangerous

plugins/rollbar_dashboards/lib/rollbar_dashboards/client.rb
```diff
@@ -0,0 +1,40 @@
...
+      nil
+    end
+
+    def items(additional_params = {})
+      params = {
```
grosser: since additional_params is not reused I would normally do something like:

```
params = {foo: bar}.merge(params)
````

Comment (grosser) [reply plugins/rollbar_dashboards/lib/rollbar_dashboards/client.rb]: also don't name different data-structures the same .. params -> hash ... query -> string

File: plugins/rollbar_dashboards/lib/rollbar_dashboards/client.rb [hunk 13]
```diff
@@ -0,0 +1,40 @@
...
+        access_token: @token
+      }.merge(additional_params).to_query
+
+      response = Faraday.get("#{@base}/items/?#{params}")
+      if response.success?
+        data = ::JSON.parse(response.body)
```
grosser: prefer parsing json with `symbolize_names: true` so everything is symbols ✨

plugins/rollbar_dashboards/lib/samson_rollbar_dashboards/samson_plugin.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module SamsonRollbarDashboards
+  class Engine < Rails::Engine
+    NUM_ITEMS = 4
+
+    initializer "refinery.assets.precompile" do |app|
```
grosser: refinery ?

dragonfax:
All-in-all. Very cool.

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
ragurney: The time zone of the Rollbar project is independent of the time zone of the user, so we'd want to store that separately in order to get the correct items for the deploy dashboard.

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
ragurney: Rollbar uses the term 'Item' in it's dashboard. Do we want to copy their terminology or would 'Error' be more intuitive?

plugins/rollbar_dashboards/config/routes.rb
ragurney: Remanent of a previous version of this feature 😅

plugins/rollbar_dashboards/app/models/rollbar_dashboards/setting.rb
ragurney: If someone were to update the token and the token was cached then the feature would be broken until the cached value expired? I could see that leading to some confusion. I could cache the result to the `RollbarDashboards::Client` call on the deploy page as well? Then the entire dashboard is essentially cached for 2 minutes, including the token.

grosser:
if the dates returned from rollbar are in utc then we don't need to worry
about that ?

grosser:
ah yeah then that should be enough caching ...

ragurney:
The dates in Rollbar aren't in UTC. From my understanding the query params are used as if they are in whatever time zone the project is set to 😕

grosser:
then play with it a bit to make sure ... or check api docs ... ideally we'd
automatically fill the timezone value in ...
but then still: convert to utc and display with out time helpers

ragurney:
Addressed comments, but I'm going to have to re-examine the logic of getting the items by date range. I'm getting some odd results that weren't expected. Might have to resort to RQL to get it to work.

ragurney:
After some more digging, confirmed I'll have to use RQL (create a job, then get the job result in two separate requests). The method I was using previously isn't officially supported and gives erroneous results.

dragonfax:
weird.

lib/samson/model_utility_methods.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Samson
+  module ModelUtilityMethods
+    def self.resolved_attribute(model_instance, attribute, project)
```
grosser: pass in attribute_value instead of instance+ attribute

grosser: can live on the KeyResolver

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,23 @@
+<% unless deploy.project.rollbar_dashboards_settings.empty? || !deploy.succeeded?
```
grosser: prefer to use if instead of unless with multiple conditions

`if settings.any? && deploy.succeeded?`

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    occurrence_term = :occurrences_count
+    @data = deploy.project.rollbar_dashboards_settings.map do |dashboard|
+      params = {
```
grosser: maybe the dashboard could know how to get items ?

`dashboard.items(deploy)`

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_no_dashboards.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="alert alert-info alert-dismissible" role="alert">
```
grosser: no longer used

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_project.html.erb
```diff
@@ -0,0 +1,19 @@
+<%
+  unless project.rollbar_dashboards_settings.empty?
```
grosser: avoid double negative -> `if any`

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_project.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    @dashboard_data = project.rollbar_dashboards_settings.map do |dashboard|
+      items = Rails.cache.fetch("project_#{dashboard.cache_key}", expires_in: 10.minutes) do
```
grosser: FYI passing models to `.fetch` uses cache_key on them, so can do `.fetch(['project', deployment], expires_in : 10.minutes)` ...

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_project.html.erb
```diff
@@ -0,0 +1,19 @@
...
+        RollbarDashboards::Client.new(dashboard).top_errors&.first(SamsonRollbarDashboards::Engine::NUM_ITEMS)
+      end
+      [dashboard, items]
+    end
+%>
```
grosser: might be simpler to cache the html instead of caching the data and then passing it in
then it could fetch the data in the partial

grosser:
A deploy should only show items from the environment it deploys to.
So dashboard should be on stage ... and project should display top-errors from the first stage sorted by `production?` right ?

ragurney:
I think it makes sense to have the dashboard for a deploy live on the deploy show page? That way you could go back and look at a past deploy and its errors, which is how the desired functionality/design described to me. I'm not sure what you mean by: 
> and project should display top-errors from the first stage sorted by production

Right now, the project dashboard is displaying the top errors in the production environment for it's rollbar project which is what we'd want, right? Just a summary of the current state of production

ragurney:
Deploys should display items only for it's environment. I think Jason and I talked about that being a later PR, but I could incorporate it into this RQL stuff.

grosser:
the deploy does not know it's environment (master/staging, production works ... if users use strange names then that fails ...) so we could configure it on the stage, that's how the rollbar_hook plugin works ... having to copy-paste the config to multiple stages would be cumbersome though :/

telling master deployers that they caused production errors is misleading ... so either don't show unless it's production or say "errors in all environments" ... maybe we can get environment from the deploy groups `Environment` it is deploying too ...

grosser:
I guess since it's opt-in it does not really matter and we can play around with it ...

dragonfax:
The project page can show just production, (or everything). Either way is
fine with me. But I think production is more relavant for the project as a
whole.

The stage page, should show errors from the environment that it deploys to.
You can find this by following Stage->deploy_groups->first->environment

dragonfax:
deploy groups know which environment they are in.

ragurney:
Updated deploy dashboard logic, now with RQL and async job result loading ✨

plugins/rollbar_dashboards/app/controllers/rollbar_dashboards/dashboards_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+      locals: { title: "Top #{RollbarDashboards::Setting::NUM_ITEMS} Items That Occurred This Deploy" }
+    )
+  end
+
+  private
+
+  def wait_for_rql_job_results(job_id, client)
+    MAX_RQL_JOB_WAIT_TIME.times do
+      result = client.rql_job_result(job_id)
+      return result if result.nil? || result != RollbarDashboards::Client::RQL_JOB_WAITING_TERM
```
grosser: rql_job_result should return nil when the result is RQL_JOB_WAITING_TERM ? <-> keep logic local

plugins/rollbar_dashboards/app/controllers/rollbar_dashboards/dashboards_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+      sleep(1)
```
grosser: it's kinda strange to sleep and then stop the loop :D 
`times do |i| ... break if i >= (MAX_RQL_JOB_WAIT_TIME - 1)`

plugins/rollbar_dashboards/app/controllers/rollbar_dashboards/dashboards_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+             item.counter as item_id,
+             item.title as title,
+             Count(*) as occurrences,
+             item.environment as environment
+      FROM   item_occurrence
+      WHERE  environment = "#{environment}"
+             AND #{timestamp_query}
+      GROUP  BY 1,
+                item.counter
+      ORDER  BY 4 DESC
```
grosser: can we put the limit here ?

plugins/rollbar_dashboards/app/models/rollbar_dashboards/setting.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module RollbarDashboards
+  class Setting < ActiveRecord::Base
+    NUM_ITEMS = 4
```
grosser: the model should not have display logic ... maybe better in the controller ?

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_dashboard.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="panel panel-default">
+  <div class="panel-heading">
+    <%= image_tag 'rollbar_dashboards/icon', height: '30px' %>
+    <%= "#{title} (#{dashboard_data.first.base_url})" %>
+  </div>
+  <% items = dashboard_data.second %>
+  <% if items %>
```
grosser: FYI could turn this into a single `if items = ` line

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_dashboard.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <tr>
+        <th>Count</th>
+        <th>Title</th>
+        <th>Environment</th>
+      </tr>
+      <%= render partial: 'rollbar_dashboards/item', collection: items %>
+    </table>
+  <% else %>
+    <div class="panel-body">
+      <p>There are no items to display at this time...</p>
```
grosser: `to display at this time` -> `found` / `occurred`

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_deploy.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if deploy.project.rollbar_dashboards_settings.any? && deploy.succeeded? %>
```
grosser: a comment here like `<%# will be replaced by foobar.js %>` could be nice

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_project.html.erb
```diff
@@ -0,0 +1,18 @@
+<%
+  if project.rollbar_dashboards_settings.any?
+    @dashboard_data = project.rollbar_dashboards_settings.map do |dashboard|
+      items = Rails.cache.fetch(['project', dashboard], expires_in: 10.minutes) do
+        RollbarDashboards::Client.new(dashboard).top_errors&.first(RollbarDashboards::Setting::NUM_ITEMS)
```
grosser: this could potentially cache nil ... which means the cache won't work ...
might be best to move this somewhere in the helper/model/client code to properly test it too

plugins/rollbar_dashboards/app/views/samson_rollbar_dashboards/_fields.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  </legend>
+  <% @project.rollbar_dashboards_settings.build %>
+
+  <b>Base URL:</b> https://api.rollbar.com/api/1 unless you're using a self-hosted instance of Rollbar.<br>
+  <b>Access token:</b> Your Rollbar project's <b>read</b> token, supports secret lookup (e.g. secret://rollbar_read_token)<br>
+  <br/><br/>
+
+  <div class="col-lg-offset-2">
+    <%= form.fields_for :rollbar_dashboards_settings do |setting| %>
+
```
grosser: extra newline

plugins/rollbar_dashboards/app/views/samson_rollbar_dashboards/_fields.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <div class="form-group">
+        <div class="col-lg-4">
+          <%= setting.url_field :base_url, class: "form-control", placeholder: "Base URL"  %>
+        </div>
+        <div class="col-lg-3">
+          <%= setting.text_field :read_token, class: "form-control", placeholder: "Project Read Token" %>
+        </div>
+        <div class="col-lg-10" style="margin-left: .5em">
```
grosser: size 10 for a delete checkbox seems strange

plugins/rollbar_dashboards/app/views/rollbar_dashboards/_project.html.erb
grosser: possibly also good to make this a lazy-load so the page stays fast and everything has the same workflow

plugins/rollbar_dashboards/app/controllers/rollbar_dashboards/dashboards_controller.rb
ragurney: So move the waiting logic into `rql_job_result`?

grosser:
yeah

app/assets/javascripts/responsive_load.js.erb
```diff
@@ -1,14 +1,26 @@
 // show loading animation and errors inside the container that is being replaced
+$(document).ready(function () {
+  $('.lazy-load').each(function () {
```
grosser: `$('.lazy-load'). responsiveLoad()` is not the same ?

lib/samson/form_builder.rb
```diff
@@ -54,6 +54,24 @@ def actions(delete: false, label: 'Save', &block)
       end
     end
 
+    # Creates multi-row input field
+    def fields_for_many(association, description, add_rows_allowed: false)
```
grosser: good stuff! ❤️

plugins/flowdock/app/views/samson_flowdock/_fields.html.erb
```diff
@@ -1,21 +1,22 @@
 <fieldset>
   <legend>Flowdock</legend>
-  <p class="col-lg-offset-2">Name and API token of the Flowdock flow which should be notified of deploys.</p>
+  <p class=></p>
 
-  <% stage = form.object %>
-  <% stage.flowdock_flows.build if stage.flowdock_flows.last.try(:name).blank? %>
+  <% 
+    stage = form.object
+    stage.flowdock_flows.build if stage.flowdock_flows.last.try(:name).blank?
```
grosser: `unless stage.flowdock_flows.last&.new_record?` maybe ?

plugins/rollbar_dashboards/app/controllers/rollbar_dashboards/dashboards_controller.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+class RollbarDashboards::DashboardsController < ApplicationController
+  NUM_ITEMS = 4
+
+  def project_dashboard
+    project = Project.find_by_permalink(params.require(:project_id))
```
grosser: `find_by_permalink!` since we don't want nils

plugins/rollbar_dashboards/app/controllers/rollbar_dashboards/dashboards_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+      )
+    end
+
+    render plain: dashboard_html
+  end
+
+  def deploy_dashboard
+    deploy = Deploy.find(params.require(:deploy_id))
+
+    dashboard_html = Rails.cache.fetch(deploy, expires_in: 10.minutes) do
```
grosser: this needs to be a more unique cache-key like `["deploy_dashboard", deploy]`

plugins/rollbar_dashboards/app/controllers/rollbar_dashboards/dashboards_controller.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+class RollbarDashboards::DashboardsController < ApplicationController
+  NUM_ITEMS = 4
+
+  def project_dashboard
+    project = Project.find_by_permalink(params.require(:project_id))
+
+    dashboard_html = Rails.cache.fetch(project, expires_in: 10.minutes) do
```
grosser: needs a more unique cache key

plugins/rollbar_dashboards/lib/rollbar_dashboards/client.rb
```diff
@@ -0,0 +1,84 @@
...
+        environments: environments.join(',')
+      }.to_query
+
+      response = Faraday.get("#{@base}/reports/top_active_items?#{args}") do |req|
+        req.options.timeout = req.options.open_timeout = 1
+      end
+
+      if response.success?
+        data = ::JSON.parse(response.body, symbolize_names: true)
+        data[:result].map { |item_hash| item_hash[:item] }
```
grosser: FYI usually nice to avoid nils via `data.fetch(:result).map { |i| i.fetch(:item) }`

plugins/rollbar_dashboards/lib/rollbar_dashboards/client.rb
```diff
@@ -0,0 +1,84 @@
...
+    # returns array hashes of top items
+    def top_errors(hours: 24, environments: ['production'])
+      args = {
+        access_token: @token,
+        hours: hours,
+        environments: environments.join(',')
+      }.to_query
+
+      response = Faraday.get("#{@base}/reports/top_active_items?#{args}") do |req|
+        req.options.timeout = req.options.open_timeout = 1
```
grosser: might want to make that a `ROLLBAR_API_TIMEOUT` constant

PR #2663, opened 2018-03-27
Fix rollbar dashboard

ragurney:

### DIFF
--- a/lib/samson/form_builder.rb
+++ b/lib/samson/form_builder.rb
@@ -64,7 +64,7 @@ module Samson
       content << fields_for(association) do |a|
         content_tag(:div, class: 'form-group') do
           yield(a)
-          @template.delete_checkbox a
+          @template.concat @template.delete_checkbox a
         end
       end
       content << @template.link_to("Add row", "#", class: "duplicate_previous_row") if add_rows_allowed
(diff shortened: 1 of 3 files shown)


test/lib/samson/form_builder_test.rb
```diff
@@ -145,28 +145,38 @@ def fake_erb_rendering
   end
 
   describe '#fields_for_many' do
-    before { template.expects(:delete_checkbox).returns('delete_checkbox') }
     def render(args)
+      project.rollbar_dashboards_settings.build
       builder.fields_for_many(*args) do |p|
-        p.text_field :name
+        p.text_field :base_url, placeholder: 'thing!'
```
dragonfax: thing?

lib/samson/form_builder.rb
```diff
@@ -64,7 +64,7 @@ def fields_for_many(association, description, add_rows_allowed: false)
       content << fields_for(association) do |a|
         content_tag(:div, class: 'form-group') do
           yield(a)
-          @template.delete_checkbox a
+          @template.concat @template.delete_checkbox a
```
dragonfax: Does this make it a link?

test/lib/samson/form_builder_test.rb
dragonfax: Sorry, forgot I was reading a test.

lib/samson/form_builder.rb
ragurney: It creates a checkbox input

PR #2664, opened 2018-03-29
Save env state on deploy

ragurney:

### DIFF
--- a/app/models/concerns/group_scope.rb
+++ b/app/models/concerns/group_scope.rb
@@ -19,6 +19,17 @@ module GroupScope
     "#{scope_type}-#{scope_id}"
   end
 
+  def matches_scope?(deploy_group)
+    return true unless scope_id # for all
+    return false unless deploy_group # unscoped -> no specific groups
+
+    case scope_type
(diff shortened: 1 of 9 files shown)


app/models/deploy.rb
```diff
@@ -266,4 +268,18 @@ def deploy_buddy
   def trim_reference
     self.reference = reference.strip if reference.present?
   end
+
+  def serialize_env_state
```
grosser: stage has a cheaper/compact env serialization :(

grosser: https://github.com/zendesk/samson/blob/master/plugins/env/app/decorators/project_decorator.rb#L22

... also it should not blow up when `env` plugin is not enabled :trollface:

grosser: having the hash would make deploying an old deploy easier ... unless there is a new deploy-group 🤷‍♂️

ragurney: Where?

grosser:
test

grosser:
- make matches_scope public if it's used from the outside instead of `send`

 - serialized_environment_variables can be private ?

 - variables_to_string -> serialize ?

Issue #2372, opened 2017-11-08
One-off jobs executed in Stages with Kubernetes/Macros to Stages


klaus993:
Hello,

At Restorando we use Samson a lot for executing and monitoring deployments and we are quite happy with it. We are currently migrating our infrastructure from Chef to Kubernetes and we need to make sure that Samson has us covered in terms of the use cases we have today.

### Current Use Case

Today we use Capistrano along with the Macros feature of Samson to execute one-off jobs in any environment we have (Staging, QA, Production). Since our apps are mostly written in Rails, these one-off jobs are usually specific data migrations that we wish to run (we should support running any type of command in fact).

We execute these data migrations with a generic Macro for each environment that takes command line arguments that we configure with the feature: configure and execute. 

We usually configure the `SCRIPT` and `SCRIPT_ARGV` environment variables which executes the following macro:

```bash
export SCRIPT=SCRIPT_NAME.rb
export SCRIPT_ARGV=≈arg1 arg2 arg3…
export USER="$DEPLOYER_NAME"
/root/create_ssh_environment.example.sh && . /root/.ssh/environment
bundle install --with deployment --without default test development production staging qa  sandbox console --path=$CACHE_DIR/bundle
export ENVIRONMENT=production
bin/cap $ENVIRONMENT run:data_migration
```
#### Actual Screenshot:

<img width="1153" alt="screen shot 2017-11-08 at 10 43 43" src="https://user-images.githubusercontent.com/18153834/32553935-ff3ec0a0-c476-11e7-82cb-bcb3e1094497.png">

### Problem

The problem we have today is two-fold. 
1. Considering that Macros is getting deprecated and replaced with Stages, we need to find a way of making "configure and execute a one-off job" work well with Stages.
2. For K8s projects, Samson reads the Objects and bundles all of them as K8s Roles, executing them together in every release.

### Stages instead of Macros

So far, we haven't seen a simple way of configuring environment variables for each deploy of a Stage. We have found UI inputs in the Project Level configuration panel, but this wouldn't work because it wouldn't scale with many developers executing one-off jobs every time. This applies to all Projects, whether they are running on K8s or not.

We found this PR (https://github.com/zendesk/samson/pull/1866) which describes how Macros are now deprecated by Stages. We also ran the data migrations until `v1914` (we are currently using `v1556`) with a copy of our DB to see how Macros end up converted in Stages. Unfortunately, we are unable to find a way of executing the use case above. What we would like help with is:
[...]

PR #1866, opened 2017-04-02
macros to stages

grosser:
/cc @zendesk/samson @craig-day @jcheatham 

### Problems
 - big code duplication around macros/stages
 - duplicate queries to look for macros and stages commands
 - macros lack many features stages have
 - no tracking of who executed a macro / what it did / what deploy groups it interacted with
 - macros and jobs take up 2 tabs in the project view

### Solution
Add a new `no_reference_selection` boolean that then allows a user to instantly `Deploy!` a stage when used with `no confirm` and `does not deploy code`.

Users are free to move them wherever they please ... so either at the top of the project page or the bottom ... or right after the stage they deal with.

Phase 1 (this PR):
 - [X] UI hints on how to combine flags
 - [X] show `Deploy!` button for `direct` stages
 - [x] bypasses reference selection when used without other flags
 - [ ] Update a few projects to use this and gather feedback

Phase 2:
 - [ ] migrate macros to stages with 'no code deployed' + 'no confirm' + 'no reference selection'
 - [ ] remove jobs view
 - [ ] remove macros

Phase 3:
 - [ ] inline modules we extracted to deal with stage/macro duplication

screen shot 2017-04-01 at 9 21 27 pm
screen shot 2017-04-01 at 9 23 19 pm

PR #2666, opened 2018-04-03
Create Deploy Environment Variables plugin

nogates:
## Description

In this PR, I am creating a plugin to add environment variables to deploys, same as projects or stages.

This functionality is required to provide a way to use samson to run one-off tasks that need to be parameterized. Something similar as Macros used to have before they were merged into Stages here https://github.com/zendesk/samson/pull/1866

Our use case, and the reason why I am adding this plugin, is that we want to give some of our users the ability to run specific tasks in our servers through samson, wihout having to create an specific UI in our API for this specific task. 

Also, it seems that I am not the only one missing this feature (as you can see here https://github.com/zendesk/samson/issues/2372), and even at some point, this was considered to be implemented, but never completed (https://github.com/zendesk/samson/pull/1803, although in that PR the environment variables are being added as part of the core code, while the environment variables models and functionality is actually being included as a plugin, so that's why I've decided to follow that approach and create a plugin for this.). 

## Screenshots

Deploy Form
screenshot_20180403_114249

Deploy Output
screenshot_20180403_114320

## Caveats

There is one main problem that I've found while working on this PR, and that's the [Redeploy button](https://github.com/zendesk/samson/blob/master/app/helpers/deploys_helper.rb#L63). As you can see there, that logic is expecting that all the `deploy permitted params` are strings, but here, I am introducing a `has_many` relation between the deploy and the environment variables, and using `environment_variables_attributes` to pass the variables in the form. 

This causes issues when rendering the redeploy button, which I've solved by overriding the default method of the `DeploysHelper`, so the nested attributes can be added to the link. 

Not sure if this is the best way of implementing this. Maybe, the deploy should have a model method that copies the parameters from a previous deploy? instead of using some sort of form to create a new complete deploy. If we do that, that would much easier to modify using model decorators, instead of the helper which cannot be modified that easily.

I would be happy to change this if you think there is a better way of doing it.

Thanks anyway for your work in this project! It's great and I was very surprised to see how easy was to add new functionalities through hooks. So yeah, :clap::clap: !

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


plugins/deploy_env_vars/app/helpers/deploys_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+      ),
+      html_options
+  end
+
+  private
+
+  # Applies different logic depending on the class of each of the deploy
+  # parameters, so it supports nested paramaters based on object relations
+  def redeploy_params
+    params = Samson::Hooks.fire(:deploy_permitted_params).flatten(1) || []
```
grosser: don't need the `|| []` since .flatten never returns nil

plugins/deploy_env_vars/app/helpers/deploys_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+    params.each_with_object(reference: @deploy.reference) do |param, collection|
+      case param
+      when String, Symbol
+        collection[param] = @deploy.public_send(param)
+        collection
```
grosser: don't need this line

plugins/deploy_env_vars/app/helpers/deploys_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+      when Hash
+        nested_redeploy_params(collection, param)
+      else
```
grosser: don't need the else

plugins/deploy_env_vars/app/helpers/deploys_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+        collection
+      end
+    end
+  end
+
+  def nested_redeploy_params(collection, params)
+    params.each_with_object(collection) do |(key, attributes), nested|
```
grosser: collection and nested are the same, no reason to use each_with_object

plugins/deploy_env_vars/app/helpers/deploys_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+      if key.to_s.ends_with?('attributes')
```
grosser: prefer `end_with` ... there should be a rubocop complaining about that ... maybe that got disabled :/

plugins/deploy_env_vars/app/helpers/deploys_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+  end
+
+  # currently, one has_many relations are supported
+  def deploy_relation_attributes(key, attributes)
+    relation_name = key.to_s.gsub(/_attributes$/, '')
+    @deploy.public_send(relation_name).map do |item|
+      (attributes - [:id, :_destroy]).each_with_object({}) do |attribute, hash|
+        hash[attribute] = item.public_send(attribute)
+        hash
```
grosser: don't need this line

plugins/deploy_env_vars/app/helpers/deploys_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+  def nested_redeploy_params(collection, params)
+    params.each_with_object(collection) do |(key, attributes), nested|
+      if key.to_s.ends_with?('attributes')
+        nested[key] = deploy_relation_attributes(key, attributes)
+      end
+      nested
+    end
+  end
+
+  # currently, one has_many relations are supported
```
grosser: maybe give an example of what this does ... I don't quiet get it :D

plugins/deploy_env_vars/app/views/samson_deploy_env_vars/_deploy_form.html.erb
```diff
@@ -0,0 +1,33 @@
+<fieldset>
+  <legend>
+    Specific Environment Variables for this deploy <%= additional_info SamsonEnv::HELP_TEXT %>
```
grosser: maybe Specific -> Extra

plugins/deploy_env_vars/app/views/samson_deploy_env_vars/_deploy_form.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  </legend>
+
+  <% current_vars = form.object.environment_variables %>
+  <%= form.fields_for :environment_variables, current_vars << EnvironmentVariable.new do |fields| %>
```
grosser: prefer `(current_vars << EnvironmentVariable.new)` to make it less like the `<<` is using the `do |fields|`
... maybe better to just do it in the line above ... and not do it if the last one is `.new_record?` already

plugins/deploy_env_vars/app/views/samson_deploy_env_vars/_deploy_form.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    <div class="form-group">
```
grosser: the whole section smells like copy-paste ... can we reuse a partial instead ?

plugins/deploy_env_vars/lib/samson_deploy_env_vars/samson_plugin.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+module SamsonDeployEnvVars
+  class Engine < Rails::Engine
+    config.after_initialize do
+      # Decorates default deploys helper. Maybe there is a better way to do this?
+      require_relative '../../app/helpers/deploys_helper'
```
grosser: the regular `inherited` hack won't work since they don't inherit anything
I'd recommend using a `:decorate_helper` hook to make that work ... so the original `deploys_helper.rb` gets a `.fire :decorate_helper, :deploys_helper` for example and here we'd add a callback to trigger the require

plugins/deploy_env_vars/lib/samson_deploy_env_vars/samson_plugin.rb
```diff
@@ -0,0 +1,30 @@
...
+end
+
+# Adds the deploy env vars view to the deploy form in order to add
+# specific environment vars per deploy
+Samson::Hooks.view :deploy_form, 'samson_deploy_env_vars/deploy_form'
+
+# Allows environment vars as valid parameters for the deploy model
+Samson::Hooks.callback :deploy_permitted_params do
+  AcceptsEnvironmentVariables::ASSIGNABLE_ATTRIBUTES
+end
```
grosser: ❤️

plugins/deploy_env_vars/lib/samson_deploy_env_vars/samson_plugin.rb
```diff
@@ -0,0 +1,30 @@
...
+
+# Injects specific environment variables for the deploy if any
+Samson::Hooks.callback :job_additional_vars do |job|
+  if job.deploy
+    job.deploy.environment_variables.each_with_object({}) do |var, collection|
+      collection[var.name] = var.value
+      collection
+    end
+  else
+    {}
```
grosser: should not need that ... make whoever calls job_additional_vars do a `.compact` ?

test/helpers/deploys_helper_test.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 1
+SingleCov.covered! uncovered: 11
```
grosser: ouch

plugins/deploy_env_vars/app/helpers/deploys_helper.rb
nogates: can do!

plugins/deploy_env_vars/app/helpers/deploys_helper.rb
nogates: I mean that this only (not one) support `has_many` relations. If someones writes a plugin that includes a `has_one` relation, this code won't fully work since the public_method we call on the object is expected to return an array. does this make sense?

plugins/deploy_env_vars/app/views/samson_deploy_env_vars/_deploy_form.html.erb
nogates: yeah! I actually removed the warning from the `flay` rake task :blush: I will move this to a partial

plugins/deploy_env_vars/lib/samson_deploy_env_vars/samson_plugin.rb
nogates: I see. Will investigate tomorrow how can I do that (I guess that in a similar way as controllers and models are already doing)

grosser:
yeah, but update the docs to say that :)

grosser:
will need it's own hook with fire/callback ... won't be as nice, but should
ideally be automagic for plugin authors :)

test/helpers/deploys_helper_test.rb
nogates: yeah! this is actually not true, but given that I am overriding the method with the new deploys helper, singlecov thinks these lines are not covered anymore, but they actually are... nevertheless, I hope I can revert this change if I take the decorator + hook path instead of overriding the method directly.

plugins/deploy_env_vars/lib/samson_deploy_env_vars/samson_plugin.rb
nogates: I saw that if I didn't do that I was breaking this test: https://github.com/zendesk/samson/blob/master/test/models/job_execution_test.rb#L165 That's why I implemented this

nogates:
@grosser I've implemented most of the suggested changes. However, I was thinking that maybe, instead of creating a decorator in the plugin, or even having to add new hooks, it might be better to add the logic of parsing different types of parameters of the redeploy button directly into the core deploys helper, and also move the tests there.

I think this way the implementation will be clearer, and other plugins can benefit of this enhancement without having to use the `deploy_env_vars` plugin.

What do you think about that? If you don't like the idea, I will find a way to make the decorator work

nogates:
@grosser done in https://github.com/zendesk/samson/pull/2666/commits/6ec0e310fa22d2319b89bbf8f2d05bea8320fe65

app/helpers/deploys_helper.rb
```diff
@@ -109,4 +106,42 @@ def deploy_favicon_path(deploy)
...
+      else
+        raise "Unsupported deploy param class: `#{param.class}` for `#{param}`."
+      end
+    end
+  end
+
+  def nested_redeploy_params(collection, params)
+    params.each_with_object(collection) do |(key, attributes), nested|
+      if key.to_s.end_with?('attributes')
+        nested[key] = deploy_relation_attributes(key, attributes)
```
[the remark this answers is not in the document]
nogates: not sure what you mean @grosser... `params` could have multiple entries, and all of them need to be parsed and included as part of the response of this method... so yeah, I think `each_with_object` fits here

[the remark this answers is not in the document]
grosser: `params.each ... collection[key] = `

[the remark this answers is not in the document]
grosser: maybe `add_nested_redeploy_params!` to signal that it does in-place edit

lib/samson/redeploy_params.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
```
grosser: beautiful 😍

plugins/deploy_env_vars/test/samson_deploy_env_vars/samson_plugin_test.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      context "if the deploy does not have any env variables" do
+        it "returns an empty hash" do
+          Samson::Hooks.fire(:job_additional_vars, job).must_equal([{}])
+        end
+      end
+
+      context "if the deploy has environment variables" do
+        before do
+          job.deploy.environment_variables << EnvironmentVariable.create(
```
grosser: fyi same as `job.deploy.environment_variables.create!(name: foo, value: bar)`

plugins/env/app/views/samson_env/_environment_variables_fields.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    <div class="col-lg-3">
+      <%= fields.text_field :name, class: "form-control", placeholder: "Name" %>
+    </div>
+
+    <div class="col-lg-5">
+      <%# using a text area so users can resize them with browser controls %>
+      <%= fields.text_area :value, class: "form-control", placeholder: "Value", rows: 1 %>
+    </div>
+
+    <% if DeployGroup.enabled? %>
```
grosser: the scope should not be shown on the deploy page since it's not being used, right ? ... so `&& !@deploy`

nogates: true that! I don't have the DeployGroup enabled, so I didn't see that... will add that condition

nogates: although, will use `@deploy.nil?`

nogates: Although, this is not gonna work because we are not passing the `scopes` variable here from the parent view, and we do need that `scope` var to generate the actual env variables list for the project.

So I've decided to use the presence of that variable `scope`, as the condition to show or hide the `scope_type_and_id` param

Also, I did enable / disabled the feature with `DEPLOY_GROUP_FEATURE=true` and ensure everything worked fine.

app/helpers/deploys_helper.rb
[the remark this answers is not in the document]
nogates: oh! yes, sorry! have been doing some functional programming lately and now I want to return the transformed data all the time! :D

grosser:
thx for the awesome addition :)

nogates:
Thanks to you for spending time reviewing this!! :-)

grosser:
Np, thx for fixing it all up :D

nogates:
btw @grosser, could you please at some point restart the master build on travis? because it didn't pass when the PR was merged, and the docker image hasn't been built for the latest version. 
but yeah, if it's not a problem. I can also wait until a new release is generated :)

nogates:
awesome! thanks for the heads up @grosser

rata:
Hi @grosser @nogates 

Is there any interest to make the env. variables available to kubernetes jobs too?

I deployed and played with this and, if I'm not missing anything, it's not possible to set env variables in a kubernetes pod. I'd like to do this, or something else, to run 1-off jobs in kubernetes.

I really don't know ruby, but I can try to help to code it, if you are okay with the feature.

Are you okay with deployment env vars for kubernetes pods to have this 1-off jobs functionality?

Another option I'm exploring (if this is not welcome or hard), just use a server with capistrano to run 1-off jobs (even for projects running on kubernetes). But, eventually, I'd like to just use kubernetes :)


Thanks!

grosser:
the env input vars input seems to exist for kubernetes too
screen shot 2018-10-01 at 8 13 53 am

but then it does not add them to the release ... I tried adding them, but the place they'd need to go is the template_filler ... but that does not know about the job/deploy ... I'll take another peak later

rata:
On Mon, Oct 01, 2018 at 09:00:15AM -0700, Michael Grosser wrote:
> the env input vars input seems to exist for kubernetes too
> screen shot 2018-10-01 at 8 13 53 am

Yes, I see that too. Good :)

> 
> but then it does not add them to the release ... I tried adding them, but the place they'd need to go is the template_filler ... but that does not know about the job/deploy ... I'll take another peak later

Thank you very much! Let me know if I can help with testing or something else :)

rata:
If it is possible to add, it would be great! Thanks anyways for trying, really :)

PR #2670, opened 2018-04-03
retry at a more basic level to avoid random timeouts stopping deploys

grosser:


plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -109,18 +109,24 @@ def raw_events
...
+              @client.get_pod_log(name, namespace, container: container)
+            end
           else
             # still waiting, stream logs
             result = +""
             begin
               timeout_logs(wait) do
-                @client.watch_pod_log(name, namespace, container: container).each do |log|
-                  result << log << "\n"
+                SamsonKubernetes.retry_on_connection_errors do
```
grosser: this might end up duplicating the logs ... but we'll see ...

PR #2671, opened 2018-04-04
Add more Datadog metrics

ragurney:
[DEVEX-368](https://zendesk.atlassian.net/browse/DEVEX-368)

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -155,7 +155,9 @@ class Job < ActiveRecord::Base
   end
 
   def status!(status)
-    update_column(:status, status)
+    success = update_column(:status, status)
+    report_state if finished?
+    success
   end
 
   def short_reference
@@ -164,5 +166,10 @@ class Job < ActiveRecord::Base
     else
       commit
     end
+  end
+
+  def report_state
+    type = deploy ? 'deploy' : 'build'
+    ActiveSupport::Notifications.instrument('job_status.samson', type: type, status: status)
   end
 end
--- a/app/models/job_queue.rb
+++ b/app/models/job_queue.rb
@@ -142,8 +142,18 @@ class JobQueue
   def instrument
     ActiveSupport::Notifications.instrument(
       "job_queue.samson",
-      threads: @executing.length,
-      queued: @queue.length
+      jobs: {
+        executing: @executing.length,
+        queued: @queue.length,
+      },
+      deploys: {
+        executing: @executing.values.select { |je| deploy? je }.size,
+        queued: @queue.select { |i| deploy? i[:job_execution] }.size
+      }
     )
+  end
+
+  def deploy?(job_execution)
+    job_execution.respond_to?(:job) && job_execution.job.deploy
   end
 end
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -40,7 +40,9 @@ ActiveSupport::Notifications.subscribe("process_action.action_controller") do |*
   action = "action:#{event.payload.fetch(:action)}"
   format = "format:#{event.payload[:format] || 'all'}"
   format = "format:all" if format == "format:*/*"
-  status = event.payload[:status] || 401 # unauthorized redirect/error has no status because it is a `throw`
+  # Unauthorized and 500s have no status because it is a `throw`
+  # samson/vendor/bundle/gems/actionpack-5.1.4/lib/action_controller/metal/instrumentation.rb:35
+  status = event.payload[:status] || 'THR'
   tags = [controller, action, format]
 
   # db and view runtime are not set for actions without db/views
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -33,6 +33,16 @@ ActiveSupport::Notifications.subscribe("system_stats.samson") do |*, payload|
   payload.each { |key, value| Samson.statsd.gauge key.to_s, value }
 end
 
+ActiveSupport::Notifications.subscribe("wait_for_build.samson") do |*args|
+  event = ActiveSupport::Notifications::Event.new(*args)
+  tags = [
+    "project:#{event.payload.fetch(:project)}",
+    "external:#{event.payload.fetch(:external)}"
+  ]
+
+  Samson.statsd.timing "builds.time.wait_time", event.duration, tags: tags
+end
+
 # basic web stats
 ActiveSupport::Notifications.subscribe("process_action.action_controller") do |*args|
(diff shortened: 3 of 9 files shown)


app/models/job.rb
```diff
@@ -165,4 +166,11 @@ def short_reference
       commit
     end
   end
+
+  def instrument
+    if succeeded? || failed?
+      type = deploy ? 'deploys' : 'builds'
```
grosser: type should be singular ?

app/models/job.rb
```diff
@@ -156,6 +156,7 @@ def execution
 
   def status!(status)
     update_column(:status, status)
+    instrument
```
grosser: `report_state if finished?`

app/models/job_queue.rb
```diff
@@ -143,7 +143,15 @@ def instrument
...
+      queued: @queue.length,
+      deploys: {
+        executing: @executing.values.select { |je| deploy? je }.size,
+        queued: @queue.select { |i| deploy? i[:job_execution] }.size
+      }
     )
   end
+
+  def deploy?(job_execution)
+    job_execution.respond_to?(:job) ? job_execution.job.deploy.present? : false
```
grosser: associations don't need a present check, they are either there or nil

app/models/job_queue.rb
```diff
@@ -143,7 +143,15 @@ def instrument
     ActiveSupport::Notifications.instrument(
       "job_queue.samson",
       threads: @executing.length,
```
grosser: should change `threads` to `executing` for consistency

app/models/job_queue.rb
grosser: could simplify to `job_execution.respond_to?(:job) && job_execution.job.deploy`

config/initializers/statsd.rb
```diff
@@ -26,21 +26,42 @@ class << Samson
 end
 
 ActiveSupport::Notifications.subscribe("job_queue.samson") do |*, payload|
+  # deploys running/queued
+  deploys_metrics = payload.delete(:deploys)
```
grosser: modifying the payload means nobody else can use it

config/initializers/statsd.rb
```diff
@@ -26,21 +26,42 @@ class << Samson
...
+  deploys_metrics.each { |key, value| Samson.statsd.gauge "deploys.#{key}", value }
+
+  # total jobs queued/running reporting
   payload.each { |key, value| Samson.statsd.gauge "job.#{key}", value }
```
grosser: would make sense to reuse the metric and report a `deploy: true/false` tag ?

config/initializers/statsd.rb
```diff
@@ -26,21 +26,42 @@ class << Samson
...
 end
 
+ActiveSupport::Notifications.subscribe("job_status.samson") do |*, payload|
+  Samson.statsd.increment "#{payload[:type]}.#{payload[:status]}"
```
grosser: have a command prefix like `jobs.` so it's easy to figure out what belongs into it

grosser: prefer .fetch since it's useless if anything is missing

config/initializers/statsd.rb
```diff
@@ -26,21 +26,42 @@ class << Samson
...
 end
 
+ActiveSupport::Notifications.subscribe("wait_for_build.samson") do |*args|
+  event = ActiveSupport::Notifications::Event.new(*args)
+  tags = [
+    "project:#{event.payload.fetch(:project)}",
+    "gcb:#{event.payload.fetch(:gcb)}"
+  ]
+
+  Samson.statsd.histogram "builds.time.wait_time", event.duration, tags: tags
```
grosser: should report a timeing instead of a histogram ... or are they the same under the hood ?

config/initializers/statsd.rb
```diff
@@ -26,21 +26,42 @@ class << Samson
...
 ActiveSupport::Notifications.subscribe("process_action.action_controller") do |*args|
   event = ActiveSupport::Notifications::Event.new(*args)
   controller = "controller:#{event.payload.fetch(:controller)}"
   action = "action:#{event.payload.fetch(:action)}"
   format = "format:#{event.payload[:format] || 'all'}"
   format = "format:all" if format == "format:*/*"
-  status = event.payload[:status] || 401 # unauthorized redirect/error has no status because it is a `throw`
+  # Unauthorized and 500s have no status because it is a `throw`
+  # samson/vendor/bundle/gems/actionpack-5.1.4/lib/action_controller/metal/instrumentation.rb:35
+  status = event.payload[:status] || '401_500'
```
grosser: maybe something shorter like `THR` / `ERR` so they are all the same length ?

config/initializers/statsd.rb
ragurney: I think they are essentially the same thing under the hood going by the datadog docs

config/initializers/statsd.rb
```diff
@@ -26,21 +26,43 @@ class << Samson
 end
 
 ActiveSupport::Notifications.subscribe("job_queue.samson") do |*, payload|
-  payload.each { |key, value| Samson.statsd.gauge "job.#{key}", value }
+  # deploys running/queued
```
grosser: `[[:deploys, true], [:jobs, false]].each` would be nice here also to make sure both are tagged for simpler display logic (true/false vs true/"")

lib/samson/build_finder.rb
```diff
@@ -28,7 +28,10 @@ def cancelled!
     def ensure_successful_builds
       builds = find_or_create_builds
       builds.compact.each do |build|
-        wait_for_build_completion(build)
+        payload = { project: build.project.name, gcb: build.external? }
```
grosser: external does not necessary mean GCB since builds could come from anywhere ... so maybe use `external:` ?

test/models/job_execution_test.rb
```diff
@@ -288,14 +288,14 @@ def last_line_of_output
   it 'can run with a block' do
     x = :not_called
     execution = JobExecution.new('master', job) { x = :called }
-    assert execution.send(:run)
+    execution.send(:run)
```
grosser: so it fails now ? ... that seems fishy ...

ragurney: It didn't seem like we were using the return value of `status!` anywhere except for these tests, so it seemed unnecessary. I've added it back in.

grosser: afaik it was in the tests to make sure it did not crash to have a fast indicator for 'something blew up' instead of just missed `expects`

PR #2672, opened 2018-04-05
improve test coverage for key-resolver

grosser:
@ragurney 

not having a project is not a valid usecase anyway so better fail fast before we add more bugs ...


dragonfax:
@grosser Its unlike you to leave a broken unit test for so long.

grosser:
fixed up by making terminal-executor always require a project since that improves visibility too

PR #2673, opened 2018-04-05
Support multi project search for deploys and allow post to deploys index

ragurney:

### DIFF
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -10,6 +10,8 @@ class DeploysController < ApplicationController
   before_action :find_deploy, except: [:index, :active, :active_count, :new, :create, :confirm]
   before_action :stage, only: :new
 
+  skip_before_action :verify_authenticity_token, only: [:index]
+
   def active
     @deploys = deploys_scope.active
     render partial: 'deploys/table', layout: false if params[:partial]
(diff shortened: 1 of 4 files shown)


app/controllers/deploys_controller.rb
```diff
@@ -10,6 +10,8 @@ class DeploysController < ApplicationController
   before_action :find_deploy, except: [:index, :active, :active_count, :new, :create, :confirm]
   before_action :stage, only: :new
 
+  skip_before_action :verify_authenticity_token, only: [:index]
```
grosser: index should be get anyway, why does it need a skip ?

app/controllers/deploys_controller.rb
```diff
@@ -135,10 +137,13 @@ def search
       ).pluck(:id)
     end
 
-    if project_name = search[:project_name].presence
-      projects = Project.where(
-        Project.arel_table[:name].matches("%#{ActiveRecord::Base.send(:sanitize_sql_like, project_name)}%")
-      ).pluck(:id)
+    if project_names = search[:project_names].presence
```
grosser: could do: `if project_names = search[:project_names].to_s.split(",").presence`

config/routes.rb
```diff
@@ -68,6 +68,8 @@
     end
   end
 
+  match '/deploys', to: 'deploys#index', via: :post
```
grosser: deploys index exists ? ... use that ?

app/controllers/deploys_controller.rb
```diff
@@ -141,6 +141,13 @@ def search
       ).pluck(:id)
     end
 
+    # For external searches
+    if project_permalinks = search[:project_permalinks].to_s.split(",").presence
+      projects = Project.where(
```
grosser: `where(permalink: project_permalinks)`

app/controllers/deploys_controller.rb
```diff
@@ -141,6 +141,11 @@ def search
       ).pluck(:id)
     end
 
+    # For external searches
+    if project_permalinks = search[:project_permalinks].to_s.split(",").presence
+      projects = Project.where(permalink: project_permalinks).pluck(:id)
```
grosser: bug: when search for names and project_permalinks it won't work :D

grosser: `projects &= Project.where(permalink: project_permalinks).pluck(:id)`

PR #2674, opened 2018-04-10
Add route to create deploys from the API

jacobat:
* Allow creating deploys through the API

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low, just adding another endpoint.


grosser:
can you use `POST /deploys` instead ?

jacobat:
Of course.

PR #2677, opened 2018-04-11
DEVEX511 - Handle invalid bytes appearing in output streams

ragurney:


app/models/terminal_executor.rb
```diff
@@ -102,6 +102,7 @@ def timeout_execution(time, &block)
 
   def stream(from:, to:)
     from.each(256) do |chunk|
+      chunk = chunk.scrub
```
grosser: `scrub!` ?

PR #2678, opened 2018-04-12
[WIP] Stagger Deploys

ragurney:

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -102,7 +102,7 @@ class Job < ActiveRecord::Base
   end
 
   def queued?
-    JobQueue.queued?(id)
+    JobQueue.queued?(id) || DeployStaggerer.queued?(id)
   end
 
   def active?
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -40,7 +40,7 @@ class DeployService
       send_after_notifications(deploy)
     end
 
-    JobQueue.perform_later(job_execution, queue: deploy.job_execution_queue_name)
+    DeployStaggerer.process(job_execution, deploy)
 
     send_sse_deploy_update('start')
   end
--- a/app/models/deploy_staggerer.rb
+++ b/app/models/deploy_staggerer.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+class DeployStaggerer
+  include Singleton
+
+  STAGGER_INTERVAL = Integer(ENV['DEPLOY_STAGGER_INTERVAL'] || '0').seconds
+
+  class << self
+    delegate :process, :dequeue, :queued?, to: :instance
+  end
+
+  def process(job_execution, deploy)
+    @lock.synchronize do
+      if @queue.empty? && can_deploy?
+        send_to_job_queue(job_execution, deploy)
+      else
+        enqueue(job_execution, deploy)
+      end
+    end
+  end
+
+  def dequeue
+    @lock.synchronize do
+      if @queue.present? && can_deploy?
+        job_execution, deploy = @queue.shift
+        send_to_job_queue(job_execution, deploy)
+      end
+    end
+  end
+
+  def queued?(id)
+    @queue.any? { |(job_execution, _deploy)| job_execution.id == id }
+  end
+
+  private
+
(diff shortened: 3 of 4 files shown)


app/models/deploy_queue.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+class DeployQueue
```
grosser: better name would be good since we already have a queue ... DeployStaggerer or so ...

app/models/deploy_queue.rb
```diff
@@ -0,0 +1,56 @@
...
+  include Singleton
+
+  STAGGER_INTERVAL = 10.seconds
```
grosser: most likely want this ENV configureable

grosser: default should be off, small deployments don't need this

app/models/deploy_queue.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  class << self
+    delegate :process, :dequeue, :queued?, to: :instance
+  end
+
+  def process(job_execution, deploy)
+    @lock.synchronize do
+      if @queue.empty? && can_deploy?
```
grosser: `if blocked?` or `if full?`

app/models/deploy_queue.rb
```diff
@@ -0,0 +1,56 @@
...
+        send_to_job_queue(job_execution, deploy)
+      else
+        enqueue(job_execution, deploy)
+      end
```
grosser: send_to_job_queue and enqueue kinda sound the same :)

app/models/deploy_queue.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  def dequeue
+    @lock.synchronize do
+      if @queue.present? && can_deploy?
```
grosser: same check from above again ...

app/models/deploy_queue.rb
```diff
@@ -0,0 +1,56 @@
...
+  def can_deploy?
+    @last_deploy_started_at.nil? || Time.now - @last_deploy_started_at >= STAGGER_INTERVAL
+  end
+
+  def send_to_job_queue(job_execution, deploy)
+    @last_deploy_started_at = Time.now
+    JobQueue.perform_later(job_execution, queue: deploy.job_execution_queue_name)
+  end
+
+  private
```
grosser: private again

app/models/job.rb
```diff
@@ -102,7 +102,7 @@ def finished?
   end
 
   def queued?
-    JobQueue.queued?(id)
+    JobQueue.queued?(id) ||  DeployQueue.queued?(id)
```
grosser: race condition ... could be just moved from queued into JobQueue and it would not be marked as queued ?

app/views/deploys/_queued.html.erb
```diff
@@ -3,7 +3,7 @@
     <% if @deploy.job.executing? %>
       Deploy is running, refresh this page.
     <% elsif @deploy.job.queued? %>
-      Deploy is queued and it will be started when previous deploys have finished.
+      Deploy is queued and it will be started soon.
```
grosser: before was better ...

grosser:
maybe the job-queue or job-execution should get a "min start time" per job that is assigned when the job is queued depending on the current load ?

app/models/deploy_queue.rb
ragurney: You mean this env would be set up on a per project/stage basis?

grosser: samson env

app/models/deploy_queue.rb
ragurney: Are they the same? One is `A ∧ B` and the other is `¬A ∧ B`?

app/models/job.rb
ragurney: Hm, how would I fix that? Have some sort of global lock on both queues when this method is called?

grosser: - possibly handling queueing in job-queue
 - reuse job-queue lock

ragurney:
Superseded by: #2680

PR #2681, opened 2018-04-13
Do not tag an image with 'production' if its stage does not deploy code

ragurney:

### DIFF
--- a/plugins/gcloud/lib/samson_gcloud/image_tagger.rb
+++ b/plugins/gcloud/lib/samson_gcloud/image_tagger.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 module SamsonGcloud
   class ImageTagger
+    PRODUCTION_TAG = 'production'
+
     class << self
       # Note: not tagging builds from different project since that would be confusing ...
       # ideally do not tag any builds for projects that use shared builds ... but that is hard to know atm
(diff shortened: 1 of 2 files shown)


plugins/gcloud/lib/samson_gcloud/image_tagger.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 module SamsonGcloud
   class ImageTagger
+    PRODUCTION_TAG = 'production'
```
sathishavm: can we freeze this??

plugins/gcloud/lib/samson_gcloud/image_tagger.rb
```diff
@@ -15,11 +17,20 @@ def tag(deploy)
...
+
+          if env_permalink = stage.environments&.first&.permalink
+            tags << "env-#{env_permalink}"
           end
 
-          tag_image(deploy.stage.permalink, deploy, base, digest)
+          tags << stage.permalink
+
+          # From docker distro: https://tinyurl.com/ycgh7qsp
+          tags.grep(/^[\w][\w.-]{0,127}$/).each { |tag| tag_image(tag, deploy, base, digest) }
```
grosser: `.uniq` since the permalink can be production too

plugins/gcloud/lib/samson_gcloud/image_tagger.rb
```diff
@@ -15,11 +17,20 @@ def tag(deploy)
           next unless digest.match?(/(^|\/|\.)gcr.io\//) # gcr.io or https://gcr.io or region like asia.gcr.io
           base = digest.split('@').first
 
-          if deploy.stage.production?
-            tag_image('production', deploy, base, digest)
+          tags = []
```
sathishavm: These tags are common for all the build in the deploy, Then we need to bring the whole 'tags' array formation out of the build loop.

plugins/gcloud/lib/samson_gcloud/image_tagger.rb
ragurney: Would we want to avoid this collision completely by prefixing the stage permalink tags?

plugins/gcloud/lib/samson_gcloud/image_tagger.rb
ragurney: Isn't that taken care of with `frozen_string_literal`?

plugins/gcloud/lib/samson_gcloud/image_tagger.rb
ragurney: Each build would have a separate digest and base though, which are arguments passed to `tag_image`, so it makes sense to call that for each build?

sathishavm: That's right.. but following code is common for all the builds.
```
  tags = []
  stage = deploy.stage
  if stage.production? && !stage.no_code_deployed?
    tags << PRODUCTION_TAG
  end
  if env_permalink = stage.environments&.first&.permalink
    tags << "env-#{env_permalink}"
 end
  tags << stage.permalink
 # From docker distro: https://tinyurl.com/ycgh7qsp
   tags.grep(/^[\w][\w.-]{0,127}$/)
```

plugins/gcloud/lib/samson_gcloud/image_tagger.rb
sathishavm: hmmm, got confused on Constant, You are right!!

grosser:
yep

PR #2692, opened 2018-04-19
Disable fields on Env Group edit page if user lacks permissions to avoid confusion

ragurney:

### DIFF
--- a/plugins/env/app/views/environment_variable_groups/form.html.erb
+++ b/plugins/env/app/views/environment_variable_groups/form.html.erb
@@ -1,6 +1,10 @@
-<% title = (@group.new_record? ? "New group" : "Edit #{@group.name}") %>
+<%
+  title = (@group.new_record? ? "New group" : "Edit #{@group.name}")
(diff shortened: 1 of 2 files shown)


plugins/env/app/views/environment_variable_groups/form.html.erb
```diff
@@ -1,6 +1,10 @@
-<% title = (@group.new_record? ? "New group" : "Edit #{@group.name}") %>
+<%
+  title = (@group.new_record? ? "New group" : "Edit #{@group.name}")
+  can_edit = can?(:write, :environment_variable_groups, @group)
```
grosser: could call this `editable`

plugins/env/app/views/environment_variable_groups/form.html.erb
```diff
@@ -10,8 +14,14 @@
 <%= page_title title %>
 
 <section>
+  <% if !can_edit %>
+    <div class="alert alert-warning">
+      You must either be an admin or have project admin privileges for all projects that use this environment group to
```
grosser: ❤️

PR #2693, opened 2018-04-19
make sure all elements in the config have the same project and unique…

grosser:
… role per file

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_verifier.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_verifier.rb
@@ -36,14 +36,22 @@ module Kubernetes
       @errors.presence
     end
 
-    def self.verify_group(elements)
+    def self.verify_groups(element_groups)
+      elements = element_groups.flatten(1)
       return if elements.empty?
       return if elements.any? { |r| r.dig(:metadata, :annotations, :"samson/multi_project") }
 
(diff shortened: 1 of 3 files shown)


grosser:
fyi @zendesk/compute @zendesk/devex this might break a few deploys but should be easy to fix ...

plugins/kubernetes/app/models/kubernetes/role_verifier.rb
```diff
@@ -36,14 +36,22 @@ def verify
...
-      errors << "metadata.labels.role must be set and unique" if roles.uniq.size != elements.size
+      element_groups.each do |element_group|
+        roles = element_group.map { |r| r.dig(:metadata, :labels, :role) }.uniq
+        if roles.size != 1 || roles == [nil]
+          errors << "metadata.labels.role must be set and consistent in each config file"
+        end
+      end
+
+      roles = element_groups.map(&:first).map { |r| r.dig(:metadata, :labels, :role) }
+      errors << "metadata.labels.role must be set and unique" if roles.uniq.size != element_groups.size
```
ragurney: 2 error checks for one 👍

grosser: it must be the same inside each role and unique across all roles

PR #2703, opened 2018-04-27
PAAS-1500 support PodDisruptionBudget

grosser:
@ragurney @jonmoter

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/cluster.rb
+++ b/plugins/kubernetes/app/models/kubernetes/cluster.rb
@@ -39,6 +39,10 @@ module Kubernetes
       @batch_client ||= build_client 'batch/v1'
     end
 
+    def policy_client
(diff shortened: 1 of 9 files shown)


dragonfax:
What does this actually do?

grosser:
it allows deploying this new resource, it makes deployments always keep a certain number of replicas when we do things like take down all nodes

plugins/kubernetes/app/models/kubernetes/cluster.rb
```diff
@@ -39,6 +39,10 @@ def batch_client
       @batch_client ||= build_client 'batch/v1'
     end
 
+    def policy_client
+      @policy_client ||= build_client 'policy/v1beta1'
+    end
```
jonmoter: Have you confirmed this works in k8s 1.5?  Sandbox is on 1.6

PR #2714, opened 2018-04-30
 Added new env DOCKER_FORCE_EXTERNAL_BUILD to force disabling image build 

sathishavm:
* Added new env DOCKER_FORCE_EXTERNAL_BUILD, that force disabling the image build options both in UI and business logic.
screen shot 2018-04-30 at 1 37 49 pm

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/DEVEX-616

### Risks
- Level: Med


app/views/projects/_docker_fields.html.erb
```diff
@@ -2,14 +2,23 @@
...
-        Project::DOCKER_BUILD_METHODS.map { |method| [method[:label], method[:method]] },
-        {},
-        class: "form-control"
-    %>
-  <% end %>
+  <% form_obj = form.object %>
+  <% if form_obj.force_external_build? %>
+    <%= form.hidden_field :docker_build_method, value: "docker_image_building_disabled" if form_obj.new_record?%>
+    <span>
+      <%= lock_icon %>
```
grosser: lock kinda looks like there must be a way to unlock it ... maybe no icon ?

grosser:
there are still a lot of places that use the boolean flags underneath docker_build_method ... they need to be changed ?

app/views/projects/_docker_fields.html.erb
```diff
@@ -2,14 +2,23 @@
...
+      Configured to use externally built docker images only.
```
ragurney: Maybe add some help text on how to set up external builds?

ragurney:
Do we still need to show the dockerfiles field for `builds_in_environment`? Even if people are using GCB to create those images, we still need the dockerfiles list because that's how Samson currently finds matching builds.

PR #2715, opened 2018-04-30
auto-set dynamic PodDisruptionBudget from annotations

grosser:
Absolute minAvailable is not good enough, we want to say that for example 80% of the replicas should be available.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -77,11 +77,44 @@ module Kubernetes
 
     # dynamically fill out the templates and store the result
     def store_resource_template
+      add_pod_disruption_budget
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -77,13 +77,46 @@ def resource_template=(value)
...
     end
 
+    def add_pod_disruption_budget
+      min_available = nil
+      raw_template.detect do |t|
+        min_available ||= t.dig(:metadata, :annotations, :"samson/minAvailable")
+      end
+      return unless min_available
+
+      target = if percent = min_available.to_s[/\A(\d+)\s*%\z/, 1]
```
jonmoter: When there's an inline regex like this, useful to have a comment explaining what the regex is supposed to be doing.

PR #2720, opened 2018-05-01
Handle Stage deletes without pipeline plugin

### LINKED ISSUE #2719: Deleting stage fails (tries to delete nil)
On version v2236, deleting a stage fails with:
```
NoMethodError (undefined method `delete' for nil:NilClass):

app/models/stage.rb:241:in `block in destroy_stage_pipeline'
app/models/stage.rb:240:in `each'
app/models/stage.rb:240:in `destroy_stage_pipeline'
app/controllers/stages_controller.rb:75:in `destroy'
```
At https://github.com/zendesk/samson/blob/82f72d32a6904669ab761348b620e3cdfceeca67/app/models/stage.rb#L241 the next_stage_id is nil.

viraptor:
When the pipeline plugin is disabled, the next_stage_ids is `nil`
rather than an empty array.
The test simulates it with a stub since we can't unload a plugin
modification of the model.

Closes #2719 

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -238,7 +238,7 @@ class Stage < ActiveRecord::Base
 
   def destroy_stage_pipeline
     (project.stages - [self]).each do |s|
-      if s.next_stage_ids.delete(id)
+      if s.next_stage_ids&.delete(id)
         s.save(validate: false)
       end
     end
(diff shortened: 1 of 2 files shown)


grosser:
that logic should live in the pipeline plugin :/

grosser:
I guess good enough for now, with a rubocop fix ... but it should get moved.

grosser:
might also be smart to define the column as not-null + default value that it expects ... but that's for another PR

grosser:
but please keep going with the PRs, always need more of them ❤️ ✨

PR #2724, opened 2018-05-01
Minor UX changes in project edit page

hdt91:
- Add dynamic links to preview each included env group
- Fix few grammar issues

#### Before
before_preview

#### After
after_preview

/cc @zendesk/samson @zendesk/devex 

### References
 - Jira link: https://zendesk.atlassian.net/browse/DEVEX-527

### Risks
- Level: Low - only UI changes

### DIFF
--- a/plugins/env/app/assets/javascripts/env/application.js
+++ b/plugins/env/app/assets/javascripts/env/application.js
@@ -49,6 +49,28 @@
     return $new_row;
   }
 
+  // Update the query string of the associated preview link
+  // with the given group ID, and toggle its visibility
+  function formatEnvGroupPreviewLink($preview, val) {
+    $preview[0].search = $.param({ group_id: val });
+    $preview.toggle(!!val);
+  }
+
+  // Select the pair of env group input and preview link to:
+  // - Format the query string of the preview link
+  // - Update the query string when the env group input changes
+  function setupEnvGroupPreview($input) {
+    var $select = $input.find('select');
+    var $preview = $input.find('.checkbox a');
+
+    $select.on('change', function(e) {
+      e.preventDefault();
+      formatEnvGroupPreviewLink($preview, $select.val());
+    });
+
+    $select.trigger('change');
+  }
+
   function withoutSelectpicker($row, callback){
     var $picker = $row.find('.selectpicker');
     $picker.selectpicker('destroy').addClass('selectpicker'); // normalize existing so they are ready to copy
@@ -56,7 +78,10 @@
     var rows = callback().concat([$row]);
 
     // add selectpicker to copied and original rows
-    $.each(rows, function(k,v) { v.find(".selectpicker").selectpicker(); });
+    $.each(rows, function(idx, $row) {
+      $row.find(".selectpicker").selectpicker();
+      setupEnvGroupPreview($row);
+    });
   }
 
   $(document).on("click", ".duplicate_previous_row", function(e){
@@ -92,4 +117,14 @@
       });
     });
   });
+
+  // Display preview link dynamically when env group dropdown changes
+  $(document).ready(function() {
+    var $inputs = $(".env_group_inputs");
+
+    $inputs.each(function(idx, input) {
+      setupEnvGroupPreview($(input));
+    });
+  });
+
 }());
--- a/plugins/env/app/views/samson_env/_fields.html.erb
+++ b/plugins/env/app/views/samson_env/_fields.html.erb
@@ -15,16 +15,20 @@
   <% environment_variable_groups = EnvironmentVariableGroup.order(:name).map { |g| [g.name, g.id] } %>
   <% if environment_variable_groups.any? %>
     <% ids.each do |id| %>
-      <div class="form-group">
-        <div class="col-lg-2 col-lg-offset-2">
+      <div class="form-group env_group_inputs">
+        <div class="col-lg-3">
           <%= live_select_tag "project[environment_variable_group_ids][]",
             options_for_select([[nil, nil]] + environment_variable_groups, id), placeholder: "Name"
           %>
+        </div>
+        <div class="col-lg-2 checkbox">
+          <%# updated dynammically in the client-side JS %>
+          <%= link_to "Preview", preview_environment_variable_groups_path %>
         </div>
       </div>
     <% end %>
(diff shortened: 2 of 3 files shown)


plugins/env/app/views/samson_env/_fields.html.erb
```diff
@@ -15,16 +15,21 @@
...
-        <div class="col-lg-2 col-lg-offset-2">
+      <div class="form-group env_group_inputs">
+        <div class="col-lg-3">
           <%= live_select_tag "project[environment_variable_group_ids][]",
             options_for_select([[nil, nil]] + environment_variable_groups, id), placeholder: "Name"
           %>
         </div>
+        <div class="col-lg-2 checkbox">
+          <!-- updated dynammically in the client-side JS -->
+          <%= link_to "Preview this group", preview_environment_variable_groups_path %>
```
grosser: "Preview" ... maybe even "Previous group" would be enough

plugins/env/app/views/samson_env/_environment_variables_fields.html.erb
```diff
@@ -29,5 +29,5 @@
     <% end %>
   </div>
 <% end %>
-<%= link_to "Add row", "#", class: "duplicate_previous_row" %> |
+<%= link_to "Add a row", "#", class: "duplicate_previous_row" %> |
```
grosser: I liked the less-words variant :)

grosser:
maybe always render the link and then make that have a handler that looks to the right and then goes to that page

plugins/env/app/assets/javascripts/env/application.js
```diff
@@ -92,4 +117,16 @@
       });
     });
   });
+
+  // Display preview link dynamically when env group dropdown changes
+  $(document).ready(function() {
+    var $inputs = $(".env_group_inputs");
```
ragurney: Another possible approach would be: 
Listen to 'env_group_inputs` for the on change: 
```
$('.env_group_inputs').on('change', function(e) {
  $input = $(e.trigger);
  updatePreviewLink($input);
}
```
and then render the previews for any existing inputs on page load. Might be a little cleaner, and avoids attaching listeners for each input. I'm ok with either way, just a thought ¯\_(ツ)_/¯

plugins/env/app/views/samson_env/_fields.html.erb
```diff
@@ -15,16 +15,21 @@
...
-      <div class="form-group">
-        <div class="col-lg-2 col-lg-offset-2">
+      <div class="form-group env_group_inputs">
+        <div class="col-lg-3">
           <%= live_select_tag "project[environment_variable_group_ids][]",
             options_for_select([[nil, nil]] + environment_variable_groups, id), placeholder: "Name"
           %>
         </div>
+        <div class="col-lg-2 checkbox">
+          <!-- updated dynammically in the client-side JS -->
```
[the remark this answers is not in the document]
hdt91: Will update

plugins/env/app/assets/javascripts/env/application.js
hdt91: what is this `e.trigger` thing? Cannot find it on `MDN` or in my dev env (Chrome btw).

ragurney: Oops, I meant `e.target`. Sorry I always want to call it trigger

hdt91: If that is the case, that is similar to my lengthy response above. Basically, doing this way `e.target` is the DOM element that can be buried deep into a bunch of wrapped HTML and navigating up/down to find the associated `href` is generally more brittle. This indeed reduces the amount of binding but it only matters if the list is super big and dynamic. Also at the initial loading page there is no `change` event to piggy-back also :)

ragurney: > Also at the initial loading page there is no change event to piggy-back also :)

Right, but you could render the links with the correct urls on page load? Then use the js to only handle updating the links? 

I agree that the jquery would be slightly more complicated to locate the link. I'm ok with the code as-is, just a suggestion.

plugins/env/app/assets/javascripts/env/application.js
```diff
@@ -92,4 +117,16 @@
...
+
+    if (!$inputs) { return; }
```
grosser: it's never false ... can check `.length` though ... or just delete this line since the .each works

plugins/env/app/assets/javascripts/env/application.js
```diff
@@ -49,14 +49,39 @@
...
+  function formatPreview($preview, val) {
+    if (val) {
+      $preview[0].search = $.param({ group_id: val });
+      $preview.show();
+    } else {
+      $preview.hide();
+    }
+  }
+
+  function setupPreview($input) {
```
grosser: add a comment explaining what it does and pick a less generic name `setupEnvGroupPreview`

grosser: like "When the user does A we do B"

plugins/env/app/assets/javascripts/env/application.js
```diff
@@ -49,14 +49,39 @@
...
+    var $select = $input.find('select');
+    var $preview = $input.find('.checkbox a');
+
+    formatPreview($preview, $select.val());
+
+    $select.on('change', function(e) {
+      e.preventDefault();
+      formatPreview($preview, $select.val());
+    });
```
grosser: might be able to remove the duplication by appending a `.trigger('change')` here so it always gets executed once

plugins/env/app/assets/javascripts/env/application.js
```diff
@@ -49,14 +49,39 @@
...
   }
 
+  // Update the query string of the associated preview link
+  function formatPreview($preview, val) {
+    if (val) {
+      $preview[0].search = $.param({ group_id: val });
+      $preview.show();
+    } else {
+      $preview.hide();
+    }
```
grosser: could do:

```
$preview[0].search = $.param({ group_id: val });
$preview.toggle(val)
```

hdt91:
thanks @grosser Updated the PR after all of your suggestions. Please comment if you have any other concern, otherwise I'll merge later today :)

PR #2730, opened 2018-05-02
Improve Rollbar plugin behaviour and config

viraptor:
Parameter ROLLBAR_URL is not required - the default urls are already
configured by the gem. Additionally it was incorrectly used for
the web_base parameter, which may not be the same url.

Make both optional, separate, and document all the parameters used.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -197,3 +197,9 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # https://github.com/omniauth/omniauth-ldap/blob/master/README.md
 # https://github.com/omniauth/omniauth-ldap/blob/master/lib/omniauth/strategies/ldap.rb#L17
 # USE_LDAP_UID_AS_EXTERNAL_ID=1
+
+## Feature: Rollbar error reporting
+# Report Samson's internal failures to Rollbar service
+# ROLLBAR_ACCESS_TOKEN= # API token, required
+# ROLLBAR_URL=https://api.rollbar.com # optional, needed only for custom/test deployments
+# ROLLBAR_WEB_BASE=https://rollbar.com # optional, needed only for custom/test deployments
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -197,3 +197,9 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
 # https://github.com/omniauth/omniauth-ldap/blob/master/README.md
 # https://github.com/omniauth/omniauth-ldap/blob/master/lib/omniauth/strategies/ldap.rb#L17
 # USE_LDAP_UID_AS_EXTERNAL_ID=1
+
+## Feature: Rollbar error reporting
+# Report Samson's internal failures to Rollbar service
+# ROLLBAR_ACCESS_TOKEN= # API token, required
+# ROLLBAR_URL=https://api.rollbar.com # optional, needed only for custom/test deployments
```
grosser: the ENV.fetch will blow up without this ?

grosser: ah nvm ... just figured it out :D

.env.example
```diff
@@ -197,3 +197,9 @@ PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:6
...
+# ROLLBAR_WEB_BASE=https://rollbar.com # optional, needed only for custom/test deployments
```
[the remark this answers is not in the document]
grosser: ... or I guess just `optional` since that's what it kinda means ... set this if you want to set it ...

plugins/rollbar/config/initializers/rollbar.rb
```diff
@@ -4,8 +4,8 @@
   Rollbar.configure do |config|
     config.access_token = token
     config.environment = Rails.env
-    config.endpoint = ENV.fetch('ROLLBAR_URL') + '/api/1/item/'
-    config.web_base = ENV.fetch('ROLLBAR_URL')
+    config.endpoint = ENV.fetch('ROLLBAR_URL') + '/api/1/item/' if ENV.key?('ROLLBAR_URL')
```
grosser: would prefer dry like:
```
if url = ENV['ROLLBAR_URL']
  config.endpoint = url + x
```

viraptor:
Is there anything else I should add for this to be merged?

grosser:
❤️

PR #2740, opened 2018-05-08
Fix readme formatting and display actual job status

yifeiwu:
Previously, the jenkins statuses would show as unstable for in progress jobs. Now we print what the exact status of the job is if it's not passing(blue) so deployers can focus on the red jobs(if there are any). 

Also fix markup for readme.

@zendesk/samson

### Risks
- Level: Low, changes how statuses are displayed.


yifeiwu:
Thanks @grosser, I think this is good to go :)

PR #2741, opened 2018-05-10
Add link back to triggering deploy

ragurney:
There have been a couple of cases where people don't know if a stage was triggered manually or via pipeline. This adds a nice convenience message that tells people the stage was triggered via a pipelined stage and links to the deploy that kicked it off.

screen shot 2018-05-10 at 12 09 58 pm

### DIFF
--- a/app/views/deploys/_header.html.erb
+++ b/app/views/deploys/_header.html.erb
@@ -1,5 +1,5 @@
 <h1>
-  <%= @project.name %>
+  <%= "#{@project.name} - Deploy ##{@deploy.id}" %>
 
   <div class="pull-right">
     <% if deployer_for_project? %>
@@ -23,5 +23,5 @@
   <%= render 'jobs/restarting' %>
 <% else %>
   <%= status_panel @deploy %>
-  <%= Samson::Hooks.render_views(:deploys_header, self) %>
+  <%= Samson::Hooks.render_views(:deploys_header, self, deploy: @deploy) %>
 <% end %>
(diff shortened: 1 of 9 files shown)


app/views/deploys/_header.html.erb
```diff
@@ -23,5 +23,5 @@
   <%= render 'jobs/restarting' %>
 <% else %>
   <%= status_panel @deploy %>
-  <%= Samson::Hooks.render_views(:deploys_header, self) %>
+  <%= Samson::Hooks.render_views(:deploys_header, self, deploy: @deploy) %>
```
grosser: partials can use `@deploy`

plugins/pipelines/db/migrate/20180510172630_add_pipelined_to_deploys.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPipelinedToDeploys < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :deploys, :triggering_deploy, foreign_key: {to_table: :deploys}, type: :integer
```
grosser: we don't use the index since we don't show which deploys were triggered

grosser: generally prefer the more sql-style migrations since it makes it more obvious what will be done ... but ok either way

plugins/pipelines/lib/samson_pipelines/samson_plugin.rb
```diff
@@ -14,12 +14,13 @@ def start_pipelined_stages(job, success, output)
 
     private
 
-    def deploy_to_stage(stage, template_deploy, output)
-      deploy_service = DeployService.new(template_deploy.user)
+    def deploy_to_stage(stage, previous_deploy, output)
```
grosser: not really an obvious 2nd argument, also should be named triggering_deploy and not previous ?

grosser: nvm this was just a rename :D

grosser:
don't care much, either way is fine ... more clutter or more verification :)

PR #2746, opened 2018-05-14
Add logging to JobExecution

ragurney:
Part of [DEVEX-598](https://zendesk.atlassian.net/browse/DEVEX-598). Adding some logging to try to pin down exactly what is going on here, as I have yet been able to reproduce this locally :(

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -29,6 +29,7 @@ class JobExecution
     @repository.executor = @executor
 
     on_finish do
+      Rails.logger.info("Calling finish callback for Job Execution #{id}")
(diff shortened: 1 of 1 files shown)


app/models/job_execution.rb
```diff
@@ -29,6 +29,7 @@ def initialize(reference, job, env: {}, output: OutputBuffer.new, &block)
     @repository.executor = @executor
 
     on_finish do
+      Rails.logger.info("Calling finish callback for Job Execution #{id}")
```
grosser: might be easier to read/parse when using a consistent prefix like `JobExecution#on_finish`

PR #2747, opened 2018-05-15
Replace tooltip that has passive aggressive tone

Gwash3189:
/cc @zendesk/samson

### Description 

The tooltip that appears below the `Redeploy` button has a passive aggressive tone. There are many reasons why a successful deploy might need to be redeployed. For instance, if the deploy was based off a branch, maybe the branch has new commits and a redeploy would deploy those changes. 

This pull request changes the tooltip to have a neutral tone but still provide the same value (warn against redeploying a successful deploy).  

### Screenshots

why

screen shot 2018-05-16 at 7 55 46 am

### Tasks
 - [x] :+1: from team

### References
 - Jira link: N/A

### Risks
- Level: Low, changes the display value of a tooltip

### DIFF
--- a/test/helpers/deploys_helper_test.rb
+++ b/test/helpers/deploys_helper_test.rb
@@ -144,7 +144,7 @@ describe DeploysHelper do
   end
 
   describe "#redeploy_button" do
-    let(:redeploy_warning) { "Previous deploy succeeded" }
+    let(:redeploy_warning) { "Previous deploy succeeded." }
 
     before do
       @deploy = deploy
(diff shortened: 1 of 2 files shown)


test/helpers/deploys_helper_test.rb
```diff
@@ -144,7 +144,7 @@
   end
 
   describe "#redeploy_button" do
-    let(:redeploy_warning) { "Why? This deploy succeeded." }
+    let(:redeploy_warning) { "Previous deploy succeeded" }
```
ragurney: Missing period?

Gwash3189: Good eye 😄. Thanks!

Issue #2744, opened 2018-05-14
RBENV_DIR in builds


viraptor:
I've run into an unexpected Samson behaviour.

When running a ruby script inside the deployment actions, RBENV_DIR is inherited from Samson's runtime in my deployment. That means anything expecting the right ruby to be set via `.ruby-version` will fail :(

While the fix is easy (`unset RBENV_DIR` at the beginning of the build), it needs to be done for every project separately and all teams need to learn about it. It would be great if I could configure Samson to drop some environment variables globally. For example never pass the `RBENV_DIR` / `PATH` / `RAILS_ENV` / `HOME` to the jobs it's running.

grosser:
whitelisted env vars are defined here:

https://github.com/zendesk/samson/blob/master/app/models/terminal_executor.rb#L166-L175

we might be able to remove RBENV_DIR ... 
PATH needs to stay or nothing works like `ls` etc
RAILS_ENV etc is already not passed in

grosser:
keep these issues coming ... such a basic flaw that we never addresses since we added that into the base commands years ago :D

PR #2748, opened 2018-05-18
do not pass on the RBENV_DIR since that breaks automated ruby sitching

grosser:
fixes https://github.com/zendesk/samson/issues/2744


grosser:
confirmed working in staging 🎉

PR #2753, opened 2018-05-22
allow reverting PodDisruptionBudget, they need to be created without …

grosser:
…resourceVersion

I was rolling back a failed deploy when I saw:

```
[17:01:31] Rolling back PodX role kube-stats
[17:01:35] FAILED: resourceVersion may not be set on objects to be created
```

which means the rollback failed and the resource stayed deleted (deploy is delete+create)
so need to fix that :)

/cc @zendesk/compute @ragurney

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/resource_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/resource_test.rb
@@ -49,10 +49,11 @@ describe Kubernetes::Resource do
 
   before { Kubernetes::Resource::Base.any_instance.expects(:sleep).with { raise }.never }
 
-  it "does modify passed in template" do
+  it "does not modify passed in template" do
     content = File.read(File.expand_path("../../../app/models/kubernetes/resource.rb", __dir__))
-    reset_code_usages = 3
-    content.scan(/@template.*(=|dig_set)/).size.must_equal content.scan('restore_template do').size + reset_code_usages
+    restore_usages = content.scan('restore_template do').size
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -805,5 +806,18 @@ def deployment_stub(replica_count)
         end
       end
     end
+
+    describe "#revert" do
+      it "deletes and then creates without resourceVersion because that is not allowed" do
+        with = ->(request) { request.body.wont_include "resourceVersion"; true }
```
adammw: I find this test case rather hard to understand.. not sure if it's the variable names or the nesting of random requests

plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -805,5 +806,18 @@ def deployment_stub(replica_count)
...
+        assert_request(:get, url, to_return: [{body: '{}'}, {status: 404}]) do
+          assert_request(:delete, url, to_return: {body: '{}'}) do
+            assert_request(:post, File.dirname(url), with: with, to_return: {body: '{}'}) do
```
adammw: i.e. would like to see this be a variable like create_url - that way i at least can track what's going on

grosser: dried it up a bit

grosser:
confirmed rolling back in staging :)

PR #2756, opened 2018-05-24
alert users visibly if they are deploying an env diff

grosser:
@ragurney 
/cc @dragonfax 

before: user has no idea what is going out
screen shot 2018-05-23 at 4 54 26 pm

after: user is alerted
screen shot 2018-05-23 at 4 59 27 pm


plugins/env/app/views/samson_env/_deploy_tab_nav.html.erb
```diff
@@ -1 +1,23 @@
-<li><a href="#environment-variables" data-toggle="tab">Environment variables</a></li>
+<%
+  previous_env = @deploy.stage ? @deploy.previous_successful_deploy&.env_state : nil
+  current_env = @deploy.project ? @deploy.retrieve_env_state : nil
+  diff = (previous_env.presence != current_env.presence)
```
ragurney: Might be a little more clean to have this as a cached value in a helper method rather than using `content_for` between views?

plugins/env/app/views/samson_env/_deploy_tab_nav.html.erb
```diff
@@ -1 +1,23 @@
...
+  <% else %>
+    No changes.
+  <% end %>
+<% end %>
+
+<li>
+  <a href="#environment-variables" data-toggle="tab">
+    Environment variables
+    <% if diff %>
+      <span class="badge">!</span>
```
ragurney: ✨

plugins/env/app/views/samson_env/_deploy_tab_nav.html.erb
grosser: yeah was my first thought too, but then it's a shared instance var that gets cached ... which is kinda funky/complex too ... I'll check out how it looks and merge if it's uglier ...

grosser: yeah it's ugly ... need to cache false then and assign arrays and stuff ...

grosser: but testing breaks ... so I guess helper it is ...

PR #2757, opened 2018-05-24
remove multi origin stuff that broke in rails 5.2

grosser:
@ragurney @dragonfax @henders 

reverts https://github.com/zendesk/samson/pull/586 which I hope is no longer an issue ...


grosser:
fixed it by unsetting the env vars I'm removing here

PR #2758, opened 2018-05-24
[DEVEX-388] Add deploy start event to datadog notification

ragurney:
This is to allow the use of event overlays to show when a deploy has started. Can select deploy start/finished events by tag `started` or `finished`. 

/cc @zendesk/samson @jasonmartens 

[DEVEX-388](https://zendesk.atlassian.net/browse/DEVEX-388)

### DIFF
--- a/plugins/datadog/app/models/datadog_notification.rb
+++ b/plugins/datadog/app/models/datadog_notification.rb
@@ -8,10 +8,17 @@ class DatadogNotification
     @stage = @deploy.stage
   end
 
-  def deliver
+  def deliver(additional_tags: [], now: false)
     Rails.logger.info "Sending Datadog notification..."
 
-    status = @deploy.succeeded? ? "success" : "error"
+    status =
+      if @deploy.active?
+        "info"
+      elsif @deploy.succeeded?
+        "success"
+      else
+        "error"
+      end
 
     event = Dogapi::Event.new(
       body,
@@ -20,8 +27,8 @@ class DatadogNotification
       event_object: Digest::MD5.hexdigest("#{Time.new}|#{rand}"),
       alert_type: status,
       source_type_name: "samson",
-      date_happened: @deploy.updated_at,
-      tags: @stage.datadog_tags_as_array + ["deploy"]
+      date_happened: now ? Time.now : @deploy.updated_at,
+      tags: @stage.datadog_tags_as_array + ["deploy", *additional_tags]
     )
 
     client = Dogapi::Client.new(api_key, nil, "")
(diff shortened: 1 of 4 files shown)


plugins/datadog/app/models/datadog_notification.rb
```diff
@@ -20,8 +20,8 @@ def deliver
       event_object: Digest::MD5.hexdigest("#{Time.new}|#{rand}"),
       alert_type: status,
       source_type_name: "samson",
-      date_happened: @deploy.updated_at,
-      tags: @stage.datadog_tags_as_array + ["deploy"]
+      date_happened: now ? Time.now : @deploy.updated_at,
```
ragurney: Alternative would be to create a new hook `:during_deploy` in addition to the existing `:before_deploy` and `:after_deploy` to use the `@deploy.updated_at` but this seemed sufficient / more clean and reusable for an event timestamp.

ragurney: Alternative would be to create a new hook :during_deploy in addition to the existing :before_deploy and :after_deploy to use the @deploy.updated_at but this seemed sufficient / more clean and reusable for an event timestamp.

plugins/datadog/app/models/datadog_notification.rb
```diff
@@ -8,10 +8,10 @@ def initialize(deploy)
     @stage = @deploy.stage
   end
 
-  def deliver
+  def deliver(additional_tags: [], now: false)
     Rails.logger.info "Sending Datadog notification..."
 
-    status = @deploy.succeeded? ? "success" : "error"
+    status = @deploy.active? || @deploy.succeeded? ? "success" : "error"
```
grosser: wut ... status is "success" when it's active ?

plugins/datadog/test/models/datadog_notification_test.rb
```diff
@@ -1,7 +1,90 @@
 # frozen_string_literal: true
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 12
+SingleCov.covered!
 
 describe DatadogNotification do
```
grosser: a tiny PR to just bump coverage would have been awesome :)

plugins/datadog/app/models/datadog_notification.rb
ragurney: Would `info` be more appropriate?

grosser:
was thinking "running"/"success"/"error" if that makes sense in datadog

ragurney:
From these [docs](https://docs.datadoghq.com/api/?lang=ruby#post-an-event) only supports error, warning, info, and success

grosser:
Maybe "info" ?

PR #2762, opened 2018-05-26
use page for pagination ... faster/simpler

grosser:
@zendesk/devex 

before:
screen shot 2018-05-25 at 5 25 16 pm

after:
screen shot 2018-05-25 at 5 41 42 pm


app/controllers/locks_controller.rb
```diff
@@ -7,7 +7,7 @@ class LocksController < ApplicationController
   before_action :authorize_resource!
 
   def index
-    render json: {locks: Lock.page(page).per(1000)}
+    render json: {locks: pagy(Lock, page: page, items: 1000).last}
```
ragurney: Kind of confusing if you don't know this returns `[@pagy, @collection]`

app/helpers/application_helper.rb
```diff
@@ -277,9 +278,11 @@ def live_select_tag(name, values, options = {})
     select_tag name, values, Samson::FormBuilder::LIVE_SELECT_OPTIONS.merge(options)
   end
 
-  def paginate(objects, *)
-    result = super
-    result << " #{objects.total_count} records" if objects.total_pages > 1
+  def paginate(pagy, _, *)
+    result = (pagy.pages > 1 ? pagy_nav_bootstrap(pagy) : "").html_safe
+    if pagy.pages > 1
```
ragurney: Combine these two same checks into one if/else?

app/views/audits/_table.html.erb
```diff
@@ -52,6 +52,6 @@
   </tbody>
 </table>
 
-<%= paginate audits if audits.respond_to?(:total_pages) %>
+<%= paginate @pagy, audits if @pagy %>
```
ragurney: Why do we pass the collection into `paginate`? It looks like it only uses the `pagy` object? Could maybe change this method to simply `paginate` which would only take `@pagy`, or expect an instance var `@pagy` to be defined and raise if not?

grosser: copy-pasted from cerebro where we say "15 Users" ... but I guess that's kinda useless anyway ... removing

PR #2767, opened 2018-05-29
Deploy all releases in parallel

dasch:
This will totally work.

/cc @zendesk/samson

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -372,14 +372,18 @@ module Kubernetes
 
     # updates resources via kubernetes api
     def deploy(release_docs)
-      release_docs.each do |release_doc|
+      # We deploy each resource in parallel.
+      resources = release_docs.flat_map do |release_doc|
         puts_action "Deploying", release_doc
+
         if release_doc.blue_green_color
-          non_service_resources(release_doc).each(&:deploy)
+          non_service_resources(release_doc)
         else
-          release_doc.deploy
+          release_doc.deploy_group.kubernetes_cluster # cache before threading
+          [release_doc]
         end
       end
+      Samson::Parallelizer.map(resources, db: true, &:deploy)
     end
 
     def deploy_and_watch(release_docs)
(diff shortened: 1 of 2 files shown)


grosser:
we did some work to make less deploys start in parallel, so this would kinda undo that ... also it should not take that long to start the deploys, they are just put into the queue

/cc @ragurney

dasch:
Because of Kubernetes API server load? We currently spend 2.5 minutes sending out the release docs, and this will only grow as we add more pods (6 more coming).

How many can safely be sent in parallel to a single API server?

grosser:
ah this is inside the deploy-executor ... that should be fine

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -372,14 +372,17 @@ def grouped_deploy_group_roles
...
-      release_docs.each do |release_doc|
+      release_docs.map do |release_doc|
         puts_action "Deploying", release_doc
-        if release_doc.blue_green_color
-          non_service_resources(release_doc).each(&:deploy)
-        else
-          release_doc.deploy
+
+        Thread.new do
+          if release_doc.blue_green_color
```
grosser: prefer to map first and then threaded `.deploy` to separate concerns a bit better

grosser: also use `Parallel.map` so exceptions are handled (we don't need to DB multithreading that Samson::Parallelizer provides)

dasch:
@grosser I don't have time to get Samson local development set up – will you be able to finish this once Bootstrapgate is finished?

dasch:
Thanks!

grosser:
```
before:
[20:12:47] Deploying Pod99 role worker
[20:12:52] Deploying Pod99 role console
[20:12:53] Deploying Pod98 role worker
[20:12:57] Deploying Pod98 role console

after:
[18:51:01] Deploying Pod99 role worker
[18:51:01] Deploying Pod99 role console
[18:51:01] Deploying Pod98 role worker
[18:51:01] Deploying Pod98 role console

```

dasch:
❤️❤️❤️

PR #2770, opened 2018-05-29
DEVEX-721 wait for release to be readable to avoid commit not found e…

grosser:
…rrors

@zendesk/samson @bquorning @benhass @dadah89 @anthonywoo @clairekrogers

### DIFF
--- a/app/models/release_service.rb
+++ b/app/models/release_service.rb
@@ -7,7 +7,8 @@ class ReleaseService
   def release(attrs = {})
     release = @project.releases.create(attrs)
     if release.persisted?
-      push_tag_to_git_repository(release)
+      push_tag_to_git_repository(release.version, release.commit)
+      ensure_tag_in_git_repository(release.version)
       start_deploys(release)
     end
     release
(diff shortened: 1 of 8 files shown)


app/models/release_service.rb
```diff
@@ -21,8 +22,14 @@ def can_release?
...
   private
 
-  def push_tag_to_git_repository(release)
-    GITHUB.create_release(@project.repository_path, release.version, target_commitish: release.commit)
+  def push_tag_to_git_repository(version, commit)
+    GITHUB.create_release(@project.repository_path, version, target_commitish: commit)
+  end
+
+  def ensure_tag_in_git_repository(tag)
+    Samson::Retry.with_retries [Octokit::NotFound], 3, sleep: 1 do
```
craig-day: is 3 enough? or is 1 second enough? Maybe make ENV configurable for easier adjustment?

lib/samson/retry.rb
```diff
@@ -5,12 +5,16 @@ def self.retry_when_not_unique(&block)
...
 
-    def self.with_retries(errors, count, options = nil)
-      check = options && options[:if]
+    def self.with_retries(errors, count, only_if: nil, sleep: nil)
       yield
     rescue *errors
       count -= 1
-      count >= 0 && (!check || check.call($!)) ? retry : raise
+      if count >= 0 && (!only_if || only_if.call($!))
+        sleep(sleep) if sleep
```
craig-day: super confusing variable names here, maybe name the argument `sleep_duration` or `wait_time` or something?

app/models/release_service.rb
grosser: I'd bet that's enough ... we currently do the new deploy immediately so the slightest wait would fix it

lib/samson/retry.rb
grosser: looks so neat though ... changed 😿

PR #2771, opened 2018-05-29
do not auto-add pod disruption budgets when replica size is unknown

grosser:
@jonmoter @adammw @bquorning 

/cc @zendesk/skyfox


grosser:
not needed for now ...

PR #2773, opened 2018-05-31
DEVEX-721 fix missing tag again by actually using git instead of api

grosser:
previous PR https://github.com/zendesk/samson/pull/2770 did not work out :/

### DIFF
--- a/app/models/release_service.rb
+++ b/app/models/release_service.rb
@@ -27,8 +27,8 @@ class ReleaseService
   end
 
   def ensure_tag_in_git_repository(tag)
-    Samson::Retry.with_retries [Octokit::NotFound], 3, wait_time: 1 do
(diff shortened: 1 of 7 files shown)


app/models/release_service.rb
```diff
@@ -27,8 +27,11 @@ def push_tag_to_git_repository(version, commit)
   end
 
   def ensure_tag_in_git_repository(tag)
-    Samson::Retry.with_retries [Octokit::NotFound], 3, wait_time: 1 do
-      GITHUB.release_for_tag(@project.repository_path, tag)
+    tries = 4
```
ragurney: Might be nice to either refactor `with_retries` to retry if a passed in proc evaluates to nil/false or if rescued an expected error, or make this a new method in `Samson::Retry`

PR #2774, opened 2018-05-31
do not fail to add budget when first resource is a service

grosser:
@libo @ragurney 

https://rollbar-us.zendesk.com/Zendesk/Samson/items/259/occurrences/805784192/

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -86,10 +86,15 @@ module Kubernetes
     end
 
     def add_pod_disruption_budget
-      min_available = nil
-      raw_template.each do |t|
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -86,10 +86,15 @@ def store_resource_template
     end
 
     def add_pod_disruption_budget
-      min_available = nil
-      raw_template.each do |t|
-        min_available ||= t.dig(:metadata, :annotations, :"samson/minAvailable")
-      end
+      min_available_key = [:metadata, :annotations, :"samson/minAvailable"]
+      match_labels_key = [:spec, :selector, :matchLabels]
```
libo: shouldn't these be plural as in `min_available_keys`

grosser: yeah ... renamed to _path

PR #2784, opened 2018-06-04
update the mirror between tries or the ref will never show up

grosser:
@ragurney


app/models/release_service.rb
```diff
@@ -29,6 +29,7 @@ def push_tag_to_git_repository(version, commit)
   def ensure_tag_in_git_repository(tag)
     tries = Integer(ENV["RELEASE_TAG_IN_REPO_RETRIES"] || 4)
     Samson::Retry.until_result tries: tries, wait_time: 1, error: "Unable to find ref" do
```
ragurney: Would we want to increase the wait time now that we are doing an update each time?

grosser:
meh ... does not really matter, the updates are cheapish and if they take
long then that increases the wall time too :D

PR #2788, opened 2018-06-05
override project labels in budgets too

grosser:
currently resources that use project-overrides get budgets created that do not match their labels since the "multi-project" setting is in an annotation

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -126,7 +126,8 @@ module Kubernetes
         metadata: {
           name: kubernetes_role.resource_name,
           namespace: resource.dig(:metadata, :namespace),
-          labels: resource.dig_fetch(:metadata, :labels).dup
+          labels: resource.dig_fetch(:metadata, :labels).dup,
+          annotations: (resource.dig(:metadata, :annotations) || {}).dup
         },
         spec: {
           minAvailable: target,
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -126,7 +126,8 @@ def add_pod_disruption_budget
         metadata: {
           name: kubernetes_role.resource_name,
           namespace: resource.dig(:metadata, :namespace),
-          labels: resource.dig_fetch(:metadata, :labels).dup
+          labels: resource.dig_fetch(:metadata, :labels).dup,
+          annotations: (resource.dig(:metadata, :annotations) || {}).dup
```
ragurney: Could move the `dup` inside parens to `resource.dig(:metadata, :annotations).dup`

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -92,15 +92,13 @@ def set_resource_blue_green
...
-      project_label = project.permalink
-      template.dig_set([:metadata, :labels, :project], project_label)
-
-      kind = template.fetch(:kind)
-      if kind == "Service"
-        template.dig_set([:spec, :selector, :project], project_label)
-      elsif kind != "Pod"
-        template.dig_set([:spec, :selector, :matchLabels, :project], project_label)
-        template.dig_set([:spec, :template, :metadata, :labels, :project], project_label)
+      [
```
ragurney: ✨

grosser:
would blow up on `nil.dup` then :)

plugins/kubernetes/app/models/kubernetes/release_doc.rb
grosser: could do `.dig()&.dup || {}` ... but I"m not making another build for that :D

PR #2794, opened 2018-06-07
[DEVEX-731] Add links to Rollbar dashboard items

ragurney:
/cc @zendesk/samson

### References
 - Jira link: [DEVEX-731](https://zendesk.atlassian.net/browse/DEVEX-731)

### DIFF
--- a/plugins/rollbar_dashboards/app/helpers/rollbar_dashboards/dashboards_helper.rb
+++ b/plugins/rollbar_dashboards/app/helpers/rollbar_dashboards/dashboards_helper.rb
@@ -20,5 +20,17 @@ module RollbarDashboards
         }
       )
     end
+
+    def item_link(item_title, item_id, dashboard_setting)
+      if account_and_project_name = dashboard_setting.account_and_project_name
+        # Removes path and handles https://api.rollbar.com cases
+        domain = URI.join(dashboard_setting.base_url, '/').to_s.sub('://api.', '://')
+
+        item_url = "#{domain}#{account_and_project_name}/items/#{item_id}"
+        link_to item_title, item_url
+      else
+        item_title
+      end
+    end
   end
 end
--- a/plugins/rollbar_dashboards/app/views/samson_rollbar_dashboards/_fields.html.erb
+++ b/plugins/rollbar_dashboards/app/views/samson_rollbar_dashboards/_fields.html.erb
@@ -6,9 +6,10 @@
 
   <%
      description = <<~HTML.html_safe
-      <b>Base URL:</b> https://api.rollbar.com/api/1 unless you're using a self-hosted instance of Rollbar.<br>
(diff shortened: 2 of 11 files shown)


plugins/rollbar_dashboards/app/helpers/rollbar_dashboards/dashboards_helper.rb
```diff
@@ -20,5 +20,16 @@ def rollbar_dashboard_container(item_path, settings)
         }
       )
     end
+
+    def item_link(item_title, item_id, dashboard_setting)
+      account_name, project_name = dashboard_setting.account_and_project_name&.split('/')
```
grosser: can add `, 2` to the split just in case

plugins/rollbar_dashboards/app/helpers/rollbar_dashboards/dashboards_helper.rb
```diff
@@ -20,5 +20,16 @@ def rollbar_dashboard_container(item_path, settings)
...
+      return item_title if account_name.nil? || project_name.nil?
+
+      # Removes path and handles https://api.rollbar.com cases
+      domain = URI.join(dashboard_setting.base_url, '/').to_s.sub('api.', '')
```
grosser: might be safer to replace `://api.`

plugins/rollbar_dashboards/app/helpers/rollbar_dashboards/dashboards_helper.rb
```diff
@@ -20,5 +20,16 @@ def rollbar_dashboard_container(item_path, settings)
...
+
+      item_url = "#{domain}#{account_name}/#{project_name}/items/#{item_id}"
```
grosser: why split them if you just end up joining them again ? :D

plugins/rollbar_dashboards/app/views/samson_rollbar_dashboards/_fields.html.erb
```diff
@@ -19,5 +20,11 @@
     <div class="col-lg-3">
       <%= setting.text_field :read_token, class: "form-control", placeholder: "Project Read Token" %>
     </div>
+    <div class="col-lg-3">
+      <%= setting.text_field :account_and_project_name,
+            class: "form-control",
+            placeholder: "Account Name/Project Name"
```
grosser: kinda looks like "put this or that here" ... maybe `{Account Name}/{Project Name}` to be more explicit

plugins/rollbar_dashboards/app/helpers/rollbar_dashboards/dashboards_helper.rb
ragurney: You mean account_name and project_name? Splitting them allows me to somewhat verify that the passed in `account_and_project_name` is valid -- that a '/' exists in the input and both were passed in.

grosser:
generating a wrong url would let user now why it's not working instead of
being unlinked ... and make it simpler :)

PR #2795, opened 2018-06-07
DEVEX-741 cache secrets per cluster so we do not have to fetch them for every r…

grosser:
…esoure of every release doc of every release

@ragurney @dasch


dasch:
:heart:

PR #2798, opened 2018-06-09
Fix a race-condition with streams being requested after the job is finished

adammw:
Occasionally you can get a job finishing before the request to the stream has been made, at which point it infinitely retries (especially if the error is within the first step of cloning the repo). This change enables the stream to update the page and log the final output. Also fixes the tests which weren't asserting correctly.

/cc @zendesk/samson

### Risks
- Level: Low - affects job output streaming only

### DIFF
--- a/app/controllers/streams_controller.rb
+++ b/app/controllers/streams_controller.rb
@@ -16,14 +16,17 @@ class StreamsController < ApplicationController
     @job = Job.find(params[:id])
     @execution = JobQueue.find_by_id(@job.id)
 
-    return response.stream.close unless @job.active? && @execution
+    event_streamer = EventStreamer.new(response.stream, &method(:event_handler))
 
-    @execution.viewers.push(current_user)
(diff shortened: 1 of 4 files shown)


app/controllers/streams_controller.rb
```diff
@@ -16,14 +16,17 @@ def show
...
+      @execution.viewers.push(current_user)
 
-    ActiveRecord::Base.clear_active_connections!
+      ActiveRecord::Base.clear_active_connections!
 
-    EventStreamer.new(response.stream, &method(:event_handler)).
-      start(@execution.output)
+      event_streamer.start(@execution.output)
+    else
+      event_streamer.start(@job.output)
```
grosser: alternatively could send a `reload the page` event and the js handle it

adammw:
bump @grosser, is this okay to merge? I don't have merge access in this repo.

PR #2801, opened 2018-06-11
Nullify Vault server dependency after destroy

sathishavm:
Removes the Vault server foreign_key after destroying.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/DEVEX-691

### Risks
- Level: Low

### DIFF
--- a/lib/samson/secrets/vault_server.rb
+++ b/lib/samson/secrets/vault_server.rb
@@ -36,7 +36,7 @@ module Samson
         read_timeout: 2
       }.freeze
 
-      has_many :deploy_groups
+      has_many :deploy_groups, dependent: :nullify
 
       attribute :token
       attr_encrypted :token
--- a/test/lib/samson/secrets/vault_server_test.rb
+++ b/test/lib/samson/secrets/vault_server_test.rb
@@ -50,10 +50,10 @@ describe Samson::Secrets::VaultServer do
       ) { refute_valid server }
     end
 
-    it "nullifies the association on destroy" do
+    it "removes valut servers from associated deploy group" do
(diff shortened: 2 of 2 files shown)


test/lib/samson/secrets/vault_server_test.rb
```diff
@@ -49,6 +49,15 @@
         to_raise: Vault::HTTPError.new("address", stub(code: '200'))
       ) { refute_valid server }
     end
+
+    it "nullifies the association on destroy" do
```
grosser: kinda vague and focussed on the implementation

`it "removes unfindable servers from associated deploy group"`

test/lib/samson/secrets/vault_server_test.rb
```diff
@@ -49,6 +49,15 @@
...
+      deploy_group = deploy_groups(:pod100)
+      server.deploy_groups = [deploy_group]
+      server.save
```
grosser: `save!` so we don't get false-positives if the `save` silently failed

lib/samson/secrets/vault_server.rb
```diff
@@ -36,7 +36,7 @@ class VaultServer < ActiveRecord::Base
         read_timeout: 2
       }.freeze
 
-      has_many :deploy_groups
+      has_many :deploy_groups, dependent: :nullify
```
grosser: this is the default behavior too, so maybe that test already passes before changing the code ?

sathishavm: Nope, this is not the default behavior, I Just tested, It's needed.

PR #2802, opened 2018-06-11
Deploy group dependency fixes

sathishavm:
Fixes for Deploy group dependency.
1) Destroy deploy_group_roles
2) override association for release_doc

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/DEVEX-668

### Risks
- Level: Low

### DIFF
--- a/app/models/deploy_group.rb
+++ b/app/models/deploy_group.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 class DeployGroup < ActiveRecord::Base
-  has_soft_deletion default_scope: true
+  has_soft_deletion default_scope: true unless self < SoftDeletion::Core
   audited
 
   include Permalinkable
--- a/plugins/kubernetes/app/decorators/deploy_group_decorator.rb
+++ b/plugins/kubernetes/app/decorators/deploy_group_decorator.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 DeployGroup.class_eval do
+  has_soft_deletion default_scope: true
   has_one(
     :cluster_deploy_group,
     class_name: 'Kubernetes::ClusterDeployGroup',
@@ -7,6 +8,7 @@ DeployGroup.class_eval do
     inverse_of: :deploy_group
   )
   has_one :kubernetes_cluster, class_name: 'Kubernetes::Cluster', through: :cluster_deploy_group, source: :cluster
+  has_many :kubernetes_deploy_group_roles, class_name: 'Kubernetes::DeployGroupRole', dependent: :destroy
 
   accepts_nested_attributes_for(
     :cluster_deploy_group,
@@ -15,7 +17,13 @@ DeployGroup.class_eval do
     reject_if: lambda { |h| h[:kubernetes_cluster_id].blank? }
   )
 
+  after_soft_delete :delete_kubernetes_deploy_group_roles
+
   def kubernetes_namespace
     cluster_deploy_group.try(:namespace)
+  end
+
+  def delete_kubernetes_deploy_group_roles
+    kubernetes_deploy_group_roles.destroy_all
   end
 end
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -184,8 +184,8 @@ module Kubernetes
         if uid = resource.uid
           selector << "involvedObject.uid=#{uid}"
         end
-
-        events = doc.deploy_group.kubernetes_cluster.client.get_events(
+        deploy_group = DeployGroup.with_deleted{ doc.deploy_group }
+        events = deploy_group.kubernetes_cluster.client.get_events(
           namespace: resource.namespace,
           field_selector: selector.join(',')
         ).fetch(:items)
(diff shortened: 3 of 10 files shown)


app/models/deploy_group.rb
```diff
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 class DeployGroup < ActiveRecord::Base
-  has_soft_deletion default_scope: true
```
grosser: needs to stay here too in case a user does not have all plugins, see https://github.com/zendesk/samson/blob/master/app/models/stage.rb#L5

plugins/kubernetes/app/decorators/deploy_group_decorator.rb
```diff
@@ -1,12 +1,14 @@
...
 DeployGroup.class_eval do
+  has_soft_deletion default_scope: true
   has_one(
     :cluster_deploy_group,
     class_name: 'Kubernetes::ClusterDeployGroup',
     foreign_key: :deploy_group_id,
     inverse_of: :deploy_group
   )
   has_one :kubernetes_cluster, class_name: 'Kubernetes::Cluster', through: :cluster_deploy_group, source: :cluster
+  has_many :kubernetes_deploy_group_roles, class_name: 'Kubernetes::DeployGroupRole'
```
grosser: use `dependent: :destroy` to make them auto-cleanup

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -64,6 +64,10 @@ def blue_green_color
       kubernetes_release.blue_green_color if kubernetes_role.blue_green?
     end
 
+    def deploy_group
```
grosser: kinda breaks the expectation of soft_deletion and might break eager loading too ...
prefer to either support nils or delete dependencies when nils are not supported

plugins/kubernetes/app/decorators/deploy_group_decorator.rb
sathishavm: Do we need this? this is not triggering anywhere. You mean that we may run destroy manually?

plugins/kubernetes/app/models/kubernetes/release_doc.rb
sathishavm: This is not breaking the eager loading, I just checked.
This also triggering the expectation of soft_deletion, I just checked by deleting the 'deleted_at' from table.
Here, I only try to support nils, There are lots of places we need to fix to support nils, I think this is the easiest and safest way to support nils.

plugins/kubernetes/app/decorators/deploy_group_decorator.rb
grosser: It's needed to make cascading deletes work https://github.com/grosser/soft_deletion/blob/master/lib/soft_deletion/core.rb#L17 also it reads nicer/looks safer

plugins/kubernetes/app/models/kubernetes/release_doc.rb
grosser: I'd prefer a pattern that can be used for everything so it's no guesswork on how associations will behave.
Adding with_deleted for this 1 association might mean less changes are necessary right now, but it makes things inconsistent / adds tech-debt. 

Maybe as minimal compromise make all deploy-group belongs-to use with_deleted ?
(search for with_deleted to remove existing with_deleted calls that will become obsolete and add a cleanliness test that checks for belongs_to :deploy_group + def deploy_group)

Changing all belongs_to sounds kinda hacky/dangerous.

Using with_deleted in views as needed is hacky/1-off too, but at least consistent.

sathishavm: The following are the places needs the change:
1) /plugins/kubernetes/app/views/samson_kubernetes/_deploy_tab_body.html.erb
2) /plugins/kubernetes/app/views/kubernetes/releases/show.html.erb
3) /plugins/kubernetes/app/models/kubernetes/template_filler.rb
4) /plugins/kubernetes/app/models/kubernetes/release.rb
5) /plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
 May be more..

sathishavm:
@grosser Need review!!

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -10,6 +10,7 @@ class TemplateFiller
 
     def initialize(release_doc, template, index:)
       @doc = release_doc
+      @deploy_group = DeployGroup.with_deleted { @doc.deploy_group }
```
grosser: we should not do this, a deleted deploy group does not need new deployments ... only view related code needs to be addressed

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -184,8 +184,8 @@ def print_resource_events(doc)
         if uid = resource.uid
           selector << "involvedObject.uid=#{uid}"
         end
-
-        events = doc.deploy_group.kubernetes_cluster.client.get_events(
+        deploy_group = DeployGroup.with_deleted { doc.deploy_group }
```
grosser: same here, we are not deploying to deleted groups

PR #2803, opened 2018-06-12
[DEVEX-137] Implement expected duration for a stage's deploys

ragurney:
Adds an `average_deploy_time` field to `Stage`, backfilling any existing stage averages. This allows tracking of overall deployment time for a particular stage, and estimates on a new deploy's duration.

/cc @zendesk/devex 

Status of active deploy:
screen shot 2018-06-11 at 4 30 27 pm

Status of active deploy - no previous average: 
screen shot 2018-06-11 at 4 30 55 pm

Deploying page:
screen shot 2018-06-11 at 5 11 55 pm

### References
 - Jira link: [DEVEX-137](https://zendesk.atlassian.net/browse/DEVEX-137)
 - [Formula for running average calculation](https://math.stackexchange.com/questions/22348/how-to-add-and-subtract-values-from-an-average)

### DIFF
--- a/app/helpers/status_helper.rb
+++ b/app/helpers/status_helper.rb
@@ -24,26 +24,32 @@ module StatusHelper
     "label-#{LABEL_STATUS_MAPPING.fetch(key, 'info')}"
   end
 
-  def status_panel(deploy)
-    content = h deploy.summary
-    if deploy.active?
+  def status_panel(deployable)
+    is_deploy = deployable.is_a?(Deploy)
+
+    content = h deployable.summary
+    if deployable.active?
       content << content_tag(:br)
       content << "Started "
-      content << relative_time(deploy.start_time)
-      job = (deploy.respond_to?(:job) ? deploy.job : deploy)
+      content << relative_time(deployable.start_time)
+      job = (is_deploy ? deployable.job : deployable)
       content << " [Process ID: #{job.pid}]"
+      if is_deploy
+        content << content_tag(:br)
+        content << "ETA: #{duration_text deployable.stage.average_deploy_time}"
+      end
     end
 
-    if deploy.finished?
+    if deployable.finished?
       content << ". Started "
-      content << render_time(deploy.start_time, current_user.time_format)
-      content << ", took #{duration_text deploy.duration}."
+      content << render_time(deployable.start_time, current_user.time_format)
+      content << ", took #{duration_text deployable.duration}."
     end
 
-    content_tag :div, content, class: "alert #{status_alert(deploy.status)}"
+    content_tag :div, content, class: "alert #{status_alert(deployable.status)}"
   end
 
   def duration_text(duration)
-    Time.at(duration).utc.strftime('%H:%M:%S')
+    duration ? Time.at(duration).utc.strftime('%H:%M:%S') : ''
   end
 end
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -38,6 +38,7 @@ class DeployService
     end
     job_execution.on_finish do
       send_after_notifications(deploy)
+      update_average_deploy_time(deploy)
     end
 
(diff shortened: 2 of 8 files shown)


app/helpers/status_helper.rb
```diff
@@ -24,14 +24,21 @@ def status_label(key)
     "label-#{LABEL_STATUS_MAPPING.fetch(key, 'info')}"
   end
 
+  # accepts either a job or a deploy
   def status_panel(deploy)
```
grosser: rename to `deployable` ?

app/helpers/status_helper.rb
```diff
@@ -24,14 +24,21 @@ def status_label(key)
...
     if deploy.active?
       content << content_tag(:br)
       content << "Started "
       content << relative_time(deploy.start_time)
-      job = (deploy.respond_to?(:job) ? deploy.job : deploy)
+      job = (is_deploy ? deploy.job : deploy)
       content << " [Process ID: #{job.pid}]"
+      if is_deploy
+        content << content_tag(:br)
+        content << "Estimated completion time: #{duration_text deploy.stage.average_deploy_time}"
```
grosser: maybe just "ETA: "

grosser: maybe cleaner to not add duration if there is know known duration ?

app/models/deploy_service.rb
```diff
@@ -47,6 +48,17 @@ def confirm_deploy(deploy)
...
+  def update_average_deploy_time(deploy)
+    stage = deploy.stage
+
+    old_average = stage.average_deploy_time || 0.00
+    number_of_deploys = stage.deploys.size
+
+    new_average = (old_average + ((deploy.duration - old_average) / number_of_deploys))
+
+    stage.update_column(:average_deploy_time, new_average)
+  end
```
grosser: kinda sketchy math ... can you explain why this is needed or switch to something simple like average from last 10 ?

db/migrate/20180611172637_add_average_deploy_time_to_stages.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class AddAverageDeployTimeToStages < ActiveRecord::Migration[5.2]
+  def up
+    ActiveRecord::Base.transaction do
+      add_column :stages, :average_deploy_time, :float
```
grosser: maybe use an integer that means seconds then we don't need float math ?

grosser: doubt mysql can add a column in an transaction, so prefer to keep that outside

db/migrate/20180611172637_add_average_deploy_time_to_stages.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class AddAverageDeployTimeToStages < ActiveRecord::Migration[5.2]
+  def up
```
grosser: define all used models in the migration to future-proof it

app/models/deploy_service.rb
ragurney: It gives a more accurate picture of the overall average deploy time of that particular stage. If there were 5 failed deploys or more for a particular stage because of some bad config or something, then the average would be greatly skewed. That might be pretty confusing to some people when their stage ETA changes so drastically due to a couple of failed or hung deploys, even though they might have hundreds of past deploys with normal run times.

db/migrate/20180611172637_add_average_deploy_time_to_stages.rb
ragurney: We need to use float math if we're keeping a running average, otherwise the formula above would not work

grosser:
and you ran migrations ?

danihodovic:
> and you ran migrations ?

Seems like I didn't. Thanks @grosser

grosser:
phew, easy fix 🎉

PR #2805, opened 2018-06-12
[DEVEX-137] Fix migration issue with postgres

ragurney:
Fixes migration issue with postgres. 

### References
 - Jira link: [DEVEX-137](https://zendesk.atlassian.net/browse/DEVEX-137)


db/migrate/20180611172637_add_average_deploy_time_to_stages.rb
```diff
@@ -1,19 +1,16 @@
 # frozen_string_literal: true
 
+class Deploy < ActiveRecord::Base
+  self.table_name = :deploys
```
grosser: don't need this

PR #2809, opened 2018-06-16
add mass select for kubernetes resources

grosser:
screen shot 2018-06-15 at 5 15 24 pm

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/views/kubernetes/deploy_group_roles/edit_many.html.erb
+++ b/plugins/kubernetes/app/views/kubernetes/deploy_group_roles/edit_many.html.erb
@@ -10,6 +10,7 @@
         [:limits_cpu, 'Limits CPU Cores'],
         [:limits_memory, 'Limits Memory in MB']
     ] %>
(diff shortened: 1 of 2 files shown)


adammw:
can you make it a different colour so it's obvious?

grosser:
color looks kinda gimmicky ... how about no role so it stands out:

screen shot 2018-06-15 at 5 29 43 pm

plugins/kubernetes/app/views/kubernetes/deploy_group_roles/edit_many.html.erb
```diff
@@ -32,17 +33,20 @@
...
           <%= fields_for "kubernetes_deploy_group_roles[#{deploy_group_role.id}]", deploy_group_role do |field| %>
             <% rows.each do |attribute, _, type| %>
-              <td>
-                <% if type == :boolean %>
-                  <%= field.check_box attribute %>
-                <% else %>
-                  <%= field.text_field attribute, size: 10 %>
-                <% end %>
-              </td>
+              <td><%= row_builder.(field, type, attribute) %></td>
```
adammw: why do you need the dot? is that the same as .call?

grosser:
yes

PR #2811, opened 2018-06-18
show full content of textareas

grosser:
/cc @zendesk/samson @adammw

### DIFF
--- a/app/assets/javascripts/expanded_textarea.js
+++ b/app/assets/javascripts/expanded_textarea.js
@@ -1,0 +1,7 @@
+// make textareas expand to their contents size
+// adjusting the math (+3) to make 1-line have regular size
+$(function(){
+  $('textarea').each(function(i, element){
+    element.style.height = element.scrollHeight + 3 +"px";
+  });
+});


grosser:
especially env var view is annoying if you have to expand every deploy group ...

screen shot 2018-06-18 at 1 09 13 pm

adammw:
Reading up on alternatives: "Another trick here is to not use a <textarea> at all, but <div contenteditable>. That will grow automatically with no fancy JavaScript help at all - except that it's not a real form element anymore so you'll need JavaScript to extract/send the value."

Think this is preferable to adding in a library to do it, or making JS do the form.

app/assets/javascripts/expanded_textarea.js
```diff
@@ -0,0 +1,7 @@
+// make textareas expand to their contents size
+// adjusting the math (+3) to make 1-line have regular size
+$(function(){
+  $('textarea').each(function(i, element){
+    element.style.height = element.scrollHeight + 3 +"px";
```
adammw: might be worth using jquery here too..

grosser: doing it for the speed :D 
... if it somehow fails we can change it over to jquery ...

PR #2815, opened 2018-06-19
PAAS-1530 unify 1-off client methods

grosser:
@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -233,7 +233,7 @@ module Kubernetes
       private
 
       def client
-        @deploy_group.kubernetes_cluster.autoscaling_client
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -233,7 +233,7 @@ class HorizontalPodAutoscaler < Base
       private
 
       def client
-        @deploy_group.kubernetes_cluster.autoscaling_client
+        @deploy_group.kubernetes_cluster.client('autoscaling/v1')
```
ragurney: Maybe worth putting these types (i.e. 'autoscaling/v1')  in a constant hash in case they change? Then use symbols for all types?

grosser: apis are backwards compatible, so they should not change ... someone might need new apis, but that does not mean we should change everyone who used v1 before

PR #2817, opened 2018-06-20
[BRE-749] Improve regexp to capture section content

ragurney:
* Fixes regex used to get risk content from PRs to make it more robust and enable it to handle edge cases.
* Now shows all PRs under 'Risks' and alerts if there was no risk section found (either it was not defined or not able to be parsed)

screen shot 2018-06-20 at 1 21 32 pm
screen shot 2018-06-20 at 1 21 42 pm

/cc @zendesk/samson

### References
 - Jira link: [BRE-749](https://zendesk.atlassian.net/browse/BRE-749)

### DIFF
--- a/app/assets/javascripts/tabs.js
+++ b/app/assets/javascripts/tabs.js
@@ -8,4 +8,9 @@ $(function () {
   $('.nav-tabs a[href^="#"]').click(function(){
     window.location.hash = this.hash;
   });
+
+  // Initialize popover tooltips in tab when tab is shown
+  $('a[data-toggle="tab"]').on('shown.bs.tab', function() {
+    $('.tab-content i[data-toggle="popover"]').popover();
+  });
 });
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -6,12 +6,6 @@ class Changeset::PullRequest
 
   WEBHOOK_FILTER = /(^|\s)\[samson review\]($|\s)/i
 
-  # Matches a markdown section heading named "Risks".
-  RISKS_SECTION = /^\s*#*\s*Risks?\s*#*\s*\n(?:\s*[-=]*\s*\n)?/i
-
-  # Matches a markdown section heading
-  SECTION_HEADING = /^\s*#*\s*\w+.*\n(?:\s*[-=]*\s*\n)?/i
-
   # Matches URLs to JIRA issues.
   JIRA_ISSUE_URL = %r[https?:\/\/[\da-z\.\-]+\.[a-z\.]{2,6}\/browse\/#{CODE_ONLY}(?=#{PUNCT}|$)]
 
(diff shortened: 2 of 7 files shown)


app/views/changeset/_risks.html.erb
```diff
@@ -1,10 +1,20 @@
...
     <h5><strong>#<%= pr.number %></strong> <%= link_to pr.title, pr.url %> <%= github_users(pr.users) %></h5>
     <%= markdown pr.risks %>
+  <% elsif pr.missing_risks? %>
+    <div>
+      <% help_icon = additional_info(
+            "Missing 'Risks' section or failed to parse risks",
+            class: 'glyphicon glyphicon-alert deployment-alert'
+          )
+      %>
+      <h5><strong>#<%= pr.number %></strong> <%= link_to pr.title, pr.url %> <%= github_users(pr.users) %><%= help_icon %></h5>
```
grosser: that'll be super spammy :(

grosser: ... but I guess if you are already on the risks tag then maybe that's ok ...

app/models/changeset/pull_request.rb
```diff
@@ -129,9 +129,17 @@ def message
 
   private
 
+  def section_content(section_title, text)
+    desired_header_regexp = "^(?:\\s*#+\\s*#{section_title}.*|\\s*#{section_title}.*\\n\\s*(?:-{2,}|={2,}))\\n"
+    content_regexp = '([\W\w]*?)'
```
grosser: how is that different from `.*?` ?

test/controllers/deploys_controller_test.rb
```diff
@@ -417,15 +421,59 @@ def self.with_and_without_project(&block)
...
+
+        confirm
+
+        assert_response :success
+
+        assert_select 'h5' do
+          assert_select 'strong', text: '#1'
+          assert_select 'a[href=?]', 'https://github.com/foo/bar/pull/1', text: 'Cool Stuff!'
+        end
+        assert_select 'p', text: 'The sky is falling!'
```
grosser: tests that break when the markup changes ... not a fan ... assert that it includes `Cool Stuff` and the url ?

app/assets/javascripts/tabs.js
```diff
@@ -8,4 +8,8 @@ $(function () {
   $('.nav-tabs a[href^="#"]').click(function(){
     window.location.hash = this.hash;
   });
+
+  $('a[data-toggle="tab"]').on('shown.bs.tab', function() {
```
grosser: add a comment to explain what this does

app/models/changeset/pull_request.rb
ragurney: `.` matches everything but a newline, while `\W` will match the newline.

grosser:
better document that :D

PR #2818, opened 2018-06-20
improve desired pod checking

grosser:
- say which deploy group failed
 - try 1 more time
 - dry up lookup logic

@zendesk/compute


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -303,21 +303,29 @@ def deploy
...
         @desired_pod_count ||= begin
-          desired = resource.dig_fetch :status, :desiredNumberScheduled
-          return desired unless desired.zero?
+          desired = 0
 
-          # in bad state or does not yet know how many it needs
-          loop_sleep
-          expire_cache
+          3.times do |i|
+            if i != 0
```
adammw: nit: i > 0 would be a little clearer I think

PR #2824, opened 2018-06-26
[BRE-443] Add Project lock functionality and refactor code

ragurney:
screen shot 2018-06-25 at 4 52 07 pm
screen shot 2018-06-25 at 5 31 19 pm

/cc @zendesk/samson @zendesk/bre 

### References
 - Jira link: [BRE-443](https://zendesk.atlassian.net/browse/BRE-443)

### DIFF
--- a/app/models/access_control.rb
+++ b/app/models/access_control.rb
@@ -18,12 +18,11 @@ class AccessControl
         case action
         when :read then true
         when :write
-          return user.admin? if scope.nil? # global locks
-          return false unless scope.is_a?(Stage) # stage locks
-          if ENV['PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN'] && scope.production
-            user.admin_for?(scope.project)
-          else
-            user.deployer_for?(scope.project)
+          case scope.class.to_s
+          when 'NilClass' then user.admin? # global locks
+          when 'Project' then user.admin_for?(scope) # project locks
+          when 'Stage' then can_lock_stage?(user, scope) # stage locks
+          else false
           end
         else raise ArgumentError, "Unsupported action #{action}"
         end
@@ -64,6 +63,14 @@ class AccessControl
 
     def can_deploy_anything?(user)
       user.deployer? || user.user_project_roles.where('role_id >= ?', Role::DEPLOYER.id).exists?
+    end
+
+    def can_lock_stage?(user, stage)
+      if ENV['PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN'] && stage.production?
+        user.admin_for?(stage.project)
+      else
+        user.deployer_for?(stage.project)
+      end
     end
   end
 end
--- a/app/models/concerns/lockable.rb
+++ b/app/models/concerns/lockable.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module Lockable
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :lock, as: :resource
+  end
+
+  def locked_by?(lock)
+    lock.global? || lock.resource_equal?(self)
+  end
+end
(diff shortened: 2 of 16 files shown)


app/models/access_control.rb
```diff
@@ -18,12 +18,12 @@ def can?(user, action, resource_namespace, scope = nil)
...
-          return false unless scope.is_a?(Stage) # stage locks
-          if ENV['PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN'] && scope.production
-            user.admin_for?(scope.project)
+          case scope.class.to_s
+          when 'NilClass' then user.admin? # global locks
+          when 'Project' then user.admin_for?(scope) # project locks
+          when 'Stage' then can_lock_stage?(user, scope) # stage locks
           else
-            user.deployer_for?(scope.project)
+            false
```
grosser: for more symmetry could even do `else false`

app/models/access_control.rb
```diff
@@ -65,5 +65,13 @@ def can?(user, action, resource_namespace, scope = nil)
     def can_deploy_anything?(user)
       user.deployer? || user.user_project_roles.where('role_id >= ?', Role::DEPLOYER.id).exists?
     end
+
+    def can_lock_stage?(user, stage)
+      if ENV['PRODUCTION_STAGE_LOCK_REQUIRES_ADMIN'] && stage.production
```
grosser: I think `.production` != `.production?` (column vs logic)

grosser: (pre-existing bug I guess, can still solve)

app/models/stage.rb
```diff
@@ -204,6 +205,10 @@ def url
     Rails.application.routes.url_helpers.project_stage_url(project, self)
   end
 
+  def affected_by_lock?(lock)
```
grosser: you sneaky :D

app/models/stage.rb
```diff
@@ -254,4 +259,12 @@ def append_new_command
...
     stage_commands.build(command: new_command, position: previous + 1)
   end
+
+  def environment_lock?(lock)
+    lock.resource_type == "Environment" && environments.any? { |e| lock.resource_equal(e) }
+  end
+
+  def project_lock?(lock)
+    lock.resource_type == "Project" && lock.resource_equal(project)
+  end
```
grosser: this feels kinda wrong ... locking logic is now all over the place :(

app/models/concerns/lockable.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Lockable
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :lock, as: :resource
+  end
+
+  def affected_by_lock?(lock)
```
grosser: `def locked_by?(lock)`

app/models/lock.rb
```diff
@@ -23,8 +23,36 @@ class Lock < ActiveRecord::Base
...
+    # normally there are very few locks, so we grab them all and filter down to avoid lookups
+    # sorted by priority(warning) and specificity(type)
+    def for_resource(resource)
+      matching_locks = all_cached.select { |l| resource.affected_by_lock?(l) }
+      matching_locks.sort_by! { |l| [l.warning? ? 1 : 0, RESOURCE_TYPES.index(l.resource_type)] }
+    end
+
+    def locked_for?(resource, user)
+      locks = for_resource(resource)
+      locks.any? { |l| !l.warning? && l.user != user }
```
grosser: compare the ids to avoid lookups ?

app/models/access_control.rb
ragurney: Good catch! Will fix

app/models/stage.rb
ragurney: Yeah... I was on the fence about this. It kind of smells bad, but as you say it allows for locks to be used in plugins and it moves model-specific logic to the models.

Issue #2820, opened 2018-06-22
Allow not using CPU limits on k8s apps


rata:
Hi!

We've just hit an issue with latency sensitive applications on kubernetes. We wanted to make sure you were aware too, just in case it helps you. We are using an old version of samson, so not sure this really applies now. But as far as I can see, it applies (will look in detail next week).

The thing is just setting CPU limits for pods. When this is done, as the default CFS period is big (100ms) in case of k8s, this can cause quite some big delays (100ms :)) on containers just to cap the CPU. This affects a lot p99 or p95 response times (in our case, at least).

There is a bug that was opened in kubernetes here[1], that I think it wasn't quite clear what it implied for latency sensitive apps at first. The documentation (as that bug says) is misleading too. Is quite long, but interesting :) 

Here is a screenshot of some charts (the deploy was just before 13hs) after removing the CPU limits on one of our apps. The mean and p95 queue time was reduced a lot and the response time by half, too. 

![image](https://user-images.githubusercontent.com/70861/41795382-e54bb260-7638-11e8-9786-ebe3cae2ad28.png)

As reported in that bug, others have seen some incredible performance improvements in p99/p95 too[2]

Some comments I'd highlight about that long issue are these:
https://github.com/kubernetes/kubernetes/issues/51135#issuecomment-388617204
https://github.com/kubernetes/kubernetes/issues/51135#issuecomment-387139466

There are several ways to work on this. Changing the CFS period, disabling CFS quota entirely or not using CPU limits. At least that are the paths I understand from the issue so far, although is quite recent. The way to change the CFS period has been merged a few days ago, so not really an option on a released version yet. And there is a strong case that the doc is misleading and probably CPU limits don't need to be set on most cases.

The schedule CFS and quota are explained here:
https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt

And here a question that explain just that, just in case: https://unix.stackexchange.com/questions/213971/what-is-difference-between-different-cpu-cfs-period-us-values

We are using v1556, so the code is quite different. At that version only limits is set (no requests), and we just changed one line to use requests instead of limit. I see the code changed a lot, but as far as I tried you can't set the limit to 0 or something to disable (kubernetes won't accept the yaml).

So, unless I'm missing something in samson code, I guess this is not possible to change (i.e. not set cpu limit).

I think it's worth having a way to not set CPU limits on the yaml sent, given the consequences it can have for latency sensitive apps. What do you think?

[...]

PR #2825, opened 2018-06-26
Allow not using CPU limits on k8s apps

grosser:
see https://github.com/zendesk/samson/issues/2820 for details
/cc @zendesk/samson @jonmoter @rata 

quick dump of how I think this might wok ... looks alright ?

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -215,6 +215,14 @@ ActiveSupport::TestCase.class_eval do
   ensure
     Thread.report_on_exception = old
   end
+
+  def stub_const(base, name, value)
+    old = base.const_get(name)
+    silence_warnings { base.const_set(name, value) }
+    yield
+  ensure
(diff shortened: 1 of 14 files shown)


grosser:
screen shot 2018-06-26 at 9 01 41 am
screen shot 2018-06-26 at 9 02 02 am

rata:
Awesome! But not sure mem limit might be desirable, at least in some cases. It does not cause such a penalty, just enforces the mem limit.  What do you think?

grosser:
updated!

rata:
@grosser LGTM! :)

test/test_helper.rb
```diff
@@ -215,6 +215,14 @@ def silence_thread_exceptions
   ensure
     Thread.report_on_exception = old
   end
+
+  def stub_const(base, name, value)
```
grosser: fyi @ragurney

PR #2830, opened 2018-06-28
Samson processes reported to statsd

sathishavm:
Job to report Samson processes to statsd

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-782

### Risks
- Level: MED

### DIFF
--- a/lib/samson/process_utils.rb
+++ b/lib/samson/process_utils.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module Samson
+  module ProcessUtils
+    ATTRIBUTES = ['pid', 'ppid', 'gid', 'pcpu', 'user', 'start', 'args'].freeze
+
+    class << self
+      def ps_list
+        whitelists = ENV.fetch('PROCESS_WHITELIST', '').split(',')
+        pipe = IO.popen("ps -eo #{ATTRIBUTES.join(',')}")
+        # Ignore known long-running processes so report is meaningful
+        filtered_processes = pipe.readlines[1..-1].reject do |lines|
+          whitelists.any? { |key| lines.include?(key) }
+        end
+        filtered_processes.map do |line|
+          Hash[ATTRIBUTES.zip line.lstrip.split(/\s+/, ATTRIBUTES.size)]
+        end
+      end
+
+      def report_to_statsd
+        ps_list.each do |process|
+          runtime = begin
+            Time.now.to_i - Time.parse(process.fetch('start')).to_i
+          rescue ArgumentError
+            0
+          end
+          tags = ATTRIBUTES.map { |attr| "#{attr}:#{process.fetch(attr)}" }
+          Samson.statsd.gauge("process.runtime", runtime, tags: tags)
+        end
+      end
+    end
+  end
+end
--- a/test/lib/samson/process_utils_test.rb
+++ b/test/lib/samson/process_utils_test.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require_relative '../../test_helper'
+
+SingleCov.covered!
+
+describe Samson::ProcessUtils do
+  describe '.ps_list' do
+    it 'returns a list of running processes' do
+      IO.popen("echo samson_testing; sleep 5")
+      Samson::ProcessUtils.ps_list.map { |ps| ps['args'] }.must_include "sh -c echo samson_testing; sleep 5\n"
+    end
+
+    it 'skips the completed processes' do
+      IO.popen("echo samson_completed")
+      Samson::ProcessUtils.ps_list.map { |ps| ps['args'] }.wont_include "sh -c echo samson_completed\n"
+    end
+  end
+
+  describe '.report_to_statsd' do
+    it 'report the processes to statsd' do
+      IO.popen("echo samson_statsd; sleep 5")
+      Samson::ProcessUtils.ps_list.each do |process|
+        tags = [
+          "pid:#{process['pid']}",
+          "ppid:#{process['ppid']}",
+          "gid:#{process['gid']}",
+          "pcpu:#{process['pcpu']}",
+          "user:#{process['user']}",
+          "start:#{process['start']}",
+          "args:#{process['args']}"
+        ]
+        Samson.statsd.stubs(:gauge).with('process.runtime', anything, tags: tags)
+      end
+      Samson.statsd.stubs(:gauge).with('process.runtime', anything, anything)
+      assert Samson::ProcessUtils.report_to_statsd
+    end
+
+    it 'report different value for timeout processes' do
+      Samson.statsd.stubs(:gauge).with('process.runtime', anything, anything)
+      Time.stubs(:parse).raises ArgumentError
+      assert Samson::ProcessUtils.report_to_statsd
+    end
+  end
+end
--- a/.env.example
+++ b/.env.example
@@ -59,10 +59,11 @@ GITHUB_TOKEN=
 # remove_expired_locks:10
 # renew_vault_token:86400
 # report_system_stats:60
+# report_process_stats:60
 # periodical_deploy:86400
 # cancel_stalled_builds:3600
 # Recommended setup:
-PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:60,periodical_deploy:86400,cancel_stalled_builds:3600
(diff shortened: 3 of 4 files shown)


lib/samson/process_utils.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Samson
+  module ProcessUtils
+    ATTRIBUTES = ['pid', 'ppid', 'gid', 'user', 'start', 'args'].freeze
+
+    class Process
```
grosser: smells like unnecessary abstraction ... lots of magic to set and then call it ... might as well do it inline ?

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,34 @@
...
+        pipe = IO.popen("ps -eo #{ATTRIBUTES.join(',')}")
+        pipe.readlines[1..-1].map do |line|
+          Process.new line.lstrip.split(/\s+/, 6)
+        end
+      end
+
+      def report_to_statsd
+        ps_list.each do |process|
+          timeout = Time.parse(process.start) < Time.now - Rails.application.config.samson.deploy_timeout
+          tags = ATTRIBUTES.map { |attr| "#{attr}:#{process.send(attr)}" }
```
grosser: args/pid/gpid etc are to diverse to be good tags ... datadog has a 1k tag limit afaik
sending all processes makes monitoring useless anyway since some of them will be samson itself
if we can somehow narrow this down to "processes started by samson" it could be actionable

grosser: let's discuss tomorrow

sathishavm:
@grosser Needs Review!!
Added CPU tag and WHITELISTS, for now, it's only puma for now.
While creating statsd monitor, We are going to exclude the tags "user" with ('root', 'dd-agent') and 
pcup with '0.0'

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Samson
+  module ProcessUtils
+    ATTRIBUTES = ['pid', 'ppid', 'gid', 'pcpu', 'user', 'start', 'args'].freeze
+    WHITELISTS = ['puma'].freeze
+
+    class << self
+      def ps_list
+        pipe = IO.popen("ps -eo #{ATTRIBUTES.join(',')} | grep -v #{WHITELISTS.join(',')}")
```
grosser: - joining a grep with comma won't work
 - prefer a whitelist from ENV since every user will need a different whitelist (also allows fast adjustment)

grosser: prefer filtering in ruby for more flexibility

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,25 @@
...
+        pipe.readlines[1..-1].map do |line|
+          Hash[ATTRIBUTES.zip line.lstrip.split(/\s+/, 7)]
```
grosser: 7 <- magic number ... use ATTRIBUTES.size

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,25 @@
...
+        end
+      end
+
+      def report_to_statsd
+        ps_list.each do |process|
+          timeout = Time.parse(process.fetch('start')) < Time.now - Rails.application.config.samson.deploy_timeout
```
grosser: more like `timed_out`

grosser: now - time -> time.ago

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,25 @@
...
+          tags = ATTRIBUTES.map { |attr| "#{attr}:#{process.fetch(attr)}" }
```
grosser: pass them in as an array (no hash coversion) and then `ATTRIBUTES.zip(process).map { |k, v| "#{k}:#{v}" }`

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,25 @@
...
+          Samson.statsd.gauge("process.start.time", (timeout ? 2.0 : 1.0), tags: tags)
```
grosser: why report 1 or 2 ? ... report the real runtime ?
`Samson.statsd.gauge("process.overdue.runtime", Time.now.to_i - Time.parse(process.fetch('start').to_i, tags: tags)`

test/lib/samson/process_utils_test.rb
```diff
@@ -0,0 +1,38 @@
...
+      Samson::ProcessUtils.ps_list.map { |ps| ps['args'] }.wont_include "sh -c echo samson_completed\n"
+    end
+  end
+
+  describe '.report_to_statsd' do
+    it 'report the processes to statsd' do
+      IO.popen("echo samson_statsd; sleep 5")
+      Samson::ProcessUtils.ps_list.each do |process|
+        tags = ["pid:#{process['pid']}", "ppid:#{process['ppid']}", "gid:#{process['gid']}", "pcpu:#{process['pcpu']}",
+                "user:#{process['user']}", "start:#{process['start']}", "args:#{process['args']}"]
```
grosser: 2-space indent plz so things don't move when the var name changes

lib/samson/process_utils.rb
sathishavm: It's hard to access attributes without hash conversion, I hope HASH formation shouldn't create much difference.

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Samson
+  module ProcessUtils
+    ATTRIBUTES = ['pid', 'ppid', 'gid', 'pcpu', 'user', 'start', 'args'].freeze
+
+    class << self
+      def ps_list
+        whitelists = ENV.fetch('PROCESS_WHITELIST', [])
```
grosser: env can only be a string, so the default should not be an array

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,29 @@
...
+        pipe = IO.popen("ps -eo #{ATTRIBUTES.join(',')}")
+        # Filter the whitelist processes
```
grosser: `# Ignore known long-running processes so report is meaningful`

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,29 @@
...
+        end
+        filtered_processes.map do |line|
+          Hash[ATTRIBUTES.zip line.lstrip.split(/\s+/, ATTRIBUTES.size)]
+        end
+      end
+
+      def report_to_statsd
+        ps_list.each do |process|
+          timed_out = Time.now.to_i - Time.parse(process.fetch('start')).to_i
+          tags = ATTRIBUTES.map { |attr| "#{attr}:#{process.fetch(attr)}" }
```
grosser: can do:
`start = process[ATTRIBUTES.index('start')]`
and
`tags = ATTRIBUTES.each_with_index.map { |attr, i| "#{attr}:#{process[i]}" }`
to avoid hash coversion
... but :whatever 🤷‍♀️

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,29 @@
...
+          whitelists.any? { |key| lines.include?(key) }
+        end
+        filtered_processes.map do |line|
+          Hash[ATTRIBUTES.zip line.lstrip.split(/\s+/, ATTRIBUTES.size)]
+        end
+      end
+
+      def report_to_statsd
+        ps_list.each do |process|
+          timed_out = Time.now.to_i - Time.parse(process.fetch('start')).to_i
```
grosser: prefer `Integer(value)` over `value.to_i` since it will fail when it is not an integer

grosser: `runtime` instead of `timed_out` ?

lib/samson/process_utils.rb
```diff
@@ -0,0 +1,29 @@
...
+          tags = ATTRIBUTES.map { |attr| "#{attr}:#{process.fetch(attr)}" }
+          Samson.statsd.gauge("process.overdue.runtime", timed_out, tags: tags)
```
grosser: they are not overdue by definition ... it's just runtime

lib/samson/process_utils.rb
sathishavm: Time parse will not return nil, It may return ArgumentError, I have rescued it and will add a monitor for this date issues.

grosser:
link the PR that adds the monitor to this one so it's easy to follow up in case something went wrong :)

PR #2831, opened 2018-06-30
[WIP] Allow Multiple Stage Commands to be Created/Reordered on Stage Edit

ragurney:

### DIFF
--- a/app/assets/stylesheets/stages.scss
+++ b/app/assets/stylesheets/stages.scss
@@ -89,6 +89,10 @@
   }
 }
 
+.stage-command-button {
+  margin: 1em 0 0 0;
+}
+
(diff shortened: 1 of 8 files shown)


grosser:
This whole commands thing is already complicated and adding more layers on top won't make it better.
The problem is kinda simple and using `command` was actually the wrong direction ... much easier to have a single assignable and then build ordered stage-commands from that, just like a mass-assignment but without allowing to assign attributes

POC https://github.com/zendesk/samson/pull/2832
needs polish/love and maybe a `duplicate` button for the new command, but gets rid of the old cruft and does not need more actions/js/workarounds

ragurney:
Superseded by #2833

PR #2833, opened 2018-07-02
[BRE-761] Allow for multiple stage command creation

ragurney:
screen shot 2018-07-02 at 12 45 26 pm

Based off of #2832 

/cc @zendesk/samson

### References
 - Jira link: [BRE-761](https://zendesk.atlassian.net/browse/BRE-761)

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 class Stage < ActiveRecord::Base
   AUTOMATED_NAME = 'Automated Deploys'
 
@@ -37,7 +38,6 @@ class Stage < ActiveRecord::Base
   validate :validate_not_auto_deploying_without_buddy
 
   before_create :ensure_ordering
-  before_save :append_new_command
   after_destroy :destroy_deploy_groups_stages
   after_soft_delete :destroy_deploy_groups_stages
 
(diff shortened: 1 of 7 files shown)


app/models/stage.rb
```diff
@@ -159,30 +159,21 @@ def environments
...
   def command_ids
     stage_commands.sort_by(&:position).map(&:command_id)
   end
 
-  def commands
-    stage_commands.sort_by(&:position).map(&:command).compact
-  end
-
-  def command=(c)
+  # reconcile new commands and positioning from command checkboxes
```
grosser: `or textare`

app/models/stage.rb
```diff
@@ -252,19 +243,25 @@ def validate_not_auto_deploying_without_buddy
...
   end
 
   def project_lock?(lock)
     lock.resource_type == "Project" && lock.resource_equal?(project)
   end
+
+  def find_or_create_stage_command(command_or_id, position)
+    command_or_id = command_or_id.to_s
+
+    if existing_command = command_or_id.match?(/\A\d+\z/) && Command.find_by(id: Integer(command_or_id))
```
grosser: n+1 ... only rely on `.detect`

grosser: if you want to make sure it exists, could do `ids = Command.pluck(:id)` ... or some other way of making sure typing an integer does not mess it up :D

app/models/stage.rb
```diff
@@ -252,19 +243,25 @@ def validate_not_auto_deploying_without_buddy
...
+      if stage_command = stage_commands.detect { |sc| sc.command_id == existing_command.id }
+        stage_command
+      else
+        StageCommand.new(command: existing_command)
```
grosser: did not assign project ?

app/views/stages/_commands.html.erb
```diff
@@ -10,38 +10,49 @@
...
+</fieldset>
 
-  <hr />
+<hr/>
 
-  <p>Add a command which is specific to your project</p>
-  <%= form.text_area :command, class: "form-control", rows: 4, value: nil, placeholder: 'cap production deploy' %>
-</fieldset>
+<%= button_tag(
+        'Add Stage Command',
```
grosser: strange indent

app/models/stage.rb
ragurney: I think if the command already exists then it should either have the correct `project_id` already assigned to it, or it's a global command and so shouldn't have a `project_id` assigned. Is that right?

ragurney:
I don't think `StageCommand` has a `project_id` field.. just a `stage_id`.

grosser:
ah right!

PR #2834, opened 2018-07-02
Allow triggering deploys to be destroyed

ragurney:
Ran across debugging another issue. Allows `Deploy.destroy_all` again, which was being prevented by the `triggering_deploy_id` fk. 

/cc @zendesk/samson


grosser:
idk ... deploys should never be destroyed especially not cascading
maybe disable the fk ?

ragurney:
Sure, we can also just close this too if you'd rather not introduce it. It was more of a convenience thing when I was messing with my db during development ¯\_(ツ)_/¯

grosser:
ok

ragurney:
🙅‍♂️

PR #2836, opened 2018-07-04
POC replace sse-rails-engine / angular with actioncable

grosser:
seems relatively simple/stable and we'll have 1 less 1-off tool in our gemfile :)

initial progress was good, but noticed a few issues that need to be addressed:

 - actioncable is not a controller, so rendering things is kinda awkward (was weird before too)
 - actioncable broadcasts to all subscribers, so info cannot be personalized, so again we should remove html rendering since it includes current-user etc
 - viewer list should be filtered clientside
 - refactor desktop notifications to not use angular

TODO:
 - [ ] viewer tracking needs to be done when subscribing, so fetch authenticate the connection and fetch it from there
 - [ ] deleted some viewer logic from job-eecution, need that back
 - [ ] add auth / test how bad auth responds
 - [ ] add secure-web-sockets
 - [ ] covert profile to jquery


grosser:
split into multiple PRs

PR #2837, opened 2018-07-05
covert deploys badge and active deploys to actioncable

grosser:


grosser:
@zendesk/samson 

extracted from https://github.com/zendesk/samson/pull/2836 which was getting a little too full

grosser:
@henders

app/models/deploy_service.rb
```diff
@@ -131,7 +131,9 @@ def notify_outbound_webhooks(deploy)
     deploy.stage.outbound_webhooks.each { |webhook| webhook.deliver(deploy) }
   end
 
-  def send_sse_deploy_update(type)
-    SseRailsEngine.send_event('deploys', type: type)
+  def send_deploy_update(finished = false)
```
ragurney: Kinda funky signature / method use combo passing in a symbol and defaulting to a boolean. Maybe make it a keyword arg `finished: false`? Or else rename it to `exclude_self` or `decrement` to make things more clear?

app/views/layouts/_header.html.erb
```diff
@@ -43,15 +43,11 @@
         <li class="<%= 'active' if request.path =~ %r{^/deploys[/\?]?$} %>">
           <%= link_to "Deploys", deploys_path %>
         </li>
-        <li class="current-deploys <%= 'active' if current_page?(controller: '/deploys', action: 'active') %>" ng-controller="currentDeployBadgeCtrl">
+        <li class="<%= 'active' if current_page?(controller: '/deploys', action: 'active') %>">
```
ragurney: Would this be cleaner as logic in a ViewModel e.g. 
```
...
def initialize(active = false)
  @active = active
end

def display_active_class
  @active ? 'active' : ''
end
```

 or test for the presence of an instance variable instantiated in the controller instead?

app/models/deploy_service.rb
```diff
@@ -131,7 +131,9 @@ def notify_outbound_webhooks(deploy)
...
+    count = Deploy.active_count
+    count -= 1 if finished # deploy is still active, so we substract one
```
ragurney: Very slight possibility of an off by one due to caching of `active_count`, but that's only if a deploy is shorter than 10 seconds :p

grosser:
need to add it, for now it's not critical info, but next PR that actually
adds deploy data will need it

app/models/deploy_service.rb
grosser: yeah but eventually consistent ... previous code did a +1/-1 counter so it could have been off forever 🤷‍♀️

app/views/layouts/_header.html.erb
grosser: `@active` is not a global concept, any link can be active/inactive ... would be nice to do a `link_to_with_active` where it checks if the url is the current url or so ... but that's for another PR

grosser:
per https://samsaffron.com/archive/2015/12/29/websockets-caution-required

"The infamous 6 connections per host limit does not apply to WebSockets. Instead a far bigger limit holds (255 in Chrome and 200 in Firefox)."
... but I can still reproduce it in staging ... but not on localhost

adammw:
maybe nginx is limiting in someway?

grosser:
seems unlikely, but possible I guess ... that's a problem for next PR :)

grosser:
turns out it actually fixes the issue, but staging did not restart on deploy 🎉 😞

PR #2843, opened 2018-07-09
use uid fix for all resources since we saw it for daemonset rollback too

grosser:
@zendesk/compute


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -47,6 +47,10 @@ def deploy
 
       def revert(previous)
         if previous
+          # Avoid "the object has been modified" / "Precondition failed: UID in precondition" error
+          # by removing internal attributes kubernetes adds
+          previous = previous.deep_dup
+          previous[:metadata].except! :selfLink, :uid, :resourceVersion, :generation, :creationTimestamp
```
adammw: make this array a constant/variable with a descriptive name?

PR #2847, opened 2018-07-10
kubernetes: show exception on connection error

raittes:
* Shows the exception when failed to connect to Kubernetes API (anon user, no permission, etc)

/cc @zendesk/samson

### Risks
- Level: Low


grosser:
neat!

plugins/kubernetes/test/models/kubernetes/cluster_deploy_group_test.rb
```diff
@@ -22,7 +22,9 @@
     it "is invalid when namespace lookup fails" do
       assert_request(:get, "http://foobar.server/api/v1/namespaces", to_return: {status: 404}) do
         refute_valid group
-        group.errors.full_messages.must_equal ["Namespace error looking up namespaces"]
+        group.errors.full_messages.any? do |m|
```
grosser: this does not make a lot of sense since it will stop on the first message :)

grosser: also you did not test the actual change ... this test would have passed before too

PR #2849, opened 2018-07-13
Allow users who can deploy to a project to cancel deploys

elaineoc:
* This change will give any user with deploy permissions to a project the permission to cancel any deploy on that project

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: n/a

### Risks
- Level: Low - Issues cancelling deploys

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -60,10 +60,6 @@ class Job < ActiveRecord::Base
     updated_at - created_at
   end
 
-  def can_be_cancelled_by?(user)
-    started_by?(user) || user.admin? || user.admin_for?(project)
-  end
-
   def commands
     commands = []
     raw = command.split(/\r?\n|\r/)
--- a/app/views/shared/_output.html.erb
+++ b/app/views/shared/_output.html.erb
@@ -10,9 +10,7 @@
       <button class="btn btn-default only-finished" id="output-expand-toggle">Expand</button>
       <%= link_to "Download log", [@project, deployable, {format: :text}], class: "btn btn-primary only-finished" %>
 
-      <% if deployable.can_be_cancelled_by?(current_user) %>
-        <%= cancel_button deploy: deployable, project: @project, class: "btn btn-danger only-active" %>
-      <% end %>
+      <%= cancel_button deploy: deployable, project: @project, class: "btn btn-danger only-active" %>
     </div>
   </div>
 
--- a/test/controllers/deploys_controller_test.rb
+++ b/test/controllers/deploys_controller_test.rb
@@ -502,22 +502,8 @@ describe DeploysController do
           delete :destroy, params: {project_id: project.to_param, id: deploy.to_param}
         end
 
-        it "cancels a deploy" do
-          flash[:error].must_be_nil
-        end
-      end
-
-      describe "with a deploy not owned by the user" do
-        before do
-          deploy_service.expects(:cancel).never
-          Deploy.any_instance.stubs(:started_by?).returns(false)
-          User.any_instance.stubs(:admin?).returns(false)
-
-          delete :destroy, params: {project_id: project.to_param, id: deploy.to_param}
-        end
-
-        it "doesn't cancel the deloy" do
-          flash[:error].wont_be_nil
+        it "redirects to project page" do
+          assert_redirected_to "/projects/foo/deploys/#{deploy.id}"
         end
       end
     end
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -108,12 +108,7 @@ class DeploysController < ApplicationController
   end
 
   def destroy
-    if @deploy.can_be_cancelled_by?(current_user)
-      @deploy.cancel(current_user)
-    else
-      flash[:error] = "You do not have privileges to cancel this deploy."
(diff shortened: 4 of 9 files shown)


app/models/job.rb
```diff
@@ -61,7 +61,7 @@ def duration
   end
 
   def can_be_cancelled_by?(user)
-    started_by?(user) || user.admin? || user.admin_for?(project)
+    started_by?(user) || user.admin? || user.deployer_for?(project)
```
grosser: I think this can just be `user.deployer_for?(project)` since only deployers can start a deploy ...
@ragurney sounds alright ?

ragurney:
@elaineoc are you still interested in adding this change?

elaineoc:
@ragurney Yes I am, I hit a very small blocker on the tests and didn't have time to come back to it yet.

app/models/job.rb
elaineoc: I changed it to `user.admin? || user.deployer_for?(project)` so that all admins and all deployers on the project will have permission to cancel the deploy.

elaineoc:
💇‍♂️ @ragurney @grosser

test/controllers/deploys_controller_test.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 1
+SingleCov.covered! uncovered: 2
```
elaineoc: I incremented this because it was not straightforward to test the case where the deployer of another project can not cancel a deploy of this project. I assumed that would fall into the 'viewer' role, but it seems that role will be denied before it gets to the `can_be_cancelled_by` point.

ragurney:
Looking at this a little more, it seems to me that the before action which eventually calls:
https://github.com/zendesk/samson/blob/1641af84662072beef5a69ed63e2fb281b1f9b18/app/models/user.rb#L129-L131
is equivalent to the new `can_be_canceled_by?`
```ruby
user.admin? || user.deployer_for?(project)
```

So we could just get rid of the usage of `can_be_canceled_by` and that if/else statement because that `else` path will never be taken.

test/controllers/deploys_controller_test.rb
```diff
@@ -509,15 +509,15 @@ def pull_request_stub(data_overrides = {})
 
       describe "with a deploy not owned by the user" do
```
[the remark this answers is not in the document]
elaineoc: I believe this test should change because now a user who is not the owner of the deploy can cancel a deploy - any deployer of the project should be able to cancel any deploy within the project, regardless of who owns it.

app/models/job.rb
```diff
@@ -61,7 +61,7 @@ def duration
   end
 
   def can_be_cancelled_by?(user)
-    started_by?(user) || user.admin? || user.admin_for?(project)
+    user.admin? || user.deployer_for?(project)
```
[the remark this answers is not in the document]
elaineoc: I was trying not to change the behaviour too much from what it currently is, so I thought it made sense for any admin and any deployer of the project to be able to cancel the deploy. But I don't have a clear view of the difference between these roles in Samson. Do you think it would be better for any deployer at all to be able to cancel a deploy?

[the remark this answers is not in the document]
grosser: `deployer?` means `global_deployer?` so yes :)

test/controllers/deploys_controller_test.rb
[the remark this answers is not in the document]
grosser: hmmm yes, but from the perspective of `what if this code never existed` the test feels redundant ... and the test we'd need would be `as a viewer I cannot cancel` ... but as ryan said below, this can all go away :)

elaineoc:
@ragurney @grosser 💇

test/controllers/deploys_controller_test.rb
ragurney: I think this can be decremented now 🎉

elaineoc:
I made that change :) Thanks for reviewing. According to the Readme you will deploy, is that still the case? Do you also merge to master, or should I do that?

app/views/shared/_output.html.erb
```diff
@@ -10,9 +10,7 @@
       <button class="btn btn-default only-finished" id="output-expand-toggle">Expand</button>
       <%= link_to "Download log", [@project, deployable, {format: :text}], class: "btn btn-primary only-finished" %>
 
-      <% if deployable.can_be_cancelled_by?(current_user) %>
-        <%= cancel_button deploy: deployable, project: @project, class: "btn btn-danger only-active" %>
-      <% end %>
+      <%= cancel_button deploy: deployable, project: @project, class: "btn btn-danger only-active" %>
```
grosser: this should only be visible for deployers (not viewers) ... or clicking the button should give a nice error message/prompt for login

grosser:
hmmm nvm atm I get 
screen shot 2018-09-04 at 9 36 13 am
when trying to cancel as viewer, which is pretty good

ragurney:
@elaineoc I'll go ahead and merge and get this deployed to production soon.

elaineoc:
thanks @ragurney :)

PR #2850, opened 2018-07-13
docs: add deploy group flag to kubernetes guide

raittes:
* Add to Readme required environment variable in order to get Kubernetes plugin working.

/cc @zendesk/samson


grosser:
thx ... never realized that since I always turn that on :)

PR #2858, opened 2018-07-19
expire secrets cache when a vault server is added or has a connection…

grosser:
… related attribute changed

I updated a bad vault server and no secrets were shown, which looked like it was still broken

@zendesk/compute @ragurney

### DIFF
--- a/lib/samson/secrets/vault_server.rb
+++ b/lib/samson/secrets/vault_server.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 require 'vault'
 
-# replace once https://github.com/hashicorp/vault-ruby/pull/118 is released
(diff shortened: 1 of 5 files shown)


lib/samson/secrets/vault_server.rb
```diff
@@ -101,6 +102,11 @@ def create_client
 
       private
 
+      def expire_secrets_cache
+        return if (previous_changes.keys - ["name"]).empty?
```
adammw: extract ["name"] into a constant, like ALLOWED_CHANGES or something...

grosser:
also moved the db url first so it's easy to see what failed
screen shot 2018-07-19 at 3 52 39 pm

PR #2861, opened 2018-07-23
Scrub session from rollbar

adammw:
Scrub session from rollbar. Doesn't seem like something we need to be storing. 
(Haven't tested that this patch actually works, needs Rollbar to be set up for that)

/cc @zendesk/samson @grosser 

### Tasks
 - [ ] :+1: from team

### DIFF
--- a/plugins/rollbar/config/initializers/rollbar.rb
+++ b/plugins/rollbar/config/initializers/rollbar.rb
@@ -14,7 +14,6 @@ if token = ENV['ROLLBAR_ACCESS_TOKEN']
     config.code_version = Rails.application.config.samson.version&.first(7)
     config.populate_empty_backtraces = true
     config.logger = Rails.logger
-    config.scrub_fields |= Rails.application.config.filter_parameters + ['HTTP_AUTHORIZATION']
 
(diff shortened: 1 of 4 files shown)


grosser:
can use a token from rollbar settings page

grosser:
`      "_samson_session_development": "****", `

grosser:
rollbar issue https://github.com/rollbar/rollbar-gem/issues/765

plugins/rollbar/config/initializers/rollbar.rb
```diff
@@ -14,7 +14,6 @@
     config.code_version = Rails.application.config.samson.version&.first(7)
     config.populate_empty_backtraces = true
     config.logger = Rails.logger
-    config.scrub_fields |= Rails.application.config.filter_parameters + ['HTTP_AUTHORIZATION']
```
grosser: we don't need this since it's rollbar default behavior via https://github.com/rollbar/rollbar-gem/blob/24c4de23c4150d19b408ed14f7839f8cc6e1bc69/lib/rollbar/request_data_extractor.rb#L240

adammw:
:+1:

PR #2863, opened 2018-07-24
do not fail seeding with an exception when too many roles have errors

grosser:
@ragurney 
/cc @zendesk/compute 

previously would result in a cookieoverflow since the limit message is quiet long ... there seems to be no generic way to cut down alerts to not overflow cookies, so using this until it happens again ...


plugins/kubernetes/app/controllers/kubernetes/deploy_group_roles_controller.rb
```diff
@@ -88,7 +88,7 @@ def seed
         errors = ["Roles failed to seed, fill them in manually."]
         errors.concat(created.map do |dgr|
           "#{dgr.kubernetes_role.name} for #{dgr.deploy_group.name}: #{dgr.errors.full_messages.to_sentence}"
-        end)
+        end.first(3))
```
ragurney: Maybe add a '...' or something to indicate the messages are being truncated if there are more than 3?

PR #2865, opened 2018-07-26
stream deploy output via actioncable

grosser:
/cc @zendesk/samson @ragurney 

WIP wil break a bunch of tests but seems to work ... let's go over this tomorrow to see if anything seems obviously broken ...

basically making job output a single stream that all users consume, then request the header as necessary and filter themselves from the viewers


app/views/shared/_output_stream.html.erb
```diff
@@ -1,6 +1,6 @@
 <% if job.active? %>
```
ragurney: So `viewers` tab only works on active jobs. Maybe we hide that tab if job is inactive?

app/assets/javascripts/channels/job_outputs_channel.js
```diff
@@ -0,0 +1,76 @@
...
+    },
+    viewers: function(users){
+      users = users.filter(function(user) { return user.id != userId});
+      if (users.length > 0) {
+        var viewers = $.map(users, function(user) {
+          return user.name;
+        }).join(', ') + '.';
+
+        $('#viewers-link .badge').html(users.length);
+        $("#viewers").html('Other viewers: ' + viewers);
```
ragurney: XSS vulnerability 😱

grosser:
ready!

ragurney:
What about removing the route / other stream-urls?

grosser:
done

grosser:
streaming works everywhere now, so here goes nothing ...

PR #2870, opened 2018-08-03
use http as workaround for expired cert

grosser:
https://github.com/tenex/rails-assets/issues/430


Kenneth-KT:
As of 3 Aug 2018 15:00 UTC, https://rails-assets.org now has a valid SSL certificate, so this workaround might not be necessary.
https://github.com/tenex/rails-assets/issues/430#issuecomment-410246733

grosser:
thx!

PR #2871, opened 2018-08-06
Get Actioncable logs in JSON format

sathishavm:
* Convert and display action-cable logs in JSON format.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: @zendesk/bre 

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-219

### Risks
- Level: Low


lib/samson/logging.rb
```diff
@@ -23,9 +23,11 @@
     }
   end
 
-  require 'syslog/logger'
   config.logger = Syslog::Logger.new('samson')
+  config.action_cable.logger = Syslog::Logger.new('samson')
```
grosser: reuse the logger from above ?

lib/samson/syslog_formatters.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'syslog/logger'
+module Syslog
```
grosser: should be in `Samson` module

grosser:
```
require 'syslog/logger/formatter'
module Samson
  class SyslogFormatter < Syslog::Logger::Formatter
```

lib/samson/logging.rb
grosser: ah I guess it needs a different formatter ? ... Logstash did not work ?

test/lib/samson/syslog_formatters_test.rb
```diff
@@ -0,0 +1,17 @@
...
+SingleCov.covered!
+
+describe Syslog::Formatters do
+  describe '.json' do
+    it 'returns a json formatted log' do
+      output = StringIO.new
+      logger = Logger.new(output)
+      logger.formatter = Syslog::Formatters::Json.new
+      logger.info('test')
+      output.string.must_equal "{:severity=>\"INFO\", :time=>#{Time.now}, :message=>\"test\"}"
```
grosser: that's not json :)
... need to add a .to_json ?

lib/samson/logging.rb
sathishavm: yup, Logstash is not working because of call method mismatch.

test/lib/samson/syslog_formatters_test.rb
sathishavm: This is because I have used 'StringIO' as Logger input, I tried 'STDOUT' as Logger input, but can't able to capture output.

test/lib/samson/syslog_formatters_test.rb
```diff
@@ -0,0 +1,17 @@
...
+SingleCov.covered!
+
+describe Samson::SyslogFormatter do
+  describe '.json' do
+    it 'returns a json formatted log' do
+      output = StringIO.new
+      logger = Logger.new(output)
+      logger.formatter = Samson::SyslogFormatter.new
+      logger.info('test')
+      output.string.must_equal "{:severity=>\"INFO\", :time=>#{Time.now}, :message=>\"test\"}"
```
grosser: this still needs to be json and not ruby-style hash

PR #2877, opened 2018-08-13
[PAAS-1939] always set spec.initContainers regardless of k8s version

jonmoter:
We're upgrading our K8S clusters from 1.5.7 to 1.10, and this is causing problems when Deployments that only have the annotations get new pods scheduled on Nodes running 1.10. The end result is no init containers at all.

I'll test to confirm it's safe to have `.spec.initContainers` set for earlier versions of Kubernetes.

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-1938

### Risks
- Level: Low

/cc @zendesk/samson, @zendesk/compute


jonmoter:
Actually, I suspect this won’t work after all. I tried doing this via kubectl, and the initContainer definition vanished from the template.

I think we need to do an intermediary upgrade to 1.7.

PR #2880, opened 2018-08-16
Bump mysql2 gem

adammw:
Bump mysql2 gem to fix compatibility with Ruby 2.5

/cc @zendesk/samson @pschambacher 

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


grosser:
worked fine for me locally and on the servers ...

PR #2881, opened 2018-08-16
Add mass rollout for k8s stages

adammw:
Refactor mass rollout to include k8s stage options.

screen shot 2018-09-10 at 1 20 11 pm

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-1970

### Risks
- Level: Medium
- While shouldn't break any existing functionality, the new functionality has the potential to create many K8s deploys so should be rigorously reviewed/tested.

### DIFF
--- a/app/controllers/mass_rollouts_controller.rb
+++ b/app/controllers/mass_rollouts_controller.rb
@@ -38,7 +38,7 @@ class MassRolloutsController < ApplicationController
     end
 
     if deploys.empty?
-      flash[:error] = "There were no stages that matched the mass rollout deploy criteria."
+      flash[:error] = "No deployable stages found."
       redirect_to deploys_path
     else
       redirect_to deploys_path(ids: deploys.map(&:id))
--- a/app/controllers/mass_rollouts_controller.rb
+++ b/app/controllers/mass_rollouts_controller.rb
@@ -16,23 +16,37 @@ class MassRolloutsController < ApplicationController
   end
 
   def deploy
-    environment = deploy_group.environment
-    template_stages = environment.template_stages.all
-    missing_only = params[:missing_only] == "true"
-    stages_to_deploy = missing_only ? deploy_group.stages.reject(&:last_successful_deploy) : deploy_group.stages
-    deploys = stages_to_deploy.map do |stage|
-      template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
-      next unless template_stage
+    successful, missing = deploy_group.stages.partition(&:last_successful_deploy)
+    stages_to_deploy = []
 
-      last_success_deploy = template_stage.last_successful_deploy
-      next unless last_success_deploy
+    if params[:successful] == "true"
+      stages_to_deploy += successful.map { |stage| [stage, stage.last_successful_deploy.reference] }
+    end
 
+    if params[:missing] == "true"
+      stages_to_deploy += missing.map do |stage|
+        ref = stage.template_stage.last_successful_deploy&.reference if stage.template_stage&.is_template?
+        [stage, ref] if ref
+      end.compact
+    end
+
+    if defined?(SamsonKubernetes::Engine)
+      if params[:kubernetes] != "true"
+        stages_to_deploy.reject! { |stage, _| stage.kubernetes? }
+      end
+
+      if params[:non_kubernetes] != "true"
+        stages_to_deploy.select! { |stage, _| stage.kubernetes? }
+      end
+    end
+
+    deploys = stages_to_deploy.map do |stage, reference|
       deploy_service = DeployService.new(current_user)
-      deploy_service.deploy(stage, reference: last_success_deploy.reference)
-    end.compact
+      deploy_service.deploy(stage, reference: reference)
+    end
 
     if deploys.empty?
-      flash[:error] = "There were no stages ready for deploy."
+      flash[:error] = "There were no stages that matched the mass rollout deploy criteria."
       redirect_to deploys_path
     else
       redirect_to deploys_path(ids: deploys.map(&:id))
--- a/app/controllers/mass_rollouts_controller.rb
+++ b/app/controllers/mass_rollouts_controller.rb
@@ -19,28 +19,20 @@ class MassRolloutsController < ApplicationController
     successful, missing = deploy_group.stages.partition(&:last_successful_deploy)
     stages_to_deploy = []
 
-    if params[:successful] == "true"
-      stages_to_deploy += successful.map { |stage| [stage, stage.last_successful_deploy.reference] }
-    end
-
-    if params[:missing] == "true"
-      stages_to_deploy += missing.map do |stage|
-        ref = stage.template_stage.last_successful_deploy&.reference if stage.template_stage&.is_template?
-        [stage, ref] if ref
-      end.compact
-    end
+    stages_to_deploy += successful if params[:successful] == "true"
+    stages_to_deploy += missing if params[:missing] == "true"
 
     if defined?(SamsonKubernetes::Engine)
-      if params[:kubernetes] != "true"
-        stages_to_deploy.reject! { |stage, _| stage.kubernetes? }
-      end
-
-      if params[:non_kubernetes] != "true"
-        stages_to_deploy.select! { |stage, _| stage.kubernetes? }
-      end
+      stages_to_deploy.reject!(&:kubernetes?) if params[:kubernetes] != "true"
+      stages_to_deploy.select!(&:kubernetes?) if params[:non_kubernetes] != "true"
     end
 
-    deploys = stages_to_deploy.map do |stage, reference|
+    deploy_references = stages_to_deploy.map do |stage|
+      reference = get_reference_from_template_stage(stage)
+      [stage, reference] if reference
+    end.compact
+
+    deploys = deploy_references.map do |stage, reference|
       deploy_service = DeployService.new(current_user)
       deploy_service.deploy(stage, reference: reference)
     end
@@ -160,6 +152,11 @@ class MassRolloutsController < ApplicationController
     end
 
     stage
+  end
+
+  def get_reference_from_template_stage(stage)
+    template_stage = deploy_group.environment.template_stages.find_by(project_id: stage.project_id)
+    template_stage&.last_successful_deploy&.reference
   end
 
   def deploy_group
--- a/app/controllers/mass_rollouts_controller.rb
+++ b/app/controllers/mass_rollouts_controller.rb
@@ -16,30 +16,28 @@ class MassRolloutsController < ApplicationController
   end
 
   def deploy
-    stages_to_deploy = deploy_group.stages
+    stages_to_deploy = deploy_group.stages.to_a
 
-    case params[:status]
-    when ->(type) { type.blank? }
-      # no-op
+    case params[:status].presence
+    when nil # rubocop:disable Lint/EmptyWhen
+      # all
     when 'successful'
       stages_to_deploy.select!(&:last_successful_deploy)
     when 'missing'
       stages_to_deploy.reject!(&:last_successful_deploy)
     else
-      return render status: :bad_request
+      return unsupported_option(:status)
     end
 
-    if defined?(SamsonKubernetes::Engine)
-      case params[:type]
-      when ->(type) { type.blank? }
-        # no-op
-      when 'kubernetes'
-        stages_to_deploy.select!(&:kubernetes?)
-      when 'non_kubernetes'
-        stages_to_deploy.reject!(&:kubernetes?)
-      else
-        return render status: :bad_request
-      end
+    case params[:kubernetes].to_s.presence
+    when nil # rubocop:disable Lint/EmptyWhen
+      # all
+    when 'true'
+      stages_to_deploy.select!(&:kubernetes?)
+    when 'false'
+      stages_to_deploy.reject!(&:kubernetes?)
+    else
+      return unsupported_option(:kubernetes)
     end
 
     deploy_references = stages_to_deploy.map do |stage|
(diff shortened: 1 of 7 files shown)


grosser:
I'd prefer not to add 1-off buttons but more of a query interface for stages

Deploy all: `[] Kubernetes [] Previously successful`

app/controllers/mass_rollouts_controller.rb
```diff
@@ -148,6 +145,16 @@ def create_stage_with_group(template_stage)
     stage
   end
 
+  def deploy_stage_reference(template_stages, stage)
+    template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
```
grosser: can turn this into a nice 1-line `return unless template_stage =`
or even better:
`template_stages.detect { |ts| ts.project_id == stage.project.id }&.last_successful_deploy&.reference`

adammw:
Hmm, any idea how we add that in given that Kubernetes is a plugin? Rewrite the mass deploy to accept hooks?

plugins/kubernetes/test/controllers/kubernetes/mass_rollouts_controller_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+
+      it 'ignores stages with no successful deployment' do
+        @k8s_stage.deploys.delete_all
+        deploys(:failed_staging_test).update(stage: @k8s_staget)
+
+        assert_no_difference '@k8s_stage.deploys.count' do
+          post :deploy, params: {deploy_group_id: deploy_group}
+        end
+        assert_match /no stages ready/, flash[:error]
```
grosser: prefer .must_include / assert_includes since it is simpler and has a more readable error message

grosser:
hardcode or add a new hook, once we get it working/agreed on moving it out
should be easy

adammw:
updated. can't get my head around the fixtures in these tests, they're rather complicated.

grosser:
looks awesome, I'll take a look

adammw:
done. added uncovered tests somehow but not sure why...

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,37 @@ def create
...
+
+    if params[:successful] == "true"
+      stages_to_deploy += successful.map { |stage| [stage, stage.last_successful_deploy.reference] }
+    end
 
-      last_success_deploy = template_stage.last_successful_deploy
-      next unless last_success_deploy
+    if params[:missing] == "true"
+      stages_to_deploy += missing.map do |stage|
+        ref = stage.template_stage.last_successful_deploy&.reference if stage.template_stage&.is_template?
```
adammw: note this behaviour is slightly different.. it'll use the actual template stage referenced under the stage **NOT** just the first template stage that matches the project ID and environment. not sure is a breaking change or not. /cc @zendesk/bre

adammw:
An even better UI for this would be where mass rollout took you to a separate page with a list of every stage it would deploy, and you can manually check/uncheck certain stages, with js buttons to check/uncheck certain presets like kubernetes only, missing only, deployed only, etc.

grosser:
not sure ...  that page could be messy with all the stages we have

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,37 @@ def create
...
-    missing_only = params[:missing_only] == "true"
-    stages_to_deploy = missing_only ? deploy_group.stages.reject(&:last_successful_deploy) : deploy_group.stages
-    deploys = stages_to_deploy.map do |stage|
-      template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
-      next unless template_stage
+    successful, missing = deploy_group.stages.partition(&:last_successful_deploy)
+    stages_to_deploy = []
+
+    if params[:successful] == "true"
+      stages_to_deploy += successful.map { |stage| [stage, stage.last_successful_deploy.reference] }
```
grosser: this is changing the behavior ... before it was deploying the template stages last deploy

grosser:
something more like this ?

```Ruby
stages = deploy_group.stages

    # kubernetes filter (cheap, so do it first)
    if selected = params[:kubernetes]
      stages.select! { |s| s.kubernetes.to_s == selected }
    end

    if selected = params[:success]
      stages.select! { |s| !!(s.last_successful_deploy).to_s == selected }
    end

    environment = deploy_group.environment
    template_stages = environment.template_stages.all
    deploy_targets = stages.map do |stage|
      template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
      next unless template_stage
      next unless reference = template_stage.last_successful_deploy&.reference
      [stage, reference]
    end

    deploys = deploy_targets.map do |stage, reference|
      deploy_service.deploy(stage, reference: reference)
    end
```

app/controllers/mass_rollouts_controller.rb
grosser: which is the indented way this should work since it's assuming the template stage is up to date

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,37 @@ def create
...
+    deploys = stages_to_deploy.map do |stage, reference|
       deploy_service = DeployService.new(current_user)
-      deploy_service.deploy(stage, reference: last_success_deploy.reference)
-    end.compact
+      deploy_service.deploy(stage, reference: reference)
+    end
 
     if deploys.empty?
-      flash[:error] = "There were no stages ready for deploy."
+      flash[:error] = "There were no stages that matched the mass rollout deploy criteria."
```
grosser: maybe "No deployable stages found"

app/controllers/mass_rollouts_controller.rb
adammw: /cc @zendesk/bre for input on if the behaviour change is okay or if there's a use-case/need to keep overwriting the last deploy with what is on the template stage. to me, it seems more risky if the template stage is out-of-sync somehow.

grosser:
can we keep current behavior and have that discussion/change in another PR ?

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,29 @@ def create
...
 
-      last_success_deploy = template_stage.last_successful_deploy
-      next unless last_success_deploy
+    stages_to_deploy += successful if params[:successful] == "true"
+    stages_to_deploy += missing if params[:missing] == "true"
 
-      deploy_service = DeployService.new(current_user)
-      deploy_service.deploy(stage, reference: last_success_deploy.reference)
+    if defined?(SamsonKubernetes::Engine)
+      stages_to_deploy.reject!(&:kubernetes?) if params[:kubernetes] != "true"
```
grosser: smells like it should be a select ? (blank/yes/no)

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,29 @@ def create
...
-    deploys = stages_to_deploy.map do |stage|
-      template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
-      next unless template_stage
+    successful, missing = deploy_group.stages.partition(&:last_successful_deploy)
+    stages_to_deploy = []
 
-      last_success_deploy = template_stage.last_successful_deploy
-      next unless last_success_deploy
+    stages_to_deploy += successful if params[:successful] == "true"
+    stages_to_deploy += missing if params[:missing] == "true"
```
grosser: select for all/successful/missing so we never end up with none ?

app/controllers/mass_rollouts_controller.rb
```diff
@@ -148,6 +154,11 @@ def create_stage_with_group(template_stage)
     stage
   end
 
+  def get_reference_from_template_stage(stage)
```
grosser: <strike>get_<strike>

grosser: `def successful_reference_from_template_stage` ?

grosser:
yep ... make the select say all/successful/missing

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,34 @@ def create
...
-      next unless template_stage
+    stages_to_deploy = deploy_group.stages
 
-      last_success_deploy = template_stage.last_successful_deploy
-      next unless last_success_deploy
+    if params[:status] == 'successful'
+      stages_to_deploy.select!(&:last_successful_deploy)
+    elsif params[:status] == 'missing'
+      stages_to_deploy.reject(&:last_successful_deploy)
+    end
```
grosser: make this a case and fail fast on unknown ?

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,34 @@ def create
...
-      template_stage = template_stages.detect { |ts| ts.project_id == stage.project.id }
-      next unless template_stage
+    stages_to_deploy = deploy_group.stages
 
-      last_success_deploy = template_stage.last_successful_deploy
-      next unless last_success_deploy
+    if params[:status] == 'successful'
+      stages_to_deploy.select!(&:last_successful_deploy)
+    elsif params[:status] == 'missing'
+      stages_to_deploy.reject(&:last_successful_deploy)
```
grosser: should be `reject!` ?

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,34 @@ def create
...
 
-      deploy_service = DeployService.new(current_user)
-      deploy_service.deploy(stage, reference: last_success_deploy.reference)
+    if defined?(SamsonKubernetes::Engine)
+      if params[:type] == 'kubernetes'
+        stages_to_deploy.select!(&:kubernetes?)
+      elsif params[:type] == 'non_kubernetes'
+        stages_to_deploy.reject!(&:kubernetes?)
+      end
+    end
```
grosser: prefer a single select per param like this (to avoid duplication and bugs)

```
if kubernetes = params[:kubernetes].presence
  stages_to_deploy.select! { |s| s.kubernetes?.to_s == kubernetes }
end
```

app/controllers/mass_rollouts_controller.rb
grosser: ^

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,23 +16,34 @@ def create
...
+
+    deploy_references = stages_to_deploy.map do |stage|
+      reference = get_reference_from_template_stage(stage)
```
grosser: maybe even `last_successful_template_refernece`

adammw:
updated.

app/views/deploy_groups/show.html.erb
```diff
@@ -114,13 +99,80 @@
...
       </div>
-    <% end %>
-    <br>
+    </div>
+    <div class="form-group" id="mass_rollout" style="display: none">
+      <label class="col-lg-2 control-label">Mass Rollout</label>
+      <div class="col-lg-10">
+        <%= form_tag deploy_deploy_group_mass_rollouts_path(@deploy_group) do %>
+
+          <% if defined?(SamsonKubernetes::Engine) %>
```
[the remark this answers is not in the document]
adammw: not sure I understand.

adammw:
swapped 'all' for nil/blank

app/views/deploy_groups/show.html.erb
```diff
@@ -114,13 +99,80 @@
...
+            <div class="form-group">
+              <div class="col-lg-10">
+                <label>Stage Type</label>
+
+                <div class="form-group">
+                  <div class="col-lg-4 radio">
+                    <%= label_tag do %>
+                      <%= radio_button_tag :type, 'all', :checked %>
```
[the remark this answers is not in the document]
adammw: original idea was to add more types later. can always change it when we get there.

grosser:
moved a few things + streamlined tests ... @adammw

app/controllers/mass_rollouts_controller.rb
```diff
@@ -16,30 +16,28 @@ def create
...
 
-    case params[:status]
-    when ->(type) { type.blank? }
-      # no-op
+    case params[:status].presence
+    when nil # rubocop:disable Lint/EmptyWhen
+      # all
     when 'successful'
-      stages_to_deploy.select!(&:last_successful_deploy)
+      stages_to_deploy = stages_to_deploy.select(&:last_successful_deploy)
```
adammw: any reason not `select!` here but still using reject! and select! below?

test/controllers/mass_rollouts_controller_test.rb
```diff
@@ -47,7 +60,29 @@ def create_stages
...
+          assert_response :redirect
+        end
+      end
+
+      it "complains about unknown status" do
+        post :deploy, params: {deploy_group_id: deploy_groups(:pod100), status: "wut"}
+        assert_response :bad_request
+      end
+
+      describe "successful deploys" do
```
adammw: maybe "successful only mass rollout"?

test/controllers/mass_rollouts_controller_test.rb
```diff
@@ -57,88 +92,38 @@ def create_stages
...
-              project: stage101.project,
-              user: User.first,
-              status: "failed",
-              command: 'blah'
-            )
-          )
+          create_deploy stage100, "v121", "succeeded"
+          create_deploy stage100, "v123", "failed"
+          create_deploy stage101, "master", "succeeded"
+          create_deploy stage101, "v123", "failed"
```
adammw: nice cleanup

adammw:
:+1: merge when ready

PR #2882, opened 2018-08-17
Feature/delegate deploy commit to job model

torukMnk:
* TODO: remove this an delegate to job directly, a commit is not a reference

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: None

### Risks
- Level: Low

### DIFF
--- a/test/models/deploy_test.rb
+++ b/test/models/deploy_test.rb
@@ -134,9 +134,10 @@ describe Deploy do
       deploy.commit.must_equal "abcdef"
     end
 
-    it "falls back to deploys reference" do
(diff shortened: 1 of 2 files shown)


test/models/deploy_test.rb
```diff
@@ -134,9 +134,10 @@
       deploy.commit.must_equal "abcdef"
     end
 
-    it "falls back to deploys reference" do
```
grosser: can just remove this since we don't add test for other code that does not exist either :D

grosser:
why do you want to remove it ?

PR #2884, opened 2018-08-20
PAAS-1245 make people aware that old resources need to be taken down …

grosser:
…when changing names

(since automatic deletion/re-create is super tricky and dangerous)

@zendesk/compute 

screen shot 2018-08-20 at 3 09 45 pm

### DIFF
--- a/plugins/kubernetes/app/views/kubernetes/roles/_form.html.erb
+++ b/plugins/kubernetes/app/views/kubernetes/roles/_form.html.erb
@@ -7,9 +7,15 @@
     <%= form.input :config_file %>
     <%= form.input :service_name %>
     <%= form.input :resource_name,
-          pattern: Kubernetes::RoleValidator::VALID_LABEL,
-          help: "Name for the resource this role will create (service/job), a-Z 0-9 and -"
+        pattern: Kubernetes::RoleValidator::VALID_LABEL,
+        help: "Name for the resource this role will create (service/job), a-Z 0-9 and -"
     %>
+    <% if @role.manual_deletion_required? %>
+      <div class="alert-danger">
+        <%= form.input :manual_deletion_acknowledged, as: :check_box, label: "I will manually delete the changed resource/service from all deploy groups before deploying (this will cause a short outage, deleting after deploying will result in undefined behavior with 2 copies running)" %>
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/views/kubernetes/roles/_form.html.erb
```diff
@@ -7,9 +7,15 @@
...
     <%= form.input :service_name %>
     <%= form.input :resource_name,
-          pattern: Kubernetes::RoleValidator::VALID_LABEL,
-          help: "Name for the resource this role will create (service/job), a-Z 0-9 and -"
+        pattern: Kubernetes::RoleValidator::VALID_LABEL,
+        help: "Name for the resource this role will create (service/job), a-Z 0-9 and -"
     %>
+    <% if @role.manual_deletion_required? %>
+      <div class="alert-danger">
+        <%= form.input :manual_deletion_acknowledged, as: :check_box, label: "I will manually delete the changed resource/service from all deploy groups before deploying" %>
```
adammw: before? wouldn't that cause an outage? or will renames always cause an outage?

plugins/kubernetes/test/models/kubernetes/role_test.rb
```diff
@@ -88,6 +88,24 @@ def commit
...
+      it "is invalid when not acknowledged" do
+        refute_valid role
+      end
+
+      it "is valid when acknowledged" do
+        role.manual_deletion_acknowledged = true
+        assert_valid role
+      end
+
+      it "is valid when acknowledged" do
```
adammw: dupe

grosser:
changing service names will always cause an outage
changing deployment names without removing the old first will create 2
deployments with the same matchlabels which is kinda scary too (and
whatever other resources do like statefulset/job/pod) ...
so unless we make sure that is ok I'd go the safe route and delete first

plugins/kubernetes/app/views/kubernetes/roles/_form.html.erb
adammw: in that case.. maybe mention the word "outage" / "undefined behaviour" in the check box label...

PR #2885, opened 2018-08-20
Datadog APM

sathishavm:
* Added Datadog APM in matching with Newrelic
/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-865

### Risks
- Level: MED

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,4 +45,4 @@
 
 # .bak files
 *.bak
-/vendor/cache/**/
+/vendor/cache
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -22,6 +22,12 @@ PATH
       aws-sdk-ecr
 
 PATH
+  remote: plugins/datadog_tracer
+  specs:
+    samson_datadog_tracer (0.0.1)
+      ddtrace
+
+PATH
   remote: plugins/datadog
   specs:
     samson_datadog (0.0.0)
@@ -230,6 +236,8 @@ GEM
       safe_yaml (~> 1.0.0)
     crass (1.0.4)
     dalli (2.7.6)
+    ddtrace (0.14.2)
+      msgpack
     diffy (3.2.0)
     docker-api (1.33.6)
       excon (>= 0.38.0)
@@ -609,6 +617,7 @@ DEPENDENCIES
   samson_assertible!
   samson_aws_ecr!
   samson_datadog!
+  samson_datadog_tracer!
   samson_deploy_env_vars!
   samson_docker_binary_builder!
   samson_env!
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -61,6 +61,7 @@ class DockerBuilderService
     Samson::Hooks.fire(:before_docker_build, tmp_dir, @build, @output)
     execute_build_command(tmp_dir, @build.project.build_command)
   end
+  add_tracer :before_docker_build
 
   def build_image(tmp_dir, tag_as_latest:)
     File.write("#{tmp_dir}/REVISION", @build.git_sha)
@@ -82,5 +83,5 @@ class DockerBuilderService
       tmp_dir, @build, @execution.executor, tag_as_latest: tag_as_latest, cache_from: cache
     )
   end
-  add_method_tracers :before_docker_build, :build_image
+  add_tracer :build_image
 end
--- a/app/models/docker_builder_service.rb
+++ b/app/models/docker_builder_service.rb
@@ -2,7 +2,7 @@
 require 'docker'
 
 class DockerBuilderService
-  include ::NewRelic::Agent::MethodTracer
+  include ::Samson::PerformanceTracer
 
   def initialize(build)
     @build = build
@@ -61,7 +61,6 @@ class DockerBuilderService
     Samson::Hooks.fire(:before_docker_build, tmp_dir, @build, @output)
     execute_build_command(tmp_dir, @build.project.build_command)
   end
-  add_method_tracer :before_docker_build
 
   def build_image(tmp_dir, tag_as_latest:)
     File.write("#{tmp_dir}/REVISION", @build.git_sha)
@@ -83,5 +82,5 @@ class DockerBuilderService
       tmp_dir, @build, @execution.executor, tag_as_latest: tag_as_latest, cache_from: cache
     )
   end
-  add_method_tracer :build_image
+  add_method_tracers :before_docker_build, :build_image
 end
--- a/lib/samson/performance_tracer.rb
+++ b/lib/samson/performance_tracer.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+module Samson
+  module PerformanceTracer
+    # It's used to trace the hook fire event,
+    # We can't use the samson hook to fetch the available plugins.
+    TRACER_PLUGINS = ['SamsonNewRelic', 'SamsonDatadogTracer::APM'].freeze
+
+    class << self
+      def included(clazz)
+        clazz.extend ClassMethods
+      end
+
+      def trace_execution_scoped(scope_name)
+        plugins = TRACER_PLUGINS.map(&:safe_constantize).compact
+        using_plugins plugins, scope_name do
+          yield
+        end
+      end
+
+      def using_plugins(plugins, scope_name, &block)
+        plugins.inject(block) { |inner, plugin| plugin.trace_method_execution_scope(scope_name) { inner } }[]
+      end
+    end
+
+    # Common class methods for Newrelic and Datadog.
+    module ClassMethods
+      def add_method_tracers(*methods)
+        Samson::Hooks.fire(:performance_tracer, self, methods)
+      end
+
+      # TODO: Add asynchronous tracer for Datadog.
+      def add_asynchronous_tracer(method, options)
+        Samson::Hooks.fire(:asynchronous_performance_tracer, self, method, options)
+      end
+    end
+  end
+end
--- a/lib/samson/performance_tracer.rb
+++ b/lib/samson/performance_tracer.rb
@@ -31,8 +31,8 @@ module Samson
 
     # Common class methods for Newrelic and Datadog.
     module ClassMethods
-      def add_method_tracers(*methods)
-        Samson::Hooks.fire(:performance_tracer, self, methods)
+      def add_tracer(method)
+        Samson::Hooks.fire(:performance_tracer, self, method)
       end
 
       # TODO: Add asynchronous tracer for Datadog.
--- a/lib/samson/performance_tracer.rb
+++ b/lib/samson/performance_tracer.rb
@@ -29,7 +29,7 @@ module Samson
       end
     end
 
-    # Common class methods for Newrelic and Datadog.
+    # for Newrelic and Datadog -> for tracer plugins.
     module ClassMethods
       def add_tracer(method)
         Samson::Hooks.fire(:performance_tracer, self, method)
--- a/lib/samson/performance_tracer.rb
+++ b/lib/samson/performance_tracer.rb
@@ -11,14 +11,21 @@ module Samson
       end
 
       def trace_execution_scoped(scope_name)
-        plugins = TRACER_PLUGINS.map(&:safe_constantize).compact
-        using_plugins plugins, scope_name do
+        # Tracing the scope is restricted to avoid into slow startup
+        # Refer Samson::BootCheck
+        if ['staging', 'production'].include?(Rails.env)
+          plugins = TRACER_PLUGINS.map(&:safe_constantize).compact
+          execution = using_plugins plugins, scope_name do
+            yield
+          end
+          execution.call
+        else
           yield
         end
       end
 
       def using_plugins(plugins, scope_name, &block)
-        plugins.inject(block) { |inner, plugin| plugin.trace_method_execution_scope(scope_name) { inner } }[]
+        plugins.inject(block) { |inner, plugin| plugin.trace_method_execution_scope(scope_name) { inner } }
       end
     end
 
--- a/plugins/datadog_tracer/config/initializers/datadog_tracer.rb
+++ b/plugins/datadog_tracer/config/initializers/datadog_tracer.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+if SamsonDatadogTracer.enabled?
+
+  SamsonDatadogTracer::IGNORED_URLS = Set[
+    '/ping',
+    '/streams',
+  ].freeze
+
+  require 'ddtrace'
+  Datadog.configure do |c|
+    # Tracer
+    c.tracer(
+      hostname:                ENV['STATSD_HOST'] || '127.0.0.1',
+      tags: {
+        env:                   ENV['RAILS_ENV'],
+        pod:                   ENV['ZENDESK_POD_ID'],
+        'rails.version':       Rails.version,
+        'ruby.version':        RUBY_VERSION
+      }
+    )
+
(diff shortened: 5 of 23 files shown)


grosser:
I don't like that we now have 2 measuring frameworks ... how about a generic approach like we did for error handling, where we use a common interface and then hook into it ? (can start with just 1 file that implements newrelic+datadog depending on which is active)

also both should track the same methods for consistency ?

dragonfax:
I was told that newrelic sticks its own handlers in there. We don't set them up ourselves. 

But data dog does no such thing, and we have to do the instrumentation ourselves.

In otherwords that nots a common interface to extract.

dragonfax:
@sathishavm is out for a few weeks. I'm taking this over, or assigning it to someone else.

@grosser Response on my above comment? Do we need to abstract this?

grosser:
Would still like to see the common things turned into a hook so we can have newerelic/datadog as plugins instead of making everyone install and use both.

grosser:
can pair on that with whoever takes this over ...

sathishavm:
@grosser It is ready for re-review, please do a review.

Gemfile.lock
```diff
@@ -21,6 +21,12 @@ PATH
     samson_aws_ecr (0.0.0)
       aws-sdk-ecr
 
+PATH
+  remote: plugins/datadog_tracer
+  specs:
+    samson_datadog_tracer (0.0.1)
```
grosser: how about samson_datadog_apm ?

lib/samson/performance_tracer.rb
```diff
@@ -0,0 +1,37 @@
...
+    TRACER_PLUGINS = ['SamsonNewRelic', 'SamsonDatadogTracer::APM'].freeze
+
+    class << self
+      def included(clazz)
+        clazz.extend ClassMethods
+      end
+
+      def trace_execution_scoped(scope_name)
+        plugins = TRACER_PLUGINS.map(&:safe_constantize).compact
+        using_plugins plugins, scope_name do
```
grosser: `using_plugins  a, b do yield` -> using_plugins(a, b, &block)

lib/samson/performance_tracer.rb
```diff
@@ -0,0 +1,37 @@
...
+          yield
+        end
+      end
+
+      def using_plugins(plugins, scope_name, &block)
+        plugins.inject(block) { |inner, plugin| plugin.trace_method_execution_scope(scope_name) { inner } }[]
```
grosser: what is the `[]` for ?

plugins/datadog_tracer/config/initializers/datadog_tracer.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+if SamsonDatadogTracer.enabled?
+
+  SamsonDatadogTracer::IGNORED_URLS = Set[
+    '/ping',
+    '/streams',
```
grosser: streams is dead ... maybe `/cable` ?
... also this setting should be in the base ?

plugins/datadog_tracer/config/initializers/datadog_tracer.rb
```diff
@@ -0,0 +1,44 @@
...
+  ].freeze
+
+  require 'ddtrace'
+  Datadog.configure do |c|
+    # Tracer
+    c.tracer(
+      hostname:                ENV['STATSD_HOST'] || '127.0.0.1',
+      tags: {
+        env:                   ENV['RAILS_ENV'],
+        pod:                   ENV['ZENDESK_POD_ID'],
```
grosser: don't think this is universal / don't think this matters ... can we leave it out ?

plugins/datadog_tracer/config/initializers/datadog_tracer.rb
grosser: nvm about the base comment ...

plugins/datadog_tracer/lib/samson_datadog_tracer/apm.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+module SamsonDatadogTracer
+  module APM
+    class << self
+      def included(clazz)
+        clazz.extend ClassMethods
+      end
+
+      def trace_method_execution_scope(scope_name)
+        if SamsonDatadogTracer.enabled? && ['staging', 'production'].include?(Rails.env)
```
grosser: can enabled check for staging/production ?

plugins/datadog_tracer/lib/samson_datadog_tracer/apm.rb
```diff
@@ -0,0 +1,103 @@
...
+          yield
+        end
+      end
+    end
+
+    module ClassMethods
+      def trace_method(method)
+        return unless SamsonDatadogTracer.enabled?
+
+        @__apm_module ||= begin
```
grosser: don't need the `__` in the name ?

plugins/datadog_tracer/lib/samson_datadog_tracer/samson_plugin.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+module SamsonDatadogTracer
+  class Engine < Rails::Engine
+  end
+
+  def self.enabled?
+    !!ENV['STATSD_TRACER']
```
grosser: would prefer something that is connected to the plugin like `DATADOG_TRACER=1`

plugins/datadog_tracer/lib/samson_datadog_tracer/samson_plugin.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+end
+
+Samson::Hooks.callback :performance_tracer do |klass, methods|
+  if SamsonDatadogTracer.enabled?
+    klass.is_a?(Class) && klass.class_eval do
```
grosser: `klass.is_a?(Class)` is always true ?

.gitignore
```diff
@@ -45,4 +45,4 @@
 
 # .bak files
 *.bak
-/vendor/cache/**/
+/vendor/cache
```
grosser: nice!

app/models/docker_builder_service.rb
```diff
@@ -61,7 +61,6 @@ def before_docker_build(tmp_dir)
     Samson::Hooks.fire(:before_docker_build, tmp_dir, @build, @output)
     execute_build_command(tmp_dir, @build.project.build_command)
   end
-  add_method_tracer :before_docker_build
```
grosser: the calls close to the method were nice to not forget to rename them / to be aware that they are instrumented

Gemfile.lock
sathishavm: The plugin just gives overlay for gem ddtrace, so I prefer to keep this name.

app/models/docker_builder_service.rb
sathishavm: right!!, but if there are multiple methods calls, defining into single trace method call, reduces the execution timing.

grosser:
readability is usually more important then saving 0.001ms :)

grosser:
🤗

sathishavm:
Hmmm.. Ok, Given important to readability, changed the needed :v:

lib/samson/performance_tracer.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+module Samson
+  module PerformanceTracer
+    # It's used to trace the hook fire event,
+    # We can't use the samson hook to fetch the available plugins.
+    TRACER_PLUGINS = ['SamsonNewRelic', 'SamsonDatadogTracer::APM'].freeze
```
grosser: having the main code know about what plugins are available is kinda backwards ... should call a hook to get the list or do something like `plugins.map(&:tracer)`

grosser: also avoids doing the constantize

lib/samson/performance_tracer.rb
```diff
@@ -0,0 +1,44 @@
...
+          yield
+        end
+      end
+
+      def using_plugins(plugins, scope_name, &block)
+        plugins.inject(block) { |inner, plugin| plugin.trace_method_execution_scope(scope_name) { inner } }
+      end
+    end
+
+    # Common class methods for Newrelic and Datadog.
```
grosser: `for Newrelic and Datadog` -> `for tracer plugins`

plugins/datadog_tracer/config/initializers/datadog_tracer.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+if SamsonDatadogTracer.enabled?
+
+  SamsonDatadogTracer::IGNORED_URLS = Set[
```
grosser: should be in the base

plugins/datadog_tracer/lib/samson_datadog_tracer/apm.rb
```diff
@@ -0,0 +1,103 @@
...
+            yield
+          end
+        else
+          yield
+        end
+      end
+    end
+
+    module ClassMethods
+      def trace_method(method)
```
grosser: this looks kinda funky ... can you add a code comment on what is going on here ?

plugins/datadog_tracer/lib/samson_datadog_tracer/apm.rb
```diff
@@ -0,0 +1,103 @@
...
+        @apm_module.module_eval do
+          _wrap_method(method, klass)
+        end
+      end
+    end
+
+    module Helpers
+      class << self
+        def sanitize_name(name)
+          name.to_s.tr_s('^a-zA-Z0-9', '_')
```
grosser: maybe better to reuse the builtin `.parameterize.tr('-', '_')`

plugins/datadog_tracer/lib/samson_datadog_tracer/apm.rb
```diff
@@ -0,0 +1,103 @@
...
+        end
+
+        def tracer_method_name(method_name)
+          "#{sanitize_name(method_name)}_with_apm_tracer"
+        end
+
+        def untracer_method_name(method_name)
+          "#{sanitize_name(method_name)}_with_apm_untracer"
```
grosser: usually the aliases are named `with` and `without`

plugins/datadog_tracer/lib/samson_datadog_tracer/apm.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      def _set_visibility(method, visibility)
+        method_name = Helpers.tracer_method_name(method)
+        case visibility
+        when :protected
+          protected(method_name)
+        when :private
+          private(method_name)
+        else
+          public(method_name)
```
grosser: isn't that the default ?

plugins/datadog_tracer/lib/samson_datadog_tracer/apm.rb
```diff
@@ -0,0 +1,103 @@
...
+        end
+
+        def untracer_method_name(method_name)
+          "#{sanitize_name(method_name)}_with_apm_untracer"
+        end
+      end
+
+      private
+
+      def _wrap_method(method, klass)
```
grosser: why is the `_` needed ?

plugins/datadog_tracer/lib/samson_datadog_tracer/samson_plugin.rb
```diff
@@ -0,0 +1,24 @@
...
+end
+
+Samson::Hooks.callback :performance_tracer do |klass, method|
+  if SamsonDatadogTracer.enabled?
+    klass.class_eval do
+      include SamsonDatadogTracer::APM
+      helper = SamsonDatadogTracer::APM::Helpers
+      trace_method method
+
+      if method_defined?(method) || private_method_defined?(method)
```
grosser: why not just call alias and let it blow up if method is not defined ?

grosser: also ideally don't class_eval but prepend a module that calls super

lib/samson/performance_tracer.rb
sathishavm: I have commented that we can' use hook, because this is used to trace the hook, this will create a endless loop.

grosser:
ah ok, already a step forward anyway, maybe we find a way to do it nicer
later

plugins/datadog_tracer/config/initializers/datadog_tracer.rb
sathishavm: ok, maybe later!!

plugins/datadog_tracer/lib/samson_datadog_tracer/apm.rb
sathishavm: done

sathishavm:
Can I get +1?

plugins/datadog_tracer/test/samson_datadog_tracer/apm_test.rb
```diff
@@ -0,0 +1,138 @@
...
+
+  describe "skips APM trace methods" do
+    let(:klass) { TestKlass1.new }
+    it "skips tracker when apm is not enabled" do
+      Datadog.expects(:tracer).never
+      klass.send(:pub_method)
+    end
+  end
+
+  ENV.store("DATADOG_TRACER", "1")
```
grosser: ... you ran into a validation that told you why this should not be done and you decided the best cause of action was to hack around it instead of asking for help 🤷‍♂️

plugins/datadog_tracer/test/samson_datadog_tracer/apm_test.rb
```diff
@@ -0,0 +1,138 @@
...
+    end
+
+    it "raises with NoMethodError when undefined method call" do
+      assert_raise NoMethodError do
+        apm.send(:not_method)
+      end
+    end
+
+    it "preserves method visibility" do
+      assert apm.class.public_method_defined?(:pub_method)
```
grosser: FYI using a simple assert give unhelpful error message when it fails, prefer `foo.methods.must_include :bar` so it fails and shows which methods existed

plugins/datadog_tracer/test/samson_datadog_tracer/apm_test.rb
```diff
@@ -0,0 +1,138 @@
...
+    trace_method :pub_method
+    trace_method :pri_method
+    trace_method :pro_method
+    trace_method :not_method
+    alias_method :pub_method_with_apm_untracer, :pub_method
+  end
+
+  describe ".trace_method" do
+    let(:apm) { TestKlass2.new }
+    Datadog.expects(:tracer).returns(Datadog.tracer)
```
grosser: never do expects outside of it blocks

grosser: cleaning this up atm, so don't make a PR ;)

plugins/datadog_tracer/test/samson_datadog_tracer/samson_plugin_test.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+require_relative "../test_helper"
+
+SingleCov.covered!
+
+describe SamsonDatadogTracer do
+  describe ".enabled?" do
+    before(:all) do
+      ENV.delete('DATADOG_TRACER')
+    end
```
grosser: same here ... global change and I also doubt `before :all` works

PR #2887, opened 2018-08-21
Install docker into Dockerfile

adammw:
Install docker into Dockerfile, so that builds/other docker commands will work.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -2,7 +2,7 @@
 class Build < ActiveRecord::Base
   SHA1_REGEX = /\A[0-9a-f]{40}\Z/i
   SHA256_REGEX = /\A(sha256:)?[0-9a-f]{64}\Z/i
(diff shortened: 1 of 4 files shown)


app/models/build.rb
```diff
@@ -2,7 +2,7 @@
 class Build < ActiveRecord::Base
   SHA1_REGEX = /\A[0-9a-f]{40}\Z/i
   SHA256_REGEX = /\A(sha256:)?[0-9a-f]{64}\Z/i
-  DIGEST_REGEX = /\A[\w.-]+[\w\/-]*@sha256:[0-9a-f]{64}\Z/i
+  DIGEST_REGEX = /\A[\w.-]+[\w.-]*(:\d+)?[\w\/-]*@sha256:[0-9a-f]{64}\Z/i
```
grosser: a test for "it allows shas with port" or something like that ?

adammw: added

PR #2888, opened 2018-08-21
Add JSON response to env groups controller#index

davidreuss:
This gives a quick overview of what keys any group might contain, which
is useful for figuring out exactly what environment variables need to be
present in a given environment.

### DIFF
--- a/plugins/env/app/controllers/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/environment_variable_groups_controller.rb
@@ -4,7 +4,18 @@ class EnvironmentVariableGroupsController < ApplicationController
   before_action :group, only: [:show]
 
   def index
-    @groups = EnvironmentVariableGroup.all
+    @groups = EnvironmentVariableGroup.all.includes(:environment_variables)
+    respond_to do |format|
+      format.html
+      format.json do
+        groups = @groups.map do |group|
+          json = group.as_json
+          json['variables'] = group.environment_variables.sort_by(&:id).map(&:name).uniq
+          json
+        end
+        render json: {groups: groups}
+      end
+    end
   end
 
   def new
--- a/plugins/env/app/controllers/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/environment_variable_groups_controller.rb
@@ -8,12 +8,7 @@ class EnvironmentVariableGroupsController < ApplicationController
     respond_to do |format|
       format.html
       format.json do
-        groups = @groups.map do |group|
-          json = group.as_json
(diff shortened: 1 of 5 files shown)


plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -4,7 +4,18 @@ class EnvironmentVariableGroupsController < ApplicationController
...
 
   def index
-    @groups = EnvironmentVariableGroup.all
+    @groups = EnvironmentVariableGroup.all.includes(:environment_variables)
+    respond_to do |format|
+      format.html
+      format.json do
+        groups = @groups.map do |group|
+          json = group.as_json
+          json['variables'] = group.environment_variables.sort_by(&:id).map(&:name).uniq
```
grosser: - prefer `variables` sounds like objects, maybe use `variable_names` ?
 - copy-pasted from the view ... turn into a method on the model and reuse ?
 - prefer to add to as_json and then use `render_as_json :environment_variable_groups, @groups` so we get consistently formatted replies

plugins/env/test/controllers/environment_variable_groups_controller_test.rb
```diff
@@ -79,6 +79,27 @@ def self.it_destroys
       end
     end
 
+    describe "a json GET to #index" do
+      it "succeeds" do
+        get :index, format: :json
+        assert_response :success
+        json_response = JSON.parse response.body
+        json_response.keys.must_include "groups"
+      end
```
grosser: don't need that test ?
... then can simplify to `it "renders json"` without a describe

grosser: also should be inside of `describe "#index"`

plugins/env/app/controllers/environment_variable_groups_controller.rb
grosser:
```Ruby
def variable_names
  ...
end

def as_json
  super(methods: [:variable_names])
end
```

plugins/env/app/models/environment_variable_group.rb
```diff
@@ -16,4 +16,12 @@ class EnvironmentVariableGroup < ActiveRecord::Base
   has_many :projects, through: :project_environment_variable_groups
 
   validates :name, presence: true
+
+  def variable_names
```
ragurney: Maybe add a unit test for this?

davidreuss:
Had to restart a few of the integration tests in travis, as they randomly timeout.

Otherwise i think this is good to go now. The `#variable_names` test isn't exhaustive, but good enough for now i'd say.

PR #2889, opened 2018-08-21
Make the docker-compose env work as a dev environment

russell:
* Mount the app directory into the container
* Setup the env files on start
* Seed if the sqllite db file is missing

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

PR #2890, opened 2018-08-21
Display the Github status of a release on the overview page

dasch:
Adds an icon indicating the Github Status of each release's commit, as well as a full list of statuses with their descriptions on the release page.

### Tasks
- [ ] :+1: from team
- [ ] Caching and invalidation

### Risks
- Level: Med

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -146,10 +146,13 @@ module ApplicationHelper
   end
 
   def icon_tag(type, options = {})
-    # Prepend the CSS classes to the existing `css` option if present.
-    css_class = ["glyphicon glyphicon-#{type}", options.fetch(:class, "")].join(" ")
+    css_classes = "glyphicon glyphicon-#{type}"
 
-    content_tag :i, '', options.merge(class: css_class)
+    if klass = options[:class]
+      css_classes += " #{klass}"
+    end
+
+    content_tag :i, '', options.merge(class: css_classes)
   end
 
   def link_to_delete(path, options = {})
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -146,7 +146,10 @@ module ApplicationHelper
   end
 
   def icon_tag(type, options = {})
-    content_tag :i, '', options.merge(class: "glyphicon glyphicon-#{type}")
+    # Prepend the CSS classes to the existing `css` option if present.
+    css_class = ["glyphicon glyphicon-#{type}", options.fetch(:class, "")].join(" ")
+
+    content_tag :i, '', options.merge(class: css_class)
   end
 
   def link_to_delete(path, options = {})
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -8,6 +8,17 @@ module ReleasesHelper
     )
   end
 
+  def status_glyphicon(status_state)
+    icon, text = case status_state
+           when "success" then ["ok", "success"]
+           when "failure" then ["remove", "danger"]
+           when "missing" then ["minus", "muted"]
+           when "pending" then ["hourglass", "primary"]
+           end
+
+    %(<span class="glyphicon glyphicon-#{icon} text-#{text}" aria-hidden="true"></span>).html_safe
+  end
+
   def link_to_deploy_stage(stage, release)
     deploy_params = {reference: release.version}
 
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -26,7 +26,7 @@ module ReleasesHelper
     icon = STATUS_ICONS.fetch(status_state)
     text = STATUS_TEXT_LABELS.fetch(status_state)
 
-    %(<span class="glyphicon glyphicon-#{icon} text-#{text}" aria-hidden="true"></span>).html_safe
+    icon_tag icon, class: "text-#{text}"
   end
 
   def link_to_deploy_stage(stage, release)
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -22,7 +22,7 @@ module ReleasesHelper
     )
   end
 
-  def status_glyphicon(status_state)
+  def github_commit_status_icon(status_state)
     icon = STATUS_ICONS.fetch(status_state)
     text = STATUS_TEXT_LABELS.fetch(status_state)
     title = "Github status: #{status_state}"
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -1,5 +1,19 @@
 # frozen_string_literal: true
 module ReleasesHelper
+  STATUS_ICONS = {
+    "success" => "ok",
+    "failure" => "remove",
+    "missing" => "minus",
+    "pending" => "hourglass"
+  }
+
+  STATUS_TEXT_LABELS = {
+    "success" => "success",
+    "failure" => "danger",
+    "missing" => "muted",
+    "pending" => "primary"
+  }
+
   def release_label(project, release)
     link_to(
       release.version,
@@ -9,12 +23,8 @@ module ReleasesHelper
   end
 
   def status_glyphicon(status_state)
-    icon, text = case status_state
-           when "success" then ["ok", "success"]
-           when "failure" then ["remove", "danger"]
-           when "missing" then ["minus", "muted"]
-           when "pending" then ["hourglass", "primary"]
-           end
+    icon = STATUS_ICONS.fetch(status_state)
+    text = STATUS_TEXT_LABELS.fetch(status_state)
 
     %(<span class="glyphicon glyphicon-#{icon} text-#{text}" aria-hidden="true"></span>).html_safe
   end
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -5,14 +5,14 @@ module ReleasesHelper
     "failure" => "remove",
     "missing" => "minus",
     "pending" => "hourglass"
-  }
+  }.freeze
 
   STATUS_TEXT_LABELS = {
     "success" => "success",
     "failure" => "danger",
     "missing" => "muted",
     "pending" => "primary"
-  }
+  }.freeze
 
   def release_label(project, release)
     link_to(
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -25,8 +25,9 @@ module ReleasesHelper
   def status_glyphicon(status_state)
     icon = STATUS_ICONS.fetch(status_state)
     text = STATUS_TEXT_LABELS.fetch(status_state)
+    title = "Github status: #{status_state}"
 
-    icon_tag icon, class: "text-#{text}"
+    icon_tag icon, class: "text-#{text}", 'data-toggle': "tooltip", 'data-placement': "right", title: title
   end
 
   def link_to_deploy_stage(stage, release)
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
 module ReleasesHelper
-  STATUS_ICONS = {
+  GITHUB_STATUS_ICONS = {
     "success" => "ok",
     "failure" => "remove",
     "missing" => "minus",
     "pending" => "hourglass"
   }.freeze
 
-  STATUS_TEXT_LABELS = {
+  GITHUB_STATUS_TEXT_LABELS = {
     "success" => "success",
     "failure" => "danger",
     "missing" => "muted",
@@ -23,8 +23,8 @@ module ReleasesHelper
   end
 
   def github_commit_status_icon(status_state)
-    icon = STATUS_ICONS.fetch(status_state)
-    text = STATUS_TEXT_LABELS.fetch(status_state)
+    icon = GITHUB_STATUS_ICONS.fetch(status_state)
+    text = GITHUB_STATUS_TEXT_LABELS.fetch(status_state)
     title = "Github status: #{status_state}"
 
     icon_tag icon, class: "text-#{text}", 'data-toggle': "tooltip", 'data-placement': "right", title: title
--- a/app/models/github_status.rb
+++ b/app/models/github_status.rb
@@ -3,6 +3,14 @@ class GithubStatus
     def success?
       latest_status.state == "success"
     end
+
+    def failure?
+      latest_status.state == "failure"
+    end
+
+    def pending?
+      latest_status.state == "pending"
+    end
   end
 
   def initialize(repo, ref, github: GITHUB)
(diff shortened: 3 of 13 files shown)


dasch:
@grosser can you take a look at this? Ignore the first commit, that was done in order to be able to run things locally with docker-compose and will be removed before merging.

app/views/releases/_release_status.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if status.success? %>
+  <span class="glyphicon glyphicon-ok text-success" aria-hidden="true"></span>
```
ragurney: Could use `icon_tag` helper here as well.

app/models/github_status.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def pending?
+    statuses.any?(&:pending?)
+  end
+
+  def missing?
+    statuses.none?
+  end
+
+  def statuses
```
ragurney: Maybe cache these on the object, otherwise `pending?`, `missing?` etc. will make this call every time in `release_status.html.erb`.  Something like:
```ruby
def statuses 
  @statuses ||= begin
    ...
  end
end

File: app/models/github_status.rb [hunk 3]
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def success?
+    statuses.all?(&:success?)
+  end
+
+  def statuses
+    response = @github.combined_status(@repo, @ref)
+
+    statuses = response.statuses
```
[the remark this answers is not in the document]
dasch: :+1:

app/helpers/releases_helper.rb
```diff
@@ -8,6 +8,17 @@ def release_label(project, release)
     )
   end
 
+  def status_glyphicon(status_state)
+    icon, text = case status_state
```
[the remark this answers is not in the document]
dasch: :+1:

app/helpers/releases_helper.rb
```diff
@@ -8,6 +8,17 @@ def release_label(project, release)
...
+           when "success" then ["ok", "success"]
+           when "failure" then ["remove", "danger"]
+           when "missing" then ["minus", "muted"]
+           when "pending" then ["hourglass", "primary"]
+           end
+
+    %(<span class="glyphicon glyphicon-#{icon} text-#{text}" aria-hidden="true"></span>).html_safe
```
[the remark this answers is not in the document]
dasch: The problem is that `icon_tag` doesn't seem to support setting additional classes, since it overwrites the `class` attribute. I frankly don't think it's necessary, since this is already declarative Bootstrap code.

app/models/github_status.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    return [] if statuses.nil?
+
+    statuses.group_by {|status| status.context }
```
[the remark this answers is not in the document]
dasch: Sure, I'll just follow the local practice. I usually use curly brackets when I care about the result rather than the side effect.

app/views/releases/_release_status.html.erb
dasch: If that method was amended to allow passing in additional tags, sure.

app/models/github_status.rb
dasch: :+1:

dasch:
@ragurney care to take another look?

dasch:
I've changed `icon_tag` so that it accepts custom `class` options.

.gitignore
```diff
@@ -46,3 +46,4 @@
 # .bak files
 *.bak
 /vendor/cache/**/
+/db/development.sqlite
```
grosser: we already have ` /db/*.sqlite3`

app/helpers/application_helper.rb
```diff
@@ -146,7 +146,10 @@ def audited_classes
   end
 
   def icon_tag(type, options = {})
-    content_tag :i, '', options.merge(class: "glyphicon glyphicon-#{type}")
+    # Prepend the CSS classes to the existing `css` option if present.
+    css_class = ["glyphicon glyphicon-#{type}", options.fetch(:class, "")].join(" ")
```
grosser: feels kinda over-engineered / un-precise ...

```Ruby
css_classes = "glyphicon glyphicon-#{type}"
if klass = options[:class]
  css_classes << " #{klass}"
end
```

app/models/github_status.rb
```diff
@@ -0,0 +1,73 @@
...
+      status_response.statuses.group_by(&:context).map do |context, statuses|
+        Status.new(context, statuses.max_by { |status| status.created_at.to_i })
+      end
+    end
+  end
+
+  private
+
+  def status_response
+    @status_response ||= @github.combined_status(@repo, @ref)
```
grosser: won't cache on error ... prefer 
```Ruby
return @status_response if defined?(@status_response)
@status_response = 
  begin
    @github.combined_status(@repo, @ref)
  rescue Octokit::Error
    nil
  end

```

app/models/github_status.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def initialize(repo, ref, github: GITHUB)
+    @github = github
+    @repo = repo
+    @ref = ref
+    @state = nil
+  end
+
+  def state
+    status_response ? status_response.state : "missing"
```
grosser: here no status_response means it's missing ... and then below it is from empty statuses .. seems kinda confusing

app/views/releases/index.html.erb
```diff
@@ -19,6 +19,7 @@
         <thead>
           <tr>
             <th></th>
+            <th>Status</th>
```
grosser: unclear that this is github status ...should say github status or have an octopus icon or a hover

script/docker_dev_server
```diff
@@ -0,0 +1,18 @@
...
+    ./bin/decode_dot_env .env
+fi
+
+if [ -f db/development.sqlite ]; then
+    bundle exec rake db:migrate
+else
+    bundle exec rake db:setup
+fi
+
+exec bundle exec puma -C config/puma.rb
```
grosser: docker things should not be part of this PR

app/models/github_status.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+class GithubStatus
```
grosser: we already have commit_status ... this feels like the same code all over again ... can we unify ?

.gitignore
dasch: Yeah, that's from #2889.

app/models/github_status.rb
dasch: Added memoization of the response on error.

app/models/github_status.rb
dasch: :+1:

app/models/github_status.rb
dasch: That would be a larger refactoring, and I mostly want to fix this UX problem right now...

app/views/releases/index.html.erb
dasch: I was actually considering removing the `th`, since it takes up more space than the icons themselves. How about a hover text saying e.g. "Github status: failed"?

script/docker_dev_server
dasch: Will remove after final review, otherwise I can't get the code to run...

app/models/github_status.rb
dasch: It seems that `CommitStatus` is deeply linked to stages...

app/models/github_status.rb
dasch: Changed the code so that we don't try to second-guess the API response's state.

app/models/github_status.rb
grosser: can commit-status use this then ?

app/helpers/releases_helper.rb
```diff
@@ -8,6 +22,14 @@ def release_label(project, release)
     )
   end
 
+  def status_glyphicon(status_state)
```
grosser: prefix method+constants with `github_` or `commit_` since we have deploy status which is different

grosser: also just `icon` since it does not matter what library we use and the other helper is called `icon_tag`

app/models/github_status.rb
```diff
@@ -0,0 +1,75 @@
...
+      status_response.statuses.group_by(&:context).map do |context, statuses|
+        Status.new(context, statuses.max_by { |status| status.created_at.to_i })
+      end
+    end
+  end
+
+  private
+
+  def status_response
+    @status_response ||= begin
```
grosser: still does not cache on failure since `nil ||=` would execute again

app/views/releases/show.html.erb
```diff
@@ -16,6 +16,13 @@
   <label>Created</label>
   <%= relative_time(@release.created_at) %>
 
+  <legend>Github Status</legend>
```
grosser: maybe name this commit-status so it's not github specific and can serve multiple backends eventually ...

grosser:
needs some `Rails.cache` love or this will be a neat N+1 on the release page :)
... also make sure we use some sane timeouts since this would block rendering that page

dasch:
@grosser I would assume Octokit has sensible timeouts already?

I wanted to ask about caching – would it be possible to have the Github event webhook invalidate the cache? We'd want short TTLs for new releases but long ones for old releases, which is tricky to do – so maybe just a long TTL and explicit invalidation from `status` events?

grosser:
we don't have that kind of event ... how about deciding how long to cache depending on if the statuses  are completed (success/failed) ?
(fetch ... fail and retrieve ... decide how long to cache)

dasch:
I guess is the status is pending we just don't cache?

grosser:
maybe cache for a minute or so ... or just don't ... there should be few commits on the release page that are pending ...

grosser:
there is a `github is down` check somewhere, would be good to use that as short-circuit too

grosser:
it's `github_ok?`

dasch:
How about this?

dasch:
@grosser as in, not trying to call the APIs in that case? For the cached entries it should be fine...

grosser:
yes!

app/models/github_status.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    cache_key = [self.class.to_s, @repo, @ref]
+
+    @status_response = @cache.read(cache_key)
+
+    # Fetch the data if the cache returned nil.
+    @status_response ||= begin
+      @github.combined_status(@repo, @ref)
+    rescue Octokit::Error
+      nil
```
grosser: should either cache that nil or return early ?

dasch: Ah yes.

dasch:
@grosser happy with the `github_ok?` code now?

app/models/github_status.rb
dasch: I've changed the code so nils aren't cached – that should do the trick.

app/models/github_status.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    # Don't cache pending statuses, since we expect updates soon.
+    unless @status_response.nil? || @status_response.state == "pending"
```
grosser: prefer to avoid unless + binary logic
`if @status_response&.state == "pending"`

app/helpers/releases_helper.rb
grosser: ^

app/models/github_status.rb
dasch: I can give it a try.

dasch:
@grosser one more go?

dasch:
I simplified GithubStatus btw, I think it's much better now that the network code is done procedurally.

app/helpers/releases_helper.rb
```diff
@@ -1,5 +1,19 @@
 # frozen_string_literal: true
 module ReleasesHelper
+  STATUS_ICONS = {
```
grosser: also prefix with GITHUB_ ?

app/models/github_status.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+class GithubStatus
+  Status = Struct.new(:context, :latest_status) do
+    def state
+      latest_status.state
```
grosser: latest_status->status ?

grosser: hmm I guess that could get confusing with `state` 🤷‍♂️

app/models/github_status.rb
```diff
@@ -0,0 +1,78 @@
...
+      GITHUB.combined_status(repo, ref)
+    rescue Octokit::Error
+      nil
+    end
+
+    # Fall back to a "missing" status.
+    return new("missing", []) if response.nil?
+
+    # Don't cache pending statuses, since we expect updates soon.
+    unless response&.state == "pending"
```
grosser: don't need the `&.`since nil is handled above

test/models/release_test.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 2
+SingleCov.covered! uncovered: 3
```
grosser: add test for github_status ?

app/models/github_status.rb
dasch: Yeah.

app/helpers/releases_helper.rb
dasch: :+1:

app/models/github_status.rb
dasch: :+1:

test/models/release_test.rb
dasch: It doesn't contain any logic, and is very unlikely to break. I think a test is superfluous, unless you're adamant about it.

dasch: The controller test already exercises the code path.

dasch:
@grosser I've included most of your feedback – I'd like to merge this thing, as I can't keep spending time on it. It significantly improves the workflow of release-based projects, so let's get it in.

app/views/releases/show.html.erb
```diff
@@ -16,6 +16,13 @@
   <label>Created</label>
   <%= relative_time(@release.created_at) %>
 
+  <legend>Commit Status</legend>
+  <p>
+    <% @release.github_status.statuses.each do |status| %>
```
grosser: inside of `github_ok` too ?

grosser:
@ragurney another look ?

test/models/release_test.rb
grosser: hmmm I'd still add the test since we are trying to keep everything fully unit-tested.

dasch:
:heart:

app/views/releases/show.html.erb
dasch: I don't think this page is critical to the workflow... let's see if it becomes a problem.

grosser:
screen shot 2018-09-10 at 11 00 42 am

just saw pending since 6 months ... either a bug or we might want to cache these too after ~1 hour or so

PR #2893, opened 2018-08-24
Add fossa scan to ensure we can release under apache v2

plonergan:
TODO:
 - [x] get the scan green
 - [x] whitelist things we want to ignore and document how to do that

/cc @zendesk/samson


grosser:
what is this and why is it needed ?

.travis.yml
```diff
@@ -4,7 +4,7 @@ cache:
   directories:
     - tmp/rubocop_cache
     - node_modules
-sudo: false
+sudo: required
```
grosser: only switch for the new job

grosser:
```
/tmp/install-fossa-cli/fossa test
No output has been received in the last 10m0s, 
```
... so ask the fossa guys how long scans usually take ...
we can add some kind of progressbar to the `test` to make it not crash, but if it takes too long than that would slow down our CI :(

grosser:
need https://github.com/fossas/fossa-cli/issues/247 to be resolved first :(

grosser:
@ragurney looks good ?

PR #2894, opened 2018-08-24
Ignore `vendor/cache/`

dasch:
The other rule didn't work for me.

/cc @grosser 

### Risks
- Level: low.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@
 # .bak files
 *.bak
 /vendor/cache/**/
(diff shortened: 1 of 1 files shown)


grosser:
this was here to ignore git reporsitories ... but now we don't do any caching, so I'd just remove it ...
why do you have a vendor/cache directory ?

dasch:
I guess at some point I've bundled with that option...

dasch:
:+1:

PR #2895, opened 2018-08-24
Show code authors for each release

dasch:
Every time I go to deploy my project, I need to answer these two questions:

1) Which of the releases contains my code?
2) If there are other undeployed releases before mine, who are the authors and reviewers of those? I need to get an OK from them before I can deploy.

This PR solves both problems by displaying the avatars of the code authors and reviewers for each release. You can hover over the avatars to get the Github name of the user.

In order to avoid extreme cases with tons of people on the list breaking the UI, there's a maximum of 8 people displayed. That should cover all normal cases.

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Med.

### DIFF
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -39,4 +39,13 @@ module ReleasesHelper
       link_to stage.name, project_stage_deploys_path(@project, stage, deploy: deploy_params), method: :post
     end
   end
+
+  def github_user_avatar(user)
+    image_tag user.avatar_url,
+      title: user.login,
+      class: "gravatar",
+      style: "border-radius: .2em",
+      width: 20,
+      height: 20
+  end
 end
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -40,11 +40,10 @@ module ReleasesHelper
     end
   end
 
-  def github_user_avatar(user)
-    image_tag user.avatar_url,
-      title: user.login,
-      class: "gravatar",
-      style: "border-radius: .2em",
+  def github_user_avatar(github_user)
+    image_tag github_user.avatar_url,
+      title: github_user.login,
+      class: "gravatar github-user-avatar",
       width: 20,
       height: 20
   end
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -39,12 +39,4 @@ module ReleasesHelper
       link_to stage.name, project_stage_deploys_path(@project, stage, deploy: deploy_params), method: :post
     end
   end
-
-  def github_user_avatar(github_user)
-    image_tag github_user.avatar_url,
-      title: github_user.login,
-      class: "gravatar github-user-avatar",
-      width: 20,
-      height: 20
-  end
 end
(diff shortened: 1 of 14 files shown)


app/helpers/releases_helper.rb
```diff
@@ -17,4 +17,13 @@ def link_to_deploy_stage(stage, release)
       link_to stage.name, project_stage_deploys_path(@project, stage, deploy: deploy_params), method: :post
     end
   end
+
+  def github_user_avatar(user)
```
grosser: maybe say `github_user` so it's clear it's not a database user

app/helpers/releases_helper.rb
```diff
@@ -17,4 +17,13 @@ def link_to_deploy_stage(stage, release)
...
+    image_tag user.avatar_url,
+      title: user.login,
+      class: "gravatar",
+      style: "border-radius: .2em",
```
grosser: use css for that ?

app/models/changeset.rb
```diff
@@ -54,6 +54,19 @@ def author_names
...
 
+  def reviewers
+    @reviewers ||= begin
+      cache_key = [self.class, "reviewers", repo, previous_commit, commit]
+
+      Rails.cache.fetch(cache_key) do
+        pull_requests.
+          flat_map { |pr| GITHUB.pull_request_reviews(repo, pr.number) }.
+          map(&:user).
+          uniq
```
ragurney: Looks like `uniq` is still allowing duplicate users. I think we might want to make this `uniq(&:login)`.

dasch:
@ragurney @grosser I agree that there's a difference between reviewers and authors – I mainly wanted a list of people to contact regarding the release, and reviewers (in my mind) are responsible for the code they review. In order to expedite this I'll remove the reviewers from the list – we can revisit at a later time.

dasch:
@ragurney @grosser I've pushed a new version, care to review again?

app/helpers/releases_helper.rb
```diff
@@ -39,4 +39,12 @@ def link_to_deploy_stage(stage, release)
       link_to stage.name, project_stage_deploys_path(@project, stage, deploy: deploy_params), method: :post
     end
   end
+
+  def github_user_avatar(github_user)
```
grosser: reuse in `app/views/changeset/_commits.html.erb:9``app/views/changelogs/show.html.erb:33` so we get consistent images ... maybe move the 20x20 into css too

dasch: :+1:

It's nice to have the size in the HTML, since it causes less jumping around. It also lets us parameterize it so that it matches the avatar size, since that's not the same in all cases.

test/models/release_test.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 3
+SingleCov.covered! uncovered: 4
```
ragurney: Could you add a test for `code_authors`?

dasch: I'd rather just remove it then. There's 0 logic in it.

dasch:
:D

PR #2896, opened 2018-08-24
Move manual release button to the bottom of the page

dasch:
Currently, the "create release manually" button takes up screen real estate at the top of the page. However, this button should hardly ever be used.

This PR:

1) Uses a separate, non-right-aligned button in case there are no releases.
2) Moves the normal button to the bottom right, next to the pagination controls.

 
/cc @zendesk/samson

### Tasks
 - [x] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/views/releases/index.html.erb
+++ b/app/views/releases/index.html.erb
@@ -36,6 +36,10 @@
         <p>No release created from release branch <%= @project.release_branch %>.</p>
         <p>Add a webhook to samson from your CI to trigger releases on green builds or from your Code host to get releases on merge.</p>
         <p>Samsons Github user needs write permisson to the repo to create tags.</p>
+
+        <p>
+          <%= link_to "Create release manually", new_project_release_path(@project), class: "btn btn-default" %>
+        </p>
       </div>
     <% end %>
   </section>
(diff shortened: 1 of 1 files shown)


app/views/releases/index.html.erb
```diff
@@ -31,11 +27,19 @@
       </table>
 
       <%= paginate @pagy %>
+
+      <p class="pull-right">
```
ragurney: I know it was there beforehand, but would a `div` tag might be more semantically correct here?

ragurney: Sure, I was just pointing it out ¯\\_(ツ)_/¯

app/views/releases/index.html.erb
```diff
@@ -31,11 +27,19 @@
...
     <% elsif @project.release_branch.present? %>
       <div class="alert alert-warning">
         <p>No release created from release branch <%= @project.release_branch %>.</p>
         <p>Add a webhook to samson from your CI to trigger releases on green builds or from your Code host to get releases on merge.</p>
         <p>Samsons Github user needs write permisson to the repo to create tags.</p>
+
+        <p>
+          <%= link_to "Create release manually", new_project_release_path(@project), class: "btn btn-default" %>
+        </p>
       </div>
```
grosser: how about something like `<% create_button = capture do %>` and then reuse that instead of copying the button ?

dasch: I honestly think that would be a lot more confusing than just having two copies. This isn't business logic, it's declarative UI code...

PR #2897, opened 2018-08-24
Render past successful deploys on the releases overview

dasch:
It's valuable to be able to see at a glance whether a given release has been successfully deployed to a stage in the past, e.g. when you want to know if a build stage has completed.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


app/helpers/application_helper.rb
```diff
@@ -320,9 +320,20 @@ def link_to_chart(name, values)
   def deployed_or_running_list(stages, reference)
     html = "".html_safe
     stages.each do |stage|
-      next unless deploy = stage.deployed_or_running_deploy
-      next unless deploy.references?(reference)
-      label = (deploy.active? ? "label-warning" : "label-success")
+      # The first deploy is the most recent one.
+      next unless deploy = stage.deploys.where(reference: reference).first
```
ragurney: Instead of `next` and doing n where queries, would it be better to do something like:
```ruby
deploys_hash = Deploy.joins(:stage).where(stage_id: stages.pluck(:id)).where(reference: 'master').group_by(&:stage_id)
stages.each do |stage|
  next unless deploy = deploys_hash[stage.id]&.first
  ...
end
```
?
```
samson(dev)> puts Benchmark.measure { deployed_or_running_list(stages, reference)  }
  Deploy Load (0.6ms)  SELECT  `deploys`.* FROM `deploys` WHERE `deploys`.`deleted_at` IS NULL AND `deploys`.`stage_id` = 4 AND `deploys`.`reference` = 'master' ORDER BY `deploys`.`id` DESC LIMIT 1
  Deploy Load (0.5ms)  SELECT  `deploys`.* FROM `deploys` WHERE `deploys`.`deleted_at` IS NULL AND `deploys`.`stage_id` = 2 AND `deploys`.`reference` = 'master' ORDER BY `deploys`.`id` DESC LIMIT 1
  Deploy Load (0.4ms)  SELECT  `deploys`.* FROM `deploys` WHERE `deploys`.`deleted_at` IS NULL AND `deploys`.`stage_id` = 3 AND `deploys`.`reference` = 'master' ORDER BY `deploys`.`id` DESC LIMIT 1
  Deploy Load (0.4ms)  SELECT  `deploys`.* FROM `deploys` WHERE `deploys`.`deleted_at` IS NULL AND `deploys`.`stage_id` = 1 AND `deploys`.`reference` = 'master' ORDER BY `deploys`.`id` DESC LIMIT 1
  Deploy Load (0.4ms)  SELECT  `deploys`.* FROM `deploys` WHERE `deploys`.`deleted_at` IS NULL AND `deploys`.`stage_id` = 5 AND `deploys`.`reference` = 'master' ORDER BY `deploys`.`id` DESC LIMIT 1
  Deploy Load (0.3ms)  SELECT  `deploys`.* FROM `deploys` WHERE `deploys`.`deleted_at` IS NULL AND `deploys`.`stage_id` = 6 AND `deploys`.`reference` = 'master' ORDER BY `deploys`.`id` DESC LIMIT 1
  Deploy Load (0.4ms)  SELECT  `deploys`.* FROM `deploys` WHERE `deploys`.`deleted_at` IS NULL AND `deploys`.`stage_id` = 7 AND `deploys`.`reference` = 'master' ORDER BY `deploys`.`id` DESC LIMIT 1
  0.005582   0.000476   0.006058 (  0.008085)
=> nil


samson(dev)> puts Benchmark.measure { deployed_or_running_list2(stages, reference)  }
  Deploy Load (1.0ms)  SELECT `deploys`.* FROM `deploys` INNER JOIN `stages` ON `stages`.`id` = `deploys`.`stage_id` AND `stages`.`deleted_at` IS NULL WHERE `deploys`.`deleted_at` IS NULL AND `deploys`.`stage_id` IN (4, 2, 3, 1, 5, 6, 7) AND `deploys`.`reference` = 'master' ORDER BY `deploys`.`id` DESC
  0.002177   0.000184   0.002361 (  0.003161)
=> nil
```

test/helpers/application_helper_test.rb
```diff
@@ -700,11 +700,23 @@ def render
       assert html.html_safe?
     end
 
-    it "renders succeeded deploys" do
+    it "renders current, succeeded deploys" do
       html = deployed_or_running_list(stage_list, "staging")
```
ragurney: Might be nice to use `assert_sql_queries` here to lock in the number of queries in case someone comes along and changes this later.

app/helpers/application_helper.rb
dasch: It depends – if we cache each row in the table based on the Release, and make sure to `touch` the release whenever it's deployed, then an _n + 1_ makes sense, as the common case will be that almost all of the releases will be cached and therefore won't trigger any queries. That's preferable to naively preloading everything.

I'd like to add that caching in a separate PR, though, unless it's a blocker for this one.

grosser:
hmm now it's doing 3 sql queries per row and before it was 1 ?

grosser:
nvm ryan already discussed that ... had this tab open for too long ...

app/helpers/application_helper.rb
grosser: something like this ?
```
deploy = stage.last_deploy
if deploy.success?
  'success'
elsif deploy.running?
  'warning'
else
  'default'
end
```

dasch: Not quite – I'll push an updated version.

test/helpers/application_helper_test.rb
dasch: :+1:

app/helpers/application_helper.rb
```diff
@@ -316,20 +316,33 @@ def link_to_chart(name, values)
...
+        if deploy == stage.last_deploy
+          "label-success"
+        else
+          # Deploy is neither active nor is it the last successful one, but it
+          # succeeded in the past.
+          "label-default"
+        end
+      end
+
+      if label.present?
```
grosser: prefer consistent `if thing` instead of `if thing.nil?` + `if thing.present?`

app/helpers/application_helper.rb
```diff
@@ -316,20 +316,33 @@ def link_to_chart(name, values)
...
+        text = stage.name.html_safe
```
grosser: this smells like xss ... I can name my stage `<script>foo</script>`

app/helpers/application_helper.rb
```diff
@@ -316,20 +316,33 @@ def link_to_chart(name, values)
...
-      next unless deploy.references?(reference)
-      label = (deploy.active? ? "label-warning" : "label-success")
-
-      text = "".html_safe
-      text << stage.name
-      html << content_tag(:span, text, class: "label #{label} release-stage")
-      html << " "
+    pieces = stages.map do |stage|
+      # The first deploy is the most recent one.
+      deploy = stage.deploys.where(reference: reference).first
```
grosser: we should fetch all the deploys in a single query with `stage.project.deploys.select('max(id) as id').where(stage_id: stages.map(&:id), reference: reference).group(:stage_id)` ?

dasch: Won't that just give the Deploy id, not the full deploy?

dasch: Okay, I've pushed an optimized version that only does 1 query per row.

app/helpers/application_helper.rb
dasch: :+1:

app/helpers/application_helper.rb
dasch: :+1:

dasch:
@grosser one more review?

app/models/deploy.rb
```diff
@@ -171,6 +171,30 @@ def self.after(deploy)
...
+  # Returns a Hash of Stage => Deploy for the given reference and stages.
+  def self.of_reference_in_stages(reference, stages)
+    # Group by stage, then select the latest deploy id.
+    deploys_and_stages = where(reference: reference).
+      in_stages(stages).
+      group(:stage_id).
+      select("MAX(id) AS id, stage_id")
+
+    # Fetch the actual deploys.
+    deploys = Deploy.where(id: deploys_and_stages.map(&:id))
```
grosser: we only use the id and stage_id ... so we could avoid an extra query ?

app/models/deploy.rb
```diff
@@ -171,6 +171,30 @@ def self.after(deploy)
     where("#{table_name}.id > ?", deploy.id)
   end
 
+  def self.in_stages(stages)
+    where(stage_id: stages.map(&:id))
+  end
+
+  # Returns a Hash of Stage => Deploy for the given reference and stages.
+  def self.of_reference_in_stages(reference, stages)
```
grosser: grouped hashes are usually built from methods with `by` not `of`

app/helpers/application_helper.rb
```diff
@@ -322,20 +322,35 @@ def link_to_chart(name, values)
...
-    stages.each do |stage|
-      next unless deploy = stage.deployed_or_running_deploy
-      next unless deploy.references?(reference)
-      label = (deploy.active? ? "label-warning" : "label-success")
-
-      text = "".html_safe
-      text << stage.name
-      html << content_tag(:span, text, class: "label #{label} release-stage")
-      html << " "
+    deploys = Deploy.of_reference_in_stages(reference, stages)
```
grosser:
```Ruby
deploy_lookup = Deploy.where(reference: reference).where(stage_id: stages.map(&:id)).group(:stage_id).      select("MAX(id) AS id, stage_id").group_by(&:stage_id)
```

app/models/deploy.rb
```diff
@@ -171,6 +171,30 @@ def self.after(deploy)
...
+
+    # Map to a hash of stage => deploy entries.
+    deploys.map do |deploy|
+      # Don't trigger another query in order to fetch the stage.
+      stage = stages.find { |stage| stage.id == deploy.stage_id }
+
+      [stage, deploy]
+    end.to_h
+  end
+
```
grosser: all pretty funky and not reused, so I'd keep it inline

app/models/deploy.rb
```diff
@@ -171,6 +171,30 @@ def self.after(deploy)
     where("#{table_name}.id > ?", deploy.id)
   end
 
+  def self.in_stages(stages)
+    where(stage_id: stages.map(&:id))
+  end
```
grosser: kinda simple query I'd rather keep inline

app/helpers/application_helper.rb
dasch: I really dislike putting SQL in the view layer – it makes testing so much harder. I'd very much prefer to keep this in the model.

app/models/deploy.rb
dasch: :+1:

app/helpers/application_helper.rb
grosser: the query logic is closely tied to the compare logic, so could lead to bugs/get out of sync when it's extracted to the model ... but I don't care that much
more important: the group_by line seems much easier to read (assuming it works as before) ?

app/helpers/application_helper.rb
```diff
@@ -322,20 +322,35 @@ def link_to_chart(name, values)
...
+
+    pieces = stages.map do |stage|
+      deploy = deploys[stage]
+
+      label = if deploy.nil?
+        nil
+      elsif deploy.running?
+        "label-warning"
+      elsif deploy.succeeded?
+        if deploy == stage.last_deploy
```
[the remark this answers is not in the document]
dasch: I mean – once you've called `Stage#last_deploy`, that should be memoized, right? If it's not, I can easily add that. The stages don't change between the release rows, so we shouldn't need to do any fancy work there.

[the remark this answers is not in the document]
grosser: yes, but it's being done once per stage -> n+1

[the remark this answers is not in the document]
dasch: How many stages do we typically have? Is it worth the added complexity?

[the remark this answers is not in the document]
grosser: projects can have tons of stages, not making it worse, so keep it like it is and open a BRE issue to make this more performant ?

PR #2898, opened 2018-08-24
Generic CI Pipeline Webhook

ragurney:
In the process of us investigating a new CI tool, we found ourselves in need of an endpoint to kick off deploys in Samson. It seemed that the most correct and convenient thing to do would be to create a generic webhook with a clearly defined acceptable payload to kick off a deploy that any future in house integrations could use.

screen shot 2018-08-24 at 10 48 43 am

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-923

### Risks
- Level: Low

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -11,7 +11,7 @@ class Integrations::BaseController < ApplicationController
   def create
     if !deploy? || skip?
       record_log :info, "Request is not supposed to trigger a deploy"
-      return render plain: @recorded_log.to_s
+      return render json: {deploy_ids: [], messages: @recorded_log.to_s}
     end
 
     if branch
@@ -27,14 +27,18 @@ class Integrations::BaseController < ApplicationController
     end
 
     stages = project.webhook_stages_for(branch, service_type, service_name)
-    failed = deploy_to_stages(release, stages)
+    deploy_results = deploy_to_stages(release, stages)
 
-    if failed
+    if failed = deploy_results[:failed_stage]
       record_log :error, "Failed to start deploy to #{failed.name}"
     else
       record_log :info, "Deploying to #{stages.size} stages"
     end
-    render plain: @recorded_log.to_s, status: (failed ? :unprocessable_entity : :ok)
+
+    render(
+      json: {deploy_ids: deploy_results[:deploy_ids], messages: @recorded_log.to_s},
+      status: (failed ? :unprocessable_entity : :ok)
+    )
   end
 
   protected
@@ -63,14 +67,18 @@ class Integrations::BaseController < ApplicationController
     ReleaseService.new(project).release(release_params)
   end
 
-  # returns stage that failed to deploy or nil
+  # returns failed stage and any successfully started deploy ids
   def deploy_to_stages(release, stages)
     deploy_service = DeployService.new(user)
-    stages.detect do |stage|
+    stages.each_with_object(failed_stage: nil, deploy_ids: []) do |stage, result_hash|
       deploy = deploy_service.deploy(stage, reference: release&.version || commit)
-      if deploy.new_record?
+
+      if deploy.persisted?
+        result_hash[:deploy_ids] << deploy.id
+      else
         record_log :error, "Deploy to #{stage.name} failed: #{deploy.errors.full_messages}"
-        true
+        result_hash[:failed_stage] = stage
+        break result_hash
       end
     end
   end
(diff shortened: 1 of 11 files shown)


app/controllers/integrations/base_controller.rb
```diff
@@ -63,14 +67,18 @@ def find_or_create_release
     ReleaseService.new(project).release(release_params)
   end
 
-  # returns stage that failed to deploy or nil
+  # returns failed stage and any successfully started deploy ids
   def deploy_to_stages(release, stages)
     deploy_service = DeployService.new(user)
-    stages.detect do |stage|
+    stages.each_with_object(failed_stage: nil, deploy_ids: []) do |stage, result_hash|
```
grosser: have 2 variables and build the hash at the end ?

app/controllers/integrations/ci_pipeline_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+module Integrations
+  class CiPipelineController < Integrations::BaseController
+    private
+
+    def deploy?
+      true
+    end
+
+    def deploy
+      @deploy ||= params.fetch(:deploy)
```
grosser: use `.require` to get nice error messages ?
(also below for commit/sha/message)

app/controllers/integrations/base_controller.rb
grosser: have 2 variables and build the hash at the end ?

app/helpers/webhooks_helper.rb
```diff
@@ -17,4 +17,28 @@ def webhook_sources_for_select(sources, none: false)
   def sources_for_select(sources)
     sources.map { |source| [source.titleize, source] }.to_a
   end
+
+  def webhook_help_text(source)
+    case source
+    when 'ci_pipeline'
+      help_text = <<~HTML.html_safe
```
grosser: maybe `CiPipelineController::HELP_TEXT` ?

config/routes.rb
```diff
@@ -154,6 +154,7 @@
     post "/jenkins/:token" => "jenkins#create", as: :jenkins_deploy
     post "/buildkite/:token" => "buildkite#create", as: :buildkite_deploy
     post "/github/:token" => "github#create", as: :github_deploy
+    post "/ci_pipeline/:token" => "ci_pipeline#create", as: :ci_pipeline_deploy
```
grosser: ci_pipeline seems very specific ... I'd have expected something like `generic` ... whatever calls this does not need to be part of a ci-pipeline

app/controllers/integrations/ci_pipeline_controller.rb
ragurney: How about something like a FormObject so we can use validations on that? Or else surround each method using `require` with a rescue so that we can give back 422s instead of blowing up with an exception?

app/controllers/integrations/base_controller.rb
ragurney: I kind of like the `each_with_object`, it feels a little cleaner to me than separately declaring two variables then explicitly returning the hash..

app/helpers/webhooks_helper.rb
ragurney: Since this string is used in only one place does it make sense to extract it to a constant? Then that's two files you have to update each time you want to add a new integration.

config/routes.rb
ragurney: 👍

config/routes.rb
```diff
@@ -154,6 +154,7 @@
     post "/jenkins/:token" => "jenkins#create", as: :jenkins_deploy
     post "/buildkite/:token" => "buildkite#create", as: :buildkite_deploy
     post "/github/:token" => "github#create", as: :github_deploy
+    post "/generic_integration/:token" => "generic_integration#create", as: :generic_integration_deploy
```
sathishavm: can you consider having different URL? 
/integrations/generic_integration little awkward,  maybe only /integrations/generic should be fine.

app/views/webhooks/index.html.erb
```diff
@@ -13,7 +13,8 @@
   <dl>
     <% Samson::Integration::SOURCES.each do |source| %>
       <dt>
-        <%= source.titleize %>:</dt><dd><%= send("integrations_#{source}_deploy_url", @project.token) %>
+        <%= source.titleize %><%= webhook_help_text source %>
+        </dt><dd><%= send("integrations_#{source}_deploy_url", @project.token) %>
```
sathishavm: Missing alignment!!

PR #2901, opened 2018-08-25
cache header projects a little to avoid heavy queries on every page load

grosser:
Avoid this query on every page
`Project Load (11.4ms)  SELECT projects.*, count(stars.id) as star_count FROM `projects` left outer join stars on stars.user_id = 1382 and stars.project_id = projects.id WHERE `projects`.`deleted_at` IS NULL GROUP BY projects.id ORDER BY star_count desc, name`

@sathishavm @ragurney 

... ideally cache this longer, but we'd have to expire on every project save+star

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -13,16 +13,18 @@
               </a>
             </li>
             <li class="divider"></li>
-            <% Project.ordered_for_user(current_user).each do |project| %>
-              <li class="filtered-projects">
-                <%= link_to project do %>
-                  <% starred = current_user.starred_project?(project) %>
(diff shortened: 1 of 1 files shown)


grosser:
expiring the cache is hard since it's 1 cache per user 🤷‍♀️ 
so we'd have to add key caching (cache header-version and use that as key), so I figured there are few deletions or permalink changes so it should be alright to just not do anything

grosser:
hmm how about we render this on-demand so the page gets lighter too ... and we save the cache round-trip :)

ragurney:
I like the idea of the on-demand render! Could use `responsive_load` helper for that?

grosser:
I'll make a PR if I find some play-time :)

PR #2904, opened 2018-08-29
document and allow opt-out from preStop since it causes warnings from…

grosser:
… containers that do not have sleep

```
Exec lifecycle hook ([sleep 3]) for Container "healthz" in Pod "-r9fnb_pod999(9ef757da-a884-11e8-8c23-0eda5c34567c)" failed - error: command 'sleep 3' exited with 126: , message: "rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:262: starting container process caused \"exec: \\\"sleep\\\": executable file not found in $PATH\"\n\r\n
```

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -163,3 +163,9 @@ Set `KUBERNETES_NO_CPU_LIMIT_ALLOWED=1`, see [#2820](https://github.com/zendesk/
 
 Knowing which team owns each component is useful, set `KUBERNETES_ENFORCE_TEAMS=true` 
 to make all kubernetes deploys that do not use a `metadata.labels.team` / `spec.template.metadata.labels.team` fail.
+
+### Preventing request loss with preStop
+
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/README.md
```diff
@@ -163,3 +163,9 @@ Set `KUBERNETES_NO_CPU_LIMIT_ALLOWED=1`, see [#2820](https://github.com/zendesk/
 
 Knowing which team owns each component is useful, set `KUBERNETES_ENFORCE_TEAMS=true` 
 to make all kubernetes deploys that do not use a `metadata.labels.team` / `spec.template.metadata.labels.team` fail.
+
+### Preventing request loss with preStop
+
+Samson auto-add a `container[].lifecycle.preStop` `sleep 3` if not preStop hook is set and
```
adammw: automatically adds .. if a preStop hook is not set

PR #2906, opened 2018-08-29
Skip caching if *any* status check is pending

dasch:
Previously, if a check has failed, we'd cache even though others were still pending. That meant that the checks were permanently shown as pending.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/models/github_status.rb
+++ b/app/models/github_status.rb
@@ -48,13 +48,13 @@ class GithubStatus
     # Fall back to a "missing" status.
     return new("missing", []) if response.nil?
 
-    # Don't cache pending statuses, since we expect updates soon.
-    unless response.state == "pending"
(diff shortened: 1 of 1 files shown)


app/models/github_status.rb
```diff
@@ -48,15 +48,15 @@ def self.fetch(repo, ref)
     # Fall back to a "missing" status.
     return new("missing", []) if response.nil?
 
-    # Don't cache pending statuses, since we expect updates soon.
-    unless response.state == "pending"
-      Rails.cache.write(cache_key, response, expires_in: 1.hour)
-    end
```
dasch: The state is `failed` if any of the checks have failed, even if others are pending.

PR #2907, opened 2018-08-29
Make caching more fine-grained

dasch:
Since the cache keys currently can't take into account the Github status checks, it's better to just cache the expensive parts of the page.

I'll try to see if there's a good way to `touch` a `Release` when the Github status changes, but we'd need to either poll the repo's Github event API or subscribe to another Github webhook. This is an immediate fix, and I think it'll be fine from a performance point of view, since we can cache the expensive parts separately. As far as I can see, only the deployed stages list actually fires off SQL queries.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/views/releases/_release.html.erb
+++ b/app/views/releases/_release.html.erb
@@ -14,7 +14,11 @@
       <%= github_user_avatar(user) %>
     <% end %>
   </td>
-  <td width="100%"><%= deployed_or_running_list @stages, release.version %></td>
+  <td width="100%">
+    <% cache ["deployed-stages", release, @stages] do %>
+      <%= deployed_or_running_list @stages, release.version %>
+    <% end %>
+  </td>
   <td><%= relative_time(release.created_at) %></td>
   <td>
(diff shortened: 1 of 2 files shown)


app/views/releases/_release.html.erb
```diff
@@ -14,7 +14,11 @@
       <%= github_user_avatar(user) %>
     <% end %>
   </td>
-  <td width="100%"><%= deployed_or_running_list @stages, release.version %></td>
+  <td width="100%">
+    <% cache ["deployed-stages", release, @stages] do %>
```
ragurney: Do we want to add an `expires_in` option to this?

dasch: There should be no need for that – if either `release` or any of `@stages` get updated, the cache key will change.

ragurney: Ah and `stage` would get updated if a deploy's `job` is updated. Sounds good 👍

PR #2910, opened 2018-08-29
PAAS-2022 support CronJob

grosser:
tested via https://github.com/samson-test-org/example-kubernetes/pull/5

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -436,9 +436,16 @@ module Kubernetes
         delete_pods { super }
       end
 
-      # FYI per docs it is supposed to use batch api, but extension api works
       def client
         @deploy_group.kubernetes_cluster.client('batch/v1')
+      end
+    end
+
+    class CronJob < Base
+      private
+
+      def client
+        @deploy_group.kubernetes_cluster.client('batch/v1beta1')
       end
     end
 
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -9,6 +9,7 @@ module Kubernetes
       ['Deployment', 'Service'],
       ['Service', 'StatefulSet'],
       ['Job'],
+      ['CronJob'],
       ['Pod'],
     ].freeze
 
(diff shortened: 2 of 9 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -436,12 +436,19 @@ def request_delete
...
       def client
         @deploy_group.kubernetes_cluster.client('batch/v1')
       end
     end
 
+    class CronJob < Base
+      private
+
+      def client
+        @deploy_group.kubernetes_cluster.client('batch/v1beta1')
```
adammw: support 'batch/v2alpha1' for k8s <v1.8, and/or fail on older?

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -187,8 +178,9 @@ def validate_containers
       primary_kinds = RoleConfigFile::PRIMARY_KINDS
       containered = templates.select { |t| primary_kinds.include?(t[:kind]) }
       containers = map_attributes([:spec, :containers], elements: containered)
-      return if containers.all? { |c| c.is_a?(Array) && c.size >= 1 }
-      @errors << "#{primary_kinds.join("/")} need at least 1 container"
+      bad = containers.select { |c| !c.is_a?(Array) || c.empty? }
+      return if bad.empty?
+      @errors << "#{containered.map { |r| r[:kind] }.join("/")} need at least 1 container"
```
adammw: nit: needs not need

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: `You need a working Kubernetes cluster at version >= 1.8 (for CronJob). For previous versions of cluster (< 1.8) you need to explicitly enable batch/v2alpha1 API by passing --runtime-config=batch/v2alpha1=true`
... so I'd rather use the official version ... also who would run such an old cluster :trollface:

plugins/kubernetes/app/models/kubernetes/role_validator.rb
grosser: could be multiple 🤷‍♂️ 
... but I guess will be 1 most of the time ...

plugins/kubernetes/app/models/kubernetes/resource.rb
adammw: ok, but then CronJob should blow up / not validate on a <1.8 cluster.

plugins/kubernetes/app/models/kubernetes/role_validator.rb
adammw: english is hard. i think it's got to do with adverb vs verb or something but :shrug:

grosser:
deployment of it will fail just like all the other things if someone
deployed to StatefulSet a 1.1 cluster

adammw:
fix the flaky test?

PR #2912, opened 2018-08-30
UI fixes for release authors

dasch:
* Make sure the avatars stack next to each other rather than on top of each other.
* Use Bootstrap's tooltip for consistency and legibility.

**Before:**

**After:**

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/views/releases/_release.html.erb
+++ b/app/views/releases/_release.html.erb
@@ -10,9 +10,11 @@
     <% end %>
   </td>
   <td>
-    <% release.changeset.authors.take(8).each do |user| %>
-      <%= github_user_avatar(user) %>
-    <% end %>
+    <span style="white-space: nowrap">
+      <% release.changeset.authors.take(8).each do |user| %>
+        <%= github_user_avatar(user) %>
+      <% end %>
+    </span>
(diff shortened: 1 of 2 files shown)


app/views/releases/_release.html.erb
```diff
@@ -10,9 +10,11 @@
     <% end %>
   </td>
   <td>
-    <% release.changeset.authors.take(8).each do |user| %>
-      <%= github_user_avatar(user) %>
-    <% end %>
+    <span style="white-space: nowrap">
```
ragurney: Would prefer adding an id here instead of inline CSS but not a big deal 🤷‍♂️

app/views/releases/_release.html.erb
```diff
@@ -10,9 +10,11 @@
...
+      <% release.changeset.authors.take(8).each do |user| %>
```
grosser: should we uniq this to remove clutter ?

ragurney: I think `authors` already has a `uniq` call

PR #2914, opened 2018-08-30
Handle Github commit status events

dasch:
A new commit status causes all releases of that commit to be touched, meaning that the status can be cached by `Release#updated_at`.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/controllers/integrations/github_controller.rb
+++ b/app/controllers/integrations/github_controller.rb
@@ -14,6 +14,8 @@ class Integrations::GithubController < Integrations::BaseController
   def create
     if github_event_type == "status"
       handle_commit_status_event
+
+      render plain: "OK", status: 200
     else
       super
     end
--- a/app/controllers/integrations/github_controller.rb
+++ b/app/controllers/integrations/github_controller.rb
@@ -12,13 +12,9 @@ class Integrations::GithubController < Integrations::BaseController
   end
 
   def create
-    if github_event_type == "status"
-      handle_commit_status_event
+    handle_commit_status_event if github_event_type == "status"
 
-      render json: {deploy_ids: [], messages: "status event received"}, status: :ok
-    else
-      super
-    end
+    super
   end
 
   protected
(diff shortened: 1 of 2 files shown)


app/controllers/integrations/github_controller.rb
```diff
@@ -11,8 +11,21 @@ def self.secret_token
     ENV['GITHUB_HOOK_SECRET']
   end
 
+  def create
+    if github_event_type == "status"
+      handle_commit_status_event
```
grosser: I don't see any kind of rendering happening here ... looks suspicious oO

app/controllers/integrations/github_controller.rb
```diff
@@ -11,8 +11,21 @@ def self.secret_token
...
+    else
+      super
+    end
+  end
+
   protected
 
+  def handle_commit_status_event
```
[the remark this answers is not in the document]
dasch: The thing is – it doesn't expire the caches. It merely marks the release as being updated. The fact that our caching system uses that information is incidental – it could be used in other contexts as well. That provides a nice level of decoupling IMO.

dasch:
I'd rather not have this code be specific to caches, since it only deals with the model layer. It really just makes sure the `updated_at` timestamp on a Release reflects the underlying data relevant to that release. We happen to base our caches on that.

dasch:
@grosser final review?

dasch:
@grosser I've implemented your feedback. Approve?

app/controllers/integrations/github_controller.rb
```diff
@@ -11,8 +11,23 @@ def self.secret_token
     ENV['GITHUB_HOOK_SECRET']
   end
 
+  def create
+    if github_event_type == "status"
+      handle_commit_status_event
+
+      render json: {deploy_ids: [], messages: "status event received"}, status: :ok
```
grosser: maybe call super here ... to not break existing behavior / not have to copy-paste the rendering ?

grosser:
```Ruby
handle_commit_status_event  if github_event_type == "status"
super
```

PR #2915, opened 2018-08-30
detect pull requests in Changeset using Github API

neerfri:
At Applicaster we deploy branches to production before we merge them to master (a practice we've borrowed from github https://githubengineering.com/deploying-branches-to-github-com/)
We are also using the built-in slack_webhooks plugin.

The problem we have is that Samson will not detect the pull request in the changeset, and so the slack plugin [will warn that a deployment is made without a pull request](https://github.com/zendesk/samson/blob/79bbedb167738006f0021ff4fe215cf4790fe93e/plugins/slack_webhooks/app/views/samson_slack_webhooks/notification.text.erb#L8-L10)

This makes everyone jump to try to find out who is deploying without a pull request, which is never the case. But never say never :trollface: 
This is especially annoying since there's a `@chanell` keyword in the message so all members of the `deployments` channel are brought in.

This PR adds a call to Github's [List pull requests endpoint](https://developer.github.com/v3/pulls/#list-pull-requests) in an attempt to resolve the PR for the deployed branch.
It will skip calling the API if it detects that the reference (called `commit` in the code) is a commit SHA reference by matching it to a simple regular expression.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -84,11 +84,26 @@ class Changeset
 
   def find_pull_requests
     numbers = commits.map(&:pull_request_number).compact
-    numbers.map { |num| PullRequest.find(repo, num) }.compact
+    numbers.map { |num| PullRequest.find(repo, num) }.
+      compact.
+      concat(find_pull_requests_for_branch)
   end
 
   def cache_key
     [self.class, repo, previous_commit, commit].join('-')
+  end
+
+  def find_pull_requests_for_branch
+    return [] if not_pr_branch?
+    org = repo.split("/", 2).first
+    GITHUB.pull_requests(repo, head: "#{org}:#{commit}")
+  rescue Octokit::Error, Faraday::ConnectionFailed => e
+    Rails.logger.warn "Failed fetching pull requests for branch #{commit}:\n#{e}"
+    []
+  end
+
+  def not_pr_branch?
+    commit =~ Build::SHA1_REGEX || commit =~ Release::VERSION_REGEX
   end
 
   class NullComparison
--- a/plugins/ledger/test/lib/samson_ledger/client_test.rb
+++ b/plugins/ledger/test/lib/samson_ledger/client_test.rb
@@ -29,6 +29,7 @@ describe SamsonLedger::Client do
       stub_github_api("repos/bar/foo/compare/abcabcaaabcabcaaabcabcaaabcabcaaabcabca1...staging", "x" => "y")
       GITHUB.stubs(:compare).with("bar/foo", "abcabcaaabcabcaaabcabcaaabcabcaaabcabca1", "staging").returns(comparison)
       Changeset::PullRequest.stubs(:find).with("bar/foo", 42).returns(pull_request)
+      GITHUB.stubs(:pull_requests).with("bar/foo", head: "bar:staging").returns([])
 
       request_lambda = ->(request) do
         results << JSON.parse(request.body)['events'].first
--- a/test/models/changeset_test.rb
+++ b/test/models/changeset_test.rb
@@ -88,15 +88,46 @@ describe Changeset do
     it "finds pull requests mentioned in merge commits" do
       comparison = Sawyer::Resource.new(sawyer_agent, commits: [commit1, commit2])
       GITHUB.stubs(:compare).with("foo/bar", "a", "b").returns(comparison)
+      GITHUB.stubs(:pull_requests).with("foo/bar", head: "foo:b").returns([])
 
       Changeset::PullRequest.stubs(:find).with("foo/bar", 42).returns("yeah!")
 
       changeset.pull_requests.must_equal ["yeah!"]
     end
 
+    it "finds pull requests open for a branch" do
+      comparison = Sawyer::Resource.new(sawyer_agent, commits: [commit2])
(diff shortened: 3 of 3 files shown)


app/models/changeset.rb
```diff
@@ -84,13 +84,25 @@ def find_comparison
...
+      .concat(find_pull_request_for_branch)
   end
 
   def cache_key
     [self.class, repo, previous_commit, commit].join('-')
   end
 
+  def find_pull_request_for_branch
+    # Bail if commit looks like a commit SHA
+    return [] if commit =~ /[0-9a-f]{40}/
```
grosser: `Build::SHA1_REGEX`

app/models/changeset.rb
```diff
@@ -84,13 +84,25 @@ def find_comparison
...
+    org = repo.split("/", 2).first
+    GITHUB.pull_requests(repo, head: "#{org}:#{commit}")
```
grosser: does not work without the `org:` ?

neerfri: https://github.com/octokit/octokit.rb/issues/865#issuecomment-280454776

app/models/changeset.rb
neerfri: 👍

app/models/changeset.rb
```diff
@@ -84,13 +84,25 @@ def find_comparison
...
+  rescue Octokit::Error
+    Rails.logger.warn "Failed fetching pull requests for branch #{commit}"
```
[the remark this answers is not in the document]
neerfri: 👍

app/models/changeset.rb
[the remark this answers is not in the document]
neerfri: This is added after the `numbers.map { |num| PullRequest.find(repo, num) }` so we save another call to `PullRequest.find(repo, num)`.
Am I missing something here?

neerfri:
Evidently this PR has a lot of "junk" commits, can this be "squash-merged"?
Is there a reason why you don't switch the repo to alway "squash"?

grosser:
you can squasch your own commits, afaik we have prefer regular merges since they make auditing easier (which PRs were merged) and allow for PRs to address multiple concerns in clean commits

app/models/changeset.rb
```diff
@@ -84,13 +84,32 @@ def find_comparison
...
+    numbers.map { |num| PullRequest.find(repo, num) }.
+      compact.
+      concat(find_pull_request_for_branch)
   end
 
   def cache_key
     [self.class, repo, previous_commit, commit].join('-')
   end
 
+  def find_pull_request_for_branch
```
grosser: the `find_` feels redundant

grosser: also should be `pull_requests` or not return an array ?

app/models/changeset.rb
```diff
@@ -84,13 +84,32 @@ def find_comparison
...
+    # Bail if commit looks like a commit SHA
+    return [] if skip_pull_request_fetch_from_api?
+    org = repo.split("/", 2).first
+    GITHUB.pull_requests(repo, head: "#{org}:#{commit}")
+  rescue Octokit::Error, Faraday::ConnectionFailed => e
+    Rails.logger.warn "Failed fetching pull requests for branch #{commit}:\n#{e}"
+    []
+  end
+
+  def skip_pull_request_fetch_from_api?
```
grosser: `def looks_like_a_branch?`
and use `&&` to join things and avoid all the if/returns

app/models/changeset.rb
neerfri: The `find_` was chosen to match the rest of the code of this class.
For example the `pull_requests` method is used to call `find_pull_requests` and cache the result:
https://github.com/zendesk/samson/blob/b8cc8aeff1abcc803ac699eccd57833564923087/app/models/changeset.rb#L33-L35

I don't mind changing the method name if that's your preference

I'll change the name to plural to indicate that it returns an array 🥇

app/models/changeset.rb
neerfri: I'll use the `&&`
Are you sure about the name change because `master` (`staging`, ...) are branches, so `looks_like_a_branch?` does not really reflect the code.

grosser: ah yes ... maybe `pr_branch?`

app/models/changeset.rb
grosser: hmm ok consistency is nice too 🤷‍♂️

app/models/changeset.rb
neerfri: Maybe we should not put these branches there at all, maybe some users use PRs to move between `staging` to `production`...

neerfri: `master` (and `develop`) won't make that request anyway cause they are converted to SHA, and it's not a big deal to make another request that will come back empty

grosser: sounds reasonable

neerfri:
@grosser all fixed and commits squashed

grosser:
travis wants some ❤️  :)

neerfri:
sent him a hug, let's hope he will return some ❤️

PR #2919, opened 2018-09-03
Display richer status tooltips

dasch:
Work in progress.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -24,10 +24,22 @@ module ReleasesHelper
     )
   end
 
-  def github_commit_status_icon(status_state)
-    icon = GITHUB_STATUS_ICONS.fetch(status_state)
-    text = GITHUB_STATUS_TEXT_LABELS.fetch(status_state)
-    title = "Github status: #{status_state}"
+  def github_commit_status_icon(status)
+    failed = status.failed
+    succeeded = status.succeeded
+    errored = status.errored
+    pending = status.pending
+
+    status_counts = []
+    status_counts << "#{failed.count} failed" if failed.any?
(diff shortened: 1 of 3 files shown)


app/helpers/releases_helper.rb
```diff
@@ -24,10 +24,22 @@ def release_label(project, release)
...
+    failed = status.failed
+    succeeded = status.succeeded
+    errored = status.errored
+    pending = status.pending
+
+    status_counts = []
+    status_counts << "#{failed.count} failed" if failed.any?
+    status_counts << "#{succeeded.count} succeeded" if succeeded.any?
+    status_counts << "#{errored.count} errored" if errored.any?
+    status_counts << "#{pending.count} pending" if pending.any?
```
grosser: how about something like that to avoid duplication here and also having to add another 4 copy-pasty methods that need to stay in sync with this code

```Ruby
[:failed?, :errored?, :pending?, :succeeded?].map do |status|
  found = status.status.select(&status)
  "#{found} #{status.to_s.sub('?', '')}" if found.any?
end.compact
```

dasch: I'll take repetition over cleverness any day of the week...

grosser: could use this to make it less clever/more repetitive but still dry, also solves all the 'missing test coverage' errors
```Ruby
[['failed, ':failed?], ['errored', :errored?], ['pending', :pending?], ['succeeded', :succeeded?]].map do |name, status|
  found = status.status.select(&status)
  "#{found} #{name}" if found.any?
end.compact
```

PR #2921, opened 2018-09-04
[BRE-943] Add styling to Rollbar tables to prevent overflow

ragurney:
Before:
screen shot 2018-09-04 at 12 09 39 pm
After:
screen shot 2018-09-04 at 12 09 32 pm

/cc @zendesk/samson

### References
 - Jira link: [BRE-943](https://zendesk.atlassian.net/browse/BRE-943)

### Risks
- Level: Low


ragurney:
It's part of the [BEM naming convention](http://getbem.com/naming/).. Thought it would be nice to start introducing it, but can remove if you'd rather not go that direction 🤷‍♂️

grosser:
meh ... I guess it's better than "do whatever" ... still kinda looks weird
to me :D

PR #2924, opened 2018-09-05
show hero as a setup option

grosser:
before it sat at the bottom of the page without any info ...

@ragurney @bcolferzd


grosser:
... also making DEFAULT_URL optional on heroku by using the HEROKU_APP_NAME

PR #2934, opened 2018-09-12
Only cache Github Status for old releases

dasch:
@grosser ?

### Risks
- Level: Low


grosser:
logic looks good though!

grosser:
logic is in commit_status already

PR #2938, opened 2018-09-14
tell people where to look when they explore the api

grosser:
also make messages a bit better


grosser:
@ragurney
/cc @zendesk/compute

PR #2940, opened 2018-09-14
Gitlab changeset support

pingfr:
* These changes add plugin support for both GitHub and GitLab change sets.  In addition to the addition of the plugins, there is support for a factory that selects the correct remote repo based on environmental setup.  The default setup remains GitHub for backward compatibility.  I was the sole developer working this for PlanSource.  We use GitLab but not GitHub.  I had time to do live testing on GitLab, but not GitHub.    The goal was to emit data in a common format (GitHub) so that minimal changes were needed in the UI.  Tests were added for the plugins.  

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High


config/application.rb
```diff
@@ -108,6 +108,9 @@ class Application < Rails::Application
     # The directory in which repositories should be cached.
     config.samson.cached_repos_dir = Rails.root.join("cached_repos")
 
+    # Choose the remote repository for this Samson implementation.  Options are github or gitlab
+    config.samson.remote_repository = ENV["REMOTE_REPO"] || 'github'
```
grosser: maybe something like `git_host` ? or `git_provider`

app/models/changeset_factory.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def self.commit
+    "#{changeset_discriptor}::Commit".constantize
+  end
+
+  private
+
+  def self.changeset_discriptor
+    "Samson::#{remote_repository.camelize}::Changeset"
```
grosser: better cache that

config/application.rb
pingfr: Either of those sound like a good choice.  Do you have a preference?  I think I am leaning toward git_provider.

grosser:
this looks pretty good ... can you break out smaller PRs from this to make it easier to review ?
(like "move github to be a plugin" / "make " etc)

ideally it should not hardcode gitlab or github but pick whatever the current project uses to be flexible ... not sure if that's feasible

config/initializers/gitlab.rb
```diff
@@ -0,0 +1,7 @@
+Gitlab.configure do |config|
```
grosser: should be in the gitlab plugin ? (... and github should be in a github plugin)

pingfr:
The choice of which git provider is pivoted off of the .env file, based on the value set config.samson.remote_repository = ENV["REMOTE_REPO"] || 'github'

If you look at the factory that picks up the plugin, it will decide based on that value.

plugins/github_changeset/app/models/samson/github/changeset.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+module Samson
+  module Github
+    class Changeset
+      attr_reader :repo, :previous_commit, :commit
+      BRANCH_TAGS = ["master", "develop"].freeze
+      ATTRIBUTE_TABS = %w[files commits pull_requests risks jira_issues].freeze
```
grosser: this is the same / needs to stay the same ... so how about we don't duplicate them ?

grosser:
the gitlab plugin is mostly a copy-paste with a few changes ... if we don't want this to get out of sync we need some kind of base that github/gitlab can inherit from (or include)

config/initializers/gitlab.rb
pingfr: My patch from my working directory moved the GitLab config from where it was.  I will fix that.  There are still a couple of places in the Review process that reference GitHub, so the gem reference is still in the main application.

grosser:
yeah I saw the remote_repository logic, but I was hoping it could be decided based on what the individual project uses so github/lab can be used at the same time

config/application.rb
grosser: I like em both 👍

pingfr:
Base classing the plugins is a good idea and can be done reasonably quickly.  It also addresses you concern for some duplicated constants.

Regarding the pivot of provider based on the project, let me take a look at that.  Currently the projects DB only has the repo url.  That could be used to determine the repo and possibly pivot as you suggested.  

Unfortunately breaking this up into smaller pieces for review would be very difficult.  There was so much Github logic in the core application, that the majority of changes had to be done together to allow it to work.

grosser:
Smaller PRs might work with steps like `extract github into a plugin` etc and then adding a new gitlab plugin in a PR ... might even work with taking this PR and deleting all gitlab code so it would basically be a big move + refactor

grosser:
https://github.com/zendesk/samson/pull/3227 is a good start ... maybe build on that

PR #2943, opened 2018-09-20
[BRE-887] Split out 'after deploy' notification conditions

ragurney:
screen shot 2018-09-20 at 3 25 59 pm

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: [BRE-887](https://zendesk.atlassian.net/browse/BRE-887)

### Risks
- Level: Low


dragonfax:
You're going to need a backfill for this, right?

ragurney:
Yep! The migration should take care of it.

ragurney:
Admitting defeat and using database adapters instead of raw sql

grosser:
so now we can notify only on success ... that seems like a weird usecase oO

ragurney:
¯\_(ツ)_/¯

PR #2944, opened 2018-09-21
include counter in daemonset error message

grosser:
@zendesk/compute


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -361,12 +360,12 @@ def client
...
           loop_sleep
-          expire_cache
-          return if no_pods_running?
+          expire_resource_cache
+          return if pods_running_count == 0
         end
         raise(
           Samson::Hooks::UserError,
-          "Unable to terminate DaemonSet #{name} #{namespace} #{@deploy_group.name} because it still has pods.\n" \
+          "Unable to terminate DaemonSet #{error_location} because it still has #{pods_running_count} pods.\n" \
```
adammw: #{'pods'.pluralize(pods_running_count)}

PR #2946, opened 2018-09-21
[BRE-969] Allow Stages to have a default reference

### LINKED ISSUE #2745: Default reference for a stage
Hi there,

I was thinking in implementing a feature so a stage can have a default reference, different from master.

The use case of this feature is that we currently have a `staging` stage, that always deploys the code from the `staging` branch, as opposed from the `production` deploy, which deploys from the `master` branch, in some projects, automatically, in others manually.

In the latter ones, we can check the `Disable reference selection`, so it automatically pick ups the `master` branch, but for the others, we always have to type `staging` in the reference box

It would be awesome if the stages could have an persistent attribute (`default_reference`) where we could store the default reference of that stage...

So yeah, what do you think? just touching base before I start to implementing it... don't think is gonna be specially complicated 

/cc @zendesk/samson

ragurney:
Resolves #2745 

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: [BRE-969](https://zendesk.atlassian.net/browse/BRE-969)

### Risks
- Level: Low

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -35,8 +35,10 @@ module ApplicationHelper
     elsif Lock.locked_for?(stage, current_user)
       content_tag :a, "Locked", class: "btn btn-primary disabled", disabled: true
     elsif stage.direct?
-      path = project_stage_deploys_path(project, stage, deploy: {reference: "master", stage_id: stage.id})
-      link_to "Deploy!", path, role: "button", class: "btn btn-warning", data: {method: :post}
+      path = project_stage_deploys_path(
+        project, stage, deploy: {reference: stage.default_reference.presence || "master", stage_id: stage.id}
+      )
+      link_to "Deploy", path, role: "button", class: "btn btn-warning", data: {method: :post}
     else
       path = new_project_stage_deploy_path(project, stage)
       link_to "Deploy", path, role: "button", class: "btn btn-primary"
--- a/db/migrate/20180921170100_add_default_reference_to_stages.rb
+++ b/db/migrate/20180921170100_add_default_reference_to_stages.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDefaultReferenceToStages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :stages, :default_reference, :string
+  end
+end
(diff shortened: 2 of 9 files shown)


test/controllers/deploys_controller_test.rb
```diff
@@ -368,6 +368,20 @@ def changeset(overrides = {})
         deploy.stage.must_equal stage
         deploy.reference.must_equal "abcd"
       end
+
+      describe "with default reference" do
+        before { stage.update_column(:default_reference, 'pandas4lyfe') }
+
+        it "sets default reference" do
+          get :new, params: {project_id: project.to_param, stage_id: stage.to_param}
+          assigns(:deploy).reference.must_equal 'pandas4lyfe'
```
dragonfax: 🐼

app/helpers/application_helper.rb
```diff
@@ -35,8 +35,10 @@ def deploy_link(project, stage)
     elsif Lock.locked_for?(stage, current_user)
       content_tag :a, "Locked", class: "btn btn-primary disabled", disabled: true
     elsif stage.direct?
-      path = project_stage_deploys_path(project, stage, deploy: {reference: "master", stage_id: stage.id})
-      link_to "Deploy!", path, role: "button", class: "btn btn-warning", data: {method: :post}
+      path = project_stage_deploys_path(
+        project, stage, deploy: {reference: stage.default_reference.presence || "master", stage_id: stage.id}
```
grosser: default_reference should return "master" when not set ?

db/migrate/20180921170100_add_default_reference_to_stages.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDefaultReferenceToStages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :stages, :default_reference, :string
```
grosser: `default: 'master'`

ragurney: Do we always want a default though? Might be kind of annoying for people to have to delete "master" each time.

nogates:
Nice! Thank you for this! :)

PR #2948, opened 2018-09-21
PAAS-2142 do not allow setting 0/2+ replicas for resources that do not support …

grosser:
…replicas

@zendesk/compute 

the only place were 0 replicas makes somewhat sense is for deployments that we can later scale up / use as template for consoles ... for everything else it should be either 1 or deleted

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -85,7 +85,7 @@ module Kubernetes
       end
 
       def desired_pod_count
-        replica_source.dig_fetch(:spec, :replicas)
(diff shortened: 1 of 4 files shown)


adammw:
what items have replicas but not templates? :/

grosser:
none ... but we set fake replicas for things like pods/jobs/daemonsets

adammw:
that's confusing...

grosser:
the UI allows setting replicas for each resource, but at that level samson does not know if the thing is a job/pod/deployment. If someone sets `0` or `2` replicas for a pod then that does not make any sense and ends up deploying something the user did not expect. So we fail early an tell the user that this won't work.

PR #2952, opened 2018-09-22
combine commit-status and github_status

grosser:
/cc @zendesk/samson 

screen shot 2018-09-22 at 9 54 05 am
screen shot 2018-09-22 at 9 51 39 am
screen shot 2018-09-22 at 9 51 29 am

also adds smarter caching logic to the status cache (atm does not cache old releases if they are still pending)


app/models/commit_status.rb
```diff
@@ -2,24 +2,23 @@
...
+  # - fatal is our own state that blocks deploys
+  # - missing is our own state that means we could not determine the status
+  STATE_PRIORITY = [:success, :pending, :missing, :failure, :error, :fatal].freeze
+  UNDETERMINED = ["pending", "missing"].freeze
+
+  def initialize(project, reference, stage: nil)
+    @project = project
     @reference = reference
+    @stage = stage
   end
```
dasch: If you're going to do this, why not make CommitStatus _less_ aware of our domain? I'd just give it a Git sha., since I don't think you need project/stage in the cache key, right?

grosser:
I tried that at first, but that means I have to either:
 - a pull+git-show
 - resolve via github api

which are both not great

dasch:
Ah, because sometimes non-sha's are passed?

app/assets/javascripts/ref_status_typeahead.js
```diff
@@ -39,7 +39,6 @@ function refStatusTypeahead(options){
 
   function show_status_problems(status_list, isDanger) {
     $ref_status_container.removeClass("hidden");
-console.log(isDanger);
```
ragurney: 😅

app/models/commit_status.rb
```diff
@@ -2,24 +2,23 @@
...
-    fatal: 4
-  }.freeze
-
-  def initialize(stage, reference)
-    @stage = stage
+  # See https://developer.github.com/v3/repos/statuses for api details
+  # - fatal is our own state that blocks deploys
+  # - missing is our own state that means we could not determine the status
+  STATE_PRIORITY = [:success, :pending, :missing, :failure, :error, :fatal].freeze
+  UNDETERMINED = ["pending", "missing"].freeze
```
ragurney: Nice 👍

PR #2954, opened 2018-09-25
remove indirections from references controller/service/settings

grosser:
also makes tests not use stubs / reuse common git setup

@ragurney 

was looking at other git foobar and noticed this convoluted web ... much simpler now and less brittle since it tests the whole thing from request to response without stubbing

also making sure REFERENCES_CACHE_TTL is always a string -> int so we don't get surprises late into the code and only deal with 1 datatype


app/controllers/references_controller.rb
```diff
@@ -1,11 +1,15 @@
 # frozen_string_literal: true
 class ReferencesController < ApplicationController
   include CurrentProject
+  CACHE_TTL = Integer(ENV['REFERENCES_CACHE_TTL'].presence || 10.minutes.to_s)
```
ragurney: Not sure you need `to_s` here

grosser:
I want it to make both be a string for consistency

PR #2955, opened 2018-09-25
buddy request fix for modified brach

sathishavm:
* Trigger a buddy request for the modified branch within the buddy grace_period.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-971

### Risks
- Level: Low

### DIFF
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -78,6 +78,7 @@ class DeployService
 
   def copy_approval_from_last_deploy(deploy)
     return false unless last_deploy = latest_approved_deploy(deploy.reference, deploy.stage.project)
+    return false if deploy.changeset_to(last_deploy).commits.any?
     return false if stage_script_changed_after?(last_deploy)
 
     deploy.buddy = (last_deploy.buddy == @user ? last_deploy.job.user : last_deploy.buddy)
--- a/test/models/deploy_service_test.rb
+++ b/test/models/deploy_service_test.rb
@@ -33,8 +33,9 @@ describe DeployService do
       before { BuddyCheck.stubs(:enabled?).returns(true) }
       let(:deploy) { deploys(:succeeded_production_test) }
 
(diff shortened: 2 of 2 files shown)


app/models/deploy_service.rb
```diff
@@ -78,6 +78,7 @@ def latest_approved_deploy(reference, project)
 
   def copy_approval_from_last_deploy(deploy)
     return false unless last_deploy = latest_approved_deploy(deploy.reference, deploy.stage.project)
+    return false if deploy.changeset_to(last_deploy).commits.any?
```
grosser: might be cheaper to resolve commit and check if it is not the same ... also make sure changeset_to does not compare refs which would be the same

test/models/deploy_service_test.rb
```diff
@@ -85,6 +86,15 @@ def create_previous_deploy(ref, stage, successful: true, bypassed: false)
         end
       end
 
+      describe "similar deploy reference with different commit sha" do
```
grosser: similar -> same

test/models/deploy_service_test.rb
```diff
@@ -33,8 +33,9 @@
       before { BuddyCheck.stubs(:enabled?).returns(true) }
       let(:deploy) { deploys(:succeeded_production_test) }
 
-      def create_previous_deploy(ref, stage, successful: true, bypassed: false)
-        job = project.jobs.create!(user: user, command: "foo", status: successful ? "succeeded" : 'failed')
+      def create_previous_deploy(ref, stage, successful: true, bypassed: false, commit: '')
```
grosser: empty commit seems kinda strange ... maybe nil ?

app/models/deploy_service.rb
sathishavm: We may cache the changeset in the previous Review step, it should re-use the cache.

PR #2957, opened 2018-09-27
Disabled `Does not deploy code` for non admin users

sathishavm:
As per compliance, `Does not deploy code` gives a way to skip buddy check and so we need to restrict the functionality to global admins.
As **no_code_deployed** is depended to `no_reference_selection` and `Confirm before deployment`, so it's best to show the feature and restrict to use.
 
screen shot 2018-09-27 at 4 29 38 pm

/cc @zendesk/samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-926

### Risks
- Level: Low


app/views/stages/_fields.html.erb
```diff
@@ -23,8 +23,10 @@
       " When selected with '#{no_code_label}' and '#{confirm_label}' is disabled the stage can be directly executed."
   %>
 
-  <% help = "Bypass " + [("buddy check" if BuddyCheck.enabled?), "release tracking"].compact.to_sentence %>
-  <%= form.input :no_code_deployed, as: :check_box, label: no_code_label, help: help %>
+  <% help = "Bypass " + [("buddy check" if BuddyCheck.enabled?), "release tracking"].compact.to_sentence \
+      + (", Only global admin can alter." if !current_user.admin?).to_s
```
ragurney: `,` should be a period `.`. 

`Only global admin can alter` -> `Must be a global admin to change`?

test/controllers/stages_controller_test.rb
```diff
@@ -172,6 +172,12 @@
         it 'adds no commands by default' do
           assigns(:stage).command_ids.must_equal []
         end
+
+        it 'disabled to alter `does not deploy code`' do
+          assert_select "#stage_no_code_deployed" do |input|
+            assert input.attr("disabled").present?
```
ragurney: Nice!

app/controllers/stages_controller.rb
```diff
@@ -142,6 +141,9 @@ def stage_permitted_params
         deploy_group_ids: [],
         command_ids: []
       }
-    ] + Samson::Hooks.fire(:stage_permitted_params).flatten
+    ].tap do |params|
+      params << :no_code_deployed if current_user.admin?
+      params << Samson::Hooks.fire(:stage_permitted_params)
```
grosser: - only `stage_permitted_params` needs the flatten
 - prefer assigning a variable and then adding things to it over `tap`-ing

app/controllers/stages_controller.rb
```diff
@@ -142,6 +141,9 @@ def stage_permitted_params
...
+    end.flatten
```
ragurney: I don't think we want to flatten the entire param array... move this up to L#146?

app/views/stages/_fields.html.erb
grosser: can always add that note, no matter what user level

test/controllers/stages_controller_test.rb
```diff
@@ -396,4 +402,26 @@ def clone(method, format, extra = {})
...
+      end
+    end
+
+    describe '#create' do
+      subject { assigns(:stage) }
+      it 'permits `no_code_deployed` as params' do
+        params = {project_id: projects(:test).to_param, stage: {name: 'test', no_code_deployed: true}}
+        post :create, params: params
+        subject.reload
+        subject.persisted?.must_equal(true)
```
grosser: anything that can be reloaded is persisted ... so this test does not make much sense ?

test/controllers/stages_controller_test.rb
```diff
@@ -172,6 +172,12 @@
         it 'adds no commands by default' do
           assigns(:stage).command_ids.must_equal []
         end
+
+        it 'disabled to alter `does not deploy code`' do
+          assert_select "#stage_no_code_deployed" do |input|
+            assert input.attr("disabled").present?
+          end
+        end
```
grosser: also should test that the create fails when submitting no_code_deployed

test/controllers/stages_controller_test.rb
```diff
@@ -396,4 +402,26 @@ def clone(method, format, extra = {})
...
     end
   end
+
+  as_an_admin do
+    describe '#new' do
+      it 'can alter `does not deploy code`' do
+        get :new, params: {project_id: subject.project.to_param}
+        assert_select "#stage_no_code_deployed" do |input|
+          refute input.attr("disabled").present?
+        end
```
grosser: copy-pasted, put into a method so we can assert true/false and don't get accidental diffs

grosser: like `assert stage_no_code_deployed_disabled` / `refute stage_no_code_deployed_disabled`

app/controllers/stages_controller.rb
```diff
@@ -142,6 +141,9 @@ def stage_permitted_params
         deploy_group_ids: [],
         command_ids: []
       }
-    ] + Samson::Hooks.fire(:stage_permitted_params).flatten
+    ].tap do |params|
+      params << :no_code_deployed if current_user.admin?
```
grosser: would actually be nice to do `alias_method :can_change_no_code_deployed?, :admin?` in User and then use `can_change_no_code_deployed?` everywhere so the consistency between view and controller is obvious
... alternatively do `helper_method def can_change_no_code_deployed?` in the controller and reuse in the view

app/controllers/stages_controller.rb
sathishavm: we are pushing into array in L#146? and so it will create a nested array, that's the reason we do flatten for entire params.

ragurney: I think only the result from `:stage_permitted_params` needs the flatten here..

sathishavm: `samson(dev)> [9]+[8]
=> [9, 8]
samson(dev)> [9]<<[8]
=> [9, [8]]
`

ragurney: Then why not use `concat` and avoid the flatten?

app/views/stages/_fields.html.erb
```diff
@@ -23,8 +23,10 @@
       " When selected with '#{no_code_label}' and '#{confirm_label}' is disabled the stage can be directly executed."
   %>
 
-  <% help = "Bypass " + [("buddy check" if BuddyCheck.enabled?), "release tracking"].compact.to_sentence %>
-  <%= form.input :no_code_deployed, as: :check_box, label: no_code_label, help: help %>
+  <% help = "Bypass " + [("buddy check" if BuddyCheck.enabled?), "release tracking"].compact.to_sentence \
+      + ". Must be a global admin to change"
```
ragurney: Missing period

sathishavm: it's missing for all other help text? I thought to make it consistency.

sathishavm: hmm.. will add for remaining too.

sathishavm:
@grosser +1 ?

test/controllers/stages_controller_test.rb
```diff
@@ -222,6 +233,12 @@ def create_stage(overrides = {})
         end
       end
 
+      it "fails with `no_code_deployed` in params" do
```
grosser: tests are easier to read/understand if they are less technical and more "user story" in the test cases like `it "fails when trying to set no code deployed"`

PR #2958, opened 2018-09-28
make scope tracing less overhead and use hooks to do it

grosser:
@zendesk/samson @sathishavm @ragurney

### DIFF
--- a/lib/samson/performance_tracer.rb
+++ b/lib/samson/performance_tracer.rb
@@ -2,13 +2,9 @@
 module Samson
   module PerformanceTracer
     class << self
-      # NOTE: this cannot be a hook since it is used from Hooks#fire
-      def handlers
-        @handlers ||= []
-      end
-
-      def trace_execution_scoped(scope_name, &block)
-        handlers.inject(block) { |inner, plugin| -> { plugin.trace_execution_scoped(scope_name, &inner) } }.call
+      def trace_execution_scoped(scope, &block)
+        tracers = Samson::Hooks.fire(:trace_scope, scope).compact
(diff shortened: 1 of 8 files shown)


lib/samson/performance_tracer.rb
```diff
@@ -2,13 +2,9 @@
...
-      # NOTE: this cannot be a hook since it is used from Hooks#fire
-      def handlers
-        @handlers ||= []
-      end
-
-      def trace_execution_scoped(scope_name, &block)
-        handlers.inject(block) { |inner, plugin| -> { plugin.trace_execution_scoped(scope_name, &inner) } }.call
+      def trace_execution_scoped(scope, &block)
+        tracers = Samson::Hooks.fire(:trace_scope, scope).compact
+        tracers.inject(block) { |inner, tracer| -> { tracer.call(&inner) } }.call
```
grosser: nice benefit here is that it no longer needs to know about method names

plugins/datadog_tracer/lib/samson_datadog_tracer/samson_plugin.rb
```diff
@@ -51,8 +43,11 @@ def method_visibility(klass, method)
   end
 end
 
-require 'samson/performance_tracer'
-Samson::PerformanceTracer.handlers << SamsonDatadogTracer
```
grosser: ... and here we don't need requires/accessors anymore

grosser: ... and it behaves more like a regular hook

PR #2960, opened 2018-09-28
stop complaining about pods that are in undeletable state and delete …

eatwithforks:
…them

* Add description here
* Remember to add unit tests

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -348,7 +348,8 @@ def request_delete
           update
```
adammw: what's the chance of two successive calls returning the same number?

grosser: what numbers are you talking about ?

grosser:
hmmm since we don't wait for the pods to shut down this results in old and new pods running side-by-side

```
kube-node-monitor-termination-watcher-64995dbfd8-j4b9q               0/1       Terminating   0          11m
kube-node-monitor-termination-watcher-744765ff85-pqfjj               1/1       Running       0          14s
```

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -366,12 +367,6 @@ def wait_for_termination_of_all_pods
           expire_resource_cache
           return if pods_running_count == 0
```
grosser: pods_running_count needs a rename, it's misleading

grosser:
nvm ... that's a deployment :D

PR #2966, opened 2018-10-01
[EDGE-761] STS samson plugin 

gerardc:
WIP

Adds a plugin that modifies the ENV variable with STS session values

/cc @zendesk/sledgehammer 

### References
 - Jira link:

### Risks
- Level: Low adds a plugin that modifies the ENV variable with STS session values

### DIFF
--- a/db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
+++ b/db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
@@ -1,0 +1,5 @@
+class AddAwsStsArnToStages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :stages, :aws_sts_iam_role_arn, :string
+  end
+end
--- a/plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
+++ b/plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
@@ -4,6 +4,15 @@
   </legend>
 
   <div class="col-lg-offset-2">
-  YAY
+    <div class="form-group">
+      <div class="col-lg-6">
+        <%= form.text_field :aws_sts_iam_role_arn, class: "form-control", placeholder: "IAM Role ARN" %>
+
+        <span class="help-block">
+          Will expose the following ENV variables...
+        </span>
+      </div>
+    </div>
   </div>
+
 </fieldset>
--- a/plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
+++ b/plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
@@ -1,0 +1,9 @@
+<fieldset>
+  <legend>
+    AWS STS
+  </legend>
+
+  <div class="col-lg-offset-2">
+  YAY
+  </div>
+</fieldset>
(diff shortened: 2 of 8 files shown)


db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
```diff
@@ -0,0 +1,5 @@
+class AddAwsStsArnToStages < ActiveRecord::Migration[5.2]
```
grosser: goes into plugins/*/db/migrate

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    AWS STS
+  </legend>
+
+  <div class="col-lg-offset-2">
+    <div class="form-group">
+      <div class="col-lg-6">
+        <%= form.text_field :aws_sts_iam_role_arn, class: "form-control", placeholder: "IAM Role ARN" %>
+
+        <span class="help-block">
+          Will expose the following ENV variables...
```
grosser: missing some info ?

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,38 @@
...
+  class << self
+    def set_env_vars(sts_client:, stage:)
+      creds = sts_client.assume_role(
+        role_arn: stage.aws_sts_iam_role_arn,
+        role_session_name: ROLE_SESSION_NAME
+      ).credentials
+
+      ENV['STS_AWS_ACCESS_KEY_ID']     = creds.access_key_id
+      ENV['STS_AWS_SECRET_ACCESS_KEY'] = creds.secret_access_key
+      ENV['STS_AWS_SESSION_TOKEN']     = creds.session_token
```
grosser: should not set global env vars ... they will never expire and be shared between all deploys/stagges

plugins/aws_sts/test/samson_aws_sts/samson_plugin_test.rb
```diff
@@ -0,0 +1,37 @@
...
+require_relative '../test_helper'
+
+SingleCov.covered! uncovered: 2
+
+describe SamsonAwsSts do
+  let(:stage) { stages(:test_staging) }
+
+  describe '.set_env_vars' do
+    it 'sets values against ENV' do
+      %w(
```
grosser: use the with_env helper for things like this ... can do `with_env FOO: nil` to enforce things are unset

grosser:
who will use these env vars ? ... a bit more on usecase/background would be nice too

gerardc:
@grosser thanks for the feedback, I had included "WIP" in the description as I was planning on pinging the samson team shortly. 

Yep modifying ENV is a hack. The idea is for an application to specify a role/policy ARN that grants permission for samson to write assets to the shard (and secure) static assets pipeline s3 bucket. Using Amazon STS we can grant _tempory_ access to the policy.

In short, the change exposes ENV variables that temporarily grant us access to upload assets to an S3 bucket.

I'll catch up with the team see if there's a different way we can do this without modifying ENV

grosser:
If these env vars are for the script that is deploying, they should go into job_execution and not samsons environment (samsons environment is not shared with the job execution)

gerardc:
ok thanks again, we've decided on using a bash script to do this work instead, I'll close this PR

PR #2967, opened 2018-10-01
remove commit_range env var since it is expensive and unused

grosser:
reverts https://github.com/zendesk/samson/pull/2225

@zendesk/samson @ragurney


app/models/job_execution.rb
```diff
@@ -84,13 +84,7 @@ def descriptor
   def base_commands(dir, env = {})
     artifact_cache_dir = File.join(@job.project.repository.repo_cache_dir, "artifacts")
     FileUtils.mkdir_p(artifact_cache_dir)
-
-    env = {
-      PROJECT_NAME: @job.project.name,
-      PROJECT_PERMALINK: @job.project.permalink,
-      PROJECT_REPOSITORY: @job.project.repository_url,
```
grosser: also moving these project vars into the other env foobar since they seem out of place here

PR #2970, opened 2018-10-02
allow setting deploy env vars on kubernetes deploys

grosser:
/cc @zendesk/samson @rata @nogates 

fixups for https://github.com/zendesk/samson/pull/2666


rata:
@grosser seems to be working fine, thanks! The variables are added to all containers in a pod.

I still have to iron out how the flow would be in kubernetes, but the most critical part seems solved by this!

(I need to check if running a data-migration will imply a deploy, as I think all roles are always deployed, but seems like "details" :))

Thanks a lot, really!

PR #2973, opened 2018-10-03
use faster worktree instead of cloning

grosser:
extracted from https://github.com/zendesk/samson/pull/944

@zendesk/samson @zendesk/bre @mariovisic @rata


config/initializers/git.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+raise "Git not found" unless version = `git --version`[/\d+\.\d+\.\d+/]
+raise "Need git v2.0.0+" if Gem::Version.new(version) < Gem::Version.new("2.0.0")
```
rata: I think worktree was added in git 2.5.0, if I searched correctly. See: https://github.com/git/git/blob/master/Documentation/RelNotes/2.5.0.txt#L45-L47

Also, see GitHub posts: https://blog.github.com/2015-07-29-git-2-5-including-multiple-worktrees-and-triangular-workflows/ The footnote says that the contrib script was improved to become the worktree subcommand.

But maybe I'm missing something?

Although, the changeling says it's experimental and it might change. Maybe it's safer to check for newer versions, as not all subcommands used might have been implemented by 2.5.0. But might probably be correct.

I can try to test it with git 2.5.0, if you want (and I find a git 2.5.0 build easy to install :D)

grosser:
yeah looks like we need 2.5.0 ... I thought I found it on 2.0 but that's wrong ...
I'd prefer to not use `--detach` since that makes `git br` not work in the checkout

app/models/git_repository.rb
```diff
@@ -143,25 +142,23 @@ def sha_exist?(sha)
     !!capture_stdout("git", "cat-file", "-t", sha)
   end
 
-  def checkout(git_reference, pwd)
-    executor.execute("cd #{pwd}", "git checkout --quiet #{git_reference.shellescape}")
+  def checkout(git_reference, work_dir)
+    executor.execute(
+      "cd #{repo_cache_dir}",
```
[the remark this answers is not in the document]
grosser: yes, it is created before this step

app/models/git_repository.rb
```diff
@@ -91,6 +90,11 @@ def update_mirror
     exclusive { (mirrored? ? update! : clone!) }
   end
 
+  # clear worktrees that are deleted
+  def prune_worktree
+    executor.execute("cd #{repo_cache_dir}", "git worktree prune")
+  end
```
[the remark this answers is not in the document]
grosser: the checkout is deleted after a job is done ... the current code-flow is kinda funky that way ... I think checkout should maybe yield to create the tempfile instead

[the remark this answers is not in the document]
grosser: ... that's for another PR, trying to keep this simple for now :)

grosser:
benchmark with samson itself showed 1s vs 0.5s for worktree, I'd expect that to be even better for larger repos

app/models/git_repository.rb
```diff
@@ -24,9 +24,8 @@ def checkout_workspace(work_dir, git_reference)
     executor.output.write("# Beginning git repo setup\n")
 
     ensure_mirror_current &&
-    create_workspace(work_dir) &&
-    checkout(git_reference, work_dir) &&
-    checkout_submodules(work_dir)
+      checkout(git_reference, work_dir) &&
```
ragurney: Could put all of these on one line?

config/initializers/git.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+min = "2.5.0"
```
ragurney: Constant?

app/models/job_execution.rb
```diff
@@ -279,5 +280,7 @@ def make_tempdir
   rescue Errno::ENOTEMPTY, Errno::ENOENT
     ErrorNotifier.notify("Notify: make_tempdir error #{$!.message.split('@').first}")
     result # tempdir ensure sometimes fails ... not sure why ... return normally
+  ensure
+    @repository.prune_worktree
```
ragurney: Is there a test for this?

config/initializers/git.rb
grosser: not reused, so prefer not adding anything to the global namespace

app/models/git_repository.rb
grosser: would get kinda crowded and the alternative is `return unless foo; return unless bar` 🤷‍♂️

app/models/job_execution.rb
grosser: added

PR #2974, opened 2018-10-04
Add STS env vars to deploy

gerardc:
Adds a plugin to allow Samson to inject temporary security credentials for an AWS Role.

The plugin uses the [assume role](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) feature of STS and exposes the generated credentials as environment variables:
  - STS_AWS_ACCESS_KEY_ID
  - STS_AWS_SECRET_ACCESS_KEY
  - STS_AWS_SESSION_TOKEN

Specify the Amazon Resource Name (ARN) of the role to assume in the stage's settings page.

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/models/terminal_executor.rb
+++ b/app/models/terminal_executor.rb
@@ -15,6 +15,9 @@ require 'pty'
 #
 class TerminalExecutor
   SECRET_PREFIX = "secret://"
+  HIDDEN_PREFIX = "hidden://"
+  HIDDEN_TXT = "HIDDEN"
+
   CURSOR = /\e\[\d*[ABCDK]/
 
   attr_reader :pid, :pgid, :output, :timeout
@@ -131,7 +134,19 @@ class TerminalExecutor
   end
 
   def print_and_execute(c, resolve: true)
-    "echo » #{c.shellescape}\n#{resolve ? resolve_secrets(c) : c}"
+    print_and_execute_hidden_command(c) ||
+      print_and_execute_raw(c, resolve ? resolve_secrets(c) : c)
+  end
+
+  def print_and_execute_hidden_command(c)
+    # hides secrets by replacing lines like export FOO="hidden://secret" into export FOO="HIDDEN"
+    return unless print_command = c.dup.sub!(/#{HIDDEN_PREFIX}[^"]+/, HIDDEN_TXT)
+    resolved_command = c.sub(/#{HIDDEN_PREFIX}/, '')
+    print_and_execute_raw(print_command, resolved_command)
+  end
+
+  def print_and_execute_raw(print_command, real_command)
+    "echo » #{print_command.shellescape}\n#{real_command}"
   end
 
   def resolve_secrets(command)
--- a/plugins/aws_sts/README.md
+++ b/plugins/aws_sts/README.md
@@ -1,0 +1,18 @@
+# AWS STS Plugin
+
+Inject temporary AWS Role credentials into a deploy environment using [AWS STS](https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html) and [Assume role](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html).
+
+## Overview
+
+The plugin uses the assume role feature of STS and exposes the generated credentials
+as environment variables:
+  - STS_AWS_ACCESS_KEY_ID
+  - STS_AWS_SECRET_ACCESS_KEY
+  - STS_AWS_SESSION_TOKEN
+
+The plugin authenticates with AWS using the following samson environment variables:
+  - SAMSON_STS_AWS_ACCESS_KEY_ID
+  - SAMSON_STS_AWS_SECRET_ACCESS_KEY
+  - SAMSON_STS_AWS_REGION
+
+Specify the Amazon Resource Name (ARN) of the role to assume in the stage's settings page.
--- a/plugins/aws_sts/app/decorators/stage_decorator.rb
+++ b/plugins/aws_sts/app/decorators/stage_decorator.rb
@@ -29,7 +29,7 @@ Stage.class_eval do
         client.assume_role(
           role_arn: aws_sts_iam_role_arn,
           role_session_name: "validate_can_assume_role_#{SecureRandom.hex(4)}",
-          duration_seconds: 10
+          duration_seconds: SamsonAwsSts::SESSION_DURATION_MIN
         )
       rescue => e
         errors.add(:aws_sts_iam_role_arn, "Unable to assume role: #{e.message}")
--- a/plugins/aws_sts/app/decorators/stage_decorator.rb
+++ b/plugins/aws_sts/app/decorators/stage_decorator.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+Stage.class_eval do
+  validates(
+    :aws_sts_iam_role_arn,
+    allow_blank: true,
+    format: {
+      with: /\Aarn:aws:iam::\d+:role\/.+\Z/,
+      message: "Must be of the form: arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME"
+    }
+  )
+
+  validates(
+    :aws_sts_iam_role_session_duration,
+    allow_blank: true,
+    numericality: {
+      only_integer: true,
+      greater_than_or_equal_to: SamsonAwsSts::SESSION_DURATION_MIN,
+      less_than_or_equal_to:    SamsonAwsSts::SESSION_DURATION_MAX
+    }
+  )
+
+  validate :validate_can_assume_role, if: -> { aws_sts_iam_role_arn.present? }
+
+  before_validation :set_default_session_duration
+
+  private
+
+  def set_default_session_duration
+    self.aws_sts_iam_role_session_duration ||= SamsonAwsSts::SESSION_DURATION_MIN
+  end
+
+  def validate_can_assume_role
+    SamsonAwsSts::Client.new(SamsonAwsSts.sts_client).assume_role(
+      role_arn: aws_sts_iam_role_arn,
+      role_session_name: "validate_can_assume_role_#{SecureRandom.hex(4)}"
+    )
+  rescue => e
+    errors.add(:aws_sts_iam_role_arn, "Unable to assume role: #{e.message}")
+  end
+end
--- a/plugins/aws_sts/app/decorators/stage_decorator.rb
+++ b/plugins/aws_sts/app/decorators/stage_decorator.rb
@@ -19,9 +19,8 @@ Stage.class_eval do
     }
   )
 
-  validate :validate_can_assume_role, if: -> { aws_sts_iam_role_arn.present? }
-
   before_validation :set_default_session_duration
+  validate :validate_can_assume_role, if: :aws_sts_iam_role_arn?
 
   private
 
--- a/plugins/aws_sts/app/decorators/stage_decorator.rb
+++ b/plugins/aws_sts/app/decorators/stage_decorator.rb
@@ -19,21 +19,23 @@ Stage.class_eval do
     }
   )
 
-  before_validation :set_default_session_duration
   validate :validate_can_assume_role, if: :aws_sts_iam_role_arn?
 
   private
 
-  def set_default_session_duration
-    self.aws_sts_iam_role_session_duration ||= SamsonAwsSts::SESSION_DURATION_MIN
-  end
-
   def validate_can_assume_role
-    SamsonAwsSts::Client.new(SamsonAwsSts.sts_client).assume_role(
-      role_arn: aws_sts_iam_role_arn,
-      role_session_name: "validate_can_assume_role_#{SecureRandom.hex(4)}"
-    )
-  rescue => e
-    errors.add(:aws_sts_iam_role_arn, "Unable to assume role: #{e.message}")
+    if client = SamsonAwsSts.sts_client
+      begin
+        client.assume_role(
+          role_arn: aws_sts_iam_role_arn,
+          role_session_name: "validate_can_assume_role_#{SecureRandom.hex(4)}",
+          duration_seconds: 10
+        )
+      rescue => e
+        errors.add(:aws_sts_iam_role_arn, "Unable to assume role: #{e.message}")
+      end
+    else
+      errors.add(:aws_sts_iam_role_arn, "SAMSON_STS_AWS_* env vars not set")
+    end
   end
 end
--- a/plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
+++ b/plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
@@ -3,7 +3,21 @@
   <div class="col-lg-offset-2">
     <div class="form-group">
       <div class="col-lg-6">
-        <p>Tokens will be requested using the following user: <%= SamsonAwsSts::Client.new(SamsonAwsSts.sts_client).caller_user_id rescue "error connecting to aws, check SAMSON_STS_AWS_* env vars" %></p>
+        <p>
+          <% if client = SamsonAwsSts.sts_client %>
+            Tokens will be requested using the following user:
+            <%= Rails.cache.fetch("sts-caller-user-id-#{Process.pid}") do
+              begin
+                client.caller_user_id
+              rescue => e
+                "error connecting to aws, check SAMSON_STS_AWS_* env vars (#{e})"
+              end
+            end
+            %>
+          <% else %>
+            SAMSON_STS_AWS_* env vars not set.
+          <% end %>
+        </p>
         <span class="help-block">
           Use the following samson environment variables to change which user to authenticate with:
           <ul>
@@ -31,7 +45,7 @@
     <div class="form-group">
       <div class="col-lg-6">
         <p>The duration, in seconds, that the credentials should remain valid:</p>
-        <%= form.number_field :aws_sts_iam_role_session_duration, class: "form-control", placeholder: 900 %>
+        <%= form.text_field :aws_sts_iam_role_session_duration, class: "form-control", placeholder: 900 %>
         <span class="help-block">
           Optional. Defaults to <%= SamsonAwsSts::SESSION_DURATION_MIN %>.
           <ul>
--- a/plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
+++ b/plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
@@ -1,0 +1,45 @@
+<fieldset>
+  <legend>AWS STS Tokens</legend>
+  <div class="col-lg-offset-2">
+    <div class="form-group">
+      <div class="col-lg-6">
+        <p>Tokens will be requested using the following user: <%= SamsonAwsSts::Client.new(SamsonAwsSts.sts_client).caller_user_id rescue "error connecting to aws, check SAMSON_STS_AWS_* env vars" %></p>
+        <span class="help-block">
+          Use the following samson environment variables to change which user to authenticate with:
+          <ul>
+            <li>SAMSON_STS_AWS_ACCESS_KEY_ID</li>
+            <li>SAMSON_STS_AWS_SECRET_ACCESS_KEY</li>
+            <li>SAMSON_STS_AWS_REGION</li>
+          </ul>
+        </span>
+      </div>
+    </div>
+    <div class="form-group">
+      <div class="col-lg-6">
+        <p>AWS IAM Role ARN that should be assumed by the stage:</p>
+        <%= form.text_field :aws_sts_iam_role_arn, class: "form-control", placeholder: "e.g. arn:aws:iam::123456789012:role/S3Access" %>
+        <span class="help-block">
+          This will create the following environment variables which can be used by stage commands:
+          <ul>
+            <li>STS_AWS_ACCESS_KEY_ID</li>
+            <li>STS_AWS_SECRET_ACCESS_KEY</li>
+            <li>STS_AWS_SESSION_TOKEN</li>
+          </ul>
+        </span>
+      </div>
+    </div>
+    <div class="form-group">
+      <div class="col-lg-6">
+        <p>The duration, in seconds, that the credentials should remain valid:</p>
+        <%= form.number_field :aws_sts_iam_role_session_duration, class: "form-control", placeholder: 900 %>
+        <span class="help-block">
+          Optional. Defaults to <%= SamsonAwsSts::SESSION_DURATION_MIN %>.
+          <ul>
+            <li>Min: <%= SamsonAwsSts::SESSION_DURATION_MIN %> (<%= SamsonAwsSts::SESSION_DURATION_MIN / 60%> minutes)</li>
+            <li>Max: <%= SamsonAwsSts::SESSION_DURATION_MAX %> (<%= SamsonAwsSts::SESSION_DURATION_MAX / 60%> minutes)</li>
+          </ul>
+        </span>
+      </div>
+    </div>
+  </div>
+</fieldset>
--- a/plugins/aws_sts/db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
+++ b/plugins/aws_sts/db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddAwsStsArnToStages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :stages, :aws_sts_iam_role_arn, :string
+    add_column :stages, :aws_sts_iam_role_session_duration, :integer, default: 900, null: false
+  end
+end
--- a/plugins/aws_sts/db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
+++ b/plugins/aws_sts/db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
@@ -3,6 +3,6 @@
 class AddAwsStsArnToStages < ActiveRecord::Migration[5.2]
   def change
     add_column :stages, :aws_sts_iam_role_arn, :string
-    add_column :stages, :aws_sts_iam_role_session_duration, :integer, default: 900, null: false
+    add_column :stages, :aws_sts_iam_role_session_duration, :integer
   end
 end
--- a/plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
+++ b/plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
@@ -2,7 +2,7 @@
 
 module SamsonAwsSts
   SESSION_DURATION_MIN = 900 # 15 minutes
-  SESSION_DURATION_MAX = Rails.application.config.samson.deploy_timeout
+  SESSION_DURATION_MAX = [SESSION_DURATION_MIN, Rails.application.config.samson.deploy_timeout].max
 
   class Engine < Rails::Engine
   end
@@ -17,7 +17,7 @@ module SamsonAwsSts
 
     options[:stub_responses] = Rails.env.test?
 
-    Aws::STS::Client.new(options)
+    Client.new(Aws::STS::Client.new(options))
   end
 
   class Client
@@ -26,10 +26,10 @@ module SamsonAwsSts
     end
 
     def caller_user_id
-      sts_client.get_caller_identity[:user_id]
+      @sts_client.get_caller_identity[:user_id]
     end
 
-    def deploy_env_vars(deploy:)
+    def deploy_env_vars(deploy)
       creds = assume_role(
         role_arn: deploy.stage.aws_sts_iam_role_arn,
         role_session_name: session_name(deploy),
@@ -43,8 +43,8 @@ module SamsonAwsSts
       }
     end
 
-    def assume_role(role_arn:, role_session_name:, duration_seconds: SESSION_DURATION_MIN)
-      sts_client.assume_role(
+    def assume_role(role_arn:, role_session_name:, duration_seconds:)
+      @sts_client.assume_role(
         role_arn: role_arn,
         role_session_name: role_session_name,
         duration_seconds: duration_seconds
@@ -52,8 +52,6 @@ module SamsonAwsSts
     end
 
     private
-
-    attr_reader :sts_client
 
     def session_name(deploy)
       project_name = deploy.project.name.parameterize[0..15].tr('_', '-')
@@ -76,5 +74,5 @@ end
 Samson::Hooks.callback :deploy_env do |deploy|
   next {} if deploy.stage.aws_sts_iam_role_arn.blank?
 
-  SamsonAwsSts::Client.new(SamsonAwsSts.sts_client).deploy_env_vars(deploy: deploy)
+  SamsonAwsSts::Client.new(SamsonAwsSts.sts_client).deploy_env_vars(deploy)
 end
--- a/plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
+++ b/plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+module SamsonAwsSts
+  SESSION_DURATION_MIN = 900 # 15 minutes
+  SESSION_DURATION_MAX = Rails.application.config.samson.deploy_timeout
+
+  class Engine < Rails::Engine
+  end
+
+  def self.sts_client
+    options = {
+      access_key_id:     ENV['SAMSON_STS_AWS_ACCESS_KEY_ID'],
+      secret_access_key: ENV['SAMSON_STS_AWS_SECRET_ACCESS_KEY'],
+      region:            ENV['SAMSON_STS_AWS_REGION'],
+    }
+    return if options.any? { |_, v| v.blank? }
+
+    options[:stub_responses] = Rails.env.test?
+
+    Aws::STS::Client.new(options)
+  end
+
+  class Client
+    def initialize(sts_client)
+      @sts_client = sts_client
+    end
+
+    def caller_user_id
+      sts_client.get_caller_identity[:user_id]
+    end
+
+    def deploy_env_vars(deploy:)
+      creds = assume_role(
+        role_arn: deploy.stage.aws_sts_iam_role_arn,
+        role_session_name: session_name(deploy),
+        duration_seconds: deploy.stage.aws_sts_iam_role_session_duration || SESSION_DURATION_MIN
+      ).credentials
+
+      {
+        STS_AWS_ACCESS_KEY_ID:     "hidden://#{creds.access_key_id}",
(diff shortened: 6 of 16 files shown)


gerardc:
@grosser you beat me to the fix for `job_additional_vars` 😎

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,53 @@
...
+Samson::Hooks.callback :stage_permitted_params do
+  [
+    :aws_sts_iam_role_arn
+  ]
+end
+
+Samson::Hooks.callback :deploy_env do |deploy|
+  options = {}
+  if Rails.env.test?
+    options[:stub_responses] = true
```
grosser: can do `{stub_responses: Rails.env.test?}`

plugins/aws_sts/samson_aws_sts.gemspec
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+Gem::Specification.new 'samson_aws_sts', '0.0.0' do |s|
+  s.summary = 'Samson AwsSts plugin'
+  s.authors = ['Gerard Cahill']
+  s.email = ['gcahill@zendesk.com']
+  s.files = Dir['{app,config,db,lib}/**/*']
```
grosser: `add_runtime_dependency 'aws-sdk-core'`

plugins/aws_sts/test/samson_aws_sts/samson_plugin_test.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.covered! uncovered: 5
```
grosser: how about 0 ? :trollface:

grosser:
that looks about right logic wise.

with this all deploys share the same aws credentials (samsons) which is kinda fishy ...
it would be better if the project would have it's own credentials instead

using `Samson::Secrets::KeyResolver.new(project, deploy_groups).read('aws_access_key_id')`
would be good for that (and fail descriptively if the secret does not exist)

gerardc:
@grosser feedback addressed if you're happy to take another look?

plugins/aws_sts/test/samson_aws_sts/samson_plugin_test.rb
gerardc: all the important stuff is tested ;)

gerardc:
@grosser regarding using `Samson::Secrets::KeyResolver.new(project, deploy_groups).read('aws_access_key_id')` the thinking behind using Samson's AWS key is for self service, i.e. all the deployer needs to specify is the ARN of the amazon role and samson takes care of the rest. Let me know what you think

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      {
+        STS_AWS_ACCESS_KEY_ID:     "hidden://#{creds.access_key_id}",
+        STS_AWS_SECRET_ACCESS_KEY: "hidden://#{creds.secret_access_key}",
+        STS_AWS_SESSION_TOKEN:     "hidden://#{creds.session_token}"
+      }
+    end
+
+    def session_name(deploy)
+      "#{deploy.project.name}:#{deploy.stage.name}:deploy/#{deploy.id}"
```
grosser: project and stage names can have all kinds of characters/lengths ... better check the docs on what a valid name is, then `.parameterize` and cut down to length if needed

plugins/aws_sts/test/samson_aws_sts/samson_plugin_test.rb
grosser: Still prefer 0, the other plugins can serve as a copy-paste example, should be pretty straight forward, if you run into any trouble let me know and I'll help out

plugins/aws_sts/test/samson_aws_sts/samson_plugin_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  let(:stage) { stages(:test_staging) }
+
+  describe '.deploy_env_vars' do
+    it 'returns a hash of additional env variables' do
+      stage.aws_sts_iam_role_arn = "some_arn"
+
+      result = SamsonAwsSts.deploy_env_vars(
+        sts_client: Aws::STS::Client.new(stub_responses: true),
+        deploy: stage.deploys.last
+      )
```
grosser: turn into a method/let instead of copy-pasting so it stays in sync

plugins/aws_sts/test/samson_aws_sts/samson_plugin_test.rb
```diff
@@ -0,0 +1,35 @@
...
+      result.must_equal(
+        STS_AWS_ACCESS_KEY_ID:     'hidden://accessKeyIdType',
+        STS_AWS_SECRET_ACCESS_KEY: 'hidden://accessKeySecretType',
+        STS_AWS_SESSION_TOKEN:     'hidden://tokenType'
+      )
+    end
+  end
+
+  describe 'for a stage with a blank ARN' do
+    it 'returns an empty hash' do
```
grosser: can simplify the nesting to:
`it 'returns nothing for a stage with a blank ARN' do`

plugins/deploy_env_vars/test/samson_deploy_env_vars/samson_plugin_test.rb
```diff
@@ -18,15 +18,13 @@
   end
 
   describe :deploy_env do
-    it "returns an empty hash" do
-      Samson::Hooks.fire(:deploy_env, deploy).must_equal([{}])
```
grosser: simpler to use `only_callbacks_for_plugin` so other plugins don't break this test again

app/models/terminal_executor.rb
```diff
@@ -131,7 +134,18 @@ def script(commands)
...
+      "echo » #{c.shellescape}\n#{resolve ? resolve_secrets(c) : c}"
+    end
+  end
+
+  def print_and_execute_hidden_command(c)
+    # Replace `hidden://abcd` up until the first doublequote with HIDDEN_TXT
+    hidden_command   = c.sub(/#{HIDDEN_PREFIX}[^"]+/, HIDDEN_TXT)
+    resolved_command = c.sub(/#{HIDDEN_PREFIX}/, '')
+    "echo » #{hidden_command.shellescape}\n#{resolved_command}"
   end
```
grosser: - echo logic is duplicated in 2 places
 - hidden logic is checked in 2 spaces
 - initial hidden check might not match hidden replaement

```Ruby
def print_and_execute(c, resolve: true)
  print_and_execute_hidden_command(c) ||
    print_and_execute_raw(c, resolve ? resolve_secrets(c) : c)
end

def print_and_execute_hidden_command(c)
  return unless hidden_command = c.dup.sub!(/#{HIDDEN_PREFIX}[^"]+/, HIDDEN_TXT)
  resolved_command = c.sub(HIDDEN_PREFIX, '')
  print_and_execute_raw hidden_command, resolved_command
end

def print_and_execute_raw(print_command, real_command)
  "echo » #{print_command.shellescape}\n#{real_command}"
end
```

grosser:
@ragurney @jonmoter thoughts on allowing users to assume any role via samsons aws credentials ?

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,51 @@
...
+  class << self
+    def deploy_env_vars(sts_client:, deploy:)
+      if deploy.stage.aws_sts_iam_role_arn.blank?
+        return {}
+      end
+
+      creds = sts_client.assume_role(
+        role_arn: deploy.stage.aws_sts_iam_role_arn,
+        role_session_name: session_name(deploy)
+      ).credentials
```
jonmoter: You might want a mechanism to set `duration_seconds` for this call. Either to allow for shorter-lived credentials for security, or for long-lived credentials if your deploy takes an exceptionally long time.

app/models/terminal_executor.rb
```diff
@@ -131,7 +134,18 @@ def script(commands)
...
+    if c.include? HIDDEN_PREFIX
+      print_and_execute_hidden_command(c)
+    else
+      "echo » #{c.shellescape}\n#{resolve ? resolve_secrets(c) : c}"
+    end
+  end
+
+  def print_and_execute_hidden_command(c)
+    # Replace `hidden://abcd` up until the first doublequote with HIDDEN_TXT
+    hidden_command   = c.sub(/#{HIDDEN_PREFIX}[^"]+/, HIDDEN_TXT)
```
jonmoter: Are we sure that double-quotes isn't a valid character in an STS token?

plugins/aws_sts/README.md
```diff
@@ -0,0 +1,14 @@
+# AWS STS Plugin
+
+This plugin allows integration between Samson and [AWS STS](https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html).
+
+## Overview
+
+Allows Samson to inject temporary security credentials for an AWS Role. The plugin uses
```
jonmoter: What are the credentials that Samson is using to authenticate, and then assume a different role? Is the assumption that ENV vars are set in the Samson process itself? And that Samson has permission to assume into the role specified?

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
```diff
@@ -0,0 +1,17 @@
+<fieldset>
+  <legend>AWS STS Tokens</legend>
+  <div class="col-lg-offset-2">
+    <div class="form-group">
+      <div class="col-lg-6">
+        <p>AWS IAM Role ARN that should be assumed by the stage:</p>
+        <%= form.text_field :aws_sts_iam_role_arn, class: "form-control", placeholder: "e.g. arn:aws:iam::123456789012:role/S3Access" %>
```
jonmoter: Does this get validated in any way? How would a user know if the ARN is valid, or whether Samson has permission to assume into that role?

app/models/terminal_executor.rb
gerardc: Good catch with the double hidden check, frozen strings led me down that route but the `dup.sub!` approach is really nice.

plugins/aws_sts/README.md
gerardc: @jonmoter Samson will use it's own AWS_ACCESS_KEY/SECRET to assume the role referenced by the ARN. The assumption is that Samson _should_ have permission to assume the role.

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
gerardc: I haven't included validation. At run time we get the following exceptions:

`Aws::STS::Client.new.assume_role(role_arn: "1234", role_session_name: "nope")` results in: 
`Aws::STS::Errors::ValidationError: 1 validation error detected: Value '1234' at 'roleArn' failed to satisfy constraint: Member must have length greater than or equal to 20
from /usr/local/bundle/gems/aws-sdk-core-2.10.21/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'`

`Aws::STS::Client.new.assume_role(role_arn: "arn:aws:iam::589470541234:user/samson-sts-test-user", role_session_name: "nope")` results in:
`Aws::STS::Errors::AccessDenied: User arn:aws:iam::589470541234:user/some_user is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::589470541234:user/samson-sts-test-user`

I guess we could validate that the ARN "looks like" an ARN, and allow the deploy to fail and surface the exception if there is a permissions error

grosser:
I'd rather not do the validation (easy to get it wrong / aws has that logic
built in anyway / it will fail for other reasons too) but make the deploy
fail nicely by maybe rescuing and re-raising the error as
Samson::Hooks::UserError

gerardc:
@grosser @jonmoter feedback implemented, ready for re-review 🙇

app/models/terminal_executor.rb
```diff
@@ -131,7 +134,19 @@ def script(commands)
   end
 
   def print_and_execute(c, resolve: true)
-    "echo » #{c.shellescape}\n#{resolve ? resolve_secrets(c) : c}"
+    print_and_execute_hidden_command(c) ||
+      print_and_execute_raw(c, resolve ? resolve_secrets(c) : c)
+  end
+
+  def print_and_execute_hidden_command(c)
+    # Replace `hidden://abcd` up until the first doublequote with HIDDEN_TXT
```
grosser: prefer "why it does it" over "what it does"
`# hides secrets by replacing lines like export FOO="hidden://secret" into export FOO="HIDDEN"`

plugins/aws_sts/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+Stage.class_eval do
+  validate :validate_arn_syntax, if: -> { self&.aws_sts_iam_role_arn.present? }
+  validates :aws_sts_iam_role_session_duration, allow_blank: true,
+                                                numericality: {
```
grosser: so much wasted whitespace, 2-space indent plz :D 
something like:
```
validates(
  :aws_sts_iam_role_session_duration, 
  allow_blank: true,
  foo: bar
)
```

plugins/aws_sts/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+Stage.class_eval do
+  validate :validate_arn_syntax, if: -> { self&.aws_sts_iam_role_arn.present? }
```
grosser: `validates : validate_arn_syntax, format: /\Aarn:aws:iam::\d+:role\/.+\Z/, allow_blank: true`

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
grosser: you can also do `pattern: /\Aarn:aws:iam::\d+:role\/.+\Z/` here and not bother with the model validation :)

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      </div>
+    </div>
+    <div class="form-group">
+      <div class="col-lg-6">
+        <p>The duration, in seconds, that the credentials should remain valid:</p>
+        <%= form.number_field :aws_sts_iam_role_session_duration, class: "form-control", placeholder: 900 %>
+        <span class="help-block">
+          Optional. Defaults to 900.
+          <ul>
+            <li>Min: 900 (15 minutes)</li>
```
grosser: use the constants from the plugin here to keep it in sync `"#{min} (#{min / 60} minutes)"`

plugins/aws_sts/db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddAwsStsArnToStages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :stages, :aws_sts_iam_role_arn, :string
+    add_column :stages, :aws_sts_iam_role_session_duration, :integer
```
grosser: would prefer not having nil numbers, maybe `default: 900, null: false`

grosser: if not, then make sure that saving `""` produces nil and not 0 ...

plugins/aws_sts/test/samson_aws_sts/samson_plugin_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    it 'returns an empty hash for a stage with a blank ARN' do
+      deploy_env_vars(stage).must_equal({})
+    end
+  end
+
+  describe 'stage_permitted_params callback' do
+    it 'returns attributes used by the plugin' do
+      Samson::Hooks.only_callbacks_for_plugin('samson_aws_sts', :stage_permitted_params) do
+        Samson::Hooks.fire(:stage_permitted_params).must_equal([%i[
```
grosser: 2-space here too plz

plugins/aws_sts/README.md
grosser: add a note here that `Samson has to be authenticated to aws for this to work, for example AWS_ACCESS_KEY/SECRET`

jonmoter: How do we know what IAM user Samson authenticates as? Is that shown in the UI somewhere?

grosser:
it's not, but we can open a console in staging/production to find out

jonmoter:
I worry this is going to become another feature that becomes a huge pain to manage. Some team is going to come up to us, saying they want to assume into a role, and there will be a conversation like:

Them: "We need to assume into an Amazon role to do our code. How should we do that?"
Us: "Use a role that you can assume into."
Them: "What should the role be?"
Us: "I dunno. Ask the Cloud Team."
Them: "The Cloud Team said to talk to you. They need to know what user Samson is authenticating as."
Us: "Actually, I don't know. There's just an secret keypair used to authenticate."

Remember all the pain we have dealing with IAM users and assumed roles in sfn_kubernetes? I'd rather not repeat that in Samson.

So if we implement this feature, I want this to have really good UI.  Like something that displays what user Samson has authenticated as, validates the assumed role works on input (not on deploy), ideally shows what permissions that assumed role has. Have some mechanism to share a role across multiple projects and/or stages. Heck, maybe have some generalized mechanism for Samson to implement auth profiles, so you can enter and manage credentials for AWS or Artifactory, and have them managed centrally.

You're probably saying "wow, that sounds like a lot of work for a little feature where I want to assume as another user..." and you'd be right. But the BRE and Compute teams are the ones who will be responsible for managing all this if more users start relying on it.

jonmoter:
> it's not, but we can open a console in staging/production to find out [what the IAM user is]

That's an automatic 👎 for me. Rails console should not be required to configure a stage in Samson.

grosser:
We can simplify this by making the project supply it's own aws credentials
via secrets.
... but then the project could also just do the auth itself via a script
instead of needing samson for that ...

jonmoter:
At a high level, I like the idea of Samson having auth credentials of its own, and some mechanism to add extra credentials on a per-project basis.  Developers shouldn't have to worry about authenticating with AWS, in the same way they don't have to worry about Samson authenticating with the Kubernetes clusters.

I'm just worried about doing a half-assed job at it, and ending up with a confusing credentials management mess 6 months down the road.

grosser:
Then we could make shared global secrets that users can select by
specifying the secrets path in the stage config.

plugins/aws_sts/app/decorators/stage_decorator.rb
gerardc: I know, I wasn't going to argue with rubocop :P I'll see what it thinks when I use brackets

plugins/aws_sts/db/migrate/20181001083613_add_aws_sts_arn_to_stages.rb
gerardc: Add just `default: 900, null: false` to the migration results in: `SQLite3::ConstraintException: NOT NULL constraint failed` when a nil value is submitted in the form - which is surprising, I expected the DB to apply the default value. So I've added a before save to set the default too.

gerardc:
Thanks for the feedback. How about using the approach @grosser suggested above?

We could call the STS assume role function with an AWS key/secret dedicated to the plugin. 
Then depending on deploy, the _value_ of the secret would change according to deploy group for example. 

These “global” secrets could be:
 - SAMSON_STS_AWS_ACCESS_KEY_ID
 - SAMSON_STS_AWS_ACCESS_SECRET_ACCESS_KEY
 - SAMSON_STS_AWS_REGION

Let me know what you think @jonmoter

garrethmcdaid-zd:
> At a high level, I like the idea of Samson having auth credentials of its own, and some mechanism to add extra credentials on a per-project basis. Developers shouldn't have to worry about authenticating with AWS, in the same way they don't have to worry about Samson authenticating with the Kubernetes clusters.
> 
> I'm just worried about doing a half-assed job at it, and ending up with a confusing credentials management mess 6 months down the road.

To some extent, that's where we are right now. For example, we have ~20 projects using Static Assets Pipeline, each of which has a different AWS S3 key/secret for Production and Staging, none of which are being rotated/tracked. The primary driver behind the plugin is to make keys/secrets ephemeral, to remove the administrative burden and security vulnerability that inevitably arises from this.

jonmoter:
I think this works okay as a feature. Though it's my preference to have some improvements to the UI. For example:

* on the input form where you can enter the role, it should say something like "Samson is authenticated to AWS using the IAM user/role [name of user]. To allow this feature, enter the ARN of an IAM user that can be assumed into. For more information see ____."
* immediately upon entering/changing that value, Samson should attempt to assume into that role, as part of validation. If that fails, do not save the change.  (If you save and don't check until deploy time, you have effectively broken that stage.)

garrethmcdaid-zd:
> * immediately upon entering/changing that value, Samson should attempt to assume into that role, as part of validation. If that fails, do not save the change.  (If you save and don't check until deploy time, you have effectively broken that stage.)

The role is not immutable. Whereas Samson could do a validation check on initial input, the role ARN becomes a saved configuration value in the stage. If it changes, or disappears, or Samson's permissions to use it change after that point, that will break the stage. This is unavoidable when dealing with AWS access and authentication.

grosser:
I'd assume whoever changes that config would test it by deploying right
after anyway ...
also very few people would use it as first, so we can ship a simpler
version before we ad UI niceties.
Only thing we need to get right is the storage/architecture so not everyone
who uses is will get broken when we change it :)

gerardc:
@grosser @jonmoter ready for re-review

I've added backend validation which asserts that the user can access the role, and I've explained in the UI which user (or set of credentials) is being used when generating the tokens

app/models/terminal_executor.rb
```diff
@@ -131,7 +134,19 @@ def script(commands)
...
 
   def print_and_execute(c, resolve: true)
-    "echo » #{c.shellescape}\n#{resolve ? resolve_secrets(c) : c}"
+    print_and_execute_hidden_command(c) ||
+      print_and_execute_raw(c, resolve ? resolve_secrets(c) : c)
+  end
+
+  def print_and_execute_hidden_command(c)
+    # hides secrets by replacing lines like export FOO="hidden://secret" into export FOO="HIDDEN"
+    return unless print_command = c.dup.sub!(/#{HIDDEN_PREFIX}[^"]+/, HIDDEN_TXT)
```
grosser: looking at this again I think it needs to get better eventually since some secrets will include `"`, maybe `hidden://fooo//`, but as long as it's only used internally thats good enough for now ...

plugins/aws_sts/README.md
grosser: first sentence is repeat of the title ... can be removed ?

plugins/aws_sts/README.md
```diff
@@ -0,0 +1,20 @@
...
+the assume role feature of STS and exposes the generated credentials as environment variables:
+  - STS_AWS_ACCESS_KEY_ID
+  - STS_AWS_SECRET_ACCESS_KEY
+  - STS_AWS_SESSION_TOKEN
+
+The plugin authenticates with AWS using the following environment variables:
```
grosser: maybe say `samson environment variables` to make it clear that they are coming from samsons env

plugins/aws_sts/README.md
```diff
@@ -0,0 +1,20 @@
...
+  - SAMSON_STS_AWS_ACCESS_KEY_ID
+  - SAMSON_STS_AWS_SECRET_ACCESS_KEY
+  - SAMSON_STS_AWS_REGION
+
+These should be configured in the secret manager against the deploy group or stage using the plugin.
```
grosser: this sounds like they would be configured per project ... maybe just remove this part and hope people know how to add secrets to samson :)

plugins/aws_sts/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,36 @@
...
+    :aws_sts_iam_role_session_duration,
+    allow_blank: true,
+    numericality: {
+      only_integer: true,
+      greater_than_or_equal_to: SamsonAwsSts::SESSION_DURATION_MIN,
+      less_than_or_equal_to:    SamsonAwsSts::SESSION_DURATION_MAX
+    }
+  )
+
+  validate :can_assume_role, if: -> { self&.aws_sts_iam_role_arn.present? }
```
grosser: self can be nil ?

grosser: also same as `if: aws_sts_iam_role_arn?`

plugins/aws_sts/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  before_save do
```
grosser: prefer named callbacks so they are easy to see in the backtrace and stubable in tests

plugins/aws_sts/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,36 @@
...
+    self.aws_sts_iam_role_session_duration ||= SamsonAwsSts::SESSION_DURATION_MIN
+  end
+
+  def can_assume_role
```
grosser: private

grosser: also `def validate_can_assume_role` since it's a validation

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
```diff
@@ -0,0 +1,45 @@
+<fieldset>
+  <legend>AWS STS Tokens</legend>
+  <div class="col-lg-offset-2">
+    <div class="form-group">
+      <div class="col-lg-6">
+        <p>Tokens will be requested using the following user: <%= SamsonAwsSts::Client.new.caller_user_id %></p>
```
grosser: good chance that this can fail ... needs a `rescue "error connecting to aws, check SAMSON_STS_AWS env vars"` ?

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
```diff
@@ -0,0 +1,45 @@
...
+        <span class="help-block">
+          Use the following env variables to change which user to authenticate with:
```
grosser: "in the project settings"

grosser: "(set secret key via secret:// and secret manager)"

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module SamsonAwsSts
+  SESSION_DURATION_MIN = 900  # 15 minutes
+  SESSION_DURATION_MAX = 3600 # 1 hour
```
grosser: would bump that since deploy could be long ... maybe 2 hours or max deploy length `Rails.application.config.samson.deploy_timeout`

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+end
+
+Samson::Hooks.view :stage_form, 'samson_aws_sts/fields'
+
+Samson::Hooks.callback :stage_permitted_params do
+  %i[
+    aws_sts_iam_role_arn
+    aws_sts_iam_role_session_duration
+  ]
```
grosser: I find `[:aws_sts_iam_role_arn, :aws_sts_iam_role_session_duration]` more readable 🤷‍♂️

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,76 @@
...
+end
+
+Samson::Hooks.callback :deploy_env do |deploy|
+  SamsonAwsSts::Client.new(sts_client: SamsonAwsSts.sts_client).deploy_env_vars(deploy: deploy)
```
grosser: same here, will break every deploy when not configured correctly ?

grosser: also should only do any work if the arn is set on the stage so users can opt in/out easily

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,76 @@
...
+
+module SamsonAwsSts
+  SESSION_DURATION_MIN = 900  # 15 minutes
+  SESSION_DURATION_MAX = 3600 # 1 hour
+
+  class Engine < Rails::Engine
+  end
+
+  def self.sts_client
+    Aws::STS::Client.new(
```
grosser: prefer to not do any work if the env vars are nil ... just return nil I guess

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
gerardc: done

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
gerardc: these are actually samson env vars, I'll update to make it clearer

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
gerardc: Added a guard clause with `next`, return was causing a localjumperror from the `fire` method in the tests

gerardc:
@grosser ready for re-review

gerardc:
ping @grosser @jonmoter ready for review

plugins/aws_sts/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+Stage.class_eval do
+  validates(
+    :aws_sts_iam_role_arn,
+    allow_blank: true,
+    format: {
+      with: /\Aarn:aws:iam::\d+:role\/.+\Z/,
+      message: "does not have the correct syntax. Must be of the form: arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME"
```
grosser: `does not have the correct syntax.` is kinda redundant :)

plugins/aws_sts/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,40 @@
...
+    :aws_sts_iam_role_session_duration,
+    allow_blank: true,
+    numericality: {
+      only_integer: true,
+      greater_than_or_equal_to: SamsonAwsSts::SESSION_DURATION_MIN,
+      less_than_or_equal_to:    SamsonAwsSts::SESSION_DURATION_MAX
+    }
+  )
+
+  validate :validate_can_assume_role, if: -> { aws_sts_iam_role_arn.present? }
```
grosser:
```suggestion
  validate :validate_can_assume_role, if: :aws_sts_iam_role_arn?
```

plugins/aws_sts/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  before_save :set_default_session_duration
```
grosser:
```suggestion
  before_validation :set_default_session_duration
```

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      {
+        STS_AWS_ACCESS_KEY_ID:     "hidden://#{creds.access_key_id}",
+        STS_AWS_SECRET_ACCESS_KEY: "hidden://#{creds.secret_access_key}",
+        STS_AWS_SESSION_TOKEN:     "hidden://#{creds.session_token}"
+      }
+    end
+
+    def assume_role(role_arn:, role_session_name:, duration_seconds: SESSION_DURATION_MIN)
+      sts_client.assume_role(
```
grosser: inline this method, does not do much ...

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -0,0 +1,80 @@
...
+    }
+    return if options.any? { |_, v| v.blank? }
+
+    options[:stub_responses] = Rails.env.test?
+
+    Aws::STS::Client.new(options)
+  end
+
+  class Client
+    def initialize(sts_client: SamsonAwsSts.sts_client)
```
grosser: don't need the default value here or the keyword-arg

grosser: ... or: either always pass the client or never pass the client

plugins/aws_sts/app/views/samson_aws_sts/_fields.html.erb
```diff
@@ -0,0 +1,45 @@
+<fieldset>
+  <legend>AWS STS Tokens</legend>
+  <div class="col-lg-offset-2">
+    <div class="form-group">
+      <div class="col-lg-6">
+        <p>Tokens will be requested using the following user: <%= SamsonAwsSts::Client.new.caller_user_id rescue "error connecting to aws, check SAMSON_STS_AWS env vars" %></p>
```
grosser:
```suggestion
        <p>Tokens will be requested using the following user: <%= SamsonAwsSts::Client.new.caller_user_id rescue "error connecting to aws, check SAMSON_STS_AWS_* env vars" %></p>
```

grosser: might be nice to not cause an exception+rescue+error here when aws_sts is not configured

```
if client
  blah blah rescue error
else
  samson env vars are not configured
end
```

gerardc:
@zendesk/bre this PR is ready to be merged 🙏(build is red due to pre-existing flakey test, I don't have permissions to retry)

When can we expect this behaviour to be deployed internally?

grosser:
- rebased and added the db/schema.rb change
 - fixing more issues atm like caching / unclear messages / always being expanded etc ... so do not merge yet

grosser:
ready to merge if you like my additions :)

gerardc:
@zendesk/bre can we please merge this PR?

gerardc:
@zendesk/bre when can we expect this behaviour to be deployed internally? Edge is blocked on this.

grosser:
I though you could merge once it's approved by any code-owner

gerardc:
Thanks!

PR #2975, opened 2018-10-04
Convert more logs to json format and added manditory fields to log

sathishavm:
Convert logs to JSON format and added some mandatory fields to log.

/cc @zendesk/samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-979

### Risks
- Level: Med

### DIFF
--- a/lib/samson/logging.rb
+++ b/lib/samson/logging.rb
@@ -19,14 +19,14 @@ elsif Samson::EnvCheck.set?("RAILS_LOG_TO_SYSLOG")
     {
       params: params,
       user_id: request.env['warden']&.user&.id,
-      ip: request.remote_ip
+      client_ip: request.remote_ip
     }
   end
 
   config.logger = Syslog::Logger.new('samson')
   config.action_cable.logger = Syslog::Logger.new('samson')
-  config.action_cable.logger.formatter = Samson::SyslogFormatter.new
   config.lograge.formatter = Lograge::Formatters::Logstash.new
+  config.logger.formatter = config.action_cable.logger.formatter = Samson::SyslogFormatter.new
 
 elsif ENV["SERVER_MODE"] # regular file logger that needs rotating
   # Reopen logfile when we receive a SIGHUP for log-rotation
--- a/lib/samson/syslog_formatters.rb
+++ b/lib/samson/syslog_formatters.rb
@@ -4,13 +4,14 @@ require 'syslog/logger'
 module Samson
   class SyslogFormatter < Syslog::Logger::Formatter
     def call(severity, timestamp, _progname, message)
-      message_h = begin
-        message = message.to_json unless message.is_a?(String)
-        # Fix for Integer and Array as a return object
-        (parsed = JSON.parse(message)).is_a?(Hash) ? parsed : raise(JSON::ParserError)
-      rescue JSON::ParserError
-        {"message": message&.squish}
-      end
+      message_h =
+        begin
+          message = message.to_json unless message.is_a?(String)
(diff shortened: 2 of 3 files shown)


lib/samson/syslog_formatters.rb
```diff
@@ -4,11 +4,20 @@
 module Samson
   class SyslogFormatter < Syslog::Logger::Formatter
     def call(severity, timestamp, _progname, message)
+      message_h =
+        begin
+          message = message.to_json unless message.is_a?(String)
+          # Fix for Integer and Array as a return object
```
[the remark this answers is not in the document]
sathishavm: Right, We allow strings / hashes, here we just parse the hash and merge into the default log hash.

lib/samson/syslog_formatters.rb
```diff
@@ -4,11 +4,21 @@
 module Samson
   class SyslogFormatter < Syslog::Logger::Formatter
     def call(severity, timestamp, _progname, message)
+      message_h =
+        begin
+          message = message.to_json unless message.is_a?(String)
```
grosser: so message comes in as a string or hash ? ... blow up when it's anything else ?
.... does it ever come in as string with json ? / can we avoid that ?

sathishavm: Not all times, we may also get a different object as input. 
I wish not to blow up because this formatter is now used by external gems, we just try to prettify the messages as much as possible, if not then we just squish them and embed in the message attributes of hash.

grosser: converting things to json and then parsing them right after is inefficient, how can we avoid that ?
who is passing json-ified objects into the logger / can we opt out of supporting that ?

grosser: why do we have to squish things ? .... that has the potential to modify important message details, so I'd rather avoid that if possible

sathishavm: ok, made alterations.

lib/samson/logging.rb
```diff
@@ -19,14 +19,14 @@
     {
       params: params,
       user_id: request.env['warden']&.user&.id,
-      ip: request.remote_ip
+      client_ip: request.remote_ip
```
grosser: why did this change, just readability or something else ?

lib/samson/syslog_formatters.rb
```diff
@@ -4,11 +4,25 @@
 module Samson
   class SyslogFormatter < Syslog::Logger::Formatter
     def call(severity, timestamp, _progname, message)
+      message_h =
+        begin
```
grosser: raising on every non-hash/json passed in is wasteful ... also code is kinda convoluted
better to dumb it down to this which not as dry but easier to read I hope:

```Ruby
if message.is_a?(Hash)
  message
elsif message.is_a?(String) && message.start_with?('{')
  begin
    JSON.parse(message)
  rescue JSON::ParserError
    {message: message}
  end
else
  {message: message}
end

```

lib/samson/logging.rb
dragonfax: I would guess readability. I mean, I wouldn't know looking at the logs if it was the server's IP (so you can track down which server is bad) or the clients IP (who causing the problem).

lib/samson/syslog_formatters.rb
grosser: old: raise on any non-json string + on any non-string-and-non-hash
new: raise when a string looks like json but is not parseable

sathishavm: fine, altered!!

PR #2978, opened 2018-10-05
Bump pg

ragurney:
`pg` gem was pretty outdated, and was giving a seg fault when attempting to migrate on rails 2.5.1. 
screen shot 2018-10-05 at 8 54 32 am
Curiously, 2.5.0 does not have the same issue. Bumping pg seems to solve the issue. 

/cc @zendesk/samson


grosser:
an general `bundle update` might also be nice :)

PR #2982, opened 2018-10-06
notify compute when kubernetes is changed

grosser:


grosser:
@zendesk/compute

adammw:
should it be bre and compute?

grosser:
I think it works since bre has * ... we'll find out soon :)

PR #2983, opened 2018-10-07
do not check github during our requests

grosser:
currently we block each request if caching it somehow did not work,
this moves it to our periodical background thread so we do not block our requests

also prepares for multiple repo providers in the future

screen shot 2018-10-06 at 8 15 39 pm
screen shot 2018-10-06 at 8 12 24 pm

@zendesk/samson 
/fyi @pingfr one thing that's no longer hardcoded to github :)

# Risks:
 - low: annoying warning when not adding `repo_provider_status` to PERIODICAL

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -86,8 +86,9 @@ GITHUB_TOKEN=
 # report_process_stats:60
 # periodical_deploy:86400
 # cancel_stalled_builds:3600
+# repo_provider_status:60 (for plugins that use repo_provider_status hook, like samson_github)
 # Recommended setup:
-PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:60,report_process_stats:60,periodical_deploy:86400,cancel_stalled_builds:3600
+PERIODICAL=stop_expired_deploys:60,remove_expired_locks:60,report_system_stats:60,report_process_stats:60,periodical_deploy:86400,cancel_stalled_builds:3600,repo_provider_status:60
 
 ## Buddy Check feature: deploys to production require a buddy
 # BUDDY_CHECK_FEATURE=1 # enable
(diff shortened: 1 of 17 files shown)


.env.example
```diff
@@ -66,8 +65,9 @@ BADGE_TOKEN_BASE={bundle exec rake secret} # optional
 # report_process_stats:60
 # periodical_deploy:86400
 # cancel_stalled_builds:3600
+# repo_provider_status:60
```
ragurney: Add a note that this depends on the `samson_github` plugin (or a plugin in general with the correct hook implemented) being enabled?

plugins/github/test/samson_github/samson_plugin_test.rb
```diff
@@ -58,4 +58,38 @@
...
+      Samson::Hooks.fire(:repo_provider_status).first
+    end
+
+    let(:status_url) { "#{SamsonGithub::STATUS_URL}/api/status.json" }
+
+    around { |t| Samson::Hooks.only_callbacks_for_plugin('github', :repo_provider_status, &t) }
+
+    it "reports good response" do
+      assert_request(:get, status_url, to_return: {body: {status: 'good'}.to_json}) do
+        refute fire
```
ragurney: Might be good to be a little more specific, and assert on the actual object we expect to get back, such as `nil` or the expected array of errors?

lib/samson/repo_provider_status.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+module Samson
+  class RepoProviderStatus
+    CACHE_KEY = name
+
+    class << self
+      def errors
+        (
+          Rails.cache.read(CACHE_KEY) ||
+          ["Add repo_provider_status:60 to PERIODICAL environment variable."]
```
ragurney: Could we add something explaining why users should add this to the environment variable?

PR #2986, opened 2018-10-08
show deploy_url in kubernetes resources since it is more useful than …

eatwithforks:
…owner info/repo

### DIFF
--- a/lib/samson/extra_dig.rb
+++ b/lib/samson/extra_dig.rb
@@ -11,7 +11,7 @@ module Samson
       raise ArgumentError, "No key given" if keys.empty?
       keys = keys.dup
       last = keys.pop
-      failed = ->(*) { raise KeyError, "key not found: #{(keys << last).inspect}" }
+      failed = ->(*) { raise KeyError, "key not found: #{keys.inspect}" }
       nested = keys.inject(self) { |h, k| h.fetch(k, &failed) }
       nested[last] = value
     end
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -20,6 +20,7 @@ module Kubernetes
 
         set_namespace
         set_project_labels if template.dig(:metadata, :annotations, :"samson/override_project_label")
+        set_deploy_url
 
         case kind
         when 'HorizontalPodAutoscaler'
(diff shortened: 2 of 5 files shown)


lib/samson/extra_dig.rb
```diff
@@ -11,7 +11,7 @@ def dig_set(keys, value)
       raise ArgumentError, "No key given" if keys.empty?
       keys = keys.dup
       last = keys.pop
-      failed = ->(*) { raise KeyError, "key not found: #{(keys << last).inspect}" }
+      failed = ->(*) { raise KeyError, "key not found: #{keys.inspect}" }
```
adammw: seems odd the error message for dig_fetch didn't change...

grosser:
dig_set was wrong, since when setting a,b,c it should say "a,b" did not
exist and not a,b,c since c is never the problem since it is what is being
set

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -85,6 +84,15 @@ def build_selector_for_container(container)
       end
     end
 
+    def set_deploy_url
+      templates = [template]
+      templates << pod_template if Kubernetes::RoleConfigFile::PRIMARY_KINDS.include?(template[:kind])
```
adammw: what about deploy kinds?

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -825,9 +797,9 @@ def secret_annotations(hash)
     end
 
     describe "when something is required" do
-      let(:annotations) { raw_template[:spec][:template][:metadata][:annotations] = {} }
+      let(:pod_annotations) { raw_template[:spec][:template][:metadata][:annotations] = {} }
```
adammw: probably didn't need to rename these, but it doesn't hurt i guess

grosser:
included in primary

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -161,49 +168,14 @@ def add_init_container(container)
...
       it "only modifies namespace since there is no template" do
         raw_template[:kind] = "ConfigMap"
         raw_template.delete(:spec)
         raw_template[:metadata][:namespace] = 'old'
-        old = raw_template.deep_dup
-        old[:metadata][:namespace] = 'pod1'
-        template.to_hash.must_equal old
+        result = template.to_hash
+        result[:metadata][:namespace].must_equal 'pod1'
+        refute result[:spec]
```
adammw: i don't really understand this change, is it related?

grosser:
just for consistency to code

grosser:
previously we made no changes to ConfigMap, this now changes since we
always set this annotation

PR #2987, opened 2018-10-08
Fix APIService no method error

eatwithforks:


eatwithforks:
https://github.com/abonas/kubeclient/pull/355

PR #2988, opened 2018-10-09
allow falling back to full checkout for stages that do weird stuff

grosser:
followup to https://github.com/zendesk/samson/pull/2973 which broke stages that
switch branches (they should not do that ever ...) or try to create tags

@zendesk/samson @zendesk/bre @craig-day @adam-hart 

screen shot 2018-10-08 at 7 33 34 pm


sathishavm:
wish to notify this change in engineering channel, so that we don't expect surprises.

PR #2990, opened 2018-10-09
allow opting out of dockerfile requirement for additional containers …

### LINKED ISSUE #2985: support builds for multiple dockerfiles forces samson/dockerfile usage
Hi!

I'm trying to upgrade samson, but I think commit a236012267657236067b3372909fc7bbc3629a4e forces to use `samson/dockerfile` on the `containers[]`.

Before the upgrade, samson used to only update the image sha on the first container. This worked fine for me because the apps I use contain only one container that needs to be built and then several side-car containers (typically an nginx and a log collector).

When trying to use the new version (for https://github.com/zendesk/samson/pull/2970, thanks again @grosser !), a deploy fails. The issue is that samson updates the images in all containers (the nginx and log collector too). And this fails with address already in used, as all bind to localhost. But of course, the issue is that sidecars containers are not used (images are specified for them).

The way I found to prevent this is setting `samson/dockerfile: none` on the sidecars containers. But for me use case, it has several problems:

- In a way, it's not backwards compatible as yaml that were working are not working any more
- So, it needs to be done on all applications for sidecars containers (this is not trivial, though)
- The syntax is not valid in k8s, so if you try to use kubectl or need some kind of compatibility, you need to add a hack. As trying to deploy with kubectl, for example, says:

```
error: error validating "kubernetes/wido-web.yml": error validating data: [ValidationError(Deployment.spec.template.spec.containers[1]): unknown field "samson/dockerfile" in io.k8s.api.core.v1.Container, ValidationError(Deployment.spec.template.spec.containers[2]): unknown field "samson/dockerfile" in io.k8s.api.core.v1.Container]; if you choose to ignore these errors, turn validation off with --validate=false
```

I think that maintaining the syntax compatible with k8s is importante, as if not the kubernetes dashboard or some other tools complementing this can't be easily used without patches.

What do you think of using a `metadata.annotation` (that are intended to be used by software) to mark either multi dockerfile builds enabled or disabled (with some default) and maybe specify the dockerfile for each container there, when enabled?

Like `metadata.annotations.zendesk/samson.multi-dockerfiles-enabled: true` to mark this behavior. Or having to mark it disabled to return to previous behavior?

That would keep the yaml syntax valid for deployments that do not use multi docker files, at least.

But, if you want also, something like:

```
metadata.annotations.zendesk/samson.dockerfile[container-name]: Dockerfile.new
```

To keep the syntax valid and specify which containers use which docker files for samson builds.

The main issue for me is to change all projects (all use sidecars :-/) and have the hard requirement of doing it in a way that breaks the kubernetes syntax, so it won't be possible to interoperate with other tools easily.

What do you think?

grosser:
…to ease migration

fixes https://github.com/zendesk/samson/issues/2985

@rata does this work ?

if yes I'll add some tests and merge it ...

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -77,8 +77,8 @@ module Kubernetes
     private
 
     def build_selector_for_container(container, first:)
-      dockerfile = container[:"samson/dockerfile"] ||
-        (first && ENV['KUBERNETES_ADDITIONAL_CONTAINERS_WITHOUT_DOCKERFILE'] ? 'none' : 'Dockerfile')
+      dockerfile = samson_container_config(container, :"samson/dockerfile") ||
+        (!first && ENV['KUBERNETES_ADDITIONAL_CONTAINERS_WITHOUT_DOCKERFILE'] ? 'none' : 'Dockerfile')
 
       return if dockerfile == 'none'
 
@@ -90,6 +90,12 @@ module Kubernetes
       else
         [dockerfile, nil]
       end
+    end
+
+    # samson/ keys in containers trigger validation warnings in kubectl, so we allow using annotations too
+    # NOTE: containers always have a name see role_validator.rb
+    def samson_container_config(container, key)
+      pod_annotations[:"container-#{container[:name]}-#{key}"] || container[key]
     end
 
     def set_deploy_url
@@ -495,7 +501,7 @@ module Kubernetes
 
     def set_pre_stop
       containers.each do |container|
-        next if container[:"samson/preStop"] == "disabled"
+        next if samson_container_config(container, :"samson/preStop") == "disabled"
         (container[:lifecycle] ||= {})[:preStop] ||= {exec: {command: ["sleep", "3"]}}
       end
     end
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -71,13 +71,15 @@ module Kubernetes
 
     def build_selectors
       all = containers + init_containers
-      all.map { |c| build_selector_for_container(c) }.compact
+      all.each_with_index.map { |c, i| build_selector_for_container(c, first: i == 0) }.compact
     end
 
     private
 
-    def build_selector_for_container(container)
-      dockerfile = container[:"samson/dockerfile"] || 'Dockerfile'
+    def build_selector_for_container(container, first:)
+      dockerfile = container[:"samson/dockerfile"] ||
+        (first && ENV['KUBERNETES_ADDITIONAL_CONTAINERS_WITHOUT_DOCKERFILE'] ? 'none' : 'Dockerfile')
+
       return if dockerfile == 'none'
 
       if project.docker_image_building_disabled?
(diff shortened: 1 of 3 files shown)


grosser:
also added support for getting container config from annotations via `container_foo_samson/dockerfile`

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -71,14 +71,14 @@ def verify
...
     end
 
     private
 
-    def build_selector_for_container(container, first:)
-      dockerfile = container[:"samson/dockerfile"] ||
-        (first && ENV['KUBERNETES_ADDITIONAL_CONTAINERS_WITHOUT_DOCKERFILE'] ? 'none' : 'Dockerfile')
+    def build_selector_for_container(container, index:)
+      dockerfile = samson_container_config(container, :"samson/dockerfile") ||
+        (index == 0 && ENV['KUBERNETES_ADDITIONAL_CONTAINERS_WITHOUT_DOCKERFILE'] ? 'none' : 'Dockerfile')
```
klaus993: I analysed and tested this code and came to the conclusion that this code should be `index != 0` instead.

We want the `dockerfile` variable to be `'none'` only with the sidecar containers, don't we? So they don't get their image tag replaced. It's the first one the only one that has to have the hash inserted. This way it happens the other way round.

klaus993 (changes requested):
Thanks, @grosser ! It works fine with the change I propose in my comment on the code, I suppose it must've been a simple error.

grosser:
ah yes :D

... I'll whip up some tests and docs then and get some more feedback!

rata:
@grosser great! Doing the change for `index != 0` works for me too.

I've tried not using the env variable and just using the new annotations, and it didn't work. I see the annotations in the pod, like:

```
kind: Pod
metadata:
  annotations:
    container-heka-samson/dokerfile: none
    container-nginx-samson/dokerfile: none
```

But not sure why that seems not to be working. I've been trying to debug samson for a while now. Will continue and report bug, it sure is something silly :D

rata:
It's not a big problem, as I plan to use the variable first. But, to check the patch as a whole, I tried to check annotations too :)

grosser:
ah I did the annotation with underscores, but dashes are probably better
anyway

grosser:
all fixed up and documented, looks good ?

grosser:
@zendesk/compute 👀

klaus993 (approved):
🚀 Works like a charm, although we will use the env var for simplicity, the other option works as well 😄 

It seems @rata was testing it with a typo in the annotation, he missed the `c` in `dockerfile`, I spotted that, changed it, tested, and only the first container is replaced.

Thanks!

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -71,13 +71,15 @@ def verify
...
       all = containers + init_containers
-      all.map { |c| build_selector_for_container(c) }.compact
+      all.each_with_index.map { |c, i| build_selector_for_container(c, index: i) }.compact
     end
 
     private
 
-    def build_selector_for_container(container)
-      dockerfile = container[:"samson/dockerfile"] || 'Dockerfile'
+    def build_selector_for_container(container, index:)
```
ragurney: `index` is a little confusing as an argument at first glance. What do you thing about passing a boolean `first` or `first_container` instead?

ragurney (approved):

grosser:
yeah, good catch

PR #2992, opened 2018-10-10
use periodical for heroku and for bootstrapped projects

grosser:
@zendesk/bre


grosser:
just forgot and nobody ever complained :)

PR #2994, opened 2018-10-11
Support setting a default user role for new users

### LINKED ISSUE #1184: Enable setting default new user role
We're using github authentication and authorise by restricting our users to a particular group in our organisation. When they first log in, users are assigned the "Viewer" role, and since all our users are trusted (because we've put them in the right group in github), we want them all to have the role "Super admin". Currently someone has elevate permissions for folks manually.

It'd be great to be able to set the default user role that applies to all new users, so we could skip the manual steps.

ragurney:
Resolves #1184 

Adds support for setting a default user role for new users via `DEFAULT_USER_ROLE`. I also refactored some of the logic around creating/updating a `User` on login. The existing logic would have silently upgraded a downgraded user's role. For example if `DEFAULT_USER_ROLE=2` and I decided to downgrade a user to `0`, it would have set the user role back to 2 on re-login. It seems kind of funky to be silently changing an existing user's attributes on login. Once created, it makes sense to me to have authorization management be manual, and centralized in one place (in this case the user dashboard) to avoid unexpected behavior. 

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


app/controllers/sessions_controller.rb
```diff
@@ -11,27 +11,27 @@ def new
...
   def ldap
     return show_login_restriction unless allowed_to_login
-    login(role_id: Role::VIEWER.id)
+    login(role_id: custom_role_or_default(Role::VIEWER.id))
   end
 
   def gitlab
     return show_login_restriction unless allowed_to_login
-    login(role_id: Role::VIEWER.id)
+    login(role_id: custom_role_or_default(Role::VIEWER.id))
```
grosser: should login take care of custom_role_or_default ?

ragurney: I was on the fence about that... I'll make the change 👍

ragurney:
I was on the fence about that too 😄 though I figured the changes are related enough, given the scenario I described above. I don't mind splitting it out though so we can merge the refactor forst

grosser:
just merge, more for future PRs, this was still readable :)

app/controllers/sessions_controller.rb
ragurney: Actually, now I think about it I think it makes sense to leave this logic in each action, in case it ever needs to change for a specific authentication method?

app/controllers/sessions_controller.rb
```diff
@@ -114,4 +114,15 @@ def login(options = {})
 
     redirect_to_origin_or_default
   end
+
+  def find_or_create_user_from_hash(hash)
+    # first user will be promoted to super admin
+    hash[:role_id] = Role::SUPER_ADMIN.id unless User.where.not(email: 'seed@example.com').exists?
```
sathishavm: Can you use User.count??.. for any reason if I delete the 'seed@example.com', then all the user's created after that will be Super admin.

ragurney: This just checks if there are any users _not_ named `seed@example.com`. Deleting a user with that name wouldn't result in all users being given super admin privileges.

sathishavm: oh!! yes. Missed the `not`.

PR #2996, opened 2018-10-12
K8s prereq timeout

rata:
Having a different timeout for prerequisites is useful for us, as some jobs we run take more than what we want to wait for a deploy.

For backwards compatibility, the default value is the same as KUBE_WAIT_FOR_LIVE (so no workflow is affected unless explicitly changed).

@grosser What do you think? This is useful for us, and should not change anything for people not using it. And the code is not very complicated.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -6,6 +6,7 @@ require 'vault'
 module Kubernetes
   class DeployExecutor
     WAIT_FOR_LIVE = ENV.fetch('KUBE_WAIT_FOR_LIVE', 10).to_i.minutes
+    WAIT_FOR_PREREQ = ENV.fetch('KUBE_WAIT_FOR_PREREQ', 10).to_i.minutes
     STABILITY_CHECK_DURATION = 1.minute
     TICK = 2.seconds
     RESTARTED = "Restarted"
@@ -59,11 +60,11 @@ module Kubernetes
       prerequisites, deploys = @release.release_docs.partition(&:prerequisite?)
       if prerequisites.any?
         @output.puts "First deploying prerequisite ..." if deploys.any?
-        return false unless deploy_and_watch(prerequisites)
+        return false unless deploy_and_watch(prerequisites, WAIT_FOR_PREREQ)
         @output.puts "Now deploying other roles ..." if deploys.any?
       end
       if deploys.any?
-        return false unless deploy_and_watch(deploys)
+        return false unless deploy_and_watch(deploys, WAIT_FOR_LIVE)
       end
       true
     end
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -72,7 +73,7 @@ def execute(*)
 
     # check all pods and see if they are running
     # once they are running check if they are stable (for apps only, since jobs are finished and will not change)
-    def wait_for_resources_to_complete(release_docs)
+    def wait_for_resources_to_complete(release_docs, timeout = WAIT_FOR_LIVE)
```
rata: Maybe the default value can be removed here, just noticed 😱

grosser: prefer keyword args and no default value

grosser:
looks good
would a project setting be better so it is not globally configured ?

grosser:
I guess global default is enough for now ... project setting would mean a
new column / form-field and can be an additional PR if needed

adammw (approved):

rata:
@grosser cool, thanks!
@adammw thanks!

Removed the default arg and fixed tests. Now all test pass, except `TASK=fossa`. Is it unrelated to this PR? The log didn't point anything related to this, AFAICS.

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -6,6 +6,7 @@
 module Kubernetes
   class DeployExecutor
     WAIT_FOR_LIVE = ENV.fetch('KUBE_WAIT_FOR_LIVE', 10).to_i.minutes
+    WAIT_FOR_PREREQ = ENV.fetch('KUBE_WAIT_FOR_PREREQ', 10).to_i.minutes
```
grosser: should fallback to WAIT_FOR_LIVE when not set ?
`WAIT_FOR_PREREQ = ENV.fetch('KUBE_WAIT_FOR_PREREQ', WAIT_FOR_LIVE / 1.minute).to_i.minutes`

rata: Yeah, makes sense. The code gets a little tricker with the division and conversion, IMHO. Wonder if can be made simpler...

Would this work:
```
WAIT_FOR_LIVE = ENV.fetch('KUBE_WAIT_FOR_LIVE', 10)
WAIT_FOR_PREREQ = ENV['KUBE_WAIT_FOR_PREREQ'] ||  WAIT_FOR_LIVE
WAIT_FOR_LIVE = WAIT_FOR_LIVE.to_i.minutes
WAIT_FOR_PREREQ = WAIT_FOR_PREREQ.to_i.minutes
```

But let me know which one you prefer (or just modify it, if GitHub default lets you change my branch) or if you can think of something simpler :)


Thanks!

PR #3002, opened 2018-10-15
[BRE-1009] Support GitHub's new Checks API for reference status on deploy review

ragurney:
screen shot 2018-10-15 at 12 25 43 pm

We've switched over to using the new Travic CI integration for all Zendesk repositories, which uses GitHub's new Checks API. This broke Samson's reference status logic on the Deploy review page and elsewhere. This PR introduces ~~a new environment variable `USE_GITHUB_CHECKS_API` which will allow users to use the Checks API to get the reference status instead of the Status API.~~ logic to show results from both the Status and Checks APIs.

/cc @zendesk/samson

### Tasks
 - [x] :+1: from team
 - [ ] Test in Staging

### References
 - Jira link: [BRE-1009](https://zendesk.atlassian.net/browse/BRE-1009)

### Risks
 - Low

### DIFF
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -7,6 +7,22 @@ class CommitStatus
   # - missing is our own state that means we could not determine the status
   STATE_PRIORITY = [:success, :pending, :missing, :failure, :error, :fatal].freeze
   UNDETERMINED = ["pending", "missing"].freeze
+  CHECK_STATE = {
+    error: ['action_required', 'canceled', 'timed_out'],
+    failure: ['failed'],
+    success: ['success', 'neutral']
+  }.freeze
+  NO_STATUSES_REPORTED_RESULT = {
+    state: 'pending',
+    statuses: [
+      {
+        state: 'pending',
+        description:
+          "No status was reported for this commit on GitHub. See https://developer.github.com/v3/checks/ and " \
+            "https://github.com/blog/1227-commit-status-api for details."
+      }
+    ]
+  }.freeze
 
   def initialize(project, reference, stage: nil)
     @project = project
@@ -19,16 +35,7 @@ class CommitStatus
   end
 
   def statuses
-    list = combined_status.fetch(:statuses).map(&:to_h)
-    if list.empty?
-      list << {
-        state: 'pending',
-        description:
-          "No status was reported for this commit on GitHub. " \
-          "See https://github.com/blog/1227-commit-status-api for details."
-      }
-    end
-    list
+    combined_status.fetch(:statuses).map(&:to_h)
   end
 
   def expire_cache(commit)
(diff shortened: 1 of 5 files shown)


grosser:
it should use both by default since not all things on all repos are in the checks api ?

ragurney:
So just combine statuses from both locations and get rid of the env var? Or have an option to be able to choose one or the other, such as `REFERENCE_STATE_API=status/checks`, both if not set?

grosser:
use both just like the github UI does ... unless there is a reason not to
... less config for our users :)

ragurney:
@grosser ready for another look 👀

grosser:
unrelated: now that the deployments are more visible maybe we should revisit the use of deployment api :)
screen shot 2018-10-16 at 11 18 40 am

app/models/commit_status.rb
```diff
@@ -5,8 +5,13 @@ class CommitStatus
   # See https://developer.github.com/v3/repos/statuses for api details
   # - fatal is our own state that blocks deploys
   # - missing is our own state that means we could not determine the status
-  STATE_PRIORITY = [:success, :pending, :missing, :failure, :error, :fatal].freeze
+  STATE_PRIORITY = [:pending, :success, :missing, :failure, :error, :fatal].freeze
```
grosser: that looks wrong ... if something is pending then it cannot be marked as success

app/models/commit_status.rb
```diff
@@ -5,8 +5,13 @@ class CommitStatus
...
   UNDETERMINED = ["pending", "missing"].freeze
+  STATE_TO_CHECK_CONCLUSION_MAPPINGS = {
```
grosser: maybe just CHECK_STATE ?

app/models/commit_status.rb
```diff
@@ -39,33 +46,70 @@ def expire_cache(commit)
...
-  def github_status
+  def github_state
     static = @reference.match?(Build::SHA1_REGEX) || @reference.match?(Release::VERSION_REGEX)
     expires_in = ->(reply) { cache_duration(reply) }
     cache_fetch_if static, cache_key(@reference), expires_in: expires_in do
-      GITHUB.combined_status(@project.repository_path, @reference).to_h
+      checks_result = with_octokit_client_error_rescue { github_check }
+      status_result = with_octokit_client_error_rescue { github_status }
+
+      merge_statuses([checks_result, status_result])
```
grosser: don't merge here but let change combined_status to be `statuses  = github_state ` ?

app/models/commit_status.rb
```diff
@@ -39,33 +46,70 @@ def expire_cache(commit)
...
+  def github_check
+    base_url = "repos/#{@project.repository_path}/commits/#{@reference}"
+    preview_header = {Accept: 'application/vnd.github.antiope-preview+json'}
+
+    check_suites = GITHUB.get("#{base_url}/check-suites", headers: preview_header)[:check_suites]
+    checks = GITHUB.get("#{base_url}/check-runs", headers: preview_header)
+
+    overall_state = check_suites.
+      map { |suite| check_state_equivalent(suite[:conclusion]) }.
+      max_by { |state| STATE_PRIORITY.index(state.to_sym) }
```
grosser: could add a `|| raise("Unknown: #{state}")` for faster debugging when a new state is added

ragurney:
@grosser would appreciate another look 🙏 took me a while to wrap my head around the behavior of the two different APIs and the right way to handle missing statuses. Think I got it this time.

app/models/commit_status.rb
```diff
@@ -39,38 +46,77 @@ def expire_cache(commit)
...
   end
 
-  def cache_duration(github_status)
-    statuses = github_status[:statuses]
-    if statuses.empty? # does not have any statuses, chances are commit is new
+  # Gets commit statuses using GitHub's check API. Currently parsing it to match status structure to better facilitate
+  # transition to new API. See https://developer.github.com/v3/checks/runs/ and
+  # https://developer.github.com/v3/checks/suites/ for details
+  def github_check
+    base_url = "repos/#{@project.repository_path}/commits/#{@reference}"
```
grosser: I think reference needs to be `CGI.escape`-ed or branches with / will break it

app/models/commit_status.rb
grosser: ^ this ?

app/models/commit_status.rb
```diff
@@ -39,38 +46,77 @@ def expire_cache(commit)
...
+
+    check_suites = GITHUB.get("#{base_url}/check-suites", headers: preview_header)[:check_suites]
+    checks = GITHUB.get("#{base_url}/check-runs", headers: preview_header)
+
+    overall_state = check_suites.
+      map { |suite| check_state_equivalent(suite[:conclusion]) }.
+      max_by { |state| STATE_PRIORITY.index(state.to_sym) }
+
+    statuses = checks[:check_runs].each_with_object([]) do |check_run, result|
+      result << {
```
grosser: .map instead if each + << ?

test/models/commit_status_test.rb
```diff
@@ -17,156 +17,302 @@ def self.deploying_a_previous_release
...
   end
 
-  def status(stage_param: stage, reference_param: reference)
-    CommitStatus.new(stage_param.project, reference_param, stage: stage_param)
+  def status(stage_param: stage, reference_param: reference, new: false)
+    if new
+      @status = CommitStatus.new(stage_param.project, reference_param, stage: stage_param)
+    else
+      @status ||= CommitStatus.new(stage_param.project, reference_param, stage: stage_param)
+    end
```
grosser: methods with arguments should not cache without the arguments ... change the tests to have local variables or do a `let(:status)` that calls `build_status` with default values or so

app/models/commit_status.rb
```diff
@@ -39,38 +46,77 @@ def expire_cache(commit)
...
-  def github_status
+  def github_state
     static = @reference.match?(Build::SHA1_REGEX) || @reference.match?(Release::VERSION_REGEX)
     expires_in = ->(reply) { cache_duration(reply) }
     cache_fetch_if static, cache_key(@reference), expires_in: expires_in do
-      GITHUB.combined_status(@project.repository_path, @reference).to_h
+      checks_result = with_octokit_client_error_rescue { github_check }
+      status_result = with_octokit_client_error_rescue { github_status }
+
+      results_with_statuses = [checks_result, status_result].select { |result| result[:statuses].any? }
```
grosser: could do the merge first and then check empty after ... but either way is fine 🤷‍♂️

app/models/commit_status.rb
ragurney: Right now we default to 'pending' if there are no `check_suites` to be consistent with the States API..

grosser: I meant for inside the max_by :D

ragurney: Ah gotcha. Currently returning 'pending' if the `conclusion` is `nil` so I don't have to check both 'status' and 'conclusion'. Added an explicit check for `nil` in `check_state_equivalent` for pending, and a raise if an unexpected state is found.

Issue #3003, opened 2018-10-16
Configuration knob to see output of prerequisite containers


rata:
Hi!

After https://github.com/zendesk/samson/pull/2996 I realize that seeing the output of the containers is useful for us.

The thing is: if we use the prerequisite jobs just to run a 1-off job (like a data migration or something), then if all goes well, there is no way to know which "deploy" run which script, what was it's output, etc. The only case were it is shown is if the 1-off job fails.

I tried to create a configuration knob (env var) to enable this, but I'm missing something that should be really simple to iterate the correct pods and get it's output. The code is pretty simple (just add a `show_logs_on_success` param to `deploy_and_watch`) and inspired the functions to get the logs from `show_failure_cause`.

Do you think it's okay to have a config know like that?

Another alternative I saw was something like this: https://github.com/zendesk/samson/pull/2665/ that references this: https://github.com/zendesk/samson/issues/2608

In any case, seeing the output for a 1-off job seems better to me (is more easy to know which script run, which things it processed, etc.). The other issue seems, honestly, like another route cause and different solutions seem better.

If this approach is fine (env var to always show logs after prerequisite runs), I'll continue to play with this tomorrow :)

Thanks!

grosser:
I like the idea of showing logs for 1-off pods, super useful for migrations
:)
... metadata.annotations.samson/show_logs_on_success or so ... ideally it
would stream the logs while it is executing, but that might be tricky

PR #3005, opened 2018-10-17
Add annotation to show logs of prerequisite pods

rata:
* Adds annotation to show logs of executed prerequisite pods

Fixes: #3003 

@grosser does this make sense to you?

The only disadvantage I see is that the annotation is on the pod, not the job, so you need one annotation for prerequisite at the job kind and another at the pod spec inside the job.

I mean, something like:

```
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    samson/prerequisite: "true"
  name: data-migration
  labels:
    project: project
    role: data-migration
spec:
  template:
    metadata:
      labels:
        project: project
        role: data-migration
      annotations:
        samson/show_logs_on_success: "true"
    spec:
      restartPolicy: Never
      ...
```

I tried first using an env var, when I had that working I tried the annotation on the pod. But then I realized it's on the pod, not the job or something else, so tried to do it on the job. But not sure that part of the code can check easily for that. I think it's kind of tricky, as it needs to check for prerequisites (not sure that part of the code knows about them), but then check for annotations there and print for all pods that belong to that resource (following the owner reference, hopefully, but not sure that works fine in Kubernetes. On not sure on which versions)

Using just an annotation for the pod seems simpler, now that I think about it. But will try to look at adding the annotation at the job level.  

What do you think? Let me know if at the pod level it's fine (seems simpler, but not sure :))

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -18,6 +18,10 @@ module Kubernetes
         @pod.dig(:metadata, :namespace)
       end
 
+      def annotations
+        @pod[:metadata][:annotations] ||= {}
+      end
+
       def live?
         completed? || (phase == 'Running' && ready?)
       end
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -63,11 +63,11 @@ module Kubernetes
       prerequisites, deploys = @release.release_docs.partition(&:prerequisite?)
       if prerequisites.any?
         @output.puts "First deploying prerequisite ..." if deploys.any?
-        return false unless deploy_and_watch(prerequisites, WAIT_FOR_PREREQUISITES)
+        return false unless deploy_and_watch(prerequisites, WAIT_FOR_PREREQUISITES, show_logs_if_requested: true)
         @output.puts "Now deploying other roles ..." if deploys.any?
       end
       if deploys.any?
-        return false unless deploy_and_watch(deploys, WAIT_FOR_LIVE)
+        return false unless deploy_and_watch(deploys, WAIT_FOR_LIVE, show_logs_if_requested: false)
       end
       true
     end
@@ -140,6 +140,20 @@ module Kubernetes
         pods = SamsonKubernetes.retry_on_connection_errors { client.get_pods(query).fetch(:items) }
         pods.map! { |p| Kubernetes::Api::Pod.new(p, client: client) }
       end
+    end
+
+    def show_pods_logs_if_requested
+      log_end_time = 0.seconds.from_now
+      pods = fetch_pods
+      pods.each do |pod|
+        if pod.annotations[:'samson/show_logs_on_deploy'] == 'true'
+          print_pod_logs(pod, log_end_time)
+          @output.puts "\n------------------------------------------\n"
+        end
+      end
+    rescue StandardError
+      info = ErrorNotifier.notify($!, sync: true)
+      @output.puts "Error showing logs: #{info}"
     end
 
     def show_failure_cause(release_docs, statuses)
@@ -392,12 +406,15 @@ module Kubernetes
       Samson::Parallelizer.map(resources, db: true, &:deploy)
     end
 
-    def deploy_and_watch(release_docs, timeout)
+    def deploy_and_watch(release_docs, timeout, show_logs_if_requested:)
       deploy(release_docs)
       result = wait_for_resources_to_complete(release_docs, timeout)
       if result == true
         if blue_green = release_docs.select(&:blue_green_color).presence
           finish_blue_green_deployment(blue_green)
+        end
+        if show_logs_if_requested
+          show_pods_logs_if_requested
         end
         true
       else
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -144,8 +144,11 @@ module Kubernetes
 
     def show_pods_logs_if_requested
       log_end_time = 0.seconds.from_now
+      @output.puts "\nXXX\n"
       pods = fetch_pods
       pods.each do |pod|
+        @output.puts "\nXXX: pod iter\n"
+        @output.puts "\n XXX: #{pod.annotations}\n"
         if pod.annotations[:'samson/show_logs_on_deploy'] == 'true'
           print_pod_logs(pod, log_end_time)
           @output.puts "\n------------------------------------------\n"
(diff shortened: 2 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -392,13 +406,16 @@ def deploy(release_docs)
       Samson::Parallelizer.map(resources, db: true, &:deploy)
     end
 
-    def deploy_and_watch(release_docs, timeout)
+    def deploy_and_watch(release_docs, timeout, show_logs_on_success)
```
grosser: show_logs_on_success is not a logical 3rd argument, so should use keyword args

`, show_logs_on_success:` and then call with `show_logs_on_success: true/false`

plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -18,6 +18,10 @@ def namespace
         @pod.dig(:metadata, :namespace)
       end
 
+      def annotations
+        @pod.dig(:metadata, :annotations)
```
grosser: needs to be `||= {}` or it will blow up when there are no annotations
```suggestion
        @pod[:metadata][:annotations] ||= {}
```

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
grosser: also the flag does not make it show logs, just 'maybe' show logs, so maybe better to use `prerequisite: true`

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -392,13 +406,16 @@ def deploy(release_docs)
...
       deploy(release_docs)
       result = wait_for_resources_to_complete(release_docs, timeout)
       if result == true
         if blue_green = release_docs.select(&:blue_green_color).presence
           finish_blue_green_deployment(blue_green)
         end
+        if show_logs_on_success
+          show_pods_logs
```
grosser: maybe even `show_pods_logs_if_requested`

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
grosser: ... we can also show logs for pods that are not prerequisites if the user asked for them ... so don't need this flag at all

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -142,6 +142,20 @@ def fetch_pods
       end
     end
 
+    def show_pods_logs
+      log_end_time = 0.seconds.from_now
+      pods = fetch_pods
+      pods.each do |pod|
+        if pod.completed? && pod.annotations[:'samson/show_logs_on_success']
```
grosser: better check for
`pod.annotations[:'samson/show_logs_on_success'] == 'true'`

grosser: ... since showing logs has a timeout, there is no real reason to not allow it for running pods if the user requested it

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -142,6 +142,20 @@ def fetch_pods
...
+          print_pod_logs(pod, log_end_time)
+          @output.puts "\n------------------------------------------\n"
+        end
+      end
+    rescue
```
grosser:
```suggestion
    rescue StandardError
```

grosser:
needs an actual test of `it "shows pod logs when requested"`

grosser:
let me know if you get stuck, I can help out or take it over if it gets to funky :)

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
rata: Sure! I'll change it to keyword args, thanks for the tips (really valuable for non rubyst :))

Do you think we don't need a third flag? Or a flag to only show prerequisites and show them all (i.e. without the pod.completed? filter)?

I changed it to the latter. The reason is that if we print the same after the deploy (not prerequisites, of deploys deploys) ends, it will still find the logs of the prerequisites pods and it will be printed again. I'm not sure how to filter those, nor if it's worth it.

But if we do the print after the deploys ends, without any more filters, then if something was print after the prerequisites run, then it will be printed again. And that seems confusing, it's just a duplicated print (most of the times).

Maybe after the deploys deploys, we can print but excluding pods that were completed when we printed the first time. But it seems a little messier than I'd like, and not sure the functionality it's worth the complexity.

Also, after the deploys ends, most of the time you deploy some always living objects (like deployments). The ones that are short-lived are probably printed in the first interruption, after prerequisites deploys, so there is no much gain, I think, printing after the deploy ends again.

What do you think?

plugins/kubernetes/app/models/kubernetes/api/pod.rb
rata: Good catch, thanks! Applied and squashed now :)

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
rata: Yeah, makes sense! I was in doubt, just because of this (it maybe shows logs), but didn't come up with a better name. Thanks!

I changed the func to that name, and changed the variable to `show_logs_if_requested`.

Now the conditional on the var name and on the function name make me doubt, but well... naming is hard. Let me know what you think :)

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
rata: Ohh, it makes sense! I removed that check! :)

rata:
@grosser thanks a lot for the review! And the help. I think I managed to address almost all of the comments. Added some tests too, but please check :)

I'm waiting to see if tests pass correctly now :-D

rata:
@grosser so, tests are not working. If you can help me, it would be awesome.

I modified the `batch/v1` job, adding the annotation (IIUC, without typos), and the output from pods is not there. I added some prints to debug (now on this PR, but just to show) and I really don't see why. For some reason, the pod is not having the pod annotation. But not sure why.

Running samson on my own and manually doing a deploy of a job with the annotation works. So I'm not sure I'm missing something really silly (like some typo or something obvious) or just something that changes in the tests.

Any help is appreciated :-)

rata:
My guess is that in the simulated environment for the test, something is not working. Maybe fetch_pods nor returning correctly or for some reason the pod not having the annotation.

I think is probably I don't know how to use the test infrastructure :-/

grosser:
I think you found a fundamental bug, fetch_pods should only fetch pods from given release-docs
it never caused a bug because we don't fail if more than requested pods are created, which is also a bug :D 
... I'll fix these and than this PR should be simpler :)

grosser:
ah that never was a problem since release_statuses selects filters by role_id

grosser:
taking this over now, stay tuned

PR #3009, opened 2018-10-18
Add 'failure' to failure commit status list

ilkkao:
GitHub is now returning this new status

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low


app/models/commit_status.rb
```diff
@@ -9,7 +9,7 @@ class CommitStatus
   UNDETERMINED = ["pending", "missing"].freeze
   CHECK_STATE = {
     error: ['action_required', 'canceled', 'timed_out'],
-    failure: ['failed'],
+    failure: ['failure', 'failed'],
```
grosser: remove failed ?

PR #3012, opened 2018-10-19
Show logs on success when requested

grosser:
this is meant for pre-requisite pods like migrations, but can also be useful when wanting to see what pods started up with

refactor of https://github.com/zendesk/samson/pull/3005

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -184,3 +184,9 @@ To override it, set the namespace to either `default` or `kube-system` and set `
 Samson automatically adds `container[].lifecycle.preStop` `sleep 3` if a preStop hook is not set and
 `container[].samson/preStop` is not set to `disabled`, to prevent in-flight requests from getting lost when taking a pod
 out of rotation (alternatively set `metadata.annoations.container-nameofcontainer-samson/preStop: disabled`).
+
+### Showing logs on successful deploys
+
+Set `metadata.annoations.samson/show_logs_on_deploy: 'true'` on pods, to see logs when the deploy succeeds.
+This can be useful for Migrations (see above).
+(On failure, samson always shows all pod logs)
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -63,11 +63,11 @@ module Kubernetes
       prerequisites, deploys = @release.release_docs.partition(&:prerequisite?)
       if prerequisites.any?
         @output.puts "First deploying prerequisite ..." if deploys.any?
-        return false unless deploy_and_watch(prerequisites, WAIT_FOR_PREREQUISITES)
+        return false unless deploy_and_watch(prerequisites, WAIT_FOR_PREREQUISITES, show_logs_if_requested: true)
         @output.puts "Now deploying other roles ..." if deploys.any?
       end
       if deploys.any?
-        return false unless deploy_and_watch(deploys, WAIT_FOR_LIVE)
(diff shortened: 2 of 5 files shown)


adammw (approved):

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -85,39 +85,39 @@ def wait_for_resources_to_complete(release_docs, timeout)
         statuses = pod_statuses(release_docs)
         if statuses.none?
           @output.puts "No pods were created"
-          return success
+          return success, statuses
```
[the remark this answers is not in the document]
grosser: hmm it's an internal method, and I kinda like the conciseness of it

plugins/kubernetes/README.md
```diff
@@ -184,3 +184,8 @@ To override it, set the namespace to either `default` or `kube-system` and set `
 Samson automatically adds `container[].lifecycle.preStop` `sleep 3` if a preStop hook is not set and
 `container[].samson/preStop` is not set to `disabled`, to prevent in-flight requests from getting lost when taking a pod
 out of rotation (alternatively set `metadata.annoations.container-nameofcontainer-samson/preStop: disabled`).
+
+### Showing logs on successful deploys
+
+Set `metadata.annoations.samson/show_logs_on_deploy: 'true'` on pods, to see logs when the deploy succeeds.
```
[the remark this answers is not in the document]
grosser: added

grosser:
@rata give it a try :)

rata:
@grosser will try it, thanks a lot!

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -142,22 +142,36 @@ def fetch_pods
       end
     end
 
+    def show_logs_on_deploy_if_requested(statuses)
+      pods = statuses.map(&:pod).compact.select { |p| p.annotations[:'samson/show_logs_on_deploy'] == 'true' }
+      pods.each { |pod| print_pod_details(pod, Time.now, events: false) }
```
rata: I think `Time.now` is calculated on each iteration, and as retrieving the pods can take some seconds (if some timeout happens, for example), the logs are not for the same ""snapshot"" on different pods. I don't think this is a problem and this way is simple and nice, IMHO.

But just something worth noting, perhaps.

rata: I think `Time.now` is calculated on each iteration, and as retrieving the pods can take some seconds (if some timeout happens, for example), the logs are not for the same ""snapshot"" on different pods. I don't think this is a problem and this way is simple and nice, IMHO.

But just something worth noting, perhaps.

rata: I think `Time.now` is calculated on each iteration, and as retrieving the pods can take some seconds (if some timeout happens, for example), the logs are not for the same ""snapshot"" on different pods. I don't think this is a problem and this way is simple and nice, IMHO.

But just something worth noting, perhaps.

rata (commented):
Awesome, I tested it with showing migration output on a prerequisite on staging and works just fine! Thanks!

rata:
@grosser well, GitHub seems to be having problems so sorry in advance if this appears multiple times or something. I've just tried it with printing logs for a prerequisite and works fine. Thanks a lot, really! :)

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
grosser: good idea, fixed in https://github.com/zendesk/samson/pull/3016 :)

PR #3017, opened 2018-10-22
Remove CGI escape from ref in GitHub check API call

ragurney:
Removes CGI escape which breaks the API call to GitHub's API. Turns out it actually expects branches with '/'s to keep them, for example:
```
https://api.github.com/repos/zendesk/explore_rails/commits/bboccara/chore/cleanup-bimeanalytics-references/check-runs
```
`combined_status` does the same thing:
```
https://api.github.com/repos/zendesk/explore_rails/commits/bboccara/chore/cleanup-bimeanalytics-references/status
```
/cc @zendesk/samson


ragurney:
Yeah, pretty strange 🤷‍♂️

PR #3018, opened 2018-10-23
make streaming from pending deploys work

grosser:
by waiting until the execution is ready

fixes fallout from https://github.com/zendesk/samson/pull/3013
which declared the job done when it ran on a pending deploy

@zendesk/bre @adammw

### DIFF
--- a/app/channels/job_outputs_channel.rb
+++ b/app/channels/job_outputs_channel.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 class JobOutputsChannel < ActionCable::Channel::Base
+  MAX_WAIT = 1200 # max seconds to wait for job to start when streaming
+  WAIT_DURATION = 5 # seconds between re-checking
+
   # pretends to be a controller to reuse partials and helpers
   class EventBuilder
     include ApplicationHelper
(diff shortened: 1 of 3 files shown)


app/channels/job_outputs_channel.rb
```diff
@@ -50,27 +52,28 @@ def status_response(event)
...
-      job = Job.find(id)
-      builder = EventBuilder.new(job)
-      transmit event: :started, data: builder.payload(:started, nil)
-      job.output.each_line do |line|
-        transmit event: :message, data: builder.payload(:message, line)
-      end
-      transmit event: :finished, data: builder.payload(:finished, nil)
     end
+
+    # TODO: kill threads when user stops streaming
```
grosser: got a branch for that ready `grosser/start-2`but it's kinda tricky since it would cancel streams when user opens multiple tabs and might have a race condition if user reloads and sends the disconnect too late .. so not doing that for now

grosser: kk, fixed :)

app/helpers/deploys_helper.rb
```diff
@@ -12,7 +12,7 @@ module DeploysHelper
   }.freeze
 
   def deploy_output
-    output = ActiveSupport::SafeBuffer.new
+    output = "".html_safe
```
adammw: why the change?

app/channels/job_outputs_channel.rb
```diff
@@ -50,33 +53,49 @@ def status_response(event)
...
+          execution = JobQueue.find_by_id(id)
+          break stream_execution(execution) if execution
+
+          # cancelled or we missed it running
+          job = Job.find(id)
+          break stream_finished_output(job) if job.finished?
+
+          # pending
+          sleep WAIT_DURATION
         end
```
adammw: does it close the connection if it gets here (times out waiting)?

app/helpers/deploys_helper.rb
grosser: just readability

app/channels/job_outputs_channel.rb
grosser: the channel does not get closed automatically, but since it's not doing anything it does not really matter (same as before)

adammw: seems like it should get closed if it gets to here, unless there's a reason not to (like it'll reconnect or something)

grosser: if we do a `connection.close` then the client reconnects and streams the whole output again :)
... there might be a way to make this work, but it involves a lot of digging/checking for little gain (closing connections that do nothing, not sure if it even removes the channel and which are auto-closed when user closes the tab)

PR #3020, opened 2018-10-26
allow replacing template parts via env

grosser:
for ShardedDeployment I need to configure `spec.shards.list` with an array, I could add custom logic for that, but that does not really scale well ... so how about something generic ?

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -190,3 +190,13 @@ out of rotation (alternatively set `metadata.annoations.container-nameofcontaine
 Set `metadata.annoations.samson/show_logs_on_deploy: 'true'` on pods, to see logs when the deploy succeeds.
 This can be useful for Migrations (see above).
 (On failure, samson always shows all pod logs)
+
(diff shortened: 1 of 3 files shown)


adammw:
seems a little strange.. what's the usecase/story?

grosser:
see PR description ;)

plugins/kubernetes/README.md
```diff
@@ -190,3 +190,13 @@ out of rotation (alternatively set `metadata.annoations.container-nameofcontaine
...
 This can be useful for Migrations (see above).
 (On failure, samson always shows all pod logs)
+
+### Changing templates via ENV
+
+For custom things that need to be different between environments/deploy-groups.
+
+Use an annotation to configure what will to be replaced:
+```
+metadata.annoations.samson/set_via_env_json/metadata.labels.custom: SOME_ENV_VAR
```

grosser:
needed to change to `samson/set_via_env_json-` since there can only be a single / in an annotation key ...

adammw (approved):

PR #3027, opened 2018-10-29
do not pick a random context if saved context is no longer available

grosser:
@zendesk/compute 

screen shot 2018-10-29 at 2 40 23 pm
screen shot 2018-10-29 at 2 43 03 pm

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/cluster.rb
+++ b/plugins/kubernetes/app/models/kubernetes/cluster.rb
@@ -23,10 +23,6 @@ module Kubernetes
       (@client ||= {})[type] ||= build_client(type)
     end
 
-    def context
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/cluster.rb
```diff
@@ -54,7 +54,7 @@ def server_version
 
     def connection_valid?
       client('v1').api_valid?
-    rescue *SamsonKubernetes.connection_errors
+    rescue KeyError, *SamsonKubernetes.connection_errors
```
adammw: capture earlier and reraise as a custom exception?

adammw (approved):
👍 but would like to see the KeyError fixed.

PR #3028, opened 2018-10-30
show logs from init containers on 1.7+ clusters

grosser:
also unifies init container logic which caused this bug / https://github.com/zendesk/samson/pull/2616

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -5,6 +5,14 @@ module Kubernetes
       INIT_CONTAINER_KEY = :'pod.beta.kubernetes.io/init-containers'
       INGORED_AUTOSCALE_EVENT_REASONS = %w[FailedGetMetrics FailedRescale].freeze
 
+      def self.init_containers(pod)
+        containers = pod.dig(:spec, :initContainers) || []
+        if json = pod.dig(:metadata, :annotations, Kubernetes::Api::Pod::INIT_CONTAINER_KEY)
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -5,6 +5,14 @@ class Pod
       INIT_CONTAINER_KEY = :'pod.beta.kubernetes.io/init-containers'
       INGORED_AUTOSCALE_EVENT_REASONS = %w[FailedGetMetrics FailedRescale].freeze
 
+      def self.init_containers(pod)
+        containers = pod.dig(:spec, :initContainers) || []
+        if json = pod.dig(:metadata, :annotations, Kubernetes::Api::Pod::INIT_CONTAINER_KEY)
+          containers += JSON.parse(json, symbolize_names: true)
```
adammw: won't this end up with duplicates if both are specified (e.g. on 1.7)?

grosser: yeah kinda ... migrated pods will have both, maybe some other conditions ... I'll add a uniq

adammw: I have a feeling kubernetes actually picks one or the other rather than merging, but i'm not sure which / can't find documentation on that.

grosser: would depend on the cluster I guess ... but does not hurt to try to find logs for both I'd guess

adammw (approved):

PR #3031, opened 2018-10-30
docs

grosser:
@adammw


adammw:
can we add a comment to the resource edit form too (if it matches that starts with pattern)?

adammw (approved):

grosser:
screen shot 2018-10-30 at 1 23 15 pm

PR #3032, opened 2018-10-30
allow filtering for limits that are on all projects or all scopes and…

grosser:
… disable seperator

 - before "All" just showed all projects/scopes but not "where project is all"
 - searching for no scope would match the `----` option and show that which was ugly

@zendesk/bre @zendesk/compute 

screen shot 2018-10-30 at 2 14 25 pm

### DIFF
--- a/app/models/environment.rb
+++ b/app/models/environment.rb
@@ -17,7 +17,7 @@ class Environment < ActiveRecord::Base
   def self.env_deploy_group_array(include_all: true)
     all = include_all ? [["All", nil]] : []
     envs = Environment.all.map { |env| [env.name, "Environment-#{env.id}"] }
-    separator = [["----", nil]]
+    separator = [["----", "disabled", {disabled: true}]]
     deploy_groups = DeployGroup.all.sort_by(&:natural_order).map { |dg| [dg.name, "DeployGroup-#{dg.id}"] }
     all + envs + separator + deploy_groups
   end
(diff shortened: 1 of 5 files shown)


app/models/environment.rb
```diff
@@ -17,7 +17,7 @@ class Environment < ActiveRecord::Base
   def self.env_deploy_group_array(include_all: true)
     all = include_all ? [["All", nil]] : []
     envs = Environment.all.map { |env| [env.name, "Environment-#{env.id}"] }
-    separator = [["----", nil]]
+    separator = [["----", "disabled", {disabled: "disabled"}]]
```
adammw: does `class: 'divider'` work?

plugins/kubernetes/app/views/kubernetes/usage_limits/index.html.erb
```diff
@@ -4,8 +4,8 @@
   <%= page_title "Kubernetes Limits" %>
 
   <%= search_form do %>
-    <%= search_select :project_id, [['All', '']] + @projects.map { |p| [p.name, p.id] }, live: true, size: 3 %>
-    <%= search_select :scope_type_and_id, @env_deploy_group_array, live: true, size: 3, label: "Scope" %>
+    <%= search_select :project_id, [['All', 'all']] + @projects.map { |p| [p.name, p.id] }, live: true, size: 3 %>
```
adammw: maybe make 'all' a constant and less likely to collide with a real id (or are id's numeric)?

grosser:
added

adammw (approved):

PR #3038, opened 2018-11-05
most new resources require a resourceVersion so make that the default

grosser:
seeing `JobExecution failed: Kubernetes error  "example" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update`

opened https://github.com/kubernetes/kubernetes/issues/70674 to get a solution in the api, but doubt that will work out
`
@zendesk/compute


adammw (approved):

grosser:
using 0 / -1 / 999999999 does not work :(

PR #3044, opened 2018-11-13
Use ruby 2.5.3 (security update)

rata:
* Updates ruby version because of the security update: https://www.ruby-lang.org/en/news/2018/10/17/ruby-2-5-2-released/ and https://www.ruby-lang.org/en/news/2018/10/18/ruby-2-5-3-released/

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


rata:
@grosser oh, sorry I missed it! Now `.ruby-version` too :)

rata:
I'm not sure how to the version heroku expects. I can change the Gemfile.lock file manually, but not sure what versions heroku uses (right now the Gemfile.lock has 2.5.1p57, not sure which suffix should be for 2.5.3 in heroku).

If anyone with access to heroku zendesk projects can check and modify this, please do!

grosser:
bundle once so the Gemfile.lock changes ... heroku will use .ruby-version

rata:
@grosser thanks! The "FOSSA" test is failing, but seems unrelated. Do you happen to know why?  I just rebased. Can't find what was causing FOSA to fail in some PR in the past...

PR #3050, opened 2018-11-16
rescue harder to make it not blow up migrations

grosser:
fixes issue on https://github.com/zendesk/samson/pull/1889

@esselius can you provide more details on what the exception was exactly ?

PG::InFailedSqlTransaction

### DIFF
--- a/db/migrate/20181111003546_remove_binary_builder.rb
+++ b/db/migrate/20181111003546_remove_binary_builder.rb
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 class RemoveBinaryBuilder < ActiveRecord::Migration[5.2]
+  class Stage < ActiveRecord::Base
+  end
+
   def change
-    remove_column :stages, :docker_binary_plugin_enabled
-  rescue StandardError # rubocop:disable Lint/HandleExceptions
-    # user might have never installed the plugin
+    # postgres cannot rescue aleter statements, so we need to be sure the up migration for this ran
+    if Stage.column_names.include? 'docker_binary_plugin_enabled'
+      remove_column :stages, :docker_binary_plugin_enabled
+    end
   end
 end


esselius:
@grosser https://gist.github.com/esselius/7726d0e3c386a988ba61c0a798d979ad

grosser:
that triggered a PG::InFailedSqlTransaction ... which is a child of StandardError ... so this should have worked oO

grosser:
also the logs show that RemoveBinaryBuilder finished ... so maybe this is something else ?

grosser:
also the logs show that RemoveBinaryBuilder finished ... so maybe this is something else ?

grosser:
also the logs show that RemoveBinaryBuilder finished ... so maybe this is something else ?

grosser:
ah I read that backwards ...

grosser:
ok, fixed it ... issue is https://github.com/rails/rails/issues/12330 ... connection gets poisoned once a statement fails ... rolling back also gets it into a weird state, so let's just never break the connection in the first place :)

grosser:
@ragurney 👁

grosser:
thx for reporting this, these kind of bugs are hard to track down :D

db/migrate/20181111003546_remove_binary_builder.rb
```diff
@@ -1,8 +1,12 @@
...
 class RemoveBinaryBuilder < ActiveRecord::Migration[5.2]
+  class Stage < ActiveRecord::Base
+  end
+
   def change
-    remove_column :stages, :docker_binary_plugin_enabled
-  rescue StandardError # rubocop:disable Lint/HandleExceptions
-    # user might have never installed the plugin
+    # postgres cannot rescue aleter statements, so we need to be sure the up migration for this ran
+    if Stage.column_names.include? 'docker_binary_plugin_enabled'
```
ragurney: Nice, seems like a good idea 👍

PR #3058, opened 2018-11-26
prerequisite_stages: refactor strange yield logic and better test

grosser:
@zendesk/bre 

/cc @zendesk/compute

### DIFF
--- a/plugins/prerequisite_stages/lib/samson_prerequisite_stages/samson_plugin.rb
+++ b/plugins/prerequisite_stages/lib/samson_prerequisite_stages/samson_plugin.rb
@@ -2,31 +2,30 @@
 
 module SamsonPrerequisiteStages
   class Engine < Rails::Engine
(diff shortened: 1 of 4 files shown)


plugins/prerequisite_stages/lib/samson_prerequisite_stages/samson_plugin.rb
```diff
@@ -3,30 +3,28 @@
 module SamsonPrerequisiteStages
   class Engine < Rails::Engine
     def self.execute_if_unmet_prereq_stages(stage, reference)
```
ragurney: Maybe change the name of this to something else now? e.g. `check_unmet_prerequisite_stages`?

grosser: done, also renamed prereq -> prerequisite and moved from `Engine` to base

PR #3062, opened 2018-11-27
remove dead methods from OutputBuffer / never block forever when adding listener during closing

grosser:
@zendesk/bre @zendesk/compute 

 - remove invalid utf8
 - remove unused methods
 - remove `white chunk = ` logic which never worked since queue.pop blocks
 - make sure we never add a listener after we close since that listener would hang forever on the .pop (race condition)

### DIFF
--- a/app/models/output_buffer.rb
+++ b/app/models/output_buffer.rb
@@ -31,6 +31,7 @@ class OutputBuffer
     @previous = ThreadSafe::Array.new
     @closed = false
     @line_finished = true
+    @mutex = Mutex.new
   end
 
   def puts(line = "")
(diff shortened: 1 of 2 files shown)


app/models/output_buffer.rb
```diff
@@ -63,34 +57,36 @@ def to_s
...
+      return @previous.each(&block) if closed?
       @listeners << queue
+    end
 
-      # race condition: possibly duplicate messages when message comes in between adding listener and this
-      @previous.each(&block)
+    # race condition: possibly duplicate messages when message comes in between adding listener and this
+    @previous.each(&block)
 
-      while (chunk = queue.pop) && chunk.first != :close
```
adammw: were there no tests checking the behaviour when the chunk was :close ?

grosser:
the tests were checking that it closes / stops, so just removing it and not
replacing would have broken them

PR #3067, opened 2018-11-28
fix vulnerability scanning is broken because google changed the api

grosser:
following https://cloud.google.com/container-registry/docs/reference/rest/v1alpha1/projects.occurrences#analysisstatus after seeing reports that it does not work anymore

/cc @zendesk/bre @zendesk/compute


plugins/gcloud/lib/samson_gcloud/image_scanner.rb
```diff
@@ -11,11 +11,18 @@ class ImageScanner
...
-      # the occurance with type DISCOVERY tells us if the scan is done, so we do not mark pending as "Nothing found"
+      # which has many thousand occurrence for very simple images
+      # the occurrence with type DISCOVERY tells us the scan status
+      # https://cloud.google.com/container-registry/docs/reference/rest/v1alpha1/projects.occurrences#analysisstatus
       def scan(build)
         return ERROR unless result = request(build, "occurrences", "kind=\"DISCOVERY\"")
-        return WAITING unless result.dig("occurrences", 0, "discovered", "operation", "done")
+
+        case result.dig("occurrences", 0, "discovered", "analysisStatus")
+        when "FINISHED_SUCCESS" # rubocop:disable Lint/EmptyWhen fall through
```
adammw: would prefer something like since the empty fall through isn't that clear...
```ruby
unless (status = result.dig("occurrences", 0, "discovered", "analysisStatus")) == "FINISHED_SUCCESS"
  return WAITING if ["PENDING", "SCANNING"].include?(status)
  return ERROR
end
```

PR #3068, opened 2018-11-29
do not fail jenkins when any comitter/user had no email

grosser:
was browsing rollbar for sth else and saw this ...
```
NoMethodError: undefined method `include?' for nil:NilClass
1
File "plugins/jenkins/app/models/samson/jenkins.rb" line 252 in block in notify_emails
```

@zendesk/compute


ragurney:
Maybe add accompanying test?

grosser:
added ... these jenkins tests are 🤢

PR #3071, opened 2018-11-30
make us not mix up dev/staging/prod samson instances

grosser:
@zendesk/bre 
/cc @zendesk/compute 

restarted the wrong one by accident this morning ... we need some way of visually distinguishing them :)

screen shot 2018-11-30 at 9 27 57 am


app/views/layouts/_header.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="navbar navbar-inverse">
```
adammw: I would even say make it a different colour depending on the env...

grosser:
screen shot 2018-11-30 at 10 04 40 am

PR #3072, opened 2018-11-30
do not catch all file errors, it hid an unrelated error

grosser:
@zendesk/bre @adammw

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -273,14 +273,10 @@ class JobExecution
 
   # TODO: @repository.checkout_workspace should manage the tempdir and prune after
   def make_tempdir
-    result = nil
-    Dir.mktmpdir("samson-#{@job.project.permalink}-#{@job.id}") do |dir|
-      result = yield dir
-    end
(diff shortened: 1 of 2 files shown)


app/models/job_execution.rb
```diff
@@ -273,14 +273,10 @@ def build_finder
...
-    Dir.mktmpdir("samson-#{@job.project.permalink}-#{@job.id}") do |dir|
-      result = yield dir
-    end
-  rescue Errno::ENOTEMPTY, Errno::ENOENT
-    ErrorNotifier.notify("Notify: make_tempdir error #{$!.message.split('@').first}")
-    result # tempdir ensure sometimes fails ... not sure why ... return normally
+    dir = Dir.mktmpdir("samson-#{@job.project.permalink}-#{@job.id}")
+    yield dir
   ensure
+    FileUtils.rm_rf dir # mktmpdir with block often raised errors when tempfolder got removed early
```
adammw: won't this raise as well? or is that okay?

grosser: won't raise since it's `rm -rf`

PR #3074, opened 2018-11-30
do not add pod-disruption-budgets for DaemonSet

grosser:
it should already have deleted it since the math came out as 0, but also no need to add it in the first place

@zendesk/compute


jonmoter (approved):
👍 on 💚

PR #3075, opened 2018-12-01
simplify cancel logic to use exceptions

grosser:
this makes everything more serial and less "threads that work in parallel to do stuff"
also removes all the cancellation logic that @adammw pointed out as being 💩 

@zendesk/bre @zendesk/compute

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -77,7 +77,7 @@ class Job < ActiveRecord::Base
     update_attribute(:canceller, canceller) unless self.canceller
 
     if ex
-      ex.cancel # switches job status in the runner thread for consistent status in after_deploy hooks
+      JobQueue.cancel(id) # switches job status in the runner thread for consistent status in after_deploy hooks
     else
       cancelled!
     end
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -13,7 +13,14 @@ class JobExecution
 
   def initialize(reference, job, env: {}, output: OutputBuffer.new, &block)
     @output = output
-    @executor = TerminalExecutor.new(@output, verbose: true, deploy: job.deploy, project: job.project)
+    @executor = TerminalExecutor.new(
+      @output,
+      verbose: true,
+      deploy: job.deploy,
+      project: job.project,
+      timeout: Rails.application.config.samson.deploy_timeout,
+      cancel_timeout: cancel_timeout
+    )
     @viewers = JobViewers.new(@output)
 
     @start_callbacks = []
@@ -22,7 +29,6 @@ class JobExecution
     @job = job
     @reference = reference
     @execution_block = block
-    @cancelled = false
     @finished = false
 
     @repository = @job.project.repository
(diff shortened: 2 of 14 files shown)


app/models/job_execution.rb
```diff
@@ -13,7 +13,14 @@ class JobExecution
...
   def initialize(reference, job, env: {}, output: OutputBuffer.new, &block)
     @output = output
-    @executor = TerminalExecutor.new(@output, verbose: true, deploy: job.deploy, project: job.project)
+    @executor = TerminalExecutor.new(
+      @output,
+      verbose: true,
+      deploy: job.deploy,
+      project: job.project,
+      timeout: Rails.application.config.samson.deploy_timeout,
+      cancel_timeout: cancel_timeout
```
bcolferzd: Where is this value set?

bcolferzd (approved):
LGTM

app/models/job_execution.rb
grosser: class attr-accessor

app/models/job.rb
```diff
@@ -77,7 +77,7 @@ def cancel(canceller)
     update_attribute(:canceller, canceller) unless self.canceller
 
     if ex
-      ex.cancel # switches job status in the runner thread for consistent status in after_deploy hooks
+      JobQueue.cancel(id) # switches job status in the runner thread for consistent status in after_deploy hooks
     else
       cancelled!
```
adammw: wouldn't you want this always called?

app/models/job.rb
```diff
@@ -77,7 +77,7 @@ def cancel(canceller)
     update_attribute(:canceller, canceller) unless self.canceller
 
     if ex
```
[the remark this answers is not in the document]
adammw: or just always call it and have jobqueue nop if it can't find the id

app/models/job_queue.rb
```diff
@@ -104,7 +110,13 @@ def perform_job(job_execution, queue)
 
     @threads[job_execution.id] = Thread.new do
       begin
-        job_execution.perform
+        ActiveRecord::Base.connection_pool.with_connection do
+          begin
+            job_execution.perform
+          rescue Cancel
```
adammw: what about successful deploys, don't want to use `ensure` and throw it away too?

app/models/terminal_executor.rb
```diff
@@ -75,6 +75,20 @@ def quiet
 
   private
 
+  def cancel(timeout:)
+    kill "INT"
+    timeout.times do
+      return unless kill(0) # stop when not running
```
adammw: so you signal 0 each second? that seems odd...

app/models/terminal_executor.rb
```diff
@@ -22,43 +22,43 @@ class TerminalExecutor
...
-          ensure
-            input.close
           end
+        rescue Timeout::Error
+          @output.puts "Timeout: execution took longer then #{timeout}s and was terminated"
+          cancel timeout: 1
+          false
+        rescue Errno::ECHILD
+          @output.puts "#{$!.class}: #{$!.message}"
+          cancel timeout: 1
```
adammw: make 1 a constant..

app/models/terminal_executor.rb
```diff
@@ -22,43 +22,43 @@ class TerminalExecutor
...
   def execute(*commands, timeout: @timeout)
-    return false if @cancelled
-    options = {in: '/dev/null', unsetenv_others: true}
     script_as_executable(script(commands)) do |command|
-      output, input, pid = PTY.spawn(whitelisted_env, command, options)
-      record_pid(pid) do
-        timeout_execution(timeout) do
-          stream from: output, to: @output
+      output, input, pid = PTY.spawn(whitelisted_env, command, in: '/dev/null', unsetenv_others: true)
+      input.close # close early so nothing can hang forever
```
adammw: this is new - have you tested this doesn't break weird streaming / process behaviour?

app/models/terminal_executor.rb
adammw: huh. TIL: https://stackoverflow.com/questions/11012527/what-does-kill-0-pid-in-a-shell-script-do

adammw: maybe make a `pgid_running?` command that internally does `kill(0)` so it's more obvious?

adammw (approved):
Nice, I like it.

app/models/job.rb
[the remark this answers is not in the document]
adammw: but it should always find the same object no, just the values on the object change?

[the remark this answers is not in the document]
grosser: it's the same record, but a different ruby object, so it won't update without a reload

Issue #3080, opened 2018-12-05
How does the preventing with preStop hook works?


rata:
Hi!

I don't follow how adding a 'sleep 3' as pre-stop hook makes a deploy not lose requests. Can you please explain what am I missing?

IIUC, If all the containers in a pod have this (as it may happen), then they will all die ~3 seconds later. Why doing it would change anything regarding the coming requests? 

I can see it happens if the pod is removed from the service and this sleep is enough to finish the inflight requests. But this is very fragile, and there is an inherent race condition there, as it is quite possible to not finish the inflight requests in that time.

We are using nginx and as a prestop we do an `nginx -s quit`. That will finish nginx when the inflight requests have been processed, I think.

Am I missing something? How is this supposed to work?

Also, I'm not sure we want to inject this in all pods. Would it be okay to add an env var to disable this behavior by default? (instead of changing all our yamls to use the annotation)

grosser:
you should totally use `nginx -s quit`

the 3s is for the time our routing layer takes to react to service changes since we don't send to the kubernetes service but to consul which knows about all the endpoints directly, but has a slight lag

(feel free to reopen if that's still unclear/confusing)

rata:
Not sure I followed in detail, but I understand it's something quite specific to your setup and we will continue to use `nginx -s quit`. That is enough, I think :)

Thanks!

PR #3081, opened 2018-12-05
plugins/kubernetes: Add env var to never add a preStop

rata:
* Add env variable to disable preStop injection in kubernetes plugin

Relates to #3080

The reason is the one explained in the bug: it's TOO much to modify all yamls to add this annotation. Just having an option to change the default is way more easy for us.  What do you think? Does it make sense? 

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -181,6 +181,8 @@ To override it, set the namespace to either `default` or `kube-system` and set `
 
 ### Preventing request loss with preStop
 
+To enable the following functionality you need to set `KUBERNETES_ADD_PRESTOP=true`.
+
 Samson automatically adds `container[].lifecycle.preStop` `sleep 3` if a preStop hook is not set and
 `container[].samson/preStop` is not set to `disabled`, to prevent in-flight requests from getting lost when taking a pod
 out of rotation (alternatively set `metadata.annoations.container-nameofcontainer-samson/preStop: disabled`).
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -6,6 +6,7 @@ module Kubernetes
 
     CUSTOM_UNIQUE_LABEL_KEY = 'rc_unique_identifier'
     SECRET_PULLER_IMAGE = ENV['SECRET_PULLER_IMAGE'].presence
+    KUBERNETES_ADD_PRESTOP = !!(ENV.fetch('KUBERNETES_ADD_PRESTOP', false))
     SECRET_PREFIX = "secret/"
 
     def initialize(release_doc, template, index:)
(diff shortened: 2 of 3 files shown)


plugins/kubernetes/README.md
```diff
@@ -185,6 +185,8 @@ Samson automatically adds `container[].lifecycle.preStop` `sleep 3` if a preStop
 `container[].samson/preStop` is not set to `disabled`, to prevent in-flight requests from getting lost when taking a pod
 out of rotation (alternatively set `metadata.annoations.container-nameofcontainer-samson/preStop: disabled`).
 
+The default behavior can be to never add this by setting the `KUBERNETES_DONT_ADD_PRESTOP=false`
```
grosser: better to let people opt-in since this is kinda hacky anyway so only do it if`KUBERNETES_ADD_PRESTOP=true`

rata: Oh, will do it that way. Thanks!

rata: I didn't want to force changes on your side =)

rata:
@grosser what about now? :)

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -6,6 +6,7 @@ class TemplateFiller
 
     CUSTOM_UNIQUE_LABEL_KEY = 'rc_unique_identifier'
     SECRET_PULLER_IMAGE = ENV['SECRET_PULLER_IMAGE'].presence
+    KUBERNETES_ADD_PRESTOP = !!(ENV.fetch('KUBERNETES_ADD_PRESTOP', false))
```
grosser:
```suggestion
    KUBERNETES_ADD_PRESTOP = ENV['KUBERNETES_ADD_PRESTOP'] == 'true'
```

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -509,6 +510,7 @@ def set_image_pull_secrets
     end
 
     def set_pre_stop
+      return if ! KUBERNETES_ADD_PRESTOP
```
grosser:
```suggestion
      return unless KUBERNETES_ADD_PRESTOP
```

rata:
Oh, this 4 lines make the test coverage go under the threshold. Will try to add some test tomorrow :)

rata:
@grosser not sure how to add a test for that (sorry, I really should learn something about ruby sometime). Can you give me a hint? :)

grosser:
looks pretty close, I'll fix it up :)

rata:
@grosser thanks a lot! I'll take a look afterwards, I'm curious :)

Thanks again

PR #3083, opened 2018-12-06
 plugins/kubernetes: Change preStop to disabled by default

grosser:
replaces https://github.com/zendesk/samson/pull/3081

@rata 

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -186,6 +186,8 @@ To override it, set the namespace to either `default` or `kube-system` and set `
 
 ### Preventing request loss with preStop
 
+To enable the following functionality you need to set `KUBERNETES_ADD_PRESTOP=true`.
+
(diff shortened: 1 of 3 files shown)


rata (approved):

plugins/kubernetes/README.md
```diff
@@ -186,6 +186,8 @@ To override it, set the namespace to either `default` or `kube-system` and set `
 
 ### Preventing request loss with preStop
 
+To enable the following functionality you need to set `KUBERNETES_ADD_PRESTOP=true`.
```
jonmoter: Do we use `true` consistently throughout Samson? Or do we have some "boolean" ENV vars that also accept `1`, or just having the environment variable set?

jonmoter (approved):
Seems reasonable to me.

plugins/kubernetes/README.md
grosser: we use it in a few places ... I guess Samson::EnvCheck is the best

PR #3086, opened 2018-12-06
docs for service_selector_across_roles

grosser:
followup to https://github.com/zendesk/samson/pull/2927

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -5,7 +5,7 @@ Allows Samson to be able to deploy to [Kubernetes](kubernetes.io), and includes 
 ## Overview
 
 Samson will talk to N Kubernetes clusters via their API.
-Clusters can be running locally (Docker For Mac / Minikube), in a datacenter, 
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/README.md
```diff
@@ -211,3 +211,9 @@ Use an annotation to configure what will to be replaced:
 metadata.annotations.samson/set_via_env_json-metadata.labels.custom: SOME_ENV_VAR
 ```
 Then configure an ENV var with that same name and a value that is valid JSON.
+
+### Disabeling service selector validation
```

plugins/kubernetes/README.md
```diff
@@ -211,3 +211,9 @@ Use an annotation to configure what will to be replaced:
...
+
+To debug services or to create Gateways it can be useful to disable selector validations.
```
adammw: create resources that need to reference a selector that doesn't include team/role, you can disable selector validation with:
...

(and then and example of a gateway)

grosser:
fixed

adammw (approved):

PR #3089, opened 2018-12-07
allow some percentage of failed pods so big deployments do not random…

grosser:
…ly fail

@zendesk/compute 

```
[19:44:59] Now deploying other roles ...
[19:44:59] Deploying GroupK role server
[19:44:59] Waiting for pods to be created
[19:44:59] Ignored 1 notready pods
[19:44:59] Deploy status:
[19:44:59]   GroupK server: Missing
[19:44:59] SUCCESS
```

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -211,3 +211,8 @@ Use an annotation to configure what will to be replaced:
 metadata.annotations.samson/set_via_env_json-metadata.labels.custom: SOME_ENV_VAR
 ```
 Then configure an ENV var with that same name and a value that is valid JSON.
+
+### Allow randomly not-ready pods during redines check
+
+Set `KUBERNETES_ALLOW_NOT_READY_PERCENT=10` to allow up to 10% of pods being not-ready,
+this is useful when dealing with large deployments that have some random failures.
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -87,19 +87,21 @@ module Kubernetes
           @output.puts "No pods were created"
           return success, statuses
         end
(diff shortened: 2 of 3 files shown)


plugins/kubernetes/README.md
```diff
@@ -211,3 +211,8 @@ Use an annotation to configure what will to be replaced:
 metadata.annotations.samson/set_via_env_json-metadata.labels.custom: SOME_ENV_VAR
 ```
 Then configure an ENV var with that same name and a value that is valid JSON.
+
+### Ignore randoly not-ready pods during redines check
```

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -88,7 +88,8 @@ def wait_for_resources_to_complete(release_docs, timeout)
           @output.puts "No pods were created"
           return success, statuses
         end
-        not_ready = statuses.reject(&:live)
+
+        not_ready = not_ready_by_percent(statuses)
```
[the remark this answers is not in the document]
grosser: tried that at first and looked complicated ... ended up refactoring it and now this seems like the nicer solution ... will change

grosser:
updated

plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
```diff
@@ -675,6 +675,26 @@ def worker_is_unstable
...
   end
 
+  describe "#allowed_not_ready" do
+    let(:log_string) { "Ignored" }
+
+    it "ignores none when percent is not set" do
+      executor.send(:allowed_not_ready, 10).must_equal 0
+    end
+
+    it "ignores given percentage" do
```
adammw: Ignores sounds wrong in this context, should be like returns number based on ..

adammw (approved):

PR #3092, opened 2018-12-12
do not store the windows newlines browsers send in

grosser:
@zendesk/compute @mustafaturan 

/fyi @zendesk/bre

### DIFF
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -108,11 +108,15 @@ class SecretsController < ApplicationController
   private
 
   def secret_params
-    @secret_params ||= params.require(:secret).permit(
-      *Samson::Secrets::Manager::ID_PARTS,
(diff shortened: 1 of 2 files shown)


app/controllers/secrets_controller.rb
```diff
@@ -108,11 +108,15 @@ def destroy
...
-      *UPDATEDABLE_ATTRIBUTES,
-      :allow_duplicates
-    )
+    @secret_params ||= begin
+      sent = params.require(:secret).permit(
+        *Samson::Secrets::Manager::ID_PARTS,
+        *UPDATEDABLE_ATTRIBUTES,
+        :allow_duplicates
+      )
+      sent[:value] = sent[:value].delete("\r") if sent[:value]
```
adammw: wouldn't `gsub("\r\n", "\n")` be safer?

grosser: added

PR #3095, opened 2018-12-14
improve dockerignore / gitignore / schema dumping

grosser:
previously images were pushed to github with whatever local config/database.yml the builder had
which would most likely not work for anyone else or worst-case could even include remote db credentials

@zendesk/bre @dndungu 
/cc @zendesk/compute


.dockerignore
```diff
@@ -1,12 +1,41 @@
...
-node_modules
+
+# user might chose to have local gem installs, they can be quiet big
 vendor/bundle
 
-public/assets
+# editor-specific folders that can be big
+.idea
+
+## copied from .gitignore
```
adammw: doesn't dockerconfig support including the gitignore? why copy it?

grosser: nope

adammw: I must be thinking of the gcloud ignore...

grosser:
yep

PR #3097, opened 2018-12-19
allow any namespace to be used when using override

grosser:
we'll need that for istio / nginx namespaces

@zendesk/compute


adammw (approved):
This one makes more sense to me than making another annotation.

grosser:
yeah, the other one felt cleaner to me since it's not really a system service, but meh ...

PR #3098, opened 2018-12-19
more generic way of keeping namespace

grosser:
alternative for https://github.com/zendesk/samson/pull/3097

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -186,7 +186,8 @@ to make all kubernetes deploys that do not use a `metadata.labels.team` / `spec.
 ### Using custom namespace
 
 Samson always sets a namespace according to the deploygroups `kubernetes_namespace`.
-To override it, set the namespace to either `default` or `kube-system` and set `metadata.kubernetes.io/cluster-service: 'true'`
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/README.md
```diff
@@ -186,7 +186,8 @@ to make all kubernetes deploys that do not use a `metadata.labels.team` / `spec.
 ### Using custom namespace
 
 Samson always sets a namespace according to the deploygroups `kubernetes_namespace`.
-To override it, set the namespace to either `default` or `kube-system` and set `metadata.kubernetes.io/cluster-service: 'true'`
+If `metadata.labels.kubernetes.io/cluster-service: 'true'` is set, it will keep `default` or `kube-system`. 
+If `metadata.annotations.samson/keep_namespace: 'true'` is set, it will keep anything.
```
hgokavarapuz: I am not completely sure if keeps_namespace to true will be useful. What if two other teams use the same namespace name.... We will not be able to control that name collision right.

grosser: yep :D

grosser:
closing for now since https://github.com/zendesk/samson/pull/3097 got us covered for now

PR #3104, opened 2018-12-28
do not require a build for secret-puller

grosser:
this should be about the fix we need, but I still don't know why it currently works or why adding an additional container broke it, so trying to figure that out first and then adding a failing test for that ...

@zendesk/bre @zendesk/compute


hgokavarapuz (approved):

grosser:
I think the bug happened because the verification template also assigns secrets and that added the init container, fixed that now too by always passing a copy to templatefiller

PR #3107, opened 2019-01-07
Update _log.html.erb

sandeepparhadonror:
This is my first OSC .  :) 
done the code indentation.

/cc @zendesk/samson

### Risks
- Level: Low


grosser:
that's there on purpose to not mess up the printed indentation (leading whitespace)
if you confirmed that it's still the same, let me know and I'll reopen, but afaik that's why it's there ... I already tried removing it once :(

grosser:
keep trying, good luck :)
docs for things that are hard to understand are also good, or even better
refactoring them :)

PR #3111, opened 2019-01-08
Allow deploying stages on new builds

### LINKED ISSUE #3110: Trigger deploy on new Build
It's possible to trigger a deploy whenever a new Release is created, but it's not possible to trigger a deploy when there's a new Build.

The reason for wanting this is that a Build may have side effects (e.g. uploading artifacts to a registry) and some pipelines need to wait for the Build to complete. A Release is created immediately on merges, so a deploy that depends on those artifacts will fail.

dasch:
WIP

Fixes #3110.

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -58,6 +58,15 @@ class BuildsController < ApplicationController
     start_docker_build if saved && !registering_external_build?
 
     status = new ? :created : :ok
+
+    if new
+      deploy_service = DeployService.new(build.creator)
+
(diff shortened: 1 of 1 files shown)


grosser:
(repeating slack comment here for findability)
why can't the deploy wait for the build as we normally do ?
if it's not kubernetes, check the `wait for docker image` (or so) checkbox

dasch:
Ah, didn't realize that's what that checkbox did!

PR #3115, opened 2019-01-09
PAAS-1240 allow ignoring roles from kubernetes stages

grosser:
We can store the "is this role ignored on this stage" information on a new role-stage join table that is managed via stage edit view.
(trying to avoid having to add one for every possible permutation, but instead defining them when users want them)

screen shot 2019-01-09 at 5 44 00 pm

@zenhao @aglover-zendesk @adammw @tmcinerney 

/cc @zendesk/compute


zenhao:
I think the UI make sense to me. Thanks for working on this.

grosser:
@zendesk/compute ready for review

hgokavarapuz (approved):

Issue #3108, opened 2019-01-08
Kubernetes and the order of environment variables


thrawny:
In a kubernetes manifest you can reference an env variable from another using the `$(FOO)` notation. However, `FOO` in this case needs to be defined before it can be referenced.

This causes problems in samson when combining env variables from a yaml file, project and env variable groups. Also of course the ones samson adds, like `DEPLOY_GROUP`. 

As far as I know without having dug deeply into the code, the order is as follows;
- yaml file
- samson (i.e. deploy_group, revision)
- project
- env variable group (in the order they were added for the project?)

So. A project env variable cannot reference an env variable group one and so on. A common case for us that we have an env variable group with some global style variables, like `ENVIRONMENT_NAME`. And then for a specific project we want to construct an env variable using `ENVIRONMENT_NAME`. Currently this is impossible without redefining `ENVIRONMENT_NAME` over and over.
Example:
Project -> `SOME_URL: $(ENVIRONMENT_NAME).example.com`
Env variable group -> `ENVIRONMENT_NAME: foo`
The order will be wrong in the kubernetes manifest so the `SOME_URL` variable will not be interpolated.

Here is a gist showing an example configuration:
https://gist.github.com/esselius/031bdb4d24bc49d77bb4d240c05784a2

My suggestion is that env variables are sorted based on how they are referencing each other, perhaps throwing an error if an impossible situation is encountered. I.e. if two env variables are referencing each other.

I would be happy to make a PR unless someone has a workaround or something that alleviates the need for this.

grosser:
Easy workaround would be to have the app that uses this url build the url itself with then ENVIRONMENT_NAME env var.

To make the ordering of env variables take into account dependencies would be complex and could cause loops, so I'd prefer to avoid that if possible. Also the current resolution should work just fine unless ENVIRONMENT_NAME itself is referencing another env variable, since the resolution is done after all env variables are merged.

If some kind of parsing is needed though, I think `resolve_dollar_variables` in `plugins/env/app/models/environment_variable.rb` could do something like:
 - match all used variables
 - sort variables by how often they are used
 - do the resolution

/cc @zendesk/bre

grosser:
the bug might be that the supported notation is `${FOO}` and not `$(FOO)`

thrawny:
[...]

PR #3116, opened 2019-01-10
Sort environment variables based on kubernetes style referencing

esselius:
Fixes https://github.com/zendesk/samson/issues/3108

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -222,3 +222,11 @@ this is useful when dealing with large deployments that have some random failure
 To debug services or to create resources that needs to reference a selector that doesn't include team/role (like a Gateway), you can disable selector validation with:
 
 `metadata.annotations.samson/service_selector_across_roles: "true"`
+
+### Ordering environment variables based on references
+
+In kubernetes, you can reference an environment variable from another using the syntax: `FOO: $(BAR)`. However, in this case `BAR` needs to be defined before `FOO` for it to work. This is due to an implementation detail in kubernetes itself.
+
+Set `KUBERNETES_SORT_ENV_BY_REFERENCE=true` to make samson ensure the proper order when generating kubernetes manifests.
+
+Beware that cyclic references (setting `FOO: $(BAR)` and `BAR: $(FOO)`) will cause an error like this: `Could not sort environment variables, topological sort failed: ["BAR", "FOO"]`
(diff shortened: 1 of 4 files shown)


esselius:
@grosser Does this seem appropriate? :)

plugins/kubernetes/README.md
```diff
@@ -222,3 +222,11 @@ this is useful when dealing with large deployments that have some random failure
 To debug services or to create resources that needs to reference a selector that doesn't include team/role (like a Gateway), you can disable selector validation with:
 
 `metadata.annotations.samson/service_selector_across_roles: "true"`
+
+### Ordering environment variables based on references
+
+In kubernetes, you can reference an environment variable from another using the syntax: `FOO: $(BAR)`. However, in this case `BAR` needs to be defined before `FOO` for it to work. This is due to an implementation detail in kubernetes itself.
```
grosser:
```suggestion
In kubernetes, you can reference an environment variable from another using the syntax: `FOO: ${BAR}`. However, in this case `BAR` needs to be defined before `FOO` for it to work. This is due to an implementation detail in kubernetes itself.
```

grosser: ... maybe I understood your proposal wrong ... 
samson does env variable replacement too with `${FOO}` which works for non-kubernetes projects too

esselius: Basically, what we want is to be able to reference `$DEPLOY_GROUP`, which is not available when the `env`-plugin runs it's `resolve_dollar_variables()`-method in `EnvironmentVariable`, triggered by the `Samson::Hook` in the `kubernetes`-plugin.

We decided to make this an optional feature to please kubernetes, enabled by setting a environment variable, rather than refactoring how the `kubernetes` and `env` (and possibly other) plugins pass around env vars through hooks without knowing if that was a appropriate design decision.

What do you think @grosser ?

grosser: I think the problem is that resolve_dollar_variables (and preview / resolve-secrets) belongs into base and not into the env plugins since we should only resolve once we have collected all env vars from base+plugins.

A solution would be to add a new `Samson::EnvResolver` class that collects env (can be given a base env too) and handles preview/merge/resolve/secrets ... then use that from template_filler and from plugins/env/app/controllers/env/environment_controller.rb.

... with that any other plugin can add env vars via deploy_group_env hook and they automatically get resolved/previewed

... sounds good ?

(base make a fresh PR for that so this here does not get too cluttered)

grosser:
I'd do the refactor first and then the sorting after so it's cleaner :)

PR #3079, opened 2018-12-04
[BRE-1125] Support Hashicorp Vault versioned key / value store

ragurney:
Refactors the secret logic in Samson to support Hashicorp's versioned Key / Value store (KV Secrets Engine - Version 2).
 
#### Notable details about this implementation:
1. This is currently using a [forked version](https://github.com/zendesk/vault-ruby) of [vault-ruby](https://github.com/hashicorp/vault-ruby) containing changes introduced by [this PR](https://github.com/hashicorp/vault-ruby/pull/188). Once that pull request has been merged, the Gemfile should be updated to point to the original Gem again.
2. To support the new API, two new constructs -- `logical(:versioned)` and `logical(:unversioned)` -- were introduced which have different method signatures out of necessity but require the same data. To have one common API that all code could use within Samson to interact with the Vault client, this PR introduces wrappers for both the client and logicals which cleans up the code and reduces a bunch of branching logic. This also gives us the benefit of having more flexibility in the future for adding custom logic / refactoring the existing behavior as needed. 

#### Changes to the UI:
screen shot 2018-12-04 at 11 19 12 am

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team
 - [ ] Thoroughly test on staging

### References
 - Jira link: [BRE-1125](https://zendesk.atlassian.net/browse/BRE-1125)

### Risks
- Level: High - this should be thoroughly tested in staging to ensure secrets are functioning normally, as this is core Samson logic for deployment.

PR #3117, opened 2019-01-10
[BRE-1125] Support Hashicorp Vault versioned key / value store

ragurney:
This is an updated version of the original implementation to support KV v2 (#3079).

Refactors the secret logic in Samson to support Hashicorp's versioned Key / Value store (KV Secrets Engine - Version 2).
 
#### Notable details about this implementation:
1. This is currently using a [forked version](https://github.com/zendesk/vault-ruby) of [vault-ruby](https://github.com/hashicorp/vault-ruby) containing changes introduced by [this PR](https://github.com/hashicorp/vault-ruby/pull/194). Once that pull request has been merged, the Gemfile should be updated to point to the original Gem again.
2. To support the new API, the vault-ruby PR introduced a new module `KeyValue`. In Samson we must interface with servers using either KV v1 or KV v2. To reduce the branching logic in Samson's code, I've created some wrappers around the client so that we can have a unified interface in Samson to call Vault methods. Eventually this could be refactored to something like what Pat suggests [here](https://github.com/hashicorp/vault-ruby/pull/194#issuecomment-448448359), which would allow us to get rid of the wrappers. 

#### Changes to the UI:
screen shot 2018-12-04 at 11 19 12 am

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team
 - [ ] Thoroughly test on staging

### References
 - Jira link: [BRE-1125](https://zendesk.atlassian.net/browse/BRE-1125)

### Risks
- Level: High - this should be thoroughly tested in staging to ensure secrets are functioning normally, as this is core Samson logic for deployment.


lib/samson/secrets/vault_client_manager.rb
```diff
@@ -3,30 +3,29 @@
...
-        with_retries { vault.logical.read(wrap_id(id)) }
+        vault_client = responsible_clients(id).first
+        with_retries { vault_client.kv.read(id) }
       end
 
       # different servers have different ids so combine all
-      def list_recursive(path)
-        path = wrap_id(path)
-        all = Samson::Parallelizer.map(clients.values) do |vault|
+      def list_recursive
```
grosser: I liked the idea of this being reusable with any prefix path

grosser (approved):
looks about right and is not too ugly, good ob 👍

config/initializers/periodical.rb
```diff
@@ -10,7 +10,7 @@
 end
 
 Samson::Periodical.register :renew_vault_token, "Renew vault token" do
-  Samson::Secrets::VaultClient.client.renew_token
+  Samson::Secrets::VaultClientManager.instance.renew_token
```
grosser: making these methods static might also work out

lib/samson/secrets/vault_client_wrapper.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'vault'
+
+module Samson
+  module Secrets
+    class VaultClientWrapper < Vault::Client
+      # TODO: make these configurable via env vars and send env vars to samson_secret_puller
+      MOUNT = 'secret'
+      PREFIX = 'apps'
```
grosser: I'd rather have the manager own the mount + prefix ... this thing here is more of a crutch and will be deleted eventually

lib/samson/secrets/vault_kv_wrapper.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+    end
+    keys.each { |k| k.slice!(0, path.size) } if root
+    keys
+  end
+
+  # Limit scope to logical after subclass call
+  alias_method :original_list, :list
+  alias_method :original_list_recursive, :list_recursive
+end
```
grosser: are these 2 list_recursive versions different ... if not, a module to prepend in both would be nice

grosser: ... and if not, call out the diff in the docs

ragurney: I'll extract it to a shared module 👍

lib/samson/secrets/vault_client_manager.rb
ragurney: It seemed like we really don't have a use for that right now though in Samson, so I thought it would be clearer to have the signature without the path. It would be easy to change later if we ever need to support other prefixes.

grosser:
having a generic patch we can share would be kinda nice and it gives us a
method we can use from console to do random listings

lib/samson/secrets/shared/list_recursive.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
```
grosser: `shared` folder seems kinda overkill 🤷‍♂️

lib/samson/secrets/shared/list_recursive.rb
```diff
@@ -0,0 +1,30 @@
...
+
+# Add recursive listing which was rejected by vault-ruby see https://github.com/hashicorp/vault-ruby/pull/118
+# using a monkey patch so `vault_action :list_recursive` works in the backend
+module Samson
+  module Secrets
+    module Shared
+      module ListRecursive
+        # Limit scope to logical after subclass call
```
grosser: idk why this is needed, could you explain more or try to remove ?

lib/samson/secrets/vault_client_manager.rb
```diff
@@ -3,30 +3,33 @@
...
-        with_retries { vault.logical.read(wrap_id(id)) }
+        vault_client = responsible_clients(id).first
+        with_retries { vault_client.kv.read(id) }
       end
 
       # different servers have different ids so combine all
-      def list_recursive(path)
-        path = wrap_id(path)
-        all = Samson::Parallelizer.map(clients.values) do |vault|
+      def list_recursive(path = '')
```
grosser: signature for list has a require path, so should have that here too for symmetry ?
https://github.com/hashicorp/vault-ruby/blob/cfc499b00e610804cbde9bebe64beb23296541d1/lib/vault/client.rb#L189

test/lib/samson/secrets/shared/list_recursive_test.rb
```diff
@@ -0,0 +1,26 @@
...
+
+SingleCov.covered!
+
+describe Samson::Secrets::Shared::ListRecursive do
+  class DummyVaultClient
+    def list(_path)
+    end
+  end
+
+  describe 'client' do
```
grosser:
```suggestion
  describe '#list_recursive' do
```

test/lib/samson/secrets/vault_client_wrapper_test.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      client = Samson::Secrets::VaultClientWrapper.new(client_args)
+
+      client.instance_variable_get(:@ssl_verify).must_equal true
+      client.instance_variable_get(:@timeout).must_equal 543
+      client.instance_variable_get(:@ssl_timeout).must_equal 345
+    end
+  end
+
+  describe 'versioned' do
```
grosser: should be inside `describe "#initialize"`

test/lib/samson/secrets/vault_client_wrapper_test.rb
```diff
@@ -0,0 +1,40 @@
...
+    let(:client) { Samson::Secrets::VaultClientWrapper.new(versioned_kv: false).kv }
+
+    it 'instantiates an unversioned logical wrapper if server is unversioned' do
+      client.must_be_instance_of Samson::Secrets::VaultLogicalWrapper
+    end
+  end
+
+  describe 'unversioned' do
```
grosser: same

lib/samson/secrets/shared/list_recursive.rb
ragurney: It's needed because the wrappers are inheriting from the classes they are wrapping, so for example `VaultLogicalWrapper` inherits from `Vault::Logical`. `VaultLogicalWrapper` has `list` and `list_recursive` defined to provide the unified interface for calling vault functions in Samson code. When we call `VaultLogicalWrapper#list_recursive`, that will call `super()`, at which point that method will call `list_recursive` and `list` multiple times as it reads secrets. We want these calls to be `Logical`'s methods, not the wrappers, so I use the aliasing to limit scope to the parent once we enter that method.

lib/samson/secrets/shared/list_recursive.rb
ragurney: I think it's nice to have, and makes it obvious what lives there.

lib/samson/secrets/shared/list_recursive.rb
grosser: add something like that as inline comment plz :)

lib/samson/secrets/vault_client_manager.rb
ragurney: Is that necessary for us though? In the context of Samson, we'll always be reading from something starting with `secret/apps`, so it made sense to me to have that as a default to save time / be more dry.

grosser: nah, not necessary, was just thinking "if this wad part of the gem it should look like this"

lib/samson/secrets/shared/list_recursive.rb
ragurney: Will do 👍

ragurney:
Tested using the new Secret Service on Staging and it looks like it's working 🎉

ragurney:
@grosser could I get another 👀? There are failures on Travis but they aren't related to this PR. Seems like there are some flaky tests.

grosser (approved):

PR #3120, opened 2019-01-12
scan hardcoded images for vulnerabilities

eatwithforks:
might break projects that use vulnerability scanner but have non-gcr sidecars/other containers, those can disable the flag on the stage to make their deploys work while they move everything to gcr.

flag can be disabled on the stage edit page.
ie. projects/truth_service/stages/master/edit
screen shot 2019-01-14 at 9 26 08 am

@zendesk/compute @zendesk/bre

### DIFF
--- a/plugins/gcloud/test/samson_gcloud/samson_plugin_test.rb
+++ b/plugins/gcloud/test/samson_gcloud/samson_plugin_test.rb
@@ -111,6 +111,43 @@ describe SamsonGcloud do
         end
       end
     end
+
+    describe "gcloud vulnerabilty scanning" do
+      with_env GCLOUD_IMAGE_SCANNER: "true", GCLOUD_ACCOUNT: 'acc', GCLOUD_PROJECT: 'example'
+
+      def fire
+        Samson::Hooks.fire(:ensure_docker_image_has_no_vulnerabilities, stage, image)
+      end
+
+      let(:image) { +'foo.com/example/bar' }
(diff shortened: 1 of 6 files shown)


hgokavarapuz:
Can please update in the PR how to do it, it would be helpful.

grosser (approved):

plugins/gcloud/test/samson_gcloud/samson_plugin_test.rb
```diff
@@ -111,6 +111,43 @@ def expects_sleep(times)
...
     end
+
+    describe "gcloud vulnerabilty scanning" do
+      with_env GCLOUD_IMAGE_SCANNER: "true", GCLOUD_ACCOUNT: 'acc', GCLOUD_PROJECT: 'example'
+
+      def fire
+        Samson::Hooks.fire(:ensure_docker_image_has_no_vulnerabilities, stage, image)
+      end
+
+      let(:image) { +'foo.com/example/bar' }
```
hgokavarapuz: what is that "+" here ? Is that intended ?

hgokavarapuz (approved):
small nits i guess, other than that looks fine to me.

plugins/gcloud/test/samson_gcloud/samson_plugin_test.rb
eatwithforks: to make variable nonfrozen. because `# frozen_string_literal: true` is at top of file.

PR #3122, opened 2019-01-16
pass mount/prefix/kv-v2 to secert-puller to reduce hardcoding and pre…

grosser:
…pare for kv v2

@zendesk/compute @zendesk/bre @ragurney


adammw (approved):

ragurney (approved):
:shipit:

PR #3123, opened 2019-01-16
fix: samson/prerequisites have to be consistent for all resources

eatwithforks:
@zendesk/compute @zendesk/bre

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -31,7 +31,10 @@ module Kubernetes
       validate_not_matching_team
       validate_stateful_set_service_consistent
       validate_stateful_set_restart_policy
-      validate_annotations || validate_prerequisites
+      unless validate_annotations
+        validate_prerequisites_kinds
+        validate_prerequisites_consistency
+      end
       validate_env_values
       validate_host_volume_paths
       @errors.presence
@@ -226,14 +229,21 @@ module Kubernetes
       @errors << "Env values #{bad.join(', ')} must be strings." if bad.any?
     end
 
-    def validate_prerequisites
+    # samson waits for prerequisites to finish, so only resources that complete can be prerequisites
+    def validate_prerequisites_kinds
       allowed = ["Job", "Pod"]
-      bad = (@elements + templates).select do |e|
-        e.dig(*RoleConfigFile::PREREQUISITE) && !allowed.include?(e[:kind]) && RoleConfigFile.primary?(e)
-      end
-      bad.compact!
-      bad.map! { |e| e[:kind] }
-      @errors << "Elements with type #{bad.join(", ")} cannot be prerequisites." if bad.any?
+      return if map_attributes(RoleConfigFile::PREREQUISITE).compact.empty?
+      return if (map_attributes([:kind]) & allowed).any? || templates.empty?
+
+      @errors << "Prerequisites only support #{allowed.join(', ')}"
+    end
+
+    # only a whole role is supported to be a prerequisites, so prerequisites flag needs to be consistent
+    def validate_prerequisites_consistency
+      used = map_attributes(RoleConfigFile::PREREQUISITE).compact
+      return if used.empty? || used.size == @elements.size && used.uniq.size <= 1
+
+      @errors << "Prerequisite annotation must be used consistently across all resources of each role"
     end
 
     # comparing all directories with trailing / so we can use simple matching logic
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -31,7 +31,7 @@ def validate
       validate_not_matching_team
       validate_stateful_set_service_consistent
       validate_stateful_set_restart_policy
-      validate_annotations || validate_prerequisites
+      validate_annotations || validate_prerequisites_kinds || validate_prerequisites_consistency
```
grosser: the || hack is here since we cannot run prerequisites checks when annotations are broken, so `||` chaining them makes no sense

```suggestion
      if validate_annotations
        validate_prerequisites_kinds
        validate_prerequisites_consistency
      end
```

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -226,16 +226,23 @@ def validate_env_values
       @errors << "Env values #{bad.join(', ')} must be strings." if bad.any?
     end
 
-    def validate_prerequisites
+    # prerequisite means that we deploy it before we deploy other things, this is only supported for deployable
```
grosser:
```suggestion
    # when prerequisites are deploy, samsons waits for it to finish, so resources that complete can be prerequisites
```

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -226,16 +226,23 @@ def validate_env_values
...
-      end
-      bad.compact!
-      bad.map! { |e| e[:kind] }
+      bad =
+        @elements.
+          select { |e| e.dig(*RoleConfigFile::PREREQUISITE) && !allowed.include?(e[:kind]) }.map { |e| e[:kind] }
       @errors << "Elements with type #{bad.join(", ")} cannot be prerequisites." if bad.any?
     end
 
+    # resources that finish like jobs and pods. It can only be set consistently across a single role.
```
grosser:
```suggestion
    # only a whole role is supported to be a prerequisites, so prerequisites flag needs to be consistent
```

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -226,16 +226,23 @@ def validate_env_values
...
+    def validate_prerequisites_consistency
+      used = map_attributes(RoleConfigFile::PREREQUISITE)
+      if (!used.empty? && used.size != @elements.size) || used.uniq.size > 1
```
grosser:
```suggestion
      if (used.any? && used.size != @elements.size) || used.uniq.size > 1
```

plugins/kubernetes/test/models/kubernetes/role_validator_test.rb
```diff
@@ -332,10 +332,36 @@
...
       end
     end
 
+    describe "#validate_prerequisites_consistency" do
+      before do
+        role.first[:kind] = "Job"
+        role.first[:metadata][:annotations] = {"samson/prerequisite": "true"}
+        role.first[:spec][:template][:spec][:restartPolicy] = "Never"
+
+        role.last[:kind] = "Pod"
```
grosser: having 2 deployables is kinda weird, best make the second one a ConfigMap or so ... but then that would break the other validation ... so something is not correct here :/

... so logic needs to be:
 - prerequisite needs to be consistent
 - at least one of the prerequisites must be Job/Pod (ideally not anything else deployable, but that's tricky, so let's not do that)

grosser: also missing test for the `size == 0` condition

eatwithforks:
done

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -226,14 +229,21 @@ def validate_env_values
...
+      return if map_attributes(RoleConfigFile::PREREQUISITE).compact.empty?
+      return if (map_attributes([:kind]) & allowed).any? || templates.empty?
+
+      @errors << "Prerequisites only support #{allowed.join(', ')}"
+    end
+
+    # only a whole role is supported to be a prerequisites, so prerequisites flag needs to be consistent
+    def validate_prerequisites_consistency
+      used = map_attributes(RoleConfigFile::PREREQUISITE).compact
+      return if (used.empty? || used.size == @elements.size) && used.uniq.size <= 1
```
grosser: same effect, but easier to read:
```suggestion
      return if used.empty? || (used.size == @elements.size && used.uniq.size == 1)
```

grosser (approved):

PR #3126, opened 2019-01-17
make output-buffer aggregate itself to remove invalid to_s and get ri…

grosser:
…d of a class

old .to_s should never have been shown to any user

/cc @zendesk/bre @zendesk/compute


grosser:
wrong approach

PR #3131, opened 2019-01-18
do not split dots in labels and annotations so they get replaved dire…

### LINKED ISSUE #3127: Evaluating environment variables in k8s manifests
After having set up samson to deploy containers to kubernetes, using externally build images and enabling the appropriate configuration flags, I am very happy about how things work so far.

We've run into a slight issue with environment variables not being evaluated and substituted in the manifests, which we'd like to have, for our load balancer to have desired endpoints defined on a per environment basis, so it can fetch certs from lets encrypt.

The manifest looks like this:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: hello-kube
  labels:
    app: hello-kube
    project: hello-kube
    role: hello-kube
  annotations:
    kubernetes.io/ingress.class: traefik
spec:
  rules:
  - host: $DOMAIN_POSTFIX
    http:
      paths:
      - backend:
          serviceName: hello-kube
          servicePort: 80
```

this is only a part, see the full yaml in [this gist](https://gist.github.com/nayena/bdd7eb6b58e46b05ab4676ecb2419e72)

For testing purposes, the `$DOMAIN_POSTFIX` variable is alone, but i'd like to use it as a postfix later, so it would be something along: `host: hello.$DOMAIN_POSTFIX`.

When deploying our role, this is the error line, indicating that the variable isn't evaluted:
```
[10:18:24] Deploying Stage role hello-kube
[10:18:24] JobExecution failed: Kubernetes error hello-kube default Stage: Ingress.extensions "hello-kube" is invalid: spec.rules[0].host: Invalid value: "$DOMAIN_POSTFIX": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```
As you can see, the environment tab shows the variable as present, and it is configured in a deploy-group.
environment

Looking at the group previewer (at /environment_variable_groups/preview?project_id=3), I see the variable as well:

preview

and it is also selected in the settings for the project:
project settings

Are we doing it wrong or is this not an option in samson?

grosser:
…ctly

fixes #3127

@zendesk/compute @nayena

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
@@ -881,6 +881,20 @@ describe Kubernetes::TemplateFiller do
         template.to_hash[:spec][:foo].must_equal "bar"
       end
 
+      it "sets annotations with ." do
+        raw_template[:metadata][:annotations] = {
+          "samson/set_via_env_json-metadata.annotations.foo.bar/baz" => "FOO"
+        }
+        template.to_hash[:metadata][:annotations][:"foo.bar/baz"].must_equal "bar"
+      end
+
+      it "sets labels with ." do
+        raw_template[:metadata][:annotations] = {
+          "samson/set_via_env_json-metadata.labels.foo.bar/baz" => "FOO"
+        }
+        template.to_hash[:metadata][:labels][:"foo.bar/baz"].must_equal "bar"
+      end
+
       it "sets in arrays" do
         raw_template[:metadata][:annotations] = {
           "samson/set_via_env_json-spec.template.spec.containers.0.foo" => "FOO"
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -78,10 +78,12 @@ module Kubernetes
     private
 
     def set_via_env_json
-      (template[:metadata][:annotations] || {}).each do |k, v|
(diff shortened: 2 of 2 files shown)


plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -881,6 +881,20 @@ def secret_annotations(hash)
...
 
+      it "sets annotations with ." do
+        raw_template[:metadata][:annotations] = {
+          "samson/set_via_env_json-metadata.annotations.foo.bar/baz" => "FOO"
+        }
+        template.to_hash[:metadata][:annotations][:"foo.bar/baz"].must_equal "bar"
+      end
+
+      it "sets labels with ." do
+        raw_template[:metadata][:annotations] = {
```
hgokavarapuz: shouldn't this be under labels instead of annotations?

hgokavarapuz (approved):
small nit, other than that looks fine to me.

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
grosser: the replacement always gets set as an annotation, but it can replace labels

PR #3135, opened 2019-01-18
hide details link that only matters for zendesk

### LINKED ISSUE #3128: Documentation for external building is internal for Zendesk
The link for the guide to set up external builds is pointing to an internal zendesk document.

the link

The link in question can be found at [/app/views/projects/_docker_fields.html.erb#L10](https://github.com/zendesk/samson/blob/master/app/views/projects/_docker_fields.html.erb#L10)

Is there any possibility to extract something for the open source community to read?

grosser:
fixes #3128

@sathishavm @nayena

/cc @zendesk/bre


nayena:
Are there plans to document how the external building process ideally would be configured? Right now I just hacked something together with cURL

grosser:
our process is having a bot that sends all GCR build results to samson

nayena:
@grosser Yeah, I understood that from reading the code, but it would be nice to have examples of those calls (and how the relationship between docker-sha, git-sha, git-ref and such) in the documentation, for others in the future, so they wouldn't have to dig deep into the code :)

grosser:
how about a docs PR ? :)
... glad the code was at least understandable :D

PR #3138, opened 2019-01-22
use a more ideomatic way of setting namespaces

grosser:
declaring things as cluster-service is misleading and not obvious

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -192,7 +192,8 @@ to make all kubernetes deploys that do not use a `metadata.labels.team` / `spec.
 ### Using custom namespace
 
 Samson overrides each resources namespace with to the deploygroups `kubernetes_namespace`.
-To make samson not override the namespace, set `metadata.kubernetes.io/cluster-service: 'true'`
+To make samson not override the namespace, set `metadata.annotations.samson/keep_namespace: 'true'` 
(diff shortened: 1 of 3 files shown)


adammw:
Is now a good time to rethink how samson does namespaces completely rather than another stop-gap? Since if we will want to be able to do namespaces per project, this will all change...

adammw (approved):

grosser:
another interesting way would be to say "if you set a namespace, we will not override it"
that will need a bit of cleanup, but could be neat

grosser:
for now simple stop-gaps/fixes ... per-project namespaces will be complicated enough / let's not rush it

plugins/kubernetes/README.md
```diff
@@ -192,7 +192,8 @@ to make all kubernetes deploys that do not use a `metadata.labels.team` / `spec.
 ### Using custom namespace
 
 Samson overrides each resources namespace with to the deploygroups `kubernetes_namespace`.
-To make samson not override the namespace, set `metadata.kubernetes.io/cluster-service: 'true'`
+To make samson not override the namespace, set `metadata.annotations.samson/keep_namespace: 'true'` 
+(or `metadata.kubernetes.io/cluster-service: 'true'`)
```
adammw: missing labels here in the path...

grosser: https://github.com/zendesk/samson/pull/3141

PR #3139, opened 2019-01-23
review stages before doing a mass-deploy

grosser:
screen shot 2019-01-22 at 5 21 43 pm
screen shot 2019-01-22 at 5 21 33 pm

@zendesk/compute @dragonfax 
/cc @aglover-zendesk this will be the UI to do the mass-deploys for secrets update

### DIFF
--- a/app/controllers/mass_rollouts_controller.rb
+++ b/app/controllers/mass_rollouts_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 class MassRolloutsController < ApplicationController
-  before_action :authorize_super_admin!
+  before_action :authorize_deployer!, except: [:new, :create, :merge, :destroy]
+  before_action :authorize_super_admin!, except: [:review_deploy, :deploy]
   before_action :deploy_group
 
   def new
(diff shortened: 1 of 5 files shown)


aglover-zendesk:
@grosser one point of confusion - since the old Vault servers will still be around, and the secret_puller will still be valid for those Vault servers, do we need to do a mass rollout to update the secret_puller? Or is the secret_puller update a breaking change, meaning the last version of secret_puller will no longer work?

app/views/mass_rollouts/review_deploy.html.erb
```diff
@@ -0,0 +1,38 @@
...
+    <small>Selected stages will be deployed.</small>
+    <section class="clearfix">
+      <div class="table table-hover table-condensed">
+        <table class="table">
+          <thead>
+          <tr>
+            <th></th>
+            <th>Project</th>
+            <th>Stage</th>
+          </tr>
```
adammw: Ideally would be good to show the reference that would be deployed here. (maybe another TODO)

config/routes.rb
```diff
@@ -89,6 +89,7 @@
     resource :mass_rollouts, only: [:new, :create, :destroy] do
       collection do
         post :merge
+        get :review_deploy
```
adammw: little unusual this is a get, but I guess it makes sense since there's no side-effects.

test/controllers/mass_rollouts_controller_test.rb
```diff
@@ -59,193 +59,113 @@ def create_deploy(stage, reference, status)
...
       end
 
-      let(:status) { nil }
-
-      it "deploys all" do
-        assert_difference('Deploy.count', 1) do
-          post :deploy, params: {deploy_group_id: deploy_groups(:pod100)}
-          assert_response :redirect
-        end
+      it "can drive list from console by building a url with needed ids" do
```
adammw: i have no idea what that test description means.. how about just deploys with list of stage ids.

app/controllers/mass_rollouts_controller.rb
```diff
@@ -9,37 +9,48 @@ def new
...
+      end
     end
+  end
 
+  # TODO: show when deploys are waiting for buddy
+  # TODO: show when deploys were invalid
+  # TODO: show when stage had no reference
+  def deploy
+    stages_to_deploy = Stage.find(params.require(:stage_ids))
     deploy_references = stages_to_deploy.map do |stage|
```
adammw: Would be nice to do this in the review stage as well for the UI, show the reference and show a list of "unable to deploy" projects as well.

app/controllers/mass_rollouts_controller.rb
```diff
@@ -9,37 +9,48 @@ def new
...
+        @stages.select!(&:kubernetes?)
+      when 'false'
+        @stages.reject!(&:kubernetes?)
+      else
+        return unsupported_option(:kubernetes)
+      end
     end
+  end
 
+  # TODO: show when deploys are waiting for buddy
```
adammw: How did this ever work with buddy requests? Did someone have to go and manually buddy each deploy one by one?

grosser: yep :D

app/controllers/mass_rollouts_controller.rb
grosser: yeah, I was a bit scared of this one since if you keep the page open too long you end up deploying a old release ... so this could only be a preview or so

test/controllers/mass_rollouts_controller_test.rb
grosser: I'm trying to explain the usecase, which is generating list of stage-ids from a console sessions (for secrets migration)

app/controllers/mass_rollouts_controller.rb
adammw: which projects/stages get deployed will only be a preview too.. alternatively you could send the references with the POST request..

app/views/mass_rollouts/review_deploy.html.erb
grosser: yeah, todo

grosser:
@aglover-zendesk new secret puller is already rolling out atm, that's why we need the mass-re-deploy, once everybody has it we can just switch the secret-puller-vault-data and it will talk to new vault servers

adammw:
Global deployer or project-level deployer? Is there possibility of global deployer and project-level viewer?

grosser:
global deployer for now, can be more fine-grained later if we need it

PR #3140, opened 2019-01-23
allow selecting what to deploy during mass-rollout

grosser:
screen shot 2019-01-23 at 11 06 14 am

should make this UI a little easier to understand and adds the option to just re-deploy which helps if projects don't have template stages

@dragonfax @zendesk/compute

### DIFF
--- a/app/views/mass_rollouts/review_deploy.html.erb
+++ b/app/views/mass_rollouts/review_deploy.html.erb
@@ -28,8 +28,18 @@
         </table>
       </div>
 
-      <div class="admin-actions">
-        <div class="pull-right">
+      <div class="form-group">
+        <div class="col-lg-2">
+          <label>Deploy reference from</label>
+        </div>
+        <%= select_tag :reference_source, options_for_select([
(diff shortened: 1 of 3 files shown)


test/controllers/mass_rollouts_controller_test.rb
```diff
@@ -110,6 +114,14 @@ def deploy
         deploy.reference.must_equal template_deploy.reference
       end
 
+      it "re-deploys last successful deploy" do
```
adammw: .. for that stage?

app/views/mass_rollouts/review_deploy.html.erb
```diff
@@ -28,6 +28,16 @@
         </table>
       </div>
 
+      <div class="col-lg-4">
+        <label>Deploy reference from:</label>
+        <%= select_tag :reference_source, options_for_select([
```
adammw: apply some bootstrap classes to make it look nice?

adammw: I'd also position this right aligned closer to the actual button, but either/or.

PR #3142, opened 2019-01-24
SAMSON-393 bump template limit so we can store endless resources

grosser:
@zendesk/bre @zendesk/compute

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_01_11_195416) do
+ActiveRecord::Schema.define(version: 2019_01_24_171921) do
 
   create_table "audits" do |t|
     t.integer "auditable_id", null: false
@@ -235,7 +235,7 @@ ActiveRecord::Schema.define(version: 2019_01_11_195416) do
     t.integer "deploy_group_id"
     t.decimal "limits_cpu", precision: 6, scale: 2, null: false
     t.integer "limits_memory", null: false
-    t.text "resource_template"
+    t.text "resource_template", limit: 1073741823
     t.decimal "requests_cpu", precision: 6, scale: 2, null: false
     t.integer "requests_memory", null: false
     t.boolean "delete_resource", default: false, null: false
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -584,8 +584,8 @@
     t.text "ca_cert"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["name"], name: "index_vault_servers_on_name", unique: true, length: 191
     t.boolean "versioned_kv", default: false, null: false
+    t.index ["name"], name: "index_vault_servers_on_name", unique: true, length: 191
```
grosser: fallout from earlier merge conflict, indexes always go last

ragurney (approved):

plugins/kubernetes/db/migrate/20190124171921_bump_limit.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class BumpLimit < ActiveRecord::Migration[5.2]
+  def change
+    change_column :kubernetes_release_docs, :resource_template, :text, limit: 1073741823
```
adammw: what is this magic number, and why doesn't it match the number in the schema.rb?

grosser: max mysql column length

db/schema.rb
```diff
@@ -235,7 +235,7 @@
     t.integer "deploy_group_id"
     t.decimal "limits_cpu", precision: 6, scale: 2, null: false
     t.integer "limits_memory", null: false
-    t.text "resource_template"
+    t.text "resource_template", limit: 4294967295
```
grosser: that's a bug and will be caught by cleanliness test :)

plugins/kubernetes/db/migrate/20190124171921_bump_limit.rb
adammw: comment / use a constant?

yizhang-zen: Same question on the numbers, is it safe to set "unlimited" limits?

grosser: 🤷‍♂️ not super safe, but we have been doing it for job output all these years and nothing bad happened

grosser: comment added

craig-day (approved):

PR #3144, opened 2019-01-24
allow deleting a resource to be able to change labels

grosser:
error message says it should be deleted, but then it does not allow deleting 😖 

```
[23:05:06] JobExecution failed: Updating spec.selector.matchLabels from {:project=>"rollup", :role=>"runner"} to {:project=>"rollup", :role=>"resque-runner"} can only be done by deleting and redeploying or old pods would not be deleted.
```

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/resource_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/resource_test.rb
@@ -152,11 +152,16 @@ describe Kubernetes::Resource do
       end
 
       describe "updating matchLabels" do
+        def assert_old(&block)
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -152,11 +152,16 @@ def assert_create_and_delete_requests(**args, &block)
       end
 
       describe "updating matchLabels" do
+        def assert_old(&block)
```
adammw: shouldn't it be assert_match_labels ?

adammw (approved):

grosser:
yeah this makes no sense ... code never is reached when deleting ... damn single_cov ruining my party :/

PR #3145, opened 2019-01-24
Render commands as json

danihodovic:
Allow the commands controller to render JSON, similar to the projects controller.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low

### DIFF
--- a/app/controllers/commands_controller.rb
+++ b/app/controllers/commands_controller.rb
@@ -26,10 +26,10 @@ class CommandsController < ApplicationController
       end
 
       format.json do
-        render json: {commands: @commands.as_json}
+        @pagy, @commands = pagy(@commands, page: page, items: 50)
+        render_as_json :commands, @commands
       end
     end
-
   end
 
   def new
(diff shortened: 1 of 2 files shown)


danihodovic:
Bare with me as this is the first time I write Ruby (outside of chef). I copied the logic from the projects controller (and the JSON tests) into the commands controller.

app/controllers/commands_controller.rb
```diff
@@ -20,7 +20,16 @@ def index
...
       end
     end
-    @pagy, @commands = pagy(@commands, page: page, items: 15)
+    respond_to do |format|
+      format.html do
+        @pagy, @commands = pagy(@commands, page: page, items: 15)
+      end
+
+      format.json do
+        render json: {commands: @commands.as_json}
```
grosser: - use pagination since this will blow up when someone has 10k commands :)
 - use `render_as_json :commands, @commands`

danihodovic:
https://github.com/zendesk/samson/pull/3153

PR #3146, opened 2019-01-24
show errors when mass-rolling stages

grosser:
screen shot 2019-01-24 at 2 07 02 pm

@dragonfax @adammw

### DIFF
--- a/app/controllers/mass_rollouts_controller.rb
+++ b/app/controllers/mass_rollouts_controller.rb
@@ -47,34 +47,41 @@ class MassRolloutsController < ApplicationController
     end
   end
 
-  # TODO: show when deploys are waiting for buddy
-  # TODO: show when deploys were invalid
-  # TODO: show when stage had no reference
   def deploy
-    stages_to_deploy = Stage.find(params.require(:stage_ids))
-    deploy_references = stages_to_deploy.map do |stage|
+    warnings = []
+
+    stages = Stage.find(params[:stage_ids] || [])
+    deploys = stages.map do |stage|
       reference =
         case params[:reference_source]
         when 'template' then last_successful_template_reference(stage)
         when 'redeploy' then stage.last_successful_deploy&.reference
-        else
-          return unsupported_option(:reference_source)
+        else return unsupported_option(:reference_source)
         end
 
-      [stage, reference] if reference
+      unless reference
+        warnings << [stage, "No reference found"]
+        next
+      end
+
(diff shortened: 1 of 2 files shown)


app/controllers/mass_rollouts_controller.rb
```diff
@@ -47,34 +47,41 @@ def review_deploy
...
         end
 
-      [stage, reference] if reference
+      unless reference
+        warnings << [stage, "No reference found"]
+        next
+      end
+
+      deploy = DeployService.new(current_user).deploy(stage, reference: reference)
+      if deploy.new_record?
```
adammw: why not `unless deploy.valid?`

app/controllers/mass_rollouts_controller.rb
```diff
@@ -47,34 +47,41 @@ def review_deploy
...
     end.compact
 
-    deploys = deploy_references.map do |stage, reference|
-      deploy_service = DeployService.new(current_user)
-      deploy_service.deploy(stage, reference: reference)
-    end
+    # prevent cookie overflow / keep the message short
+    ignored = warnings.slice!(10..-1)
+    warnings.map! { |stage, message| "#{message} #{stage.url}" }
+    warnings << "... #{ignored.size} more" if ignored
```
adammw: not really a fan of dropping this info... but can't easily see a way around it.

app/controllers/mass_rollouts_controller.rb
```diff
@@ -47,34 +47,41 @@ def review_deploy
     end
   end
 
-  # TODO: show when deploys are waiting for buddy
```
adammw: this is still a todo isn't it?

app/controllers/mass_rollouts_controller.rb
grosser: that would run validations again ... we could check `errors` but that would not  catch things callbacks throwing ... only alternative is `id` but that's even less readable

app/controllers/mass_rollouts_controller.rb
grosser: we could put it into the url, but that opens lots of xss holes 🤷‍♂️ 
... rendering is also not great since then reload would redeploy

app/controllers/mass_rollouts_controller.rb
grosser: decided to not do that since we have limited room and `no buddy` is visible in the deploy list we redirect to, so it's not that urgent

Issue #191, opened 2014-11-10
Get project configuration from a ".samson.yml"


xakraz:
It could be great to be able to _define the stages_ and associated commands to run from a YML files like travis and co...

``` yml
deploy:

  stages:
    demo:
      command:
      permalink:
      confirm:
      e-mail:
...
```

Sansom will 
- clone the project
- Get and parse the `.samson.yml`
- Identify and create the different stages / commands

grosser:
you could set up a rake task that deletes everything and sets it up by reading samson.yml

xakraz:
@grosser : Yes, or have it that way as every good product do : Descriptive and version controlled among the project source :)

grosser:
I was suggesting to check in the samson.yml and then have a rake task delete all existing commands and fill the commands from the yml file, would that work ?

xakraz:
Yes for sure, let's try that and make it evolve later if needed.

Do you want me to do that or some body of the samson Team will take it ?
Because I am kind a nooby in development, but I will be glad to have some directions and try :)

grosser:
You should give it a try :)

danihodovic:
I've started working on a project that allows you to build up your projects / stages / commands from yaml using Ansible. It's still a work in progress, but projects, stages and commands seem to work OK. Programming around the Samson API is tricky because it's not restful and not all endpoints support JSON. It might be useful to you if you're still using Samson.

https://github.com/danihodovic/ansible-samson

PR #3150, opened 2019-01-26
Provide restful JSON endpoints for projects

danihodovic:
I'm currently working on a Samson module for Ansible. The use case of the module is to streamline project, stage and command setup and specify the Samson object hierarchy as code. An example might explain this better:

```yaml
---
- name: Create Samson project
  samson_project:
    samson_url: https://mysamson.org.com
    token: xxxx
    state: present
    name: My project
    repository_url: https://github.com/zendesk/samson
    permalink: samson

- name: Create staging stage
  samson_stage:
    samson_url: https://mysamson.org.com
    token: xxxx
    state: present
    name: staging
    project_id: '{{ samson_project.project.id }}'
    production: false

- name: Remove the testing stage
  samson_stage:
    samson_url: https://mysamson.org.com
    token: xxxx
    state: absent
    premalink: testing
    project_id: '{{ samson_project.project.id }}'
```

Ansible uses the HTTP API to create these resources and the module is currently fairly complex because the Samson API is not restful. I rely on redirect status codes and parsing html for errors in order to know if a resource was created as intended.

Having a more restful-like approach to status codes and errors when he JSON endpoints are targeted simplifies the http client quite a lot.

There has been another attempt at creating Samson projects through Terraform which I think would also benefit from this PR - https://github.com/tolgaakyuz/samson-go.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -46,15 +46,30 @@ class ProjectsController < ApplicationController
   def create
     @project = Project.new(project_params)
     @project.current_user = current_user
+    saved = @project.save
 
-    if @project.save
+    if saved
       if Rails.application.config.samson.project_created_email
         ProjectMailer.created_email(@current_user, @project).deliver_now
       end
-      redirect_to @project
       Rails.logger.info("#{@current_user.name_and_email} created a new project #{@project.to_param}")
-    else
-      render :new
+    end
+
+    respond_to do |format|
+      format.html do
+        if saved
+          redirect_to @project
+        else
+          render :new
+        end
+      end
+      format.json do
+        if saved
+          render_as_json :project, @project
+        else
+          render_as_json :errors, @project.errors, status: :bad_request
+        end
+      end
     end
   end
 
@@ -74,10 +89,22 @@ class ProjectsController < ApplicationController
   end
 
   def update
-    if @project.update_attributes(project_params)
-      redirect_to @project
-    else
-      render :edit
+    updated = @project.update_attributes(project_params)
+    respond_to do |format|
+      format.html do
+        if updated
+          redirect_to @project
+        else
+          render :edit
+        end
+      end
+      format.json do
+        if updated
+          render_as_json :project, @project
+        else
+          render_as_json :errors, @project.errors, status: :bad_request
+        end
+      end
     end
   end
 
@@ -87,7 +114,15 @@ class ProjectsController < ApplicationController
     if Rails.application.config.samson.project_deleted_email
       ProjectMailer.deleted_email(@current_user, @project).deliver_now
     end
-    redirect_to projects_path, notice: "Project removed."
+
+    respond_to do |format|
+      format.html do
+        redirect_to projects_path, notice: "Project removed."
+      end
+      format.json do
+        render json: {message: "Project removed."}
+      end
+    end
   end
 
   def deploy_group_versions
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -27,7 +27,7 @@ class ProjectsController < ApplicationController
       end
 
       format.json do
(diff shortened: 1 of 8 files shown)


danihodovic:
https://github.com/zendesk/samson/issues/191 may be relevant

app/controllers/projects_controller.rb
```diff
@@ -46,15 +46,30 @@ def new
...
+          redirect_to @project
+        else
+          render :new
+        end
+      end
+      format.json do
+        if saved
+          render_as_json :project, @project
+        else
+          render_as_json :errors, @project.errors, status: :bad_request
```
grosser: should be a 422

app/controllers/projects_controller.rb
```diff
@@ -74,10 +89,22 @@ def edit
...
+          redirect_to @project
+        else
+          render :edit
+        end
+      end
+      format.json do
+        if updated
+          render_as_json :project, @project
+        else
+          render_as_json :errors, @project.errors, status: :bad_request
```
grosser: 422

app/controllers/projects_controller.rb
grosser: actually these should be `render_json_error 422, exception.record.errors` for consistency
also this is getting kinda repetitive, so some kind of method extraction would be great :)

app/controllers/projects_controller.rb
```diff
@@ -87,7 +114,15 @@ def destroy
...
       ProjectMailer.deleted_email(@current_user, @project).deliver_now
     end
-    redirect_to projects_path, notice: "Project removed."
+
+    respond_to do |format|
+      format.html do
+        redirect_to projects_path, notice: "Project removed."
+      end
+      format.json do
+        render json: {message: "Project removed."}
```
grosser: would prefer a simple `head :ok` here ... or something else generic

grosser:
this looks like the right direction overall, just need some helper methods to make this consistent so we can apply this everywhere ... I don't want to end up with different controllers behaving in inconsistent ways :)

danihodovic:
> this looks like the right direction overall, just need some helper methods to make this consistent so we can apply this everywhere ... I don't want to end up with different controllers behaving in inconsistent ways :)

I was thinking of refactoring / using helper methods once I make a pull request for stages and commands later on, WDYT?

grosser:
see locks_controller for an example
maybe a RestfulReplies concern that can handle these methods would be good, can be simple/stupid for now

grosser:
the helper methods would be pretty straight forward, so I'd love to see them in this PR to not merge a bunch of duplication 🤷‍♂️

grosser:
... flay is failing because of the duplication anyway :D

danihodovic:
OK sounds good. I'll see what I can do with my limited Ruby / Rails knowledge :)

Thank you for the quick reviews!

danihodovic:
>  maybe a RestfulReplies concern that can handle these methods would be good, can be simple/stupid for now

> the helper methods would be pretty straight forward, so I'd love to see them in this PR to not merge a bunch of duplication man_shrugging

I'm not sure what the helper methods or RestfulReplies concern would look like. If you would like it to look similar to the locks_controller.rb I could do that. If not, can you elaborate with a code snippet?

danihodovic:
I'll close this until it's ready for review

danihodovic:
@grosser I've added a helper method and provided JSON interfaces for projects, stages and commands. Let me know if this looks better.

grosser:
I like where this is going, so I'll steal this and improve it a bit, and make my own PR, hopefully done tomorrow, need a bit of cleanup merged first to rebase onto

danihodovic:
https://github.com/zendesk/samson/pull/3153

PR #3153, opened 2019-01-28
start unifying controller for consistent html/json/csv behavior

grosser:
as replacement for https://github.com/zendesk/samson/pull/3150 and as future basis for most of the controllers

TODO:
 - use it for more controllers (commands/stages/deploys)
 - add more test coverage


grosser:
@danihodovic

danihodovic:
:clap: :clap: 

What do you think about applying resource_controller.rb to stages and commands to see if it's as generic as it should be?

danihodovic:
Great :+1: . This will simplify building tooling around Samson since the API is now restful-ish. cc @elblivion @tolgaakyuz.

test/controllers/resource_controller_test.rb
```diff
@@ -0,0 +1,276 @@
...
+    json.keys.must_equal [:project]
+  end
+
+  def json
+    JSON.parse(response.body, symbolize_names: true)
+  end
+
+  let(:project) { projects(:test) }
+
+  as_a :admin do
```
bcolferzd: Do you need "as_a :user" or "as_not_a :admin" ?

dragonfax:
Its just a cosmetic test failure.

test/controllers/resource_controller_test.rb
grosser: need admin login here so all urls work like deleting commands etc

danihodovic:
Any plans on applying this for stages and deploy groups?

grosser:
all controller should get these, I'm doing them 1-by-1 when I have a bit of time to kill :D

PR #3154, opened 2019-01-28
clarity readme

grosser:
@adam-hart 
/cc @zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -132,11 +132,11 @@ Via [Template filler](/plugins/kubernetes/app/models/kubernetes/template_filler.
 
 ### Migrations / Prerequisite
 
-Add a role with only a `Pod`, `metadata.annotations.samson/prerequisite: 'true'`, and command to run a migrations.
-It will be executed before the rest is deployed.
-
-For default it waits for 10 minutes before timeout, you can change the timeout
-using KUBERNETES_WAIT_FOR_PREREQUISITES env variable (specified in seconds).
+Should be added as a separate role with the annotation  `samson/prerequisite: 'true'` set on the `Job`/`Deployment`/`Pod`
+(annotation should be added to the 'root' object, not the template).
+This role will be deployed/executed before any other role is deployed.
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/README.md
```diff
@@ -132,11 +132,11 @@ Via [Template filler](/plugins/kubernetes/app/models/kubernetes/template_filler.
 
 ### Migrations / Prerequisite
 
-Add a role with only a `Pod`, `metadata.annotations.samson/prerequisite: 'true'`, and command to run a migrations.
-It will be executed before the rest is deployed.
+Add a role with a `Pod` or `Job`, `metadata.annotations.samson/prerequisite: 'true'`, and `command` to run a migrations.
+It will be executed before any other role is deployed.
```
adammw: still not clear you need to add this to the job and not the pod template

grosser:
added "root" ... better ?

plugins/kubernetes/README.md
hgokavarapuz: same here...That description is still not clear.

plugins/kubernetes/README.md
```diff
@@ -132,11 +132,11 @@ Via [Template filler](/plugins/kubernetes/app/models/kubernetes/template_filler.
 
 ### Migrations / Prerequisite
 
-Add a role with only a `Pod`, `metadata.annotations.samson/prerequisite: 'true'`, and command to run a migrations.
-It will be executed before the rest is deployed.
+Add a role with a `Pod` or `Job`, root `metadata.annotations.samson/prerequisite: 'true'`, and command to run a migrations.
+It will be executed before any other role is deployed.
```
adammw: Suggestion:

> Migrations/prerequisites should be added as a separate role with the annotation  `samson/prerequisite: 'true'` set on your `Job`/`Deployment`/`Pod` (annotation should be added to the 'root' object, not the template). This role will be deployed/executed before any other role is deployed.

grosser: updated

adammw (approved):

PR #3155, opened 2019-01-29
use json for js commands edit to unify on json replies

grosser:
split out from my PR that uses generic json logic

@zendesk/compute


test/controllers/commands_controller_test.rb
```diff
@@ -136,7 +136,7 @@
         end
 
         it "cannot update invalid as js" do
```
adammw: json

PR #3158, opened 2019-01-30
PAAS-2771 do not blow up when cluster config is bad/unreadable

grosser:
allows fixing config or deleting the cluster

@zendesk/compute


plugins/kubernetes/app/views/kubernetes/clusters/edit.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= page_title(@cluster.new_record? ? "New Kubernetes Cluster" : "Edit Kubernetes Cluster") %>
-<% config_contexts = (@cluster.config_filepath ? @cluster.kubeconfig.contexts : []) %>
+<% config_contexts = (@cluster.config_filepath ? @cluster.kubeconfig.contexts : []) rescue [] %>
```
adammw: wow inline rescue is weird.

adammw (approved):

plugins/kubernetes/app/views/kubernetes/clusters/edit.html.erb
grosser: you spelled "amazing" wrong :)

PR #3159, opened 2019-01-30
make clusters controller reuse restful logic

grosser:
getting rid of more repetitive code and making things behave the same (+ free json api)

@zendesk/compute


plugins/kubernetes/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -1,79 +1,42 @@
...
-    else
-      render :edit
-    end
+    super
+    @kubernetes_cluster.config_filepath ||= new_config_filepath
   end
 
   def index
-    @clusters = ::Kubernetes::Cluster.all.sort_by { |c| Samson::NaturalOrder.convert(c.name) }
+    @kubernetes_clusters = ::Kubernetes::Cluster.all.sort_by { |c| Samson::NaturalOrder.convert(c.name) }
```
bcolferzd: I guess it is more descriptive to identify the cluster a kubernetes

plugins/kubernetes/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -1,79 +1,42 @@
...
-  def new_cluster_params
-    params.require(:kubernetes_cluster).permit(
+  def resource_params
+    super.permit(
       :name, :config_filepath, :config_context, :description, :ip_prefix, deploy_group_ids: []
     )
   end
 
-  def new_config_file_path
+  def new_config_filepath
```
bcolferzd: Interesting to remove the "_" in file path... okay

plugins/kubernetes/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -121,6 +84,6 @@ def secret_exist?(secret)
   end
 
   def secrets_client
-    @cluster.client('v1')
+    @kubernetes_cluster.client('v1')
```
bcolferzd: In the future should the client version be a parameter or env var?

plugins/kubernetes/test/controllers/kubernetes/clusters_controller_test.rb
```diff
@@ -166,7 +166,7 @@ def self.use_example_config
 
       it "updates" do
         patch :update, params: {id: cluster.id, kubernetes_cluster: {name: "NEW"}}
-        assert_redirected_to "/kubernetes/clusters"
+        assert_redirected_to cluster
```
bcolferzd: should this be kubernetes_cluster?

bcolferzd: I assume the redirect is to the cluster route so it should work.

bcolferzd (approved):
Looks to me like it will work.

plugins/kubernetes/app/controllers/kubernetes/clusters_controller.rb
grosser: that was to make all the file+path methods have the same name :)

plugins/kubernetes/app/controllers/kubernetes/clusters_controller.rb
grosser: secrets are v1 .. so either it's a constant or just stays this way 🤷‍♂️

PR #3161, opened 2019-01-30
expire vault list when deleting a server

grosser:
@zendesk/compute 

/cc @zendesk/bre

### DIFF
--- a/lib/samson/secrets/vault_server.rb
+++ b/lib/samson/secrets/vault_server.rb
@@ -29,8 +29,8 @@ module Samson
       validate :validate_cert
       validate :validate_connection
 
-      after_save :refresh_vault_clients
-      after_save :expire_secrets_cache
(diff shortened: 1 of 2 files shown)


lib/samson/secrets/vault_server.rb
```diff
@@ -29,8 +29,8 @@ class VaultServer < ActiveRecord::Base
       validate :validate_cert
       validate :validate_connection
 
-      after_save :refresh_vault_clients
-      after_save :expire_secrets_cache
+      after_commit :refresh_vault_clients
+      after_commit :expire_secrets_cache
```
yizhang-zen: What's the difference here?

grosser: after_save applies to create+update
after_commit applies to create+update+destroy

PR #3164, opened 2019-02-05
Add keep_name annotation to keep resource names

adammw:
Add keep_name annotation to keep resource names.

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -156,7 +156,12 @@ module Kubernetes
       end
     end
 
+    def keep_name?
+      template.dig(:metadata, :annotations, :'samson/keep_name') == 'true'
+    end
+
     def set_service_name
+      return if keep_name?
       template[:metadata][:name] = generate_service_name(template[:metadata][:name])
     end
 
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -157,6 +157,7 @@ def set_project_labels
     end
 
     def set_service_name
+      return if template.dig(:metadata, :annotations, :'samson/keep_name') == 'true'
```
grosser: test

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -329,7 +330,12 @@ def validate_replica_target_is_supported
     end
 
     def set_name
-      name = @doc.kubernetes_role.resource_name
+      keep_name = template.dig(:metadata, :annotations, :'samson/keep_name') == 'true'
```
grosser: extract method to dry up

grosser (approved):

grosser:
unintentional copy-paste might still get us in trouble with this ... but hopefully we are soon on namespace-per-project ...
what exactly was the kind you had trouble with ? ... add it to the exception list ?

adammw:
service account, since we wanted it to be shared but also had to be a prerequisite

grosser:
"shared" sounds kinda strange ... should only be in a single project 🤷‍♂️

adammw:
single project, shared between multiple roles..

grosser:
should still work then to use the name samson wants, but less obvious I guess 🤷‍♂️

adammw:
merged on red since FOSSA is acting up... all other tests are green.

PR #3166, opened 2019-02-06
Listed Soft Deleted deploys for compliance.

sathishavm:
* Listed Soft deleted deploys for compliance.

screen shot 2019-02-06 at 12 56 42 pm

/cc @zendesk/samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-1251

### Risks
- Level: Med

### DIFF
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -2,13 +2,11 @@
 require 'csv'
 
 class DeploysController < ApplicationController
-  include CurrentProject
-
-  skip_before_action :require_project, only: [:active, :active_count, :changeset]
-
+  before_action :require_project, except: [:active, :active_count, :changeset]
   before_action :authorize_project_deployer!, except: [:index, :show, :active, :active_count, :changeset]
   before_action :find_deploy, except: [:index, :active, :active_count, :new, :create, :confirm]
   before_action :stage, only: :new
+  helper_method :current_project
 
   def active
     @deploys = deploys_scope.active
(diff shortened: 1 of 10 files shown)


app/controllers/deploys_controller.rb
```diff
@@ -201,13 +201,21 @@ def deploy_params
...
   def deploys_scope
     current_project&.deploys || Deploy
   end
 
+  def current_project
+    @project
+  end
+
+  def require_project
+    @project = (Project.with_deleted { Project.find_by_param!(params[:project_id]) } if params[:project_id])
```
grosser: `Project.with_deleted {super }` ?

app/helpers/application_helper.rb
```diff
@@ -72,14 +72,23 @@ def breadcrumb(*items)
   # tested via link_to_resource
   def link_parts_for_resource(resource)
     case resource
-    when Project, DeployGroup, User, Samson::Secrets::VaultServer then [resource.name, resource]
+    when Project, DeployGroup, User, Samson::Secrets::VaultServer then
```
grosser: this is not good ... makes every link on the whole page slower :(

app/models/deploy.rb
```diff
@@ -64,6 +64,14 @@ def summary_for_timeline
     end
   end
 
+  def stage
+    super || (Stage.with_deleted { Stage.find(stage_id) } if stage_id)
```
grosser: same here ... not a good pattern / we need consistent behavior

grosser:
let's talk about this on Thursday and see if we can find a cleaner approach

PR #3169, opened 2019-02-08
List and filter Soft Deleted deploys for compliance.

sathishavm:
* List soft-deleted deploys with params `with_deleted=true` (eg: https://samson.zende.sk/deploys?with_deleted=true)
* Filter by deleted deploys

screen shot 2019-02-08 at 3 15 52 pm

/cc @zendesk/samson

### References
 - Jira link:   https://zendesk.atlassian.net/browse/BRE-1251

### Risks
- Level: Low

### DIFF
--- a/app/assets/stylesheets/deploys.scss
+++ b/app/assets/stylesheets/deploys.scss
@@ -61,6 +61,12 @@
   padding: 0 50px;
 }
 
+.deploy-search{
+  input[type="checkbox"] {
+    margin: 10px 0 0 15px;
+  }
+}
+
 #deploy-confirmation {
   display: none;
 
--- a/app/controllers/concerns/wrap_in_with_deleted.rb
+++ b/app/controllers/concerns/wrap_in_with_deleted.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+module WrapInWithDeleted
+  extend ActiveSupport::Concern
+
+  included do
+    around_action :wrap_in_with_deleted, if: proc {
+      request.get? && (with_deleted? || search_with_deleted?)
+    }
+  end
+
+  protected
+
+  def wrap_in_with_deleted
+    Project.with_deleted do
+      Stage.with_deleted do
+        Deploy.with_deleted do
+          yield
+        end
+      end
+    end
+  end
+
+  # For both deleted and non-deleted resources
+  def with_deleted?
+    params[:with_deleted] == "true"
+  end
+
+  def search_with_deleted?
+    params.dig(:search, :deleted) == "true"
+  end
+end
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -2,6 +2,8 @@
 require 'csv'
 
 class DeploysController < ApplicationController
+  around_action :wrap_in_with_deleted, if: proc { request.get? && params[:with_deleted] == "true" }
+
   include CurrentProject
 
   skip_before_action :require_project, only: [:active, :active_count, :changeset]
@@ -237,6 +239,16 @@ class DeploysController < ApplicationController
       if deploy_count > csv_limit
         csv << ['-', 'There are more records. Generate a full report at']
         csv << ['-', new_csv_export_url]
+      end
+    end
+  end
+
+  def wrap_in_with_deleted
+    Project.with_deleted do
+      Stage.with_deleted do
+        Deploy.with_deleted do
+          yield
+        end
       end
     end
   end
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -2,7 +2,9 @@
 require 'csv'
 
 class DeploysController < ApplicationController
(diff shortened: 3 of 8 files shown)


app/assets/stylesheets/deploys.scss
```diff
@@ -61,6 +61,12 @@
   padding: 0 50px;
 }
 
+.deploy-search{
```
grosser: this would be good to be more general for .search or whatever class we wrap around the search forms

app/controllers/deploys_controller.rb
```diff
@@ -2,6 +2,10 @@
 require 'csv'
 
 class DeploysController < ApplicationController
+  around_action :wrap_in_with_deleted, if: proc {
```
grosser: can you turn this into a concern and then move into ApplicationController so we can reuse it anywhere ?

app/controllers/deploys_controller.rb
```diff
@@ -2,6 +2,10 @@
...
+    request.get? && (params[:with_deleted] == "true" || params.dig(:search, :deleted_at) == "true")
```
grosser: can we use a single param ... otherwise passing it on with get complicated ...

app/controllers/deploys_controller.rb
```diff
@@ -181,6 +185,7 @@ def search
     if updated_at = search[:updated_at].presence
       deploys = deploys.where("updated_at between ? AND ?", *updated_at)
     end
+    deploys = deploys.where.not(deleted_at: nil) if search[:deleted_at].presence
```
grosser: use same logic here ... `== "true"`

grosser: extract method for the "is_this_deleted?" and reuse in around and here

app/views/deploys/_deploy.html.erb
```diff
@@ -11,7 +11,7 @@
       <% end %>
     </td>
     <td>
-      <%= link_to deploy.summary, [@project || deploy.project, deploy] %>
+      <%= link_to deploy.summary, [@project || deploy.project, deploy, with_deleted: deploy.deleted? ? true : nil] %>
```
grosser: I though it would be `soft_deleted_at?`

app/views/deploys/index.html.erb
```diff
@@ -27,9 +27,13 @@
 
       <%= search_select :production, [["Yes", "true"], ["No", "false"]], size: 1 %>
       <%= search_select :code_deployed, [["Yes", "true"], ["No", "false"]], title: "Code deployed?", label: "Code", size: 1 %>
-      <%= search_select :status, Job::VALID_STATUSES %>
+      <%= search_select :status, Job::VALID_STATUSES, size: 1 %>
       <%= search_select :group, Environment.env_deploy_group_array %>
       <%= search_select :time_format, User::TIME_FORMATS, selected: params.dig(:search, :time_format) || current_user.time_format, label: 'Time', size: 1 %>
+      <div class="col-sm-1">
+        <%= label_tag :Deleted %>
+        <%= check_box_tag 'search[deleted_at]', true, params.dig(:search, :deleted_at) %>
```
grosser: deleted_at looks like a date ... `deleted` or `soft_deleted`

app/assets/stylesheets/deploys.scss
sathishavm: We don't have a general CSS for `search`, Moreover search_form for deploys have different design patterns compare to other searches.

app/controllers/deploys_controller.rb
sathishavm: Right, but we have 2 different functionalities here.
1. list all deploys along with deleted deploys.
2. Filter and show only deleted deploys along with search results.

app/views/deploys/_deploy.html.erb
sathishavm: It is only deleted?, as per soft_delete gem.

app/controllers/deploys_controller.rb
sathishavm: Converted into reusable methods

app/controllers/concerns/wrap_in_with_deleted.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+  # For both deleted and non-deleted resources
+  def with_deleted?
+    params[:with_deleted] == "true"
+  end
+
+  def search_with_deleted?
+    params.dig(:search, :deleted) == "true"
```
grosser: change the input to `with_deleted` and then get rid of this ?

sathishavm: Fine, but we will miss the functionality of displaying the combined result of deleted and non-deleted deploys.
This combined results may help the user to view the deploys in exact order.
What's your thought on this functionality?

grosser: ah, I did not think of that usecase :D

PR #3175, opened 2019-02-13
allow any class to be requested via with_deleted

grosser:
- existing code will pass on the value of with_deleted
 - search input is not sent when not clicked, so we don't need a "== true" check

/cc @zendesk/samson @zendesk/bre @sathishavm

### DIFF
--- a/app/controllers/concerns/wrap_in_with_deleted.rb
+++ b/app/controllers/concerns/wrap_in_with_deleted.rb
@@ -3,29 +3,27 @@ module WrapInWithDeleted
   extend ActiveSupport::Concern
 
   included do
-    around_action :wrap_in_with_deleted, if: proc {
-      request.get? && (with_deleted? || search_with_deleted?)
-    }
+    around_action :wrap_in_with_deleted
   end
 
   protected
 
-  def wrap_in_with_deleted
-    Project.with_deleted do
-      Stage.with_deleted do
-        Deploy.with_deleted do
-          yield
-        end
-      end
+  def search_deleted
+    params.dig(:search, :deleted)
+  end
+
+  def wrap_in_with_deleted(&block)
+    # when searching for deleted, make sure results are shown
+    if deleted = search_deleted
+      params[:with_deleted] ||= deleted
     end
-  end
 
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/wrap_in_with_deleted.rb
```diff
@@ -3,29 +3,21 @@ module WrapInWithDeleted
...
-  def wrap_in_with_deleted
-    Project.with_deleted do
-      Stage.with_deleted do
-        Deploy.with_deleted do
-          yield
-        end
-      end
+  def wrap_in_with_deleted(&block)
+    if request.get? && value = (params[:with_deleted].presence || search_with_deleted?)
+      klasses = value.split(",").map(&:safe_constantize)
```
sathishavm: It this is safe?

grosser:
I'm not planing to change the view, just adding this as an escape-hatch for admins/auditors to do funky things without going through the console

app/controllers/concerns/wrap_in_with_deleted.rb
grosser: it says "safe_" :trollface: 

users could summon any constant here, but we only call `with_deleted` on it ... so kinda hard to abuse ... we could limit it to AR models 🤷‍♂️

app/controllers/concerns/wrap_in_with_deleted.rb
```diff
@@ -3,29 +3,21 @@ module WrapInWithDeleted
...
   end
 
-  # For both deleted and non-deleted resources
-  def with_deleted?
-    params[:with_deleted] == "true"
-  end
-
   def search_with_deleted?
-    params.dig(:search, :deleted) == "true"
+    params.dig(:search, :deleted)
```
sathishavm: Need to change the view to pass "Project,Stage,Deploy" for `search[:deleted]` field or else the value will be `nil` when we try to filter in deploys page.

grosser: ah yes, good catch!
I'm setting with_deleted from search query now to avoid having to check multiple params and so links work

PR #3176, opened 2019-02-16
Fix to project locks for project admin

sathishavm:
* Fix for user with role as `project_admin` can't able to lock the project.
* Fix for locks controller to fetch the current project and current stage as per resource.

/cc @zendesk/samson

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-1268

### Risks
- Level: Med

### DIFF
--- a/app/controllers/concerns/current_user.rb
+++ b/app/controllers/concerns/current_user.rb
@@ -58,8 +58,7 @@ module CurrentUser
 
   # tested via access checks in the actual controllers
   def authorize_resource!
-    scope = current_stage || current_project if controller_name.to_sym == :locks
-    unauthorized! unless can?(resource_action, controller_name.to_sym, scope)
+    unauthorized! unless can?(resource_action, controller_name.to_sym)
   end
 
   def resource_action
--- a/app/controllers/concerns/current_user.rb
+++ b/app/controllers/concerns/current_user.rb
@@ -58,7 +58,8 @@ module CurrentUser
 
   # tested via access checks in the actual controllers
   def authorize_resource!
-    unauthorized! unless can?(resource_action, controller_name.to_sym)
+    scope = current_stage || current_project if controller_name == 'locks'
+    unauthorized! unless can?(resource_action, controller_name.to_sym, scope)
   end
 
   def resource_action
@@ -66,11 +67,7 @@ module CurrentUser
   end
 
   def can?(action, resource_namespace, scope = nil)
-    scope ||= if resource_namespace == :locks
-      current_stage || current_project
-    elsif respond_to?(:current_project, true)
-      current_project
-    end
+    scope ||= current_project if respond_to?(:current_project, true)
     AccessControl.can?(current_user, action, resource_namespace, scope)
   end
 end
--- a/app/controllers/concerns/current_user.rb
+++ b/app/controllers/concerns/current_user.rb
@@ -66,8 +66,8 @@ module CurrentUser
   end
 
   def can?(action, resource_namespace, scope = nil)
-    scope ||= if resource_namespace == :locks && respond_to?(:current_stage)
-      current_stage
+    scope ||= if resource_namespace == :locks
+      current_stage || current_project
     elsif respond_to?(:current_project, true)
       current_project
     end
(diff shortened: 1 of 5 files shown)


grosser:
some description would be nice, even if it's copy-paste from jira since nobody else on the pulic repo can read this :D

app/controllers/locks_controller.rb
```diff
@@ -59,32 +58,30 @@ def format_errors(object)
...
+    @project = find_resource
   end
 
-  # Overrides CurrentStage#require_stage
-  def require_stage
+  def find_resource
     case action_name
     when 'create' then
-      @stage = Stage.find(params[:lock][:resource_id])
+      resource_class.find(params[:lock][:resource_id]) if (params[:lock] || {})[:resource_type] == @resource_class
```
grosser: I doubt this works ... or is the `@resource_class` a string ?

grosser: ... looks like it is :(

grosser: prefer `params.dig(:lock, :resource_type)`

grosser:
needs some new tests ... because apparently something was broken before and it's not ?

app/views/locks/_button.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if can? :write, :locks, @stage %>
+<% if can? :write, :locks, resource %>
```
grosser: make sure to click through environment/deploy-group/project and check that it still works, idk why `@stage` was here, but now the logic will be a lot different ...

app/controllers/locks_controller.rb
```diff
@@ -59,32 +58,30 @@ def format_errors(object)
...
-      (params[:lock] || {})[:resource_type] == "Stage"
-    when 'destroy'
-      lock.resource_type == "Stage"
-    when 'index', 'destroy_via_resource'
-      false
-    else
-      raise 'Unsupported action'
-    end
+  def require_stage
+    @resource_class = "Stage"
```
grosser: better set the actual class here `Stage` and then compare to `@resource_class.name` ? (instead of constantizing)

app/controllers/locks_controller.rb
```diff
@@ -59,32 +58,30 @@ def format_errors(object)
...
+  def find_resource
     case action_name
     when 'create' then
-      @stage = Stage.find(params[:lock][:resource_id])
+      resource_class.find lock_params[:resource_id] if lock_params[:resource_type] == @resource_class
     when 'destroy' then
-      @stage = lock.resource
-    else
-      raise 'Unsupported action'
+      lock.resource if lock.resource_type == @resource_class
```
grosser: we don't need this before-action on destroy ?

test/controllers/locks_controller_test.rb
```diff
@@ -188,13 +178,38 @@ def create_lock(resource = nil, options = {})
...
+  as_a :project_admin do
+    describe '#create' do
+      it 'creates a project lock' do
+        delete_at = Time.now + 1.hour
+        create_lock project, delete_at: delete_at
+        assert_redirected_to '/back'
+        assert flash[:notice]
+
+        lock = project.lock
+        (lock.delete_at&.to_i).must_equal delete_at.to_i
```
grosser: simpler to do a `assert_difference 'Lock.count', +1 do`

grosser: add test for `it "does not allow creating a deploy-group lock` ?

app/controllers/concerns/current_user.rb
```diff
@@ -58,19 +58,16 @@ def authorize_project_deployer!
 
   # tested via access checks in the actual controllers
   def authorize_resource!
-    unauthorized! unless can?(resource_action, controller_name.to_sym)
+    scope = current_stage || current_project if controller_name.to_sym == :locks
+    unauthorized! unless can?(resource_action, controller_name.to_sym, scope)
```
grosser: how about we get rid of this hack and override `def authorize_resource!` in locks_controller ?
... the all the resource finding logic can go in there too ...

app/controllers/locks_controller.rb
sathishavm: we need this to determine, whether we need to assign the found resource to instance variable in require_stage and require_project methods.

PR #3177, opened 2019-02-16
Webhooks to use ResourceController

danihodovic:
Allows us to use JSON endpoints for webhooks.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -20,10 +20,6 @@ class WebhooksController < ResourceController
   end
 
   def resource_params
-    puts(current_project.inspect)
-    puts("super is #{super}")
     super.permit(:branch, :source).merge(project: current_project)
-
-    params.require(:webhook).permit(:branch, :stage_id, :source)
   end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/webhooks_controller.rb
```diff
@@ -1,33 +1,42 @@
...
+    @project.webhooks
   end
 
-  private
+  def resources_path
+    project_webhooks_path(current_project)
+  end
+
+  def resource_params
+    super.permit(:branch, :source)
```
grosser: I think when an added `.merge(project: current_project)` you can delete all of `def create`

config/routes.rb
```diff
@@ -47,7 +47,7 @@
 
     resource :changelog, only: [:show]
     resource :stars, only: [:create]
-    resources :webhooks, only: [:index, :create, :destroy]
+    resources :webhooks, only: [:index, :create, :update, :destroy]
```
grosser: should add a test for `:update` then

grosser:
also fixed in https://github.com/zendesk/samson/pull/3179

PR #3178, opened 2019-02-16
Outbound webhooks to use ResourceController

danihodovic:
Allows us to render the outbound webhooks as JSON.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High


grosser:
that looks pretty good ... let me see if I can make it work ...

PR #3179, opened 2019-02-16
Danihodovic outbound webhooks resource controller

grosser:
@zendesk/bre 

replaces https://github.com/zendesk/samson/pull/3178
@danihodovic 

Now showing inline errors on the page too:
screen shot 2019-02-16 at 8 51 42 am

### DIFF
--- a/app/controllers/outbound_webhooks_controller.rb
+++ b/app/controllers/outbound_webhooks_controller.rb
@@ -1,34 +1,47 @@
 # frozen_string_literal: true
 require 'samson/integration'
 
-class OutboundWebhooksController < ApplicationController
+class OutboundWebhooksController < ResourceController
   include CurrentProject
-
   before_action :authorize_project_deployer!
 
   def create
-    webhook = OutboundWebhook.new(outbound_webhook_params.merge(project_id: current_project.id))
+    webhook = OutboundWebhook.new(resource_params)
 
-    if webhook.save
-      flash.delete(:error)
-      redirect_to project_webhooks_path(current_project)
-    else
-      flash[:error] = webhook.errors.full_messages.join(', ')
-      @new_outbound_webhook = webhook
-      render 'webhooks/index'
+    respond_to do |format|
+      format.html do
+        if webhook.save
+          flash.delete(:error)
+          redirect_to project_webhooks_path(current_project)
+        else
+          flash[:error] = webhook.errors.full_messages.join(', ')
+          @new_outbound_webhook = webhook
+          render 'webhooks/index'
+        end
+      end
+      format.json do
+        webhook.save!
+        render_as_json :webhook, webhook
+      end
     end
-  end
-
-  def destroy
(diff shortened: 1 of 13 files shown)


app/controllers/outbound_webhooks_controller.rb
```diff
@@ -1,34 +1,28 @@
...
-
-    if webhook.save
-      flash.delete(:error)
-      redirect_to project_webhooks_path(current_project)
-    else
-      flash[:error] = webhook.errors.full_messages.join(', ')
-      @new_outbound_webhook = webhook
-      render 'webhooks/index'
-    end
+    super(template: 'webhooks/index')
```
grosser: FYI new pattern I'm starting since I'm getting sick of all these 1-off helper methods ... this seems to work fine and means we have less methods flying around

app/controllers/outbound_webhooks_controller.rb
```diff
@@ -1,34 +1,28 @@
...
   end
 
-  private
+  def resources_path
+    [@project, 'webhooks']
+  end
 
-  def outbound_webhook_params
-    params.require(:outbound_webhook).permit(:stage_id, :project_id, :url, :username, :password)
+  def resource_params
```
grosser: project_id was never sent in and should not be sent in

grosser:
... also added json index for both and made sure not to send the passwords out :D

danihodovic (approved):

danihodovic:
This is not working as expected
```
$ echo '{"source": "mysrc", "branch": "master", "stage_id": 1}'| http --headers POST 'localhost:9080/projects/be5343b7-d558-5385-8be5-f314a8ef972f/webhooks.json' "Authorization: Bearer token"

HTTP/1.1 500 Internal Server Error
```

Samson logs
```
Completed 500 Internal Server Error in 44ms (ActiveRecord: 1.4ms)



NoMethodError (undefined method `admin_for?' for nil:NilClass):

app/models/access_control.rb:14:in `can?'
app/controllers/concerns/current_user.rb:74:in `can?'
app/controllers/concerns/current_user.rb:61:in `authorize_resource!'
```

Suspect change: https://github.com/zendesk/samson/pull/3179/files#diff-f1be353a0a407c7748e57e2e397809fdR7

If you modify that to
```ruby

```

grosser:
yeah that's because the filters are now infront of the `login user` filter :(

grosser:
FYI tried doing some "insert action at location" magic but that did not work out :(

```
  def self.insert_before_action(filter, **options, before:)
    chain = __callbacks[:process_action].send(:chain)
    location = chain.index { |callback| callback.filter == before } || raise("Did not find #{before} in the chain")
    chain.insert(location, ActiveSupport::Callbacks::Callback.build(filter))
  end
```

grosser:
better now ✨

danihodovic (approved):

sathishavm (approved):
Nice!!, Good improvement.

PR #3181, opened 2019-02-18
Fix ActionMailer for Gmail

danihodovic:
This fixes two things:

1) parsing the smtp settings when they contain a user with an @ symbol.

```
smtp_url = URI.parse(ENV["SMTP_URL"] || "smtp://localhost")
smtp_url.user = 'test@gmail.com' # process crash here
```
the URI.parse call fails with any @ symbols. By assigning the smtp_url
variable with a string containing '@' the process crashes.

2) allows the user to configure smtp over tls if provided in the
environment. This is required for gmails smtp servers.

Both changes should be backwards compatible.

gmail

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low/Med/High

### DIFF
--- a/config/initializers/smtp.rb
+++ b/config/initializers/smtp.rb
@@ -2,16 +2,17 @@
 require 'uri'
 
 smtp_url = URI.parse(ENV["SMTP_URL"] || "smtp://localhost")
-smtp_url.port ||= 25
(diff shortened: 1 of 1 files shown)


grosser:
thx, good details :)

Issue #3172, opened 2019-02-10
Inconsistent stages API between GET and PATCH


danihodovic:
The Samson API differs in the data that is returned from a GET /stages/stage.json` and what you can supply in a POST/PATCH request. This inconsistency in the API makes it difficult to build an idempotent Ansible modules because I need to be able to compare what values the user has provided and the values that exist in the remote state. Because the GET request hides parameters in the stage it's impossible to compare the local and remote state.

E.g you can PATCH `command_ids: [1,2,3]`, but when you GET a stage it does not contain the `command_ids` property. The same applies for slack webhooks and a few other properties.

grosser:
can you pass extra arguments when doing the get, so if we add these extra
things then it's not expensive for everyone ?
like `get stages/foo.json?gimme_everything=true`

danihodovic:
Yes I can. Ansible uses a http client under the hood. However I wonder what the use case is exactly? The JSON endpoints don't affect normal users page loads (html), but is used by programs that access the Samson API. Do you guys use some sort of automated tooling at Zendesk that already uses these endpoints and needs to be quick?

So the issues I see with it are:
1) Additional complexity
2) More inconsistency in the API endpoints, e.g projects and commands wouldn't have this extra parameter but stages would.

Let me know what you think.

grosser:
The basic issue is that there are lots of parameters that can be passed into the stage and returning all of them would mean rendering lots of the stages associations. Rendering association is normally done via `includes`, and we want `index` to be `[show, show, show, show]` which then would make `index` very big/slow :(
... so there needs to be an inconsistency somewhere, either it's `index!=[show]` or show gets an additional `?return_all_attributes=true` parameter where we then go over all assignable associations and render them inline (and not how we normally do `include`)

danihodovic:
If consistency would require a major overhaul of the API I think it's better to use your proposition in the short term and figure out a longer term plan if more users consider the API complex to use :) 

The HTTP client is thicker than I would like it to be because it needs to handle the inconsistencies, but I'll be reporting them here over time because I think there are a lot of small issues that can easily be fixed.

grosser:
Another alternative we be to could consider the "_attributes" params not
part of the api and then create each association individually,
but that would mean lots of new controllers for each 1-off association,
which would also mean a bunch of work.

Let's start with a list of all the things you are using, so we get a
pattern going and then add things as needed.

danihodovic:
:+1:

[...]

PR #3182, opened 2019-02-18
allow inlining association attributes for idempotent api updates

grosser:
for https://github.com/zendesk/samson/issues/3172

@zendesk/bre

### DIFF
--- a/app/controllers/stages_controller.rb
+++ b/app/controllers/stages_controller.rb
@@ -23,6 +23,11 @@ class StagesController < ResourceController
           if params[:include] == "kubernetes_matrix"
             reply[:stage][:kubernetes_matrix] = Kubernetes::DeployGroupRole.matrix(@stage)
           end
+
+          # for idempotent updates via api
+          if params[:inlines].to_s.split(",").include?("slack_webhooks_attributes")
+            reply[:stage][:slack_webhooks_attributes] = @stage.slack_webhooks.map(&:attributes)
+          end
         end
       end
       format.svg do
--- a/test/controllers/stages_controller_test.rb
+++ b/test/controllers/stages_controller_test.rb
@@ -139,6 +139,17 @@ describe StagesController do
           get :show, params: {project_id: subject.project.to_param, id: 123123}
         end
       end
+
+      it "inlines slack_webhooks_attributes" do
+        subject.slack_webhooks.create!(webhook_url: "https://xyz.com", before_deploy: true)
+        get :show, params: {
+          project_id: subject.project.to_param, id: subject.to_param,
+          inlines: "slack_webhooks_attributes"
+        }, format: :json
+        json["stage"]["slack_webhooks_attributes"].first.slice("webhook_url", "before_deploy").must_equal(
+          "webhook_url" => "https://xyz.com", "before_deploy" => true
+        )
+      end
     end
 
     describe "#index" do


danihodovic:
I'll give this a try. Is it possible to inline **all** related entities instead of having to specify what you want?

danihodovic:
So I could theoretically read the upstream webhooks and read the webhooks the user has provided, see how the state diffs and update based on that, but I don't think it's good DX. It would either require the user to specify a webhook ID up front (so that a new webhook isn't created) or I would have to perform some magic in the Ansible module to compare webhook attributes and if a webhook with the parameters doesn't exist create one.

Both of these solutions are not good enough because specifying `slack_webhook_attributes: []` in a PUT or a PATCH does not remove the existing webhooks, since the Slack webhooks are their own DB entity. The UI sort of hacks around this by specifying a property called `_delete` nested inside of the Slack webhook to remove a webhook. This isn't great when dealing with the API directly.

I think we should either ensure that PUTs replace the entire stage (including all the nested sub-resources) or separate Slack webhooks so that they're a  REST resource with it's own endpoints.

danihodovic:
I'm also a little confused of the difference between PUT and PATCH. I imagined that PUT would replace the resource entirely while PATCH would work similar to json-patch, i.e update individual attributes in place.

grosser:
- patch and put are the same in rails, just aliased
 - rebuilding something that assigns associations without checking ids sounds dangerous/complicated
 - we can have an api controller without the UI actually using that 🤷‍♂️

grosser:
... having separate rest resource get's us into trouble with "who has access to this" duplication :/

grosser:
I'm also open to making it return all associations that can be assigned, we just need to sanity-check that there are no secrets in the settings ...
so ideally you could figure out some client logic that stores the extra ids that get returned ...

danihodovic:
Is it possible to return these properties on any method? After a PUT I'll need to perform another GET instead of just reading the returned payload.

grosser:
yeah, that's all the same `render_as_json` logic ... so it's possible to do wherever

grosser:
... so sounds like what you really need is either apis for all associations or assignable arrays without ids.
... assignable arrays could be pretty simple hack and reusable for other projects, so I'll give that a try
.

danihodovic:
@grosser I don't want to enforce design on the API just so that it can suite my little Ansible module ;) . More complex infra-as-code tools such as Terraform can likely work around this because it can save state between runs.

If you think assignable arrays might solve the problem and doesn't impose too many changes in the API go for it. I'm not really sure what you mean by assignable arrays - but would something like this allow the stage attributes to be updated in place (including deleting webhooks)?

grosser:
If you find nice Ansible pattern that would be preferable ... or how
terraform does this ...
I want samson to be usable by terraform mostly since we also use that
internally
assignable array would mean that we update all attributes that come in and
delete the ones that do not exist
hooks = [{url: foo}, {url: bar}]
... but I guess that would produce a lot of updates if something from the
middle was missing and then all other hooks would be updated ...
so maybe independent api is best ... but it's some boilerplate + new
routing, so not fun :)

danihodovic:
> ... but I guess that would produce a lot of updates if something from the middle was missing and then all other hooks would be updated ...

This is implying that a stage would have many webhooks. I suspect that 99% of the use cases people have 1 or 3 webhooks active, usually to a deployment channel.

grosser:
yes, but it's a bad pattern to build if it cannot scale :)

PR #3187, opened 2019-02-25
Add a multi-layer dockerfile

lox:
This adds a multi-layer dockerfile that can build both production and test docker images.

These are reflected in the `docker-compose.yml` file, which include the database dependencies for running tests against sqlite, mysql and postgres environments:

```
docker-compose run samson-test-sqlite
docker-compose run samson-test-mysql
docker-compose run samson-test-postgres
```

Both sqlite and mysql run and pass, postgres doesn't. Is this working in Travis? 

I figured this might be a useful thing for reproducible test environments in CI vs local development. Feel free to close if not!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,9 +1,6 @@
 # git history is big, do not add it
 .git
 
-# tests are big and not needed
-test
-
 # user might chose to have local gem installs, they can be quiet big
 vendor/bundle
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,70 @@
-FROM ruby:2.5.3-slim
+FROM ruby:2.5.3-slim AS base
 
-RUN apt-get update && apt-get install -y build-essential default-libmysqlclient-dev libpq-dev libsqlite3-dev wget apt-transport-https git curl
-RUN curl -sL https://deb.nodesource.com/setup_6.x | bash - && apt-get install nodejs -y
-RUN curl -fsSL https://get.docker.com | bash -
+# Install dependencies
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+      build-essential \
+      default-libmysqlclient-dev \
+      libpq-dev \
+      libsqlite3-dev \
+      wget \
+      apt-transport-https \
+      git \
+      curl \
+      gnupg2 \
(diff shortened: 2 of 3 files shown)


grosser:
In general the direction looks good, but it feels a little too big of a change ... how about we start with just multi-layer and keep the rest the same for now ?

I'd prefer to also not set the RAILS_ENV/PRECOMPILE so the images stay clean ... not sure if that has any big downsides though

lox:
What is the current Dockerfile / docker-compose.yml used for @grosser?

lox:
My aim was basically to get things setup so that I could run the test suites locally against mysql / sqlite / postgres, how are y'all doing that currently?

grosser:
Dockerfile is for:
- releasing to docker
- local dev server (change things locally and then `compose up`)

grosser:
I'm mostly running it locally and then just swap out the database.yml to what I need

lox:
> I'm mostly running it locally and then just swap out the database.yml to what I need

Fair enough! This PR might be an unmerited level of complexity then! 

> I'd prefer to also not set the RAILS_ENV/PRECOMPILE so the images stay clean ... not sure if that has any big downsides though

I believe the sprockets/assets stuff needs these set. The tests seem to depend on non-production assets.

lox:
Shall I close this out for now @grosser?

grosser:
yeah, can close for now, some incremental improvements would be nice though, but that can be smaller PRs :)

grosser:
afaik the reason to break out different stages in the dockerfile is to make the final image smaller ... but this does not happen here since everything uses the same big base image 🤷‍♂️

lox:
There are a few reasons to do it, and a few reasons why it makes things more complicated. The main benefit is cacheability. In your image, the bundle process is the slowest, so putting it in its own stage means that even if the node modules change the cache on the bundle stage remains intact. In the future when [Buildkit](https://docs.docker.com/develop/develop-images/build_enhancements/) is a thing then you get concurrent image building too. As you say, you also can get image size improvements as you don't drag along dependencies from the build process from one layer into another. It wasn't a big priority for me though.

grosser:
parallel sounds interesting :)

PR #3188, opened 2019-02-25
cleanup dockerfile a bit

grosser:
mostly cherry-picked from https://github.com/zendesk/samson/pull/3187

@zendesk/bre @lox

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,47 @@
 FROM ruby:2.5.3-slim
 
-RUN apt-get update && apt-get install -y build-essential default-libmysqlclient-dev libpq-dev libsqlite3-dev wget apt-transport-https git curl
-RUN curl -sL https://deb.nodesource.com/setup_6.x | bash - && apt-get install nodejs -y
-RUN curl -fsSL https://get.docker.com | bash -
+# Install dependencies
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends \
+    build-essential \
+    default-libmysqlclient-dev \
+    libpq-dev \
+    libsqlite3-dev \
+    wget \
+    apt-transport-https \
+    git \
+    curl \
+    gnupg2 \
+  && curl -sL https://deb.nodesource.com/setup_6.x | bash - \
+  && apt-get install nodejs -y \
+  && rm -rf /var/lib/apt/lists/* \
+  && curl -fsSL https://get.docker.com | bash -
 
 WORKDIR /app
 
 # Mostly static
-COPY config.ru /app/
-COPY Rakefile /app/
-COPY bin /app/bin
-COPY public /app/public
-COPY db /app/db
-COPY .env.bootstrap /app/.env
-COPY .env.virtualbox /app/
-COPY .ruby-version /app/.ruby-version
+COPY .ruby-version config.ru Rakefile .env.virtualbox ./
+COPY bin bin
+COPY public public
+COPY db db
+COPY .env.bootstrap .env
 
 # NPM
-COPY package.json /app/package.json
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -34,7 +47,14 @@ COPY lib /app/lib
 
 # Assets
 COPY vendor/assets /app/vendor/assets
-RUN echo "takes 5 minute" && ./bin/decode_dot_env .env && RAILS_ENV=production PRECOMPILE=1 bundle exec rake assets:precompile
+RUN echo "takes 5 minute" && ./bin/decode_dot_env .env && RAILS_ENV=production PRECOMPILE=1 bundle exec rake assets:precompile 2>/dev/null
+
+# run as non-root, but allow it to access docker, use id so kubernetes non-root check passes
+RUN groupadd -r samson \
+  && useradd -r -g samson -d /app -u 1000 samson \
+  && chown -R samson:samson /app \
```
lox: This will duplicate the entire source code layer unfortunately, check out the output of `docker history samson`.

lox:
This image ends up with a HEAP of layers, I'd suggest combining as many of those `COPY` directives as possible, unlike `ADD` they can have multiple params.

lox:
Out of interest, why do you silence the output of bundler and npm?

grosser:
output was annoying, I wanted to get rid of it :D

lox:
I view it as a good indicator that you've got a cache miss :) Also when things break it's helpful.

grosser:
I'll check the layers, the image size was the same so I though this is fine
...

lox:
This is what I ended up with after flattening the multi-stage down in to one: 

https://github.com/jwswj/samson/blob/master/.buildkite/Dockerfile

lox:
For the Dockerfile in this PR I get:

```bash
docker build --tag samson --no-cache .  0.92s user 0.68s system 0% cpu 4:29.87 total

$ docker images samson
REPOSITORY          TAG                 IMAGE ID            SIZE
samson              latest              6e690d8b0d2d        914MB

$ docker history samson | wc -l
      40
```

For the one I linked to I get:

```bash
docker build -f .buildkite/Dockerfile --tag samson-bk --no-cache .  1.00s user 0.59s system 0% cpu 4:31.84 total

$ docker images samson-bk
REPOSITORY          TAG                 IMAGE ID            SIZE
samson-bk           latest              6cae0d5bc807        914MB

$ docker history samson-bk | wc -l
      29
```

Doesn't end up a huge difference in size or build time, but the latency of less layers in push and pull ends up being meaningful.

grosser:
... removed the non-root for now and this is just a cleanup PR / reduces layers

Dockerfile
```diff
@@ -1,30 +1,38 @@
...
+  && rm -rf /var/lib/apt/lists/* \
+  && curl -fsSL https://get.docker.com | bash -
 
 WORKDIR /app
 
 # Mostly static
-COPY config.ru /app/
-COPY Rakefile /app/
+COPY .ruby-version config.ru Rakefile .env.virtualbox /app/
 COPY bin /app/bin
```
sathishavm: We no need to prepend '/app' everywhere, as we are in `WORKDIR /app`, just specify target file or folder.
eg: COPY bin bin
COPY .env.bootstrap .env

grosser:
good catch :D

Dockerfile
sathishavm: Still some https://github.com/zendesk/samson/blob/5e97920ef5ed1b1529df91a91154da062730cd8f/Dockerfile#L39

grosser:
got em all now :)

PR #3190, opened 2019-02-25
Setup Fossa to analyze code and send results to our internal tool

plonergan:
* This change simplifies the Fossa setup to just analyze the code and send the results to our internal tool.  The use of only the "fossa" command initializes the Fossa client (removing the need for the .fossa.yml configuration file), analyzes the code (so fossa --analyze isn't needed anymore), and sends the results to our internal tool.  The status of the scan is still shown through the Fossa scan status badge in the README.  I've also allowed failures of the Fossa job so that it should stop blocking builds if Fossa encounters any issues.

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low - This change simplifies the prior Fossa setup and since failures are now allowed, it should stop blocking builds if Fossa encounters any issues.


grosser:
This disables `test` mode, it just runs `analyze`

if that is intended:
 - change description to say something like "send to internal reporting tool" or so
 - change to `./fossa analyze`

if that is not intended:
 - bring back `test` with `travis_wait`

sathishavm:
Do we need Fossa in Travis?. Fossa is in GitHub Apps and can we configure it to test through that?? 
I hope some projects have integrated it.

grosser:
that would be neat :)

grosser:
any update ? ... want to just change it to analyze ?

plonergan:
@grosser I've updated the comments and description to explain the setup change.  I'm happy with this setup.

grosser:
the comment in .travis.yml needs an update too :)

sathishavm:
Added an IT support ticket https://zendeskit.zendesk.com/hc/en-us/requests/194196

grosser:
Pushed an update to simplify setup/docs ... let me know if that looks alright.
 - explicitly calling `analyze`
 - removing allow-failure since analyze never failed

PR #3193, opened 2019-02-27
PAAS-2867 show user when cluster is waiting for resources

grosser:
we start using autoscaler now, so when no resources are available the deploy has to wait

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
@@ -394,8 +394,8 @@ describe Kubernetes::DeployExecutor do
             items: [
               {
                 type: 'Warning',
-                reason: 'FailedScheduling',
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
```diff
@@ -394,8 +394,8 @@ def worker_is_unstable
             items: [
               {
                 type: 'Warning',
-                reason: 'FailedScheduling',
-                message: "fit failure on node (ip-1-2-3-4)\nfit failure on node (ip-2-3-4-5)",
+                reason: 'BadStuff',
```
adammw: Can we find an actual error reason kubernetes might emit to test against?

adammw:
How long does it wait?

adammw (approved):

grosser:
for now the maximum wait time of the deploy (20 min)
if we need we can make that shorter, but that's more logic, so keeping it
simple for now

PR #3197, opened 2019-03-01
convert locks controller to resource

grosser:
getting rid of lots of repetitive code / normalizing behavior
also removing destroy_via_resource since that's just a destroy with a different route ...


app/controllers/deploys_controller.rb
```diff
@@ -184,7 +184,7 @@ def search
       deploys = deploys.where("updated_at between ? AND ?", *updated_at)
     end
     deploys = deploys.where.not(deleted_at: nil) if search_deleted
-    pagy(deploys, page: page, items: 30)
+    pagy(deploys, page: params[:page], items: 30)
```
adammw: do you need to permit :page or are you not using strong params?

app/controllers/locks_controller.rb
```diff
@@ -50,29 +20,20 @@ def lock_params
...
-        nil
-      else
-        raise
-      end
+  def set_resource
+    if action_name == 'destroy' && !params[:id]
+      # NOTE: using .fetch instead of .require since we support "" as meaning "global"
+      id = params.fetch(:resource_id).presence
+      type = params.fetch(:resource_type).presence
+      raise if !type ^ !id # global or exact are ok, but not just id or just type
```
adammw: raise with error

app/controllers/locks_controller.rb
```diff
@@ -50,29 +20,20 @@ def lock_params
...
+      assign_resource Lock.where(resource_id: id, resource_type: type).first!
+    else
+      super
     end
   end
 
   # TODO: make CurrentUser handle dynamic scopes and remove this
   def authorize_resource!
-    unauthorized! unless can?(resource_action, controller_name.to_sym, lock&.resource)
+    unauthorized! unless can?(resource_action, controller_name.to_sym, @lock&.resource)
```
adammw: how come there's no attr_reader here?

app/controllers/resource_controller.rb
```diff
@@ -100,6 +110,7 @@ def destroy
   private
 
   def search_resources
+    resource_class
```
adammw: where is this used / tested?

test/controllers/locks_controller_test.rb
```diff
@@ -270,12 +270,12 @@ def create_lock(resource = nil, options = {})
       end
     end
 
-    describe "#destroy_via_resource" do
+    describe "#destroy via resource" do
```
adammw: not sure I understand the "via resource" part here

adammw (approved):

app/controllers/deploys_controller.rb
grosser: no, params are freely useable unless they are assigned to models

app/controllers/locks_controller.rb
grosser: not needed :D

app/controllers/resource_controller.rb
grosser: tested by new test in resouce_controller_test ... the default index response is "all resources"

test/controllers/locks_controller_test.rb
grosser: old name was kinda weird ... changed it to `destroy via query`

grosser: ... we don't really need this since api users could just hit index and find the lock they want to deleted ... old cruft :(

PR #3202, opened 2019-03-01
PAAS-2924 reload events before displaying them since some are added a…

grosser:
…fter pod signals restart

before:

```
[22:46:55] POD EVENTS:
[22:46:55]   Normal Scheduled: Successfully assigned default/example-server-7d9dcf68c-nn6dj to docker-desktop
```

after:
```
[22:57:21] POD EVENTS:
[22:57:21]   Warning BackOff: Back-off restarting failed container
```

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -91,7 +91,8 @@ module Kubernetes
         events.any? && events_indicating_failure.all? { |e| e[:reason] == "FailedScheduling" }
       end
 
-      def events
+      def events(reload: false)
+        @events = nil if reload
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -91,7 +91,8 @@ def waiting_for_resources?
         events.any? && events_indicating_failure.all? { |e| e[:reason] == "FailedScheduling" }
       end
 
-      def events
+      def events(reload: false)
+        @events = nil if reload
```
adammw: would personally prefer `return @events if @events && !reload`, then just an equal later

adammw (approved):

PR #3203, opened 2019-03-01
PAAS-2924 tell users to use describe for tricky errors

grosser:
@zendesk/compute


hgokavarapuz (approved):

adammw:
:/ okay as a temporary fix, but really samson should provide all the information that kubectl describe would otherwise provide...

grosser:
it already does ... events+logs ... but it's guiding users to deeper debugging if that is still not enough 🤷‍♂️

PR #3204, opened 2019-03-02
Stage reorder fails if we paginate

sathishavm:
* Stage list page has draggable reorder function and so it doesn't require pagination

/cc @zendesk/samson

### Risks
- Level: Low

### DIFF
--- a/app/controllers/resource_controller.rb
+++ b/app/controllers/resource_controller.rb
@@ -5,9 +5,9 @@ require 'csv'
 class ResourceController < ApplicationController
   include JsonRenderer
 
-  def index(with_pagination: true)
+  def index(paginate: true)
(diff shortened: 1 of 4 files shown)


app/controllers/resource_controller.rb
```diff
@@ -5,13 +5,17 @@
 class ResourceController < ApplicationController
   include JsonRenderer
 
-  def index
+  def index(with_pagination: true)
```
grosser:
```suggestion
  def index(paginate: true)
```

grosser:
how about html only and we keep json consistent ?

grosser:
I changed the title since I guess the real reason for this PR is that ordering broke, right ?

PR #3206, opened 2019-03-05
Handle icon for non-standard statuses

goyalmunish:
## Handle Status Icon for keys other than success/failure/missing/pending/error

**`CommitStatus#statuses`** can produce values such as ` {:state=>"Old Release", :description=>"v4.3 was deployed to deploy groups in this stage by Production"}, {:state=>"Production Only Reference", :description=>"v4.2 has not been deployed to a non-production stage."}` in case of combined status using **`release_status`** and **`ref_statuses`**.

**Idea** is to suppress the exceptions and highlight concerns such as `Old Release` and `Production Only Reference` as `"error"/"warning"`.

**Other options** to handle is to use only `github_state` instead of [`combined_status`](https://github.com/zendesk/samson/blob/mugoyal/fix_key_error/app/models/commit_status.rb#L47), but that way we would suppress concerns such as `Old Release` and `Production Only Reference`.

cc @zendesk/samson, @zendesk/sustaining, @zendesk/bre 

### Tasks
 - [x] Refactor names (as `github` in the name is mis-leading) 
 - [ ] Test on Staging

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-1340
 - Rollbar: https://rollbar-us.zendesk.com/Zendesk/Samson/items/1324/?item_page=0&#instances

### Risks
- Level: Low

### DIFF
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 module ReleasesHelper
-  GITHUB_STATUS_ICONS = {
+  STATUS_ICONS = {
     "success" => "ok",
     "failure" => "remove",
     "missing" => "minus",
@@ -8,7 +8,7 @@ module ReleasesHelper
     "error" => "exclamation-sign"
   }.freeze
 
-  GITHUB_STATUS_TEXT_LABELS = {
+  STATUS_TEXT_LABELS = {
     "success" => "success",
     "failure" => "danger",
     "missing" => "muted",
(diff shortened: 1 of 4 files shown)


goyalmunish:
Hi @ragurney, Hi @grosser 

I see that related code was changes latest by you. Could you have a glimpse on this PR and set the expected behavior?

app/helpers/releases_helper.rb
```diff
@@ -24,9 +26,9 @@ def release_label(project, release)
     )
   end
 
-  def github_commit_status_icon(status_state)
-    icon = GITHUB_STATUS_ICONS.fetch(status_state)
-    text = GITHUB_STATUS_TEXT_LABELS.fetch(status_state)
+  def commit_status_icon(status_state)
+    icon = STATUS_ICONS.fetch(status_state, STATUS_ICONS[DEFAULT_STATUS])
+    text = STATUS_TEXT_LABELS.fetch(status_state, STATUS_TEXT_LABELS[DEFAULT_STATUS])
     title = "Github status: #{status_state}"
```
grosser:
```suggestion
    title = "Commit status: #{status_state}"
```

test/helpers/releases_helper_test.rb
```diff
@@ -15,42 +15,49 @@
...
-      html.must_include "text-primary"
-      html.must_include "Github status: pending"
+    it "renders an icon for error status" do
+      html = commit_status_icon("error")
+      html.must_include "glyphicon-exclamation-sign"
+      html.must_include "text-warning"
+      html.must_include "Github status: error"
+    end
+
+    it "assumes 'error' as default status" do
```
grosser: might be more descriptive to do:
```suggestion
    it "assumes 'error' for 1-off states produced by release_status" do
```

app/helpers/releases_helper.rb
goyalmunish: thanks. I missed it.

test/helpers/releases_helper_test.rb
goyalmunish: sure..

grosser:
@goyalmunish merge ?

goyalmunish:
Hi @grosser, yes it can be merged. But for me "Merge pull request" button is disabled:

image

PR #3207, opened 2019-03-05
fix rake breaking in local development because of RollbarTestController

grosser:
```
rake db:migrate
rake aborted!
RollbarTestController should not be loaded
lib/samson/boot_check.rb:20:in `check'
config/application.rb:153:in `block in <class:Application>'
```

rollbar gem now defines a controller in it's rake tasks, which are loaded because we use the rollbar engine ... going to ignore it instead of hunting it down :(


grosser:
if you see errors, please fix them or let me know, will otherwise go unnoticed like this one :D

PR #3209, opened 2019-03-06
covert kubernetes roles to be based on resources-controller for unifi…

grosser:
…ed behavior

@zendesk/compute


hgokavarapuz:
What is the related PAAS ticket for this change?

hgokavarapuz (approved):
LGTM. would be ideal to link with a PAAS ticket useful for sprint planning.

grosser:
no paas ticket, just something I'm doing for fun to make controllers more generic

PR #3216, opened 2019-03-07
make deploy-group deletion not be 1-off controller logic, but use mod…

grosser:
…el validations

 - controllers should only use "save/destroy/soft_delete" and not know about model validations
 - models should know when they are valid / when they are allowed to be deleted

 - fixed that we should not destroy dependencies before soft_delete hook is started (SoftDeleteWithDestroy bug) 
 - fixed we should not set a permalink before validations are done so redirects do not break and forms do not show weird permalinks

screen shot 2019-03-07 at 1 44 40 pm

### DIFF
--- a/test/controllers/deploy_groups_controller_test.rb
+++ b/test/controllers/deploy_groups_controller_test.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative '../test_helper'
 
-SingleCov.covered! uncovered: 1
+SingleCov.covered!
 
 describe DeployGroupsController do
   let(:deploy_group) { deploy_groups(:pod100) }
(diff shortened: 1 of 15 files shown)


test/controllers/deploy_groups_controller_test.rb
```diff
@@ -75,10 +75,15 @@
 
       it "compares environment" do
         var = EnvironmentVariable.create!(scope: deploy_groups(:pod2), parent: stage.project, name: "Foo", value: "bar")
+        group = EnvironmentVariableGroup.create!(name: "G1")
+        var2 = EnvironmentVariable.create!(scope: deploy_groups(:pod2), parent: group, name: "Bar", value: "baz")
```
grosser: unrelated cleanup since this had missing coveragge

grosser:
... now  it still runs validations when we use `validate: false`, so will need a workaround for that ...

sathishavm (approved):
LGTM

PR #3224, opened 2019-03-10
Add support for a deploy_group usage report

thilgen:
**Use Case**
- I want a report that shows all the `deploy_groups` where all `projects` are deploying code.
- This is partially supported today by going to the `deploy_groups` page and inspecting each individual `deploy_group`, however this is kind of a pain when I am trying to do management over a large number of `projects`, each with a number of `stages`, and each of those going to a variable number of `deploy_groups`.

**Solution**
- Enable a new report option (when `DeployGroup.enabled?`) that generates a CSV containing a list of all `projects`, their `stages`, and information about their `deploy_groups`
- Based on code block previously (and manually) run against the database
```
Project.all.each do |project|
  project.stages.each do |stage|
    stage.deploy_groups.each do |deploy_group|
      puts "#{project.name}\t#{stage.name}\t#{deploy_group.name}\t#{deploy_group.environment_id}\t#{deploy_group.environment.name}\t#{deploy_group.env_value}\t#{deploy_group.permalink}"
    end
  end
  puts "#{project.name}" if project.stages.empty?
end; nil
```
**Screenshots**
image
_Reports screen when DEPLOY_GROUP_FEATURE is not enabled_

image
_Reports screen when DEPLOY_GROUP_FEATURE is enabled_

image
_The New Deploy Group Usage Report screen_

image
_CSC Download generated after clicking on the Download Deploy Group Usage Report button_

image
_Example CSV results_

/cc @zendesk/samson, @grosser, @pduchemin

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Low

### DIFF
--- a/app/presenters/deploy_group_usage_csv_presenter.rb
+++ b/app/presenters/deploy_group_usage_csv_presenter.rb
@@ -32,15 +32,17 @@ class DeployGroupUsageCsvPresenter
   end
 
   def self.csv_line(project, stage = nil, deploy_group = nil)
-    result = []
+    result = [project.name]
     if stage && deploy_group
-      result << stage.name
-      result << deploy_group.name
-      result << deploy_group.environment_id
-      result << deploy_group.environment.name
-      result << deploy_group.env_value
-      result << deploy_group.permalink
+      result.concat [
+        stage.name,
+        deploy_group.name,
+        deploy_group.environment_id,
+        deploy_group.environment.name,
+        deploy_group.env_value,
+        deploy_group.permalink
+      ]
     end
-    result.unshift(project.name)
+    result
   end
 end
--- a/app/presenters/deploy_group_usage_csv_presenter.rb
+++ b/app/presenters/deploy_group_usage_csv_presenter.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class DeployGroupUsageCsvPresenter
+
+  def self.to_csv
+    CSV.generate do |csv|
+      csv << csv_header
+      Project.all.each do |project|
+        if project.stages.empty?
+          csv << csv_line(project)
+        else
+          project.stages.each do |stage|
+            stage.deploy_groups.each do |deploy_group|
+              csv << csv_line(project, stage, deploy_group)
+            end
+          end
+        end
+      end
+    end
+  end
+
+  def self.csv_header
+    [
+      "project_name",
+      "stage_name",
+      "deploy_group_name",
+      "deploy_group_environment_id",
+      "deploy_group_environment_name",
+      "deploy_group_env_value",
+      "deploy_group_permalink"
+    ]
+  end
+
+  def self.csv_line(project, stage=nil, deploy_group=nil)
+    result = []
+    if stage && deploy_group
+      result << stage.name
+      result << deploy_group.name
+      result << deploy_group.environment_id
+      result << deploy_group.environment.name
+      result << deploy_group.env_value
+      result << deploy_group.permalink
+    end
+    return result.unshift(project.name)
+  end
+end
--- a/app/presenters/deploy_group_usage_csv_presenter.rb
+++ b/app/presenters/deploy_group_usage_csv_presenter.rb
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
-
 require 'csv'
 
 class DeployGroupUsageCsvPresenter
-
   def self.to_csv
     CSV.generate do |csv|
       csv << csv_header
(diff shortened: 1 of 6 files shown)


app/presenters/deploy_group_usage_csv_presenter.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+require 'csv'
+
+class DeployGroupUsageCsvPresenter
+  def self.to_csv
```
grosser: use `class << self` and then `def to_csv` and all the other methods can be in `private`

app/presenters/deploy_group_usage_csv_presenter.rb
```diff
@@ -0,0 +1,46 @@
...
+    CSV.generate do |csv|
+      csv << csv_header
+      Project.all.each do |project|
+        if project.stages.empty?
+          csv << csv_line(project)
+        else
+          project.stages.each do |stage|
+            stage.deploy_groups.each do |deploy_group|
+              csv << csv_line(project, stage, deploy_group)
```
grosser: same logic here when a stage has no deploy group ?

app/presenters/deploy_group_usage_csv_presenter.rb
```diff
@@ -0,0 +1,46 @@
...
+      "deploy_group_name",
+      "deploy_group_environment_id",
+      "deploy_group_environment_name",
+      "deploy_group_env_value",
+      "deploy_group_permalink"
+    ]
+  end
+
+  def self.csv_line(project, stage = nil, deploy_group = nil)
+    result = []
```
grosser:
```Ruby
result = [project.name]
if stage && deploy_group
  result.concat [
    stage.name,
    ...
  ]
end
```

app/views/csv_exports/new_deploy_group_usage.erb
```diff
@@ -0,0 +1,16 @@
+<%= page_title "Deploy Group Usage Report" %>
+
+<section>
+
+  <div class="csv-users-filters clearfix">
+    <%= form_tag new_csv_export_path(format: :csv, type: :deploy_group_usage), method: :get, class: "form-horizontal" do %>
+      <%= hidden_field_tag :type, "deploy_group_usage" %>
```
grosser: it's in the url so no need for it here ?

app/presenters/deploy_group_usage_csv_presenter.rb
thilgen: I looked at sorta doing it this way - but the eval of the `if statement` blows away the auto return of `result` - so something like this would be necessary

```
def self.csv_line(project, stage = nil, deploy_group = nil)
  result = [project.name]
  if stage && deploy_group
    result.concat [
      stage.name,
      ...
    ]
  end
  result
end
```

It seemed more clever :D to instead flip the logic around and make the last statement (which will always have a good return value) to shove the always present first value into the array. 

But, I guess I am not really saving any lines here, still need to have the init then the return, so I will make the proposed change.

app/presenters/deploy_group_usage_csv_presenter.rb
thilgen: if `DEPLOY_GROUP_FEATURE` is enabled - Can you have a `stage` without a `deploy_group`? The UI seemed to enforce that relationship...

grosser: ah true!

app/presenters/deploy_group_usage_csv_presenter.rb
thilgen: done

app/views/csv_exports/index.html.erb
```diff
@@ -4,6 +4,9 @@
   <div>
     <p><%= link_to "New Deploys Report", new_csv_export_path %></p>
     <p><%= link_to "New Users Report", new_csv_export_path(type: :users) %></p>
+    <%if DeployGroup.enabled? %>
```
grosser:
```suggestion
    <% if DeployGroup.enabled? %>
```

app/presenters/deploy_group_usage_csv_presenter.rb
thilgen: done

app/views/csv_exports/new_deploy_group_usage.erb
thilgen: When I remove `hidden_field_tag` the csv generation+download fails - seems to be part of the weird html and csv routing in the `csv_exports_controller` that makes this dance work. I poked around a little but couldn't find a way to remove it - however since this is the same pattern employed in `new_users.html.erb` - hopefully not too egregious to leave it as-ss.

app/views/csv_exports/index.html.erb
thilgen: done

PR #3225, opened 2019-03-11
Add baseline metrics

stephennyu:
* Add staging and production deployment lead time metrics
* Add staging to production deployment latency metrics

/cc @zendesk/bre

### References
 - Jira link: https://zendesk.atlassian.net/browse/BRE-1367

### DIFF
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -98,6 +98,10 @@ class Changeset::PullRequest
     users.compact.map { |user| Changeset::GithubUser.new(user) }.uniq
   end
 
+  def created_at
+    @data['created_at']
+  end
+
   def risky?
     risks.present?
   end
--- a/app/models/deploy_metrics.rb
+++ b/app/models/deploy_metrics.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+# Used to collect metrics of cycle time(elapsed time needed to deliver a project to production)
+# Check out https://about.gitlab.com/2019/03/04/reduce-cycle-time/ for more details on cycle time
+# The main objective for this is to gain more insight on the efficiency of the end-to-end project development process
+# There are 2 key metrics:
+# 1. PR - Production: this will show the time needed for a pull request to get deployed to production environment
+# 2. Staging - Production: this will show the time needed to deliver from staging(non-production environment)
+#                          to production environment.
+class DeployMetrics
+  def initialize(deploy)
+    @deploy = deploy
+  end
+
+  def cycle_time
+    return {} unless @deploy&.production && @deploy&.status == "succeeded"
+
+    times = {}
+    if time = pr_production
+      times[:pr_production] = time
+    end
+    if time = staging_production
+      times[:staging_production] = time
+    end
+    times
+  end
+
+  private
+
+  # Calculated by averaging deployment end time minus pull requests' creation time
+  def pr_production
+    prs = @deploy.changeset.pull_requests
+    return nil if prs.empty?
+
+    prs.sum { |pr| @deploy.updated_at.to_i - pr.created_at.to_i } / prs.size
+  end
+
+  # Time it took from first staging deploy to finish until the first production deploy finished
+  def staging_production
+    stages = @deploy.project.stages
+    production, staging = stages.partition(&:production?)
+
+    scope = Deploy.successful.where(jobs: {commit: @deploy.commit}).reorder(:id)
+    return nil unless first_staging_deploy = scope.where(stage: staging).first
+    # should always exists since we have a check in cycle_time method
+    first_production_deploy = scope.where(stage: production).first!
+
+    first_production_deploy.updated_at.to_i - first_staging_deploy.updated_at.to_i
+  end
+end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -157,7 +157,13 @@ class Job < ActiveRecord::Base
   end
 
   def report_state
-    type = deploy ? 'deploy' : 'build'
-    ActiveSupport::Notifications.instrument('job_status.samson', type: type, status: status)
+    payload = {
+      stage: deploy&.stage&.permalink,
+      project: project.permalink,
+      type: deploy ? 'deploy' : 'build',
+      status: status,
+      cycle_time: DeployMetrics.new(deploy).cycle_time
+    }
+    ActiveSupport::Notifications.instrument('job_status.samson', payload)
   end
 end
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -33,7 +33,15 @@ ActiveSupport::Notifications.subscribe("job_queue.samson") do |*, payload|
 end
 
 ActiveSupport::Notifications.subscribe("job_status.samson") do |*, payload|
-  Samson.statsd.increment "jobs.#{payload.fetch(:type)}.#{payload.fetch(:status)}"
+  tags = [
+    "project:#{payload.fetch(:project)}",
+    "stage:#{payload.fetch(:stage)}",
+  ]
+
+  payload.fetch(:cycle_time).each do |key, value|
+    Samson.statsd.histogram "jobs.deploy.cycle_time.#{key}", value, tags: tags
+  end
+  Samson.statsd.increment "jobs.#{payload.fetch(:type)}.#{payload.fetch(:status)}", tags: tags
 end
 
 ActiveSupport::Notifications.subscribe("system_stats.samson") do |*, payload|
--- a/test/models/deploy_metrics_test.rb
+++ b/test/models/deploy_metrics_test.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+describe DeployMetrics do
+  let(:now) { Time.now }
+  let(:project) { projects(:test) }
+  let(:stage) { stages(:test_staging) }
+  let(:user) { users(:deployer) }
+  let(:deploy) { deploys(:succeeded_test) }
+  let(:deploy_metrics) { DeployMetrics.new(deploy) }
+
+  describe "#cycle_time" do
+    describe "production deployment with no staging deployment" do
+      let(:deploy) { deploys(:succeeded_production_test) }
+      let(:deploy_metrics) { DeployMetrics.new(deploy) }
+      before do
+        deploy.updated_at = now + 25
+        deploy.changeset.expects(:pull_requests).
+          returns([stub("commit 1", created_at: now), stub("commit 2", created_at: now + 10)])
+      end
+
+      it "returns pr_production cycle time" do
+        deploy_metrics.cycle_time[:pr_production].must_equal 20
+      end
+
+      it "returns nil for staging_production cycle time" do
+        deploy_metrics.cycle_time[:staging_production].must_equal nil
+      end
+    end
+
+    describe "production deployment with staging deployment" do
+      let(:deploy1) do
+        create_deploy!(reference: deploy.commit, stage: stages(:test_production), updated_at: now + 25)
+      end
+      let(:deploy2) do
+        create_deploy!(reference: deploy1.commit, stage: stages(:test_production), updated_at: now + 50)
+      end
+      let(:deploy_metrics) { DeployMetrics.new(deploy2) }
+      before do
+        deploy2.changeset.expects(:pull_requests).
+          returns([stub("commit 1", created_at: now), stub("commit 2", created_at: now + 10)])
+      end
+
+      it "returns pr_production cycle time" do
+        deploy_metrics.cycle_time[:pr_production].must_equal 45
+      end
+
+      it "returns staging_production cycle time" do
+        expected = deploy1.updated_at.to_i - deploy.updated_at.to_i
+        deploy_metrics.cycle_time[:staging_production].must_equal expected
+      end
+    end
+
+    describe "production deployment with no changes deployed" do
+      let(:deploy) { deploys(:succeeded_production_test) }
+      let(:deploy_metrics) { DeployMetrics.new(deploy) }
+      before do
+        deploy.updated_at = now + 25
+        deploy.changeset.expects(:pull_requests).returns([])
+      end
+
+      it "returns nil for pr_production cycle time" do
+        deploy_metrics.cycle_time[:pr_production].must_equal nil
+      end
+
+      it "returns nil for staging_production cycle time" do
+        deploy_metrics.cycle_time[:staging_production].must_equal nil
+      end
+    end
+
+    describe "production deployment with unsuccessful staging deployment" do
+      let(:deploy) { deploys(:failed_staging_test) }
+      let(:deploy1) do
+        create_deploy!(reference: deploy.commit, stage: stages(:test_production), updated_at: now + 25)
+      end
+      let(:deploy_metrics) { DeployMetrics.new(deploy1) }
+      before do
+        deploy1.changeset.expects(:pull_requests).
+          returns([stub("commit 1", created_at: now), stub("commit 2", created_at: now + 10)])
+      end
+
+      it "returns pr_production cycle time" do
+        deploy_metrics.cycle_time[:pr_production].must_equal 20
+      end
+
+      it "returns nil for staging_production cycle time if no successful staging deployment" do
+        deploy_metrics.cycle_time[:staging_production].must_equal nil
+      end
+    end
+
+    describe "staging deployment" do
+      it "returns {} for cycle time" do
+        deploy_metrics.cycle_time.blank?.must_equal true
+      end
+    end
+  end
+
+  def create_deploy!(attrs = {})
+    default_attrs = {
+      reference: "baz",
+      job: create_job!(commit: attrs[:reference]),
+      project: project
+    }
+
+    deploy_stage = attrs.delete(:stage) || stage
+
+    deploy_stage.deploys.create!(default_attrs.merge(attrs))
+  end
+
+  def create_job!(attrs = {})
+    default_attrs = {
+      project: project,
+      command: "echo hello world",
+      status: "succeeded",
+      user: user
+    }
+
+    Job.create!(default_attrs.merge(attrs))
+  end
+end
--- a/test/models/job_test.rb
+++ b/test/models/job_test.rb
@@ -380,15 +380,27 @@ describe Job do
     let(:job) { jobs(:succeeded_test) }
 
     it 'reports for deploy' do
-      assert_instrument(type: 'deploy', status: 'succeeded')
+      assert_instrument(
+        stage: job&.deploy&.stage&.permalink,
+        project: project.permalink,
+        type: job&.deploy ? 'deploy' : 'build',
+        status: 'succeeded',
+        cycle_time: DeployMetrics.new(job&.deploy).cycle_time
+      )
 
       job.send(:report_state)
     end
 
     it 'reports for build' do
-      assert_instrument(type: 'build', status: 'succeeded')
       job = jobs(:running_test) # job without deploy
       job.update_column(:status, 'succeeded')
+      assert_instrument(
+        stage: job&.deploy&.stage&.permalink,
+        project: project.permalink,
+        type: job&.deploy ? 'deploy' : 'build',
+        status: 'succeeded',
+        cycle_time: DeployMetrics.new(job&.deploy).cycle_time
+      )
 
       job.send(:report_state)
     end
--- a/test/models/changeset/pull_request_test.rb
+++ b/test/models/changeset/pull_request_test.rb
@@ -53,13 +53,15 @@ describe Changeset::PullRequest do
           "head" => {
             "ref" => 'a/ref',
             "sha" => 'abcd123'
-          }
+          },
+          "created_at" => '2019-03-13T00:00:00Z'
         }
       }
       pr = Changeset::PullRequest.changeset_from_webhook(project, params)
       pr.state.must_equal 'open'
       pr.branch.must_equal 'a/ref'
       pr.sha.must_equal 'abcd123'
+      pr.created_at.must_equal '2019-03-13T00:00:00Z'
     end
   end
 


app/models/changeset/pull_request.rb
```diff
@@ -127,6 +127,10 @@ def message
     nil
   end
 
+  def lead_time
+    @data['updated_at'] - @data['created_at']
```
grosser: If I let a PR sit for a week that does not count into lead-times ?
... link to a definition of lead-time too, so explain what this is used for/means

app/models/deploy.rb
```diff
@@ -135,6 +135,27 @@ def duration
     updated_at - start_time
   end
 
+  # time from when PRs were created until get deployed
+  def total_lead_time
+    changeset.pull_requests_lead_time + duration
```
grosser: this math does not make sense to me ... total-lead-time would be `prs.sum { deploy.end - pr.start }`

docker-compose.yml
```diff
@@ -1,7 +1,10 @@
 version: "2"
 services:
   samson:
-    image: zendesk/samson:latest # replace with `build: .` to use Dockerfile (--build to rebuild it)
+    # image: zendesk/samson:latest # replace with `build: .` to use Dockerfile (--build to rebuild it)
+    build: .
```
grosser: don't commit that

docker-compose.yml
```diff
@@ -1,7 +1,10 @@
...
+    tty: true
+    stdin_open: true
```
grosser: why does it need tty/stdin ?

config/initializers/statsd.rb
```diff
@@ -33,7 +33,20 @@ class << Samson
 end
 
 ActiveSupport::Notifications.subscribe("job_status.samson") do |*, payload|
-  Samson.statsd.increment "jobs.#{payload.fetch(:type)}.#{payload.fetch(:status)}"
+  tags = []
+  if payload[:deploy]
```
grosser: avoid duplication
```suggestion
  if deploy = payload[:deploy]
```

config/initializers/statsd.rb
```diff
@@ -33,7 +33,20 @@ class << Samson
...
+    tags.push("project:#{payload[:deploy].project.name}", "stage:#{payload[:deploy].stage.name}")
```
grosser: prefer permalink, it won't have spaces/colons etc things that break tagging

config/initializers/statsd.rb
```diff
@@ -33,7 +33,20 @@ class << Samson
...
+    if ['succeeded', 'failed', 'errored'].include?(payload[:status])
+      total_lead_time = payload[:deploy].total_lead_time
```
grosser: how does lead time make sense if the deploy failed ? ... it's not done yet, so there is no lead-time ?

config/initializers/statsd.rb
```diff
@@ -33,7 +33,20 @@ class << Samson
...
+      if payload[:deploy].production
+        to_production_latency = payload[:deploy].to_production_latency
+        Samson.statsd.histogram "jobs.deploy.to_production_latency", to_production_latency, tags: tags
+      end
+      Samson.statsd.histogram "jobs.deploy.total_lead_time", total_lead_time, tags: tags
+    end
+  end
```
grosser: this whole logic is rather complicated, prefer making a new class and then putting all the logic in there

stephennyu:
so for multiple production stages, we have decided to calculate each stage, send to datadog, and average all out

app/models/deploy_baseline.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+class DeployBaseline
```
grosser: this is some made up math, so explain what is going on in detail and why these metrics were chosen "it's the best we can do / it's what we want because"

grosser: class name is kinda awkward ... DeployMetrics ?
... and call it `cycle time` ... and calculate it from first commit to prod deploy https://about.gitlab.com/2019/03/04/reduce-cycle-time/

app/models/deploy_baseline.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize(deploy)
+    @deploy = deploy
+  end
+
+  def total_latency
+    @total_latency ||= begin
```
grosser: only cache it if that's needed

app/models/deploy_baseline.rb
```diff
@@ -0,0 +1,35 @@
...
+      {
+        pr_production_latency: valid_deployment ? pr_production_latency : 0,
+        staging_production_latency: valid_deployment ? staging_production_latency : 0
+      }
+    end
+  end
+
+  private
+
+  def valid_deployment
```
grosser: inline this to avoid calling it twice

app/models/deploy_baseline.rb
```diff
@@ -0,0 +1,35 @@
...
+    @deploy&.production && @deploy&.status == "succeeded"
+  end
+
+  def pr_production_latency
+    pull_requests = @deploy&.changeset&.pull_requests || []
+    pull_requests.sum { |pr| @deploy.updated_at.to_i - pr.created_at.to_i }
```
grosser: should be the average and not the sum ?
deploy 1 pr -> 1 minute
deploy 10 prs -> 10 minutes ... that seems wrong

app/models/deploy_baseline.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def staging_production_latency
+    jobs = Job.where(commit: @deploy&.commit)
```
grosser:
```suggestion
    jobs = Job.where(commit: @deploy&.commit).pluck(:id)
```

app/models/deploy_baseline.rb
```diff
@@ -0,0 +1,35 @@
...
+    stages = Stage.where(project: @deploy&.project)
+    stages = stages.reject(&:production?)
+
+    first_staging_deploy = Deploy.where(job: jobs, stage: stages).reorder(:id).first
+    first_staging_deploy ? @deploy.updated_at.to_i - first_staging_deploy.created_at.to_i : 0
```
grosser: again,. needs explanation on what this means exactly, but I was thinking if it's time it takes from staging to prod then we should measure "staging done to production done" 

```suggestion
    first_staging_deploy ? @deploy.updated_at.to_i - first_staging_deploy.updated_at.to_i : 0
```

app/models/job.rb
```diff
@@ -157,7 +157,13 @@ def short_reference
   end
 
   def report_state
-    type = deploy ? 'deploy' : 'build'
-    ActiveSupport::Notifications.instrument('job_status.samson', type: type, status: status)
+    payload = {
+      stage: deploy&.stage&.name,
```
grosser:
```suggestion
      stage: deploy&.stage&.permalink,
```

config/initializers/statsd.rb
```diff
@@ -33,7 +33,21 @@ class << Samson
 end
 
 ActiveSupport::Notifications.subscribe("job_status.samson") do |*, payload|
-  Samson.statsd.increment "jobs.#{payload.fetch(:type)}.#{payload.fetch(:status)}"
+  tags = [
+    "project:#{payload.fetch(:project)}",
+    "stage:#{payload.fetch(:stage)}",
+  ]
+  total_latency = payload.fetch(:total_latency)
+  pr_production_latency = total_latency.fetch(:pr_production_latency)
```
grosser: maybe the structure should be `latencies: {pr_production: 123, staging_productiom: 234}`

config/initializers/statsd.rb
```diff
@@ -33,7 +33,21 @@ class << Samson
...
+  staging_production_latency = total_latency.fetch(:staging_production_latency)
+
+  if pr_production_latency > 0
```
grosser: `!= 0` so we see when the math goes wrong ?
... also 0 is a magic value, so better use a constant like "NOT_APPLICABLE" or so ?

test/models/job_test.rb
```diff
@@ -380,15 +380,29 @@ def assert_instrument(with)
...
     it 'reports for deploy' do
-      assert_instrument(type: 'deploy', status: 'succeeded')
+      payload = {
+        stage: job&.deploy&.stage&.name,
+        project: project.permalink,
+        type: job&.deploy ? 'deploy' : 'build',
+        status: 'succeeded',
+        total_latency: DeployBaseline.new(job&.deploy).total_latency
+      }
+      assert_instrument(payload)
```
grosser: assigning the extra variable does not really make this more readable 🤷‍♂️

app/models/deploy_baseline.rb
stephennyu: got it. thanks

app/models/deploy_baseline.rb
stephennyu: yup, should be average

stephennyu:
@grosser please help to review again. Thanks

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,40 @@
...
+      {pr_production: 0, staging_production: 0}
+    end
+  end
+
+  private
+
+  # Calculated by averaging deployment end time minus pull requests' creation time
+  def pr_production
+    prs = @deploy&.changeset&.pull_requests || []
+    prs.any? ? prs.sum { |pr| @deploy.updated_at.to_i - pr.created_at.to_i } / prs.size : 0
```
grosser: FYI careful with integer math in ruby since it can be incorrect `19 / 10 = 1`

grosser: also prefer handling the simple case first: `return 0 if prs.empty?`

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  # Calculated by taking production deployment end time minus the very first staging deployment end time
```
grosser: prefer making the comment human readable instead of repeating the code 1-to-1
`Time it took from first staging deploy to finish until the first production deploy finished`

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,40 @@
...
+    else
+      {pr_production: 0, staging_production: 0}
+    end
+  end
+
+  private
+
+  # Calculated by averaging deployment end time minus pull requests' creation time
+  def pr_production
+    prs = @deploy&.changeset&.pull_requests || []
```
grosser: deploy can never be nil here

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,40 @@
...
+    prs.any? ? prs.sum { |pr| @deploy.updated_at.to_i - pr.created_at.to_i } / prs.size : 0
+  end
+
+  # Calculated by taking production deployment end time minus the very first staging deployment end time
+  def staging_production
+    jobs = Job.where(commit: @deploy&.commit).pluck(:id)
```
grosser: deploy can never be nil here

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,40 @@
...
+    stages = Stage.where(project: @deploy&.project)
+    stages = stages.reject(&:production?)
+
+    first_staging_deploy = Deploy.where(job: jobs, stage: stages).reorder(:id).first
+    first_staging_deploy ? @deploy.updated_at.to_i - first_staging_deploy.updated_at.to_i : 0
```
grosser: we can only count the first production deploy, so need to do the same logic as the first_staging_deploy but `first_production_deploy `

config/initializers/statsd.rb
```diff
@@ -5,6 +5,8 @@ class << Samson
   attr_accessor :statsd
 end
 
+NOT_APPLICABLE = 0
```
grosser: should use that inside of DeployMetrics so they are linked together via this constant

app/models/deploy_metrics.rb
stephennyu:
```
  def report_state
    payload = {
      stage: deploy&.stage&.permalink,
      project: project.permalink,
      type: deploy ? 'deploy' : 'build',
      status: status,
      deploy_metrics: DeployMetrics.new(deploy)
    }
    ActiveSupport::Notifications.instrument('job_status.samson', payload)
  end
```

I thought on this part `deploy` can be nil ? (e.g. build)

app/models/deploy_metrics.rb
stephennyu: alright, agree

app/models/deploy_metrics.rb
grosser: yes, but it won't get into this method while still being nil

stephennyu: oh, got your point. will update. thanks

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def initialize(deploy)
+    @deploy = deploy
+  end
+
+  def cycle_time
+    if @deploy&.production && @deploy&.status == "succeeded"
+      {pr_production: pr_production, staging_production: staging_production}
+    else
+      {pr_production: NOT_APPLICABLE, staging_production: NOT_APPLICABLE}
```
grosser: get rid of NOT_APPLICABLE and return {} here ... then for the reporting just report each key/value ... to get less repetition/code that needs to be in sync all the time
```suggestion
      {}
```

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  private
+
+  # Calculated by averaging deployment end time minus pull requests' creation time
+  def pr_production
+    prs = @deploy.changeset.pull_requests || []
```
grosser: is pull_requests ever nil ? -> remove `|| []` ... and if not, make pull_requests have the `|| []`

app/models/deploy_metrics.rb
grosser: return `{}` here and when reporting iterate all keys, so we can avoid having to keep things in sync

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,49 @@
...
+    return NOT_APPLICABLE if prs.empty?
+
+    prs.sum { |pr| @deploy.updated_at.to_i - pr.created_at.to_i } / prs.size
+  end
+
+  # Time it took from first staging deploy to finish until the first production deploy finished
+  def staging_production
+    jobs = Job.where(commit: @deploy.commit).pluck(:id)
+    stages = Stage.where(project: @deploy.project)
```
grosser:
```suggestion
    stages = @deploy.project.stages
```

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,49 @@
...
+  def pr_production
+    prs = @deploy.changeset.pull_requests || []
+    return NOT_APPLICABLE if prs.empty?
+
+    prs.sum { |pr| @deploy.updated_at.to_i - pr.created_at.to_i } / prs.size
+  end
+
+  # Time it took from first staging deploy to finish until the first production deploy finished
+  def staging_production
+    jobs = Job.where(commit: @deploy.commit).pluck(:id)
```
grosser: `return if jobs.size == 1` <- if this was only deployed once we don't need to do all this math

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,49 @@
...
+    stages = Stage.where(project: @deploy.project)
+
+    staging = stages.reject(&:production?)
```
grosser: production, staging = stages.partition(&:production?)

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,49 @@
...
+    production = stages - staging
+
+    first_staging_deploy = Deploy.where(job: jobs, stage: staging).reorder(:id).first
```
grosser: - needs to only count success
 - might as well match on the commit since we have to join either way

```Ruby
scope = Deploy.successful.where(jobs: {commit: @deploy.commit}).reorder(:id)
first_staging_deploy = scope.where(stage: staging).first
first_production_deploy = scope.where(stage: production).first
```

config/initializers/statsd.rb
```diff
@@ -33,7 +33,21 @@ class << Samson
...
-  Samson.statsd.increment "jobs.#{payload.fetch(:type)}.#{payload.fetch(:status)}"
+  tags = [
+    "project:#{payload.fetch(:project)}",
+    "stage:#{payload.fetch(:stage)}",
+  ]
+  cycle_time = payload.fetch(:cycle_time)
+  pr_production = cycle_time.fetch(:pr_production)
+  staging_production = cycle_time.fetch(:staging_production)
+
+  if pr_production != DeployMetrics::NOT_APPLICABLE
```
grosser:
```Ruby
payload.fetch(:cycle_time).each do |key, value|
  Samson.statsd.histogram "jobs.deploy.cycle_time.#{key}", value, tags: tags if value != DeployMetrics::NOT_APPLICABLE
end
```

app/models/deploy_metrics.rb
stephennyu: thanks! awesome

app/models/deploy_metrics.rb
stephennyu: just `first_production_deploy = scope.where(stage: production).first` is enough maybe ? since if production deployment is unsuccessful this line will never be reached

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  # Time it took from first staging deploy to finish until the first production deploy finished
+  def staging_production
+    stages = @deploy.project.stages
+    production, staging = stages.partition(&:production?)
+
+    scope = Deploy.successful.where(jobs: {commit: @deploy.commit}).reorder(:id)
+    return NOT_APPLICABLE unless first_staging_deploy = scope.where(stage: staging).first
+    first_production_deploy = scope.where(stage: production).first
```
grosser: need this or it can be `nil.updated_at` below
```suggestion
    return NOT_APPLICABLE unless first_production_deploy = scope.where(stage: production).first
```

stephennyu: if there is no successful production deployment, it seems to me that this line will never be reached ?

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,44 @@
...
+  NOT_APPLICABLE = 0
+
+  def initialize(deploy)
+    @deploy = deploy
+  end
+
+  def cycle_time
+    return {} unless @deploy&.production && @deploy&.status == "succeeded"
+
+    {pr_production: pr_production, staging_production: staging_production}
```
grosser: atm there are 2 ways of excluding a metric, either by not setting the key or by setting it to NOT_APPLICABLE
ideally there should be only 1

so how about:
```ruby
times = {}
if time = pr_production
  times[:pr_production] = time
end
if time = staging_production
  times[:staging_production] = time
end
times
```

and inside the methods just do `return unless foo` so the methods return `nil`

app/models/deploy_metrics.rb
grosser: `first_production_deploy = scope.where(stage: production).first` does not return oO

stephennyu: sorry for the confusion, I'm referring to this line
`return {} unless @deploy&.production && @deploy&.status == "succeeded"`

so I assume this `scope` must already have at least 1 production deploy with succeeded status ?

app/models/deploy_metrics.rb
stephennyu: okay sure

app/models/deploy_metrics.rb
grosser: ah true!

grosser: maybe use `.first! # should always be there since we check for that in foo method`

stephennyu:
@grosser any idea about this https://travis-ci.org/zendesk/samson/jobs/506084439 ?

app/models/deploy_metrics.rb
```diff
@@ -0,0 +1,51 @@
...
+
+# Used to collect metrics of cycle time(elapsed time needed to deliver a project to production)
+# Check out https://about.gitlab.com/2019/03/04/reduce-cycle-time/ for more details on cycle time
+# The main objective for this is to gain more insight on the efficiency of the end-to-end project development process
+# There are 2 key metrics:
+# 1. PR - Production: this will show the time needed for a pull request to get deployed to production environment
+# 2. Staging - Production: this will show the time needed to deliver from staging(non-production environment)
+#                          to production environment.
+class DeployMetrics
+  NOT_APPLICABLE = 0
```
grosser: don't need this anymore now :)

config/initializers/statsd.rb
```diff
@@ -33,7 +33,15 @@ class << Samson
...
 
 ActiveSupport::Notifications.subscribe("job_status.samson") do |*, payload|
-  Samson.statsd.increment "jobs.#{payload.fetch(:type)}.#{payload.fetch(:status)}"
+  tags = [
+    "project:#{payload.fetch(:project)}",
+    "stage:#{payload.fetch(:stage)}",
+  ]
+
+  payload.fetch(:cycle_time).each do |key, value|
+    Samson.statsd.histogram "jobs.deploy.cycle_time.#{key}", value, tags: tags if value != DeployMetrics::NOT_APPLICABLE
```
grosser:
```suggestion
    Samson.statsd.histogram "jobs.deploy.cycle_time.#{key}", value, tags: tags
```

grosser:
ignore the rails failure, that's another JIRA

test/models/deploy_metrics_test.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+describe DeployMetrics do
+  include GitRepoTestHelper
```
grosser: doubt this is needed

test/models/deploy_metrics_test.rb
```diff
@@ -0,0 +1,124 @@
...
+    describe "production deployment with no staging deployment" do
+      let(:deploy) { deploys(:succeeded_production_test) }
+      let(:deploy_metrics) { DeployMetrics.new(deploy) }
+      before do
+        deploy.updated_at = date_time_now + 25
+        deploy.changeset.expects(:pull_requests).
+          returns([stub("commit 1", created_at: date_time_now), stub("commit 2", created_at: date_time_now + 10)])
+      end
+
+      it "must return pr_production cycle time" do
```
grosser: can simplify this to:
```suggestion
      it "returns pr_production cycle time" do
```

test/models/deploy_metrics_test.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+describe DeployMetrics do
+  include GitRepoTestHelper
+
+  let(:date_time_now) { Time.now }
```
grosser:
```suggestion
  let(:now) { Time.now }
```

test/models/deploy_metrics_test.rb
stephennyu: right. thanks

PR #3227, opened 2019-03-11
GitLab integration

isharshun:
* Add GitLab plugin
* Add Changeset support for GitLab projects as first part of plugin

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low


app/models/changeset.rb
```diff
@@ -71,11 +72,20 @@ def find_comparison
...
+          end
       end
 
       Rails.cache.fetch(cache_key) do
-        GITHUB.compare(repo, previous_commit, commit)
+        if project.gitlab?
+          Presenters::GitlabChangeset.new(Gitlab.compare(repo, previous_commit, commit)).build
+        else
+          GITHUB.compare(repo, previous_commit, commit)
+        end
```
grosser: ideally these 2 would be hooks, but good enough for now ...

plugins/gitlab/README.md
```diff
@@ -0,0 +1 @@
+GitLab integration.
```
grosser: Maybe say a bit more like "Allows deploying gitlab projects, viewing their diff, blah blah"

plugins/gitlab/app/models/presenters/gitlab_commit.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module Presenters
```
grosser: would prefer `Gitlab::CommitPresenter`

grosser:
this looks pretty clean, did not think that was possible without rewiring big parts of the app ...

got some UI screenshots too ?

isharshun:
UI is the same as for GitHub projects.
image
image

Sample error message for the GitLab API:
image

.env.test
```diff
@@ -27,3 +27,4 @@ DOCKER_REGISTRIES=docker-registry.example.com
 DEPRECATED_TOKEN_AUTH=1
 GCLOUD_GKE_CLUSTERS_FOLDER=/tmp
 CACHE_STORE=memory
+GITLAB_TOKEN=
```
grosser: newline after :D

grosser:
let me know when you are happy with it and I'll merge ❤️

isharshun:
Thanks, I`m happy this to be merged to master!

grosser:
thx so much ... already 2+ failed PRs for this :)

grosser:
For more integrations we should have a dedicated remote_repository on the project and then plug in whichever we need, hooks are kinda the wrong tool for the job (then we can also support "no remote = local" and have fallback logic there)

PR #3233, opened 2019-03-12
Environment variables for project stages

sathishavm:
User can add environment variables specific to each stage in project, This allows multiple configurations for the same deploy group.
Stage environment-variables can add both in projects and stages edit pages.

**Projects Edit Page**

Screen Shot 2019-03-11 at 11 20 18 PM

**Stages Edit Page**

Screen Shot 2019-03-11 at 11 21 41 PM

/cc @zendesk/samson

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -72,10 +72,7 @@ class ProjectsController < ResourceController
   end
 
   def allowed_includes
-    [
-      :environment_variable_groups,
-      :environment_variables_with_scope,
-    ]
+    Samson::Hooks.fire(:project_allowed_includes).flatten
   end
 
   def resource_params
--- a/app/models/concerns/group_scope.rb
+++ b/app/models/concerns/group_scope.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module GroupScope
   def self.included(base)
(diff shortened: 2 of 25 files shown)


grosser (changes requested):
- should only be editable in a single place
 - afaik edit via stage-edit would be simpler to implement and not require adding stage logic to so many places

grosser (commented):
use https://github.com/zendesk/samson/pull/2666 as template, it's much simpler / needs fewer changes (no need for a new plugin, maybe just put the env var stuff into env plugin directly too)

sathishavm:
The stage env logic is only in env plugin, just like deploy and project env, rather than creating a new plugin.
The reason we have the stage env both in the project edit page and stage edit page, so that it is more reliable and the user has more understanding about the priority and changes to the env.
The deploy_env plugin is much more simpler, as we can't able to edit the environment variables once the deploy is finished.
Along with this, I have converted the usage of plugins inside the application to hooks.

grosser:
Yes, it should not be a new plugin 👍 
Having multiple edit UIs for the same data does not sound "more reliable" to me, but confusing / error-prone. Let's do the "plugin conversion" and "deploy env -> env" first, so this will be a smaller PR, I'll give it a try and see if it's simple.

grosser:
https://github.com/zendesk/samson/pull/3234 works and is dead-simple ... let's do that instead

PR #3234, opened 2019-03-12
Stage env

grosser:
Screen Shot 2019-03-12 at 12 12 39 PM
Screen Shot 2019-03-12 at 12 12 18 PM

### DIFF
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -174,7 +174,7 @@ module Samson
       def only_callbacks_for_plugin(plugin_name, hook_name)
         original_hooks = @hooks[hook_name]
         @hooks[hook_name] = @hooks[hook_name].select do |proc|
-          proc.source_location.first.include?(plugin_name)
+          proc.source_location.first.include?("/#{plugin_name}/")
         end
         yield
       ensure
--- a/plugins/env/app/views/samson_env/_stage_form.html.erb
+++ b/plugins/env/app/views/samson_env/_stage_form.html.erb
@@ -1,0 +1,8 @@
+<fieldset>
+  <legend>
+    Extra Environment Variables for this stage <%= additional_info SamsonEnv::HELP_TEXT %>
+  </legend>
+
+  <%= render "samson_env/environment_variables_fields",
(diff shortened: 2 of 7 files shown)


sathishavm (commented):
Can you add some test cases??

grosser:
coming, was just a POC to see if you like the approach :)

grosser:
done!

sathishavm (approved):

hgokavarapuz (approved):

lib/samson/hooks.rb
```diff
@@ -174,7 +174,7 @@ def with_callback(name, *hook_blocks)
       def only_callbacks_for_plugin(plugin_name, hook_name)
         original_hooks = @hooks[hook_name]
         @hooks[hook_name] = @hooks[hook_name].select do |proc|
-          proc.source_location.first.include?(plugin_name)
+          proc.source_location.first.include?("/#{plugin_name}/")
```
adammw: explain?

plugins/env/app/views/samson_env/_stage_form.html.erb
```diff
@@ -0,0 +1,8 @@
+<fieldset>
+  <legend>
```
adammw: Probably for another PR, but would be nice to have a collapsable read-only view of the globally-set and deploy group-set environment variables here too for visibility.

lib/samson/hooks.rb
grosser: using exact matching since we had multiple plugins that included the letters "env"

plugins/env/app/views/samson_env/_stage_form.html.erb
grosser: make a BRE jira ;)

lib/samson/hooks.rb
adammw: sorry, for some reason thought this was a regex not a string, so you're filtering by directory name...

plugins/env/app/views/samson_env/_stage_form.html.erb
adammw: 😿

PR #3237, opened 2019-03-12
run build validations via api since a missing checkout would take too…

### LINKED ISSUE #3217: Registering build fails when cloning takes more then 5 seconds
Relates to #3198 but did not know exactly what the problem was when I created that issue.

How to reproduce;
Find a "big" repo, one that takes 10 or so seconds to clone. Then register a build when the repo is not in the cache. The request will hang for about 6 seconds (on my machine) then return with a 422 "not a valid git reference".

I don't know how to solve this problem in a good way, but I think a 5 second timeout is a bit small in this case. Something like 20-30 seconds timeout would be better and then fail with a better error message than the very confusing "not a valid git reference".

grosser:
… long to download

alternative for https://github.com/zendesk/samson/pull/3228
fixes #3217

 - ideally we'd check if the local one is mirrored and then use local/remote ... but that's a race-condition with `GitRepository#exclusive` and could still take significant time
 - ideally this should also work for gitlab or a local checkout ... but not as important
 - cleans up logic a fair bit and makes sure that set sha is also valid when ref was given

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -98,25 +98,25 @@ class Build < ActiveRecord::Base
   end
 
   def validate_git_reference
-    if git_ref.blank? && git_sha.blank?
-      errors.add(:git_ref, 'must be specified')
-      return
-    end
-
+    return errors.add(:git_ref, 'must be specified') if git_ref.blank? && git_sha.blank?
     return if errors.include?(:git_ref) || errors.include?(:git_sha)
+    return validate_git_sha if git_ref.blank?
+    commit = commit_from_ref(git_ref)
+    return errors.add(:git_ref, 'is not a valid reference') unless commit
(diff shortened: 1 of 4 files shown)


app/models/build.rb
```diff
@@ -106,19 +106,25 @@ def validate_git_reference
     return if errors.include?(:git_ref) || errors.include?(:git_sha)
 
     if git_ref.present?
-      commit = project.repository.commit_from_ref(git_ref)
-      if commit
-        self.git_sha = commit unless git_sha.present?
+      if commit = commit_from_ref(git_ref)
+        self.git_sha = commit if git_sha.blank?
```
grosser: this is actually wrong, we need to validate that the current commit exists like we do in the elsif

PR #3240, opened 2019-03-13
Update GitHub status API

hdt91:
GitHub deprecated its old status page API and probably moved to Atlassian product. New status API is here: 
https://www.githubstatus.com/api#status

/cc @zendesk/samson @kiranprakash154 @ben @lennardseah 

### Tasks
 - Update the default URL and status check field

### Risks
- Level: Low

### DIFF
--- a/plugins/github/lib/samson_github/samson_plugin.rb
+++ b/plugins/github/lib/samson_github/samson_plugin.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 module SamsonGithub
-  STATUS_URL = ENV["GITHUB_STATUS_URL"] || 'https://status.github.com'
+  # Reference: https://www.githubstatus.com/api#status
+  # Official doc example is: https://kctbh9vrtdwd.statuspage.io',
+  # however 'https://www.githubstatus.com' also works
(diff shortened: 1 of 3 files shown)


plugins/github/lib/samson_github/samson_plugin.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 module SamsonGithub
-  STATUS_URL = ENV["GITHUB_STATUS_URL"] || 'https://status.github.com'
+  # Reference: https://www.githubstatus.com/api#status
+  # Official doc example is: https://kctbh9vrtdwd.statuspage.io',
+  # however 'https://www.githubstatus.com' also works
+  STATUS_URL = ENV["GITHUB_STATUS_URL"] || 'https://www.githubstatus.com'
```
kiranprakash154: do we know what ENV["GITHUB_STATUS_URL"] value is initialized with?

hdt91: That can be set during deploy time, for example chef or your script runner to start the server. Nevertheless, the default behaviour still needs to get fixed.

PR #3243, opened 2019-03-15
Fix for projects search with url params

sathishavm:
Allow a user to search projects with different URL formats, more often application like 'GCB' supports different URL formats to clone the application, following that I made a change to accept all standard URL's to clone the application.

/cc @zendesk/samson

### Risks
- Level: Low

### DIFF
--- a/test/controllers/projects_controller_test.rb
+++ b/test/controllers/projects_controller_test.rb
@@ -41,18 +41,9 @@ describe ProjectsController do
       end
 
       describe "search" do
-        before do
-          Project.create!(name: "https_url", repository_url: "https://github.com/foo/bar.git")
-        end
-
         it "can search via query" do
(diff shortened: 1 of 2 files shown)


test/controllers/projects_controller_test.rb
```diff
@@ -41,28 +41,46 @@ def fields_disabled?
...
-          get :index, params: {search: {url: "https://github.com/test.git"}}
-          assigns(:projects).map(&:name).must_be_empty
+        describe "via url" do
+          def validate_search_url(url, result)
+            get :index, params: {search: {url: url}}
+            assigns(:projects).map(&:name).must_equal result
+          end
+
+          before do
+            Project.create!(name: "https_url", repository_url: "https://github.com/foo/bar.git")
```
grosser: prefer changing the existing fixture to keep tests fast

PR #3244, opened 2019-03-15
Upgrade rails to '5.2.2.1'

sathishavm:
Fix for security. alerts.

https://github.com/zendesk/samson/network/alert/Gemfile.lock/railties/open
https://github.com/zendesk/samson/network/alert/Gemfile.lock/actionview/open

/cc @zendesk/samson


grosser:
prefer `bundle update rails activerecord actionview activemodel activesupport actionpack actionmailer railties --conservative` which will not bump unrelated gems, so this change is safe and easy to revert

PR #3247, opened 2019-03-16
stop blowing up on bad slack messages and only accept correctly forma…

grosser:
…tted messages

fixes https://github.com/zendesk/samson/pull/3219

overall the code in this plugin seems questionable ... so refactor/cleanup/coverage welcome!

@zendesk/bre @nayena

### DIFF
--- a/plugins/slack_app/app/controllers/slack_app_controller.rb
+++ b/plugins/slack_app/app/controllers/slack_app_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 class SlackAppController < ApplicationController
   PUBLIC = [:command, :interact].freeze
+
   skip_before_action :verify_authenticity_token, :login_user, only: PUBLIC
 
   before_action :handle_slack_inputs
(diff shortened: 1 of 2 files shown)


plugins/slack_app/app/controllers/slack_app_controller.rb
```diff
@@ -41,32 +42,46 @@ def interact
...
+    #   "text"=>"foobar/master to production", <-- deploy project foobar, branch master to stage production
     #   "response_url"=>"https://hooks.slack.com/commands/T0HAGP0J2/58604540277/g0xd4K2KOsgL9zXwR4kEc0eL"
     #    ^^^^ This is how to respond later, we can also directly return some JSON
     # }
-    return unknown_user unless @current_user_from_slack
-
-    # Parse the command
-    project_name, ref_name, stage_name = @payload['text'].match(/([^\s\/]+)(?:\/(\S+))?\s*(?:to\s+(.*))?/).captures
+    project_permalink, ref_name, stage_permalink =
+      @payload['text'].match(/^([^\s\/]+)(?:\/(\S+))?(?: to (\S+))?$/)&.captures
```
nayena: Instead of this funky regexp, why not simply split on a space character and use the first 3 results? That way, any git-strategy (including git-flow) with multiple slashes can be supported as well.

grosser: - this supports multiple slashes
 - I'm trying to keep the api the same, if we want to change the api, then that's another PR

PR #3248, opened 2019-03-17
validate permalinks and do not override user supplied values

grosser:
before creating a project with permalink via api just overrode the permalink and we did not validate when changing existing

for https://github.com/zendesk/samson/issues/3172

@zendesk/bre @danihodovic


danihodovic (approved):
Good stuff :+1:

sathishavm (approved):
LGTM

PR #3252, opened 2019-03-18
unify successful and succeeded

grosser:
it was confusing to use different words for the same meaning

@zendesk/bre 

### Risks
 - low: stage#show with last_successful_deploy will fail, but the error should be obvious


sathishavm (approved):
This looks good for me, but please notify to the channel after deploying about this change, we may missed something???

PR #3258, opened 2019-03-19
stop deploys when bad events happen directly on the deployed resource

grosser:
this works now, I just don't know any way how to create a bad event ... 
jira had daemonset as an example, but that was on a pod too ... and the others I found are on replicaset which requires another change to fetch the rs from each deployment :/
parking this here for now and extracting some parts into smaller PRs

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -14,16 +14,25 @@ module Kubernetes
     TICK = Integer(ENV.fetch('KUBERNETES_STABILITY_CHECK_TICK', 2.seconds))
     RESTARTED = "Restarted"
 
-    # TODO: this logic might be able to go directly into Pod, which would simplify the code here a bit
-    class ReleaseStatus
-      attr_reader :live, :stop, :details, :pod, :role, :group
-      def initialize(stop: false, live:, details:, pod:, role:, group:)
+    class ResourceStatus
+      attr_reader :live, :stop, :details, :role, :group, :resource
+      def initialize(stop: false, live:, details:, resource:, pod:, role:, group:)
         @live = live
         @stop = stop
         @details = details
         @pod = pod
+        @resource = resource
         @role = role
         @group = group
+      end
+
+      def pod?
+        @pod
+      end
+
+      # TODO: use resource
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/event_reader.rb
```diff
@@ -0,0 +1,22 @@
...
+        SamsonKubernetes.retry_on_connection_errors do
+          selector = ["involvedObject.name=#{resource.dig_fetch(:metadata, :name)}"]
+
+          # do not query for nil uid ... rather show events for old+new resource when creation failed
+          if uid = resource.dig_fetch(:metadata, :uid)
+            selector << "involvedObject.uid=#{uid}"
+          end
+
+          client.get_events(
+            namespace: resource.dig(:metadata, :namespace),
```
adammw: ideally we never change namespace or this will break.

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -234,48 +239,64 @@ def print_events(events)
...
+    def resource_statuses(pods, release_doc)
+      deploy_group = release_doc.deploy_group
       role = release_doc.kubernetes_role
 
-      pods = pods.select { |pod| pod.role_id == role.id && pod.deploy_group_id == group.id }
-
-      statuses = Array.new(release_doc.desired_pod_count).each_with_index.map do |_, i|
-        pod = pods[i]
+      # monitor each resource we deploy directly
+      resource_statuses = release_doc.resources.map do |resource|
```
adammw: can we split up this method, a separate one for monitoring directly deployed resources, and monitoring pods we expect to appear?

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -111,7 +116,7 @@ def wait_for_resources_to_complete(release_docs, timeout)
         else
           if too_many_not_ready
             print_statuses(statuses)
-            unstable!('one or more pods are not live', not_ready_statuses)
+            unstable!('resources failed', not_ready_statuses)
```
adammw: is the reason useful anymore? maybe get rid of it

PR #3259, opened 2019-03-19
fix resource events being shown from previous deploys

grosser:
also benchmarked using `involvedObject.uid` but it's the same speed, so no reason for this complexity

extracting bits from https://github.com/zendesk/samson/pull/3258 which was getting too large

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -178,7 +178,7 @@ module Kubernetes
     end
 
     def pod_identifier(pod)
-      "#{deploy_group_for_pod(pod).name} pod #{pod.name}"
+      "#{deploy_group_for_pod(pod).name} Pod #{pod.name}"
     end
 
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -197,22 +197,28 @@ def print_pod_logs(pod, end_time)
...
+          field_selector: "involvedObject.name=#{resource.name},involvedObject.kind=#{resource.kind}"
         ).fetch(:items)
+
+        # ignore events from before the deploy, comparing strings for speed
+        events.select! { |e| e.dig(:metadata, :creationTimestamp) >= @deploy_start_time }
+
+        # https://github.com/kubernetes/kubernetes/issues/29838
+        events.sort_by! { |e| e.dig(:metadata, :creationTimestamp) }
+
         next if events.none?
```
eatwithforks: put between events.select!

eatwithforks (approved):

PR #3260, opened 2019-03-20
unify event fetching logic

grosser:
- de-spaghetti deploy-executor
 - show all pod events even if the pod restarted
 - move logic from pod.rb -> resource_status so it can get killed someday
 - move autoscaling pod logic into a single place

breaking out improvements from https://github.com/zendesk/samson/pull/3258 which was getting too large

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -5,6 +5,8 @@ module Kubernetes
       INIT_CONTAINER_KEY = :'pod.beta.kubernetes.io/init-containers'
       INGORED_AUTOSCALE_EVENT_REASONS = %w[FailedGetMetrics FailedRescale].freeze
 
+      attr_writer :events
+
       def self.init_containers(pod)
         containers = pod.dig(:spec, :initContainers) || []
         if json = pod.dig(:metadata, :annotations, Kubernetes::Api::Pod::INIT_CONTAINER_KEY)
@@ -16,18 +18,6 @@ module Kubernetes
       def initialize(api_pod, client: nil)
         @pod = api_pod
         @client = client
-      end
-
-      def name
-        @pod.dig(:metadata, :name)
-      end
-
-      def namespace
-        @pod.dig(:metadata, :namespace)
-      end
-
-      def annotations
-        @pod[:metadata][:annotations] ||= {}
       end
 
       def live?
@@ -59,16 +49,8 @@ module Kubernetes
         reasons.reject(&:blank?).uniq.join("/").presence || "Unknown"
       end
 
-      def deploy_group_id
-        Integer(labels.fetch(:deploy_group_id))
-      end
-
-      def role_id
-        Integer(labels.fetch(:role_id))
-      end
-
-      def containers
-        @pod.dig(:spec, :containers)
+      def container_names
+        (@pod.dig(:spec, :containers) + self.class.init_containers(@pod)).map { |c| c.fetch(:name) }.uniq
       end
 
       # tries to get logs from current or previous pod depending on if it restarted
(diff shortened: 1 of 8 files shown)


plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -59,16 +49,8 @@ def reason
...
-      end
-
-      def role_id
-        Integer(labels.fetch(:role_id))
-      end
-
-      def containers
-        @pod.dig(:spec, :containers)
+      def container_names
+        (@pod.dig(:spec, :containers) + self.class.init_containers(@pod)).map { |c| c.fetch(:name) }.uniq
```
adammw: why is init_containers not a method on the instance?

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -101,7 +88,7 @@ def wait_for_resources_to_complete(release_docs, timeout)
               @output.puts "TIMEOUT, pods took too long to get live"
               return false, statuses
             end
-          elsif ready_statuses.all? { |s| s.pod.completed? }
+          elsif ready_statuses.all?(&:stop)
```
adammw: what does stop mean? can you use something more descriptive?

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -154,9 +142,9 @@ def show_logs_on_deploy_if_requested(statuses)
     def show_failure_cause(release_docs, statuses)
       release_docs.each { |doc| print_resource_events(doc) }
       log_end_time = Integer(ENV['KUBERNETES_LOG_TIMEOUT'] || '20').seconds.from_now
-      debug_pods = statuses.reject(&:live).select(&:pod).group_by(&:role).map { |_, g| g.first.pod }
-      debug_pods.each do |pod|
-        print_pod_details(pod, log_end_time, events: true)
+      debug_statuses = statuses.reject(&:live).select(&:resource).group_by(&:role).each_value.map(&:first)
+      debug_statuses.each do |status|
+        print_pod_details(status, log_end_time, events: true)
```
adammw: are these resources guaranteed to be pods?

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -167,25 +155,25 @@ def show_failure_cause(release_docs, statuses)
...
       end
-      print_pod_logs(pod, log_end_time)
+      print_pod_logs(status.pod, log_end_time)
       @output.puts "\n------------------------------------------\n"
     end
 
-    def pod_identifier(pod)
-      "#{deploy_group_for_pod(pod).name} Pod #{pod.name}"
+    def pod_identifier(status)
+      "#{status.deploy_group.name} #{status.kind} #{status.resource&.dig(:metadata, :name)}"
```
adammw: extract out `status.resource&.dig(:metadata, :name)` into a variable

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -167,25 +155,25 @@ def show_failure_cause(release_docs, statuses)
...
     end
 
     # show why container failed to boot
     def print_pod_logs(pod, end_time)
       @output.puts "LOGS:"
 
-      containers_names = (pod.containers + pod.init_containers).map { |c| c.fetch(:name) }.uniq
+      containers_names = pod.container_names
```
adammw: can do inline in the each now...

adammw (approved):

plugins/kubernetes/app/models/kubernetes/api/pod.rb
grosser: used from outside too

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
grosser: yes

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
grosser: no, used below too

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
grosser: changed to `finished`

grosser:
💇‍♂️

PR #3261, opened 2019-03-21
PAAS-2990 display history for vault secrets using kv v2

grosser:
we want to know who changed what when ... kv v2 allows that 🎉 
using https://www.vaultproject.io/api/secret/kv/kv-v2.html#read-secret-metadata
adding a fake ActiveRecord backend too so it's easier to debug locally

@zendesk/bre @zendesk/compute


adammw:
Does it really have a grey background like that? Where's the typical white?

grosser:
added a jira for that, would still like to merge this as a stop-gap

grosser:
wrong PR

grosser:
Screen Shot 2019-03-21 at 10 09 12 AM

app/views/secrets/show.html.erb
```diff
@@ -69,6 +69,9 @@
           <%= submit_tag ResourceController::ADD_MORE, class: 'btn btn-default' %>
           <% if id %>
             <%= link_to "Duplicates", secrets_path(search: {value_from: id}) %>
+            <% if Samson::Secrets::Manager.backend == Samson::Secrets::HashicorpVaultBackend %>
```
adammw: Need to check for vault v2 / history support?

lib/samson/secrets/db_backend.rb
```diff
@@ -21,6 +21,19 @@ def read(id)
           secret_to_hash secret
         end
 
+        # Not implemented, just bogus values to be able to debug UI in development+test
+        def history(_id)
```
adammw: shouldn't this be stubbed in test? looks strange if this secret backend otherwise works to add dummy data here.

app/views/secrets/show.html.erb
grosser: killing that soon, so not adding extra steps now

lib/samson/secrets/db_backend.rb
grosser: I'm using this for development / styling, where I don't have stubs

PR #3268, opened 2019-03-23
prefer to debug resources that failed before resources that are still…

grosser:
… running

A deploy with multiple pods, with 1 of them running and 1 failed should show logs+events from the failed one

@zendesk/compute


adammw (approved):
no tests to update?

grosser:
just a list of things we ignore, no new logic <-> no new tests 🤷‍♂️

Issue #3269, opened 2019-03-23
500 errors on deploy group update


danihodovic:
```
Started PATCH "/deploy_groups/382f0cd6-9fde-5113-9990-cf7b98058872" for 127.0.0.1 at 2019-03-23 15:06:15 +0000
Processing by DeployGroupsController#update as HTML
  Parameters: {"deploy_group"=>{"permalink"=>"382f0cd6-9fde-5113-9990-cf7b98058872", "environment_id"=>5, "name_sortable"=>"dg 00382f00000cd00006-00009fde-05113-09990-cf00007b98058872", "env_value"=>"[FILTERED]", "vault_server_id"=>nil, "name"=>"changed 382f0cd6-9fde-5113-9990-cf7b98058872"}, "id"=>"382f0cd6-9fde-5113-9990-cf7b98058872"}
  Doorkeeper::AccessToken Load (0.1ms)  SELECT  "oauth_access_tokens".* FROM "oauth_access_tokens" WHERE "oauth_access_tokens"."token" = ? LIMIT ?  [["token", "token"], ["LIMIT", 1]]
  Doorkeeper::AccessToken Load (0.1ms)  SELECT  "oauth_access_tokens".* FROM "oauth_access_tokens" WHERE "oauth_access_tokens"."refresh_token" = ? LIMIT ?  [["refresh_token", ""], ["LIMIT", 1]]
   (0.1ms)  begin transaction
   (0.1ms)  commit transaction
  User Load (0.2ms)  SELECT  "users".* FROM "users" WHERE "users"."deleted_at" IS NULL AND "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  DeployGroup Load (0.2ms)  SELECT  "deploy_groups".* FROM "deploy_groups" WHERE "deploy_groups"."deleted_at" IS NULL AND "deploy_groups"."permalink" = ? ORDER BY "deploy_groups"."name_sortable" ASC LIMIT ?  [["permalink", "382f0cd6-9fde-5113-9990-cf7b98058872"], ["LIMIT", 1]]
Completed 500 Internal Server Error in 8ms (ActiveRecord: 0.7ms)



ActionController::UnpermittedParameters (found unpermitted parameter: :name_sortable):

app/controllers/deploy_groups_controller.rb:81:in `resource_params'
app/controllers/resource_controller.rb:67:in `block (2 levels) in update'
app/controllers/resource_controller.rb:65:in `update'

```

[...]

PR #3270, opened 2019-03-23
do not return read-only column as json so writing objects back does n…

grosser:
…ot fail on permitted params

fixes https://github.com/zendesk/samson/issues/3269
fallout from https://github.com/zendesk/samson/pull/3231

... ideally we'd verify all models as_jjson against their permitted params, but that will be a bigger PR

@zendesk/bre @danihodovic


danihodovic:
I'll can verify if the fix works tomorrow because I'm using 4g now and don't have enough bandwidth to pull the rails container for testing.

PR #3272, opened 2019-03-25
Added deploy metrics to deploy csv report

stephennyu:
1. Moved logic to fetch cycle_time into deploy model
2. Added cycle_time into Deploy CSV

/cc @zendesk/samson

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -135,6 +135,10 @@ class Deploy < ActiveRecord::Base
     updated_at - start_time
   end
 
+  def cycle_time
+    DeployMetrics.new(self).cycle_time
+  end
+
   def self.start_deploys_waiting_for_restart!
     pending.reorder(nil).reject(&:waiting_for_buddy?).each do |deploy|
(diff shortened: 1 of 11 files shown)


app/models/deploy.rb
```diff
@@ -135,6 +135,10 @@ def duration
     updated_at - start_time
   end
 
+  def cycle_time
+    @cycle_time ||= DeployMetrics.new(self).cycle_time
```
grosser: remove the nil logic from DeployMetrics then

grosser:
is this needed multiple times ... can this be a 1-off export instead ?

stephennyu:
1. Moved this to a separate presenter
2. Converted it to a one-off export (export whole projects at once)

app/presenters/deploy_metrics_csv_presenter.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+require 'csv'
+
+class DeployMetricsCsvPresenter
```
grosser: stupid-simple ❤️

PR #3273, opened 2019-03-25
also ignore FailedCreatePodSandBox since that auto-resolves itself af…

grosser:
…ter a bit

```
Warning FailedCreatePodSandBox: Failed create pod sandbox: rpc error: code = Unknown desc = NetworkPlugin cni failed to set up pod
```

@zendesk/compute


hgokavarapuz (approved):

adammw:
Can we determine if this is specific to our use case or general that `FailedCreatePodSandBox` will generally autoresolve?

adammw (approved):

grosser:
if someone complains with the opposite problem (FailedCreatePodSandBox that did not resolve) we can add more logic

PR #3274, opened 2019-03-25
PAAS-3025 debug-log shell/vault/kubernetes calls

grosser:
we log AR calls that are usually fast, so lets log this stuff too since it can slow us down significantly

 - being careful not to log script with secrets
 - being careful not to log request payloads
 - replacing existing 1-off github instrumentation that is now captured by faraday instrumentation

@zendesk/bre @zendesk/compute

### DIFF
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -25,6 +25,22 @@ ActiveSupport::Notifications.subscribe("execute_job.samson") do |*args|
   Samson.statsd.histogram "execute_shell.time", event.duration, tags: tags
 end
 
+# report and log timing, plain names so git-grep works
+[
+  ["execute.command_executor.samson", "command_executor.execute", []],
+  ["execute.terminal_executor.samson", "terminal_executor.execute", []],
+  ["request.rest_client.samson", "rest_client.request", [:method]],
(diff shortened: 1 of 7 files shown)


config/initializers/statsd.rb
```diff
@@ -25,6 +25,20 @@ class << Samson
...
+  "execute.command_executor.samson",
+  "execute.terminal_executor.samson",
+  "request.rest_client.samson",
+  "request.vault.samson",
+  "request.faraday.samson"
+].each do |metric|
+  ActiveSupport::Notifications.subscribe(metric) do |_, start, finish, _, payload|
+    duration = finish - start
+    Rails.logger.debug(message: metric, duration: duration.round(3), **payload)
+    Samson.statsd.timing metric.split(".").last(2).reverse!.join("."), duration
```
sathishavm: Can we reuse or make everything in one place?
https://github.com/zendesk/samson/blob/0342c6990b7e1aa5427d982fd12661f0e901d689/config/initializers/statsd.rb#L65

grosser: updated

sathishavm (approved):
LGTM

PR #3280, opened 2019-03-26
Revert "Added deploy metrics to deploy csv report"

stephennyu:
This reverts commit a9340956d07d1393f4167c57ab311e5f35af23ad.
Due to some timeout and github API rate limitting found in staging. Might need to be further tweaked


grosser:
just don't use it for now, nobody is going to find it :)

grosser:
one approach could be to add some caching for refs since we'll have lots of similar deploys
other would be to make users request for a certain project only
if we want one giant export we'll need some kind of busy-wait for the github client and then the eport will just take forever ... I think we got 5k requests / hour so do the math of how long it will take 🤷‍♂️

stephennyu:
Okay then I will close this PR and work on the improvement. Thanks

PR #3285, opened 2019-03-27
separate log formatter and log location options

grosser:
this file was confusing and lograge and syslog don't need to be the same ... so now I can debug lograge locally without having to use syslog

### DIFF
--- a/lib/samson/logging.rb
+++ b/lib/samson/logging.rb
@@ -1,33 +1,19 @@
 # frozen_string_literal: true
+
 config = Rails.application.config
 
 if Samson::EnvCheck.set?("RAILS_LOG_TO_STDOUT")
-  # heroku and docker: dump everything to stdout
+  # good for heroku or docker
   config.logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT))
 elsif Samson::EnvCheck.set?("RAILS_LOG_TO_SYSLOG")
-  require 'lograge'
-  require 'logstash-event'
-  # log 1 message per request to syslog in json format
-  config.lograge.enabled = true
-
-  config.lograge.custom_options = ->(event) do
-    # show params for every request
-    unwanted_keys = %w[format action controller]
-    params = event.payload[:params].reject { |key, _| unwanted_keys.include? key }
-    params['commits'] = '... truncated ...' if params['commits'] # lots of metadata from github we don't need
-    request = event.payload[:headers].instance_variable_get(:@req)
-    {
-      params: params,
-      user_id: request.env['warden']&.user&.id,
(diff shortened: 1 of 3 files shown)


lib/samson/logging.rb
```diff
@@ -1,33 +1,19 @@
...
-
+  # good for production hosts with syslog
+  # TODO: add Syslog::Logger#silence to support dev mode
+  # TODO: add tagged logging so we get request-ids
   config.logger = Syslog::Logger.new('samson')
-  config.action_cable.logger = Syslog::Logger.new('samson')
-  config.lograge.formatter = Lograge::Formatters::Logstash.new
-  config.logger.formatter = config.action_cable.logger.formatter = Samson::SyslogFormatter.new
-elsif ENV["SERVER_MODE"] # regular file logger that needs rotating
+  config.logger.formatter = Samson::SyslogFormatter.new
```
sathishavm: Need to add formatter for 'action_cable',  Samson strictly needs to log in json format if it is RAILS_LOG_TO_SYSLOG, But `action_cable` isn't doing it.

lib/samson/logging.rb
```diff
@@ -46,7 +32,25 @@
   end
 end
 
-# workaround for our server ... always gets SIGHUP even though it does not do file logging
-if ENV["IGNORE_SIGHUP"]
-  Signal.trap(:SIGHUP) {}
+# log 1 message per request to in json/syslog format
+if ENV["RAILS_LOG_WITH_LOGRAGE"]
```
sathishavm: Need to add in .env.example with explanation

lib/samson/logging.rb
sathishavm: we may break elk!!

grosser: verified in staging that it still does
```
{"level":1,"@timestamp":"2019-03-27T22:17:38.152+00:00","application":"samson","host":"XXX","message":"Successfully upgraded to WebSocket (REQUEST_METHOD: GET, HTTP_CONNECTION: upgrade, HTTP_UPGRADE: websocket)"}
```

lib/samson/logging.rb
grosser: added

sathishavm:
Pls see comments in https://zendesk.atlassian.net/browse/BRE-218, Some action cable logs is not in json format, that's why I added it!!

grosser:
https://github.com/zendesk/samson/pull/2871 added the actioncabble logger but did not include repro instructions ... all I can see is that syslog actioncable things are coming in on staging which uses syslog and locally if I `bundle open` syslog and `puts` in `syslog/logger.rb` instead of writing to syslog

PR #3288, opened 2019-03-27
fix seeding vs deleted bugs and allow seeding without primary resource

grosser:
... also improve coverage
found by @aksgithub 🎉 
@zendesk/compute


plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -65,28 +65,30 @@ def self.seed!(project, git_ref)
       if configs.empty?
         raise Samson::Hooks::UserError, "No configs found in kubernetes folder or invalid git ref #{git_ref}"
       end
+      existing = where(project: project, deleted_at: nil).to_a
```
grosser: avoiding N+1 by moving this outside

ankilosaurus (approved):

PR #3289, opened 2019-03-27
we need multiple ClusterRoleBinding per role, so they cannot have the…

grosser:
… same name

### Risks
 - high: existing role-bindings get renamed and we end up with duplicates, need to check that first

@zendesk/compute


ankilosaurus (approved):

grosser:
nvm we got `template.dig(:metadata, :annotations, :'samson/keep_name') == 'true'`

PR #3294, opened 2019-03-29
report time as time to datadog instead of histogram

grosser:
afaik timing is a special case of histogram so let's use it https://docs.datadoghq.com/developers/dogstatsd/data_types/#timers "Timers in DogStatsD are an implementation of Histograms"
(also unifying how duration is calculated)

@zendesk/bre 
/cc @zendesk/compute


grosser:
FYI @stephennyu this will report cycle time as ms now to be compatible with how datadog handles time metrics

PR #3299, opened 2019-04-01
rename misleading kubernetes_roles

grosser:
they are not "roles" but just a join-table called "StageRole"
followup to https://github.com/zendesk/samson/pull/3115

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/decorators/stage_decorator.rb
+++ b/plugins/kubernetes/app/decorators/stage_decorator.rb
@@ -6,8 +6,8 @@ Stage.class_eval do
   after_create :seed_kubernetes_roles
   validate :validate_not_using_non_kubernetes_rollback, if: :kubernetes?
 
-  has_many :kubernetes_roles, class_name: "Kubernetes::StageRole", dependent: :destroy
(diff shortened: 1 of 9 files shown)


plugins/kubernetes/app/decorators/stage_decorator.rb
```diff
@@ -6,8 +6,8 @@
   after_create :seed_kubernetes_roles
   validate :validate_not_using_non_kubernetes_rollback, if: :kubernetes?
 
-  has_many :kubernetes_roles, class_name: "Kubernetes::StageRole", dependent: :destroy
-  accepts_nested_attributes_for :kubernetes_roles,
+  has_many :kubernetes_stage_roles, class_name: "Kubernetes::StageRole", dependent: :destroy
```
adammw: inverse_of ?

adammw (approved):

plugins/kubernetes/app/decorators/stage_decorator.rb
grosser: got a test for that :)
when it's obvious rails can figure it out on it's own

yizhang-zen (approved):

PR #3300, opened 2019-04-01
PAAS-3057 calculate allowed errors per role

grosser:
if a role has a single pod and it fails, consider it a failure
if a role has 10 pods and 1 of them fails, consider it ok

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
@@ -392,7 +392,7 @@ describe Kubernetes::DeployExecutor do
     it "fails when release has errors" do
       Kubernetes::Release.any_instance.expects(:persisted?).at_least_once.returns(false)
       e = assert_raises(Samson::Hooks::UserError) { execute }
-      e.message.must_equal "Failed to create release: []" # inspected errros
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
```diff
@@ -441,6 +441,13 @@ def worker_is_unstable
       out.must_include "UNSTABLE"
     end
 
+    it "ignores allowed percentage of failures" do
```
adammw: need a test with your example listed in the description

adammw (approved):

yizhang-zen (approved):
lgtm...just simple questions: Does the value allow to set greater than 10? And is it possible/valuable to make it as an option in Deploy UI?

grosser:
value is "percent", so 0-100

grosser:
for now it's global ... until someone requests something special ... will most likely set it to 25%

plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
grosser: changed it to 1/3 since that is the lowest I can do without messing with the whole test setup

PR #3303, opened 2019-04-02
show that stage role config is optional

grosser:
before looked like you have to configure them, now shows "optional"
@zendesk/compute 
/cc @aksgithub 

Screen Shot 2019-04-01 at 8 31 23 PM


hgokavarapuz (approved):

adammw:
maybe have one of those popup "i" things explaining what this is/does? I have no idea from the UI...

PR #3304, opened 2019-04-02
cache ids to avoid hitting memcached for each secret lookup

grosser:
this reduces deploy time from ~16s -> ~1.2s

before:
Screen Shot 2019-04-02 at 10 38 32 AM

after:
Screen Shot 2019-04-02 at 10 38 46 AM

@zendesk/compute

### DIFF
--- a/lib/samson/secrets/key_resolver.rb
+++ b/lib/samson/secrets/key_resolver.rb
@@ -61,6 +61,11 @@ module Samson
 
       private
 
+      # cache since we use this for every secret
+      def ids
+        @ids ||= Samson::Secrets::Manager.ids
+      end
+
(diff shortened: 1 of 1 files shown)


lib/samson/secrets/key_resolver.rb
```diff
@@ -9,6 +9,7 @@ def initialize(project, deploy_groups)
         @project = project
         @deploy_groups = deploy_groups
         @errors = []
+        @ids = Samson::Secrets::Manager.ids # cache since we use this for every secret
```
adammw: is this ever not used? worth having a memoized method?

grosser: always used

PR #3306, opened 2019-04-03
PAAS-3056 POC kustomize

grosser:
kustomize has limited capabilities, but could be useful ... so if anyone wants to use it we should be able to add it easily and without overhead for normal deploys

https://github.com/kubernetes-sigs/kustomize

@zendesk/bre @zendesk/compute


adammw:
that was quick...

grosser:
for fuller support we most likely need to get all files in the folder or parse the files ... things like bases and automatic config-map generation look neat

PR #3308, opened 2019-04-04
Tag images in parallel

dasch:
In e.g. [this deploy](https://samson.zende.sk/projects/help_center/deploys/693573#L504) we spend more than 20 seconds tagging images. This PR parallelizes that.

I can't get Samson to run locally, and I'm off on paternity in a week, but @grosser would you be able to take a look?

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### Risks
- Level: Med

### DIFF
--- a/plugins/gcloud/lib/samson_gcloud/image_tagger.rb
+++ b/plugins/gcloud/lib/samson_gcloud/image_tagger.rb
@@ -19,7 +19,7 @@ module SamsonGcloud
           next unless digest.match?(/(^|\/|\.)gcr.io\//) # gcr.io or https://gcr.io or region like asia.gcr.io
           base = digest.split('@').first
 
-          tags.each { |tag| tag_image(tag, base, digest, output) }
+          Samson::Parallelizer.map(tags) { |tag| tag_image(tag, base, digest, output) }
         end
       end
 


grosser:
@zendesk/bre how about we get rid of all this extra tagging stuff ... idk anyone who uses it :/

dasch:
Every second counts when you're trying to deploy a fix during an incident, so we should be prudent about what we spend time on.

ragurney:
I think @zendesk/bre is using those tags for ZDI, so they might be necessary to keep around. If so, this doesn't really affect the time it takes to actually deploy code, right? The code has already been successfully deployed by the time image tagging is done, it's more of a post-deploy step from my understanding.

sathishavm:
I agree with @ragurney, ZDI is using 'production' tag, we also thinking to improvise more by collecting all organizing all metadata for each production release.

grosser:
with "extra" I meant getting rid of anything that's not "production", does that sound good ?

dasch:
> I think @zendesk/bre is using those tags for ZDI, so they might be necessary to keep around. If so, this doesn't really affect the time it takes to actually deploy code, right? The code has already been successfully deployed by the time image tagging is done, it's more of a post-deploy step from my understanding.

From the perspective of a developer, a deployment is finished when Samson marks it as finished.

ragurney:
It sounds like that could be fixed with a UI change then?

grosser:
no longer a big issue by tagging less and caching the tagging
we could also add some post-processing step, but that would make failures harder to see and things more indirect, so I doubt it's worth it

PR #3310, opened 2019-04-04
PAAS-3105 allow environment variables to be set via api

eatwithforks:
cannot merge because compliance. compliance requires to show the diff before something is deployed. will address this in a separate PR, just parking this here for now.


grosser:
rebased

eatwithforks:
rebased

PR #3311, opened 2019-04-04
make env vars reviewable

grosser:
problem: when a user enters per-deploy environment variables they did not show up in buddy review window (neither did stage var changes)

@zendesk/compute @zendesk/bre 

Screen Shot 2019-04-04 at 4 02 41 PM

also:
 - rework priority to be additive instead of callback based
 - make deploy review and confirm consistent (not removing trailing space)
 - allow any params in deploy/new
 - sort env vars correctly by deploy/stage/project/env/deploy-group

### Risks
 - medium: project-wide env vars are now directly in the job executors environment which could lead to surprises
 - low: previously all deploy/stage env vars were injected into all deploys no matter if they matched the scope

### DIFF
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -42,7 +42,6 @@ module Samson
       :build_permitted_params,
       :buildkite_release_params,
       :can,
-      :deploy_group_env,
       :deploy_group_includes,
       :deploy_group_permitted_params,
       :deploy_permitted_params,
--- a/plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
+++ b/plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
@@ -72,8 +72,8 @@ Samson::Hooks.callback :stage_permitted_params do
   ]
 end
 
-Samson::Hooks.callback :deploy_env do |deploy|
-  next {} if deploy.stage.aws_sts_iam_role_arn.blank?
+Samson::Hooks.callback :deploy_env do |deploy, _deploy_group, *|
+  next {} if deploy.stage&.aws_sts_iam_role_arn.blank?
 
   SamsonAwsSts::Client.new(SamsonAwsSts.sts_client).deploy_env_vars(deploy)
 end
--- a/plugins/env/app/controllers/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/environment_variable_groups_controller.rb
@@ -54,7 +54,7 @@ class EnvironmentVariableGroupsController < ApplicationController
       @project = Project.find(params[:project_id])
     end
 
-    @groups = SamsonEnv.env_groups(@project, deploy_groups, preview: true)
+    @groups = SamsonEnv.env_groups(Deploy.new(project: @project), deploy_groups, preview: true)
 
     respond_to do |format|
       format.html
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -3,6 +3,7 @@ require 'validates_lengths_from_database'
 
 class EnvironmentVariable < ActiveRecord::Base
   FAILED_LOOKUP_MARK = ' X' # SpaceX
+  PARENT_PRIORITY = ["Deploy", "Stage", "Project"].freeze
 
   include GroupScope
   extend Inlinable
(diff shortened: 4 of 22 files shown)


plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -54,7 +54,7 @@ def preview
       @project = Project.find(params[:project_id])
     end
 
-    @groups = SamsonEnv.env_groups(@project, deploy_groups, preview: true)
+    @groups = SamsonEnv.env_groups(Deploy.new(project: @project), deploy_groups, preview: true)
```
adammw: is deploy_groups not part of the `Deploy` object?

grosser:
no

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
```diff
@@ -72,8 +72,8 @@ def session_name(deploy)
   ]
 end
 
-Samson::Hooks.callback :deploy_env do |deploy|
-  next {} if deploy.stage.aws_sts_iam_role_arn.blank?
+Samson::Hooks.callback :deploy_env do |deploy, *|
```
adammw: _deploy_group would be clearer what the second variable you're expecting is...

lib/samson/hooks.rb
```diff
@@ -42,7 +42,6 @@ class UserError < StandardError
       :build_permitted_params,
       :buildkite_release_params,
       :can,
-      :deploy_group_env,
       :deploy_group_includes,
```
adammw: do you document the signature of each hook somewhere? would be good to say what you expect to call deploy_env with... (deploy + deploy groups)

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -468,8 +468,8 @@ def static_env
         env[:BLUE_GREEN] = blue_green_color if blue_green_color
 
         # env from plugins
-        plugin_envs = Samson::Hooks.fire(:deploy_group_env, project, @doc.deploy_group, resolve_secrets: false)
-        plugin_envs += Samson::Hooks.fire(:deploy_env, @doc.kubernetes_release.deploy) if @doc.kubernetes_release.deploy
+        deploy = @doc.kubernetes_release.deploy || Deploy.new(project: project)
+        plugin_envs = Samson::Hooks.fire(:deploy_env, deploy, @doc.deploy_group, resolve_secrets: false)
```
adammw: do you still need to pass through a separate deploy_group? can't you get that from the deploy?

lib/samson/hooks.rb
grosser: yeah, for now it's "look at the source" :)

plugins/kubernetes/app/models/kubernetes/template_filler.rb
grosser: deploys don't have a deploy-group, they only know their stage which has multiple

plugins/env/app/models/environment_variable.rb
```diff
@@ -92,13 +94,13 @@ def resolve_secrets(project, deploy_group, env, preview:)
     end
   end
 
-  # used by `priority` from GroupScope
-  def project?
-    parent_type == "Project"
+  def priority
+    [PARENT_PRIORITY.index(parent_type) || 999, super]
```
adammw: why use a lookup here when you used a case statement in another class?
also didn't you get rid of the array?

plugins/aws_sts/lib/samson_aws_sts/samson_plugin.rb
grosser: added

plugins/env/app/models/environment_variable.rb
grosser: the base class uses an integer and the sub-classes change it to an array so they can prepend their own priority

grosser: hmmm yeah I'll change the other one to array too

adammw: can you put that in a comment somewhere too...

lib/samson/hooks.rb
adammw: this is where a strongly typed language would help :p

plugins/kubernetes/test/models/kubernetes/usage_limit_test.rb
```diff
@@ -79,4 +79,17 @@ def create_limit(attributes = {})
       Kubernetes::UsageLimit.most_specific(project, deploy_group).must_equal found
     end
   end
+
+  describe "#priority" do
+    let(:limit) { create_limit(scope: environments(:staging)) }
+
+    it "is high with project" do
```
adammw: aren't there other types to check too?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
adammw: would the deploy's deploy group ever not match the stages?

adammw: oh, this is for a single deploy group? that makes more sense...

adammw (approved):
can't understand everything but looks good on the whole.

plugins/kubernetes/test/models/kubernetes/usage_limit_test.rb
grosser: usage limits only have project or are global

PR #3312, opened 2019-04-05
Support customizable GitHub deploy comment

ragurney:
Adds the ability to customize the comment left by Samson on PRs which have been deployed. This is helpful for public repositories, where I would still want to give an indication that a PR has been deployed, but not show a restricted link back to Samson. 

/cc @zendesk/samson

### Tasks
 - [x] :+1: from team

### Risks
- Level: Changes the GitHub commenting on deploy functionality

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -549,6 +549,7 @@ ActiveRecord::Schema.define(version: 2019_04_03_202316) do
     t.string "aws_sts_iam_role_arn"
     t.integer "aws_sts_iam_role_session_duration"
     t.boolean "allow_redeploy_previous_when_failed", default: false, null: false
+    t.string "github_pull_request_comment"
     t.index ["project_id", "permalink"], name: "index_stages_on_project_id_and_permalink", unique: true, length: { permalink: 191 }
     t.index ["template_stage_id"], name: "index_stages_on_template_stage_id"
   end
--- a/lib/samson/form_builder.rb
+++ b/lib/samson/form_builder.rb
@@ -36,7 +36,12 @@ module Samson
           end
         else
           input_html[:class] ||= "form-control"
-          content = label(attribute, label, class: "col-lg-2 control-label")
+          content = if label.present?
+            label(attribute, label, class: 'col-lg-2 control-label')
+          else
+            # Spacer with no label, useful for subfields or fields that are already under a different header
+            content_tag(:div, '', class: 'col-lg-2 control-label')
+          end
           content << content_tag(:div, class: 'col-lg-4', &block)
           content << help
         end
--- a/plugins/github/app/decorators/stage_decorator.rb
+++ b/plugins/github/app/decorators/stage_decorator.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+Stage.class_eval do
+  validate :validate_github_pull_request_comment_variables
+
+  private
+
+  def validate_github_pull_request_comment_variables
+    if comment = github_pull_request_comment.presence
+      comment % Hash[GithubNotification::SUPPORTED_KEYS.map { |k| [k, ''] }] # Validate no extra keys are supplied
+    end
+  rescue KeyError => e
+    errors.add(:github_pull_request_comment, e.message)
+  end
+end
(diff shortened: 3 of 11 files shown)


lib/samson/form_builder.rb
```diff
@@ -36,7 +36,11 @@ def input(
           end
         else
           input_html[:class] ||= "form-control"
-          content = label(attribute, label, class: "col-lg-2 control-label")
+          content = if label.presence
```
grosser:
```suggestion
          content = if label.present?
```

lib/samson/form_builder.rb
```diff
@@ -36,7 +36,11 @@ def input(
...
+            label(attribute, label, class: 'col-lg-2 control-label')
+          else
+            content_tag(:div, '', class: 'col-lg-2 control-label')
```
grosser: why is an empty label bad ? (addd comment for that)

plugins/github/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+Stage.class_eval do
+  validate :github_pull_request_comment_variables
+
+  private
+
+  def github_pull_request_comment_variables
```
grosser:
```suggestion
  def validate_github_pull_request_comment_variables
```

db/schema.rb
```diff
@@ -549,6 +549,7 @@
     t.string "aws_sts_iam_role_arn"
     t.integer "aws_sts_iam_role_session_duration"
     t.boolean "allow_redeploy_previous_when_failed", default: false, null: false
+    t.text "github_pull_request_comment"
```
grosser: prefer string since it should not be that long ?

plugins/github/app/decorators/stage_decorator.rb
```diff
@@ -0,0 +1,14 @@
...
+    if comment = github_pull_request_comment.presence
+      comment % {stage_name: '', reference: ''} # Validate no extra keys are supplied
```
grosser: call GithubNotification#body instead if duplicating the logic ? ... or use a constant for the supported keys or something like that ...

plugins/github/app/views/samson_github/_fields.html.erb
```diff
@@ -1,5 +1,15 @@
 <fieldset>
   <legend>GitHub</legend>
-  <%= form.input :update_github_pull_requests, as: :check_box, label: "Comment on Pull Requests after deployment" %>
-  <%= form.input :use_github_deployment_api, as: :check_box, label: "Use GitHub Deployment API" %>
+  <%= form.input :update_github_pull_requests,
+        as: :check_box,
+        label: 'Comment on Pull Requests after deployment',
+        input_html: { id: 'gh-comment-checkbox' }
```
grosser: why does it need an id ?

plugins/github/app/views/samson_github/_fields.html.erb
```diff
@@ -1,5 +1,15 @@
...
+  %>
+  <%= form.input :github_pull_request_comment,
+        as: :text_area,
+        label: '',
+        help: 'Customize comment message. Supports interpolation of `stage_name` and `reference`',
```
grosser:
```suggestion
        help: 'Customize comment message. Supports replacement of <code>%{stage_name}</code> and <code>%{reference}</code>'.html_safe',
```

plugins/github/test/models/github_notification_test.rb
```diff
@@ -4,20 +4,41 @@
...
-
     it "adds a comment" do
       assert_request(:post, endpoint) do
         notification.deliver
       end
     end
   end
+
+  describe '#body' do
+    it('uses the correct default message') do
```
grosser:
```suggestion
    it 'uses the correct default message' do
```

plugins/github/test/models/github_notification_test.rb
```diff
@@ -4,20 +4,41 @@
...
+      body = notification.send(:body)
+
+      body.must_equal <<~TEXT.squish
+        This PR was deployed to staging.
+        Reference: <a href='http://www.test-url.com/projects/3-glitter/deploys/18' target='_blank'>7e6c415</a>
+      TEXT
+    end
+
+    it('uses the user supplied message') do
```
grosser:
```suggestion
    it 'uses the user supplied message' do
```

ragurney:
@grosser thanks for the comments. Ready for a re-review!

lib/samson/form_builder.rb
grosser: is is it different from this:
`label(attribute, '', class: 'col-lg-2 control-label')`

grosser: how is it different from a label with an empty value ?

PR #3313, opened 2019-04-05
simplify repetetive plugin callback scoping

grosser:
@zendesk/compute @zendesk/bre


adammw:
this does make it a little more mysterious.. maybe put the word `around` in the method name so we know it's doing it as a block? then again, i have no idea what this does so 🤷‍♂️

grosser:
I'd expect all top-level callback to behave in an `around` way like `with_env` / `with_registries` / `with_forgery_protection` etc

adammw:
fair. i guess this is a rails project after all, the home of 🔮

PR #3314, opened 2019-04-05
improve hooks UI

grosser:
- organize data/flow
 - collapse distracting info
 - make generic info copy-able + add curl example
 - explain/link release flow

Screen Shot 2019-04-05 at 2 54 39 PM
Screen Shot 2019-04-05 at 2 54 35 PM
Screen Shot 2019-04-05 at 2 43 10 PM

@zendesk/compute @zendesk/bre

### DIFF
--- a/app/views/webhooks/index.html.erb
+++ b/app/views/webhooks/index.html.erb
@@ -3,48 +3,122 @@
 <%= render 'projects/header', project: @project, tab: "webhooks" %>
 
 <% inbound_webhooks = @project.webhooks %>
-<% outbound_webhooks = @project.outbound_webhooks %>
-
 <section class="tabs">
-  <h2>Inbound Webhooks</h2>
-  <p>In order to automatically trigger a deploy to a stage, add a branch &rarr; stage mapping below
-  and paste in the project's webhook URL to your CI service of choice.</p>
+  <h2>
+    Inbound Webhooks
+    <%= additional_info "Allows CI or source-code provider to trigger a release or deploy." %>
+  </h2>
 
-  <dl>
-    <% Samson::Integration::SOURCES.each do |source| %>
-      <dt>
-        <%= source.titleize %><%= webhook_help_text source %>
-      </dt>
-      <dd><%= send("integrations_#{source}_deploy_url", @project.token) %>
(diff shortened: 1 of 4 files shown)


app/views/webhooks/index.html.erb
```diff
@@ -3,48 +3,129 @@
...
+                {
+                  deploy: {
+                    branch: &ltname&gt,
+                    commit: {
+                      sha: &ltsha&gt,
+                      message: &ltmessage&gt
+                    }
+                  }
+                }
+                curl URL -H 'Content-Type: application/json' -d '{"deploy":{"branch":"master","commit":{"sha":"abc","message":"hi"}}}'
```
adammw: separate code block for json and curl?

adammw: either that or use 
```
cat << EOF | curl -d - ...
{
  "deploy": {}
}
EOF
```

app/views/webhooks/index.html.erb
```diff
@@ -3,48 +3,129 @@
...
-      </dt>
-      <dd><%= send("integrations_#{source}_deploy_url", @project.token) %>
-        <% if source == 'github' && token = Integrations::GithubController.secret_token %>
-          <br/>
-          <% if current_user.admin_for?(@project) %>
-            (Set <%= link_to "secret token", "javascript:prompt('Copy this token', '#{token}')" %> in <%= link_to 'hook settings', "https://github.com/#{@project.repository_path}/settings/hooks" %>)
+    <dl>
+      <% Samson::Integration::SOURCES.each do |source| %>
+        <dt>
+          <% if source == 'generic' %>
```
adammw: feel like these should really be generic hooks that the integration provides rather than supplying all of them in the template guarded with if statements...

app/views/webhooks/index.html.erb
adammw: also interpolate the URL.. we should know it right?

grosser: change to just curl + real url

app/views/webhooks/index.html.erb
grosser: maybe ... but already rewriting enough of this :D

PR #3315, opened 2019-04-05
do not auto-export all env vars to the stage env

grosser:
https://github.com/zendesk/samson/pull/3311 broken because we tried to resolve secrets from env vars
that did not have a "global" config since while generating env vars for the deploy we did not have
a deploy group and fell back to global ... changing direction to only have an explicit "deploy_execution_env"
for the 1-off usecase and not merge it into kubernetes ... which was an accident I hope

/cc users who used the "stage env" feature on a non-kubernetes stage @smedberg @fadimaali @irwingarry @hughobrien
```
es = EnvironmentVariable.where(parent_type: "Stage").all.select { |e| e.parent.project && !e.parent.kubernetes };nil
User.find(es.map { |e| e.audits.last.user_id }.uniq)
```

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -204,8 +204,7 @@ class JobExecution
       PROJECT_REPOSITORY: @job.project.repository_url
     )
 
-    deploy = @job.deploy || Deploy.new(project: @job.project)
-    Samson::Hooks.fire(:deploy_env, deploy, nil).compact.inject(env, :merge!)
+    Samson::Hooks.fire(:deploy_execution_env, @job.deploy).compact.inject(env, :merge!) if @job.deploy
     base_commands(dir, env) + @job.commands
   end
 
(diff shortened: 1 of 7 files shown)


smedberg:
Thanks for the fix!  However, if you'd like me to review this, I might need some background on how the env vars in Samson are actually supposed to work.  Most of my "understanding" is based on just trying stuff in the Samson UI and seeing if it works.

grosser:
Yeah docs were bad :(

env vars were supposed to be available to the deployed app, not "during the deploy" ... and they got somehow mixed up ... for now stage env vars will still work since I want to deploy this soon, but next I'll try to hunt down all usages of "during the deploy" and then remove them and then remove the fallback.

grosser:
... so no need to review, nothing will break, just FYI to not add more things that rely on this deprecated behavior

grosser:
@nogates are you still using deploy env variables this for deploy execution ?

originally deploy_env came from https://github.com/zendesk/samson/pull/2666
and for consistency if would be nice if all of them are in the deployed resource and not in the execution environment

PR #2679, opened 2018-04-12
DEVEX-408 Tag images with environment permalink

ragurney:
@zendesk/devex

PR #3319, opened 2019-04-16
only tag production image since we do not use the others

grosser:
alternative for https://github.com/zendesk/samson/pull/3308 to stop the tagging-spam after each deploy

reverts https://github.com/zendesk/samson/pull/2679 which came from BRE-408 which has no description as to why we wanted that.

@zendesk/bre @dasch @ragurney @dragonfax

### DIFF
--- a/plugins/gcloud/lib/samson_gcloud/image_tagger.rb
+++ b/plugins/gcloud/lib/samson_gcloud/image_tagger.rb
@@ -7,40 +7,24 @@ module SamsonGcloud
       # Note: not tagging builds from different project since that would be confusing ...
       # ideally do not tag any builds for projects that use shared builds ... but that is hard to know atm
       def tag(deploy, output)
-        return unless ENV["DOCKER_FEATURE"]
-        return unless deploy.succeeded?
-        return unless builds = deploy.project.builds.
-          where(git_sha: deploy.job.commit).where.not(docker_repo_digest: nil).to_a.presence
+        return unless needs_tag?(deploy)
 
-        tags = tags(deploy)
+        builds = deploy.project.builds.
+          where(git_sha: deploy.job.commit).
+          where.not(docker_repo_digest: nil)
 
         builds.each do |build|
           digest = build.docker_repo_digest
           next unless digest.match?(/(^|\/|\.)gcr.io\//) # gcr.io or https://gcr.io or region like asia.gcr.io
           base = digest.split('@').first
-
-          tags.each { |tag| tag_image(tag, base, digest, output) }
+          tag_image PRODUCTION_TAG, base, digest, output
         end
       end
 
       private
 
-      def tags(deploy)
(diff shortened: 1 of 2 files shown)


plugins/gcloud/lib/samson_gcloud/image_tagger.rb
```diff
@@ -11,38 +9,19 @@ def tag(deploy, output)
         return unless deploy.succeeded?
         return unless builds = deploy.project.builds.
           where(git_sha: deploy.job.commit).where.not(docker_repo_digest: nil).to_a.presence
-
-        tags = tags(deploy)
+        return unless deploy.stage.production?
```
ragurney: Could just make this one return statement?
```ruby
return if !deploy.stage.production? || deploy.stage.no_code_deployed?
```

plugins/gcloud/lib/samson_gcloud/image_tagger.rb
```diff
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 module SamsonGcloud
   class ImageTagger
-    PRODUCTION_TAG = 'production'
```
ragurney: We should keep this so it can be used [here](https://github.com/zendesk/samson/pull/3319/files#diff-ee1c2f12150e834b6ae360c7421fb2a9R19)

plugins/gcloud/lib/samson_gcloud/image_tagger.rb
ragurney: Maybe we should take all of these `return unless` statements L8 - L12 and put them into a method `needs_tag?`?

sathishavm:
We are using only **production** tag both in '**zdi**' and '**zdi-cloud**'
Let we hear from @russell if they **need** 'stage' tags for images in '**scooter**'

bcolferzd:
Speeding up the deploy process is a good thing.
It may be possible to preserve the intent of tagging the images with a post-hoc deploy tagging process.

russell:
well, i hadn't switched to using the stage tag yet, but i was going to use it for detecting if we had successfully run stages, but i'll figure out another way

pkmsingh:
We BRE , specifically wanted to setup Development environment through these Tags, so that Development environment is near to staging/production and idea is to bring consistency to environments.

grosser:
- which tags did you have in mind using ? (stage or env)
 - would a opt-in flag (per stage or project) work ?

pkmsingh:
wanted Env Tags, which I can see they are already applied there "env-staging , env-production" in some chat services, but I am not sure how these two tags get updated wrt to regional rollout(deploy groups) in both staging and production through multiple stages.

grosser:
- would a opt-in flag (per stage or project) work ?
 - is it ok to do env-staging but not do `env-production` since we already do `production`?

PR #3320, opened 2019-04-16
cleanup deletion logic and make it harder to accidentally press for important resources

grosser:
@zendesk/bre @zendesk/compute 

Screen Shot 2019-04-16 at 12 30 39 PM

/cc @mhratson let's do the same for cerebro

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -134,7 +134,9 @@ module ApplicationHelper
     content_tag :i, '', options.merge(class: css_classes)
   end
 
-  def link_to_delete(path, text: 'Delete', question: nil, remove_container: false, disabled: false, **options)
+  def link_to_delete(
+    path, text: 'Delete', question: nil, type_to_delete: false, remove_container: false, disabled: false, **options
+  )
     if disabled
       content_tag :span, text, title: disabled, class: 'mouseover'
     else
@@ -150,7 +152,12 @@ module ApplicationHelper
         else
           "Really delete ?"
         end
-      options[:data] = {confirm: message, method: :delete}
+      options[:data] = {method: :delete}
+      if type_to_delete
+        options[:data][:type_to_delete] = message
+      else
+        options[:data][:confirm] = message
+      end
       if remove_container
         options[:data][:remove_container] = remove_container
         options[:data][:remote] = true
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -124,7 +124,7 @@ module ApplicationHelper
     end
   end
 
-  def icon_tag(type, options = {})
+  def icon_tag(type, **options)
     css_classes = "glyphicon glyphicon-#{type}"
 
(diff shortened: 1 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -134,35 +134,39 @@ def icon_tag(type, options = {})
...
-      content_tag :span, text, title: disabled_reason, class: 'mouseover'
+  def link_to_delete(
+    path, text: 'Delete', question: nil, type_to_delete: false, remove_container: false, disabled: false, **options
+  )
+    if disabled
+      content_tag :span, text, title: disabled, class: 'mouseover'
     else
-      resource = Array(path).last
       message =
-        if question = options.delete(:question)
```
grosser: refactoring to keyword-args instead of deleting things from options hash

app/views/stages/show.html.erb
```diff
@@ -18,7 +18,6 @@
 
     <% if admin_for_project? %>
       <%= link_to "Edit", edit_project_stage_path(@project, @stage), class: "btn btn-default" %>
-      <%= link_to_delete_button([@project, @stage]) %>
```
grosser: only in the edit view now to be consistent with project

mhratson:
>  @mhratson let's do the same for cerebro

Sounds good!

app/helpers/application_helper.rb
```diff
@@ -134,35 +134,39 @@ def icon_tag(type, options = {})
...
+        if question
           question
-        elsif resource.is_a?(ActiveRecord::Base)
+        elsif path.is_a?(ActiveRecord::Base)
+          resource = path
+          name, path = link_parts_for_resource(resource)
+          "Delete #{resource.class.name.split("::").last} #{name} ?"
+        elsif (resource = Array(path).last) && resource.is_a?(ActiveRecord::Base)
```
eatwithforks: In this `elsif`, resource is unknown here? where is resource defined?

app/helpers/application_helper.rb
```diff
@@ -134,35 +134,39 @@ def icon_tag(type, options = {})
...
-      if container = options[:remove_container]
-        options[:data][:remove_container] = container
+      options[:data] = {method: :delete}
+      if type_to_delete
+        options[:data][:type_to_delete] = message
+      else
+        options[:data][:confirm] = message
+      end
+      if remove_container
+        options[:data][:remove_container] = remove_container
```
eatwithforks: set to `true` instead?

grosser: keeping it open for future enhancements, someone might set something else, afaik no need to preemptively break that

app/helpers/application_helper.rb
grosser: resource comes from `Array(path).last)` so if path is an array then it's the last element and otherwise it's the `path`

grosser: using `(a = 1) && 2` pattern so `a != 2`

PR #3322, opened 2019-04-16
do not tag the same image again if we already tagged it

grosser:
makes https://github.com/zendesk/samson/pull/3308 less of an issue

@dasch @zendesk/bre @zendesk/compute


plugins/gcloud/lib/samson_gcloud/image_tagger.rb
```diff
@@ -38,6 +40,16 @@ def tag_image(tag, base, digest, job_output)
           #{output.strip}
         TEXT
         job_output.puts "FAILED" unless success
+        success
+      end
+
+      def cache_last_tagged(key, value)
+        old = Rails.cache.read(key)
+        if old == value
+          Rails.cache.write(key, value) # refresh the cache expiration
```
grosser: actually a race condition here but doubt that will happen very often ... and if it does it will be fixed with the next version

PR #3323, opened 2019-04-17
PAAS-2636 start production console in sandbox mode by default

grosser:
prompt will then be `samson(sandbox prod)>` to indicate it
can disable with --no-sandbox

@zendesk/bre @zendesk/compute

### DIFF
--- a/.irbrc.rb
+++ b/.irbrc.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+# loaded by marco-polo gem, marking sandbox mode
+if Rails.application.sandbox
+  [:PROMPT_I, :PROMPT_N].each { |p| IRB.conf[:PROMPT][:RAILS_ENV][p].sub!("(", "(sandbox ") }
+end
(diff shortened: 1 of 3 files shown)


adammw:
Sandbox isn't that obvious but if that's the name of the rails feature it makes sense. Where does that come from?

grosser:
it's builtin to rails console, I was looking to add a readonly mode but most articles I found delegated to sandbox and the only other readonly thing I found were nasty sql hacks ... so I figured let's take the builtin stuff since it's easy and will be a nice transferrable skill for anyone who learns about it

adammw:
Can you print [this warning](https://dev.to/dwilkins/comment/41i9) when a sandbox console is started?

ie. that sandbox can still cause some types of changes, and relies on db transactions...

grosser:
ugh ... yeah that sounds like we should not use it then, more like a trap
for a future outage :/
... so I'll look into real readonly ...

PR #3324, opened 2019-04-18
add readonly mode to production console

grosser:
Alternative for https://github.com/zendesk/samson/pull/3323 which can lock up the whole DB

```
rails c
Loading development environment (Rails 5.2.2.1)
Running in readonly mode. Use Samson::ReadonlyDb.disable to switch to writable.
samson(readonly dev)> Samson::ReadonlyDb.disable
samson(dev)>
```

@zendesk/compute @zendesk/bre

### DIFF
--- a/lib/samson/readonly_db.rb
+++ b/lib/samson/readonly_db.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+# Make Activerecord readonly by blocking write requests
+# NOTE: not perfect and can be circumvented with multiline sql statements or `;`
+module Samson
+  module ReadonlyDb
+    ALLOWED = ["SELECT ", "SHOW ", "SET  @@SESSION.", "EXPLAIN ", "PRAGMA "].freeze
+    PROMPT_CHANGE = ["(", "(readonly "].freeze
+    PROMPTS = [:PROMPT_I, :PROMPT_N].freeze
+
+    class << self
+      def enable
+        return if @subscriber
+        @subscriber = ActiveSupport::Notifications.subscribe("sql.active_record") do |*, payload|
+          sql = payload.fetch(:sql)
+          next if sql.lstrip.start_with?(*ALLOWED)
+          raise ActiveRecord::ReadOnlyRecord, <<~MSG
+            Database is in readonly mode, cannot execute query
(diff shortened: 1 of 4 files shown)


lib/samson/readonly_db.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    class << self
+      def enable
+        return if @subscriber
+        @subscriber = ActiveSupport::Notifications.subscribe("sql.active_record") do |*, payload|
+          sql = payload.fetch(:sql)
+          next if sql.start_with?(*ALLOWED)
+          raise <<~MSG
+            Database is in readonly mode, cannot execute query
+            Switch off readonly with #{self}.#{method(:disabble).name}
```
grosser: fetching class/method so docs stay in sync

lib/samson/readonly_db.rb
```diff
@@ -0,0 +1,48 @@
...
+module Samson
+  module ReadonlyDb
+    ALLOWED = ["SELECT ", "SHOW ", "SET  @@SESSION.", "EXPLAIN "].freeze
+    PROMPT_CHANGE = ["(", "(readonly "].freeze
+    PROMPTS = [:PROMPT_I, :PROMPT_N].freeze
+
+    class << self
+      def enable
+        return if @subscriber
+        @subscriber = ActiveSupport::Notifications.subscribe("sql.active_record") do |*, payload|
```
adammw: is there any documented guarantees that the notification will fire before the actual sql gets executed?

lib/samson/readonly_db.rb
adammw: `disable`

grosser:
I'll add tests to make sure the callback blocks the actual sql, so we'll
know if rails breaks it

adammw:
not sure if possible, but I wonder if wrapping it in a transaction _as well_ would make doubly-sure that nothing bad sneaks through...

adammw:
also does transactions still work in here, do you have to allow "START TRANSACTION" and "ROLLBACK"?

grosser:
we decided against transactions in the sandbox PR because they can cause global blocking, so I don't want to bring them back

grosser:
🍿 can't wait for this to blow up

PR #3327, opened 2019-04-19
wait for FailedAttachVolume warning to go away and not fail in samson

ankilosaurus:
@zendesk/samson

https://zendesk.atlassian.net/browse/PAAS-3155


eatwithforks (approved):

adammw (approved):

grosser (approved):

grosser:
FYI a safer approach would be to only ignore the failure when the message has "creating" but until we need it current state is good enough


```
[10:38:16]   Warning FailedAttachVolume: AttachVolume.Attach failed for volume "pvc-xyz" : "Error attaching EBS volume \"vol-xyz\"" to instance "i-xyz" since volume is in "creating" state x3
```

grosser:
FYI also ignores this one `Warning FailedAttachVolume: Multi-Attach error for volume "pvc-foo" Volume is already exclusively attached to one node and can't be attached to another`

PR #3326, opened 2019-04-19
 try debugging api foobar

grosser:
we see these a lot ...

theories are:
 - fetch_entities returns bad result
 - defining the methods blows up
 - methods get removed

```
Kubernetes::DeployExecutor::#execute::an autoscaled role#test_0001_only requires one pod to go live when a role is autoscaled:
Samson::Hooks::UserError: apiVersion extensions/v1beta1 does not support Deployment. Check kubernetes docs for correct apiVersion
    plugins/kubernetes/app/models/kubernetes/resource.rb:213:in `block in request'
    lib/samson/retry.rb:9:in `with_retries'
    plugins/kubernetes/lib/samson_kubernetes/samson_plugin.rb:20:in `retry_on_connection_errors'
    plugins/kubernetes/app/models/kubernetes/resource.rb:207:in `request'
    plugins/kubernetes/app/models/kubernetes/resource.rb:186:in `block in fetch_resource'
    plugins/kubernetes/app/models/kubernetes/resource.rb:251:in `ignore_404'
    plugins/kubernetes/app/models/kubernetes/resource.rb:185:in `fetch_resource'
    plugins/kubernetes/app/models/kubernetes/resource.rb:78:in `resource'
    plugins/kubernetes/app/models/kubernetes/release_doc.rb:23:in `map'
    plugins/kubernetes/app/models/kubernetes/release_doc.rb:23:in `deploy'
    lib/samson/parallelizer.rb:10:in `block (2 levels) in map'
    lib/samson/parallelizer.rb:10:in `block in map'
bin/rails test plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb:557
```

PR #3328, opened 2019-04-20
avoid race conditions from defining methods on demand for kubeclient

grosser:
debug steps are in https://github.com/zendesk/samson/pull/3326

there might be other race conditions that this avoids too, like sending 2 requests at once

known issue cam from the call here https://github.com/zendesk/samson/blob/76901674fef7e429d650a20d7ba16f1eb005dce5/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb#L372

@zendesk/compute @zendesk/bre

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/cluster.rb
+++ b/plugins/kubernetes/app/models/kubernetes/cluster.rb
@@ -26,7 +26,7 @@ module Kubernetes
     before_destroy :ensure_unused
 
     def client(type)
-      (@client ||= {})[type] ||= begin
+      (@client ||= {})[[Thread.current.object_id, type]] ||= begin
         case auth_method
         when "context"
           context = kubeconfig.context(config_context)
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/cluster.rb
```diff
@@ -26,7 +26,7 @@ class Cluster < ActiveRecord::Base
     before_destroy :ensure_unused
 
     def client(type)
-      (@client ||= {})[type] ||= begin
+      (@client ||= {})[[Thread.current.object_id, type]] ||= begin
```
adammw: alternatively a mutex/lock might have worked...

adammw (approved):

sathishavm (approved):

plugins/kubernetes/app/models/kubernetes/cluster.rb
grosser: - the client most likely is not threadsafe internally, so 1 per thread is safer
 - the race condition is not here but when resolving the "is get_deployments defined" so I'd have to hack a mutex around that

PR #3330, opened 2019-04-20
PAAS-3170 Link locks to their unlock page + fix global lock buttons n…

grosser:
…ot showing

@zendesk/compute @zendesk/bre

Screen Shot 2019-04-20 at 2 48 51 PM

### DIFF
--- a/app/helpers/locks_helper.rb
+++ b/app/helpers/locks_helper.rb
@@ -34,4 +34,14 @@ module LocksHelper
     text = (lock.warning? ? "#{warning_icon} Warning" : "#{lock_icon} Locked")
     content_tag :span, text.html_safe, class: "label label-warning", title: strip_tags(lock.summary)
   end
+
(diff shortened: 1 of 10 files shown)


app/helpers/locks_helper.rb
```diff
@@ -34,4 +34,14 @@ def resource_lock_icon(resource)
     text = (lock.warning? ? "#{warning_icon} Warning" : "#{lock_icon} Locked")
     content_tag :span, text.html_safe, class: "label label-warning", title: strip_tags(lock.summary)
   end
+
+  def lock_affected(lock)
+    if lock.resource_type == "Stage"
+      "stage"
```
adammw: why is this not a link / name of resource?

grosser: it will only ever be shown on the page of the sage itself, so no need to link it to anything

PR #3332, opened 2019-04-23
remove a few binstubs

grosser:
afaik nobody uses them ... let's remove the clutter ...

@zendesk/bre @zendesk/compute


grosser:
random failure fixed in https://github.com/zendesk/samson/pull/3335

Issue #3251, opened 2019-03-18
Running a build, triggered by gcloud plugin


nayena:
Using gcloud sdk v 238.0.0, the build triggers the following error:
```
[05:20:38] » gcloud container builds submit . --timeout 7200 --config /tmp/samson-es-biblen-220190318-1-1fg7jop/cloudbuild.yml --account samson-builder@jdsys-229319.iam.gserviceaccount.com --project jdsys-229319
[05:20:38] ERROR: (gcloud.container) Invalid choice: 'builds'.
[05:20:38] Maybe you meant:
[05:20:38]   gcloud builds submit
[05:20:38] 
[05:20:38] Showing 1 out of 49 suggestions.
```

The command seems to have been renamed/moved. How should this be approached, to ensure backwards compability with current users?

grosser:
I think viable options are:
 - add version detection
 - install an older version of the sdk
 - enforce a more modern version in initializer

PR welcome for first option (most likely need some caching so we don't run
the version check on every execution)

PR #3333, opened 2019-04-23
detect gcloud version to allow upgrading

grosser:
fixes https://github.com/zendesk/samson/issues/3251

### DIFF
--- a/plugins/gcloud/lib/samson_gcloud/image_builder.rb
+++ b/plugins/gcloud/lib/samson_gcloud/image_builder.rb
@@ -41,8 +41,9 @@ module SamsonGcloud
 
         prevent_upload_of_ignored_files(dir, build)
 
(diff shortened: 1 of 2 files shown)


plugins/gcloud/lib/samson_gcloud/image_builder.rb
```diff
@@ -59,6 +60,12 @@ def build_image(dir, build, executor, tag_as_latest:, cache_from:)
 
       private
 
+      def gcloud_new?
```
adammw: would prefer if this method was gcloud_version and you did the comparison where you use it...

adammw: (new is relative)

grosser: yeah gcloud_version was also nice because it removes the `false` caching logic which is always annoying

PR #3335, opened 2019-04-24
fix coverage randomly being bad because cache was filled before first…

grosser:
… run

```
ruby test/helpers/application_helper_test.rb -s 14004
Lines missing coverage:
app/helpers/application_helper.rb:117:7-124:10
```

@zendesk/bre @zendesk/compute

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -219,6 +219,11 @@ ActiveSupport::TestCase.class_eval do
     plugin_name = caller(1..1).first[/\/plugins\/([^\/]+)/, 1] || raise("not called from a plugin")
     around { |t| Samson::Hooks.only_callbacks_for_plugin(plugin_name, callback, &t) }
   end
+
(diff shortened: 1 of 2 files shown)


test/test_helper.rb
```diff
@@ -219,6 +219,12 @@ def self.only_callbacks_for_plugin(callback)
     plugin_name = caller(1..1).first[/\/plugins\/([^\/]+)/, 1] || raise("not called from a plugin")
     around { |t| Samson::Hooks.only_callbacks_for_plugin(plugin_name, callback, &t) }
   end
+
+  def self.before_and_after(&block)
+    before(&block)
+    after(&block)
+    after(&block)
```
hgokavarapuz: why two after?

grosser: :trollface: oops

grosser:
and another random failure, it never ends :D

PR #3336, opened 2019-04-25
show triggered deploys on triggering deploy

grosser:
makes navigating pipelines a bit easier and shows triggered deploys status without having to open all the urls from the output log

adds 1 more query to the deploy page, but it's indexed so should be cheap

Screen Shot 2019-04-25 at 1 39 46 PM
Screen Shot 2019-04-25 at 1 39 38 PM

@zendesk/bre @zendesk/compute

### DIFF
--- a/plugins/pipelines/app/views/samson_pipelines/_deploy_header.html.erb
+++ b/plugins/pipelines/app/views/samson_pipelines/_deploy_header.html.erb
@@ -1,5 +1,17 @@
 <% if triggering_deploy = deploy.triggering_deploy %>
   <div class="alert alert-info">
-    Triggered via <%= link_to "Deploy ##{triggering_deploy.id}", project_deploy_path(deploy.project, triggering_deploy) %>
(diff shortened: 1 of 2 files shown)


plugins/pipelines/app/views/samson_pipelines/_deploy_header.html.erb
```diff
@@ -1,5 +1,17 @@
...
-    Triggered via <%= link_to "Deploy ##{triggering_deploy.id}", project_deploy_path(deploy.project, triggering_deploy) %>
+    Triggered via <%= link_to "#{triggering_deploy.short_reference} to #{triggering_deploy.stage&.name}", [@project, triggering_deploy] %>
+  </div>
+<% elsif triggered_deploys = Deploy.where(triggering_deploy_id: deploy).presence %>
+  <div class="alert alert-info">
+    Triggered deploys:
+    <ul>
+      <% triggered_deploys.each do |deploy| %>
+        <%= link_to [@project, deploy] do %>
+          <%= "#{deploy.short_reference} to #{deploy.stage&.name}" %>
```
adammw: need to include the ref? this will be the same for all...

PR #3340, opened 2019-04-29
Allow including locks when fetching stage json

davidreuss:
In scooter, we'd want to not delete resources if the stage backing them
is locked.

ref: QA-1777

/cc @zendesk/samson

Risks: Low, not user facing


davidreuss:
That’s good to know, we’ll deal with that scenario if it becomes a requirement

PR #3347, opened 2019-05-03
use a progress deadline for deployment timeouts instead of manual timer

craig-day:
Instead of having Samson watch the pods and manage the timeout, I propose that we use the [Progressing Deployment](https://v1-12.docs.kubernetes.io/docs/concepts/workloads/controllers/deployment/#progressing-deployment) mechanisms provided by Kubernetes. This will allow deployments to set their own timeouts, but keeps the safety mechanism of setting a default timeout so deploys don't sit around forever. 

/cc @zendesk/samson

### Risks
- Level: Med

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -8,7 +8,6 @@ module Kubernetes
     if ENV['KUBE_WAIT_FOR_LIVE'] && !ENV["KUBERNETES_WAIT_FOR_LIVE"]
       raise "Use KUBERNETES_WAIT_FOR_LIVE with seconds instead of KUBE_WAIT_FOR_LIVE" # uncovered
     end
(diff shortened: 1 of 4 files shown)


grosser:
hmm looks interesting ... might get us out of having to monitor all pods by hand

 - this will make deployments that don't get ready in time rollback, right ?
 - should be opt-in at first so we can try it out safely
 - won't work for non-deployments
 - would break KUBERNETES_ALLOW_NOT_READY_PERCENT

craig-day:
Yeah, so kubernetes should fail this deployment in a similar way to an image pull failure. It also handles readiness probe failures and a few other things so it might entirely eliminate the need to monitor per pod and we can monitor per deployment instead. That would be a hefty refactor, but I'm happy to pair with you on it. Might warrant a new module to encapsulate the logic and make opt-in and testing easier.

grosser:
Let's get it to a state where we can opt-in to it so we can play around ...
afaik all we have to do is disable pod monitoring + timeout and it might
just work out ... maybe next week, doubt I'll have much time today

PR #3348, opened 2019-05-06
Skipped Pagination for Locks index

sathishavm:
> Skip's pagination for the rendering Locks controller index method in JSON format.
> Limited to render only JSON format

### Risks
- Level: Low

### DIFF
--- a/app/controllers/resource_controller.rb
+++ b/app/controllers/resource_controller.rb
@@ -115,7 +115,7 @@ class ResourceController < ApplicationController
   end
 
   def search_resources
-    resource_class
+    resource_class.all
   end
 
   # hook
--- a/test/controllers/locks_controller_test.rb
+++ b/test/controllers/locks_controller_test.rb
@@ -57,6 +57,21 @@ describe LocksController do
   as_a :project_deployer do
     unauthorized :post, :create, lock: {description: 'xyz'}
 
+    describe '#index' do
+      it "doesn't paginate" do
+        2.times { create_lock stage }
+        get :index, format: :json, params: {per_page: 1}
+        json = JSON.parse(response.body)
+        json['locks'].count.must_equal 2
+      end
+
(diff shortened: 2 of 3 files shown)


app/controllers/resource_controller.rb
```diff
@@ -115,7 +115,7 @@ def redirect_after_save
   end
 
   def search_resources
-    resource_class
+    resource_class.all
```
grosser: why was this necessary ?

test/controllers/locks_controller_test.rb
```diff
@@ -57,6 +57,21 @@ def create_lock(resource = nil, options = {})
...
 
+    describe '#index' do
+      it "doesn't paginate" do
+        2.times { create_lock stage }
+        get :index, format: :json, params: {per_page: 1}
+        json = JSON.parse(response.body)
+        json['locks'].count.must_equal 2
+      end
+
+      it "refuses to render html" do
```
grosser: nice!

app/controllers/resource_controller.rb
sathishavm: Well this is optional for paginated resources, but for non-paginated resource, we need to list all the resources.
While during pagination this will get overridden by pagy.

sathishavm:
> every page without a limit could one day be dead or be a DOS ... so prefer to keep pagination and make the client iterate

Agree, but the locks have API URL '/api/locks' which doesn't have good documentation about pagination for this URL and people started complaining about this.

app/controllers/resource_controller.rb
grosser: prefer to put that into the resoruce_controller so it's fixed in general

grosser:
then improve the documentation ? :D
... make sure the reply also includes some meta attributes that says if this is the last page or so ...

PR #3351, opened 2019-05-07
allow waiting for cluster resources then reset live timeout

craig-day:
This starts an approach to changing how the live timeout check works during kubernetes deploys. 

For context, there might be a situation where a deployment happens and the kubernetes cluster doesn't have enough resources to schedule the workload. In our case, we have a cluster autoscaler in place to detect this and add nodes to the cluster. This process; however, takes about 5 minutes to get the new nodes ready to accept workloads. If we run with the default live timeout of 10 minutes, that means my deployment really only has 5 minutes to get ready after waiting for 5 minutes for the cluster to scale. 

This PR creates two timeout windows, both using the live timeout value. First, we assume we're going to wait on the cluster to provision more resources for us. Then, if none of the pods are waiting for those resources, meaning the workload has been scheduled, we reset our start time and the wait period. This should allow the actual full live timeout value for the pods to actually get live instead of waiting on the cluster to scale up. 

/cc @zendesk/samson

### Risks
- Level: Low. Deploys could take longer before timing out, or we would provide confusing messaging if we think we're waiting for resources but aren't.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -4,7 +4,7 @@ module Kubernetes
     class Pod
       INIT_CONTAINER_KEY = :'pod.beta.kubernetes.io/init-containers'
       INGORED_AUTOSCALE_EVENT_REASONS = %w[FailedGetMetrics FailedRescale].freeze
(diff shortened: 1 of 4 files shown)


grosser:
just bump the timeout 🤷‍♂

PR #3352, opened 2019-05-07
Use a popover to display detected problems

adammw:
Use a popover to display detected problems - no more pushing down the deploy button when you're about to click as the problems UI loads.

Screen Shot 2019-05-07 at 1 26 13 pm

/cc @zendesk/samson @zendesk/compute @grosser 

### Tasks
 - [ ] :+1: from team

### References
 - Jira link:

### Risks
- Level: Low - UI change

### DIFF
--- a/app/views/deploys/new.html.erb
+++ b/app/views/deploys/new.html.erb
@@ -17,12 +17,13 @@
 
       <div class="form-group">
         <%= form.label :reference, "Reference", class: "col-lg-2 control-label" %>
+        <div class="col-lg-4">
         <% if @stage.no_reference_selection %>
           <%= form.object.reference = @stage.default_reference.presence || 'master' %>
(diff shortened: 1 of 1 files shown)


dndungu:
OOT: The label placements would work better placed above the input fields. 

https://www.uxmatters.com/mt/archives/2006/07/label-placement-in-forms.php

app/views/deploys/new.html.erb
```diff
@@ -17,12 +17,13 @@
 
       <div class="form-group">
         <%= form.label :reference, "Reference", class: "col-lg-2 control-label" %>
+        <div class="col-lg-4">
```
grosser: indent

adammw: label is not a block though...?

PR #3354, opened 2019-05-08
PAAS-3303 allow configuring kubernetes rollout timeout per project

grosser:
we have a few projects that take forever, so allow them to still deploy

@zendesk/compute @zendesk/bre @craig-day 

Screen Shot 2019-05-08 at 4 37 03 PM

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -8,8 +8,8 @@ module Kubernetes
     if ENV['KUBE_WAIT_FOR_LIVE'] && !ENV["KUBERNETES_WAIT_FOR_LIVE"]
       raise "Use KUBERNETES_WAIT_FOR_LIVE with seconds instead of KUBE_WAIT_FOR_LIVE" # uncovered
     end
-    WAIT_FOR_LIVE = Integer(ENV.fetch('KUBERNETES_WAIT_FOR_LIVE', '600'))
-    WAIT_FOR_PREREQUISITES = Integer(ENV.fetch('KUBERNETES_WAIT_FOR_PREREQUISITES', WAIT_FOR_LIVE))
+    DEFAULT_ROLLOUT_TIMEOUT = Integer(ENV.fetch('KUBERNETES_WAIT_FOR_LIVE', '600'))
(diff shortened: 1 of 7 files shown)


uthark (approved):

plugins/kubernetes/app/views/samson_kubernetes/_project_form.html.erb
```diff
@@ -0,0 +1,4 @@
+<fieldset>
```
adammw: so this is per project, not per stage?

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -44,7 +44,7 @@ def execute(*)
       end
 
       if deploys.any?
-        return false unless deploy_and_watch(deploys, timeout: WAIT_FOR_LIVE)
+        return false unless deploy_and_watch(deploys, timeout: @job.project.kubernetes_rollout_timeout || WAIT_FOR_LIVE)
```
adammw: rename constant DEFAULT_WAIT_FOR_LIVE?

adammw: or default wait timeout

adammw (approved):

plugins/kubernetes/app/views/samson_kubernetes/_project_form.html.erb
grosser: usually all stages deploy the same slow software, so would make sense to have it on the project 🤷‍♂

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
grosser: -> DEFAULT_ROLLOUT_TIMEOUT

plugins/kubernetes/app/views/samson_kubernetes/_project_form.html.erb
adammw: I'm thinking more if say production or one pod has way more replicas than others, but yes i guess that's a different problem for us to solve a different way via "making progress" events or the like...

grosser: can always crank this high as an escape hatch then :D

PR #3359, opened 2019-05-13
Config Service Integration

arecker:
[CONFIG-332] env: add stubbed "use config service" option to stage
[CONFIG-354] env: add use config service option to deploy

[CONFIG-332]: https://zendesk.atlassian.net/browse/CONFIG-332
[CONFIG-354]: https://zendesk.atlassian.net/browse/CONFIG-354

### DIFF
--- a/plugins/env/app/assets/javascripts/env/application.js
+++ b/plugins/env/app/assets/javascripts/env/application.js
@@ -157,5 +157,4 @@
       setupEnvGroupPreview($(input));
     });
   });
-
 }());
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -26,9 +26,23 @@ class EnvironmentVariable < ActiveRecord::Base
     # also used by an external plugin
     def env(deploy, deploy_group, preview: false, resolve_secrets: true)
       env = {}
+
+      if deploy_group && deploy.project.config_service?
+        begin
+          url = config_service_folder(deploy.project, required: true)
+          url += "/#{deploy_group.permalink}.yml" # TODO: version ?
+          response = Samson::Retry.with_retries(Faraday::Error, 3) { Faraday.get(url) }
+          raise "Invalid response #{response.status}" unless response.status == 200
+          env.merge! YAML.safe_load(response.body)
+        rescue StandardError => e
+          raise Samson::Hooks::UserError, "Error reading env vars from config service: #{e.message}"
+        end
+      end
+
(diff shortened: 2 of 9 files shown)


plugins/env/app/assets/javascripts/env/application.js
```diff
@@ -158,4 +158,12 @@
     });
   });
 
+  // Disable environment inputs if "use config service" checked
+  $(document).ready(function() {
```
grosser: we are already in a .ready here

plugins/env/app/assets/javascripts/env/application.js
```diff
@@ -158,4 +158,12 @@
...
+    var envVarInputs = $('.environment-variable-inputs');
+    $('#use_config_service').change(function() {
+      envVarInputs.prop('disabled', this.checked);
```
grosser: should only disable them if they are empty ... otherwise the existing ones would still stay ... 
prefer allowing mixing for an easy migration

plugins/env/app/models/environment_variable.rb
```diff
@@ -26,6 +26,9 @@ class << self
     # also used by an external plugin
     def env(deploy, deploy_group, preview: false, resolve_secrets: true)
       env = {}
+      if deploy.use_config_service
+        return {'HAVING_FUN' => 'true'}
```
grosser: prefer to either just merge or merge after dollar/secrets are resolved

plugins/env/app/views/samson_env/_deploy_form.html.erb
```diff
@@ -3,6 +3,13 @@
     Extra Environment Variables for this deploy <%= additional_info SamsonEnv::HELP_TEXT %>
   </legend>
 
+  <div class="form-group">
+    <div class="col-lg-3">
+      <%= form.check_box :use_config_service, id: 'use_config_service',  checked: @deploy.stage.use_config_service %>
```
grosser: gets an automatic id that we can use

grosser: should not be settable per deploy or this gets us in trouble with auditing I think

grosser:
done ... just need some way to request the specified version 🤷‍♂ 
... and some docs on what the version field should be set to / how to find out more about versions

grosser:
removed version since that does not work per project, neither does it work per stage 🤷‍♂ 
... I doubt we want users to fiddle with 1 s3 version per deploy-group either ... so either merge this and then improve or come up with a new plan :(

PR #3362, opened 2019-05-13
PAAS-3264 only allow namespace deletion when unused and better explan…

grosser:
…ation

@zendesk/compute 

Screen Shot 2019-05-13 at 10 54 44 AM
Screen Shot 2019-05-13 at 10 54 29 AM

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/namespace.rb
+++ b/plugins/kubernetes/app/models/kubernetes/namespace.rb
@@ -11,5 +11,14 @@ module Kubernetes
     has_many :projects, dependent: nil, foreign_key: :kubernetes_namespace_id, inverse_of: :kubernetes_namespace
 
     validates :name, presence: true, uniqueness: true, format: NAME_PATTERN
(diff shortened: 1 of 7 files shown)


uthark (approved):

plugins/kubernetes/app/models/kubernetes/namespace.rb
```diff
@@ -11,5 +11,14 @@ class Namespace < ActiveRecord::Base
     has_many :projects, dependent: nil, foreign_key: :kubernetes_namespace_id, inverse_of: :kubernetes_namespace
 
     validates :name, presence: true, uniqueness: true, format: NAME_PATTERN
+    before_destroy :ensure_unused
+
+    private
+
+    def ensure_unused
+      return if projects.none?
+      errors.add :base, 'Can only delete when not used by any project.'
```
adammw: the verbage is a little strange here.. would have expected `"Used by #{projects.size} projects."`

PR #3365, opened 2019-05-15
PAAS-3264 fill namespace from project

grosser:
stop setting resource names and service names when using namespace feature (since there should be no collisions)

Screen Shot 2019-05-15 at 10 53 30 AM
Screen Shot 2019-05-15 at 10 53 16 AM

Screen Shot 2019-05-15 at 10 50 05 AM

### DIFF
--- a/plugins/kubernetes/app/decorators/project_decorator.rb
+++ b/plugins/kubernetes/app/decorators/project_decorator.rb
@@ -10,4 +10,8 @@ Project.class_eval do
     class_name: 'Kubernetes::Namespace', dependent: nil, inverse_of: :projects, optional: true
 
   scope :with_kubernetes_roles, -> { where(id: Kubernetes::Role.not_deleted.pluck(Arel.sql('distinct project_id'))) }
+
+  def override_resource_names?
+    !kubernetes_namespace_id?
+  end
 end
--- a/plugins/kubernetes/app/models/kubernetes/namespace.rb
+++ b/plugins/kubernetes/app/models/kubernetes/namespace.rb
@@ -12,6 +12,7 @@ module Kubernetes
     has_many :projects, dependent: nil, foreign_key: :kubernetes_namespace_id, inverse_of: :kubernetes_namespace
 
     validates :name, presence: true, uniqueness: true, format: NAME_PATTERN
+    after_save :remove_configured_resource_names
     before_destroy :ensure_unused
 
     private
@@ -20,6 +21,16 @@ module Kubernetes
       return if projects.none?
       errors.add :base, 'Can only delete when not used by any project.'
       throw :abort
+    end
+
+    # we will no longer use these, so clean the DB up and leave audits behind
+    def remove_configured_resource_names
+      roles = Kubernetes::Role.
+        where(project_id: projects.map(&:id)).
+        where("resource_name IS NOT NULL OR service_name IS NOT NULL")
+      roles.find_each do |role|
+        role.update_attributes!(resource_name: nil, service_name: nil, manual_deletion_acknowledged: true)
+      end
     end
   end
 end
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -90,26 +90,14 @@ module Kubernetes
 
     def add_pod_disruption_budget
       return unless deployment = raw_template.detect { |r| ["Deployment", "StatefulSet"].include? r[:kind] }
+      return unless target = disruption_budget_target(deployment)
 
-      min_available = deployment.dig(:metadata, :annotations, :"samson/minAvailable")
-      return if min_available == "disabled"
-
-      # NOTE: this is a bit of overhead for 0 or 1 replica deployments, but we don't know if a bad budget existed before
-      min_available ||= ENV["KUBERNETES_AUTO_MIN_AVAILABLE"]
-      return unless min_available
-
-      target = if percent = min_available.to_s[/\A(\d+)\s*%\z/, 1] # "30%" -> 30 / "30 %" -> 30
-        percent = Integer(percent)
-        if percent >= 100
-          raise Samson::Hooks::UserError, "minAvailable of >= 100% would result in eviction deadlock, pick lower"
+      name =
+        if kubernetes_release.project.override_resource_names?
+          kubernetes_role.resource_name
         else
-          [((replica_target.to_f / 100) * percent).ceil, replica_target - 1].min
+          deployment.dig(:metadata, :name)
         end
-      else
-        [replica_target - 1, Integer(min_available)].min
-      end
-      target = 0 if target < 0
-
       annotations = (deployment.dig(:metadata, :annotations) || {}).dup
       annotations[:"samson/updateTimestamp"] = Time.now.utc.iso8601
 
@@ -117,7 +105,7 @@ module Kubernetes
         apiVersion: "policy/v1beta1",
         kind: "PodDisruptionBudget",
         metadata: {
-          name: kubernetes_role.resource_name,
+          name: name,
           labels: deployment.dig_fetch(:metadata, :labels).dup,
           annotations: annotations
         },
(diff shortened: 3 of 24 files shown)


adammw:
I don't understand that last screenshot, 1) where is that, and 2) shouldn't that be "namespace" not projects?

grosser:
first screenshot is the project/roles overview (no longer has resource or service names)
last screenshot is the namespace edit view, where projects get added to the namespace

plugins/kubernetes/app/models/kubernetes/namespace.rb
```diff
@@ -21,5 +22,15 @@ def ensure_unused
       errors.add :base, 'Can only delete when not used by any project.'
       throw :abort
     end
+
+    # we will no longer use these, so clean the DB up and leave audits behind
+    def nil_out_configured_role_names
```
adammw: configured_resource_names?

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -117,7 +117,7 @@ def add_pod_disruption_budget
         apiVersion: "policy/v1beta1",
         kind: "PodDisruptionBudget",
         metadata: {
-          name: kubernetes_role.resource_name,
+          name: kubernetes_role.resource_name || deployment.dig(:metadata, :name),
```
adammw: wonder if you should check manual_resource_naming rather than relying on nil

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -47,12 +48,13 @@ class Role < ActiveRecord::Base
     validates :project, presence: true
     validates :name, presence: true, format: Kubernetes::RoleValidator::VALID_LABEL_VALUE
     validates :service_name,
-      uniqueness: {scope: :deleted_at, allow_nil: true},
+      uniqueness: {scope: :deleted_at},
       format: Kubernetes::RoleValidator::VALID_LABEL_VALUE,
       allow_nil: true
```
adammw: `unless: :manual_resource_naming?` here too?

plugins/kubernetes/app/decorators/project_decorator.rb
```diff
@@ -10,4 +10,8 @@
     class_name: 'Kubernetes::Namespace', dependent: nil, inverse_of: :projects, optional: true
 
   scope :with_kubernetes_roles, -> { where(id: Kubernetes::Role.not_deleted.pluck(Arel.sql('distinct project_id'))) }
+
+  def manual_resource_naming?
```
adammw: rename to make clearer... maybe - override_resource_names?

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -103,7 +103,9 @@ def validate_single_primary_kind
     # template_filler.rb sets name for everything except for IMMUTABLE_NAME_KINDS, keep_name, and Service
     # we make sure users dont use the same name on the same kind twice, to avoid them overwriting each other
     def validate_name_kinds_are_unique
-      # ignore service if we generate their names
+      return if @namespace
```
adammw: check project instead, namespace isn't really enough i think

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -195,12 +194,13 @@ def prefix_service_cluster_ip
 
     def set_namespace
       return if template[:metadata].key?(:namespace)
-      template[:metadata][:namespace] = @doc.deploy_group.kubernetes_namespace
+      template[:metadata][:namespace] = project.kubernetes_namespace&.name || @doc.deploy_group.kubernetes_namespace
```
adammw: shouldn't it be the other way around?

adammw: i think `||=` would be better...

adammw: can do as a separate PR, but we _should_ do it

grosser: I can't thin of a scenario where that would be better, atm we validate that users are using their namespace or namespace-less resources ... if we do `||=` then that adds more ambiguity as to what "nil" namespace means and we again override a configured setting

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -112,12 +112,18 @@ def add_pod_disruption_budget
 
       annotations = (deployment.dig(:metadata, :annotations) || {}).dup
       annotations[:"samson/updateTimestamp"] = Time.now.utc.iso8601
+      name =
```
adammw: tempted to suggest making this into a private method to split up this long method

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -47,12 +48,13 @@ class Role < ActiveRecord::Base
...
     validates :resource_name,
       uniqueness: {scope: :deleted_at, allow_nil: true},
-      format: Kubernetes::RoleValidator::VALID_LABEL_VALUE
+      format: Kubernetes::RoleValidator::VALID_LABEL_VALUE,
+      if: :override_resource_names?
```
adammw: remove

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -74,18 +76,20 @@ def self.seed!(project, git_ref)
         resource = config_file.primary || config_file.elements.first
         name = resource.dig_fetch(:metadata, :labels, :role)
 
-        # ensure we have a globally unique resource name
-        resource_name = resource.dig_fetch(:metadata, :name)
-        if where(deleted_at: nil, resource_name: resource_name).exists?
-          resource_name = "#{project.permalink}-#{resource_name}".tr('_', '-')
-        end
+        if project.override_resource_names?
```
adammw: maybe separate this into a separate concern? pretty big method

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -103,7 +103,9 @@ def validate_single_primary_kind
     # template_filler.rb sets name for everything except for IMMUTABLE_NAME_KINDS, keep_name, and Service
     # we make sure users dont use the same name on the same kind twice, to avoid them overwriting each other
     def validate_name_kinds_are_unique
-      # ignore service if we generate their names
+      return if @project && !@project.override_resource_names?
```
adammw: comment why we ignore this?

adammw: don't validate if doing global?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -195,12 +194,13 @@ def prefix_service_cluster_ip
 
     def set_namespace
       return if template[:metadata].key?(:namespace)
```
adammw: assumed to be validated by role_validator

plugins/kubernetes/app/views/kubernetes/roles/_form.html.erb
```diff
@@ -5,11 +5,13 @@
...
+    <% if @project.override_resource_names? %>
+      <%= form.input :service_name, help: "Override service name with this name. Uses name from yml file if it starts with this name." %>
+      <%= form.input :resource_name,
+          pattern: Kubernetes::RoleValidator::VALID_LABEL_VALUE,
+          help: "Name for the resource this role will create (service/job), a-Z 0-9 and -"
+      %>
+    <% end %>
     <% if @kubernetes_role.manual_deletion_required? %>
       <div class="alert-danger">
         <%= form.input :manual_deletion_acknowledged, as: :check_box, label: "I will manually delete the changed resource/service from all deploy groups before deploying (this will cause a short outage, deleting after deploying will result in undefined behavior with 2 copies running)" %>
```
adammw: add a "sync" button to force-add to clusters

adammw: also either in this PR or a different PR add a cron to always reconcile the namespaces

adammw (approved):

adammw:
Really don't like the multi-select UI.. but fine for now.

adammw:
Can you also add the field for which namespace in the project settings (and disabled for anyone not super admin)? That's where I would expect to look for this setting...

grosser:
allowing to set via projects controller is more complicated since there is no infrastructure to allow some specific attributes + it would require controller changes from the kubernetes plugin side ... so I'm staying out of that for now unless this gets more user-editable

grosser:
for multi-select: alternatives would be a big checkbox list or a row-based select ... I'll play with the row-based select in a different PR

PR #3367, opened 2019-05-15
LoadBalancer blocker

grosser:
nobody using it atm, so we are good 🤷‍♂ 

```ruby
roles = Kubernetes::Role.where(autoscaled: true).all.to_a; nil
ignored_groups = DeployGroup.where(name: []).pluck(:id)
roles.select!(&:project); nil
roles.each do |role|
  groups_ids = Kubernetes::ReleaseDoc.where(kubernetes_role: role).pluck(:deploy_group_id).uniq - ignored_groups

  DeployGroup.where(id: groups_ids).all.each do |group|
    next unless doc = Kubernetes::ReleaseDoc.where(kubernetes_role: role, deploy_group: group).last
    doc.resources.each do |r|
      puts "FOUND #{role.project.url} #{role.name} #{r.template.dig(:metadata, :namespace)}" if r.kind == "LoadBalancer"
    end
  end
end; nil

```

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -65,6 +65,7 @@ module Kubernetes
       validate_not_matching_team
       validate_stateful_set_service_consistent
       validate_stateful_set_restart_policy
+      validate_load_balancer
       unless validate_annotations
         validate_prerequisites_kinds
         validate_prerequisites_consistency
@@ -102,6 +103,17 @@ module Kubernetes
     end
 
     private
+
+    def validate_load_balancer
+      allowed = ENV["KUBERNETES_ALLOWED_LOAD_BALANCER_NAMESPACES"].to_s.split(",")
+      return if allowed.empty?
+      bad = @elements.map do |e|
+        next unless e[:kind] == "LoadBalancer"
+        namespace = e.dig(:metadata, :namespace) || "unset"
+        namespace unless allowed.include?(namespace)
+      end.compact
+      @errors << "LoadBalancer is not allowed in #{bad.join(", ")} namespace" if bad.any?
+    end
 
     def validate_name
       @errors << "Needs a metadata.name" unless map_attributes([:metadata, :name]).all?
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -243,3 +243,7 @@ this is useful when dealing with large deployments that have some random failure
 To debug services or to create resources that needs to reference a selector that doesn't include team/role (like a Gateway), you can disable selector validation with:
 
 `metadata.annotations.samson/service_selector_across_roles: "true"`
+
+### Blocking LoadBalancer usage
+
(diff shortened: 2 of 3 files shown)


adammw:
I thought you said you were going to do this as an admission controller?

grosser:
just POC in case we need it "now" :D

hgokavarapuz:
Sounds good to me for now till the admission controller is figured out and unblock Zorg usage.

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -103,6 +104,12 @@ def self.keep_name?(e)
 
     private
 
+    def validate_load_balancer
+      return unless allowed = ENV["KUBERNETES_ALLOWED_LOAD_BALANCER_NAMESPACE"]
+      return if @elements.none? { |e| e[:kind] == "LoadBalancer" && e.dig(:metadata, :namespace) != allowed }
```
hgokavarapuz: Would prefer to have it as a list of allowed namespaces rather than just a single namespace in case, we want to stick to this for a while and use across multiple namespaces.

grosser:
done

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -103,6 +104,13 @@ def self.keep_name?(e)
 
     private
 
+    def validate_load_balancer
+      allowed = ENV["KUBERNETES_ALLOWED_LOAD_BALANCER_NAMESPACES"].to_s.split(",")
+      return if allowed.empty?
+      return if @elements.none? { |e| e[:kind] == "LoadBalancer" && !allowed.include?(e.dig(:metadata, :namespace)) }
+      @errors << "LoadBalancer is not allowed in this namespace"
```
hgokavarapuz: It would be great if you print the namespace as well in which it is not allowed, so it will be clear.

grosser: done

hgokavarapuz (approved):
might have to rebase this branch though...

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -103,6 +104,17 @@ def self.keep_name?(e)
 
     private
 
+    def validate_load_balancer
+      allowed = ENV["KUBERNETES_ALLOWED_LOAD_BALANCER_NAMESPACES"].to_s.split(",")
+      return if allowed.empty?
+      bad = @elements.map do |e|
+        next unless e[:kind] == "LoadBalancer"
```
uthark: @grosser What is the APIGroup of this kind?
Usually LoadBalancers are part of the service (i.e. kind = 'Service' and spec.type = 'LoadBalancer') 
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#servicespec-v1-core

grosser: yeah ... thx for paying attention :D

PR #3368, opened 2019-05-15
add project permalink to webhook urls so we can quickly see where the…

grosser:
…y go

decided against:
 - not being backwards compatible
 - duplicating all integration urls in projects scope
 - changing the token format to project-token since I want a readable permalink and all possible separators can be in the permalink too

Screen Shot 2019-05-15 at 2 27 51 PM


grosser:
turns out some integrations already send a `project` param, so we can't use that :(

PR #3370, opened 2019-05-17
Added pagination links for JSON api

sathishavm:
## Description
  We have pagination for Json API, but we don't have enough details in JSON API to perform pagination, along with this story, I have added addition details like `links` and `headers` for pagination.
Screen Shot 2019-05-17 at 10 38 58 AM
Screen Shot 2019-05-17 at 10 49 18 AM

### Risks
- Level: Low

### DIFF
--- a/app/controllers/concerns/json_pagination.rb
+++ b/app/controllers/concerns/json_pagination.rb
@@ -2,12 +2,12 @@
 module JsonPagination
   private
 
-  def add_pagination_links(json, pagy)
+  def add_json_pagination_links(json, pagy)
     # add custom headers to the response
-    add_headers(pagy)
+    add_json_pagination_headers(pagy)
 
     # Insert links at beginning
-    links = links(pagy)
+    links = json_pagination_links(pagy)
     if links.present?
       json_d = json.clone
       json.clear
@@ -16,7 +16,7 @@ module JsonPagination
     end
   end
 
-  def pages(pagy)
+  def json_pagination_pages(pagy)
     paging = {}
     paging[:first] = 1          unless pagy.page == 1
     paging[:last] = pagy.pages  unless pagy.page == pagy.pages
@@ -26,19 +26,19 @@ module JsonPagination
     paging
   end
 
-  def links(pagy)
+  def json_pagination_links(pagy)
     links = {}
     uri = request.env["PATH_INFO"]
 
     # Build pagination links
-    pages(pagy).each do |key, value|
+    json_pagination_pages(pagy).each do |key, value|
       query_params = request.query_parameters.merge(page: value)
       links[key] = "#{uri}?#{query_params.to_param}"
     end
     links
   end
 
-  def add_headers(pagy)
+  def add_json_pagination_headers(pagy)
     headers['X-PER-PAGE']      = pagy.offset
     headers["X-CURRENT-PAGE"]  = pagy.page
     headers["X-TOTAL-PAGES"]   = pagy.pages
--- a/app/controllers/concerns/json_pagination.rb
+++ b/app/controllers/concerns/json_pagination.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+module JsonPagination
+  private
+
+  def add_pagination_links(json, pagy)
+    # add custom headers to the response
+    add_headers(pagy)
+
+    # Insert links at beginning
+    links = links(pagy)
+    if links.present?
+      json_d = json.clone
+      json.clear
+      json["links"] = links
+      json.merge! json_d
+    end
+  end
+
+  def pages(pagy)
+    paging = {}
+    paging[:first] = 1          unless pagy.page == 1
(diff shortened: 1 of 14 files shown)


app/controllers/concerns/json_pagination.rb
```diff
@@ -0,0 +1,47 @@
...
+      links[key] = "#{uri}?#{query_params.to_param}"
+    end
+    links
+  end
+
+  def add_headers(pagy)
+    headers['X-PER-PAGE']      = pagy.offset
+    headers["X-CURRENT-PAGE"]  = pagy.page
+    headers["X-TOTAL-PAGES"]   = pagy.pages
+    headers["X-TOTAL-RECORDS"] = pagy.count
```
grosser: is this stolen from somewhere/a common pattern ?
would it be easier to have `pagination` in the reply so that users don't have to read headers ?
(check what github or zendesk or openapi do)

app/controllers/concerns/json_renderer.rb
```diff
@@ -40,7 +40,7 @@ def add_includes(root, resources, includes)
 
   private
 
-  def render_as_json(namespace, resource, status: :ok, allowed_includes: [])
+  def render_as_json(namespace, resource, status: :ok, allowed_includes: [], pagy: nil)
```
grosser: make pagy a required argument so all call-sites have to consider adding `@pagy` or `nil`

app/controllers/concerns/json_pagination.rb
```diff
@@ -0,0 +1,47 @@
...
+    links = links(pagy)
+    if links.present?
+      json_d = json.clone
+      json.clear
+      json["links"] = links
+      json.merge! json_d
+    end
+  end
+
+  def pages(pagy)
```
grosser: prefer less generic names since this will be mixed into every controller

app/controllers/concerns/json_pagination.rb
```diff
@@ -0,0 +1,47 @@
...
+    paging = {}
+    paging[:first] = 1          unless pagy.page == 1
+    paging[:last] = pagy.pages  unless pagy.page == pagy.pages
+
+    paging[:prev] = pagy.prev   unless pagy.prev.nil?
+    paging[:next] = pagy.next   unless pagy.next.nil?
+    paging
+  end
+
+  def links(pagy)
```
grosser: less generic name

app/controllers/concerns/json_pagination.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    # Build pagination links
+    pages(pagy).each do |key, value|
+      query_params = request.query_parameters.merge(page: value)
+      links[key] = "#{uri}?#{query_params.to_param}"
+    end
+    links
+  end
+
+  def add_headers(pagy)
```
grosser: less generic name

app/controllers/concerns/json_pagination.rb
```diff
@@ -0,0 +1,47 @@
...
+    # add custom headers to the response
+    add_headers(pagy)
+
+    # Insert links at beginning
+    links = links(pagy)
+    if links.present?
+      json_d = json.clone
+      json.clear
+      json["links"] = links
+      json.merge! json_d
```
grosser: just add them add the end ?

grosser:
in general might be quicker to pair for a bit to decide on direction together before making big PRs lie this

app/controllers/concerns/json_renderer.rb
sathishavm: done

app/controllers/concerns/json_pagination.rb
sathishavm: It was just inspired by Github API and some gems.
Where GitHub adds everything including links in the header. I prefer most frequently changing items like `links` should be the body and some constant values like.. `total_Recods`, `total_page` ..etc should be in the header.

app/controllers/concerns/json_pagination.rb
sathishavm: Generally, we can add anywhere,  but some of our API results are very large along with `includes` and `inlines` its very hard to find it out while viewing in UI and also it's hard to maintain the `links` always as the last entry.. something may end-up up append in between.

grosser: then how about:
```ruby
json = {links: links}.merge(json)
```

also document this reason why we are doing it

app/controllers/concerns/json_pagination.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+module JsonPagination
+  private
+
+  def add_json_pagination_links(json, pagy)
+    # add custom headers to the response
+    add_json_pagination_headers(pagy)
+
+    # Insert links at beginning
+    links = json_pagination_links(pagy)
```
grosser: make it symmetric and remove comments ? `add_json_pagination_links`

grosser:
please squasch commits before merging or merge with squasching

PR #3371, opened 2019-05-20
unify hook view locations

grosser:
stage_form being in fields etc made it not possible to search for `files that match stage_form` ... so now they all get the same name and file search in editor works
`Samson::Hooks.view :stage_form, 'samson_aws_sts/fields'`


adammw (approved):

adammw:
Is there any plugin versioning that you need to bump / note in a changelog?

grosser:
whoever does not use the new format will get a pretty descriptive error message ... not much docs anywhere for this :)

PR #3373, opened 2019-05-22
show kkubernetes namespace comment on index page

grosser:
Screen Shot 2019-05-22 at 11 32 54 AM

### DIFF
--- a/plugins/kubernetes/app/views/kubernetes/namespaces/index.html.erb
+++ b/plugins/kubernetes/app/views/kubernetes/namespaces/index.html.erb
@@ -12,7 +12,10 @@
 
       <% @kubernetes_namespaces.each do |namespace| %>
         <tr>
-          <td><%= namespace.name %></td>
+          <td>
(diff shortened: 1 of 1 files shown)


eatwithforks (approved):

plugins/kubernetes/app/views/kubernetes/namespaces/index.html.erb
```diff
@@ -12,7 +12,10 @@
 
       <% @kubernetes_namespaces.each do |namespace| %>
         <tr>
-          <td><%= namespace.name %></td>
+          <td>
+            <%= namespace.name %>
+            <%= additional_info simple_format(namespace.comment) if namespace.comment? %>
```
adammw: why isn't comment just a column in the table?

grosser: because it could be pretty big and also noisy to ahve it in the index view

PR #3374, opened 2019-05-22
sync all namespaces button

grosser:
press:
 - when a cluster was down
 - when new cluster was added
 - when creation randomly failed

Screen Shot 2019-05-22 at 12 16 02 PM
Screen Shot 2019-05-22 at 12 14 07 PM

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/config/routes.rb
+++ b/plugins/kubernetes/config/routes.rb
@@ -32,6 +32,13 @@ Samson::Application.routes.draw do
       end
     end
     resources :usage_limits, except: [:edit]
-    resources :namespaces, except: [:edit]
+    resources :namespaces, except: [:edit] do
+      member do
+        post :sync
+      end
+      collection do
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/config/routes.rb
```diff
@@ -32,6 +32,10 @@
       end
     end
     resources :usage_limits, except: [:edit]
-    resources :namespaces, except: [:edit]
+    resources :namespaces, except: [:edit] do
+      collection do
+        post :sync_all
```
adammw: also able to sync a single resource?

adammw (approved):
would like to see syncing a single namespace as well, but 🤷‍♂

plugins/kubernetes/config/routes.rb
grosser: added a `sync` button to the edit view too

PR #3376, opened 2019-05-23
Do not allow using a Deployment without spec.selector which would fal…

grosser:
…l back to all metadata.labels

could be changing labels like team etc that would make re-deploys fail

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -104,6 +104,7 @@ module Kubernetes
 
     # template_filler.rb sets name for everything except for IMMUTABLE_NAME_KINDS, keep_name, and Service
     # we make sure users dont use the same name on the same kind twice, to avoid them overwriting each other
+    # if they run in the default namespace
     def validate_name_kinds_are_unique
       # do not validate on global since we hope to be on namespace soon
       return if !@project || !@project.override_resource_names?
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -140,7 +141,7 @@ def validate_project_and_role_consistent
         kind = resource[:kind]
 
         label_paths = metadata_paths(resource).map { |p| p + [:labels] } +
-          if resource.dig(:spec, :selector, :matchLabels)
+          if resource.dig(:spec, :selector, :matchLabels) || resource[:kind] == "Deployment"
```
adammw: aren't there a few of these like that? worth having an array of kinds, ie. job, cronjob, replicaset, etc. ?

adammw (approved):

adammw:
would like a comment there why you check for deployment explicitly...

plugins/kubernetes/app/models/kubernetes/role_validator.rb
grosser: job: no, cronjob: no, replicaset: nobody deploys that ... so just leaves Deployment ... and that's were we saw the bug 🤷‍♂

grosser:
hmm just realized with "there" you meant in the code or in the PR ? :trollface:

PR #3379, opened 2019-05-28
PAAS-3357 add kritis breakglass option

grosser:
Screen Shot 2019-05-28 at 10 55 12 AM
Screen Shot 2019-05-28 at 10 52 13 AM

```
- apiVersion: extensions/v1beta1
  kind: Deployment
  metadata:
    name: example-server
    labels:
      project: example
      role: server
      team: foo
      kritis.grafeas.io/tutorial: ''
    namespace: sdfsfsfs
    annotations:
      samson/deploy_url: http://localhost:3000/projects/example-kubernetes/deploys/515
      kritis.grafeas.io/breakglass: 'true'
```

@zendesk/compute 
/cc @zendesk/samson

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_05_14_231932) do
+ActiveRecord::Schema.define(version: 2019_05_28_173915) do
 
   create_table "audits" do |t|
     t.integer "auditable_id", null: false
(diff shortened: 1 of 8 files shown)


adammw:
Feel like this should be on an individual deploy, not the deploy group...

grosser:
this is to disable it globally so we can roll out kritis without taking down the cluster and to turn it off in emergencies ... per deploy can come later if we need it, teams can also add it to their yaml files so the next deploy does not break again

adammw:
in that case, shouldn't it be per kubernetes cluster instead than, rather than deploy group?

grosser:
hmm yeah

grosser:
updated 💇‍♂  @adammw

hgokavarapuz:
Can you please paste the full screenshot where it appears in the Samson UI. Not clear from the current one.

grosser:
Screen Shot 2019-05-29 at 2 14 21 PM

hgokavarapuz (approved):

PR #3384, opened 2019-06-02
Added GitHub username in user profile

sathishavm:
Github Username is added in a user profile, it will allow creating a relation to GitHub account and Samson users signed up without GitHub.
Will allow to search with github username.

Screen Shot 2019-06-02 at 12 19 27 AM


test/controllers/profiles_controller_test.rb
```diff
@@ -24,6 +24,24 @@
         put :update, params: {user: {name: 'Hans'}}
         assert_template :show
       end
+
+      describe 'github username' do
```
grosser: these tests belong into the model

PR #3386, opened 2019-06-03
keep file diff additions/deletions aligned even if we have long renames

grosser:
reproduce: deploy example-kubernetes 71f4147 then grosser/diff

before:
image

after:
Screen Shot 2019-06-03 at 1 59 47 PM

@zendesk/compute


adammw:
although i think the first column should be narrower to match the style before...

grosser:
mostly just padding that was on the tds anyway 🤷‍♂ 
Screen Shot 2019-06-03 at 3 58 08 PM

PR #3387, opened 2019-06-03
show some indicator while fetching references so we know when it succ…

grosser:
…eeds

previously if it crashed or did not start it looked like all was good which is kinda dangerous

Jun-03-2019 15-47-38

@zendesk/compute @zendesk/bre


adammw:
was going to suggest a green checkmark but that works.. 🤷‍♂

grosser:
yeah checkmarks needs images and background is already set as style :D

PR #3388, opened 2019-06-04
PAAS-3493 enforce requests/limits on all containers that samson is not filling …

grosser:
…out to avoid swap/oomkill etc issues

this will fail a few deploys, so I'll search for them first and try to fix them before merging

```ruby
bad = []
Kubernetes::Role.find_each do |role|
  begin
    elements.each do |e|
      templates = [e, e.dig(:spec, :template), e.dig(:spec, :jobTemplate, :spec, :template)].compact
      containers = templates.flat_map do |t|
        all = ((t.dig(:spec, :containers) || [])[1..-1] || []) + (t.dig(:spec, :initContainers) || [])
        if json = t.dig(:metadata, :annotations, :"pod.beta.kubernetes.io/init-containers")
          all += JSON.parse(json, symbolize_names: true)
        end
        all
      end.flatten

      containers.each do |c|
        [
          [:resources, :requests, :cpu],
          [:resources, :requests, :memory],
          [:resources, :limits, :cpu],
          [:resources, :limits, :memory],
        ].each do |path|
          next if c.dig(*path)
          name = c[:name] || c[:image] || "unknown"
          bad << [role, c]
          puts "BAD #{project.permalink} #{role.config_file} #{name} is missing #{path.join(".")}"
        end
      end
    end
  rescue => e
    puts "Error #{e}"
    bad << [role]
  end
end
```

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -30,11 +30,12 @@ module Kubernetes
       validate_name_kinds_are_unique
       validate_single_primary_kind
       validate_api_version
-      validate_containers
+      validate_containers_exist
       validate_container_name
+      validate_container_resources
       validate_job_restart_policy
       validate_pod_disruption_budget
-      validate_numeric_limits
+      validate_numeric_cpu_limits
       validate_project_and_role_consistent
       validate_team_labels
       validate_not_matching_team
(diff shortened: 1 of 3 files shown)


hgokavarapuz (approved):
😍 very important change..

uthark:
Why can't we use https://kubernetes.io/docs/concepts/policy/limit-range/#enabling-limit-range ?

grosser:
Interesting addon :)

To start we only want to enforce "no unlimited requests/limits", this validation will show nicely in the UI and even before someone deploys, but sadly won't be a catch-all. Adding Limiter on top might be good if it can be configured to only enforce "no unlimited".

We don't want to set a default and we don't care about min/max per pod, just that the sum of the namespace stays under a total ... so I think we want ResourceQuota for that.

uthark:
LimitRange can also be useful for setting defaults per container.

uthark (approved):

PR #3391, opened 2019-06-06
fix ValidatingWebhookConfiguration not being deployable because it do…

grosser:
…es not have a namespace

longterm fix is to stop keeping a list an actually checking against the api :/

```
[15:49:08] JobExecution failed: Kubernetes error foo bar the server could not find the requested resource

```

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -5,10 +5,33 @@ module Kubernetes
     # not perfect since the actual rules are stricter
     VALID_LABEL_VALUE = /\A[a-zA-Z0-9]([-a-zA-Z0-9.]*[a-zA-Z0-9])?\z/.freeze # also used in js ... cannot use /i
 
+    # TODO: lookup dynamically
     NAMESPACELESS_KINDS = [
-      'APIService', 'ClusterRoleBinding', 'ClusterRole', 'CustomResourceDefinition', 'Namespace', 'PodSecurityPolicy',
+      'ComponentStatus',
+      'Namespace',
+      'Node',
+      'PersistentVolume',
+      'MutatingWebhookConfiguration',
+      'ValidatingWebhookConfiguration',
+      'CustomResourceDefinition',
+      'APIService',
+      'TokenReview',
+      'SelfSubjectAccessReview',
+      'SelfSubjectRulesReview',
+      'SubjectAccessReview',
+      'CertificateSigningRequest',
+      'PodSecurityPolicy',
+      'NodeMetrics',
+      'PodSecurityPolicy',
+      'ClusterRoleBinding',
+      'ClusterRole',
+      'PriorityClass',
+      'StorageClass',
+      'VolumeAttachment',
       'ValidatingWebhookConfiguration'
     ].freeze
+
+    # for non-namespace deployments: names that should not be changed since they will break dependencies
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -6,7 +6,8 @@ class RoleValidator
     VALID_LABEL_VALUE = /\A[a-zA-Z0-9]([-a-zA-Z0-9.]*[a-zA-Z0-9])?\z/.freeze # also used in js ... cannot use /i
 
     NAMESPACELESS_KINDS = [
-      'APIService', 'ClusterRoleBinding', 'ClusterRole', 'CustomResourceDefinition', 'Namespace', 'PodSecurityPolicy'
+      'APIService', 'ClusterRoleBinding', 'ClusterRole', 'CustomResourceDefinition', 'Namespace', 'PodSecurityPolicy',
```
uthark:
```suggestion
     'ComponentStatus',
'Namespace',
'Node',
'PersistentVolume',
'MutatingWebhookConfiguration',
'ValidatingWebhookConfiguration',
'CustomResourceDefinition',
'APIService',
'TokenReview',
'SelfSubjectAccessReview',
'SelfSubjectRulesReview',
'SubjectAccessReview',
'CertificateSigningRequest',
'ENIConfig',
'PodSecurityPolicy',
'NodeMetrics',
'PodSecurityPolicy',
'ClusterRoleBinding',
'ClusterRole',
'PriorityClass',
'StorageClass',
'VolumeAttachment',
```

uthark: kubectl api-resources -o wide

uthark (dismissed):
Let's update with other kinds too. Agree, that it should be dynamic in the future.

adammw (approved):
:+1: with oleg's changes. will eventually need to do this properly if we ever want to be compatible with cluster-scoped CRDs...

plugins/kubernetes/app/models/kubernetes/role_validator.rb
adammw: I would leave out ENIConfig since that's a custom CRD, but the rest :+1:

uthark: We already work with nonstandard APIGroups (i.e. ValidatingWebhookConfiguration comes from admissionregistration.k8s.io), by the way.  ENIConfig comes from crd.k8s.amazonaws.com API Group.

uthark: Removed ENIConfig from suggestion.

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -5,9 +5,33 @@ class RoleValidator
...
+      'CertificateSigningRequest',
+      'PodSecurityPolicy',
+      'NodeMetrics',
+      'PodSecurityPolicy',
+      'ClusterRoleBinding',
+      'ClusterRole',
+      'PriorityClass',
+      'StorageClass',
+      'VolumeAttachment',
+      'ValidatingWebhookConfiguration'
```
uthark: I see that you added it, but see line 15 too. Now it's duplicated. :)

grosser:
https://github.com/zendesk/samson/pull/3394

PR #3392, opened 2019-06-06
retry more often to avoid random errors

grosser:

### DIFF
--- a/plugins/kubernetes/lib/samson_kubernetes/samson_plugin.rb
+++ b/plugins/kubernetes/lib/samson_kubernetes/samson_plugin.rb
@@ -17,7 +17,7 @@ module SamsonKubernetes
   end
 
   def self.retry_on_connection_errors(&block)
-    Samson::Retry.with_retries connection_errors, 3, only_if: NOT_A_404, &block
+    Samson::Retry.with_retries connection_errors, 5, only_if: NOT_A_404, wait_time: 0.1, &block
   end
 end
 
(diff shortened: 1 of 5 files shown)


hgokavarapuz (approved):

plugins/kubernetes/lib/samson_kubernetes/samson_plugin.rb
```diff
@@ -17,7 +17,7 @@ def self.connection_errors
   end
 
   def self.retry_on_connection_errors(&block)
-    Samson::Retry.with_retries connection_errors, 3, only_if: NOT_A_404, &block
+    Samson::Retry.with_retries connection_errors, 5, only_if: NOT_A_404, wait_time: 0.1, &block
```
uthark: is 0.1 is enough?
what about exponential backoff?

plugins/kubernetes/test/controllers/kubernetes/namespaces_controller_test.rb
```diff
@@ -135,7 +135,7 @@
 
       it "shows creation errors" do
         assert_request(:get, "http://foobar.server/api/v1/namespaces/test", to_return: {status: 404}) do
-          assert_request(:post, "http://foobar.server/api/v1/namespaces", to_timeout: [], times: 4) do
+          assert_request(:post, "http://foobar.server/api/v1/namespaces", to_timeout: [], times: 6) do
```
uthark: I'd extract 5 and 6 to constants to improve clarity and to reduce further changes, if we want to change from 5 to 10.

grosser:
0.1 for now until we run into more trouble

plugins/kubernetes/test/controllers/kubernetes/namespaces_controller_test.rb
grosser: https://github.com/zendesk/samson/pull/3395

PR #3393, opened 2019-06-06
show deploy on release checkbox when a branch is configured, no need …

grosser:
…to wait for first release

I was confused why the box did not show up :/


yizhang-zen:
Silly questions: Will the deploy branches show up? Do we really need the deploy branches to choose from, why not waiting and use release?

grosser:
the release branch gets set in the project edit `master` mostly ... once it's set stages can start deploying when a new commit is pushed to that branch

yizhang-zen:
True. But will it confuse people when showing a bunch of deploy branches?

grosser:
this is not about deploy branches

PR #3395, opened 2019-06-06
document test logic and make changing retries not blow up tests

grosser:


uthark (approved):

yizhang-zen:
hmm..where do you document? Looks more like a refactor to me..Maybe add comments?

grosser:
document by using readable constants instead of magic numbers :)

PR #3396, opened 2019-06-06
say that tagged images are unsupported and stop requesting them from …

grosser:
…the api

sadly there is no nice api to resolve tags to shas, so for now let's declare them unresolvable instead of making futile requests and throwing errors

```
JobExecution failed: GCR scan result: Error retrieving vulnerabilities for <URL>
```

now it fails with
```
JobExecution failed: GCR scan result: Only full gcr repos with shas are supported for scanning for 
```

### DIFF
--- a/plugins/gcloud/lib/samson_gcloud/image_scanner.rb
+++ b/plugins/gcloud/lib/samson_gcloud/image_scanner.rb
@@ -5,8 +5,9 @@ module SamsonGcloud
     SUCCESS = 1
     FOUND = 2
     ERROR = 3
+    UNSUPPORTED = 4
 
-    FINISHED = [SUCCESS, FOUND].freeze
+    FINISHED = [SUCCESS, FOUND, UNSUPPORTED].freeze
 
     class << self
       # found by inspecting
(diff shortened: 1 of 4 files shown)


plugins/gcloud/lib/samson_gcloud/image_scanner.rb
```diff
@@ -49,7 +47,8 @@ def status(id)
       # NOTE: If we could guarantee ordering we could make a single request for
       # "(kind=\"DISCOVERY\" OR kind=\"PACKAGE_VULNERABILITY\")"
       def uncached_scan(image)
-        return ERROR unless image.include? SamsonGcloud.project
+        return UNSUPPORTED unless image.include? SamsonGcloud.project
+        return UNSUPPORTED unless image.include? "@" # only scanning shas is supported, not tags
```
hgokavarapuz: Instead of just returning just Unsupported would be better to return the error message why it is unsupported would be helpful.

plugins/gcloud/lib/samson_gcloud/image_scanner.rb
```diff
@@ -28,14 +29,11 @@ def result_url(image)
...
-        when FOUND
-          "Vulnerabilities found"
-        when ERROR
-          "Error retrieving vulnerabilities"
+        when WAITING then "Waiting for Vulnerability scan"
+        when SUCCESS then "No vulnerabilities found"
+        when FOUND then "Vulnerabilities found"
+        when ERROR then "Error retrieving vulnerabilities"
+        when UNSUPPORTED then "Only full gcr repos with shas are supported for scanning"
         else raise
```
uthark: can we display failing id in the exception?

plugins/gcloud/lib/samson_gcloud/image_scanner.rb
```diff
@@ -5,8 +5,9 @@ class ImageScanner
     SUCCESS = 1
     FOUND = 2
     ERROR = 3
+    UNSUPPORTED = 4
 
-    FINISHED = [SUCCESS, FOUND].freeze
+    FINISHED = [SUCCESS, FOUND, UNSUPPORTED].freeze
```
uthark: ERROR is also FINISHED, no?

uthark: or we do a retry?

plugins/gcloud/lib/samson_gcloud/image_scanner.rb
grosser: The message is shown per build, so it will be clear which build failed

plugins/gcloud/lib/samson_gcloud/image_scanner.rb
grosser: retry since it could be a temporary error

plugins/gcloud/lib/samson_gcloud/image_scanner.rb
uthark: ID is a phase in this case. And raise is to catch unsupported phase.

PR #3397, opened 2019-06-07
Deploy resources in kind order

adammw:
Deploy resources in kind order specified by https://github.com/helm/helm/blob/release-2.13/pkg/tiller/kind_sorter.go#L29

/cc @zendesk/samson

### Tasks
 - [ ] :+1: from team

### References
 - Jira link: https://zendesk.atlassian.net/browse/PAAS-3365

### Risks
- Level: Low/Med/High

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -43,12 +43,18 @@ module Kubernetes
     end
 
     def resources
-      @resources ||= resource_template.map do |t|
-        Kubernetes::Resource.build(
-          t, deploy_group,
-          autoscaled: kubernetes_role.autoscaled,
-          delete_resource: delete_resource
-        )
+      @resources ||= begin
+        resources = resource_template.map do |t|
+          Kubernetes::Resource.build(
+            t, deploy_group,
+            autoscaled: kubernetes_role.autoscaled,
+            delete_resource: delete_resource
+          )
+        end
+        resources.sort_by do |r|
+          Kubernetes::RoleConfigFile::DEPLOY_SORT_ORDER.index(r.kind) ||
+            Kubernetes::RoleConfigFile::DEPLOY_SORT_ORDER.size # default to maximum value
+        end
       end
     end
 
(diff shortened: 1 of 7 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -43,12 +43,15 @@ def resource_template
...
-        )
+      @resources ||= begin
+        resources = resource_template.map do |t|
+          Kubernetes::Resource.build(
+            t, deploy_group,
+            autoscaled: kubernetes_role.autoscaled,
+            delete_resource: delete_resource
+          )
+        end
+        resources.sort_by { |r| Kubernetes::RoleConfigFile::DEPLOY_SORT_ORDER.index(r.kind) }
```
grosser:
```suggestion
        resources.sort_by { |r| Kubernetes::RoleConfigFile::DEPLOY_SORT_ORDER.index(r.kind) || 999 }
```

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -20,7 +20,7 @@ class ReleaseDoc < ActiveRecord::Base
     attr_reader :previous_resources
 
     def deploy
-      @previous_resources = resources.map(&:resource)
+      @previous_resources = resources.reverse.map(&:resource)
```
grosser: We'd still want to deploy the configmaps before we deploy anything else etc ... so keep this ?

grosser:
the deploy_executor still deploys them in parallel, need to remove that too deploy_executor.rb:370 :(

plugins/kubernetes/app/models/kubernetes/release_doc.rb
adammw: used the size instead of a hardcoded number

adammw:
Is it worth partitioning by kind and doing parallel per kind or is that needless premature optimisation?

uthark (approved):

grosser:
premature ... we don't deploy that many things of the same kind in parallel

grosser (approved):
kinda sad that we mess with the oder the users gave us, but if that's what kubectl does then 🤷‍♂

adammw:
i mean, if it was doing it in parallel then the order would be messed up anyway..

grosser:
True 😄

PR #3401, opened 2019-06-14
better error messages and set_via_env fix

grosser:
... also avoids duplicated secret lookups and all other kinds of useless annotation copy

@zendesk/compute 
/cc @craig-day


craig-day (approved):

yizhang-zen (approved):
LGTM

PR #3407, opened 2019-06-20
add monitor linking and reuse faraday

grosser:
in preparation for adding more monitor logic ... I want 1 less abstraction layer and have logging for all http requests

Screen Shot 2019-06-19 at 7 10 05 PM

### DIFF
--- a/plugins/datadog/app/models/datadog_notification.rb
+++ b/plugins/datadog/app/models/datadog_notification.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
-require 'dogapi'
+require 'faraday'
 require 'digest/md5'
 
+# Note: might be able to replace this with Samson.statsd.event
 class DatadogNotification
   def initialize(deploy)
     @deploy = deploy
@@ -9,8 +10,6 @@ class DatadogNotification
   end
 
   def deliver(additional_tags: [], now: false)
-    Rails.logger.info "Sending Datadog notification..."
-
     status =
       if @deploy.active?
         "info"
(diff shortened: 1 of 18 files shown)


plugins/datadog/app/models/datadog_notification.rb
```diff
@@ -20,34 +20,23 @@ def deliver(additional_tags: [], now: false)
         "error"
       end
 
-    event = Dogapi::Event.new(
-      body,
-      msg_title: @deploy.summary,
-      event_type: "deploy",
-      event_object: Digest::MD5.hexdigest("#{Time.new}|#{rand}"),
```
grosser: no idea why we did that, so removing when in doubt 🤷‍♂

plugins/datadog/app/models/datadog_notification.rb
```diff
@@ -20,34 +19,19 @@ def deliver(additional_tags: [], now: false)
...
 
-  private
-
-  def body
-    "#{@deploy.user.email} deployed #{@deploy.short_reference} to #{@stage.name}"
-  end
-
-  def api_key
-    ENV["DATADOG_API_KEY"]
+    if response.status != 202
```
uthark: What about statuses
- 200 OK
- 201 Created?
https://docs.datadoghq.com/api/?lang=ruby#success-and-errors

grosser: yeah let's just say between 200 and 202 ... 202 is what datadog returns, but no need to hardcode this weirdness

grosser:
found a 3 places where we were not logging request urls, so that's nice too :)

yizhang-zen:
@grosser Are you going to add more monitor logic? Or is it ready for me to add monitors in Samson after this change rollout?

grosser:
more coming

PR #3414, opened 2019-06-24
resolve docker image tags so vulnerability scanning always works

grosser:
@zendesk/compute

### DIFF
--- a/plugins/gcloud/lib/samson_gcloud/tag_resolver.rb
+++ b/plugins/gcloud/lib/samson_gcloud/tag_resolver.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+#
+# https://cloud.google.com/sdk/gcloud/reference/container/images/describe
+# Could also be done via docker api https://docs.docker.com/registry/spec/api/#existing-manifests
+# but then needs a token-auth step.
+module SamsonGcloud
+  class TagResolver
+    class << self
+      def resolve_docker_image_tag(image)
+        return unless SamsonGcloud.gcr? image
+        return if image.match? Build::DIGEST_REGEX
+
+        success, json = Samson::CommandExecutor.execute(
+          "gcloud", "container", "images", "describe", image, "--format", "json",
+          *SamsonGcloud.cli_options,
+          err: '/dev/null',
+          timeout: 10,
+          whitelist_env: ["PATH"]
+        )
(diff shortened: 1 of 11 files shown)


plugins/gcloud/lib/samson_gcloud/tag_resolver.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    class << self
+      def resolve_docker_image_tag(image)
+        return unless SamsonGcloud.gcr? image
+        return if image.match? Build::DIGEST_REGEX
+
+        tag = image.split(":", 2)[1] || DEFAULT_TAG
+
+        success, json = Samson::CommandExecutor.execute(
+          "gcloud", "container", "images", "list-tags", image, "--limit", LIMIT.to_s, "--format", "json",
```
adammw: can't you use the container registry api directly rather than using something gcr specific?

adammw: https://docs.docker.com/registry/spec/api/#listing-image-tags

adammw: (that being said, using it against gcr requires docker auth credentials - not sure if you have them in samson or not already)

grosser: it works, is easy to debug locally, reuses the gcloud executable we already use in different places, does not need new credentials ... so lots of upsides :)

plugins/gcloud/lib/samson_gcloud/tag_resolver.rb
```diff
@@ -0,0 +1,34 @@
...
+    class << self
+      def resolve_docker_image_tag(image)
+        return unless SamsonGcloud.gcr? image
+        return if image.match? Build::DIGEST_REGEX
+
+        image, tag = image.split(":", 2)
+        tag ||= DEFAULT_TAG
+
+        success, json = Samson::CommandExecutor.execute(
+          "gcloud", "container", "images", "list-tags", image, "--limit", LIMIT.to_s, "--format", "json",
```
adammw: just use `gcloud container images describe <image_name>:<tag>` -> take `fully_qualified_digest`?

grosser: done

rbayerl (approved):

PR #3415, opened 2019-06-24
PAAS-3674 allow updating matchlabels when users know about manual cle…

grosser:
…anup

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -247,3 +247,14 @@ To debug services or to create resources that needs to reference a selector that
 ### Blocking LoadBalancer usage
 
 Set `KUBERNETES_ALLOWED_LOAD_BALANCER_NAMESPACES=foo,bar` to block all other namespaces from using them.
+
+### Updating matchLabels
+
+Samson will by default block updating `matchLabels` since it leads to abandoned pods.
+
+If you still want to change a matchLabel, for example project/role:
+
+ - set `metadata.annotations.samson/allow_updating_match_labels: "true"`
+ - deploy with renamed project/role
+ - manually delete pods from old Deployment (`kubectl delete pods -l project=old-name,role=old-role`)
+ - unset annotations from step 1
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -137,8 +137,11 @@ module Kubernetes
       end
 
       def ensure_not_updating_match_labels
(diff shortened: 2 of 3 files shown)


vvicaretti (approved):
👍

plugins/kubernetes/README.md
```diff
@@ -247,3 +247,14 @@ To debug services or to create resources that needs to reference a selector that
...
+### Updating matchLabels
+
+Samson will by default block updating `matchLabels` since it leads to abandoned pods.
+
+If you still want to change a matchLabel, for example project/role:
+
+ - set `metadata.annotations.samson/allow_updating_match_labels: "true"`
+ - deploy with renamed project/role
+ - manually delete pods from old Deployment (`kubectl delete pods -l project=old-name,role=old-role`)
+ - unset annotations from step 1
```
adammw: finda kunky.. would prefer a checkbox on the deploy, but `¯\_(ツ)_/¯`

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -137,9 +137,12 @@ def update
       end
 
       def ensure_not_updating_match_labels
-        # blue-green deply is allowed to do this, see template_filler.rb + deploy_executor.rb
+        # blue-green deploy is allowed to do this, see template_filler.rb + deploy_executor.rb
         return if @template.dig(:spec, :selector, :matchLabels, :blue_green)
 
+        # allow manual migration when user is aware of the issue and wants to do manual cleanup
+        return if @template.dig(:metadata, :annotations, :"samson/allow_updating_match_labels") == "true"
```
adammw: do you have somewhere where you're putting all these annotations as constants? would be good to know what samson is looking for in one place.

adammw (approved):

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: Readme has all the possible annotations :)

uthark:
If users use apps/v1 api group, they can't update matchLabels:

> Note: In API version apps/v1, a Deployment’s label selector is immutable after it gets created.

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates

grosser:
hoho, good thing we use extensions/v1beta1 everywhere :trollface:

grosser:
... but still, I want to add an escape-hatch for this usecase, if it's not changeable in kubernetes then we need a downtime anyway

PR #3417, opened 2019-06-25
improve redeploy on error logic

grosser:
- do not re-deploy the same sha when deploy failed
 - do not declare success when deploy failed
 - improve DeployService test coverage

### DIFF
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -48,7 +48,7 @@ class DeployService
     job_execution.on_finish { notify_outbound_webhooks(deploy) }
     job_execution.on_finish do
       if deploy.redeploy_previous_when_failed? && deploy.status == "failed"
-        redeploy_previous_succeeded(deploy, job_execution.output)
+        redeploy_previous(deploy, job_execution.output)
       end
     end
     job_execution.on_finish { Samson::Hooks.fire(:after_deploy, deploy, job_execution) }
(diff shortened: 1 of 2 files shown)


uthark:
> do not re-deploy the same sha when deploy failed

What about CPU/Memory request/limit changes? If I modify them, I should be able to try deploy again with the same SHA. (i.e. because error was of incorrect req/limits)

grosser:
deploying the old deploy would not revert that since we do not persist these values per deploy
only thing that might be worth checking is per-deploy env vars, but that's a underused feature anyway ...

adammw:
i think that's a vital feature (the env vars). if it's a configuration change that's the issue, it should revert that...

adammw:
don't you already have them serialised in the deploy?

grosser:
we don't use "deploy env vars" only thing we use is stage/project env vars and they are not stored per deploy

grosser:
we serialize some "env state" but that's not really deployable ... also deploy env vars change rarely and user has to opt-in to the redeploy anyway

grosser:
I'll see if I can improve this a bit, but either way this is better than the previous "just deploy the same sha again when it failed" logic 🤷‍♂

grosser:
yeah it's not straightforward, so I'll leave it like it is, not worse than a manual redeploy of an old sha

PR #3419, opened 2019-06-25
cleanup job execution flow a bit

grosser:
@zendesk/compute

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -11,8 +11,12 @@ class JobExecution
   delegate :id, :deploy, to: :job
   delegate :pid, :pgid, to: :executor
 
-  def initialize(reference, job, env: {}, output: OutputBuffer.new, &block)
-    @output = output
+  def initialize(reference, job, &block)
+    @job = job
+    @reference = reference
+    @execution_block = block
+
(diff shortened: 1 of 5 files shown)


app/models/job_execution.rb
```diff
@@ -11,8 +11,12 @@ class JobExecution
   delegate :id, :deploy, to: :job
   delegate :pid, :pgid, to: :executor
 
-  def initialize(reference, job, env: {}, output: OutputBuffer.new, &block)
-    @output = output
+  def initialize(reference, job, &block)
+    @job = job
+    @reference = reference
+    @execution_block = block
```
grosser: doing the basic  assignments first then the rest

app/models/job_execution.rb
```diff
@@ -191,17 +187,22 @@ def commands(dir)
...
-    # for shared notification scripts
-    env.merge!(
+      # for shared notification scripts
       PROJECT_NAME: @job.project.name,
       PROJECT_PERMALINK: @job.project.permalink,
       PROJECT_REPOSITORY: @job.project.repository_url
-    )
+    }
+
+    if @stage
```
grosser: adding code that used to be in deploy-service so all env stuff is in one place

PR #3420, opened 2019-06-26
mark deploys as failed when they trigger datadog monitors

grosser:
reuse existing rollback on failure feature to roll them back

 - low: during deploy stage edit page will fail while old code is running

```
[03:23:48] Testing for stability: 1s
[03:23:51] Testing for stability: 0s
[03:23:51] SUCCESS
[03:23:52] Alert on datadog monitors:
[03:23:52] sitemap refreshes missing https://foo.datadoghq.com/monitors/1805
```

@zendesk/compute 

also: send datadog errors to notifier instead of log which nobody looks at / do not blow up the deploy when datadog is down

### DIFF
--- a/app/models/deploy_service.rb
+++ b/app/models/deploy_service.rb
@@ -36,7 +36,9 @@ class DeployService
 
   def confirm_deploy(deploy)
     job_execution = JobExecution.new(deploy.reference, deploy.job)
-    job_execution.on_start { send_before_notifications(deploy, job_execution) }
+
+    job_execution.on_start { Samson::Hooks.fire(:before_deploy, deploy, job_execution) }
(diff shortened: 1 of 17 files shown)


uthark:
What if it alerts was spurious? I.e. It just flickered once during deployment?

grosser:
afaik we'd consider a monitor that recovers by the time the deploy is done ok

adammw:
Is this only during the deploy stability check or at any time?

grosser:
only during the deploy stability check, aka at the end of the deploy

grosser:
running intermittent or checking the alert history -> new story/PR

adammw:
yeah, just because i'd argue that we shouldn't do that if it's after the deploy finished.

PR #3421, opened 2019-06-26
allow scoping datadog monitors to current deploy-groups

grosser:
make deploys only fail when their deploy groups are alerting (will be even more useful once we do monitors per project) and makes monitor badges on the stage show page only show alert when their scope is alerting

Screen Shot 2019-06-26 at 4 03 35 PM


plugins/datadog/app/models/datadog_monitor_query.rb
```diff
@@ -1,24 +1,56 @@
...
+  end
+
+  def single_monitor?
+    query.match? /\A\d+\z/
+  end
+
   def validate_query_works
     return if errors[:query].any? # do not add to the pile
-    return if monitors.any? && monitors.all?(&:state) # tag search failed or the id search returned a bad monitor
+    return if monitors.any? && monitors.all? { |m| m.state([]) } # tag search failed or id search returned a bad monitor
```
uthark: why this version is better than previous?

grosser: it's the same, `state` just needs an argument now

plugins/datadog/app/models/datadog_monitor.rb
```diff
@@ -5,18 +5,20 @@ class DatadogMonitor
   API_KEY = ENV["DATADOG_API_KEY"]
   APP_KEY = ENV["DATADOG_APPLICATION_KEY"]
   SUBDOMAIN = ENV["DATADOG_SUBDOMAIN"] || "app"
+  BASE_URL = "https://#{SUBDOMAIN}.datadoghq.com"
```
adammw:
```suggestion
  BASE_URL = ENV["DATADOG_URL"]
```
since datadoghq.eu is also possible

plugins/datadog/app/models/datadog_monitor.rb
```diff
@@ -44,28 +46,44 @@ def initialize(id, response = nil)
...
-  end
-
   private
 
+  def match_value(deploy_group)
+    case match_source
+    when "deploy_group.permalink" then deploy_group.permalink
+    when "deploy_group.env_value" then deploy_group.env_value
+    when "environment.permalink" then deploy_group.environment.permalink
+    when "kubernetes_cluster.permalink"
```
adammw: is kubernetes a plugin? do you need to guard if that's enabled?

grosser: yes, but if someone does not have the kubernetes plugin and selects cluster matching then 💥 🤷‍♂

plugins/datadog/app/models/datadog_monitor_query.rb
```diff
@@ -1,24 +1,56 @@
...
+    else
+      q = URI.encode query.tr(",", " ") # rubocop:disable Lint/UriEscapeUnescape .to_query/CGI.escape do not work here
+      "#{DatadogMonitor::BASE_URL}/monitors/manage?q=#{q}"
+    end
+  end
+
   private
 
+  def validate_source_and_target
+    errors.add :match_source, "set target too or none." if match_source? ^ match_target?
```
adammw: i don't understand what this error even means

plugins/datadog/app/models/datadog_monitor.rb
grosser: oh boy

plugins/datadog/app/models/datadog_monitor_query.rb
```diff
@@ -1,24 +1,56 @@
...
 class DatadogMonitorQuery < ActiveRecord::Base
+  # add new handling code to datadog_monitor.rb when adding
+  MATCH_SOURCES = {
+    "Group Permalink" => "deploy_group.permalink",
+    "Group Env Value" => "deploy_group.env_value",
+    "Environment Permalink" => "environment.permalink",
+    "Cluster Permalink" => "kubernetes_cluster.permalink"
+  }.freeze
+
   belongs_to :stage, inverse_of: :datadog_monitor_queries
```
adammw: since this is on the stage, why not support match_source to be freetext matching, and interpolate specials like $KUBERNETES_CLUSTER_PERMALINK ?

that way if none of the permalinks match you can still use the feature.

plugins/datadog/app/models/datadog_monitor.rb
```diff
@@ -44,28 +46,44 @@ def initialize(id, response = nil)
...
     @response = nil
   end
 
-  def alert?
-    state == "Alert"
-  end
-
   private
 
+  def match_value(deploy_group)
```
adammw: what if you wanted two groups to match (say cluster and node type)?

plugins/datadog/app/models/datadog_monitor_query.rb
grosser: changed to `Datadog monitor queries match source cannot be set when target is not set.`

plugins/datadog/app/models/datadog_monitor.rb
grosser: make a new PR ;)

adammw: better to implement it right the first place so you don't have an ugly migration

plugins/datadog/app/models/datadog_monitor_query.rb
grosser: new PR ... don't need that for now I think

plugins/datadog/app/models/datadog_monitor.rb
adammw: or just cheat and serialise it into the column :p

plugins/datadog/app/models/datadog_monitor.rb
grosser: added

plugins/datadog/app/models/datadog_monitor.rb
grosser: all we need at is pod / cluster matching ... maybe env ... so idk why there should be more here ... what usecase do you have in mind ?

adammw: hmm, i was thinking more for shell based deploys like sfn_kubernetes where we need cluster+nodetype. do they use the same code? i guess it can always come later

grosser: we can always do comma-separated entries or something like that :)
going a bit too broad might not be that bad either, if something bad happens in the cluster you are deploying to that could be related ...

grosser:
... made the alert check a bit more efficient and also now ignoring alerts when they are silenced (datadog shows the monitor scope as "Alert" but does not send alerts :/ )

PR #3422, opened 2019-06-27
ignore silenced alerts to be consistent with datadog UI

grosser:
for the "fail deploy" feature it mostly does not matter since we ignore monitors that were alerting before the deploy and the UI will now be more consistent with what datadog shows

@zendesk/compute

### DIFF
--- a/plugins/datadog/app/models/datadog_monitor.rb
+++ b/plugins/datadog/app/models/datadog_monitor.rb
@@ -51,7 +51,7 @@ class DatadogMonitor
     return unless response[:overall_state] # show fallback as warning
 
     if match_source.present?
-      return "OK" unless alerting = alerting_tags.presence # be lazy
+      return "OK" unless alerting = alerting_tags.presence
       deployed = deploy_groups.map { |dg| "#{match_target}:#{match_value(dg)}" }
       (deployed & alerting).any? ? "Alert" : "OK"
     else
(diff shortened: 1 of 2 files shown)


adammw:
I'm confused, why all the deletion? is that the default out of box behaviour?

grosser:
removing it because I want the behavior to be simpler and closer to how datadog handles it, adding the silence filtering was a mistake

plugins/datadog/app/models/datadog_monitor.rb
```diff
@@ -73,19 +73,10 @@ def reload
...
-    alerting.reject! { |scope| silenced.any? { |s| (s & scope) == s } }
-    alerting.flatten(1)
-  end
-
-  # @return [Array<Array<String>>]
-  def tags_from_response(*path)
-    hash = response.dig(*path) || {}
-    hash.keys.map { |k| k.to_s.split(",").sort }
+    groups = response.dig(:state, :groups) || {}
+    groups.keys.flat_map { |k| k.to_s.split(",") }
```
hgokavarapuz: Do you want to sort them just to keep the behavior same?

hgokavarapuz:
But one thing I wanted to clarify, are we planning to use this for the rollback strategy ?

plugins/datadog/app/models/datadog_monitor.rb
grosser: I added sorting to make silencing work, so removing it now since it has no benefit

PR #3423, opened 2019-06-27
allow user to decide whether to rollback or fail when monitor is aler…

grosser:
…ting

not all users want to "deploy previous on every failure", so adding this option, reuses the existing rollback behavior

Screen Shot 2019-06-27 at 9 16 16 AM

```
[16:35:50] Alert on datadog monitors:
[16:35:50]  alarm🔒 https://zendesk.datadoghq.com/monitors/123
[16:35:50] Trying to redeploy previous succeeded deploy
[16:35:50] Deploy failed, attempting redeploy of previous succeeded deploy ...
[16:35:50] Previous succeeded deploy is the same reference 123
```

@zendesk/compute

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_06_26_163916) do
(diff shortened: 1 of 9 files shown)


uthark:
Nice feature 💚

PR #3424, opened 2019-06-27
allow checking datadog monitors for user-defined duration after deploy

grosser:
Screen Shot 2019-06-27 at 11 03 52 AM

@zendesk/compute 

```
No datadog monitors alerting 3 min remaining
No datadog monitors alerting 2 min remaining
Alert on datadog monitors:
Foo is down https://app.datadoghq.com/monitors/123
```

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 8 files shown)


adammw:
Does that mean the deploy will take that much longer?

grosser:
yep ... so better keep it :D

PR #3427, opened 2019-06-27
Pull allowed request origins from ENV

camertron:
Unless I'm much mistaken, Samson uses action cable now to tail your deploy log. We ran into this issue while upgrading today that was solved by setting the allowed action cable request origins in production.rb. It would be nice to be able to set them via the ENV.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -34,7 +34,7 @@ Samson::Application.configure do
   # Mount Action Cable outside main process or domain
   # config.action_cable.mount_path = nil
   # config.action_cable.url = 'wss://example.com/cable'
-  # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
+  config.action_cable.allowed_request_origins = ENV['RAILS_ALLOWED_REQUEST_ORIGINS'].split(',').map(&:strip)
 
   # Set to :debug to see everything in the log.
   config.log_level = :info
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -34,7 +34,7 @@
   # Mount Action Cable outside main process or domain
   # config.action_cable.mount_path = nil
   # config.action_cable.url = 'wss://example.com/cable'
-  # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
+  config.action_cable.allowed_request_origins = ENV['RAILS_ALLOWED_REQUEST_ORIGINS'].split(',').map(&:strip)
```
grosser: if it's not configured then this would block everything right ?
```ruby
if origins = ENV['RAILS_ALLOWED_REQUEST_ORIGINS'].to_s.split(',').map(&:strip).presence
  config.action_cable.allowed_request_origins = origins
end

```

camertron: True, nice :)

camertron: Actually I guess it would set `allowed_request_origins` to `nil`, which... might break things?

grosser:
yep, don't do that :)
... in general not sure why you have to set the origin anyway since it
should be the same as samsons origin

camertron:
@grosser yeah that's what we thought too... might have something to do with our nginx setup.

PR #3431, opened 2019-06-29
validate datadog tags match the selected monitors

grosser:
otherwise the monitors will never alert since the tags will never appear in the group-state

still not perfect since the computed tag values could still not be in the queries, but a good step forward ...

this means all monitors selected by a query must have a common tag, which is usually good anyway ...

Screen Shot 2019-06-28 at 9 18 09 PM

also added alert borders so it's clear which fields are invalid
Screen Shot 2019-06-28 at 9 30 12 PM

@zendesk/compute

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -57,3 +57,10 @@ a.btn.disabled[title] {
 .github-user-avatar {
   border-radius: .2em
 }
+
+// form elements not built by the form-builder have has-error divs directly around the form elements
+// color is same as `.has-error .form-control` from bootstrap
+.inline-errors .has-error {
+  border: 1px solid #a94442;
+  display: inline-block;
+}
--- a/plugins/datadog/test/models/datadog_monitor_query_test.rb
+++ b/plugins/datadog/test/models/datadog_monitor_query_test.rb
@@ -7,7 +7,7 @@ describe DatadogMonitorQuery do
   let(:query) { DatadogMonitorQuery.new(query: '123', scope: stages(:test_staging)) }
 
   describe "validations" do
-    def assert_id_request(to_return: {body: '{"overall_state":"OK"}'}, &block)
+    def assert_id_request(to_return: {body: {overall_state: "OK", query: "foo by {foo}"}.to_json}, &block)
       assert_request(
         :get,
         "#{api_url}/monitor/123?api_key=dapikey&application_key=dappkey&group_states=alert",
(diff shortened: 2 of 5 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -57,3 +57,10 @@ a.btn.disabled[title] {
 .github-user-avatar {
   border-radius: .2em
 }
+
+// form elements not built by the form-builder have has-error divs directly around the form elements
+// color is same as `.has-error .form-control` from bootstrap
+.inline-errors .has-error{
```
uthark:
```suggestion
.inline-errors .has-error {
```

plugins/datadog/test/models/datadog_monitor_query_test.rb
```diff
@@ -31,50 +31,102 @@ def assert_tag_request(response, &block)
...
+          query.match_source = "deploy_group.permalink"
+          assert_valid query
+        end
       end
-    end
 
-    it "is invalid with bad monitor tag" do
-      assert_id_request to_return: {status: 404} do
-        refute_valid query
+      it "does not allow source without target" do
```
uthark:
```suggestion
      it "does not allow missing target or source" do
```

plugins/datadog/test/models/datadog_monitor_query_test.rb
```diff
@@ -31,50 +31,102 @@ def assert_tag_request(response, &block)
...
-    it "is valid with good monitor multi-tags" do
-      assert_tag_request([{id: 123, overall_state: "OK"}]) do
-        assert_valid query
+      it "is invalid with bad monitor id" do
+        assert_id_request to_return: {status: 404} do
+          refute_valid query
+        end
+      end
+
+      it "is invalid with bad monitor tag" do
```
uthark: duplicated with previous.

plugins/datadog/test/models/datadog_monitor_query_test.rb
```diff
@@ -31,50 +31,102 @@ def assert_tag_request(response, &block)
...
+          refute_valid query
+        end
       end
-    end
 
-    it "does not allow source without target" do
-      assert_id_request do
-        query.match_target = "foo"
-        refute_valid query
+      it "is valid with good monitor multi-tags" do
```
uthark: Let's move it up.

plugins/datadog/test/models/datadog_monitor_query_test.rb
```diff
@@ -31,50 +31,102 @@ def assert_tag_request(response, &block)
...
+        query.query = "team:foo,team:bar"
+      end
+
+      it "does not make q request when query did not change" do
```
uthark:
```suggestion
      it "does not make request when query did not change" do
```

plugins/datadog/test/models/datadog_monitor_query_test.rb
```diff
@@ -31,50 +31,102 @@ def assert_tag_request(response, &block)
...
+        end
+
+        it "is valid when tag can be in group state" do
+          assert_id_request do
+            assert_valid query
+          end
+        end
+
+        it "can parse double quotes flavors" do
+          value = {body: {overall_state: "OK", query: "(foo).by(\"pod\",\"foo\").last(1)."}.to_json}
```
uthark:
```suggestion
          value = {body: {overall_state: "OK", query: '(foo).by("pod","foo").last(1).'}.to_json}
```

PR #3434, opened 2019-07-03
[CONFIG-430] samson_env: use S3 sdk to connect to config service

arecker:
Security requires the bucket be private, so we are updating clients to
connect to the config service bucket using the SDK, instead of plain
'ol HTTP.

### References
- Jira link: https://zendesk.atlassian.net/browse/CONFIG-430

### Risks
- Low: if `CONFIG_SERVICE_REGION` and `CONFIG_SERVICE_BUCKET` are not configured, samson will throw an error for projects in which the config service integration is enabled.

### DIFF
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'validates_lengths_from_database'
+require 'aws-sdk-s3'
 
 class EnvironmentVariable < ActiveRecord::Base
   FAILED_LOOKUP_MARK = ' X' # SpaceX
@@ -29,11 +30,12 @@ class EnvironmentVariable < ActiveRecord::Base
 
       if deploy_group && deploy.project.config_service?
         begin
-          url = config_service_folder(deploy.project, required: true)
-          url += "/#{deploy_group.permalink}.yml" # TODO: version ?
-          response = Samson::Retry.with_retries(Faraday::Error, 3) { Faraday.get(url) }
-          raise "Invalid response #{response.status}" unless response.status == 200
-          env.merge! YAML.safe_load(response.body)
+          path = "samson/#{deploy.project.permalink}"
+          path += "/#{deploy_group.permalink}.yml" # TODO: version ?
+          response = Samson::Retry.with_retries(Aws::S3::Errors::ServiceError, 3) do
+            config_service_read(path)
+          end
+          env.merge! YAML.safe_load(response)
         rescue StandardError => e
           raise Samson::Hooks::UserError, "Error reading env vars from config service: #{e.message}"
         end
@@ -64,10 +66,23 @@ class EnvironmentVariable < ActiveRecord::Base
       end.join("\n")
     end
 
-    def config_service_folder(project, required:)
-      url = ENV["CONFIG_SERVICE_URL"]
-      raise KeyError, "CONFIG_SERVICE_URL not set" if required && !url
-      "#{url}/samson/#{project.permalink}"
+    def config_service_url(project)
+      path = "samson/#{project.permalink}"
+      if bucket = ENV["CONFIG_SERVICE_BUCKET"]
+        path = "s3://#{bucket}/#{path}"
+      end
+      path
+    end
+
+    def s3_client
+      @s3_client ||= Aws::S3::Client.new(region: ENV.fetch('CONFIG_SERVICE_REGION'))
+    end
+
+    def config_service_read(key)
+      bucket = ENV.fetch('CONFIG_SERVICE_BUCKET')
+      s3_client.get_object(bucket: bucket, key: key).body.read
+    rescue Aws::S3::Errors::NoSuchKey
+      raise "key \"#{key}\" does not exist in bucket #{bucket}!"
     end
 
     private
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -29,16 +29,7 @@ class EnvironmentVariable < ActiveRecord::Base
       env = {}
 
       if deploy_group && deploy.project.config_service?
-        begin
-          path = "samson/#{deploy.project.permalink}"
-          path += "/#{deploy_group.permalink}.yml" # TODO: version ?
-          response = Samson::Retry.with_retries(Aws::S3::Errors::ServiceError, 3) do
-            config_service_read(path)
-          end
-          env.merge! YAML.safe_load(response)
-        rescue StandardError => e
-          raise Samson::Hooks::UserError, "Error reading env vars from config service: #{e.message}"
-        end
+        env.merge! env_vars_from_config_service(deploy, deploy_group)
       end
 
       if deploy_group && (env_repo_name = ENV["DEPLOYMENT_ENV_REPO"]) && deploy.project.use_env_repo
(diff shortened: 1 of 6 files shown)


grosser:
`Security requires the bucket be private` <- right ?

plugins/env/app/models/environment_variable.rb
```diff
@@ -64,13 +67,19 @@ def serialize(variables, env_scopes)
...
-    def config_service_folder(project, required:)
-      url = ENV["CONFIG_SERVICE_URL"]
-      raise KeyError, "CONFIG_SERVICE_URL not set" if required && !url
-      "#{url}/samson/#{project.permalink}"
+    def config_service_folder(project)
+      "samson/#{project.permalink}"
     end
 
-    private
+    def config_service_bucket(required:)
```
grosser: always called with `required: true` so don't need that argument

plugins/env/app/models/environment_variable.rb
```diff
@@ -64,13 +67,19 @@ def serialize(variables, env_scopes)
       end.join("\n")
     end
 
-    def config_service_folder(project, required:)
-      url = ENV["CONFIG_SERVICE_URL"]
-      raise KeyError, "CONFIG_SERVICE_URL not set" if required && !url
-      "#{url}/samson/#{project.permalink}"
+    def config_service_folder(project)
```
grosser: this is called with `required: false` from the UI ... best boot the app and see if it works/makes sense

arecker: Ahhh I found it.  Didn't realize this was getting called from an ERB template.

arecker:
@grosser 🙏 📿

plugins/env/app/models/environment_variable.rb
```diff
@@ -64,13 +66,17 @@ def serialize(variables, env_scopes)
...
+      "samson/#{project.permalink}"
     end
 
-    private
```
grosser: private needs to come back ... some of these methods are private ...

plugins/env/app/models/environment_variable.rb
```diff
@@ -64,13 +66,17 @@ def serialize(variables, env_scopes)
...
+    def config_service_bucket
+      bucket_name = ENV["CONFIG_SERVICE_BUCKET"]
+      region = ENV["CONFIG_SERVICE_REGION"]
+      raise KeyError, "CONFIG_SERVICE_BUCKET not set" unless bucket_name
```
grosser: use `ENV.fetch` to get the same error and less code 🤷‍♂

plugins/env/app/views/samson_env/_project_form.html.erb
```diff
@@ -15,11 +15,11 @@
     %>
   <% end %>
 
-  <% if defined?(EnvironmentVariable) && folder = EnvironmentVariable.config_service_folder(@project, required: false) %>
+  <% if defined?(EnvironmentVariable) && folder = EnvironmentVariable.config_service_folder(@project) %>
     <%= form.input(
         :config_service,
         label: "Use Config service",
-        help: "Use config service to manage deployment environment variables, base url is #{folder}"
+        help: "Use config service to manage deployment environment variables, S3 path is #{folder}"
```
grosser: should show the bucket and the folder to be obvious ?
something like `#{bucket}:#{folder}`

grosser:
check the plugins/env/Readme.md if it is still up to date

plugins/env/app/views/samson_env/_project_form.html.erb
arecker: I can use the whole S3 path, like `s3://bucket-name/folder/path`

grosser:
even better!

plugins/env/app/models/environment_variable.rb
```diff
@@ -64,14 +66,28 @@ def serialize(variables, env_scopes)
...
-      url = ENV["CONFIG_SERVICE_URL"]
-      raise KeyError, "CONFIG_SERVICE_URL not set" if required && !url
-      "#{url}/samson/#{project.permalink}"
+    def config_service_folder(project)
+      "samson/#{project.permalink}"
+    end
+
+    def config_service_label(project)
+      path = config_service_folder(project)
+      path = "s3://#{ENV["CONFIG_SERVICE_BUCKET"]}/#{path}" if ENV.key? "CONFIG_SERVICE_BUCKET"
```
grosser: FYI dry
```suggestion
      if bucket = ENV["CONFIG_SERVICE_BUCKET"]
        path = "s3://#{bucket}/#{path}"
      end
```

plugins/env/app/models/environment_variable.rb
```diff
@@ -64,14 +66,28 @@ def serialize(variables, env_scopes)
...
+      path
+    end
+
+    def config_service_bucket(stub_responses: false)
+      bucket_name = ENV.fetch("CONFIG_SERVICE_BUCKET")
+      region = ENV.fetch("CONFIG_SERVICE_REGION")
+      Aws::S3::Resource.new(region: region, stub_responses: stub_responses).bucket(bucket_name)
```
grosser: prefer
 - caching the s3 client
 - using the direct methods instead of the fancy object wrappers since they show which permissions are needed too

https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#get_object-instance_method

```
def config_service_read(key)
  @s3_client ||= Aws::S3::Client.new(region: region)
  s3_client.get_object(bucket: bucket_name, stub_responses: stub_responses).get_object(key).body.read
ebd
```

plugins/env/app/views/samson_env/_project_form.html.erb
```diff
@@ -15,11 +15,11 @@
     %>
   <% end %>
 
-  <% if defined?(EnvironmentVariable) && folder = EnvironmentVariable.config_service_folder(@project, required: false) %>
+  <% if defined?(EnvironmentVariable) && folder = EnvironmentVariable.config_service_label(@project) %>
```
grosser: _label -> folder + path -> folder for consistency ?

arecker: Clarification?

grosser: rename label to folder ... path/label/folder/key/bucket is too many words :D

arecker:
@tamgrosser tag! 🙏

plugins/env/app/models/environment_variable.rb
```diff
@@ -64,10 +66,23 @@ def serialize(variables, env_scopes)
...
+
+    def s3_client
+      @s3_client ||= Aws::S3::Client.new(region: ENV['CONFIG_SERVICE_REGION'])
+    end
+
+    def config_service_read(key)
+      bucket = ENV.fetch('CONFIG_SERVICE_BUCKET')
+      s3_client.get_object(bucket: bucket, key: key).body.read
+    rescue Aws::S3::Errors::NoSuchKey
+      raise "key \"#{key}\" does not exist!"
```
grosser:
```suggestion
      raise "key \"#{key}\" does not exist in bucket #{bucket!"
```

plugins/env/app/models/environment_variable.rb
```diff
@@ -64,10 +66,23 @@ def serialize(variables, env_scopes)
...
+    def config_service_folder(project)
+      path = "samson/#{project.permalink}"
+      if bucket = ENV["CONFIG_SERVICE_BUCKET"]
+        path = "s3://#{bucket}/#{path}"
+      end
+      path
+    end
+
+    def s3_client
+      @s3_client ||= Aws::S3::Client.new(region: ENV['CONFIG_SERVICE_REGION'])
```
grosser:
```suggestion
      @s3_client ||= Aws::S3::Client.new(region: ENV.fetch('CONFIG_SERVICE_REGION'))
```

plugins/env/app/views/samson_env/_project_form.html.erb
```diff
@@ -15,11 +15,11 @@
     %>
   <% end %>
 
-  <% if defined?(EnvironmentVariable) && folder = EnvironmentVariable.config_service_folder(@project, required: false) %>
+  <% if defined?(EnvironmentVariable) && folder = EnvironmentVariable.config_service_folder(@project) %>
```
grosser:
```suggestion
  <% if defined?(EnvironmentVariable) && path = EnvironmentVariable.config_service_folder(@project) %>
```

grosser:
all fixed up, let me know if you need help merging/deploying

arecker:
> all fixed up, let me know if you need help merging/deploying

Thank you!  Yes, we'd love help deploying this 🙏

Issue #3436, opened 2019-07-05
Schedule Webhook Deploys in Threads


vinayhegde90:
Hi Team,

We're using Samson (**v2488**) for our deployments for about 12 apps with 4-5 stages for each. For automated deploys, we trigger the **Generic** web-hook via a scheduled bash-script to all the stages every morning. Below is a small snippet of the same:

```
#!/bin/bash
curl -v http://$webhook_url_1/integrations/generic/$random_hash_1 -d @deploy.json --header "Content-Type: application/json"
sleep 180

curl -v http://$webhook_url_2/integrations/generic/$random_hash_2 -d @deploy.json --header "Content-Type: application/json"
sleep 180

curl -v http://$webhook_url_X/integrations/generic/$random_hash_X -d @deploy.json --header "Content-Type: application/json"
```
**The Issue**: While the script just fires a **curl** call to trigger the deployments and about 50% succeed, the remaining fail with errors like below:

```
[12:37:21] Your configuration specifies to merge with the ref 'refs/heads/master'
[12:37:21] from the remote, but no such ref was fetched.
```

Can someone here please help me with enable scheduling via an option in Samson or an alternate logic in the script such as **multi-threading** so that the triggered webhook finishes deploy to all stages successfully in a phased manner?

Do let me know if any more info is required.

grosser:
samson has `periodic deploys` which might help `PERIODICAL=...,periodical_deploy:86400,...` but it does them all at once

idk what these errors are or where they come from ... sounds like git problems :(

you could try using `curl -X POST foo.com/projects/foo/stages/bar/deploys` and see if that's better

vinayhegde90:
I already had a look at periodic deploys as you said but was confused by its tooltip. 

![samson-periodic-deploys](https://user-images.githubusercontent.com/43435196/60791154-ace3b200-a180-11e9-9824-df3b8ad7d2bf.png)

Please correct me if I'm wrong but periodic deploys will only happen every 24 hrs from the last deploy that should be successful otherwise there won't be any, thus triggering a failure email. 

> For e.g: Let's say if a **successful** deploy happens for 1 stage today at **7 PM**, then the periodic one will be triggered at **7 PM** on the **next day** right?

Also, is there a way to schedule this at a particular time?

grosser:
It schedules at the same time for all deploys, I'd think it would run
midnight UTC
I think it would be easy to add a kind of offset so it runs at the time you
want.

grosser:
[...]

PR #3438, opened 2019-07-08
show when next periodical deploy will take place

grosser:
for https://github.com/zendesk/samson/issues/3436

Screen Shot 2019-07-08 at 8 07 21 AM

@zendesk/compute

### DIFF
--- a/test/lib/samson/periodical_test.rb
+++ b/test/lib/samson/periodical_test.rb
@@ -196,6 +196,21 @@ describe Samson::Periodical do
     end
   end
 
+  describe ".next_execution_in" do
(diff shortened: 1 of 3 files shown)


test/lib/samson/periodical_test.rb
```diff
@@ -196,6 +196,21 @@ def call(*args)
     end
   end
 
+  describe ".next_execution_in" do
+    before { freeze_time }
+
+    it "shows next eecution time" do
```
uthark:
```suggestion
    it "shows next execution time" do
```

PR #3439, opened 2019-07-08
Resolve Omniauth CVE-2015-9284

grosser:
allows picking a provider of choice when signing up, only exploitable when multiple providers are enabled

see https://github.com/omniauth/omniauth/pull/809

testing: making a get request (removing the `method: :post` from the link) fails with 404

@zendesk/compute @uthark 🎉


uthark:
👍

PR #3443, opened 2019-07-10
hpas do not need a version to be updated

grosser:
more of a stopgap, real fix is making version update retry ... but this still helps 🤷‍♂ 

@zendesk/compute


aksgithub:
wondering why it's failing for hpa only. May be that's updated more frequently.

ankilosaurus (approved):

grosser:
statuses are updated constantly, this failed previously for other resources too

PR #3444, opened 2019-07-10
remove versionupdate logic since that seems to be fixed

grosser:
I tried with ApiService and CRDs against a 1.10 cluster and all worked out fine, so let's try removing this or find out where exactly it is still an 

## Risk
 - Med: will break deploys that still need this, simply revert and note which ones these are so we can document better / bring the logic back ... ideally improve it to fetch before retry too

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -260,16 +260,6 @@ module Kubernetes
       end
     end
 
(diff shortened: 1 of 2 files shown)


uthark (approved):

grosser:
CRDs seem to still need it `customresourcedefinitions.apiextensions.k8s.io "proxysqlclusters.proxysql.storage.zende.sk" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update`

PR #3451, opened 2019-07-12
fail early when cluster is missing to eliminate other sources of the bug

grosser:
we had this failing in `template_filler.rb" line 499 in static_env` which does not make any sense ... so maybe the template filler did something weird ... if this raise explodes then it's a db bug, if not the the filler does something wrong

@zendesk/compute


uthark (approved):

grosser:
also fixes a n+1 query by accident, so I guess we are keeping this 🤷‍♂

PR #3455, opened 2019-07-17
when deleting an autoscaled deployment, set the replicas to 0

grosser:
noticed this old bug when testing locally

@zendesk/compute


uthark:
Why we need to set it to 0, especially if we delete it?

uthark (approved):

eatwithforks (approved):

grosser:
a delete is an update to 0 and then we wait for the pods to go away

PR #3457, opened 2019-07-19
fix check to ignore hpa events

ankilosaurus:
Description:
Checks to ignore HPA warnings were done for incorrect resource type and due to that samson was not ignoring warnings from hpa

@zendesk/compute 
@grosser

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/horizontal_pod_autoscaler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/horizontal_pod_autoscaler.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+module Kubernetes
+  module Api
+    class HorizontalPodAutoscaler
+      IGNORED_AUTOSCALE_EVENT_REASONS = [
+        "FailedGetMetrics",
+        "FailedRescale",
+        "FailedGetResourceMetric",
+        "FailedGetExternalMetric",
+        "FailedComputeMetricsReplicas"
+      ].freeze
+
+      def events_indicating_failure(events)
+        events.reject { |e| IGNORED_AUTOSCALE_EVENT_REASONS.include? e[:reason] }
+      end
+    end
+  end
+end
--- a/plugins/kubernetes/app/models/kubernetes/resource_status.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource_status.rb
@@ -43,8 +43,8 @@ module Kubernetes
       elsif kind == "HorizontalPodAutoscaler"
         hpa = Kubernetes::Api::HorizontalPodAutoscaler.new
         failures = hpa.events_indicating_failure(events(type: "Warning"))
-        if failures && !failures.empty?
-          @details = "Error event \n#{failures.join("\n")}"
+        if failures.any?
+          @details = "Error event\n #{failures.join("\n")}"
         else
           @details = "Live"
           @live = true
--- a/plugins/kubernetes/app/models/kubernetes/resource_status.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource_status.rb
@@ -40,6 +40,16 @@ module Kubernetes
         else
           @details = "Waiting (#{@pod.phase}, #{@pod.reason})"
         end
+      elsif kind == "HorizontalPodAutoscaler"
+        hpa = Kubernetes::Api::HorizontalPodAutoscaler.new
+        failures = hpa.events_indicating_failure(events(type: "Warning"))
+        if failures && !failures.empty?
+          @details = "Error event \n#{failures.join("\n")}"
+        else
+          @details = "Live"
+          @live = true
+        end
+        @finished = true
       else
         # NOTE: non-pods are never "Missing" because we create them manually
         @finished = true
--- a/plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+require_relative '../../../test_helper'
+
+SingleCov.covered!
+
+describe Kubernetes::Api::HorizontalPodAutoscaler do
+  let(:start_time) { "2017-03-31T22:56:20Z" }
+  let(:event) { {metadata: {creationTimestamp: start_time}, type: 'Normal'} }
+  let(:events) { [event] }
+  let(:hpa) { Kubernetes::Api::HorizontalPodAutoscaler.new }
+
+  describe "HorizontalPodAutoscaler events with failures" do
+    before do
+      event.merge!(kind: 'HorizontalPodAutoscaler', type: 'Warning')
+    end
+
+    it "does not ignore bad events" do
+      refute hpa.events_indicating_failure(events).empty?
+    end
+
+    it "ignores failing to get metrics" do
+      event[:reason] = 'FailedGetMetrics'
+
+      assert hpa.events_indicating_failure(events).empty?
+    end
+
+    it "ingores failures to scale" do
+      event[:reason] = 'FailedRescale'
+
+      assert hpa.events_indicating_failure(events).empty?
+    end
+
+    it "ignores failing to get resource metrics" do
+      event[:reason] = 'FailedGetResourceMetric'
+
+      assert hpa.events_indicating_failure(events).empty?
+    end
+
+    it "ignores failing to get external metrics" do
+      event[:reason] = 'FailedGetExternalMetric'
+
+      assert hpa.events_indicating_failure(events).empty?
+    end
+
+    it "ignores failing to compute metrics replicas" do
+      event[:reason] = 'FailedComputeMetricsReplicas'
+
+      assert hpa.events_indicating_failure(events).empty?
+    end
+
+    it "does not ignore an unknown HPA event" do
+      event[:reason] = 'SomeOtherFailure'
+
+      refute hpa.events_indicating_failure(events).empty?
+    end
+  end
+end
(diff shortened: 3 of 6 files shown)


plugins/kubernetes/app/models/kubernetes/api/hpa.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+module Kubernetes
+  module Api
+    class Hpa
```
adammw: Would prefer HorizontalPodAutoscaler rather than abbreviations

plugins/kubernetes/app/models/kubernetes/resource_status.rb
```diff
@@ -40,6 +40,17 @@ def check
...
+        hpa = Kubernetes::Api::Hpa.new if @resource
+        failures = hpa.events_indicating_failure(events(type: "Warning"))
+        if !failures.empty?
+          @details = "Error event \n#{failures.join("\n")}"
+        else
+          @details = "Live"
+          @live = true
+        end
+        @finished = true
+
```
adammw: nit: no need for empty line

adammw (approved):

plugins/kubernetes/app/models/kubernetes/resource_status.rb
```diff
@@ -40,6 +40,17 @@ def check
         else
           @details = "Waiting (#{@pod.phase}, #{@pod.reason})"
         end
+      elsif kind == "HorizontalPodAutoscaler"
```
eatwithforks:
```suggestion
      elsif kind == "HorizontalPodAutoscaler" && @resource
```

plugins/kubernetes/app/models/kubernetes/resource_status.rb
```diff
@@ -40,6 +40,17 @@ def check
...
+        hpa = Kubernetes::Api::Hpa.new if @resource
+        failures = hpa.events_indicating_failure(events(type: "Warning"))
```
eatwithforks: this line will fail if there's no @resource because hpa will be undefined. so combine the @resource if check to the elsif on top

plugins/kubernetes/app/models/kubernetes/resource_status.rb
```diff
@@ -40,6 +40,17 @@ def check
...
+        if !failures.empty?
```
eatwithforks:
```suggestion
        unless failures.empty?
```

adammw: or `if failures && failures.length > 0`

plugins/kubernetes/app/models/kubernetes/resource_status.rb
aksgithub: could resource be invalid ?anyways that check is not needed here.

plugins/kubernetes/app/models/kubernetes/api/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,18 @@
...
+      IGNORED_AUTOSCALE_EVENT_REASONS = [
+        "FailedGetMetrics",
+        "FailedRescale",
+        "FailedGetResourceMetric",
+        "FailedGetExternalMetric",
+        "FailedComputeMetricsReplicas"
+      ].freeze
+
+      def events_indicating_failure(events)
+        (events - IGNORED_AUTOSCALE_EVENT_REASONS)
```
grosser:
```suggestion
        events.reject { |e| IGNORED_AUTOSCALE_EVENT_REASONS.include? e[:reason] }
```

plugins/kubernetes/app/models/kubernetes/resource_status.rb
```diff
@@ -40,6 +40,16 @@ def check
         else
           @details = "Waiting (#{@pod.phase}, #{@pod.reason})"
         end
+      elsif kind == "HorizontalPodAutoscaler"
+        hpa = Kubernetes::Api::HorizontalPodAutoscaler.new
+        failures = hpa.events_indicating_failure(events(type: "Warning"))
+        if failures && !failures.empty?
```
grosser: can use `if failures = hpa.events_indicating_failure(events(type: "Warning")).presence` to get rid of the empty/nil check

plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
```diff
@@ -0,0 +1,55 @@
...
+  let(:start_time) { "2017-03-31T22:56:20Z" }
+  let(:event) { {metadata: {creationTimestamp: start_time}, type: 'Normal'} }
+  let(:events) { [event] }
+  let(:hpa) { Kubernetes::Api::HorizontalPodAutoscaler.new }
+
+  describe "HorizontalPodAutoscaler with failures we don't care about" do
+    before do
+      event.merge!(kind: 'HorizontalPodAutoscaler', type: 'Warning')
+    end
+
```
grosser: start with a test that modifies nothing and asserts things are ok `it "does not ignore bad events"` or so, then each subsequent test changes 1 aspect and tests that the outcome change

plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    it "ignores failing to get resource metrics" do
+      event[:reason] = 'FailedGetResourceMetric'
+
+      refute hpa.events_indicating_failure(events).empty?, 'FailedGetResourceMetric must not be recognized as failures'
+    end
+
+    it "ignores failing to get external metrics" do
+      event[:reason] = 'FailedGetExternalMetric'
+
+      refute hpa.events_indicating_failure(events).empty?, 'FailedGetExternalMetric must not be recognized as failures'
```
grosser: I'd recommend leaving out these extra messages since if something goes wrong usually we have to dig in anyway and they get out of date quickly :shrug:
```suggestion
      refute hpa.events_indicating_failure(events).empty?
```

plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
```diff
@@ -0,0 +1,55 @@
...
+      event[:reason] = 'FailedComputeMetricsReplicas'
+
+      refute hpa.events_indicating_failure(events).empty?,
+        'FailedComputeMetricsReplicas must not be recognized as failures'
+    end
+
+    it "does not ignore an unknown HPA event" do
+      event[:reason] = 'SomeOtherFailure'
+
+      assert !hpa.events_indicating_failure(events).empty?,
```
grosser:
```suggestion
      assert hpa.events_indicating_failure(events).any?,
```

plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    it "ignores failing to get external metrics" do
+      event[:reason] = 'FailedGetExternalMetric'
+
+      refute hpa.events_indicating_failure(events).empty?, 'FailedGetExternalMetric must not be recognized as failures'
+    end
+
+    it "ignores failing to compute metrics replicas" do
+      event[:reason] = 'FailedComputeMetricsReplicas'
+
+      refute hpa.events_indicating_failure(events).empty?,
```
grosser: be consistent with the assertions ... prefer obvious assert/refute changes instead of subtle `!foo.empty?`

plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
```diff
@@ -0,0 +1,55 @@
...
+        'FailedComputeMetricsReplicas must not be recognized as failures'
+    end
+
+    it "does not ignore an unknown HPA event" do
+      event[:reason] = 'SomeOtherFailure'
+
+      assert !hpa.events_indicating_failure(events).empty?,
+        'Events we dont explicitly ignore must be recognized as failures'
```
uthark:
```suggestion
        'Events we don't explicitly ignore must be recognized as failures'
```

uthark (approved):

plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+require_relative '../../../test_helper'
+
+SingleCov.covered!
+
+describe Kubernetes::Api::HorizontalPodAutoscaler do
+  let(:start_time) { "2017-03-31T22:56:20Z" }
+  let(:event) { {metadata: {creationTimestamp: start_time}, type: 'Normal'} }
+  let(:events) { [event] }
```
grosser: events/event/start_time can be moved into `#events_indicating_failure ` test ... and can also remove the before and have the `event` be a warning instead

plugins/kubernetes/test/models/kubernetes/api/horizontal_pod_autoscaler_test.rb
```diff
@@ -0,0 +1,57 @@
...
+  let(:hpa) { Kubernetes::Api::HorizontalPodAutoscaler.new }
+
+  describe "HorizontalPodAutoscaler events with failures" do
```
grosser:
```suggestion
  describe "#events_indicating_failure" do
```

plugins/kubernetes/app/models/kubernetes/resource_status.rb
grosser: also failures will never be nil so `if failures.any?` would do too

grosser (approved):

PR #3463, opened 2019-07-30
Increase preStop delay for 35 seconds to allow DNS TTL to expire and die gracefully without spilling over requests

uthark:
### References
- Jira link: PAAS-3874

### Risks
- Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -9,6 +9,7 @@ module Kubernetes
     KUBERNETES_ADD_PRESTOP = Samson::EnvCheck.set?('KUBERNETES_ADD_PRESTOP')
     SECRET_PREFIX = "secret/"
     DOCKERFILE_NONE = 'none'
+    DEFAULT_TERMINATION_GRACE_PERIOD = 30
 
     def initialize(release_doc, template, index:)
       @doc = release_doc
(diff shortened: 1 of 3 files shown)


grosser:
- we should make this an ENV var so we don't hardcode values that work for us
 - we should set terminationgraceperiod to time + x so we don't sigkill the processes ... ideally set to "[user-configured, 35+x].max" ... but that get's fishy too ... making the DNS cache be <30s would be ideal so we don't have to think about terminationgraceperiod

uthark:
I've added environment variable. Reducing DNS TTL will help, but setting grace period doesn't harm.

grosser (approved):

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -569,11 +570,20 @@ def set_image_pull_secrets
...
       return unless KUBERNETES_ADD_PRESTOP
       pod_containers.each do |container|
         next if samson_container_config(container, :"samson/preStop") == "disabled"
-        (container[:lifecycle] ||= {})[:preStop] ||= {exec: {command: ["sleep", "3"]}}
+        # set sleep to 35 seconds to allow for DNS TTL to expire.
+        # Default termination grace period for pods
+        (container[:lifecycle] ||= {})[:preStop] ||= {exec: {command: ["sleep", pre_stop_sleep.to_s]}}
+      end
+      if pre_stop_sleep.to_i > DEFAULT_TERMINATION_GRACE_PERIOD
+        pod_template[:spec][:terminationGracePeriodSeconds] = pre_stop_sleep+3 # add 3 extra seconds before terminating.
```
grosser:
```suggestion
        pod_template[:spec][:terminationGracePeriodSeconds] = pre_stop_sleep + 3 # make sure servers shut down after they are done waiting
```

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -850,6 +850,12 @@ def secret_annotations(hash)
           )
         end
 
+        it "sets termination grace period if sleep time > default" do
```
grosser:
```suggestion
        it "bumps termination grace period if we would sleep longer that termination allows" do
```

PR #3466, opened 2019-07-31
allow more time for pod deletion and stop early if it failed

grosser:
before we waited 2*30s ... let's bump this a bit more to make sure it's not randomness ...
also fail when waiting did not work so the error is less confusing and we don't trigger an update if the old version could not be removed

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -380,11 +380,16 @@ module Kubernetes
       end
 
       def wait_for_termination_of_all_pods
-        30.times do
+        60.times do
           sleep TICK
           expire_resource_cache
           return if pods_count == 0
         end
+
+        raise(
+          Samson::Hooks::UserError,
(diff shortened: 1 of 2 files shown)


grosser:
not sure what to do if this still does not work, I'd hate the idea of "let's just update this thing that does not work" :/

eatwithforks (approved):

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -380,11 +380,16 @@ def pods_count
       end
 
       def wait_for_termination_of_all_pods
-        30.times do
+        60.times do
```
uthark: How will it work with RollingUpdate of the DaemonSet?

uthark (approved):

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: I'll look into the whole RollingUpdate/OnDelete business next, hoping this is a good quick-fix for the meantime

PR #3467, opened 2019-07-31
Transition jira issues on deploy

AlaricCalmette:
* This is a really rough first draft of a jira plugin that will transition linked jira issues at the end of a deploy.


grosser:
so we'd need:
 - a project level jira prefix field to store `EXPOLY`
 - a stage level perform this "transition" field ... do users know what the transition id is / do they need a guide to find it ?

what happens if a ticket is already in prod and then we deploy the PR to staging -> marked as "in staging again" ? ... is that made impossible by using the transition id ?

grosser:
I pushed some more code, let me know how that looks ... also if transition ids are integers and how to get them

AlaricCalmette:
> - a project level jira prefix field to store EXPOLY

    Yes, this is the project key on JIRA

> - a stage level perform this "transition" field ... do users know what the transition id is / do they need a guide to find it ?

    It's accessible in the edit workflow screen, so JIRA Admin can see it. Otherwise, we can fetch all the transitions of their project workflows and let them choose from a dropdown?

> - env vars for user+pass
> - a new "system" user so it's not you personally changing all these tickets

    There might already been one, we'll have to ask IT.

> what happens if a ticket is already in prod and then we deploy the PR to staging -> marked as "in staging again" ? ... is that made impossible by using the transition id ?

    It really depends on your JIRA workflow configuration. Transition have a start and end point. If the issue you're trying to transition is not in the start point, then nothing happens and an error saying that the transition couldn't be performed is returned. A issue relative to a pr already in prod should have an In Production status or equivalent, therefore making it impossible to transition TO this status.

AlaricCalmette:
The transition id is passed as a string in a JSON when using the REST API. I'll check the jira-ruby gem to see how they set it.

AlaricCalmette:
[JIRA doc](https://developer.atlassian.com/cloud/jira/platform/rest/v3/#api-rest-api-3-issue-issueIdOrKey-transitions-post) says it's a string

Screenshot 2019-07-31 at 15 05 48

grosser:
Sounds good ... Making it a select with available options can be a future
improvement, will need some caching and expiration too ... And validation
that the transition exists

AlaricCalmette:
Okay to have it later, once the plugin is set up i should be able to work on it by myself hopefully.

grosser:
@zendesk/compute @zendesk/bre

AlaricCalmette:
As far as my understanding and knowledge of Samson go, it looks great!
Way better than my bash hack :)

AlaricCalmette:
If you need me to test it somehow, I'll be glad to!

grosser:
already  tested it, just need some user/token so it does not run with my
account ;)

grosser:
@zendesk/compute ready for review

Gemfile.lock
```diff
@@ -89,6 +89,12 @@ PATH
     samson_jenkins (0.0.0)
       jenkins_api_client (~> 1.3)
 
+PATH
+  remote: plugins/jira
+  specs:
+    samson_jira (0.0.0)
```
uthark: should use released version, right?

db/schema.rb
```diff
@@ -429,6 +429,7 @@
     t.integer "kubernetes_rollout_timeout"
     t.integer "kubernetes_namespace_id"
     t.boolean "config_service", default: false, null: false
+    t.string "jira_issue_prefix"
```
uthark: what if issues are come from different JIRA projects?

plugins/jira/db/migrate/20190731123954_add_jjira_fields.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+class AddJjiraFields < ActiveRecord::Migration[5.2]
```
uthark:
```suggestion
class AddJiraFields < ActiveRecord::Migration[5.2]
```

plugins/jira/lib/samson_jira/samson_plugin.rb
```diff
@@ -0,0 +1,55 @@
...
+
+require 'base64'
+
+module SamsonJira
+  class Engine < Rails::Engine
+  end
+
+  def self.transition_jira_tickets(deploy, output)
+    return unless deploy.succeeded?
+    return unless url = ENV['JIRA_BASE_URL'].to_s[/(.*)\/[a-z]/, 1]
```
uthark: ENV vars should be validated on class creation + write in logs in something is missing? To improve debuggability.

Gemfile.lock
grosser: always stays at 0.0.0 since we don't version plugins (yet)

plugins/jira/lib/samson_jira/samson_plugin.rb
grosser: not having user/token configured is a valid usecase, so I don't want to spam the logs ...
if they are not set the prefix/transation ui is not shown, so that case should not happen 🤞

yizhang-zen:
I'd suggest to get eng-productivity team involved on pr reviewing as well

PR #3468, opened 2019-07-31
Use full path for sleep

uthark:
Found during classic deploy/undeploy: 

```
13m         Warning   FailedPreStopHook        Pod    Exec lifecycle hook ([sleep 3]) for Container "raingutter" in Pod "... " failed - error: command 'sleep 3' exited with 126: , message: "OCI runtime exec failed: exec failed: container_linux.go:345: starting container process caused \"exec: \\\"sleep\\\": executable file not found in $PATH\": unknown\r\n"
```

PATH is unset, so sleep is not found.

### Risks
- Low: /bin/sleep is a canonical path for it.


grosser (approved):

grosser:
I think we saw this a bunch of times ... good fix :D

grosser:
... might still happen for go, need to re-check this actually fixes it

uthark:
Yes, it will break for containers that are built from `scratch` or `distroless`.

PR #3471, opened 2019-08-02
Daemonsets cascade deletes so we can delete and then wait for pods to…

grosser:
… go away

- faster since we no longer do N+1 delete requests
- simpler
- no longer makes rollback->delete wait for cleanup

this drops support for extensions/v1beta1 which does not cascade deletes
(I want to get rid of supporting that anyway,  less things we need to test against)

testing:

```
dg = DeployGroup.find_by_name('Podabc')
r = Kubernetes::Resource::DaemonSet.new({kind: 'DaemonSet', apiVersion: 'apps/v1', metadata: {name: "example-server", namespace: "podabc"}}, dg, autoscaled: false, delete_resource: false)
r.send(:wait_for_termination_of_all_pods) { r.delete }
```

```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: example-server
  labels:
    project: example
    role: server
    team: foo
spec:
  selector:
    matchLabels:
      project: example
      role: server
  template:
    metadata:
      labels:
        project: example
        role: server
        team: foo
        release_id: xyz
        deploy_group_id: abc
    spec:
      containers:
      - name: default
        image: busybox
        command: ["/bin/sleep", "3600"]
        resources:
          requests:
            cpu: '0.1'
            memory: '100Mi'
          limits:
            cpu: '0.5'
            memory: '300Mi'
```

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -315,8 +315,9 @@ module Kubernetes
     end
 
     class DaemonSet < Base
+      # we cannot replace or update a daemonset, so we take it down completely
       def deploy
-        delete
+        wait_for_termination_of_all_pods { delete }
         create
       end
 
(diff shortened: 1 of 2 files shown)


grosser:
closing since old behavior works for both versions, is similar to what the Deployment does and also waits while deleting (prevents bugs when doing a delete - create)

PR #3472, opened 2019-08-02
support only RollingUpdate of DaemonSet

grosser:
testing with https://github.com/samson-test-org/example-kubernetes grosser/daemonset branch and it seems to work alright

there seem to be no restrictions on which fields can be updated either (like statefulset)
afaik we'll never need OnDelete so I'm locking things down to RollingUpdate and apps/v1 only to get rid of all the old code

also we need apps/v1 to make the DaemonSet delete request delete all the pods (for rollback / un-deploy)

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -315,11 +315,6 @@ module Kubernetes
     end
 
     class DaemonSet < Base
-      def deploy
-        delete
-        create
-      end
-
       # need http request since we do not know how many nodes we will match
       # and the number of matches nodes could update with a changed template
       # only makes sense to call this after deploying / while waiting for pods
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -352,53 +347,16 @@ def desired_pod_count
...
     class StatefulSet < Base
       def patch_replace?
         return false if @delete_resource || !exist?
+
+        # TODO: default is RollingUpdate, so this is wrong
         deprecated = @template.dig(:spec, :updateStrategy) # supporting pre 1.9 clusters
-        strategy = (deprecated.is_a?(String) ? deprecated : @template.dig(:spec, :updateStrategy, :type))
-        [nil, "OnDelete"].include?(strategy)
+        strategy = deprecated.is_a?(String) ? deprecated : @template.dig(:spec, :updateStrategy, :type)
+        [nil, "OnDelete"].include? strategy
```
uthark: RollingUpdate is default strategy, so it can be nil.

https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/#daemonset-update-strategy

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -241,6 +242,20 @@ def validate_stateful_set_service_consistent
       @errors << "Service metadata.name and StatefulSet spec.serviceName must be consistent"
     end
 
+    def validate_daemon_set_supported
+      return unless daemon_set = @elements.detect { |t| t[:kind] == "DaemonSet" }
+
+      if daemon_set.dig(:apiVersion) != "apps/v1"
```
uthark: This is not required, by the way, support for apps/v1 is different issue. But leave it here, anyway.

uthark:
How to override timeout for deployment? I.e we may expect it to take >10 minutes.

uthark (approved):

plugins/kubernetes/app/models/kubernetes/role_validator.rb
grosser: extensions/v1beta1 which we used before does not do any updates to the pod when used, so we have to force apps/v1

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: that's why there is a TODO above, not changing it in this PR

grosser:
- can speedup by using a higher maxunavailable
 - can set kubernetes_rollout_timeout per project

PR #3473, opened 2019-08-02
travis is still randomly slow

grosser:
@zendesk/compute

```
test/lib/samson/time_sum_test.rb:18]
{:db=>54.905500999999994}

```

### DIFF
--- a/test/lib/samson/time_sum_test.rb
+++ b/test/lib/samson/time_sum_test.rb
@@ -15,7 +15,7 @@ describe Samson::TimeSum do
     it "records single" do
       result = Samson::TimeSum.record("sql.active_record" => :db) { User.first }
       result.keys.must_equal [:db]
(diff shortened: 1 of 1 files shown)


test/lib/samson/time_sum_test.rb
```diff
@@ -15,7 +15,7 @@
     it "records single" do
       result = Samson::TimeSum.record("sql.active_record" => :db) { User.first }
       result.keys.must_equal [:db]
-      assert result[:db].between?(0, 10), result
+      assert result[:db].between?(0, 100), result
```
adammw: what does this test actually do?

grosser: make sure the time counting returns a sane value (ms vs s) so <1000 is fine 🤷‍♂

PR #3474, opened 2019-08-02
Fixup for #3463

uthark:
Prevent FailedPreStopHook when sleep duration is set exactly to 30 seconds.

### Risks
- Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -582,9 +582,10 @@ module Kubernetes
         # Default termination grace period for pods
         (container[:lifecycle] ||= {})[:preStop] ||= {exec: {command: ["/bin/sleep", pre_stop_sleep.to_s]}}
       end
-      if pre_stop_sleep.to_i > DEFAULT_TERMINATION_GRACE_PERIOD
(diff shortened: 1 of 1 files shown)


adammw (approved):
🤷‍♂

yizhang-zen (approved):

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -582,7 +582,7 @@ def set_pre_stop
         # Default termination grace period for pods
         (container[:lifecycle] ||= {})[:preStop] ||= {exec: {command: ["/bin/sleep", pre_stop_sleep.to_s]}}
       end
-      if pre_stop_sleep.to_i > DEFAULT_TERMINATION_GRACE_PERIOD
+      if pre_stop_sleep.to_i >= DEFAULT_TERMINATION_GRACE_PERIOD
```
grosser: technically if it's 29 we should still set to 32 to maintain the diff

```suggestion
      extra = 3
      if pre_stop_sleep.to_i + extra >= DEFAULT_TERMINATION_GRACE_PERIOD
        pod_template[:spec][:terminationGracePeriodSeconds] = pre_stop_sleep + extra
```

grosser (approved):

grosser (approved):

PR #3476, opened 2019-08-03
tell users to use maxUnavailable setting for DaemonSets so they do no…

grosser:
…t time out

@zendesk/compute 

finding existing with
```Ruby
bad = []
Kubernetes::Role.find_each do |role|
  begin
    elements.each do |e|
      next unless e[:kind] == "DaemonSet"
      next if e.dig(:spec, :updateStrategy, :rollingUpdate, :maxUnavailable)
      name = e[:name] || e[:image] || "unknown"
      bad << [role, e]
      puts "BAD #{project.permalink} #{role.config_file} #{e[:kind]} #{name} needs maxUnavailable"
    end
  rescue => e
    puts "Error #{e}"
    bad << [role]
  end
end

```

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -254,6 +254,12 @@ module Kubernetes
         @errors << "set DaemonSet spec.updateStrategy.type to RollingUpdate"
         return
       end
+
+      unless daemon_set.dig(:spec, :updateStrategy, :rollingUpdate, :maxUnavailable)
+        @errors << "set DaemonSet spec.updateStrategy.rollingUpdate.maxUnavailable, the default of 1 is too slow" \
+          " (pick something between '25%' and '100%')"
+        return
+      end
     end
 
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -254,6 +254,12 @@ def validate_daemon_set_supported
         @errors << "set DaemonSet spec.updateStrategy.type to RollingUpdate"
         return
       end
+
+      unless daemon_set.dig(:spec, :updateStrategy, :rollingUpdate, :maxUnavailable)
```
adammw: so manually setting it to 1 is the override.. sounds fine.

adammw (approved):

uthark:
I think, this is not a check that Samson should do.

grosser:
if it does not check this, then deploys will fail with timeout, since I already pathed all our daemonsets (and only our team uses them) it's not required 🤷‍♂

uthark (approved):

PR #3477, opened 2019-08-04
improve prestop hook docs a bit and always ensure sufficient terminat…

grosser:
…iongrace

also allowing users to opt-out completely since the new increased sleep is kinda annoying for most non-service deployments ... ideally change it so we either opt-in or only do it for roles with services :/

replaces/includes https://github.com/zendesk/samson/pull/3474

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -205,13 +205,16 @@ To make Samson leave your resource name alone, set `metadata.annotations.samson/
 
 ### Preventing request loss with preStop
 
-To enable the following functionality you need to set `KUBERNETES_ADD_PRESTOP=true`.
+When not using kubernetes servives to route requests, requests can be lost during a deployment,
+since old pods shut down before everyone all clients are refreshed.
 
-Samson automatically adds `container[].lifecycle.preStop` `sleep 3` if a `preStop` hook is not set and
-`container[].samson/preStop` is not set to `disabled`, to prevent in-flight requests from getting lost when taking a pod
-out of rotation (alternatively set `metadata.annoations.container-nameofcontainer-samson/preStop: disabled`).
+To prevent this, samson can automatically add `container[].lifecycle.preStop` `/bin/sleep <INT>`
+and increase the `spec.terminationGracePeriodSeconds` if necessary.
 
-Set `KUBERNETES_PRESTOP_SLEEP_DURATION` in seconds if you want to override default sleep duration.
+(will only add if `preStop` hook is not set and `container[].samson/preStop` is not set to `disabled`)
+
+- Set `KUBERNETES_ADD_PRESTOP=true` to enable
+- Set `KUBERNETES_PRESTOP_SLEEP_DURATION=30` in seconds to override default sleep duration (3 seconds)
 
 ### Showing logs on succeeded deploys
 
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -570,22 +570,28 @@ module Kubernetes
       pod_template.fetch(:spec)[:imagePullSecrets] = docker_credentials
     end
 
-    def pre_stop_sleep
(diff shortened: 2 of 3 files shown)


plugins/kubernetes/README.md
```diff
@@ -205,13 +205,16 @@ To make Samson leave your resource name alone, set `metadata.annotations.samson/
...
-`container[].samson/preStop` is not set to `disabled`, to prevent in-flight requests from getting lost when taking a pod
-out of rotation (alternatively set `metadata.annoations.container-nameofcontainer-samson/preStop: disabled`).
+To prevent this, samson can automatically add `container[].lifecycle.preStop` `/bin/sleep <INT>`
+and increase the `spec.terminationGracePeriodSeconds` if necessary.
 
-Set `KUBERNETES_PRESTOP_SLEEP_DURATION` in seconds if you want to override default sleep duration.
+(will only add if `preStop` hook is not set and `container[].samson/preStop` is not set to `disabled`)
+
+- Set `KUBERNETES_ADD_PRESTOP=true` to enable
+- Set `KUBERNETES_PRESTOP_SLEEP_DURATION=30` in seconds to override default sleep duration
```
uthark:
```suggestion
- Set `KUBERNETES_PRESTOP_SLEEP_DURATION=30` in seconds to override default sleep duration. Default sleep duration is 3 seconds.
```

uthark (approved):

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -570,21 +570,28 @@ def set_image_pull_secrets
...
+      sleep_time = Integer(ENV['KUBERNETES_PRESTOP_SLEEP_DURATION'] || '3')
+      containers.each do |container|
+        (container[:lifecycle] ||= {})[:preStop] ||= {exec: {command: ["/bin/sleep", sleep_time.to_s]}}
+      end
+
+      # shut down after prestop sleeping is done
+      buffer = 3
+      grace_period = pod_template[:spec][:terminationGracePeriodSeconds] || DEFAULT_TERMINATION_GRACE_PERIOD
+      if sleep_time + buffer > grace_period
+        pod_template[:spec][:terminationGracePeriodSeconds] = sleep_time + buffer
```
adammw: don't overwrite if manually set (like you do with prestop)?

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -850,10 +850,22 @@ def secret_annotations(hash)
           )
         end
 
-        it "bumps termination grace period if we would sleep longer that termination allows" do
-          with_env(KUBERNETES_PRESTOP_SLEEP_DURATION: "50") do
+        describe "when pod would terminate before finishing to sleep" do
+          with_env(KUBERNETES_PRESTOP_SLEEP_DURATION: "50")
+
+          it "bumps termination grace period" do
```
adammw: test doesn't overwrite existing preStop hook

adammw (approved):

plugins/kubernetes/app/models/kubernetes/template_filler.rb
grosser: I want to override  or crash since the config would be invalid ... and I figured override is simpler

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
grosser: `it "overrides termination grace period when user configured too low value" do`

grosser:
... if you feel strongly about override I'd make if fail when the config is invalid, but I doubt many users will hit this anyway 🤷‍♂

adammw:
strongly about overriding prestop, don't really care about termination grace period, unless it makes it lower.

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
grosser: we got `it "does not add preStop when it was already defined" do` for that

PR #3478, opened 2019-08-06
add option to show alternative user access request instructions

anthonywoo:
After talking with Compliance, we don't need to log user access changes in JIRA for individual projects. This adds an option to display an alternative instruction instead of using the access request form in Samson. If set, `ENV['REQUEST_ACCESS_ALT_INSTRUCTION']` will now appear when trying to access the request access form. If it is not set, the access request form will show.

I have also updated the Zendesk Samson confluence pages around this with information on how project admins can assign user permissions

Banner with example instructions:
Screen Shot 2019-08-06 at 3 17 36 PM

### Risks
- Low

### DIFF
--- a/app/helpers/access_request_helper.rb
+++ b/app/helpers/access_request_helper.rb
@@ -4,10 +4,6 @@ module AccessRequestHelper
     AccessRequestsController.feature_enabled? && current_user && !current_user.super_admin?
   end
 
-  def access_request_alternative_instruction
-    ENV['ACCESS_REQUEST_ALTERNATIVE_INSTRUCTION']
-  end
-
   def link_to_request_access
     if current_user.access_request_pending?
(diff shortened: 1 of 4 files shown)


app/helpers/access_request_helper.rb
```diff
@@ -4,6 +4,10 @@ def display_access_request_link?
     AccessRequestsController.feature_enabled? && current_user && !current_user.super_admin?
   end
 
+  def access_request_alternative_instruction
+    ENV['ACCESS_REQUEST_ALTERNATIVE_INSTRUCTION']
```
grosser: just inline this ? `<% if text = ENV['ACCESS_REQUEST_ALTERNATIVE_INSTRUCTION'] %>`

PR #3479, opened 2019-08-06
do not waste time with preStop when container has no ports

grosser:
@zendesk/compute


adammw (approved):

adammw:
_technically_, only needs to be pods pointed to by a service, but that's too difficult to get right. this is easier.

grosser:
yeah, we could say "role has a Service" but the Service does not need to be in the same role so kinda risky ...

uthark (approved):

PR #3480, opened 2019-08-08
use percentage based min-available to work nicely with HPAs

grosser:
removing the "convert % to number" logic we needed for old kubernetes versions
but keeping the "make it allow at least 1 eviction" logic

@vvicaretti @aymen-chetoui @zendesk/compute


vvicaretti (approved):
🎉

uthark (approved):
May we continue support both percentage and nubmer?

grosser:
number can make sense if for example the app needs at least 1 running, so I'd keep both

PR #3486, opened 2019-08-10
Introduce soak times for a Stage

jwswj:
I thought it would be interesting to play around with the concept of soak times in Samson so when someone is deploying a Stage they can understand how long the team normally waits for the Stage to soak before moving onto the next stage.

This is what it looks like:
soak-time-in-samson

### DIFF
--- a/app/assets/javascripts/channels/job_outputs_channel.js
+++ b/app/assets/javascripts/channels/job_outputs_channel.js
@@ -59,6 +59,7 @@ function startStream(userId, projectParam, jobId){
   function updateStatusAndTitle(data) {
     $('#header').load("/projects/" + projectParam + "/jobs/" + jobId + "?header=true", function () {
       timeAgoFormat(); // header includes new dates ... show them nicely instantly
+      $(this).trigger('statusAndTitleUpdated');
     });
     window.document.title = data.title;
   }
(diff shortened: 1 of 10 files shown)


grosser:
A few people suggested a feature like that, but I always imagined it being used for pipelines and not for manual deploys.
Kinda fancy, but complicated ... how about a simpler "Soak until X" info text instead of changing the button ? (then roll this out and see if anyone actually uses that ... then maybe make fancy if it starts getting adoption)
The UI should also say something about "Tell users to wait for X" so nobody expects this to work for pipelines.
If soak time is minutes and not hours, using the datadog monitor watching feature could replace this.

jwswj:
> so nobody expects this to work for pipelines
Ah. I wondered where deploy the next stage automatically had gone. Now I see the pipeline plugin.

> how about a simpler "Soak until X"
I really enjoy the countdown. It brings it to life. It does add a little complexity though.

Thanks for the suggestions @grosser. I was mostly just playing around here, I'll stew over it for a bit.

grosser:
If you have fun playing with the code, then keep going, otherwise a short
chat/planning session might be faster :)

catkins:
I like the countdown too! ❤️ 

What would make it super awesome for me:

- integration with pipelines, perhaps as a checkbox next to the automatically deploy stage bit
- either parameterized or conditional soak times so that you could specify a soak duration at deploy time (depending on the risk) or be able to configure rules that would infer it (possibly a bit fiddly)

PR #3488, opened 2019-08-14
automatically retry from config service DR bucket

aglover-zendesk:
* automatically retry from config service DR bucket

### References
- Jira link: https://zendesk.atlassian.net/browse/CONFIG-464

### Risks
- Low: May read slightly outdated configuration data


grosser:
why is that necessary ... scared of s3 being down ?

aglover-zendesk:
> why is that necessary ... scared of s3 being down ?

Basically yes. The alternative would be a manual procedure, where if S3 in us-east-1 were done/unavailable, we'd have to restart/redeploy Samson with `CONFIG_SERVICE_REGION` and `CONFIG_SERVICE_BUCKET` set to the DR-region values. That could be a viable alternative to an automated failover, but I'm assuming that manual failover operation would have to be executed by the Engineering Productivity folks.

What do you think?

PR #3492, opened 2019-08-20
inject a uniq label to ensure PDB targets deployment, not pod

craig-day:
Been seeing lots of deployments where the `PodDisruptionBudget` can't find a controller, so it generates a warning event cause Samson to abort the deploy. 

```
Warning NoControllers: found no controllers for pod "classic-sync-attachments-85869c4d4b-9x6d9" x2
Warning CalculateExpectedPodCountFailed: Failed to calculate the number of expected pods: found no controllers for pod "classic-sync-attachments-85869c4d4b-9x6d9" x2
```

This adds a new samson-specific label to `metadata.labels` and then to `spec.selector.matchLabels` of the PDB in the hopes that this will ensure the PDB targets the Deployment object, or really the ReplicationController it creates, rather than targeting individual Pod resources. 

This is mostly a theory, but if I define my own PDB and do this same thing, the deploys are reliably successful. 

### Risks
- Low. Just injecting a new label, but that could have unintended side-effects if someone is expecting an exact set of labels.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -114,9 +114,16 @@ module Kubernetes
         },
         spec: {
           minAvailable: target,
-          selector: {matchLabels: deployment.dig_fetch(:spec, :selector, :matchLabels).dup}
+          selector: {
+            matchLabels: deployment.
+              dig_fetch(:spec, :selector, :matchLabels).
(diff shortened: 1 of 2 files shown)


grosser:
if it's only failing during the deploy and not after the deploy then that should not be the problem ... so if the deploy fails  without rollback, is this even constantly produced by the PDB

grosser:
https://github.com/zendesk/samson/pull/3493 if it turns out to be a temporary error

PR #3502, opened 2019-08-28
PAAS-4013 make sure we do not use ineffective/misleading readOnlyRoot…

grosser:
…Filesystem directive

@uthark found that gotcha thanks to Intellij plugin 🎉 

Already made PRs for all our projects, so will give a week of grace period and then merge this 🤷‍♂ 
```ruby
bad = []
Kubernetes::Role.all.uniq { |r| [r.project_id, r.config_file] }.each do |role|
  begin
    elements.each do |e|
      next if !e.dig(:spec, :template, :spec, :securityContext, :readOnlyRootFilesystem) &&
        !e.dig(:spec, :securityContext, :readOnlyRootFilesystem)
      puts "BAD #{project.permalink} #{role.config_file} #{e[:kind]} #{e.dig(:metadata, :name)}"
    end
  rescue => e
    puts "Error #{e}"
    bad << [role]
  end
end

```

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -60,6 +60,7 @@ module Kubernetes
       validate_job_restart_policy
       validate_pod_disruption_budget
       validate_numeric_cpu_limits
+      validate_security_context
       validate_project_and_role_consistent
       validate_team_labels
       validate_not_matching_team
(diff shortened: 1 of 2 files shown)


uthark (approved):
I'd also put a link to k8s doc in the error message, but it's cosmetic.

grosser:
the docs page does not really  help ... it just lists all the available options which might even be more confusing, I like the error message as it is "move it to xyz" ... clear / direct 🤷‍♂

grosser:
ugh ... did not mean to merge that  ... I guess I'll lock the stage so nobody deploys until next week ...

PR #3503, opened 2019-08-28
Add clone button to stage show page

adammw:
Add clone button to stage show page
Screen Shot 2019-08-28 at 2 37 34 pm

### Risks
- Low: show page has broken link


grosser:
... not needed that often, but can be nice for setup I guess / a bit hidden normally

PR #3506, opened 2019-08-30
PAAS-4019 allow deleting with mismatched labels

grosser:
@zendesk/compute

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/resource_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/resource_test.rb
@@ -165,7 +165,7 @@ describe Kubernetes::Resource do
           end
         end
 
-        it "explains why it is a bad idea" do
+        it "allows updating when opting out" do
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -165,7 +165,7 @@ def assert_create_and_delete_requests(**args, &block)
           end
         end
 
-        it "explains why it is a bad idea" do
+        it "allows updating when opting out" do
           old = {spec: {selector: {matchLabels: {foo: "baz"}}}}
           resource.template[:metadata][:annotations] = {"samson/allow_updating_match_labels": "true"}
```
uthark: Please document this annotation in readme.

uthark (approved):

plugins/kubernetes/test/models/kubernetes/resource_test.rb
grosser: already documented

PR #3507, opened 2019-09-03
PAAS-3866 do not allow nginx ingress config for everyone

grosser:
@zendesk/compute 

found existing usages with
```ruby
bad = []
Kubernetes::Role.all.uniq { |r| [r.project_id, r.config_file] }.each do |role|
  begin
    elements.each do |e|
      next unless e[:kind] == "Ingress"
      (e.dig(:metadata, :annotations) || {}).each_key do |key|
        next unless key.to_s.start_with?("nginx.ingress.kubernetes.io/")
        puts "BAD #{project.permalink} #{role.config_file} #{e[:kind]} #{e.dig(:metadata, :name)} #{key}"
      end
      next if !e.dig(:spec, :template, :spec, :securityContext, :readOnlyRootFilesystem) &&
        !e.dig(:spec, :securityContext, :readOnlyRootFilesystem)
    end
  rescue => e
    puts "Error #{e}"
    bad << [role]
  end
end

```

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -263,3 +263,12 @@ If you still want to change a matchLabel, for example project/role:
  - deploy with renamed project/role
  - manually delete pods from old Deployment (`kubectl delete pods -l project=old-name,role=old-role`)
  - unset annotations from step 1
+
+### Blocking Ingress Nginx annotations
+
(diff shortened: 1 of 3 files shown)


adammw (approved):

uthark (approved):

uthark:
Also, I would say that people can expose their apps using Service with type = LoadBalancer.

grosser:
LoadBalancer usage is already restricted 👍

PR #3510, opened 2019-09-11
Allow duplicate builds for a single git sha

adammw:
Removes the uniqueness constraint for build sha to docker digest, so that newer builds for the same commit can be used and deployed, for example to pick up changes from a base image or internet download in one of the layers.

### References
- Jira link: 

### Risks
- Med: assumptions about docker builds broken where a multiple builds exist for a single git sha.

### DIFF
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -17,12 +17,15 @@ class Build < ActiveRecord::Base
   validates :project, presence: true
   validates :git_sha, allow_nil: true, format: SHA1_REGEX
   validates :dockerfile, presence: true, unless: :external?
-  [:dockerfile, :image_name].each do |scope|
-    [:git_sha, :external_url].each do |attribute|
+  [:git_sha, :external_url].each do |attribute|
+    [:dockerfile, :image_name].each do |scope|
       validates(
         attribute,
         allow_nil: true,
-        uniqueness: {scope: [scope, :external_url], message: "already exists with this #{scope} and external_url"},
+        uniqueness: {
+          scope: [:git_sha, scope, :external_url].without(attribute),
+          message: "already exists with this #{attribute} and #{scope}"
+        },
         if: ->(build) { build.send(scope).present? && build.external_url.present? }
       )
     end
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -17,13 +17,15 @@ class Build < ActiveRecord::Base
   validates :project, presence: true
   validates :git_sha, allow_nil: true, format: SHA1_REGEX
   validates :dockerfile, presence: true, unless: :external?
-  [:dockerfile, :image_name].each do |attribute|
-    validates(
-      :git_sha,
-      allow_nil: true,
-      uniqueness: {scope: attribute, message: "already exists with this #{attribute}"},
-      if: attribute
-    )
+  [:dockerfile, :image_name].each do |scope|
+    [:git_sha, :external_url].each do |attribute|
+      validates(
+        attribute,
+        allow_nil: true,
+        uniqueness: {scope: [scope, :external_url], message: "already exists with this #{scope} and external_url"},
+        if: ->(build) { build.send(scope).present? && build.external_url.present? }
+      )
+    end
   end
   validates :docker_repo_digest, format: DIGEST_REGEX, allow_nil: true
   validates :external_url, format: /\Ahttps?:\/\/\S+\z/, allow_nil: true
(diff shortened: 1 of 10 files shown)


db/migrate/20190911050155_alter_build_indexes.rb
```diff
@@ -0,0 +1,9 @@
+class AlterBuildIndexes < ActiveRecord::Migration[5.2]
+  def change
+    remove_index :builds, [:git_sha, :dockerfile]
+    remove_index :builds, [:git_sha, :image_name]
+
+    add_index :builds, [:git_sha, :dockerfile, :external_url], unique: true, length: {git_sha: 80, dockerfile: 80, external_url: 255}
```
grosser: I don't think this works since and index with null in it is never matching ... so this allows unlimited git_sha-X/dockerfile-Y/nil rows

lib/samson/build_finder.rb
```diff
@@ -90,7 +90,12 @@ def possible_builds
         commits.unshift previous if previous
       end
 
-      Build.where(git_sha: commits).sort_by { |build| commits.index(build.git_sha) }
+      Build.where(git_sha: commits).
+        group_by { |build| commits.index(build.git_sha) }.
+        sort_by { |index, _| index }.to_h.
+        transform_values { |builds| builds.sort_by!(&:updated_at).reverse }.
+        values.
+        flatten
```
grosser: can sort by index+updated_at and then uniq by sha ?

grosser:
why do we want to keep old builds if we can't deploy them anyway ... could just override them 🤷‍♂

db/migrate/20190911050155_alter_build_indexes.rb
adammw: that's what I want.

adammw:
Want to keep them for auditing, visibility, rollbacks, etc.

lib/samson/build_finder.rb
adammw: I struggled with working out the right syntax for that, could you write a diff for that?

grosser: `Build.where(git_sha: commits).sort_by { |build| [commits.index(build.git_sha), -build.updated_at.to_i]  }`

adammw:
Merge this and do the UI changes in a separate PR, or want to wait and merge them all at once?

app/models/build.rb
```diff
@@ -17,13 +17,15 @@ class Build < ActiveRecord::Base
...
-    validates(
-      :git_sha,
-      allow_nil: true,
-      uniqueness: {scope: attribute, message: "already exists with this #{attribute}"},
-      if: attribute
-    )
+  [:dockerfile, :image_name].each do |scope|
+    [:git_sha, :external_url].each do |attribute|
+      validates(
+        attribute,
```
grosser: `validate :external_url scope: [foo, :external_url]` seems wrong

db/migrate/20190911050155_alter_build_indexes.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AlterBuildIndexes < ActiveRecord::Migration[5.2]
+  def change
+    remove_index :builds, [:git_sha, :dockerfile]
+    remove_index :builds, [:git_sha, :image_name]
+
+    add_index :builds, [:git_sha, :dockerfile, :external_url],
+      unique: true, length: {git_sha: 80, dockerfile: 80, external_url: 255}
```
grosser: 191 is the max for multibyte

grosser: ... update the cleanliness test too so it fails with 255

test/controllers/builds_controller_test.rb
```diff
@@ -174,16 +174,58 @@ def create(attributes = {})
...
-        it 'updates a failed external build' do
+        it 'creates a new build when external url changes for the same git sha' do
+          assert_difference 'Build.count' do
+            create(
+              git_sha: build.git_sha,
+              external_status: 'succeeded',
+              external_url: 'https://blob.com/1235',
+              docker_repo_digest: digest,
+              dockerfile: build.dockerfile,
+              format: :json
```
grosser: put the attributes in a `let` so it's easier to see what changed

test/lib/samson/build_finder_test.rb
```diff
@@ -330,6 +330,19 @@ def execute
       finder.send(:possible_builds).must_equal [build]
     end
 
+    it "find builds ordered by most recent first" do
```
grosser: better create with 2 shas and updated_ats and then assert exact order

```
builds.map { |b| [b.git_sha, b.updated_at.year] }.must_equal [["a", 2019], ["a", 2017], ["b", 2020], ["a", 2019]]
```

PR #3511, opened 2019-09-11
Swanjr/use dr region when using dr bucket

swanjr:
* Needed to add DR region to the existing Config Service related code. It was using the wrong region with the DR bucket.

### References
- Jira link: https://zendesk.atlassian.net/browse/CONFIG-551

### Risks
- Low

### DIFF
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -101,19 +101,19 @@ class EnvironmentVariable < ActiveRecord::Base
       raise Samson::Hooks::UserError, "Error reading env vars from config service: #{e.message}"
     end
 
-    def config_service_s3_client
-      @config_service_s3_client ||= Aws::S3::Client.new(region: ENV.fetch('CONFIG_SERVICE_REGION'))
-    end
-
     def config_service_read_with_failover(key)
-      bucket = ENV.fetch "CONFIG_SERVICE_BUCKET"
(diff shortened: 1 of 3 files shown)


plugins/env/app/models/environment_variable.rb
```diff
@@ -101,19 +101,19 @@ def env_vars_from_config_service(deploy, deploy_group)
...
-      bucket = ENV.fetch "CONFIG_SERVICE_BUCKET"
-      dr_bucket = ENV.fetch "CONFIG_SERVICE_DR_BUCKET"
+      bucket = ENV.fetch 'CONFIG_SERVICE_BUCKET'
+      region = ENV.fetch 'CONFIG_SERVICE_REGION'
+      dr_bucket = ENV.fetch 'CONFIG_SERVICE_DR_BUCKET'
+      dr_region = ENV.fetch 'CONFIG_SERVICE_DR_REGION'
       Samson::Retry.with_retries(Aws::S3::Errors::ServiceError, 3) do
         begin
+          config_service_s3_client = Aws::S3::Client.new(region: region)
           config_service_s3_client.get_object(bucket: bucket, key: key).body.read
```
grosser: wait were did config_service_s3_client come from before ? ... need to delete a method ?

grosser: ah nvm I'm blind

PR #3513, opened 2019-09-12
Move table checks out of cleanliness and rewrite

adammw:
I threw up in my mouth a little bit when i saw all the regexes in here. This change splits these db schema tests out and actually executes the `db/schema.rb` file to get the schema, which should catch any edge cases.

### Risks
- Low: change in test only


test/integration/database_schema_test.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+SingleCov.not_covered!
+
+describe "database schema" do
+  attr_reader :table_definitions
+
+  before(:all) do
```
grosser: I don't think that works ... not sure what the first argument does though 🤷‍♂ 

`let_all ` would work ... but if it's not too slow it does not matter either

test/integration/database_schema_test.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  it "does not have boolean limit 1 in schema since this breaks mysql" do
+    table_definitions.
+      flat_map(&:columns).
+      select { |c| c.type == :boolean && c.options[:limit] == 1 }.
+      must_be_empty
+  end
+
+  it "does not have limits too big for postgres in schema" do
```
grosser:
```suggestion
  it "does not have limits too big for mysql in schema" do
```

grosser: .... maybe postgres, not sure

PR #3515, opened 2019-09-13
Revamp build search

davidreuss:
External status makes no sense, we have multiple states.

Added search possibilities for Git SHA, Git Reference and Docker digest.

### Risks
- Low: Search for builds could break

### DIFF
--- a/app/views/builds/index.html.erb
+++ b/app/views/builds/index.html.erb
@@ -6,21 +6,21 @@
   <div class="row">
     <%= search_form do %>
       <%= search_select :status, Job::VALID_STATUSES, size: 2 %>
-      <%= search_select :time_format, User::TIME_FORMATS, selected: params.dig(:search, :time_format) || current_user.time_format, label: 'Time', size: 1 %>
-      <% projects = Project.order(:name).pluck(:name, :id) %>
-      <%= search_select :project_id, projects, live: true, label: "Project" %>
+      <% unless @project %>
+        <% projects = Project.order(:name).pluck(:name, :id) %>
+        <%= search_select :project_id, projects, live: true, label: "Project" %>
+      <% end %>
       <div class="col-sm-2">
-        <%= label_tag :git_sha, "Git SHA (commit)" %><br/>
-        <%= text_field_tag 'search[git_sha]', params.dig(:search, :git_sha), class: "form-control" %>
-      </div>
-      <div class="col-sm-2">
-        <%= label_tag :git_ref, "Git Reference" %><br/>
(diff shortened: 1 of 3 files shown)


app/views/builds/index.html.erb
```diff
@@ -5,8 +5,21 @@
 <section class="tabs">
   <div class="row">
     <%= search_form do %>
-      <%= search_select :external, [["Yes", "true"], ["No", "false"]], size: 1 %>
       <%= search_select :time_format, User::TIME_FORMATS, selected: params.dig(:search, :time_format) || current_user.time_format, label: 'Time', size: 1 %>
+      <% projects = Project.order(:permalink).pluck(:permalink, :id) %>
```
grosser: prefer `name`, not `permalink`

app/views/builds/index.html.erb
```diff
@@ -5,8 +5,21 @@
 <section class="tabs">
   <div class="row">
     <%= search_form do %>
-      <%= search_select :external, [["Yes", "true"], ["No", "false"]], size: 1 %>
```
grosser: external was used to figure out how many of our builds were still built locally ... we might need that again if someone tries to find out if we still need the docker build host 🤷‍♂

grosser:
screenshot plz

davidreuss:
image

Added external status back in.

app/views/builds/index.html.erb
```diff
@@ -5,8 +5,22 @@
...
+      <%= search_select :external_status, Job::VALID_STATUSES, size: 1, label: "Status" %>
```
grosser: Calling it status when it's only external status could lead to confusion.
Either add a `help: 'External status'` or call it `Ext Status` or make the query correctly do the "job ? job.status : external_status" (hard)

davidreuss: yeah, rails is not really my thing 😂

grosser:
I can help ya if this rabbit hole is getting to deep ;)

davidreuss:
Yeah, probably going to need some help if you want it fully polished. It should be good now though, and i'm happy with the functionality as is.

Your call 👍 - otherwise a reapprove would be appreciated before merging :)

grosser:
added fixes:
 - do not show project search when already on a project page
 - allow searching by "status" which then searches for "internal status" (basic setup where samson builds itself) and external status
 - unify git search to be by commit and support sha+ref in a single field + instructions that we do not resolve

let me know if this looks good and I'll merge

davidreuss:
looks good!

nice fixes for my flailings :-)

grosser:
yeah it's a complicated area :D
... years of spaghetti ;)

PR #3516, opened 2019-09-16
new token for some-public-token user

grosser:
`GITHUB.commit('grosser/parallel', 'master')`

before: `401 - Bad credentials`
after: works

@zendesk/compute


uthark:
Questions - how to generate this token in a future? 
And I guess we need a metric for such failures to report in datadog.

grosser:
generated via "login as that user and then generate a new token" ... it's a fake user I created/own ... can share via internal lastpass or on-demand, but cannot put the password+2FA in the repo

reporting errors does not really work since this is the community-token ... but we can make the tests fail if the token is invalid ... that should not really happen, so the overhead might not be worth the effort

PR #3517, opened 2019-09-17
Build selection UI

adammw:
Screen Shot 2019-09-17 at 6 41 23 pm

Work in progress.

### References
- Jira link: 

### Risks

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -38,11 +38,11 @@ class BuildsController < ApplicationController
   def create
     new = false
     saved = false
-    external_build_has_digest = false
+    external_build_already_has_digest = false
 
     Samson::Retry.retry_when_not_unique do
       if registering_external_build? && @build = find_external_build
-        external_build_has_digest = @build.docker_repo_digest.present?
+        external_build_already_has_digest = @build.docker_repo_digest.present?
         @build.attributes = edit_build_params(validate: false)
       else
         @build = scope.new(new_build_params.merge(creator: current_user))
@@ -51,7 +51,7 @@ class BuildsController < ApplicationController
       new = @build.new_record?
       changed = @build.changed?
 
-      return head :unprocessable_entity if external_build_has_digest && changed
+      return head :unprocessable_entity if external_build_already_has_digest && changed
       saved = !changed || @build.save # nothing has changed or save result
     end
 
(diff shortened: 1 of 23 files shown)


app/controllers/builds_controller.rb
```diff
@@ -10,6 +10,13 @@ class BuildsController < ApplicationController
 
   def index
     @builds = scope.order('id desc')
+
+    if params[:reference].present?
+      git_sha = current_project.repository.commit_from_ref(params[:reference])
+      raise "Commit not found for reference: #{params[:reference]}" unless git_sha.present?
```
grosser: unknown ref should return empty results ?

app/helpers/builds_helper.rb
```diff
@@ -21,14 +21,6 @@ def git_ref_and_sha_for(build, make_link: false)
...
-    if job = build.docker_build_job
-      status_badge(job.status)
-    elsif build.docker_repo_digest
-      status_badge("succeeded")
-    elsif (Job::VALID_STATUSES - ["succeeded"]).include?(build.external_status)
-      status_badge(build.external_status)
-    else
-      "not built"
-    end
+    status_badge build.status
```
grosser: what will no status look like ?

app/models/build.rb
```diff
@@ -17,13 +17,18 @@ class Build < ActiveRecord::Base
...
   validates :dockerfile, presence: true, unless: :external?
-  [:dockerfile, :image_name].each do |attribute|
-    validates(
-      :git_sha,
-      allow_nil: true,
-      uniqueness: {scope: attribute, message: "already exists with this #{attribute}"},
-      if: attribute
-    )
+  [:git_sha, :external_url].each do |attribute|
+    [:dockerfile, :image_name].each do |scope|
```
grosser: that's already merged ? <-> rebase ?

adammw:
yeah, it is already lazy-loaded, but definitely thinking of not showing it unless more than one build is detected, this is just getting the valuable bit out the door... the js is doing the changing of the visible builds when the ref changes, the validation is still in the server-side, however some of it is spread between record validation and job execution since we don't actually know the resolved sha until the deploy is running.

the other problem is that the builds displayed aren't the same as what build finder would find, need to think about how to deal with that.

again, this is because with samson we're trying to deploy the code commit not the artefact that it makes this so ugly/difficult.

app/helpers/builds_helper.rb
adammw: build.status always returns a string, defaults to `"not built"` - same as before

app/models/build.rb
adammw: #3510 isn't merged yet, but yes they are the same commits, will disappear once merged.

grosser:
let's get that merged / tested / deployed before we go deeper :)

PR #3518, opened 2019-09-18
Changed session store from cookie store to cache store

sathishavm:
Fix For: Cookie overflow error for big flash notice.

Changed session store to cache store.
Cache store can be either memory or memcached.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -179,9 +179,8 @@ end
 # Must be here instead of in an initializer because plugin initializers run before app initializers
 # Used in plugins/airbrake + rollbar which do not support the 'foo.bar' syntax as rails does
 # https://github.com/airbrake/airbrake-ruby/issues/137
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -179,9 +179,8 @@ class Application < Rails::Application
 # Must be here instead of in an initializer because plugin initializers run before app initializers
 # Used in plugins/airbrake + rollbar which do not support the 'foo.bar' syntax as rails does
 # https://github.com/airbrake/airbrake-ruby/issues/137
-Samson::Application.config.session_key = :"_samson_session_#{Rails.env}"
```
grosser: keep this unless there is no `session` usage in any controller

sathishavm: Yup, I checked we are not using any sessions, there may be session used in some gems.

grosser:
FYI side-effect is that in local development you are always logged out after a restart

PR #3519, opened 2019-09-19
Validate datadog annotations

adammw:
Validate datadog annotations, if present. Mistyping an annotation with the incorrect container name can leave you wondering why it's not working for an hour. Fail fast.

@zendesk/compute @cholman-zd 

### Risks
- Low: breaks deployment of weird datadog-annotated deploys

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role_validator.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role_validator.rb
@@ -40,6 +40,8 @@ module Kubernetes
     # we either generate multiple names or allow custom names
     ALLOWED_DUPLICATE_KINDS = ((['Service'] + IMMUTABLE_NAME_KINDS)).freeze
 
+    DATADOG_AD_REGEXP = %r{(?:service-discovery|ad)\.datadoghq\.com/([^.]+)\.}.freeze
+
     def initialize(elements, project:)
       @project = project
       @elements = elements.compact
@@ -67,10 +69,11 @@ module Kubernetes
       validate_stateful_set_service_consistent
       validate_daemon_set_supported
       validate_load_balancer
-      validate_ingress_annotations_allowed
       unless validate_annotations
         validate_prerequisites_kinds
         validate_prerequisites_consistency
+        validate_datadog_annotations
+        validate_ingress_annotations_allowed
       end
       validate_env_values
       validate_host_volume_paths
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -178,6 +181,22 @@ def validate_api_version
...
     end
 
+    # validate datadog-specific annotations against
+    # https://docs.datadoghq.com/agent/autodiscovery/integrations/?tab=kubernetes#configuration
+    def validate_datadog_annotations
+      templates.each do |template|
+        pod_spec = pod_template(template)
+        annotations = pod_spec.dig(:metadata, :annotations) || {}
+        containers = pod_spec.dig(:spec, :containers) || []
+        dd_container_names = annotations.keys.map { |k| (DATADOG_AD_REGEXP.match(k) || [])[1] }.compact.uniq
```
grosser:
```suggestion
        dd_container_names = annotations.keys.map { |k| k[DATADOG_AD_REGEXP, 1] }.compact.uniq
```

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -178,6 +181,22 @@ def validate_api_version
...
+        spec_container_names = containers.map { |c| c[:name] }.compact
+
+        unless (invalid = dd_container_names - spec_container_names).empty?
```
grosser: could also do this ... but about the same kind of funky 🤷‍♂ 
```suggestion
        unless invalid = (dd_container_names - spec_container_names).presence
```

plugins/kubernetes/app/models/kubernetes/role_validator.rb
```diff
@@ -398,6 +417,17 @@ def validate_host_volume_paths
 
     # helpers below
 
+    def pod_template(template)
```
grosser: that's what `def templates` already does

grosser (commented):
I like the idea, annoying error to debug  :)

plugins/kubernetes/app/models/kubernetes/role_validator.rb
adammw: i don't understand how...

grosser: delegates to role_config_file.rb self.templates ... which does some magic to do it ... there are tests, so pretty sure it works :)

uthark (approved):

plugins/kubernetes/app/models/kubernetes/role_validator.rb
adammw: updated.

PR #3520, opened 2019-09-19
PAAS-3875 allow users to ignore kritis vulnerabilities per deploy

grosser:
Screen Shot 2019-09-19 at 2 41 59 PM

Screen Shot 2019-09-19 at 2 44 22 PM

@zendesk/compute @zendesk/secure

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_08_22_232319) do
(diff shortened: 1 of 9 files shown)


eatwithforks (approved):

uthark:
People may abuse it, and the whole fix looks like a bandaid.

uthark (approved):

grosser:
yeah, easy to abuse ... but so is putting the annotation directly into the yml files :)
more for unblocking atm so every new base image vuln does not bring all deploys to a standstill

PR #3523, opened 2019-09-25
Restrict namespaces to only those managed by samson

adammw:
Restrict namespaces to only those managed by samson.
Need a way to migrate existing applications which are deploying to other namespaces....

### References
- Jira link: https://zendesk.atlassian.net/browse/PAAS-4185

### Risks
- Low: breaks deployment of apps outside their namespace, potentially breaking some complex use cases


grosser:
looks about right ... now we just need to migrate all projects first ... use the crawler logic from https://github.com/zendesk/samson/pull/3507

adammw:
i ended up writing my own, good to know for next time. results are here: https://zendesk.atlassian.net/browse/PAAS-4213

PR #3524, opened 2019-09-30
PAAS-4154 show oomkill in deploy output since it does not have logs

grosser:
Screen Shot 2019-09-30 at 4 45 56 PM

reproduce by setting memory to 10m and deploying

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -32,8 +32,12 @@ module Kubernetes
         phase == 'Failed'
       end
 
-      def restarted?
-        @pod.dig(:status, :containerStatuses)&.any? { |s| s.fetch(:restartCount) > 0 }
+      def restart_details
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/api/pod.rb
```diff
@@ -32,8 +32,14 @@ def failed?
         phase == 'Failed'
       end
 
-      def restarted?
-        @pod.dig(:status, :containerStatuses)&.any? { |s| s.fetch(:restartCount) > 0 }
+      def restart_details
+        @pod.dig(:status, :containerStatuses)&.detect do |s|
```
uthark: Need more details - which container OOMKilled

uthark (approved):

plugins/kubernetes/app/models/kubernetes/api/pod.rb
grosser: done

PR #3525, opened 2019-10-02
PAAS-4052 fail when bad events happen on non-pod resources

grosser:
before: deploy runs until timeout before error is shown
Screen Shot 2019-10-02 at 12 47 42 PM

after: deploy fails immediately
Screen Shot 2019-10-02 at 12 56 45 PM

@zendesk/compute 
/cc @madhatter215

### Testing

deployed example_kubernetes with branch grosser/high which has a bad daemonset

### Risks
 - Low: deploys that previously passed fail because we ignored warnings on non-pods

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -60,12 +60,13 @@ module Kubernetes
 
       loop do
         statuses = resource_statuses(release_docs)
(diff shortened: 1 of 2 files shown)


adammw:
> we don't want to waste resources to check events

premature optimisation? i'd prefer to check everything...

grosser:
hmm yeah I guess ... we have deployments with hundreds of pods and that still works 🤷‍♂ 
... just realized that the "interesting" logic was reused to determine percentage of failure too, so will rework this a bit but check all

grosser:
all fixed up :D

adammw (approved):

PR #3527, opened 2019-10-03
short circuit commit status logic when commit was not found

grosser:
- avoid multiple ref -> commit resolutions
 - avoids 1-off logic for empty changesets / github api errors

### Risks
- Low: errors / bugs during commit status resolution

@zendesk/compute


grosser:
Now also treating unfound commit as fatal ... since the deploy would fail anyway
Screen Shot 2019-10-04 at 8 46 04 AM

PR #3505, opened 2019-08-29
More descriptive error messages

uthark:
Before fix, errors look  like this:

```
[22:41:56] Missing project or role for ClusterRoleBinding metadata.labels

[22:41:56] Missing project or role for ClusterRoleBinding metadata.labels

[22:41:56] Missing project or role for RoleBinding metadata.labels

[22:41:56] Missing project or role for ClusterRole metadata.labels

[22:41:56] Missing project or role for ClusterRoleBinding metadata.labels

[22:41:56] Missing project or role for ClusterRole metadata.labels

[22:41:56] Missing project or role for ClusterRoleBinding metadata.labels

[22:41:56] Missing project or role for ClusterRole metadata.labels

[22:41:56] Missing project or role for ClusterRoleBinding metadata.labels

[22:41:56] Missing project or role for ClusterRole metadata.labels

[22:41:56] Missing project or role for ClusterRoleBinding metadata.labels

```

### Risks
- Low

PR #3528, opened 2019-10-04
Add Kubernetes manifest preview feature

adammw:
Kubernetes manifest preview - get rendered manifest before deploying

```
# Manifest preview for Example-kubernetes - Master. Git ref: master (536826b62edc0c6990b4a05497f73b9df0f76ec0)
# 
---
apiVersion: v1
kind: Pod
metadata:
  name: example-migrate
  labels:
    project: example
    role: migrate
    team: foo
    release_id: ''
    deploy_group_id: '2'
    deploy_id: ''
    project_id: '2'
    role_id: '1'
    deploy_group: groupk
    revision: 536826b62edc0c6990b4a05497f73b9df0f76ec0
    tag: master
```

https://zendesk.atlassian.net/browse/PAAS-1699

### Risks
- Low: Kubernetes deploys break / are not persisted correctly

### DIFF
--- a/plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
+++ b/plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+class Kubernetes::StagesController < ResourceController
+  include CurrentProject
+  include CurrentStage
+
+  DEFAULT_BRANCH = "master"
+
+  before_action :authorize_resource!
+
+  def manifest_preview
+    # find ref and sha ... sha takes priority since it's most accurate
+    git_sha = params[:git_sha]
+    git_ref = params[:git_ref] || git_sha || current_project.release_branch || DEFAULT_BRANCH
+    git_sha ||= current_project.repository.commit_from_ref(git_ref)
+    if git_sha.empty?
+      return render body: '# Git reference not found', content_type: 'application/yaml'
+    end
+
+    # build a new deploy/job in memory to process the builds/templates
+    resolve_build = params[:resolve_build] == 'true'
+    job = Job.new(
+      deploy: current_stage.deploys.new(project: current_project, reference: git_ref || git_sha),
+      project: current_project,
+      user: current_user,
+      commit: git_sha
+    )
+    output = OutputBuffer.new
+    deploy_executor = Kubernetes::DeployExecutor.new(job, output)
+    release_docs = deploy_executor.preview(resolve_build: resolve_build)
+
+    template_fillers = release_docs.map(&:verification_template)
+    log = "# Manifest preview for #{current_project.name} - #{current_stage.name}. Git ref: #{git_ref} (#{git_sha})\n" +
+      yaml_comment(output.messages)
+    yaml = log + YAML.dump_stream(*template_fillers.map { |tf| tf.to_hash(verification: !resolve_build).deep_stringify_keys })
+    render body: yaml
+  rescue Samson::Hooks::UserError
+    render status: 400, body: yaml_comment($!.message)
+  end
+
+  private
+
+  def yaml_comment(message)
+    "# #{message.split("\n").join("\n# ")}\n"
+  end
+end
--- a/plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
+++ b/plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
@@ -6,6 +6,7 @@ class Kubernetes::StagesController < ResourceController
 
   before_action :authorize_resource!
 
+  # used internally for deploy tab preview as well as an external API to render manifests for external tools
   def manifest_preview
     git_ref = params[:git_ref] || current_project.release_branch || DEFAULT_BRANCH
     git_sha = current_project.repository.commit_from_ref(git_ref)
--- a/plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
+++ b/plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
@@ -4,35 +4,34 @@ class Kubernetes::StagesController < ResourceController
   include CurrentProject
   include CurrentStage
 
-  DEFAULT_BRANCH = "master"
-
   before_action :authorize_resource!
 
   def manifest_preview
-    # find ref and sha ... sha takes priority since it's most accurate
-    git_sha = params[:git_sha]
-    git_ref = params[:git_ref] || git_sha || current_project.release_branch || DEFAULT_BRANCH
-    git_sha ||= current_project.repository.commit_from_ref(git_ref)
-    if git_sha.empty?
-      return render body: '# Git reference not found', content_type: 'application/yaml'
-    end
+    git_ref = params[:git_ref] || current_project.release_branch || DEFAULT_BRANCH
+    git_sha = current_project.repository.commit_from_ref(git_ref)
+    raise Samson::Hooks::UserError, 'Git reference not found' unless git_sha.present?
+
+    # resolving builds can take some time, best to leave it off by default
+    resolve_build = params[:resolve_build] == 'true'
 
     # build a new deploy/job in memory to process the builds/templates
-    resolve_build = params[:resolve_build] == 'true'
     job = Job.new(
-      deploy: current_stage.deploys.new(project: current_project, reference: git_ref || git_sha),
+      deploy: current_stage.deploys.new(project: current_project, reference: git_ref),
       project: current_project,
       user: current_user,
       commit: git_sha
     )
     output = OutputBuffer.new
     deploy_executor = Kubernetes::DeployExecutor.new(job, output)
-    release_docs = deploy_executor.preview(resolve_build: resolve_build)
-
+    release_docs = deploy_executor.preview_release_docs(resolve_build: resolve_build)
     template_fillers = release_docs.map(&:verification_template)
+
+    # build yaml file with comment header for debugging
     log = "# Manifest preview for #{current_project.name} - #{current_stage.name}. Git ref: #{git_ref} (#{git_sha})\n" +
       yaml_comment(output.messages)
-    yaml = log + YAML.dump_stream(*template_fillers.map { |tf| tf.to_hash(verification: !resolve_build).deep_stringify_keys })
+    template_yaml = template_fillers.map { |tf| tf.to_hash(verification: !resolve_build).deep_stringify_keys }
+    yaml = log + YAML.dump_stream(*template_yaml)
+
     render body: yaml
   rescue Samson::Hooks::UserError
     render status: 400, body: yaml_comment($!.message)
--- a/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
@@ -20,9 +20,25 @@ module Kubernetes
       @reference = job.deploy.reference
     end
 
+    def preview(resolve_build: true)
+      verify_kubernetes_templates!
+      @release = build_release(resolve_build: resolve_build)
+      unless @release.valid?
+        raise Samson::Hooks::UserError, "Failed to store manifests: #{release.errors.full_messages.inspect}"
+      end
+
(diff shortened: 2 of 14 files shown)


plugins/kubernetes/config/routes.rb
```diff
@@ -16,6 +16,13 @@
       end
       resources :usage_limits, only: [:index]
     end
+    resources :stages do
```
grosser: hmm don't we want this in the kubernetes namespace so it's `/kubernetes/stages/foo` ?

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -84,10 +82,10 @@ def resource_release_id(release_doc)
     end
 
     # Creates a ReleaseDoc per DeployGroupRole
-    def create_release_docs(grouped_deploy_group_roles)
+    def build_release_docs(grouped_deploy_group_roles)
       grouped_deploy_group_roles.each do |dgrs|
-        dgrs.each do |dgr|
-          release_docs.create!(
+        dgrs.map do |dgr|
```
grosser:
```suggestion
        dgrs.each do |dgr|
```

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -20,9 +20,25 @@ def initialize(job, output)
...
     def execute(*)
       verify_kubernetes_templates!
-      @release = create_release
+      @release = build_release
+
+      Kubernetes::Release.transaction do
+        unless @release.save
+          raise Samson::Hooks::UserError, "Failed to store manifests: #{release.errors.full_messages.inspect}"
+        end
+      end
```
grosser: these should be some kind of method maybe and then be reused .. looks very similar

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+class Kubernetes::StagesController < ResourceController
+  include CurrentProject
+  include CurrentStage
+
+  DEFAULT_BRANCH = "master"
```
grosser: bet we got that somewhere already ... reuse would be nice

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  before_action :authorize_resource!
+
+  def manifest_preview
+    # find ref and sha ... sha takes priority since it's most accurate
+    git_sha = params[:git_sha]
+    git_ref = params[:git_ref] || git_sha || current_project.release_branch || DEFAULT_BRANCH
```
grosser: why do we need to pass sha and ref ... ref should be enough ?

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    git_sha ||= current_project.repository.commit_from_ref(git_ref)
+    if git_sha.empty?
+      return render body: '# Git reference not found', content_type: 'application/yaml'
```
grosser: status 400 ?

grosser (commented):
looks like it could work :D

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
adammw: yep but they seem all to be local to their specific controller. not sure where to put it

grosser:
build.rb or something like that :D

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
grosser: ... if this is for js then maybe ignore the 400 ... otherwise add a comment that it's about js not failing

plugins/kubernetes/app/views/samson_kubernetes/_deploy_changeset_tab_body.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if @stage.kubernetes? %>
+  <div class="tab-pane" id="manifests">
+    <!-- using an iframe to lazy-load the manifest preview as it is potentially expensive / slow -->
```
grosser: can we make this actually lazy-load (instead of in async parallel) ? ... not sure how the other tabs work

plugins/kubernetes/app/views/samson_kubernetes/_deploy_tab_nav.html.erb
```diff
@@ -1,3 +1,3 @@
-<% if @deploy.kubernetes %>
+<% if @deploy&.kubernetes || @stage&.kubernetes? %>
```
grosser: if the deploy was not done on kubernetes we should not show manifests ...

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    deploy_executor = Kubernetes::DeployExecutor.new(job, output)
+    release_docs = deploy_executor.preview(resolve_build: resolve_build)
+
+    template_fillers = release_docs.map(&:verification_template)
+    log = "# Manifest preview for #{current_project.name} - #{current_stage.name}. Git ref: #{git_ref} (#{git_sha})\n" +
+      yaml_comment(output.messages)
+    yaml = log + YAML.dump_stream(*template_fillers.map { |tf| tf.to_hash(verification: !resolve_build).deep_stringify_keys })
+    render body: yaml
+  rescue Samson::Hooks::UserError
+    render status: 400, body: yaml_comment($!.message)
```
uthark: How error message will look like?

plugins/kubernetes/config/routes.rb
adammw: ¯\_(ツ)_/¯ dealers choice

plugins/kubernetes/app/views/samson_kubernetes/_deploy_tab_nav.html.erb
adammw: don't understand...

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
adammw:
```
# Error found when validating kubernetes/migrate.yml
# securityContext.readOnlyRootFilesystem can only be set at the container level
```

uthark: Can we provide extra detail? kind/namespace/name of the object? Because user may have several containers/deployments in the same file.

adammw: that's up to the validator/reporter of the error, it's no worse than what validation people would get today, since it's using the same logic.

adammw: as in, separate PR

plugins/kubernetes/app/views/samson_kubernetes/_deploy_tab_nav.html.erb
adammw: possible for deploy to be nil.

grosser: we show the tab nav for old deploys too, so one edge-case is that an old deploy might be non-kubernetes, but the stage no longer is, then leading to issue ... but not that likely, so not important

would the deploy just be new and not nil maybe ?

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
uthark: I already did PR, but it wasn't merged. https://github.com/zendesk/samson/pull/3505

plugins/kubernetes/app/views/samson_kubernetes/_deploy_tab_nav.html.erb
adammw: updated to 
```
<% if @deploy ? @deploy.kubernetes : @stage&.kubernetes? %>
```

plugins/kubernetes/app/views/samson_kubernetes/_deploy_changeset_tab_body.html.erb
adammw: I thought you hated JS? The other tabs are rendered inline, and this comes from an XHR once you select your reference and click preview.

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
adammw: similarity is pretty small, prefer to leave it for now

uthark (approved):

adammw:
rebased, hopefully tests will pass still...

plugins/kubernetes/app/controllers/kubernetes/stages_controller.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class Kubernetes::StagesController < ResourceController
+  include CurrentProject
+  include CurrentStage
+
+  before_action :authorize_resource!
+
+  def manifest_preview
```
grosser: maybe say "Use internally for deploy tab and as external api"

plugins/kubernetes/app/views/samson_kubernetes/_deploy_changeset_tab_body.html.erb
```diff
@@ -1,14 +1,20 @@
-<% if @deploy.kubernetes %>
-  <div class="tab-pane" id="manifests">
-    <% if release = @deploy.kubernetes_release %>
-      <% release.release_docs.sort_by { |rd| DeployGroup.with_deleted{ rd.deploy_group&.name_sortable }}.each do |release_doc| %>
-        <div>
-          <h3><%= release_doc.deploy_group.name %> - <%= link_to release_doc.kubernetes_role.name, [@deploy.project, release_doc.kubernetes_role] %></h3>
-          <pre><%= release_doc.resource_template.map(&:deep_stringify_keys).map(&:to_yaml).join("\n") %></pre>
-        </div>
+<% if @deploy.persisted? %>
+  <% if @deploy.kubernetes %>
```
grosser: for consistency
```suggestion
  <% if @deploy.kubernetes? %>
```

plugins/kubernetes/app/views/samson_kubernetes/_deploy_changeset_tab_body.html.erb
```diff
@@ -1,14 +1,20 @@
...
+      <% else %>
+        Manifests not yet create, reload page.
       <% end %>
-    <% else %>
-      Manifests not yet create, reload page.
-    <% end %>
+  <% end %>
+<% elsif @stage&.kubernetes? %>
+  <div class="tab-pane" id="manifests">
+    <!-- using an iframe to lazy-load the manifest preview as it is potentially expensive / slow -->
```
grosser:
```suggestion
    <!-- using an iframe to lazy-load the manifest preview as it can be slow -->
```

plugins/kubernetes/app/views/samson_kubernetes/_deploy_changeset_tab_nav.html.erb
```diff
@@ -1,3 +1,3 @@
-<% if @deploy.kubernetes %>
+<% if @deploy ? @deploy.kubernetes : @stage&.kubernetes? %>
```
grosser: revert ?

grosser (approved):

plugins/kubernetes/app/views/samson_kubernetes/_deploy_changeset_tab_nav.html.erb
adammw: still needed (although updated to @deploy.persisted) since the new deploy doesn't have kubernetes set.

PR #3530, opened 2019-10-04
use api hooks to not break gitlab + simplify them

grosser:
- fix caching logic resetting `@reference` which could remove open PRs when done in the wrong order
 - cache all branch comparisons on their SHA since they can change at any time
 - do not load open pull requests for main branches like "master" since they are never used as PR

followup to https://github.com/zendesk/samson/pull/3527 where I saw all these weirdnesses

@zendesk/compute

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 class Changeset
-  attr_reader :project, :repo, :previous_commit, :reference # used by plugins
-
-  BRANCH_TAGS = ["master", "develop"].freeze
-  ATTRIBUTE_TABS = %w[files commits pull_requests risks jira_issues].freeze
+  ATTRIBUTE_TABS = ["files", "commits", "pull_requests", "risks", "jira_issues"].freeze
+  MAIN_BRANCHES = ["master", "develop", "staging", "production"].freeze
 
   def initialize(project, previous_commit, reference)
     @project = project
@@ -33,7 +31,7 @@ class Changeset
   end
 
   def pull_requests
-    @pull_requests ||= find_pull_requests
+    @pull_requests ||= merged_pull_requests + open_pull_requests
   end
 
   def risks?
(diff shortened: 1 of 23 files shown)


app/models/changeset.rb
```diff
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 class Changeset
-  attr_reader :project, :repo, :previous_commit, :reference # used by plugins
-
-  BRANCH_TAGS = ["master", "develop"].freeze
-  ATTRIBUTE_TABS = %w[files commits pull_requests risks jira_issues].freeze
+  ATTRIBUTE_TABS = ["files", "commits", "pull_requests", "risks", "jira_issues"].freeze
+  MAIN_BRANCHES = ["master", "develop", "staging", "production"].freeze
```
adammw: eww. there has to be a better way...

grosser: catches the 99% ... good enough 🤷‍♂

grosser: ... also it's just an optimization to avoid lookups, so does not matter if it misses a few

app/models/changeset.rb
```diff
@@ -67,53 +66,54 @@ def error
...
+
+    [self.class, @repo, @previous_commit, key].join('-')
   end
 
   # github only supports finding open PRs for branches (not commits or tags)
   # https://help.github.com/en/articles/searching-issues-and-pull-requests
-  def find_pull_requests_for_branch
-    return [] if not_pr_branch?
+  def open_pull_requests
+    return [] if static? || MAIN_BRANCHES.include?(@reference)
```
adammw: is @reference guaranteed to be a branch here?

adammw: what if it's a tag?

grosser: `static?` checks that
... it's just optimization to avoid doing the request

plugins/gitlab/lib/samson_gitlab/samson_plugin.rb
```diff
@@ -11,21 +11,14 @@ class Engine < Rails::Engine
   end
 end
 
-Samson::Hooks.callback :changeset_api_request do |changeset, method|
-  next unless changeset.project.gitlab?
+Samson::Hooks.callback :repo_commit_from_ref do |project, reference|
+  next unless project.gitlab?
```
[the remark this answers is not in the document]
grosser: done

PR #3532, opened 2019-10-07
Merge deploy_tab and deploy_confirmation into deploy_changeset

adammw:
Merge deploy_tab and deploy_confirmation into deploy_changeset.
Clean up the hooks as the confirmation ones are pretty useless since they can't use the reference. New hooks need to be aware they could be used for changeset preview as well as actually rendering it, so need to check `@deploy.persisted?` if it's a saved record or not.

Also fixes this bug, where preview didn't consider env vars for the individual deploy...
Screen Shot 2019-10-07 at 3 33 39 pm

### References
- Jira link: 

### Risks
- Low: UI bugs in non-primary tabs of deploy or deploy review


app/assets/javascripts/deploys.js
```diff
@@ -13,13 +16,22 @@ $(function () {
...
       e.preventDefault();
-      var tab = $(this);
-      tab.tab("show");
 
+      var $tab = $(this);
+      var $placeholder = $container.find('.changeset-placeholder');
+      $tab.tab("show");
+
+      // abort if already loaded (either static or via previous changeset load)
+      if ($container.find($tab.attr('href')).length) { return; }
```
grosser: might be able to solve this by doing `.one('click', function)` instead of `.click` above

app/assets/javascripts/deploys.js
```diff
@@ -13,13 +16,22 @@ $(function () {
...
+
+      // prevent double-loading
       if (!changesetLoaded) {
```
grosser: this also prevents double-loading again ?

grosser (approved):
😍

app/assets/javascripts/deploys.js
adammw: we don't know if the link maps to a static or something not loaded yet, hence we need this even after the first click

app/assets/javascripts/deploys.js
adammw: yep, but also prevents requesting a second time while the ajax request is in flight (ie. a double click on the tab)

uthark (approved):

grosser (approved):
👍

PR #3536, opened 2019-10-09
PAAS-4603 expire pull requests cache so updated risks are visible

grosser:
found this while poking at risks vs commit status ticket

@zendesk/compute

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -95,6 +95,7 @@ class Changeset
 
   # github only supports finding open PRs for branches (not commits or tags)
   # https://help.github.com/en/articles/searching-issues-and-pull-requests
+  # TODO: can we just use .new here to avoid extra fetch ?
   def open_pull_requests
     return [] if static? || MAIN_BRANCHES.include?(@reference)
     org = @repo.split("/", 2).first
--- a/app/models/changeset/commit.rb
+++ b/app/models/changeset/commit.rb
@@ -36,6 +36,7 @@ class Changeset::Commit
     @data.sha.slice(0, 7)
   end
 
+  # @return [Integer, NilClass]
   def pull_request_number
     if number = summary[PULL_REQUEST_MERGE_MESSAGE, 1] || summary[PULL_REQUEST_SQUASH_MESSAGE, 1]
       Integer(number)
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -20,43 +20,54 @@ class Changeset::PullRequest
   # Docs on the pull request event: https://developer.github.com/v3/activity/events/types/#pullrequestevent
   VALID_ACTIONS = ['opened', 'edited', 'synchronize'].freeze
 
-  # Finds the pull request with the given number.
-  #
-  # repo   - The String repository name, e.g. "zendesk/samson".
-  # number - The Integer pull request number.
-  #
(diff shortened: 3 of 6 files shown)


test/models/changeset/pull_request_test.rb
```diff
@@ -70,15 +79,17 @@ def no_risks
       {
         "number" => 1,
         "pull_request" => {
-          "state" => 'open',
-          "body" => 'pr description [samson review]'
+          "state" => "open",
+          "body" => "pr description [samson review]"
         },
-        "github" => {
-          "action" => 'opened'
```
grosser: this test was just wrong, action is directly in the payload 🤷‍♂

app/models/changeset.rb
```diff
@@ -95,6 +95,7 @@ def compare_cache_key
 
   # github only supports finding open PRs for branches (not commits or tags)
   # https://help.github.com/en/articles/searching-issues-and-pull-requests
+  # TODO: can we just use .new here to avoid extra fetch ?
```
adammw: relevant?

app/models/changeset/commit.rb
```diff
@@ -36,6 +36,7 @@ def short_sha
     @data.sha.slice(0, 7)
   end
 
+  # @return [Integer, NilClass]
```
adammw:
```suggestion
  # @return [Integer|NilClass]
```

app/models/changeset/pull_request.rb
```diff
@@ -20,43 +20,54 @@ class Changeset::PullRequest
...
-      previous_desc = payload.dig('changes', 'body', 'from')
-      return false if !previous_desc || (previous_desc =~ WEBHOOK_FILTER && data['body'] =~ WEBHOOK_FILTER)
+      data['body'].match? WEBHOOK_FILTER
     end
 
-    !!(data['body'] =~ WEBHOOK_FILTER)
+    private
+
+    def cache_key(repo, number)
+      [self, repo, number].join("-")
```
adammw: does repo include org name?

test/models/changeset/pull_request_test.rb
```diff
@@ -44,6 +44,15 @@ def no_risks
     end
   end
 
+  describe ".expire" do
+    it "expires find cache" do
+      GITHUB.expects(:pull_request).with("foo/bar", 42).returns({}).times(2)
+      2.times { Changeset::PullRequest.find("foo/bar", 42) }
+      Changeset::PullRequest.expire("foo/bar", 42)
+      Changeset::PullRequest.find("foo/bar", 42)
```
adammw: assert return value?

app/models/changeset/pull_request.rb
grosser: yep

app/models/changeset/commit.rb
grosser: only guidlines I found say to use comma `# @param argname [true, false] only true or false
` https://rubydoc.info/gems/yard/file/docs/GettingStarted.md

app/models/changeset.rb
grosser: that's for a later PR :)

test/models/changeset/pull_request_test.rb
grosser: done

PR #3537, opened 2019-10-10
PAAS-4303 show missing risks as commit status so users are more aware

grosser:

### Risks
- Low: slows down commit status too much or is annoying to users / risk caching issues show more

### DIFF
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -49,12 +49,20 @@ class CommitStatus
 
   private
 
+  # TODO: do non-db lookups in parallel to save time
   # @return [Hash]
   def combined_status
     @combined_status ||= begin
       if @commit
         statuses = [github_combined_status]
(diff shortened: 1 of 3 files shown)


app/models/commit_status.rb
```diff
@@ -49,12 +49,20 @@ def expire_cache
...
   # @return [Hash]
   def combined_status
     @combined_status ||= begin
       if @commit
         statuses = [github_combined_status]
-        statuses += [release_status, *only_production_status, *plugin_statuses].compact if @stage
+        if @stage
+          statuses.concat [
+            release_status,
+            only_production_status,
```
eatwithforks:
```suggestion
            *only_production_status,
```
missing * to expand blob?

grosser: no, it's only a single status so no need to expand

PR #3538, opened 2019-10-10
Added resolve endpoint to secrets controller

davidreuss:
This allows resolving secrets to get their paths in the context of a
project and deploy group.

This could be used for external integrations, while keeping the logic of
resolving secrets, and the currently underlying path structure in place,
without needing to explicitly know where a key resides in vault.

### References
- Jira link: https://zendesk.atlassian.net/browse/GUIDEOPS-1284

### Risks
- Low: This adds a new endpoint

### DIFF
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -54,6 +54,26 @@ class SecretsController < ApplicationController
       sort_by { |_, v| -v.size }
   end
 
+  def resolve
+    params.require([:project_id, :deploy_group, :keys])
+
+    project = Project.find(params[:project_id])
+    deploy_group = DeployGroup.find_by_permalink(params[:deploy_group])
+    keys = params[:keys]
+    resolver = Samson::Secrets::KeyResolver.new(project, [deploy_group])
+
+    resolved = {}
+    keys.each do |key|
+      resolved[key] = resolver.expand('unused', key).first&.last
+    end
+
+    respond_to do |format|
+      format.json do
+        render_as_json :secrets, resolved, nil
+      end
+    end
+  end
+
   def history
     @history = Samson::Secrets::Manager.history(id, resolve: true)
   end
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -55,21 +55,38 @@ class SecretsController < ApplicationController
   end
 
   def resolve
-    params.require([:project_id, :deploy_group, :keys])
+    if params[:project_id].present?
+      project = if params[:project_id].match?(/^\d+$/)
+        Project.find_by_id(params[:project_id])
+      else
+        Project.find_by_permalink(params[:project_id])
+      end
+    elsif params[:project_permalink].present?
+      project = Project.find_by_permalink(params[:project_permalink])
+    else
+      raise "Neither project_id or project_permalink given as parameters"
+    end
 
-    project = Project.find(params[:project_id])
-    deploy_group = DeployGroup.find_by_permalink(params[:deploy_group])
-    keys = params[:keys]
+    deploy_group = DeployGroup.find_by_permalink(params.require(:deploy_group))
+    keys_param = params.require(:keys)
+    keys = keys_param.is_a?(Array) ? keys_param : keys_param.split(/, ?/)
+
     resolver = Samson::Secrets::KeyResolver.new(project, [deploy_group])
 
     resolved = {}
     keys.each do |key|
-      resolved[key] = resolver.expand('unused', key).first&.last
+      expanded = resolver.expand(key, key)
+      found = expanded.first&.last
+      if found
+        expanded.each { |k, r| resolved[k] = r }
+      else
+        resolved[key] = nil
+      end
     end
 
     respond_to do |format|
       format.json do
-        render_as_json :secrets, resolved, nil
+        render_as_json :resolved, resolved, nil
       end
     end
   end
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -56,18 +56,15 @@ class SecretsController < ApplicationController
 
   def resolve
     if params[:project_id].present?
-      project = if params[:project_id].match?(/^\d+$/)
-        Project.find_by_id(params[:project_id])
-      else
-        Project.find_by_permalink(params[:project_id])
-      end
+      project = Project.find(params[:project_id])
     elsif params[:project_permalink].present?
-      project = Project.find_by_permalink(params[:project_permalink])
+      project = Project.find_by_permalink!(params[:project_permalink])
     else
-      raise "Neither project_id or project_permalink given as parameters"
+      render_json_error 400, "Neither project_id or project_permalink given as parameters"
+      return
     end
 
-    deploy_group = DeployGroup.find_by_permalink(params.require(:deploy_group))
+    deploy_group = DeployGroup.find_by_permalink!(params.require(:deploy_group))
     keys_param = params.require(:keys)
     keys = keys_param.is_a?(Array) ? keys_param : keys_param.split(/, ?/)
 
@@ -76,8 +73,7 @@ class SecretsController < ApplicationController
     resolved = {}
     keys.each do |key|
       expanded = resolver.expand(key, key)
-      found = expanded.first&.last
-      if found
+      if expanded.any?
         expanded.each { |k, r| resolved[k] = r }
       else
         resolved[key] = nil
@@ -86,7 +82,7 @@ class SecretsController < ApplicationController
 
     respond_to do |format|
       format.json do
-        render_as_json :resolved, resolved, nil
+        render json: {resolved: resolved}
       end
     end
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -98,6 +98,8 @@ Samson::Application.routes.draw do
   resources :secrets, except: [:edit] do
     collection do
       get :duplicates
+      get :resolve
+      post :resolve
     end
     member do
       get :history
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -28,6 +28,19 @@ Secrets can be resolved to their full path in the context of a projects deploy g
 
 In the below example we have a secret configured for `a` in the deploy group `group1`, but no secret configured for `b`.
 
+The endpoint supports both GET:
+
+```
+$ curl -s -H "Content-Type: application/json" -H "Authorization: Bearer $SAMSON_ACCESS_TOKEN" $SAMSON_BASE_URL/secrets/resolve.json?project_id=2&deploy_group=group1&keys[]=a&keys[]=b
+{
+  "a": "global/example-kubernetes/global/a",
+  "b": null
+}
+```
+
+For ease of use to clients parameters can be sent in JSON format in a POST request. This way clients doesn't need to deal with url encoding semantics.
+
+
 ```
 # params.json
 {
@@ -45,3 +58,5 @@ $ curl -s -d @params.json -H "Content-Type: application/json" -H "Authorization:
   "b": null
 }
 ```
+
+This endpoint was added and is useful for resolving secrets as defined in kubernetes manifests, in the format that the `samson_secret_puller` or similar solutions use. The secrets are usually a list of names, with actual paths in the backend being different because of scoping/sharing concerns.
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -22,13 +22,13 @@ For these the `visible` flag can be used to keep them visible for everyone.
 
 Secrets can be `deprecated` first and deleted when all use outside of samson (vault backend or accessing secrets via api client) has stopped.
 
-### Resolving secrets from external tools
+### Resolving secrets via API
 
 Secrets can be resolved to their full path in the context of a projects deploy group with the `/secrets/resolve.json` endpoint.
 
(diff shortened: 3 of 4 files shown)


davidreuss:
I'm sure there are some rails'isms i am not using here in terms of validating request input, and i'd be happy to accept suggestions.

grosser:
sounds like secrets#show to me ... maybe with a `resolve` parameter 🤷‍♂

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,26 @@ def duplicates
       sort_by { |_, v| -v.size }
   end
 
+  def resolve
+    params.require([:project_id, :deploy_group, :keys])
+
+    project = Project.find(params[:project_id])
```
grosser: kinda strange that deploy group is by permalink and project is by id ... most other controller do project by permalink

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,26 @@ def duplicates
       sort_by { |_, v| -v.size }
   end
 
+  def resolve
+    params.require([:project_id, :deploy_group, :keys])
```
grosser: doing the require on each line as they are used reduces duplication

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,26 @@ def duplicates
...
+
+    project = Project.find(params[:project_id])
+    deploy_group = DeployGroup.find_by_permalink(params[:deploy_group])
+    keys = params[:keys]
+    resolver = Samson::Secrets::KeyResolver.new(project, [deploy_group])
+
+    resolved = {}
+    keys.each do |key|
+      resolved[key] = resolver.expand('unused', key).first&.last
```
grosser: resolver returns multiple results when resolving a `_*` key, you might want to support that too and do an `each { |k, r| resolved[k] = r }`

app/controllers/secrets_controller.rb
davidreuss: yep, i know -- i couldn't figure out how to wire it up, i can change this to support both if you want -- i was looking at what we currently do in scooter, and the new mutating webhook service, and they're both looking up projects by id, because they look up the project first.

What would it look like to support both? I didn't get how that was wired in the current state.

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,26 @@ def duplicates
...
+    end
+
+    respond_to do |format|
+      format.json do
+        render_as_json :secrets, resolved, nil
```
grosser: this is kinda meh since these are not really `secrets`, just a random hash ... so prefer `render json: resolved`

config/routes.rb
```diff
@@ -98,6 +98,8 @@
   resources :secrets, except: [:edit] do
     collection do
       get :duplicates
+      get :resolve
+      post :resolve
```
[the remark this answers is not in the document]
davidreuss: the idea is to use this to resolve a bunch of secrets -- projects like classic/help-center has a _lot_ of secrets, and there's usual limits to how much you can shove into a URL -- seems easier to handle with a POST body.

app/controllers/secrets_controller.rb
davidreuss: i don't understand this -- sorry 🤕

grosser:
add some docs on why it's here like `# allow posting files for resolution`
I think the url limit is crazy like 64k, but supporting files is fine ...
unless -X GET works ...

docs/secrets.md
```diff
@@ -21,3 +21,27 @@ For these the `visible` flag can be used to keep them visible for everyone.
...
+{
+  "project_id": 2,
+  "deploy_group": "group1",
+  "keys": [
+    "a",
+    "b"
+  ]
+}
+
+$ curl -s -d @params.json -H "Content-Type: application/json" -H "Authorization: Bearer $SAMSON_ACCESS_TOKEN" $SAMSON_BASE_URL/secrets/resolve.json | jq '.'
```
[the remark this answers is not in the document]
davidreuss: yep, if you use keys like `keys[]=foo&keys[]=bar` ... in the query string

app/controllers/secrets_controller.rb
grosser: yeah

app/controllers/secrets_controller.rb
grosser: supporting both is fine too ... I think we even do that somewhere already ... `id.match?(/^\d+$/) ? find_by_id : find_by_permalink`

app/controllers/secrets_controller.rb
davidreuss: oh, i read it as it would be something like

```
{
  "wildcard_*": [
    {"wildcard_foo": "global/global/global/wildcard_foo" },
    {"wildcard_blah": "global/global/global/wildcard_blah" },
    ...
  ]
}
```

took your suggestion and merge them back into the hash, so it's kept flat

it looks like there's some implicit uppercasing going on though, but -- i left that as is. 🤷‍♂

davidreuss:
I think this is ready for a re-review.

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,43 @@ def duplicates
       sort_by { |_, v| -v.size }
   end
 
+  def resolve
+    if params[:project_id].present?
+      project = if params[:project_id].match?(/^\d+$/)
```
grosser: don't need this if you are handling project_permalink separately

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,43 @@ def duplicates
...
+        Project.find_by_id(params[:project_id])
```
grosser:
```suggestion
        Project.find(params[:project_id])
```

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,43 @@ def duplicates
...
+      else
+        Project.find_by_permalink(params[:project_id])
+      end
+    elsif params[:project_permalink].present?
+      project = Project.find_by_permalink(params[:project_permalink])
```
grosser:
```suggestion
      project = Project.find_by_permalink!(params[:project_permalink])
```

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,43 @@ def duplicates
...
+    else
+      raise "Neither project_id or project_permalink given as parameters"
```
grosser: render a 400 and return

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,43 @@ def duplicates
...
+    end
+
+    deploy_group = DeployGroup.find_by_permalink(params.require(:deploy_group))
```
grosser:
```suggestion
    deploy_group = DeployGroup.find_by_permalink!(params.require(:deploy_group))
```

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,43 @@ def duplicates
...
+    keys_param = params.require(:keys)
+    keys = keys_param.is_a?(Array) ? keys_param : keys_param.split(/, ?/)
```
grosser: support for 1 flavor should be enough ... arrays seem fine 🤷‍♂

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,43 @@ def duplicates
...
+
+    resolver = Samson::Secrets::KeyResolver.new(project, [deploy_group])
+
+    resolved = {}
+    keys.each do |key|
+      expanded = resolver.expand(key, key)
+      found = expanded.first&.last
+      if found
```
grosser:
```suggestion
      if found.any?
```

app/controllers/secrets_controller.rb
```diff
@@ -54,6 +54,43 @@ def duplicates
...
+        expanded.each { |k, r| resolved[k] = r }
+      else
+        resolved[key] = nil
+      end
+    end
+
+    respond_to do |format|
+      format.json do
+        render_as_json :resolved, resolved, nil
```
grosser:
```suggestion
        render json: resolved
```

docs/secrets.md
```diff
@@ -21,3 +21,42 @@ For these the `visible` flag can be used to keep them visible for everyone.
...
 
 Secrets can be `deprecated` first and deleted when all use outside of samson (vault backend or accessing secrets via api client) has stopped.
+
+### Resolving secrets from external tools
+
+Secrets can be resolved to their full path in the context of a projects deploy group with the `/secrets/resolve.json` endpoint.
+
+In the below example we have a secret configured for `a` in the deploy group `group1`, but no secret configured for `b`.
+
+The endpoint supports both GET:
```
grosser:
```suggestion
The endpoint supports both GET and POST:
```

docs/secrets.md
```diff
@@ -21,3 +21,42 @@ For these the `visible` flag can be used to keep them visible for everyone.
...
+
+```
+$ curl -s -H "Content-Type: application/json" -H "Authorization: Bearer $SAMSON_ACCESS_TOKEN" $SAMSON_BASE_URL/secrets/resolve.json?project_id=2&deploy_group=group1&keys[]=a&keys[]=b
+{
+  "a": "global/example-kubernetes/global/a",
+  "b": null
+}
+```
+
+For ease of use to clients parameters can be sent in JSON format in a POST request. This way clients doesn't need to deal with url encoding semantics.
```
grosser: can just remove that line, the example is pretty obvious :)

docs/secrets.md
```diff
@@ -21,3 +21,42 @@ For these the `visible` flag can be used to keep them visible for everyone.
...
+}
+
+$ curl -s -d @params.json -H "Content-Type: application/json" -H "Authorization: Bearer $SAMSON_ACCESS_TOKEN" $SAMSON_BASE_URL/secrets/resolve.json | jq '.'
+{
+  "a": "global/example-kubernetes/global/a",
+  "b": null
+}
+```
+
+This endpoint was added and is useful for resolving secrets as defined in kubernetes manifests, in the format that the `samson_secret_puller` or similar solutions use. The secrets are usually a list of names, with actual paths in the backend being different because of scoping/sharing concerns.
```
grosser: can remove that line too ... whoever reads finds the heading should know this already 🤷‍♂

docs/secrets.md
```diff
@@ -21,3 +21,42 @@ For these the `visible` flag can be used to keep them visible for everyone.
 ### Deprecated
 
 Secrets can be `deprecated` first and deleted when all use outside of samson (vault backend or accessing secrets via api client) has stopped.
+
+### Resolving secrets from external tools
```
grosser:
```suggestion
### Resolving secret keys to full paths via API
```

docs/secrets.md
```diff
@@ -21,3 +21,42 @@ For these the `visible` flag can be used to keep them visible for everyone.
...
+
+Secrets can be resolved to their full path in the context of a projects deploy group with the `/secrets/resolve.json` endpoint.
+
+In the below example we have a secret configured for `a` in the deploy group `group1`, but no secret configured for `b`.
```
grosser:
```suggestion
In the below example we have a secret configured for `keya` in the deploy group `group1`, but no secret configured for `keyb`.
```

test/controllers/secrets_controller_test.rb
```diff
@@ -507,5 +516,43 @@ def do_update(extras = {})
...
+        create_secret 'production/z/pod2/bar'
+        get :resolve, params: {project_id: other_project.id, deploy_group: 'pod2', keys: ['bar', 'foo'], format: 'json'}
+        assert_response :success
+
+        result = JSON.parse(response.body)
+        resolved = result['resolved']
+        resolved['bar'].must_equal 'production/z/pod2/bar'
+        resolved['foo'].must_be_nil
+      end
+      it "supports permalink given as project_id" do
```
grosser: newlines between `it`s

davidreuss:
GitHub seems slow, not sure if they're having problems. https://travis-ci.org/zendesk/samson/builds/600611177 succeeded, but status doesn't (currently) reflect that.

This is ready for a re-review.

test/controllers/secrets_controller_test.rb
```diff
@@ -507,5 +516,39 @@ def do_update(extras = {})
...
       end
     end
+
+    describe "#resolve" do
+      it "renders resolved secrets" do
+        create_secret 'production/z/pod2/bar'
+        get :resolve, params: {project_id: other_project.id, deploy_group: 'pod2', keys: ['bar', 'foo'], format: 'json'}
+        assert_response :success
+
+        result = JSON.parse(response.body)
```
grosser: FYI usually I go with `must_equal` for these cases so it's easier to see what the expected output is and extra keys get shown too

davidreuss: i'm not exactly following -- can you point to an example?

app/controllers/secrets_controller.rb
davidreuss: this was so you could easily resolve a couple of keys in the browser by doing `keys=key1,key2,key3` .. but i can remove it if we just want to support `keys[]=key1&keys[]=key2&.....`

grosser:
users could just do 1 at a time when playing in the browser ... but does
not matter much

grosser:
JSON.parse(response.body).must_equal "foo" => "bar"

PR #3539, opened 2019-10-12
properly indent changeset subclasses so they do not cause load errors…

grosser:
… when doing things in parallel

@zendesk/compute


grosser:
load errors happen atm when I add parallelization ... never mind it still happens when rebasing on this :(

PR #3540, opened 2019-10-13
Fix regression causing prerequisite stages to stop working 

KJTsanaktsidis:
SamsonPrerequisisteStages::validate_deployed_to_all_prerequisiste_stages
was changed to accept a commit instead of a reference, but it was still
being called with a reference from the before_deploy hook. This made
prerequisiste stages never be satisfied (unless the deploy was done with
the acutal commit SHA).

Fix it by resolving the ref to a commit in the before_deploy hook.

### Risks
- Low: This only changes the `before_deploy` codepath for prerequisite stages, and this is already broken for those projects that have them. Any further breakage is easily worked around by disabling the prerequisite.

### DIFF
--- a/plugins/prerequisite_stages/lib/samson_prerequisite_stages/samson_plugin.rb
+++ b/plugins/prerequisite_stages/lib/samson_prerequisite_stages/samson_plugin.rb
@@ -14,8 +14,12 @@ module SamsonPrerequisiteStages
   Samson::Hooks.view :stage_show, 'samson_prerequisite_stages'
 
   Samson::Hooks.callback :before_deploy do |deploy, _|
+    # This check is technically redundant (undeployed_prerequisite_stages above will be empty anyway if this
+    # is not true) but a whole bunch of unrelated tests will complain about a missing HTTP stub from resolving
+    # the ref to a commit if we don't do this.
+    next unless deploy.stage.prerequisite_stage_ids?
     error = SamsonPrerequisiteStages.validate_deployed_to_all_prerequisite_stages(
-      deploy.stage, deploy.reference, deploy.commit
+      deploy.stage, deploy.reference, deploy.project.repo_commit_from_ref(deploy.reference)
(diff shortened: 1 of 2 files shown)


grosser:
yeah I think you found the bug! :D
... this used to get resolved in the method but now no longer does

 - it should resolve the reference and not the commit so it does not look all funky
 - we need a test that fails without that change so it does not come back

I can help with these unless you want to have a go at this further

grosser:
I guess the real refactor we need is that deploy.commit should no longer be a reference ... but that's a bigger undertaking

KJTsanaktsidis:
Yeah I saw the TODO for that in the `Deploy#commit` message... with code that dates from five years ago :)

I'll have a bash at turning this into a real PR now & getting it sent out today. It's Sunday for you!

Thanks for your help.

plugins/prerequisite_stages/lib/samson_prerequisite_stages/samson_plugin.rb
```diff
@@ -14,8 +14,12 @@ def self.validate_deployed_to_all_prerequisite_stages(stage, reference, commit)
   Samson::Hooks.view :stage_show, 'samson_prerequisite_stages'
 
   Samson::Hooks.callback :before_deploy do |deploy, _|
+    # This check is technically redundant (undeployed_prerequisite_stages above will be empty anyway if this
+    # is not true) but a whole bunch of unrelated tests will complain about a missing HTTP stub from resolving
+    # the ref to a commit if we don't do this.
+    next unless deploy.stage.prerequisite_stage_ids?
```
grosser: I think it's smart to do this anyway, since it saves a lookup :)

KJTsanaktsidis:
@grosser Thanks for the review. I don't seem to have permission to merge this - does it need 2x reviews or does someone from @zendesk/samson have to do the merge?

PR #3541, opened 2019-10-14
Fetch commit status parts in parallel to speed it up

grosser:
this takes it in dev from `2810ms` to `1445ms`

@zendesk/compute 

### Risks
- Low: parallel bugs making status not work

### DIFF
--- a/app/models/changeset.rb
+++ b/app/models/changeset.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 class Changeset
   ATTRIBUTE_TABS = ["files", "commits", "pull_requests", "risks", "jira_issues"].freeze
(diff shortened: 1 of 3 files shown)


app/models/changeset.rb
```diff
@@ -31,7 +32,11 @@ def files
   end
 
   def pull_requests
-    @pull_requests ||= merged_pull_requests + open_pull_requests
+    @pull_requests ||= begin
+      numbers = (merged_pull_requests + open_pull_requests)
+      PullRequest.name # avoid require race conditions in development
```
adammw:
```suggestion
      PullRequest.name # call no-op method to load class before running in parallel
```

PR #3542, opened 2019-10-16
Add secret whitespace warning

adammw:
Screen Shot 2019-10-16 at 2 53 10 pm

### Risks
- Low: modifies secret entry and env var entry fields

### DIFF
--- a/app/views/secrets/show.html.erb
+++ b/app/views/secrets/show.html.erb
@@ -53,9 +53,9 @@
       <% end %>
 
       <% needs_value = @duplicate_secret_error || !id %>
-      <%= form.input :value do %>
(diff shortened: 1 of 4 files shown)


app/views/secrets/show.html.erb
```diff
@@ -119,6 +119,17 @@ $(function () {
         $warn.text('Value is not valid json ( ' + e.toString() + ' )');
         $warn.show();
       }
+    } else if(/^\s+/.test(val) || /\s+$/.test(val)) {
```
grosser: how about putting that under `app/assets/javascripts/whitespace_stripper.js` and then reuse it for env vars ... we have the same problem there :)

PR #3545, opened 2019-10-18
Update project show page when stages change

grosser:
nameand deploy group changes were stale before

this adds an extra query for the stages, but I'd prefer that over explicit expire

@zendesk/compute

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -21,7 +21,7 @@
     </thead>
     <tbody>
       <% if @project.stages.to_a.any? %>
-        <% cache [@project, 'stages', Lock.cache_key] do %>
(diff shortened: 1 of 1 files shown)


app/views/projects/show.html.erb
```diff
@@ -21,7 +21,7 @@
     </thead>
     <tbody>
       <% if @project.stages.to_a.any? %>
-        <% cache [@project, 'stages', Lock.cache_key] do %>
+        <% cache [@project, 'stages', Lock.cache_key, @project.stages.map(&:updated_at).max] do %>
```
adammw: does .map create a n+1? pluck or a query involving sql max perhaps?

grosser: it fetches the stages in a single query ... so not really expensive ... 
we already do the `.to_a` above so the stages are preloaded

PR #3546, opened 2019-10-18
Fix build search

adammw:
https://rollbar-us.zendesk.com/Zendesk/Samson/items/2279/occurrences/1739017169/

### Risks
- Low: search with commit arg breaks

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -10,14 +10,15 @@ class BuildsController < ApplicationController
 
   def index
     builds = scope
+    search = params[:search] || ActionController::Parameters.new
 
-    if status = params[:search]&.delete(:status)&.presence
+    if status = search.delete(:status).presence
       builds = builds.
         left_outer_joins(:docker_build_job).
         where("jobs.status = ? OR external_status = ?", status, status)
     end
 
-    if commit = params[:search]&.delete(:git_commit)&.presence
+    if commit = search.delete(:git_commit).presence
       builds =
         if commit.match?(Build::SHA1_REGEX)
           builds.where(git_sha: commit)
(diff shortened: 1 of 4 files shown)


test/controllers/builds_controller_test.rb
```diff
@@ -42,12 +42,12 @@ def stub_git_reference_check(returns)
...
-        get :index, params: {search: {commit: build.git_sha}}
+        get :index, params: {search: {git_commit: build.git_sha}}
         assigns(:builds).must_equal [build]
       end
 
       it 'can search for ref' do
-        get :index, params: {search: {commit: build.git_ref}}
+        get :index, params: {search: {git_commit: build.git_ref}}
         assigns(:builds).must_equal [build]
       end
```
grosser: `it "does not pass random keys to sql"` since you already implemented that :)

app/controllers/builds_controller.rb
```diff
@@ -9,15 +9,15 @@ class BuildsController < ApplicationController
...
-    if status = search.delete(:status)
+
+    if status = params[:search]&.delete(:status)&.presence
       builds = builds.
         left_outer_joins(:docker_build_job).
         where("jobs.status = ? OR external_status = ?", status, status)
     end
 
-    if commit = search.delete(:commit)
+    if commit = params[:search]&.delete(:git_commit)&.presence
```
grosser: that was commit on purpose since it's not a direct column mapping

grosser: nvm ...

app/controllers/builds_controller.rb
```diff
@@ -9,15 +9,15 @@ class BuildsController < ApplicationController
   before_action :find_build, only: [:show, :build_docker_image, :edit, :update]
 
   def index
-    search = params[:search]&.to_unsafe_h&.except(:time_format)&.select { |_, v| v.present? } || {}
     builds = scope
-    if status = search.delete(:status)
+
+    if status = params[:search]&.delete(:status)&.presence
```
grosser: maybe simpler to do `search = params[:search] || ActionPack::Parameters.new` and then not deal with all the nils

PR #3549, opened 2019-10-22
add new feature to inject Istio annotation

jonmoter:
Docs: https://istio.io/docs/setup/additional-setup/sidecar-injection/

I want to make it easy for teams to enable Istio sidecars for certain roles, in particular deploy targets.  For example, having a sidecar for an app-server role, but not a console role, and only in non-production clusters.

This adds a column to track that bit of information, and some logic to template filler.
  
The feature is gated around an `ISTIO_INJECTION_SUPPORTED` env var.

Note: not tested yet, since I need to get Samson into a state on my laptop where I can actually run it...

### Risks
- Low: new feature gated by ENV var

cc @zendesk/compute, @zendesk/service-mesh

### DIFF
--- a/plugins/kubernetes/app/helpers/deploy_group_roles_helper.rb
+++ b/plugins/kubernetes/app/helpers/deploy_group_roles_helper.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 # NOTE: using kubernetes namespace resulted in a strange error on every page
 module DeployGroupRolesHelper
+  def show_istio_sidecar_ui?
+    Samson::EnvCheck.set?('ISTIO_INJECTION_SUPPORTED')
+  end
+
   def kubernetes_deploy_group_role_replica(role, dgr)
     if dgr.delete_resource?
       icon_tag :remove, title: "Marked for deletion"
--- a/plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
@@ -93,10 +93,6 @@ module Kubernetes
       end
     end
 
-    def self.istio_injection_supported?
-      Samson::EnvCheck.set?("ISTIO_INJECTION_SUPPORTED")
-    end
-
     def requests_below_limits
       if limits_cpu && requests_cpu && requests_cpu > limits_cpu
         errors.add :requests_cpu, "must be less than or equal to the limit"
--- a/plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
@@ -93,6 +93,10 @@ module Kubernetes
       end
     end
 
+    def self.istio_injection_supported?
+      Samson::EnvCheck.set?("ISTIO_INJECTION_SUPPORTED")
+    end
+
     def requests_below_limits
       if limits_cpu && requests_cpu && requests_cpu > limits_cpu
         errors.add :requests_cpu, "must be less than or equal to the limit"
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -24,6 +24,10 @@ module Kubernetes
       resources.each(&:deploy)
     end
 
+    def deploy_group_role
+      @deploy_group_role ||= DeployGroupRole.where(kubernetes_role: kubernetes_role, deploy_group: deploy_group).first
+    end
+
     def revert
       raise "Can only be done after a deploy" unless @previous_resources
       resources.each_with_index do |resource, i|
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -25,7 +25,7 @@ module Kubernetes
     end
 
     def deploy_group_role
-      @deploy_group_role ||= DeployGroupRole.where(kubernetes_role: kubernetes_role, deploy_group: deploy_group).first
+      @deploy_group_role ||= DeployGroupRole.where(kubernetes_role: kubernetes_role, deploy_group: deploy_group).first!
     end
 
     def revert
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -436,12 +436,12 @@ module Kubernetes
 
       # https://istio.io/docs/setup/additional-setup/sidecar-injection/#policy
       annotation_name = 'sidecar.istio.io/inject'.to_sym
-      template.dig_set([:spec, :template, :metadata, :annotations, annotation_name], "true")
+      pod_template.dig_set [:metadata, :annotations, annotation_name], init_containers
 
-      # Also add labels to the Deployment or DaemonSet, and to the Pod template
+      # Also add labels to the resource and to the Pod template.
       # This is not necessary for Istio, but makes it easier for us to select and see
       # which resources should have sidecars injected.
-      template.dig_set([:spec, :template, :metadata, :labels, annotation_name], "true")
+      pod_template.dig_set([:metadata, :labels, annotation_name], "true")
       template.dig_set([:metadata, :labels, annotation_name], "true")
     end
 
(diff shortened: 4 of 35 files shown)


plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
```diff
@@ -93,6 +93,10 @@ def self.seed!(stage)
       end
     end
 
+    def self.istio_injection_supported?
+      Samson::EnvCheck.set?("ISTIO_INJECTION_SUPPORTED")
```
uthark: `ISTIO_ENABLE_INJECTION` ?

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -1069,6 +1069,39 @@ def secret_annotations(hash)
...
+
+      before { doc.deploy_group_role.inject_istio_annotation = true }
+
+      let(:istio_annotation) { :'sidecar.istio.io/inject' }
+
+      let(:pod_annotation) { template.to_hash.dig(:spec, :template, :metadata, :annotations, istio_annotation) }
+      let(:pod_label) { template.to_hash.dig(:spec, :template, :metadata, :labels, istio_annotation) }
+      let(:resource_label) { template.to_hash.dig(:metadata, :labels, istio_annotation) }
+
+      it "modifies the Deployment" do
```
uthark: What about StatefulSet, DaemonSet, Job, CronJob?

plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
grosser: noun-action is usually nicer especially when we plan to nest more things .. `ISTIO_INJECTION_ENABLED` ... but "supported" also makes sense since users still have to opt-in

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -24,6 +24,10 @@ def deploy
       resources.each(&:deploy)
     end
 
+    def deploy_group_role
+      @deploy_group_role ||= DeployGroupRole.where(kubernetes_role: kubernetes_role, deploy_group: deploy_group).first
```
grosser: should always exist, so add `!` ?
```suggestion
      @deploy_group_role ||= DeployGroupRole.where(kubernetes_role: kubernetes_role, deploy_group: deploy_group).first!
```

plugins/kubernetes/app/models/kubernetes/deploy_group_role.rb
jonmoter: I was following the convention of the `KRITIS_BREAKGLASS_SUPPORTED` env var.

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
jonmoter: The logic should work with any of them, since it uses the logic in TemplateFiller that applies to any k8s resource that have a pod template in it:

https://github.com/zendesk/samson/blob/ec06cc05bd97453fddb8acf7c8a02ca8a2345acc/plugins/kubernetes/app/models/kubernetes/template_filler.rb#L38-L57

It's a fair point that it would be a good idea to add tests to a few other resources though.

@grosser, are there any other tests for similar logic that iterate over all the types of resources that match `RoleConfigFile.primary?`

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -439,6 +440,21 @@ def set_kritis_breakglass
       template.dig_fetch(:metadata, :annotations)[:"kritis.grafeas.io/breakglass"] = "true"
     end
 
+    def set_istio_sidecar_injection
+      return unless Samson::EnvCheck.set?('ISTIO_INJECTION_SUPPORTED')
+      return unless @doc.deploy_group_role.inject_istio_annotation?
```
grosser: `next unless template.dig(:spec, :template, :metadata)` so we don't set it for configmaps etc ... or only set it for `pod_template` which will then also support Pod and CronJob

grosser (approved):
looks clean / safe 👍

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
grosser: use `pod_template` ... I'd just test with a single type or 2 (unless you add custom logic) ... the supported types are tested in `.primary?` tests 🤷‍♂

plugins/kubernetes/app/models/kubernetes/template_filler.rb
jonmoter: I actually had that line in there, and a test for it.  But I couldn't get that line to be triggered in my tests, so it was complaining about being an uncovered line.

Then I realized that the place I was calling this method up top is in a block that is only called for resources that have some variation of a pod template.  So the `set_istio_sidecar_injection` method wouldn't ever get called for a Service or ConfigMap.

plugins/kubernetes/app/models/kubernetes/release_doc.rb
jonmoter: It's technically possible that if you load up an old ReleaseDoc, it could be for a Role or DeployGroup that has since been (soft) deleted.  But I don't know what the expected behavior should be in that case.

grosser: we would never recreate templates for old docs ... so make it blow up :)

plugins/kubernetes/app/helpers/deploy_group_roles_helper.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 # NOTE: using kubernetes namespace resulted in a strange error on every page
 module DeployGroupRolesHelper
+  def show_istio_sidecar_ui?
+    Samson::EnvCheck.set?('ISTIO_INJECTION_SUPPORTED')
```
uthark: I suggest to extract constant ISTIO_INJECTION_SUPPORTED

uthark (approved):

plugins/kubernetes/app/models/kubernetes/release_doc.rb
grosser: `!` ?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
grosser: should still set it on the `pod_template` to reuse that logic / support Pod+CronJob

plugins/kubernetes/app/views/kubernetes/deploy_group_roles/index.html.erb
```diff
@@ -41,6 +44,9 @@
           <td><%= deploy_group_role.requests_cpu %> to <%= deploy_group_role.limits_cpu %></td>
           <td><%= deploy_group_role.requests_memory %> to <%= deploy_group_role.limits_memory %></td>
           <td><%= kubernetes_deploy_group_role_replica deploy_group_role.kubernetes_role, deploy_group_role %></td>
+          <% if show_istio_sidecar_ui? %>
+            <td><%= deploy_group_role.inject_istio_annotation? ? 'Yes' : '-' %></td>
```
grosser: `'Yes' : 'No'` like `app/views/environments/index.html.erb` ... or change both 🤷‍♂

plugins/kubernetes/app/views/kubernetes/deploy_group_roles/show.html.erb
```diff
@@ -29,6 +29,15 @@
     </div>
   </div>
 
+  <% if show_istio_sidecar_ui? %>
+  <div class="form-group">
```
grosser: indent ?

plugins/kubernetes/app/views/kubernetes/deploy_group_roles/show.html.erb
```diff
@@ -29,6 +29,15 @@
     </div>
```
grosser: I think the show view is deleted now ... so needs a rebase

plugins/kubernetes/app/helpers/deploy_group_roles_helper.rb
jonmoter: constants make it harder to test both enabled and disabled cases

plugins/kubernetes/app/views/kubernetes/deploy_group_roles/index.html.erb
jonmoter: I originally had "No" here, but I liked the way it looked better with the dash. It made the text jump out more for the roles that had it enabled when visually scanning the table.

grosser:
Sounds good, let's just make it consistent :)

PR #3552, opened 2019-10-22
Only recommend valid container syntax

grosser:
@zendesk/compute 

this will come back to bite us at some point, so might as well encourage users to migrate

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -108,14 +108,14 @@ Kubernetes::Release
 
 ### Docker Images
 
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/README.md
```diff
@@ -108,14 +108,14 @@ Kubernetes::Release
 
 ### Docker Images
 
-(To opt out of this feature set `containers[].samson/dockerfile: none` or `metadata.annotations.container-nameofcontainer-samson/dockerfile: none`)
+(To opt out of this feature set `metadata.annotations.container-nameofcontainer-samson/dockerfile: none`)
```
adammw: this annotation makes me sad because it's so long and wordy, would have preferred to have seen JSON or YAML inside the annotation instead

adammw (approved):

PR #3556, opened 2019-10-23
Filters for environment variables groups

sathishavm:
We need env vars for a project, without the shared environment variables (aka groups), as json.
We want to store both the project vars and the groups so we can reuse the groups.

1. Filter environment variables groups with `project_id` and `deploy_group` parameters.
`eg: /environment_variable_groups.json?deploy_group=pod999&project_id=1`
2. Preview project environment variables without groups.
 `eg: /projects/hello_world/environment_variables_preview.json`

### Risks
- Low

### DIFF
--- a/plugins/env/app/controllers/env/environment_controller.rb
+++ b/plugins/env/app/controllers/env/environment_controller.rb
@@ -14,6 +14,19 @@ module Env
       end
     end
 
+    def preview
+      deploy_groups =
+        if deploy_group = params[:deploy_group].presence
+          [DeployGroup.find_by_permalink!(deploy_group)]
+        else
+          DeployGroup.all
+        end
+      deploy = Deploy.new(project: current_project)
+      envs = SamsonEnv.env_groups(deploy, deploy_groups, preview: true, env_group: false)
+
+      render json: {environment_variables: envs || []}
+    end
+
     private
 
     def unexpanded_secrets(env)
--- a/plugins/env/app/controllers/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/environment_variable_groups_controller.rb
@@ -5,6 +5,18 @@ class EnvironmentVariableGroupsController < ApplicationController
 
   def index
     @groups = EnvironmentVariableGroup.all.includes(:environment_variables)
+
+    if project_id = params[:project_id].presence
+      @groups = @groups.joins(:project_environment_variable_groups).
+      where("project_environment_variable_groups.project_id = ?", project_id)
+    end
+
+    if deploy_group = params[:deploy_group].presence
+      group = DeployGroup.find_by_permalink!(deploy_group)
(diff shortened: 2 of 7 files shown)


grosser:
idk what's going on here ... can you split this up into smaller PRs ?

grosser:
also a clearer "I'm trying to get abc because xyz" would be nice so we later know why this was added

sathishavm:
@grosser This is just adding filters, updated description.

grosser:
but why do we need this ?
... viewing groups by project makes little sense since all we usually care about is "what env vars does this project need for this deploy group"

plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -5,6 +5,18 @@ class EnvironmentVariableGroupsController < ApplicationController
 
   def index
     @groups = EnvironmentVariableGroup.all.includes(:environment_variables)
+
+    if project_id = params[:project_id].presence
+      @groups = @groups.joins(:project_environment_variable_groups).
+      where("project_environment_variable_groups.project_id = ?", project_id)
+    end
+
+    if deploy_group = params[:deploy_group].presence
```
grosser: kinda misleading since it should match `all` + the "deploy groups environment" too ?

grosser:
.. and why is it not using the existing preview endpoint ?

grosser:
basically Env::EnvironmentController#show ... and query it for every deploy group

sathishavm:
The shared or global configuration will be provided and managed by external service and we need only environment variables for the project i.e without groups.
We already have JSON API for fetching environment variables with `project, stage and all scopes` but fetching all separately and prioritizing externally is not feasible solution.

plugins/env/app/controllers/env/environment_controller.rb
```diff
@@ -14,6 +14,19 @@ def show
...
 
+    def preview
+      deploy_groups =
+        if deploy_group = params[:deploy_group].presence
+          [DeployGroup.find_by_permalink!(deploy_group)]
+        else
+          DeployGroup.all
+        end
+      deploy = Deploy.new(project: current_project)
+      envs = SamsonEnv.env_groups(deploy, deploy_groups, preview: true, env_group: false)
```
grosser: `resolve_secrets: false`

plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -5,6 +5,18 @@ class EnvironmentVariableGroupsController < ApplicationController
...
+      group = DeployGroup.find_by_permalink!(deploy_group)
+      @groups = @groups.references(:environment_variables).
+      where("environment_variables.scope_type = 'DeployGroup' AND environment_variables.scope_id = ?", group.id)
```
grosser: would need to say "|| env matches || all"

PR #3557, opened 2019-10-24
PAAS-4981 cache incoming PRs to avoid having to fetch them

grosser:
@zendesk/compute 

needed to convert the payload to symbols because caching a Sawyer::Response
makes it otherwise forget its keys

```
Rails.cache.write 'foo', Sawyer::Resource.new(Octokit.agent, "a" => "b")
Rails.cache.read('foo').a
=> nil
```

### Risks
 - Low: pr review crashes or being wrong because webhook payload was incomplete/bad ... but double checked that already, so we should be good

### DIFF
--- a/app/controllers/integrations/github_controller.rb
+++ b/app/controllers/integrations/github_controller.rb
@@ -13,7 +13,7 @@ class Integrations::GithubController < Integrations::BaseController
 
   def create
     expire_commit_status if github_event_type == "status"
-    expire_pull_request if github_event_type == "pull_request"
+    cache_pull_request if github_event_type == "pull_request"
     super
   end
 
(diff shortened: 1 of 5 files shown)


app/controllers/integrations/github_controller.rb
```diff
@@ -24,7 +24,7 @@ def expire_commit_status
   end
 
   def expire_pull_request
```
hgokavarapuz: can we change the def name as it is confusing

uthark:
```suggestion
  def cache_pull_request
```

alanhogan:
this change may have caused a regression as Lotus PRs are no longer getting risks pulled by Samson

grosser:
reverted for now

PR #3559, opened 2019-10-24
Project Specific environment variables in preview

sathishavm:
Added project-specific environment variables in preview, so that we will have option to differentiate the project and shared(groups) envs.
eg:
` /environment_variable_groups/preview?project_id=1&project_specific=true`

### DIFF
--- a/plugins/env/app/controllers/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/environment_variable_groups_controller.rb
@@ -55,7 +55,7 @@ class EnvironmentVariableGroupsController < ApplicationController
     end
 
     options = {project_specific: params[:project_specific]}
-    options.merge!(params[:preview] ? {preview: true, resolve_secrets: true} : {resolve_secrets: false})
+    options.merge!(params[:preview].to_s == "false" ? {resolve_secrets: false} : {preview: true, resolve_secrets: true})
 
     @groups = SamsonEnv.env_groups(Deploy.new(project: @project), deploy_groups, **options)
 
--- a/plugins/env/app/controllers/environment_variable_groups_controller.rb
+++ b/plugins/env/app/controllers/environment_variable_groups_controller.rb
@@ -54,7 +54,10 @@ class EnvironmentVariableGroupsController < ApplicationController
       @project = Project.find(params[:project_id])
     end
 
-    @groups = SamsonEnv.env_groups(Deploy.new(project: @project), deploy_groups, preview: true)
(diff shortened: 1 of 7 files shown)


sathishavm:
`loofah` gem updated because of vulnerability https://github.com/flavorjones/loofah/issues/171

plugins/env/app/controllers/environment_variable_groups_controller.rb
```diff
@@ -54,7 +54,10 @@ def preview
       @project = Project.find(params[:project_id])
     end
 
-    @groups = SamsonEnv.env_groups(Deploy.new(project: @project), deploy_groups, preview: true)
+    options = {project_specific: params[:project_specific]}
+    options.merge!(params[:preview] ? {preview: true, resolve_secrets: true} : {resolve_secrets: false})
```
grosser: before it used `preview` mode when preview was not given ... so to not break that, we have to add a new case for `preview == false` (boolean or string supported) ?
```suggestion
    options.merge!(params[:preview].to_s == "false" ? {resolve_secrets: false} : {preview: true, resolve_secrets: true})
```

plugins/env/app/models/environment_variable.rb
```diff
@@ -73,11 +73,24 @@ def config_service_location(project, deploy_group, display:)
 
     private
 
-    def env_vars_from_db(deploy, deploy_group)
+    def env_vars_from_db(deploy, deploy_group, project_specific: nil)
+      # Project Specific:
+      # nil   => project env + groups env
+      # true  => project env
```
grosser:
```suggestion
      # true/"true"  => project env
```

plugins/env/app/models/environment_variable.rb
```diff
@@ -73,11 +73,24 @@ def config_service_location(project, deploy_group, display:)
...
+      # false => groups env
+      project_envs =
+        if project_specific.to_s == "true"
+          deploy.project.environment_variables
+        elsif project_specific.to_s == "false"
+          deploy.project.environment_variable_groups.flat_map(&:environment_variables)
```
grosser: this is duplicating the logic from `nested_environment_variables` ... so prefer moving project_specific there

```
    def env_vars_from_db(deploy, deploy_group, **args)

deploy.project.nested_environment_variables **args

def nested_environment_variables(project_specific: nil)
```

grosser:
oh yeah, looking good/simple :D

sathishavm:
@grosser ready to re-review!!

PR #3560, opened 2019-10-25
PAAS-4111 for update of services by re-creating them if they complain…

grosser:
… that they cannot be updated

Tested with https://github.com/samson-test-org/example-kubernetes
968b2d1 -> 87ba8f5 which modifies `spec.healthCheckNodePort`

Before resulted in
```
[16:47:01] JobExecution failed: Kubernetes error example sdfsfsfs GroupK: Service "example" is invalid: spec.healthCheckNodePort: Invalid value: 30124: cannot change healthCheckNodePort on loadBalancer service with externalTraffic=Local during update
```

this might later be expanded to other resources since we had these issues before

... an alternative would be to require a `samson/force-update` annotation if we think that re-creating a service could have bad consequences like changing it's id

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -294,3 +294,11 @@ of a Deployment, DaemonSet, or StatefulSet. Assuming you have Istio configured t
 MutatingWebhook in the target namespace, that should trigger Istio to inject the sidecar.
 
 To enable this functionality, set the environment variable `ISTIO_INJECTION_SUPPORTED=true`.
+
+### Forcing updates when kubernetes cannot change a resource
+
+Delete old resource and create new when an update fails because it `cannot change` a resources.
+
+```
+metadata.annotation.samson/force_update: "true"
+```
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -166,6 +166,16 @@ module Kubernetes
         ensure_not_updating_match_labels
         request(:update, template_for_update)
         expire_resource_cache
+      rescue Samson::Hooks::UserError => e
+        raise unless e.message.include?("cannot change")
+
+        path = [:metadata, :annotations, :"samson/force_update"]
+        if @template.dig(*path)
+          delete
+          create
+        else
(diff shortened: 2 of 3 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -311,6 +311,17 @@ def template_for_update
     end
 
     class Service < VersionedUpdate
+      def update
+        super
+      rescue Samson::Hooks::UserError => e
+        if e.message.include?("cannot change") # original error got already translated in #request
```
uthark: make it more precise: `cannot change healthCheckNodePort`

adammw:
Should have an extra-opt-in / warning if it's going to be deleting and recreating in case it drops traffic during that period.

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: it's meant as a catch-all since there are more things that will cause the "cannot change"

uthark: we may have "can't change" because of other issues.
I don't like the idea to catch all in this case.

grosser:
warning is hard because we don't have an output inside of the resource ... but I'll add a opt-in and make it more general

grosser:
💇‍♂

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: it's opt in ... so user has to decide if they want to do force update or not ... but this way we don't have to hardocde all the millions of things that are not updateable

plugins/kubernetes/README.md
```diff
@@ -294,3 +294,9 @@ of a Deployment, DaemonSet, or StatefulSet. Assuming you have Istio configured t
 MutatingWebhook in the target namespace, that should trigger Istio to inject the sidecar.
 
 To enable this functionality, set the environment variable `ISTIO_INJECTION_SUPPORTED=true`.
+
+### Forcing updates when kubernetes cannot change a resource
+
```
uthark:
```suggestion

It will delete old resource and create new.

```

uthark (approved):

PR #3564, opened 2019-10-27
Render JSON instead of HTML for 404's on json requests

cakejelly:
Im provisioning some resources through the json api and noticed the api returns html instead of json if there's a 404. This adds a rescue for project endpoints that renders a json error for json requests and raises otherwise to keep behaviour the same.

### References
- Jira link: 

### Risks


cakejelly:
@grosser nope, setting `config.consider_all_requests_local = false` does the job 🤦‍♂ Works now, thanks

grosser:
hehe, I ran into the same issue first before I figured out why it works in
prod :D

PR #3568, opened 2019-10-28
PAAS-5121 fail deploys when init containers crashed

grosser:
@zendesk/compute 

tested with 
```
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod
  labels:
    app: myapp
spec:
  containers:
  - name: myapp-container
    image: busybox
    command: ['sleep', '1000']
  initContainers:
    - name: foo
      image: busybox
      command: ['sleepnoooo']
```

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/api/pod.rb
+++ b/plugins/kubernetes/app/models/kubernetes/api/pod.rb
@@ -34,7 +34,8 @@ module Kubernetes
       end
 
       def restart_details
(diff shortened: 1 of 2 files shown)


hgokavarapuz (approved):

grosser:
FYI saw a random deploy error `Restarted (secret-puller Error)` but all logs looked good ... so something is going on with this feature or the secret puller

PR #3572, opened 2019-10-29
Remove build_id from deploys table

adammw:
Remove build_id from deploys table. Seems unused and doesn't make sense as Deploys to Builds is a many-to-many association.

### Risks
- Medium: Deploy build_id is used somewhere that we don't know of

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_10_22_023140) do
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -134,8 +133,8 @@
     t.text "env_state", limit: 16777215
     t.integer "triggering_deploy_id"
     t.boolean "redeploy_previous_when_failed", default: false, null: false
+    t.string "selected_builds"
```
grosser: nope

adammw: ugh, damn schema rb copying my database...

adammw:
Tests pass, ready for :+1: @zendesk/compute

PR #3574, opened 2019-10-29
Add many-to-many relationship between deploys and builds

adammw:
Add many-to-many relationship between deploys and builds to capture which builds are deployed by a single deploy. This is used for showing the correct "deployed to" stage in the Builds listing, and can be used in the future for ensuring redeploys redeploy the exact same build and not the latest.

Deploys with no builds default back to the old display behaviour, ie may show multiple builds deployed when only one (or none) were.

### References
- Jira link: https://zendesk.atlassian.net/browse/PAAS-4110

### Risks
- Low: Builds are not saved if they don't go through the expected code path, may show builds as being deployed when they weren't and vice-versa.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -317,15 +317,36 @@ module ApplicationHelper
 
   # show which stages this reference is deploy(ed+ing) to
   def deployed_or_running_list(stages, reference)
-    html = "".html_safe
-    stages.each do |stage|
+    deploys = stages.map do |stage|
       next unless deploy = stage.deployed_or_running_deploy
       next unless deploy.references?(reference)
+      [stage, deploy]
+    end.compact
+
+    stage_deploy_labels(deploys)
+  end
+
+  # show which stages this build is deploy(ed+ing) to
+  def deployed_or_running_builds(stages, build)
+    deploys = stages.map do |stage|
+      next unless deploy = stage.deployed_or_running_deploy
+      next unless deploy.builds.include?(build)
+      [stage, deploy]
+    end.compact
+
+    stage_deploy_labels(deploys)
+  end
+
+  def stage_deploy_labels(deploy_map)
+    html = "".html_safe
+    deploy_map.each do |stage, deploy|
       label = (deploy.active? ? "label-warning" : "label-success")
 
       text = "".html_safe
       text << stage.name
-      html << content_tag(:span, text, class: "label #{label} release-stage")
+      html << link_to([@project || deploy.project, deploy]) do
+        content_tag(:span, text, class: "label #{label} release-stage")
+      end
       html << " "
     end
     html
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -7,6 +7,8 @@ class Build < ActiveRecord::Base
   belongs_to :project, inverse_of: :builds
   belongs_to :docker_build_job, class_name: 'Job', optional: true
   belongs_to :creator, class_name: 'User', foreign_key: 'created_by', inverse_of: :builds
+  has_many :deploy_builds, dependent: nil
+  has_many :deploys, through: :deploy_builds
 
   before_validation :nil_out_blanks
   before_validation :make_default_dockerfile_and_image_name_not_collide, on: :create
--- a/app/models/build.rb
+++ b/app/models/build.rb
@@ -7,7 +7,7 @@ class Build < ActiveRecord::Base
   belongs_to :project, inverse_of: :builds
   belongs_to :docker_build_job, class_name: 'Job', optional: true
   belongs_to :creator, class_name: 'User', foreign_key: 'created_by', inverse_of: :builds
-  has_many :deploy_builds, dependent: nil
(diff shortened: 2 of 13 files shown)


app/helpers/application_helper.rb
```diff
@@ -316,11 +316,13 @@ def link_to_chart(name, values)
...
 
   # show which stages this reference is deploy(ed+ing) to
-  def deployed_or_running_list(stages, reference)
+  def deployed_or_running_list(stages, build)
     html = "".html_safe
     stages.each do |stage|
       next unless deploy = stage.deployed_or_running_deploy
-      next unless deploy.references?(reference)
+      next unless deploy.references?(build.git_sha)
+      next if deploy.builds.empty? || !deploy.builds.include?(build)
```
grosser:
```suggestion
      next unless deploy.builds.include?(build)
```

app/helpers/application_helper.rb
```diff
@@ -316,11 +316,13 @@ def link_to_chart(name, values)
   end
 
   # show which stages this reference is deploy(ed+ing) to
-  def deployed_or_running_list(stages, reference)
+  def deployed_or_running_list(stages, build)
     html = "".html_safe
     stages.each do |stage|
       next unless deploy = stage.deployed_or_running_deploy
-      next unless deploy.references?(reference)
+      next unless deploy.references?(build.git_sha)
```
grosser:
```suggestion
      next unless deploy.commit == build.git_sha
```

app/models/build.rb
```diff
@@ -7,7 +7,8 @@ class Build < ActiveRecord::Base
   belongs_to :project, inverse_of: :builds
   belongs_to :docker_build_job, class_name: 'Job', optional: true
   belongs_to :creator, class_name: 'User', foreign_key: 'created_by', inverse_of: :builds
-  has_many :deploys, dependent: nil
+  has_many :deploy_builds, dependent: nil
```
grosser: no reason to leave join associations that don't work 🤷‍♂ 
```suggestion
  has_many :deploy_builds, dependent: :destroy
```

app/models/job_execution.rb
```diff
@@ -212,7 +212,7 @@ def commands(dir)
 
   def make_builds_available
     # wait for builds to finish
-    builds = build_finder.ensure_succeeded_builds
+    builds = deploy.builds = build_finder.ensure_succeeded_builds
```
grosser: this is nice, so it will also work for bash based deployments :)

db/migrate/20191029181037_add_deploy_builds.rb
```diff
@@ -0,0 +1,9 @@
+class AddDeployBuilds < ActiveRecord::Migration[5.2]
+  def change
+    create_table :deploy_builds do |t|
+      t.references :build, index: true, null: false
+      t.references :deploy, index: true, null: false
+      t.timestamps
```
grosser: not meant to be updated, so we could drop the updated_at 🤷‍♂ 
```suggestion
      t.timestamp :created_at, null: false
```

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -15,6 +15,8 @@ class Release < ActiveRecord::Base
 
     attr_accessor :builds
 
+    after_save :save_builds
```
grosser: deploy_executor / whoever creates the release should assign deploy.builds just like job_execution does for a symmetric flow ?

grosser:
direction looks good 👍

adammw:
updated

plugins/kubernetes/app/models/kubernetes/deploy_executor.rb
```diff
@@ -17,7 +17,8 @@ class DeployExecutor
     def initialize(job, output)
       @output = output
       @job = job
-      @reference = job.deploy.reference
+      @deploy = job.deploy
```
grosser: either use `@job.deploy` or use `@deploy` everywhere for consistency

adammw: ah didn't realise how many places it was used.

app/helpers/application_helper.rb
```diff
@@ -316,16 +316,20 @@ def link_to_chart(name, values)
...
+      next unless deploy.commit == build.git_sha
+      next unless deploy.builds.include?(build)
+
       label = (deploy.active? ? "label-warning" : "label-success")
 
       text = "".html_safe
       text << stage.name
-      html << content_tag(:span, text, class: "label #{label} release-stage")
+      html << link_to([@project || deploy.project, deploy]) do
+        content_tag(:span, text, class: "label #{label} release-stage")
```
grosser: good stuff!

grosser (approved):

db/migrate/20191029181037_add_deploy_builds.rb
adammw: going to leave it, easier to go with defaults than something custom

PR #3576, opened 2019-10-29
Fix history view

uthark:
* Add description
Fix view history in Samson.

### Risks
- Low - nothing should break.

### DIFF
--- a/test/lib/samson/secrets/hashicorp_vault_backend_test.rb
+++ b/test/lib/samson/secrets/hashicorp_vault_backend_test.rb
@@ -92,6 +92,29 @@ describe Samson::Secrets::HashicorpVaultBackend do
         result.must_equal versions: {v1: {metadata: {foo: "bar", destroyed: true}}}
       end
     end
+
+    it "does not resolve deleted versions" do
+      id = "production/foo/pod2/bar"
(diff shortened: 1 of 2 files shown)


test/lib/samson/secrets/hashicorp_vault_backend_test.rb
```diff
@@ -92,6 +92,22 @@
         result.must_equal versions: {v1: {metadata: {foo: "bar", destroyed: true}}}
       end
     end
+
+    it "does not resolve deleted versions" do
+      id = "production/foo/pod2/bar"
+      versions_body = {data: {versions: {"v1" => {foo: "bar", destroyed: false, deletion_time: "2019-10-29"}, "v2" => {foo: "bar2", destroyed: false }}}}
+      version_body = {data: {data: {vault: 1}, metadata: {v2: {metadata: {foo: "bar2", destroyed: false}}}}}
```
grosser: add the `""` we saw so nobody removes the `.present?`
```suggestion
      version_body = {data: {data: {vault: 1}, metadata: {v2: {metadata: {foo: "bar2", destroyed: false, deletion_time: ""}}}}}
```

PR #3577, opened 2019-10-29
PAAS-3990 tag datadog deploy events with kubernetes project/role/team…

grosser:
… when available

@zendesk/compute 

### Risks
 - Low: deploys failing from strange resources

### DIFF
--- a/plugins/datadog/app/models/datadog_deploy_event.rb
+++ b/plugins/datadog/app/models/datadog_deploy_event.rb
@@ -14,6 +14,18 @@ class DatadogDeployEvent
         "error"
       end
 
+    tags += ["deploy"]
+
+    # for kubernetes, we also send project+team tags ... we validate they are the same in all roles/resources
+    kubernetes_template =
+      defined?(SamsonKubernetes) &&
+      deploy.kubernetes? &&
+      deploy.kubernetes_release&.release_docs&.first&.resource_template&.first
+    if kubernetes_template
+      tags << "kube_project:#{kubernetes_template.dig(:metadata, :labels, :project)}"
+      tags << "team:#{kubernetes_template.dig(:metadata, :labels, :project)}"
+    end
+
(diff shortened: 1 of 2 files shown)


plugins/datadog/app/models/datadog_deploy_event.rb
```diff
@@ -14,6 +14,18 @@ def self.deliver(deploy, tags:, time:)
...
 
+    tags += ["deploy"]
+
+    kubernetes_template =
+      defined?(SamsonKubernetes) &&
+      deploy.kubernetes? &&
+      deploy.kubernetes_release&.release_docs&.first&.resource_template&.first
+    if kubernetes_template
+      tags << "kube_project:#{kubernetes_template.dig(:metadata, :labels, :project)}"
+      tags << "kube_role:#{kubernetes_template.dig(:metadata, :labels, :role)}"
```
adammw: how does this work with multiple roles being deployed? need to loop through the resources...

grosser: yeah that won't work ... we can do project/team first and see if that is good enough
we could tag a single event with all roles that were deployed, but that's kinda fishy / means we have to parse all the docs (5 clusters * 3 roles * 3 resources = 45 resources 🤷‍♂ )

plugins/datadog/app/models/datadog_deploy_event.rb
```diff
@@ -14,6 +14,17 @@ def self.deliver(deploy, tags:, time:)
         "error"
       end
 
+    tags += ["deploy"]
+
+    kubernetes_template =
+      defined?(SamsonKubernetes) &&
+      deploy.kubernetes? &&
+      deploy.kubernetes_release&.release_docs&.first&.resource_template&.first
```
adammw: maybe a comment on how we assume all roles / templates will be the same / already have a validaiton to enforce that so we can do this

PR #3579, opened 2019-10-29
Allow show action for inbound webhooks

cakejelly:
I'd like to be able to fetch the state of a single webhook for a given project with the json api. This PR allows requests to `GET /projects/:id/webhooks/:webhook_id`

Not sure what to do about html requests though.. previously it would 404 in the web app because the route wasn't defined but now that it is, it'll respond with 406 because there's no view. I think it's not a problem though as this view has never existed so nobody should try to reach it.

One option to handle it a bit more gracefully would be to add something like the following to webhooks controller
```ruby
def show
  respond_to do |format|
    format.html { head :not_found } # or something similar  
    format.json { render_resource_as_json } 
  end
end
```

### References
- Jira link: 

### Risks
- Low: thing that could happen

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -47,7 +47,7 @@ Samson::Application.routes.draw do
 
     resource :changelog, only: [:show]
     resource :stars, only: [:create]
-    resources :webhooks, only: [:index, :create, :update, :destroy]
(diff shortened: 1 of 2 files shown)


grosser:
looks good!

PR #3580, opened 2019-10-30
PAAS-5085 add input for namespace metadata so we can track team/proje…

grosser:
…ct/product

Screen Shot 2019-10-30 at 9 46 59 AM

@zendesk/compute 

### Risks
 - Low: namespace getting modified by accident/breaking other tools/integrations

### DIFF
--- a/app/views/commands/_form.html.erb
+++ b/app/views/commands/_form.html.erb
@@ -5,7 +5,7 @@
     <fieldset>
       <%= render 'shared/errors', object: @command %>
 
-      <%= form.input :command, as: :text_area, input_html: {rows: 10, style: 'width: 800px'} %>
+      <%= form.input :command, as: :text_area, input_html: {rows: 10, style: "width: 800px", class: "form-control monospace"} %>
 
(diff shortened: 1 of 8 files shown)


uthark:
Does it support env substitution? We may want to have different labels per cluster/deploy group.

uthark (approved):

grosser:
- so far nobody wanted different labels per cluster ... so I'd leave that for next PR
 - not sure what "values from git" is, but it only patches, so it will merge with existing values from the cluster ... for example it will not remove labels/annotations that are already set, but override them when they are set

uthark:
Ok, one more question - what about annotations/labels created by different controllers? update should be a PATCH ideally.

grosser:
- yes in the db
 - db is easier since that's where the name is stored too ... we can talk about moving config into git, but that's a different usecase / not in scope for this change 🤷‍♂

grosser:
"annotations/labels created by different controllers" will not be touched since it uses "PATCH"

PR #3581, opened 2019-10-30
New plugin - shared outbound hooks

hdt91:
This plugin is a proposed implementation to support a use case, in which some projects need to perform provisioning or setup work outside Samson before a code deployment can happen. This work can be common across projects so it might be a good idea to define them separately instead of letting each project store its own hooks using the existing outbound webhook feature.

I'd like to get the feedback about this plugin as it is currently serving as the PoC reference for the feature requesters, and of course we are open to any implementation change suggestion. Also, the PR is a bit big due to some boilerplates so I can split that into smaller PRs if there is a need for that.

Checklist
- [ ] Refine implementation or requirements based on PR feedback
- [ ] Add proper tests with the final implementation
- [ ] Add any feature toggling or safeguard knob if needed

The implementation synopsis: this plugin is almost a clone of the environment variable group, but the hook record is associated with the stage instead. Samson now fires a new event called `:before_deploy_setup` that happens before the repo setup logic, triggering the new plugin. This plugin tests if there is a hook setup with the current stage, and do the following blocking logic before moving to the normal setup work:
- Send the standard deployment payload using the setup outbound hook, with a few dumb http retries
- The plugin expects the hook returns successfully with a status poll URL, then uses this to keep polling for the progress as this might take awhile.
- The code deployment only continues if there is no hook setup, or the entire polling logic above terminates with successful status.

The plugin also adds new UIs to manage and setup these setup hooks
#### New entry in dropdown menu
Screenshot 2019-10-30 at 10 12 54 AM

#### CRUD menu for the hook records
Screenshot 2019-10-30 at 10 13 18 AM

#### New entry in stage setup
Screenshot 2019-10-30 at 10 14 37 AM

#### The polling log happens before the code deployment
Screenshot 2019-10-30 at 10 17 07 AM

### References
- Internal documentation about the proposed implementation.

### Risks
- Medium: code is buggy and breaks Samson
- Low: the setup in each stage is opt-in and user can leave the selection empty to disable the hook.

@zendesk/interface @zendesk/eng-productivity @grosser


plugins/external_setup_hook/app/decorators/stages_controller_decorator.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+StagesController.class_eval do
```
grosser: `StagesController.prepend(Module.new do` <- simpler ?

plugins/external_setup_hook/app/decorators/stages_controller_decorator.rb
```diff
@@ -0,0 +1,20 @@
...
+  prepend(
+    Module.new do
+      def resource_params
+        stage_params = super
+        stage_params[:stage_external_setup_hook_attributes]['_destroy'] = '1' if setup_hook_empty?(stage_params)
```
grosser: that would no longer be necessary with has_many ... so another good reason :)

plugins/external_setup_hook/app/models/external_setup_hook.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class ExternalSetupHook < ActiveRecord::Base
+  validates :name, presence: true
+
+  validate :validate_endpoint
+
+  has_many :stage_external_setup_hooks, dependent: :destroy
+  has_many :stages, through: :stage_external_setup_hooks, dependent: :destroy
```
grosser: yeah we don't want to delete all stages when a hook is deleted :)
```suggestion
  has_many :stages, through: :stage_external_setup_hooks
```

plugins/external_setup_hook/app/models/external_setup_hook.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def validate_endpoint
+    errors.add(:auth_type, "is unsupported") if auth_type.downcase! != 'token'
```
grosser: I'd prefer not to downcase ... better have ppl get this right instead of guessing and then ending up with strange errors

plugins/external_setup_hook/lib/samson_external_setup_hook/samson_plugin.rb
```diff
@@ -0,0 +1,106 @@
...
+  POLL_INTERVAL = 15
+  POLL_WAIT_TIME = 30
+  TRIGGER_WAIT_TIME = 30
+
+  def self.wait_for_external_setup(job, output, stage, ctx_data)
+    status_poll_url = nil
+    success = false
+
+    setup_hook = stage.external_setup_hook
+    conn = Faraday.new(:ssl => {:verify => setup_hook.verify_ssl})
```
grosser:
```suggestion
    conn = Faraday.new(ssl: {verify: setup_hook.verify_ssl})
```

plugins/external_setup_hook/lib/samson_external_setup_hook/samson_plugin.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    return true if setup_hook.nil?
```
grosser: do that before making a connection object

plugins/external_setup_hook/lib/samson_external_setup_hook/samson_plugin.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    trigger_wait_time = TRIGGER_WAIT_TIME
+    until trigger_wait_time < 0 do
+      trigger_wait_time -= POLL_INTERVAL
+
+      response = conn.post(setup_hook.endpoint, ctx_data.to_json) do |req|
+        req.options.timeout = req.options.open_timeout = 5
+        req.headers['Content-Type'] = 'application/json'
+
+        if setup_hook.auth_type == "token"
```
grosser: maybe just have a single column for "authorization header" where users can enter `Token fooo`
... or otherwise make sure to also support "Bearer" and "Token"

plugins/external_setup_hook/lib/samson_external_setup_hook/samson_plugin.rb
```diff
@@ -0,0 +1,106 @@
...
+          req.headers['Authorization'] = "token #{setup_hook.auth_token}"
+        end
+      end
+      # TODO: handle more failure case based on requirements
+      if response.status == 200
+        body = JSON.parse response.body
+        status_poll_url = body['status_poll_url'] || nil
```
grosser: should be mentioned on the hook setup page how this works / which fields are required ... feels a little magic

plugins/external_setup_hook/lib/samson_external_setup_hook/samson_plugin.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      break unless status_poll_url.nil?
+      sleep POLL_INTERVAL
+    end
+
+    return false if status_poll_url.nil?
+
+    output.write("Polling external setup status through #{status_poll_url}\n")
+
+    poll_wait_time = POLL_WAIT_TIME
```
grosser: looks kinda similar ... extract method ?

plugins/external_setup_hook/lib/samson_external_setup_hook/samson_plugin.rb
```diff
@@ -0,0 +1,106 @@
...
+
+module SamsonExternalSetupHook
+  class Engine < Rails::Engine
+  end
+
+  POLL_INTERVAL = 15
+  POLL_WAIT_TIME = 30
+  TRIGGER_WAIT_TIME = 30
+
+  def self.wait_for_external_setup(job, output, stage, ctx_data)
```
grosser: smaller methods + some method docs would help readability

grosser:
... and as usual: prefer to brainstorm together for a few minutes before making big PRs that might go in the wrong direction 🤷‍♂

hdt91:
@zendesk/interface I just had a quick chat with @grosser and there is an opportunity to explore about tweaking existing outbound webhook feature gradually to adapt to your requirement, for example, adding options to make it shareable. However this might deviate from the current workflow depicted by the UI so if you have any concern about this, let's bring it to the discussion.

PR #3584, opened 2019-10-31
copy stage deploy group roles config when cloning stages

grosser:
@zendesk/compute 

### Risks
 - None

### DIFF
--- a/plugins/kubernetes/test/samson_kubernetes/samson_plugin_test.rb
+++ b/plugins/kubernetes/test/samson_kubernetes/samson_plugin_test.rb
@@ -93,6 +93,12 @@ describe SamsonKubernetes do
       ignore = ['id', 'created_at', 'updated_at', 'deploy_group_id']
       new_stage_dgr.attributes.except(*ignore).must_equal old_stage_dgr.attributes.except(*ignore)
     end
+
+    it "copies kubernetes_stage_roles" do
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/test/samson_kubernetes/samson_plugin_test.rb
```diff
@@ -93,6 +93,12 @@ def stage_clone(old_stage, new_stage)
       ignore = ['id', 'created_at', 'updated_at', 'deploy_group_id']
       new_stage_dgr.attributes.except(*ignore).must_equal old_stage_dgr.attributes.except(*ignore)
     end
+
+    it "copies kubernetes_stage_roles" do
+      @old_stage.kubernetes_stage_roles.create!(kubernetes_role: kubernetes_roles(:app_server))
+      stage_clone(@old_stage, @new_stage)
+      @new_stage.kubernetes_stage_roles.size.must_equal 1
```
hgokavarapuz: maybe better to check if the role is equal to app_server rather than checking just the count. Will catch if there are any bugs in future.

hgokavarapuz (approved):

plugins/kubernetes/test/samson_kubernetes/samson_plugin_test.rb
grosser: good idea, done

hgokavarapuz (approved):

PR #3585, opened 2019-10-31
do not put fieldset into a fieldset or collapse_fieldsets will parse …

grosser:
…the legend twice

also only attach to legends directly in fieldset or folding logic will not work anyways

Fixes Screen Shot 2019-10-31 at 11 07 33 AM

@zendesk/compute 

### Risks
 - Low: some fieldsets not folding

### DIFF
--- a/app/views/deploys/new.html.erb
+++ b/app/views/deploys/new.html.erb
@@ -12,67 +12,65 @@
     }
   %>
   <%= form_for [@project, @stage, @deploy], html: { class: "form-horizontal collapse_fieldsets" }, data: data do |form| %>
-    <fieldset>
-      <%= render 'shared/errors', object: @deploy %>
+    <%= render 'shared/errors', object: @deploy %>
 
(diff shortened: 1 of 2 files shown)


app/views/deploys/new.html.erb
```diff
@@ -12,67 +12,65 @@
...
-        </div>
+    <div id="deploy-confirmation">
+      <ul class="nav nav-tabs">
+        <%= render 'changeset/tab_list' %>
+      </ul>
+      <div class="tab-content deploy-details">
+        <div class="tab-pane changeset-placeholder">Loading...</div>
       </div>
-    </fieldset>
+    </div>
```
hgokavarapuz: should have been easy if there were comments like  hard to check if the syntax is correct in this way. Replacing fieldset with div.

grosser: github has a "no whitespace" option for diffs

PR #3587, opened 2019-10-31
Report tag resolver error

Strainy:
As per internal discussion @grosser, would be good to know the error details here.

### Risks
- Low: should simply print error on raise


grosser:
https://github.com/zendesk/samson/pull/3605

PR #3589, opened 2019-11-01
Decorate classes after they are fully loaded to avoid bugs/hacks

grosser:

atm we have to re-define soft_deletion in plugins which is ugly/magic ... also the whole workflow was magic

### Risks
 - Low: callbacks not working because their ordering changed


yizhang-zen (approved):
🌟

PR #3591, opened 2019-11-01
do not crash with 404 errors when trying to create a namespace-less r…

grosser:
…esource with a namespace

we saw 404 when trying to create a namespace-less resource with a namespace ... kubectl is smart enough to remove the namespace, but I'd prefer if we stop user early that did not understand they are creating a global resource

```
the server could not find the requested resource
```

@zendesk/compute @swapnil-zendesk


yizhang-zen (approved):
💯

uthark (approved):

PR #3595, opened 2019-11-02
Filter env-variables group by project

sathishavm:
We need env-variable group names used by the project.
This PR gives the ability to filter the env-variable groups for the project by passing `project_id` in parameters.
`eg: /environment_variable_groups.json?project_id=1` 

### Risks
- Low


sathishavm:
@grosser The [PR](https://github.com/zendesk/samson/pull/3559) is helpful to pull values, we need Group names to resolve and manage values.

PR #3596, opened 2019-11-03
Use a single read-multi to render project show

grosser:
- single request to memcached instead of 1 (project)+N(stages)
 - fix caching logic to not cache deploy buttons when user is a viewer

```
Cache read_multi: views/projects/_stage: ...
```

see https://blog.bigbinary.com/2016/03/09/rails-5-makes-partial-redering-from-cache-substantially-faster.html for details

@zendesk/compute

### DIFF
--- a/app/views/projects/_stage.html.erb
+++ b/app/views/projects/_stage.html.erb
@@ -1,5 +1,5 @@
-<tr>
-  <% cache [@project.permalink, stage, Lock.cache_key] do %>
+<% cache [@project.permalink, stage, Lock.cache_key, deployer_for_project?] do %>
+  <tr>
     <td>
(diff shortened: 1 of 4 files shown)


app/views/projects/_stage.html.erb
```diff
@@ -1,6 +1,7 @@
-<tr>
-  <% cache [@project.permalink, stage, Lock.cache_key] do %>
+<% cache [@project.permalink, stage, Lock.cache_key, deployer_for_project?] do %>
+  <tr>
     <td>
+      CACHE2 <%= rand %>
```
adammw:
```suggestion
```

grosser: hehe

PR #3600, opened 2019-11-04
Hack: set columns for failed migration to unblock deploys

grosser:


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -154,6 +155,14 @@ def disruption_budget_target(deployment)
       end
     end
 
+    def backfill_broken_columns
+      columns = self.class.columns.map(&:name)
+      p columns
```
sathishavm: remove this puts?

sathishavm (approved):

PR #3608, opened 2019-11-05
Reusable outbound webhooks

grosser:
@hdt91 

TODO:
 - deleting a stage leaves the outboundwebhook behind 🤷‍♂

### DIFF
--- a/app/controllers/outbound_webhooks_controller.rb
+++ b/app/controllers/outbound_webhooks_controller.rb
@@ -7,27 +7,52 @@ class OutboundWebhooksController < ResourceController
   before_action :authorize_project_deployer!, except: [:index]
   before_action :set_resource, only: [:show, :edit, :update, :destroy, :new, :create]
 
-  def index
-    respond_to do |format|
-      format.json { render_as_json :webhooks, @project.outbound_webhooks, nil }
-    end
-  end
-
+  # creation is done from project webhooks page
   def create
     super(template: 'webhooks/index')
   end
 
+  def destroy
+    if @stage # unlink (does not handle json or failure)
+      @resource.outbound_webhook_stages.where(stage: @stage).destroy_all
+      @resource.destroy! unless @resource.global?
+      redirect_to project_webhooks_path(@stage.project), notice: "Deleted"
+    else
+      super
+    end
+  end
+
   private
+
+  def search_resources
+    scope = super
+    if project_id = params[:project_id]
+      stage_ids = Project.find_by_param!(project_id).stages.pluck(:id)
+      ids = OutboundWebhookStage.where(stage_id: stage_ids).pluck(:outbound_webhook_id).uniq
+      scope = scope.where(id: ids)
+    end
+    scope
+  end
 
   def resource_path
     resources_path
   end
 
   def resources_path
-    [@project, 'webhooks']
+    @project ? [@project, 'webhooks'] : super
   end
 
   def resource_params
-    super.permit(:stage_id, :url, :username, :password).merge(project: current_project)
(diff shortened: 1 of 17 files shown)


hdt91:
This PR is the start of the attempt to rework existing outbound webhook feature to support external pre-deployment setup use case. The first task is to move this type of webhook beyond the project scope to make it shareable across different stages in different projects. This should help simplify the setup for many application by reusing common setup hooks.

@zendesk/interface

app/controllers/outbound_webhooks_controller.rb
```diff
@@ -28,6 +44,20 @@ def resources_path
...
   def resource_params
-    super.permit(:stage_id, :url, :username, :password).merge(project: current_project)
+    link = OutboundWebhookStage.new(stage: @stage)
+    super.permit(:url, :username, :password).merge(outbound_webhook_stages: [link])
+  end
+
+  def require_project
+    case action_name
+    when "create"
+      @stage = Stage.find(params[:outbound_webhook].delete(:stage_id))
```
rbayerl: This looks backwards

grosser: stage_id is a fake attribute ... but yeah we can just submit it as a regular param and avoid this delete magic 👍

grosser:
Expanded the addition UI since a single line won't be enough anymore:
Screen Shot 2019-11-06 at 5 25 14 PM

"Delete" will be an unlink when the hook is global:
Screen Shot 2019-11-06 at 5 25 02 PM

New overview UI that shows all webhooks and how often they are used:
Screen Shot 2019-11-06 at 6 49 00 PM

New menu entry to find that:
Screen Shot 2019-11-06 at 5 05 33 PM

New edit UI for global webhooks (also added auditing):
Screen Shot 2019-11-06 at 5 05 29 PM

app/controllers/outbound_webhooks_controller.rb
```diff
@@ -7,27 +7,55 @@ class OutboundWebhooksController < ResourceController
   before_action :authorize_project_deployer!, except: [:index]
   before_action :set_resource, only: [:show, :edit, :update, :destroy, :new, :create]
 
-  def index
```
grosser: index json reply changed from `webhooks` -> `outbound_webhooks` ... also no more nested under `projects/foo/`

app/controllers/outbound_webhooks_controller.rb
```diff
@@ -7,27 +7,55 @@ class OutboundWebhooksController < ResourceController
...
 
   def resources_path
-    [@project, 'webhooks']
+    @project ? [@project, 'webhooks'] : super
   end
 
   def resource_params
-    super.permit(:stage_id, :url, :username, :password).merge(project: current_project)
+    allowed = [:url, :username, :password]
+    allowed << :global if action_name == "create"
```
grosser: not allowing to edit global or we could get into weird situations where something is global and then gets made non-global

app/controllers/outbound_webhooks_controller.rb
```diff
@@ -7,27 +7,55 @@ class OutboundWebhooksController < ResourceController
...
+    params = super.permit(*allowed)
+    params[:stages] = [@stage] if @stage && action_name == "create"
```
grosser: auto-linking to the stage the request came from

rbayerl:
Would it be possible to add Bearer token auth as an option? That's what we're hoping to use.

grosser:
yes, that's for next PR :)

hdt91:
👍

Issue #3402, opened 2019-06-17
"New button" on secrets page does not work on small screens


lukestephenson-zendesk:
If I split my screen 50/50, the `New` button on https://samson.zende.sk/secrets no longer responds to clicks.

![image](https://user-images.githubusercontent.com/48660846/59572764-a2d71200-90f2-11e9-871d-82e7bc80259a.png)

At full screen it works, but I wanted side by side with some instructions from confluence that I was trying to follow.

lukestephenson-zendesk:
Same on the secret sharing grants screen https://samson.zende.sk/secret_sharing_grants

grosser:
yeah this is all kinda messy, the whole search UI needs some more consistency :(

also the search_form `capture(&block) << content_tag(:div, button, class: "col-md-1 clearfix")` needs to lose the clearfix I think ...

I poked it a bit, but seems not to be easy ... adding z-index did not work either, I think first the search UIs need some more consistency before we go into more details

PR #3610, opened 2019-11-07
Fix unclickable button

rubynho:
### Description

**Problem:** Pulling the new button to the right is causing the search button div get on top of the new button when in half screen size, turning them unclickable.

**Solution:** Adjust the new and search buttons size so they can fit side by side in small screens.

### UI screenshots
image

image

### References
- Solves #3402.

### Risks
- No risks, just a UI change.


grosser:
search button should come after the search fields ... make both float left ?

rubynho:
> search button should come after the search fields ... make both float left ?

It works but I think we could just give space to the new button fits next to the search button, like so:

image

Then the search button comes first even when in half screen.

grosser:
looks good too :)

rubynho:
Hey @grosser, what's wrong with Travis? :thinking:

grosser:
random error, restarted it

PR #3618, opened 2019-11-08
Show webhook delivery in deploy log

grosser:
Screen Shot 2019-11-08 at 2 19 59 PM


grosser:
delivery timeout is the standard faraday timeout ... we can configure it if it ever causes trouble 🤷‍♂

PR #3620, opened 2019-11-08
Make outbound webhook errors stop the deploy when used before deploy

grosser:
also show connection errors in the job output just like request errors (500 etc)

@zendesk/compute @hdt91

### DIFF
--- a/app/models/outbound_webhook.rb
+++ b/app/models/outbound_webhook.rb
@@ -20,17 +20,25 @@ class OutboundWebhook < ActiveRecord::Base
     prefix = "Webhook notification:"
     output.puts "#{prefix} sending to #{url} ..."
 
-    response = connection.post url do |req|
-      req.headers['Content-Type'] = 'application/json'
-      req.body = self.class.deploy_as_json(deploy).to_json
-    end
+    error_message =
+      begin
+        response = connection.post url do |req|
+          req.headers['Content-Type'] = 'application/json'
+          req.body = self.class.deploy_as_json(deploy).to_json
+        end
 
-    if response.success?
-      output.puts "#{prefix} succeeded"
-    else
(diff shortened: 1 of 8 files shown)


hdt91:
This is a continuation of:
- Extra auth options are added in https://github.com/zendesk/samson/pull/3616
- Add an option to fire the outbound webhook before the deployment https://github.com/zendesk/samson/pull/3619

This changeset is part of the new semantics of outbound webhook. Here are the key points
- For both pre and post deployment hooks, any failure would stop all pending hooks (as they are fired linearly)
- For pre deployment hooks, a failure always stops the deployment. Basically a deployment will continue only if nothing goes wrong during the delivery, including the polling logic.
- The polling logic, including the relay of status messages from the other end, will likely be kept simple in another PR

@zendesk/interface

app/models/outbound_webhook.rb
```diff
@@ -20,17 +20,25 @@ def deliver(deploy, output)
...
-    end
+    error_message =
+      begin
+        response = connection.post url do |req|
+          req.headers['Content-Type'] = 'application/json'
+          req.body = self.class.deploy_as_json(deploy).to_json
+        end
+
+        if response.success?
+          output.puts "#{prefix} succeeded"
```
rbayerl: This is expecting the webhook to do all of it's logic end to end before reporting success or failure. Would prefer something like the image builds where we can report back a status or we can provide a url to poll. Things on our end happen asynchronously and we can't be sure that the pod that receives the complete (or error) message is the same pod with the open connection.

grosser:
makes sense, we'll add that next PR :)

app/models/outbound_webhook.rb
hdt91: There will be another PR that allows the receiver to provide a url to poll.

PR #3622, opened 2019-11-11
Allow connecting global webhooks to other stages

grosser:
Screen Shot 2019-11-10 at 3 31 55 PM

and when it fails:
Screen Shot 2019-11-10 at 4 24 09 PM


grosser:
@hdt91

PR #3623, opened 2019-11-11
Allow to patch volumes and initContainers in StatefulSet

uthark:
* Volumes can be PATCHed.
* JIRA: PAAS-5351

### Risks
- Low - build may break if trying to patch stateful set.


adammw:
any test updates?

adammw (approved):

PR #3624, opened 2019-11-11
Allow outbound webhooks to poll status url

grosser:
Screen Shot 2019-11-11 at 2 33 09 PM

@hdt91 
/cc @zendesk/compute @zendesk/interface

### DIFF
--- a/app/controllers/outbound_webhooks_controller.rb
+++ b/app/controllers/outbound_webhooks_controller.rb
@@ -56,7 +56,7 @@ class OutboundWebhooksController < ResourceController
   end
 
   def resource_params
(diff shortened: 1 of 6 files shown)


adammw:
Explain more?

adammw:
maybe a readme?

grosser:
the `i` icon has a short explanation

the outbound webhook can poll a returned url for setup steps that take long

rbayerl:
How would this support a per deploy id check url?

grosser:
each request would return it's own check url ... which should include the
deploy-id or something else unique

PR #3625, opened 2019-11-12
Remove gcr image scanning functionality since we use kritis now

grosser:
This feature has a lot of false-positives since it shows all vulnerabilities and does not offer allowing some / does not ignore "unfixable" vulernerabilites ... so let's remove it since we won't invest more into it and the current state is confusing/mostly unusable.

@zendesk/secure @zendesk/compute @zendesk/eng-productivity \

### RIsks
 - Low: projects that rely on it will start deploying with vulnerabilities if they ignored kritis


hdt91 (approved):
👍 on 🍏

PR #3626, opened 2019-11-12
Environment variable group owners

sathishavm:
Environment variable group(i.e shared environment variables) has been shared by different projects. In this PR, added a field to add owner(s) for the env group.

This will also help to identify the owner(i.e CODEOWNERS) of the shared environment variables in GitHub repository.

Screen Shot 2019-11-12 at 1 09 44 PM

Risk: **Low**


grosser:
why can't this be:
 - a string field
 - a part of the comment like `owner: fooo`

sathishavm:
Oh, This may need to compared and validated with `git codeowners` and based on our discussion with the config team and all other vendors, everyone considered to make ownership to be compulsory for all shared config(maybe future).

grosser:
multiple owners is an edge-case, so no need to make it super convenient ... I'd suggest doing comma separated string ... no need for extra UI and only a single new column, also the export via api would be simple

grosser:
https://github.com/zendesk/samson/pull/3628

PR #3627, opened 2019-11-13
Outbound hook adjustments

hdt91:
- Include deploy groups via stage in the outbound payload
- Add sleep period in the hook polling
- Treat `202` as the processing signal instead of `102` as it chokes `Faraday` with an exception.

### DIFF
--- a/app/models/outbound_webhook.rb
+++ b/app/models/outbound_webhook.rb
@@ -64,6 +64,10 @@ class OutboundWebhook < ActiveRecord::Base
 
   private
 
+  def poll_period
+    @poll_period ||= Integer(ENV['OUTBOUND_WEBHOOK_POLL_PERIOD'] || '30')
+  end
+
   def post_hook(deploy)
     connection.post url do |req|
       req.headers['Content-Type'] = 'application/json'
@@ -71,12 +75,13 @@ class OutboundWebhook < ActiveRecord::Base
     end
   end
 
-  # loop forever until user cancels the deploy or deploy times out
+  # loop forever, with sleeping in between, until user cancels the deploy or deploy times out
   def poll_status_url(url)
     loop do
       response = connection.get(url)
       yield response.body
-      break if response.success?
+      break if response.success? && response.status != 202
+      sleep poll_period
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/outbound_webhook.rb
```diff
@@ -63,19 +63,24 @@ def as_json(*)
...
   end
 
-  # loop forever until user cancels the deploy or deploy times out
+  # loop forever, with sleeping in between, until user cancels the deploy or deploy times out
   def poll_status_url(url)
     loop do
       response = connection.get(url)
       yield response.body
-      break if response.success?
+      break if response.success? && response.status != 202
```
grosser:
```
if response.success?
  break if response.status != 202
else
  raise Samson::Hooks::UserError, "error polling status endpoint"
end
```

app/models/stage.rb
```diff
@@ -211,6 +211,10 @@ def locked_by?(lock)
     end
   end
 
+  def as_json(options = {})
+    super(options).merge(deploy_groups: deploy_groups.as_json)
```
grosser: don't include things by default into every stage reply ... manually add it in the webhook or something like that

test/models/outbound_webhook_test.rb
```diff
@@ -143,14 +143,16 @@ def assert_webhook_fired(&block)
         assert_request :post, "https://testing.com/deploys", to_return: polling_target, &block
       end
 
+      with_env OUTBOUND_WEBHOOK_POLL_PERIOD: '1'
```
grosser: - test the polling work by expecting sleeps
 - add a `expects(:sleep).never` to the default before so we don't sleep forever when something goes wrong

app/models/outbound_webhook.rb
hdt91: Do you want to raise here? This is different from the current behaviour where it would retry for non-exception errors like 5xx.

grosser: I'd say just raise for now ... we can deal with the edge-cases later if we have to, for now I'd assume the service we are talking to is reliable :)

hdt91:
💇‍♂ @grosser @zendesk/eng-productivity @zendesk/interface

PR #3634, opened 2019-11-14
PAAS-5207 supported secrets with unknown project/group/env

grosser:
- no longer shows "first if not found"
 - correctly mark label as required `*`

before:
Screen Shot 2019-11-13 at 5 08 43 PM

after:
Screen Shot 2019-11-13 at 5 07 20 PM

 - show all available envs/deploy-groups in search:
Screen Shot 2019-11-13 at 5 19 29 PM

 - allow deletion when no server is deemed responsible for a secret by deleting from all + make "no vault server found" more specific

reproduce with:
```
Samson::Secrets::DbBackend::Secret.find('master/example-kubernetes/groupk/dfdffd').update_column(:id, 'nope/nope/nope/nope')
```

@zendesk/compute

### DIFF
--- a/app/views/secrets/show.html.erb
+++ b/app/views/secrets/show.html.erb
@@ -19,21 +19,35 @@
 
   <%= form_for OpenStruct.new(secret), as: :secret, url: url, method: method, html: {class: "form-horizontal"} do |form| %>
     <fieldset>
-      <%= form.input :environment_permalink, label: "Environment" do %>
-        <%= form.select :environment_permalink,
-          options_for_select(environment_permalinks, secret[:environment_permalink]),
-          {include_blank: true}, class: "form-control", required: true, disabled: !!id %>
+      <%= form.input :environment_permalink, label: "Environment", required: true do %>
+        <% if id %>
+          <%= form.text_field :environment_permalink, disabled: true, class: "form-control" %>
+        <% else %>
+          <%= form.select :environment_permalink,
+            options_for_select(environment_permalinks, secret[:environment_permalink]),
+            {include_blank: true}, required: true, **Samson::FormBuilder::LIVE_SELECT_OPTIONS %>
+        <% end %>
       <% end %>
 
-      <%= form.input :project_permalink, label: "Project" do %>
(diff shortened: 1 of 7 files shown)


lib/samson/secrets/vault_client_manager.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'samson/secrets/vault_logical_wrapper'
```
grosser: test was complaining about circular dependencies

lib/samson/secrets/vault_client_manager.rb
```diff
@@ -105,16 +116,16 @@ def responsible_clients(id)
...
-              raise "no environment with permalink #{environment_permalink} found"
+            unless deploy_group_permalinks = client_map[:environments][environment_permalink].presence
+              raise VaultServerNotConfigured, "no environment with permalink #{environment_permalink} found"
             end
             deploy_group_permalinks.map { |p| client(p) }.uniq
           end
         else
           [client(deploy_group_permalink)]
-        end.presence || raise("no vault servers found for #{id}")
+        end
```
grosser: each branch now takes care of missing servers so we never have to say "idk what happened"

app/views/secrets/show.html.erb
```diff
@@ -19,21 +19,35 @@
...
-        <%= form.select :deploy_group_permalink,
-          options_for_select([['Loading ...', secret[:deploy_group_permalink]]], secret[:deploy_group_permalink]),
-          {include_blank: true}, class: "form-control", required: true, disabled: !!id %>
+      <%= form.input :deploy_group_permalink, label: "Deploy Group", required: true do %>
+        <% if id %>
+          <%= form.text_field :deploy_group_permalink, disabled: true, class: "form-control" %>
+        <% else %>
+          <%= form.select :deploy_group_permalink,
+            options_for_select([['Loading ...', secret[:deploy_group_permalink]]], secret[:deploy_group_permalink]),
+            {include_blank: true}, class: "form-control", required: true %>
```
grosser: the only one that cannot be a live select since we dynamically change the select values

PR #3635, opened 2019-11-14
PAAS-5242 do not enforce env/secrets when deleting

grosser:
@zendesk/compute 

### Risks
 - None


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -339,7 +339,7 @@ def set_replica_target
     end
 
     def validate_replica_target_is_supported
-      return if @doc.replica_target == 1 || (@doc.replica_target == 0 && @doc.delete_resource)
+      return if @doc.replica_target == 1 || @doc.delete_resource
```
grosser: validations enforce that delete_resource is only used for 0 replicas

uthark (approved):
💇

PR #3637, opened 2019-11-15
Simplify flashes and add warning flash

grosser:
- layout should not have special logic to support certain controllers
 - flashes should be standardized
 - we need warning flashes for ignorable errors

@zendesk/compute 

Screen Shot 2019-11-15 at 1 44 41 PM
Screen Shot 2019-11-15 at 2 37 15 PM

### DIFF
--- a/plugins/kubernetes/app/controllers/kubernetes/namespaces_controller.rb
+++ b/plugins/kubernetes/app/controllers/kubernetes/namespaces_controller.rb
@@ -10,10 +10,10 @@ class Kubernetes::NamespacesController < ResourceController
 
   def sync_all
     clusters = Kubernetes::Cluster.all.to_a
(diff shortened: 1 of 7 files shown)


eatwithforks (approved):

plugins/kubernetes/app/controllers/kubernetes/namespaces_controller.rb
```diff
@@ -54,9 +54,9 @@ def upsert_namespace(clusters, namespace)
     end.compact
   end
 
-  def show_namespace_errors(errors)
+  def show_namespace_warnings(errors)
```
eatwithforks:
```suggestion
  def show_namespace_warnings(warnings)
```

PR #3643, opened 2019-11-19
PAAS-5262 stop deploys when bad events happen

grosser:
instead of running forever and not telling the user which pod failed with which events

broken through https://github.com/zendesk/samson/pull/3497 were the line was removed without breaking any tests or adding new tests

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource_status.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource_status.rb
@@ -54,6 +54,7 @@ module Kubernetes
           @details = "Waiting for resources (#{@pod.phase}, #{@pod.reason})"
         elsif @pod.events_indicate_failure?
           @details = "Error event"
+          @finished = true
         else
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/resource_status.rb
```diff
@@ -54,6 +54,7 @@ def check
           @details = "Waiting for resources (#{@pod.phase}, #{@pod.reason})"
         elsif @pod.events_indicate_failure?
           @details = "Error event"
+          @finished = true
```
yizhang-zen: Adding pod status and failure event in details for better visible?

yizhang-zen (approved):

plugins/kubernetes/app/models/kubernetes/resource_status.rb
grosser: both will be shown in the failure description

PR #3645, opened 2019-11-24
bump doorkeeper for rails 6

grosser:
@zendesk/compute 

with bugfix for https://github.com/doorkeeper-gem/doorkeeper/pull/1331

 - [doorkeeper](https://rubygems.org/gems/doorkeeper) [v4.4.0 -> v5.2.2](https://github.com/doorkeeper-gem/doorkeeper/compare/v4.4.0...v5.2.2)

 - checked that creating and using a new app works + is not allowed for non-admins

### RIsks
 - Low: auth tokens not working / having the wrong scope / expiring too early


app/controllers/access_tokens_controller.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
```
eatwithforks: trailing space change?

grosser: nah, just a newline because I added the hack here originally and then removed it :D

PR #3646, opened 2019-11-25
Add deploy_groups includes for stages endpoints

cakejelly:
When creating/reading a stage with deploy groups it would be nice to be able to retrieve the deploy group ids in the response to avoid additional requests. The main reason i'd like this is that i'm writing a terraform provider to provision samson resources and I need to handle partial updates without this, which adds a bit of complexity

### Risks


app/controllers/stages_controller.rb
```diff
@@ -56,6 +54,16 @@ def clone
     end
   end
 
+  def allowed_includes
```
grosser: move into private

Issue #3648, opened 2019-11-25
Slack webhook URL link does not support ports


benrudolph:
Currently when using the slack webhook integration, the formatted message uses the hostname without the port and leads to broken URLs.

Example:
```
[Peregrine Frontend] Samson is cancelling Ben Rudolph's deploy of master to Staging (view the deploy)
```
The `(view the deploy)` link does not include the port of the URL

grosser:
it uses `deploy.url` ... which should work ... you can verify the payload it sends works, but otherwise have to reach out to slack to get this working

```
Deploy.last.url
=> "http://localhost:3000/projects/example-kubernetes/deploys/1"
```

PR #3649, opened 2019-11-25
urls with ports are not supported in slack

grosser:
@zendesk/compute @benrudolph

fixes https://github.com/zendesk/samson/issues/3648


benrudolph:
appreciate it, thanks

PR #3651, opened 2019-11-27
link env groups to their edit page

grosser:
Often when I work with env groups, I also want to see how they are configured ... now we can do this without going to index and then searching

Screen Shot 2019-11-26 at 4 57 15 PM
Screen Shot 2019-11-26 at 5 04 55 PM
Screen Shot 2019-11-26 at 5 04 58 PM

@zendesk/compute

### DIFF
--- a/plugins/env/app/views/samson_env/_project_form.html.erb
+++ b/plugins/env/app/views/samson_env/_project_form.html.erb
@@ -43,8 +43,9 @@
           %>
         </div>
         <div class="col-lg-2 checkbox">
-          <%# updated dynammically in the client-side JS %>
-          <%= link_to "Preview", preview_environment_variable_groups_path %>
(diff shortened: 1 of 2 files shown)


plugins/env/app/views/samson_env/_project_form.html.erb
```diff
@@ -43,8 +43,9 @@
           %>
         </div>
         <div class="col-lg-2 checkbox">
-          <%# updated dynammically in the client-side JS %>
-          <%= link_to "Preview", preview_environment_variable_groups_path %>
+          <%# linked to selected group in env/app/assets/javascripts/env/application.js %>
+          <%= link_to "Vars", environment_variable_group_path(0) %> |
```
adammw: "Edit" would make more sense to me

grosser: yeah true :/

PR #3655, opened 2019-11-29
bring schema dumping back

grosser:
... broke that in https://github.com/zendesk/samson/pull/2487 🤦‍♂ 
changing a few old migrations so they produce the schema we have in production ... there seem to be some rails bugs that make the old migrations not produce what they used to do (bigint / datetime etc)

### Risks
 - Low: column types mismatch for old apps that never migrated

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -37,7 +37,7 @@ ActiveRecord::Schema.define(version: 2019_11_13_043345) do
   create_table "builds", id: :integer do |t|
     t.integer "project_id", null: false
     t.integer "number"
-    t.string "git_sha", null: false
+    t.string "git_sha", limit: 128, null: false
     t.string "git_ref", null: false
     t.string "docker_tag"
     t.string "docker_repo_digest"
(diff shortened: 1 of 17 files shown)


yizhang-zen (approved):
Set risk to medium or low?

grosser:
set to low: changing old migrations only influences apps that never ran them ... and these are all 1+ years old

db/schema.rb
```diff
@@ -154,7 +154,7 @@
     t.string "name", null: false
     t.text "comment"
     t.string "owners"
-    t.index ["name"], name: "index_environment_variable_groups_on_name", unique: true
+    t.index ["name"], name: "index_environment_variable_groups_on_name", unique: true, length: 191
```
uthark: where does this arbitrary limit come from?

grosser: bug in 9fc22d6254 which reformatted the schema

grosser: https://github.com/zendesk/samson/pull/3656 to clean things up more

PR #3662, opened 2019-12-10
Prevent mystery @_touch_attr_names nil error caused by threaded execu…

grosser:
AR keeps track of @_touch_attr_names and it looks like we are doing 2 parallel saves which results in the instance variable being unset when doing `job.running!`

/Zendesk/Samson/items/2364
`File "app/models/job.rb" line 148 in status!`

so adding a hacky fix to hopefully get rid of that

@zendesk/compute

### DIFF
--- a/app/jobs/csv_export_job.rb
+++ b/app/jobs/csv_export_job.rb
@@ -11,6 +11,9 @@ class CsvExportJob
     "csv-export-#{@csv_export.id}"
   end
 
+  def reload!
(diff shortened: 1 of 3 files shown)


uthark:
Is there an upstream bug about it?

grosser:
no, we are using it in a bad way, not their fault :(

grosser:
did not work :/

PR #3664, opened 2019-12-10
Work around job save errors by using update_columns

grosser:
https://github.com/zendesk/samson/pull/3662 did not work, so let's try making it not execute the code that causes the issue

@zendesk/compute


grosser:
so this did not work either, but points at where the problem is: touching the deploy

```

activerecord-6.0.1/lib/active_record/no_touching.rb" line 62 in touch
44
app/models/job.rb" line 150 in status!
```

PR #3665, opened 2019-12-10
Work around job save errors by using update_columns take 2

grosser:
https://github.com/zendesk/samson/pull/3664 showed that touching the deploy caused the issue

so either touching anything blows up in which case we have to revert or the deploy had the issue ...
if it is indeed the deploy then we might see strange errors further down the line, but it will either fix the issue or mean we need to load a fresh job/deploy

@zendesk/compute


grosser:
looks like that finally worked :)

PR #3667, opened 2019-12-12
PAAS-5872 create local commands when creating a stage

grosser:
@zendesk/compute


adammw:
explain what, why? how this fixes it?

grosser:
before commands were assigned before the project and then in the command code it build new commands with the current project ... since at that time project was nil it created global commands ... usecase is now covered by the new test I added

PR #3668, opened 2019-12-13
fix reversed code_deployed logic in deploy search

anthonywoo:
On the deploys page, users can search by `code_deployed`. However, the associated key on the Stage is the inverse (`no_code_deployed`). 

This fixes the query to return the correct deploys

### References
- Jira link: 

### Risks
- Low. Fixes deploy search query


grosser:
assets error is random :(

PR #3671, opened 2019-12-18
Make global webhooks more recognizable by adding names

grosser:
@zendesk/eng-productivity @zendesk/compute 

Screen Shot 2019-12-18 at 10 57 41 AM

Screen Shot 2019-12-18 at 12 26 23 PM

Also breaking up the Add UI since it was confusing with 2 buttons
Screen Shot 2019-12-18 at 10 55 49 AM

### DIFF
--- a/app/views/outbound_webhooks/_fields.html.erb
+++ b/app/views/outbound_webhooks/_fields.html.erb
@@ -1,5 +1,6 @@
 <%= render 'shared/errors', object: form.object %>
 
+<%= form.input :name, help: "For global webhooks" %>
 <%= form.input :url, required: true %>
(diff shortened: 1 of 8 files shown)


app/views/outbound_webhooks/_fields.html.erb
```diff
@@ -1,5 +1,6 @@
 <%= render 'shared/errors', object: form.object %>
 
+<%= form.input :name, help: "For global webhooks" %>
```
adammw: only global ones can have names? or can all?

grosser: local can have names, but does not have to ... I guess the db uniqueness will get weird if people start using them for local ones too ... I'll just drop the index 🤷‍♂

PR #3673, opened 2019-12-19
Support JSON CRUD operations for environment variable groups

cakejelly:
Similar to previous contributions, i'd like to be able to perform CRUD operations on environment variable groups via a json api and this PR adds support for this

Probably this could also be cleaned up and simplified to inherit from `ResourceController` but not sure about your stance on sharing code between plugins & the main codebase and also wanted to get buy-in on the change itself first

### References
- Jira link: 

### Risks

### DIFF
--- a/plugins/env/test/controllers/environment_variable_groups_controller_test.rb
+++ b/plugins/env/test/controllers/environment_variable_groups_controller_test.rb
@@ -132,6 +132,26 @@ describe EnvironmentVariableGroupsController do
         assert_response :success
       end
 
+      it "renders json" do
+        get :show, params: {id: env_group.id}, format: :json
+        json = JSON.parse(response.body)
+        group = json['environment_variable_group']
+        group.keys.must_include 'name'
+        group.keys.must_include 'variable_names'
+        assert_response :success
+      end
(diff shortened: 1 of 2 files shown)


plugins/env/test/controllers/environment_variable_groups_controller_test.rb
```diff
@@ -288,6 +332,36 @@ def unauthorized_env_group
 
       it_updates
 
+      it 'renders json' do
```
grosser:
```suggestion
      it 'fails to update with json' do
```

plugins/env/test/controllers/environment_variable_groups_controller_test.rb
```diff
@@ -288,6 +332,36 @@ def unauthorized_env_group
...
+        variable = env_group.environment_variables.first
+        refute_difference "EnvironmentVariable.count" do
+          put :update, params: {
+            id: env_group.id,
+            includes: "environment_variables",
+            environment_variable_group: {
```
grosser: could maybe submit a blank name to trigger the validation more easily

PR #3675, opened 2019-12-23
PAAS-5919 allow removing kuberntes cluster from a deploy group

grosser:
Currently submitting an empty cluster is ignored by the backend

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/decorators/deploy_group_decorator.rb
+++ b/plugins/kubernetes/app/decorators/deploy_group_decorator.rb
@@ -16,8 +16,15 @@ DeployGroup.class_eval do
   accepts_nested_attributes_for(
     :cluster_deploy_group,
     allow_destroy: true,
-    update_only: true,
-    reject_if: ->(h) { h[:kubernetes_cluster_id].blank? }
+    reject_if: ->(h) do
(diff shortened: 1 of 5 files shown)


eatwithforks (approved):

plugins/kubernetes/app/decorators/deploy_group_decorator.rb
```diff
@@ -16,8 +16,15 @@
   accepts_nested_attributes_for(
     :cluster_deploy_group,
     allow_destroy: true,
-    update_only: true,
-    reject_if: ->(h) { h[:kubernetes_cluster_id].blank? }
+    reject_if: ->(h) do
+      empty = h[:kubernetes_cluster_id].blank?
+      if h[:id]
+        h[:_destroy] = true if empty # user removed cluster -> delete connection
```
uthark: Should we delete connections when cluster is deleted?

uthark (approved):

plugins/kubernetes/app/decorators/deploy_group_decorator.rb
grosser: that is done automatically

PR #3676, opened 2019-12-23
PAAS-5913 use server-side apply to avoid weird statefulset bugs

grosser:
@zendesk/compute 

`metadata.annotation.samson/server_side_apply='true'`

should work with all resources, we'll fix bugs as they come up ... later should be the default way samson works

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -10,7 +10,7 @@ module Kubernetes
   module Resource
     module PatchReplace
       def patch_replace?
-        !@delete_resource && exist?
+        !@delete_resource && !server_side_apply? && exist?
       end
 
       # Some kinds can be updated only using PATCH requests.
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -27,15 +27,19 @@ def patch_replace
           next unless value = @template.dig(*keys)
           update.dig_set keys, value
         end
-        with_patch_header do
+        with_header 'application/json-patch+json' do
```
cben: This one is covered by `json_patch` since https://github.com/abonas/kubeclient/pull/390/files

grosser: I need the `with_header` for apply anyway, so just leaving that ... `+json` does not work for apply for some reason, it only supports `+yaml` :/

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -179,6 +182,22 @@ def update
         end
       end
 
+      # TODO: remove name hack https://github.com/abonas/kubeclient/issues/427
+      def server_side_apply(template)
+        with_header 'application/apply-patch+yaml' do # NOTE: we send json but say +yaml since +json gives a 415
```
adammw: lolwut? any upstream fixes?

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -179,6 +182,22 @@ def update
...
+          request(:patch, "#{name}?fieldManager=kubectl&force=true", template, namespace)
```
adammw: field manager samson maybe?

adammw (approved):

plugins/kubernetes/app/models/kubernetes/resource.rb
grosser: we checked kubernetes source code and this is hardcoded 🤷‍♂

PR #3683, opened 2019-12-26
Use resource discovery endpoint to determine namespaced kinds

adammw:
Use resource discovery endpoint to determine namespaced kinds, allowing namespaceless Custom Resources to work correctly (that currently return 404s since they are not known).

### References
https://zendesk.atlassian.net/browse/PAAS-5276

### Risks
- Low: could break existing deployments expecting previous namespace behaviour

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -22,7 +22,7 @@ module Kubernetes
         kind = template[:kind]
 
         set_via_env_json
-        set_namespace if namespaced_kind?(kind)
+        set_namespace
         set_project_labels if template.dig(:metadata, :annotations, :"samson/override_project_label")
         set_deploy_url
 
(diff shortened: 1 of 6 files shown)


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -85,6 +85,20 @@ def self.dig_path(path)
 
     private
 
+    def namespaced_kind?(kind)
+      cluster = @doc.deploy_group.kubernetes_cluster
+      api_version = template.fetch(:apiVersion)
+
+      resources = Rails.cache.fetch(["template-filler-resources", api_version, cluster]) do
```
uthark: how often cache is refreshed?
I.e. how to handle CRD updates?

uthark (approved):

uthark:
🏅

plugins/kubernetes/app/models/kubernetes/template_filler.rb
adammw: apparently once per day per cluster per api version. should be "good enough"

grosser (approved):

grosser:
FYI added another validation that users don't set namespace when kind does not support namespaces

PR #3684, opened 2019-12-27
remove unique label key since it makes deployments on sandbox not pic…

grosser:
…k up new pods and is not needed for other clusters

@zendesk/compute


uthark (approved):

uthark:
💇

PR #3686, opened 2019-12-27
do not fail with 404 errors when requesting an unknown api

grosser:
`[21:15:24] JobExecution failed: 404 Not Found`

@zendesk/compute


adammw (approved):

adammw:
> ehh, little meh on the negative caching but ok
> since if you install a CRD then it will still say it’s wrong until the cache expires

PR #3687, opened 2019-12-30
Fix deploy executor sum error

zechris:
/cc  @zendesk/compute @Tavio @saikambaiyyagari

* Attempted fix for https://rollbar-us.zendesk.com/Zendesk/Samson/items/2276

Which is currently causing Classic deploys to fail in `pod999`:
  https://samson.zende.sk/projects/zendesk/stages/pod-999-kubernetes
Screen Shot 2019-12-30 at 12 13 51

There's probably an underlying cause to those deployment failures, but we're thinking that this exception is hiding them from us.

### References
- Jira link: 

### Risks

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
@@ -960,6 +960,34 @@ describe Kubernetes::DeployExecutor do
     end
   end
 
+  describe "#sum_event_group" do
+    context "when counts are present" do
+      let(:events) do
+        [
+          {count: 1},
+          {count: 2}
+        ]
+      end
+
+      it "sums the counts in the event_group" do
+        executor.send(:sum_event_group, events).must_equal 3
+      end
+    end
+
+    context "when counts are missing" do
+      let(:events) do
+        [
+          {},
+          {}
+        ]
+      end
+
+      it "returns 0" do
+        executor.send(:sum_event_group, events).must_equal 0
+      end
+    end
+  end
+
   describe "#show_logs_on_deploy_if_requested" do
     it "prints errors but continues to not block the deploy" do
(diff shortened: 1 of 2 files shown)


grosser:
makes sense .. events have been known to be weird :/

grosser:
please also add a test that fails without this change ... or ping me if you need help creating it

zechris:
Thanks @grosser.
I've added a regression test now which fails without the fix in this way:
```
Error:
Kubernetes::DeployExecutor::#sum_event_group::when counts are missing#test_0001_returns 0:
NoMethodError: undefined method `+' for nil:NilClass
    plugins/kubernetes/app/models/kubernetes/deploy_executor.rb:224:in `sum_event_group'
    /plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb:986:in `block (4 levels) in <main>'
```

Tavio (approved):

plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
```diff
@@ -960,6 +960,34 @@ def expect_logs(logs = "LOG-1")
     end
   end
 
+  describe "#sum_event_group" do
```
grosser: that's kinda cheating, but whatever :D

grosser (approved):

plugins/kubernetes/test/models/kubernetes/deploy_executor_test.rb
zechris: Haha - yer sorry about that one chief.
I was going to try & figure out how to set up the right conditions for the test of an existing public method, but when I saw the `#show_failure_cause` test is already testing a private method thought I could just add this test in to get things moving again.
(Thanks for the merge though - hopefully we'll be able to see what the *real* problem is now.)

PR #3690, opened 2020-01-02
move files that do not match their classname to make zeitwerk happy

grosser:
@zendesk/compute

### Risks
None


config/logging.rb
```diff
@@ -54,3 +55,14 @@
 
 # workaround for our server ... always gets SIGHUP even though it does not do file logging
 Signal.trap(:SIGHUP) {} if ENV["IGNORE_SIGHUP"]
+
+# We saw initializers hanging boot so we are logging before every initializer is called to easily debug
+# Test: boot up rails and look at the logs
+if ENV['SERVER_MODE'] && !Rails.env.development?
```
grosser: moved code

PR #3694, opened 2020-01-03
allow force update when updating fields is forbidden

grosser:
now that we use RollingUpdate on Statefulset the "not allowed" error popped back up with a new error message ... so adding it to the list of ignores

once saw this error when delete/create was too fast, but it's easy to just press redeploy 🤷‍♂ 

```
[18:37:13]   Warning FailedCreate: create Pod example-server-0 in StatefulSet example-server failed error: The POST operation against Pod could not be completed at this time, please try again. x4
```
(the error will resolve itself since the StatefulSet keeps trying to re-create it's pods)

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/resource_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/resource_test.rb
@@ -779,11 +779,7 @@ describe Kubernetes::Resource do
       end
 
       describe "forced update" do
-        let(:error) { +"Foo is invalid: cannot change spec.bar" }
-
-        before { template[:metadata][:annotations] = {"samson/force_update": "true"} }
-
-        it "re-creates when updating is not possible" do
+        def assert_recreate(error)
           assert_request(:get, url, to_return: [{body: "{}"}, {status: 404}]) do
             assert_request(:put, url, to_return: {body: {message: error}.to_json, status: 422}) do
               assert_request(:delete, url, to_return: {body: '{}'}) do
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -795,6 +791,21 @@ def deployment_stub(replica_count)
           end
         end
 
+        let(:error) { +"Foo is invalid: cannot change spec.bar" }
```
adammw: what's the `+` for?

plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -795,6 +791,21 @@ def deployment_stub(replica_count)
...
+
+        before { template[:metadata][:annotations] = {"samson/force_update": "true"} }
+
+        it "re-creates when updating is not possible" do
```
uthark: this may be dangerous if it is a StatefulSet with PersistentVolume which has `persistentVolumeReclaimPolicy` set to `Delete`.

plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -795,6 +791,21 @@ def deployment_stub(replica_count)
           end
         end
 
+        let(:error) { +"Foo is invalid: cannot change spec.bar" }
+
+        before { template[:metadata][:annotations] = {"samson/force_update": "true"} }
```
adammw: would prefer this in a do..end block for readability

adammw (approved):

plugins/kubernetes/test/models/kubernetes/resource_test.rb
grosser: making it non-frozen

plugins/kubernetes/test/models/kubernetes/resource_test.rb
grosser: up to the user to not use it then ... I'd prefer not hardcoding more rules

uthark (approved):

PR #3695, opened 2020-01-03
filter by pods uid so pods that do not change their name like a state…

grosser:
…fulset do not find old errors

we had a deploy always fail because the previous deploy had error events for the same pod name

we already had `events.select! { |e| last_timestamp(e) >= @start }` but the previous pod kept running for a while because the statefulset took a minute to kill the previous generation of pods

@zendesk/compute


uthark (approved):

adammw:
nice

PR #3697, opened 2020-01-06
Show actual github user in warning message

jvorenkamp:

* Add description
* Add unit tests

### References
- Jira link: 

### Risks


grosser:
that might actually work ... we just need to make a request to github and ask for what the current users name is ... or even simpler configure it via an env var :D

PR #3701, opened 2020-01-07
Fix readonly mode not working on update_columns

grosser:
previously it would say "this is blocked" but not actually block it
@zendesk/compute 

### Risks
 - Low: only active on console

### DIFF
--- a/lib/samson/readonly_db.rb
+++ b/lib/samson/readonly_db.rb
@@ -4,30 +4,46 @@
 # NOTE: not perfect and can be circumvented with multiline sql statements or `;`
 module Samson
   module ReadonlyDb
+    module ReadonlyEnforcer
+      # mysql, sqlite, postgres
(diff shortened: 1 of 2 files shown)


lib/samson/readonly_db.rb
```diff
@@ -9,24 +9,35 @@ module ReadonlyDb
...
-          next if sql.lstrip.start_with?(*ALLOWED)
-          raise ActiveRecord::ReadOnlyRecord, <<~MSG
-            Database is in readonly mode, cannot execute query
-            Switch off readonly with #{self}.#{method(:disable).name}
-            #{sql}
-          MSG
+        return if @enabled
+
+        if @enabled.nil? # only add patch on first enable
+          ActiveRecord::Base.connection.class.prepend(Module.new do
```
adammw: Named module?

PR #3704, opened 2020-01-10
Remove config service integration for application env

sathishavm:
In this PR, Deprecated the config service integration for merging application environment variables during deployment.
Currently, no projects are using the config service integration for app env's.

We will have another PR for config service to fetch only the shared env variable groups.

### Risks
- Low


plugins/env/app/models/environment_variable.rb
```diff
@@ -28,9 +28,10 @@ class << self
     def env(deploy, deploy_group, resolve_secrets:, project_specific: nil)
       env = {}
 
-      if deploy_group && deploy.project.config_service?
-        env.merge! env_vars_from_config_service(deploy, deploy_group)
-      end
+      # Retained the code, to make use for config shared env_group.
+      # if deploy_group
+      #   env.merge! env_vars_from_config_service(deploy, deploy_group)
+      # end
```
grosser: just delete it ... we have git history for that

PR #3705, opened 2020-01-12
prevent writing from console builtin rails helper

grosser:
much cleaner and hopefully less edge-cases :)

found via https://blog.saeloun.com/2019/12/10/rails-block-writes-to-database-connection-while-prevent-writes

@zendesk/compute
/cc @alkesh26


alkesh26:
@grosser thanks.

PR #3706, opened 2020-01-13
cache resources shorter so we do not block projects for too long

grosser:
@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -88,7 +88,7 @@ module Kubernetes
       api_version = template.fetch(:apiVersion)
 
       resources =
-        Rails.cache.fetch(["template-filler-resources", api_version, cluster]) do
+        Rails.cache.fetch(["template-filler-resources", api_version, cluster], expires_in: 1.hour) do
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -88,15 +88,18 @@ def namespaced_kind?(kind)
       api_version = template.fetch(:apiVersion)
 
       resources =
-        Rails.cache.fetch(["template-filler-resources", api_version, cluster]) do
+        Rails.cache.fetch(["template-filler-resources", api_version, cluster], expires_in: 1.hour) do
```
uthark: This is slightly better, but it would be nice to reset cache if deploy contains new CR Definitions. 
Basically, check if there is a resource
```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
```

eatwithforks (approved):

jvorenkamp (approved):

uthark (approved):

grosser:
gaps:
- does not expire cache after new CRD was created
- does not work when CRD is created indirectly

... so ideally watch CRDs for each cluster and update cache, but that's more work

PR #3708, opened 2020-01-17
Support external service for env groups 

sathishavm:
In this story, Users can add s3 URL for **ENV Groups** on their project settings page. The s3 file consists of list of environment variables grouped by deploy_group.

The S3 files should be managed by some external service, the external service is responsible for syncing files to s3 and provide relevant s3 file URL.

Currently, It is not mandated for using only external service env groups, adding s3 files allows the system to read values and merge ENV's along with ENV stored in the database.

Screen Shot 2020-01-16 at 9 10 57 PM

We have preview options to view the values from the s3 URL, once it is validated and added to the system.

### s3 URL format  
Currently, AWS widely accepts various s3 URL format eg: s3://bucket/key, http://s3.amazonaws.com/bucket, http://s3-aws-region.amazonaws.com/bucket ..etc

In this release, we have validated only to use **https://[bucket].s3.amazonaws.com/[key]?versionId=[version_id]** which is the most common format and available in AWS s3 page for all s3 object, it allows users to just copy and use it here.

### File format
In this release, we support only JSON and YAML file format for the external env groups as it is widely known formats.

### File schema
 As mentioned above the file should have list of ENV variables(key: value) and grouped by `Deploy group Permalink`

Eg:
```
pod1:
  name: pod1
  env: development
pod2:
  name: pod2
  env: test  
```

### Risks
- Med


plugins/env/app/models/external_environment_variable_group.rb
```diff
@@ -0,0 +1,70 @@
...
+  default_scope -> { order(:name) }
+
+  belongs_to :project, inverse_of: :external_environment_variable_groups
+  validates :name, :url, presence: true
+  validates :url, format: {
+    with: S3_URL_REGEX,
+    message: "Invalid format, valid url format is #{S3_URL_FORMAT}"
+  }
+  validate :validate_s3_url
+  after_find :resolve_s3_url
```
grosser: prefer defining methods for each component and making them call `resolve_s3_url`, always doing certrain work ater finding is usually a bad pattern

plugins/env/app/models/external_environment_variable_group.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def external_service_read_with_failover
+    default_bucket = ENV.fetch 'EXTERNAL_ENV_GROUP_S3_BUCKET'
+    default_region = ENV.fetch 'EXTERNAL_ENV_GROUP_S3_REGION'
+    dr_bucket      = ENV['EXTERNAL_ENV_GROUP_S3_DR_BUCKET']
+    dr_region      = ENV['EXTERNAL_ENV_GROUP_S3_DR_REGION']
+    if default_bucket != bucket && dr_bucket != bucket
+      buckets = [default_bucket, dr_bucket].compact
+      raise "Invalid s3 bucket, acceptable buckets are #{buckets.join(',')}"
```
grosser: if we know the supported bucket, then this should happen during validation or users should only set the path

plugins/env/app/models/external_environment_variable_group.rb
```diff
@@ -0,0 +1,70 @@
...
+          s3_client = Aws::S3::Client.new(region: default_region)
+          s3_client.get_object(bucket: default_bucket, key: key, version_id: version_id)
+        rescue Aws::S3::Errors::NoSuchKey
+          raise "key \"#{key}\" does not exist in bucket #{bucket}!"
+        rescue Aws::S3::Errors::ServiceError => e
+          raise e if !dr_bucket || !dr_region
+          s3_client = Aws::S3::Client.new(region: dr_region)
+          s3_client.get_object(bucket: dr_bucket, key: key, version_id: version_id)
+        end
+      YAML.safe_load response.body.read
```
grosser: docs says it supports json ... fix the docs or add json ?
... also validate file extension is `.yml` / `.yaml`

plugins/env/app/models/external_environment_variable_group.rb
```diff
@@ -0,0 +1,70 @@
...
+  belongs_to :project, inverse_of: :external_environment_variable_groups
+  validates :name, :url, presence: true
+  validates :url, format: {
+    with: S3_URL_REGEX,
+    message: "Invalid format, valid url format is #{S3_URL_FORMAT}"
+  }
+  validate :validate_s3_url
+  after_find :resolve_s3_url
+
+  def external_service_read_with_failover
```
grosser: pretty long name . how about `read` ? :D

plugins/env/app/views/samson_env/_project_form.html.erb
```diff
@@ -20,6 +20,12 @@
   <% sorted_env_vars << EnvironmentVariable.new %>
   <%= render "samson_env/environment_variables", form: form, environment_variables: sorted_env_vars %>
 
+  <h4>
```
grosser: only show that section if the bucket or help-text is set, most users don't want this feature I'd guess

plugins/env/decorators/project_decorator.rb
```diff
@@ -2,8 +2,11 @@
 Project.class_eval do
   include AcceptsEnvironmentVariables
 
+  has_many :external_environment_variable_groups, dependent: :destroy
   has_many :project_environment_variable_groups, dependent: :destroy
   has_many :environment_variable_groups, through: :project_environment_variable_groups, inverse_of: :projects
+  accepts_nested_attributes_for :external_environment_variable_groups, allow_destroy: true,
+    reject_if: ->(a) { a[:name].blank? && a[:url].blank? }
```
grosser: we want to show `needs url` validation, so accept any partially filled form ?
```suggestion
    reject_if: ->(a) { a[:name].blank? || a[:url].blank? }
```

plugins/env/test/controllers/external_environment_variable_groups_controller_test.rb
```diff
@@ -0,0 +1,46 @@
...
+describe ExternalEnvironmentVariableGroupsController do
+  before do
+    ExternalEnvironmentVariableGroup.any_instance.
+      expects(:external_service_read_with_failover).times(2).returns("a" => "b")
+  end
+
+  let(:project) { projects(:test) }
+  let(:group) do
+    ExternalEnvironmentVariableGroup.create!(
+        name: "A",
```
grosser: 2-space indent plz

grosser:
looks pretty good, just a few nitpicks :)

plugins/env/app/models/external_environment_variable_group.rb
sathishavm: Yes, this is validated during create and update

plugins/env/app/models/external_environment_variable_group.rb
sathishavm: YAML.safe_load can parse json, yml and yaml formats, verified.

grosser: crazyyyy :D

grosser: .. maybe add a comment for it though, just in case since this looks like a bug can point to https://stackoverflow.com/questions/24608600/is-it-safe-to-parse-json-with-yaml-load

plugins/env/app/models/environment_variable.rb
```diff
@@ -74,6 +77,15 @@ def env_vars_from_repo(env_repo_name, project, deploy_group)
       raise Samson::Hooks::UserError, "Cannot download env file #{path} from #{env_repo_name} (#{e.message})"
     end
 
+    def env_vars_from_external_groups(project, deploy_group)
+      project.external_environment_variable_groups.each_with_object({}) do |group, envs|
+        group_env = group.read[deploy_group.permalink]
+        envs.merge! group_env if group_env
+      end
+    rescue StandardError => e
+      raise Samson::Hooks::UserError, "Error reading env vars from external env-groups: #{e.message}"
```
grosser: this is great!

plugins/env/app/views/external_environment_variable_groups/_fields.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <%= fields.text_field :name, class: "form-control monospace", placeholder: "Name" %>
+    </div>
+
+    <div class="col-lg-3">
+      <%# using a text area so users can resize them with browser controls %>
+      <%= fields.text_area :description, class: "form-control monospace validate_whitespace", rows: 1, placeholder: "Description" %>
+    </div>
+
+    <div class="col-lg-5">
+      <%= fields.text_field :url, class: "form-control monospace", placeholder: ExternalEnvironmentVariableGroup::S3_URL_FORMAT %>
```
grosser: FYI there is in-browser regex matching too, I thin `def input` in form-builder supports that

plugins/env/app/views/external_environment_variable_groups/preview.html.erb
```diff
@@ -0,0 +1,23 @@
...
+<section class="clearfix">
+  <dl class="dl-horizontal">
+    <dt>Name:</dt>
+    <dd><%= @group.name %></dd>
+    <dt>Description:</dt>
+    <dd><%= @group.description %></dd>
+    <dt>URL:</dt>
+    <dd><%= @group.url %></dd>
+  </dl>
+    <pre>
```
grosser: out-dent

grosser (approved):
👍

PR #3709, opened 2020-01-21
only rescue vault read error if vault is not configured

anthonywoo:
Only rescue specific errors when `read_multi` fails. This is to account for when the request is not authorized or there is some other issue with vault (issue related to an incident)

### References
- Jira link: 

### Risks
- Medium: could cause issues if something expects `read_multi` to return nil instead of an error.

### DIFF
--- a/test/lib/samson/secrets/hashicorp_vault_backend_test.rb
+++ b/test/lib/samson/secrets/hashicorp_vault_backend_test.rb
@@ -144,10 +144,6 @@ describe Samson::Secrets::HashicorpVaultBackend do
     it "leaves out vaules from deploy groups that have no vault server so KeyResolver works" do
       backend.read_multi(['production/foo/pod100/bar']).must_equal({})
     end
-
-    it "leaves out vaules from unknown deploy groups" do
-      backend.read_multi(['production/foo/pod1nope/bar']).must_equal({})
-    end
   end
 
   describe ".delete" do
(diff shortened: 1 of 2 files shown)


test/lib/samson/secrets/hashicorp_vault_backend_test.rb
```diff
@@ -144,10 +144,6 @@
     it "leaves out vaules from deploy groups that have no vault server so KeyResolver works" do
       backend.read_multi(['production/foo/pod100/bar']).must_equal({})
     end
-
-    it "leaves out vaules from unknown deploy groups" do
-      backend.read_multi(['production/foo/pod1nope/bar']).must_equal({})
-    end
```
grosser: ... we still need a testcase for this ?

grosser:
change looks good, just needs a test so we don't break that by accident

test/lib/samson/secrets/hashicorp_vault_backend_test.rb
anthonywoo: oops, I thought this was testing the same functionality as https://github.com/zendesk/samson/pull/3709/files/430a0f5d71795fde0bcca0ad610ecd4ce3a3c6c7#diff-2f2df8cde92ee51d143c321a9d28f391R145 as I didn't realize a pod100 deploy group existed for the test 😅 

I've added the test back

PR #3711, opened 2020-01-25
shorten generated labels so we do not run into kubernetes validation …

grosser:
…errors

```
[23:36:11] JobExecution failed: Kubernetes error Deployment.apps "foo" is invalid: spec.template.labels: Invalid value: "foobarbaz-edata-12881-split-drdb-tasks-out-into-dedicated-collector-instance": must be no more than 63 characters
```

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -396,7 +396,7 @@ module Kubernetes
     # Adding the Release ID to allow us to track the progress of a new release from the UI.
     def set_spec_template_metadata
       release_doc_metadata.each do |key, value|
-        pod_template.dig_fetch(:metadata, :labels)[key] ||= value.to_s.parameterize.tr('_', '-')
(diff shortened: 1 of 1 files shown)


uthark (approved):

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -396,7 +396,7 @@ def set_hpa_scale_target_name
     # Adding the Release ID to allow us to track the progress of a new release from the UI.
     def set_spec_template_metadata
       release_doc_metadata.each do |key, value|
-        pod_template.dig_fetch(:metadata, :labels)[key] ||= value.to_s.parameterize.tr('_', '-')
+        pod_template.dig_fetch(:metadata, :labels)[key] ||= value.to_s.parameterize.tr('_', '-').slice(0, 63)
```
uthark: I’d cut last 63 symbols, not first.

grosser: I meant to do a truncation, so I kinda wanted the first chars ... also the first will include the persons name, which I think is meaningful here

PR #3713, opened 2020-02-05
bump omniauth-github to stop sending access-token via query params

grosser:
github is sending emails about it "Please use the Authorization HTTP header instead as using the `access_token` query parameter is deprecated.
"

confirmed working locally by printing in config/initializers/faraday.rb

@zendesk/compute


patrobinson:
I don't think this works, because omniauth-github has not released a new gem version for this change. The current maintainers are waiting for someone to give them access https://github.com/omniauth/omniauth-github/issues/87 so we may have to use the github source.

grosser:
there might be more bugs in there, but I can see it clearly not working in 925b68766 and being better after the update 🤷‍♂ 
PR welcome if you can find another area where it uses the query

patrobinson:
The code to fix this was only committed yesterday https://github.com/omniauth/omniauth-github/pull/84

Github only send you an email every 3 days it says so maybe it'll come through then?

grosser:
yeah, could be that it just changed something and now it no longer logs ...
we'll see when it comes back,
if it  does then it's time to use it github :(

grosser:
yeah, when I add logging deeper down it shows that it still uses params ... time for git dependency ...

PR #3717, opened 2020-02-07
fix random test error

grosser:
if this test ran first it would not provide coverage or the "disabled" case ... and since reproducing the uninitialized case is not reliable or necessary, I'm just changing it to a reliable disabled

```
test/lib/samson/readonly_db_test.rb 
lib/samson/readonly_db.rb new uncovered lines introduced (1 current vs 0 configured)
Lines missing coverage:
lib/samson/readonly_db.rb:11:9-106
Run options: --seed 61123
```

@zendesk/compute

### DIFF
--- a/test/lib/samson/readonly_db_test.rb
+++ b/test/lib/samson/readonly_db_test.rb
@@ -21,7 +21,9 @@ describe Samson::ReadonlyDb do
       end
     end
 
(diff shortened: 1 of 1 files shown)


grosser:
I can explain if you care, but it's complicated :D

PR #3720, opened 2020-02-10
namespaces: tell users the template needs to be set when left empty

grosser:
previously said "it needs to be a hash" which was confusing
@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/namespace.rb
+++ b/plugins/kubernetes/app/models/kubernetes/namespace.rb
@@ -51,10 +51,9 @@ module Kubernetes
     end
 
     def validate_template
-      unless parsed_template.is_a?(Hash)
-        return errors.add :template, "needs to be a Hash"
-      end
-      errors.add :template, "needs metadata.labels.team" unless parsed_template.dig("metadata", "labels", "team")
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/namespace.rb
```diff
@@ -51,10 +51,9 @@ def parsed_template
     end
 
     def validate_template
-      unless parsed_template.is_a?(Hash)
-        return errors.add :template, "needs to be a Hash"
-      end
-      errors.add :template, "needs metadata.labels.team" unless parsed_template.dig("metadata", "labels", "team")
+      return errors.add :template, "needs to be set" if template.blank?
+      return errors.add :template, "needs to be a Hash" unless parsed_template.is_a?(Hash)
+      return errors.add :template, "needs metadata.labels.team" unless parsed_template.dig("metadata", "labels", "team")
```
grosser: notice the symmetry :trollface:

hgokavarapuz (approved):

plugins/kubernetes/app/models/kubernetes/namespace.rb
uthark: I'd prefer UI to be populated with valid empty hash.

PR #3722, opened 2020-02-12
Scan the repository URL correctly for gitlab projects.

farvour:
* Allows gitlab project URLs to be scanned properly. Previously the ref checker for GitLab didn't function with projects in subgroups under a GitLab organization. @grosser 's change to make the ref check more strict made this a blocker. In the past one could just bypass the Review button. It also fixes the URL generated with the GitLab cat icon in the project details page.

### Risks
- Low, impacts only gitlab projects.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -130,7 +130,7 @@ class Project < ActiveRecord::Base
 
   # The user/repo part of the repository URL.
   def repository_path
-    if self.gitlab?
+    if gitlab?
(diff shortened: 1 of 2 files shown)


grosser:
funky bugs ... thx for the good comments :)

farvour:
Thanks! I bet not many ran into this weird one since top-level org projects in GitLab don't exhibit the issue, but we have many.

Glad to help.

PR #3724, opened 2020-02-14
send correct canceller in slack notifications

grosser:
what I received when cancelling my own deploy:
```
Samson is cancelling Michael Grosser's deploy
```

when I reproduce via console:
```
s = SlackWebhookNotification.new(d, [])
s.send :deploy_callback_content
Michael Grosser cancelled Michael Grosser's deploy
```

... so there is a race condition :)

@zendesk/compute

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -134,7 +134,7 @@ class Project < ActiveRecord::Base
       # This is GitLab, which allows similar characters in repository names, but
       # also follows a subgroup convention which can contain an n-number of paths
       # underneath the organization group.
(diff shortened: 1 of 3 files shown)


app/models/project.rb
```diff
@@ -134,7 +134,7 @@ def repository_path
       # This is GitLab, which allows similar characters in repository names, but
       # also follows a subgroup convention which can contain an n-number of paths
       # underneath the organization group.
-      repository_url.scan(%r{[:/]([A-Za-z0-9-/]+/[\w.-]+?)(?:\.git)?$}).join
+      repository_url.scan(%r{[:/]([A-Za-z0-9\-/]+/[\w.-]+?)(?:\.git)?$}).join
```
grosser: random warning fix

PR #3735, opened 2020-03-02
Cleanup orphaned webhooks

patrobinson:
As part of #3608 `OutboundWebhook.find_each` was used to migrate existing webhooks, however deleted webhooks were not copied because they are not returned by that command.

This left some dangling `OutboundWebhookStage` objects, which meant when browsing to the Webhook tab on a Project which had previously deleted webhooks you'd get an exception.

This PR cleans up all `OutboundWebhookStage` objects who's OutboundWebhook has been incidentally dropped.

### DIFF
--- a/db/migrate/20200219043627_delete_orphaned_outbound_webhooks.rb
+++ b/db/migrate/20200219043627_delete_orphaned_outbound_webhooks.rb
@@ -1,0 +1,14 @@
+class DeleteOrphanedOutboundWebhooks < ActiveRecord::Migration[6.0]
+  class OutboundWebhook < ActiveRecord::Base
+  end
+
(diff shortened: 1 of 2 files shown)


grosser:
looks good!
... run the migration and commit the updated schema (number will change)

patrobinson:
@grosser I ran it but there was a bunch of irrelevant changes. Is this the only change I need to commit?
```
-ActiveRecord::Schema.define(version: 2020_01_10_213328) do
+ActiveRecord::Schema.define(version: 2020_02_19_043627) do
```

(Sorry not a rails developer)

grosser:
yep!

ideally there should not be any other diff when running on mysql, but I know it can be messy :)

PR #3738, opened 2020-03-04
Return error output when git commands fail

patrobinson:
While investigating an issue where git clones failed I found no command output was being logged.

There's two different classes involved in running commands via the `ensure_tag_in_git_repository` code path where the issue occurs.
Therefore I've added logging of command results to both (CommandExecutor and TerminalExecutor)

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -190,6 +190,9 @@ class GitRepository
       err: '/dev/null',
       dir: dir
     )
+    unless result.status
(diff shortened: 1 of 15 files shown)


grosser:
I like the capturing of err and out!
not a fan of openstruct though, prefer a Hash ... even better with a `success, out, err` triplet 🤷‍♂

patrobinson:
Yeah I've been writing Golang too long. Guess a Hash achieves the same thing

PR #3739, opened 2020-03-04
Bump timeout for git commands

patrobinson:
Even when using a shallow clone, we commonly hit the 5s timeout.
This is rather intermittent, but on a fresh instance with no cache it still occurs 1-2 times a day for us.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -176,7 +176,7 @@ class Project < ActiveRecord::Base
     @repository ||= GitRepository.new(
       repository_url: repository_url,
       repository_dir: repository_directory,
-      executor: TerminalExecutor.new(StringIO.new, project: self)
+      executor: TerminalExecutor.new(StringIO.new, project: self, timeout: 30)
     )
   end
 


grosser:
where do you see these errors happening ?
- when doing a deploy, the executor has a long timeout
- all other actions are not really supposed to do a full checkout since that would be longer then the request timeout anyway
... I guess ideally we'd only allow this kind of checkout for local repositories and use the api when using github/gitlab :(

patrobinson:
@grosser it's during an automatic deploy triggered via webhook. The `ReleaseService` executes `ensure_tag_in_git_repository` which does a clone, which executes
```
    executor.execute "git -c core.askpass=true clone --mirror #{repository_url} #{repo_cache_dir}"
```

https://github.com/zendesk/samson/blob/master/app/models/git_repository.rb#L134

patrobinson:
Actually for our largest repository (2.5GB mirrored) takes ~180s to clone

grosser:
better to either:
 - change the webhook to fail when repo was never cloned
 - avoid this call and use api instead to find the tag

patrobinson:
I think using the API is a much better idea. Will put that into my backlog of stuff to do

PR #3742, opened 2020-03-05
Disable serviceLinks

uthark:
To not pollute env of the containers, disable serviceLinks by default.
If the deployment explicitly has it, keep as is.

Some back story:
- [Docker legacy `--links`](https://docs.docker.com/network/links/) - what causes this.
- [K8 feature PR](https://github.com/kubernetes/kubernetes/pull/68754)
- [K8 issue raised](https://github.com/kubernetes/kubernetes/issues/60099)

### Risks
- Med - changing default behavior might break existing deployments that rely on it.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -38,6 +38,7 @@ module Kubernetes
         elsif Kubernetes::RoleConfigFile.primary?(template)
           if kind == 'Deployment'
             set_history_limit
+            set_disable_service_links
           end
 
           make_stateful_set_match_service if kind == 'StatefulSet'
@@ -55,6 +56,7 @@ module Kubernetes
           set_init_containers
           set_kritis_breakglass
           set_istio_sidecar_injection
+
         elsif kind == 'PodDisruptionBudget'
           set_name
           set_match_labels_blue_green if blue_green_color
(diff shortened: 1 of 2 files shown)


magnetikonline (approved):

magnetikonline:
I approve @uthark - but just wondering _could_ this break someone's current workflow/deployment I wonder?

If it does, I would strongly advise the owners of the project in the question to avoid using these env vars - but still, might be a breaking change?

grosser:
yeah, please update risks ... sounds like we have to first find out who uses this to disable or only turn on in staging somehow

uthark:
It might break, so we need to announce the change and ask people to explicitly set the flag if they rely on this behavior.

magnetikonline:
Turning on in staging sounds like a good idea no? It should shake out any deployments that fail - of course you need to deploy them first to find out! The vars do have a fairly predictable pattern - could grep GitHub?

```
DB_NAME=/web2/db
DB_PORT=tcp://172.17.0.5:5432
DB_PORT_5432_TCP=tcp://172.17.0.5:5432
DB_PORT_5432_TCP_PROTO=tcp
DB_PORT_5432_TCP_PORT=5432
DB_PORT_5432_TCP_ADDR=172.17.0.5
```

grosser:
we could do some opa rule that says "values must be set to either
true/false", but that would fail almost all deploys ... so maybe do that
after this rollout
for now only setting it on non-production sounds good, I'm still expecting
fallout since we at least use the kubernetes ports somewhere ...

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -267,6 +269,11 @@ def set_history_limit
       template[:spec][:revisionHistoryLimit] ||= 1
     end
 
+    # disable service links.
+    def set_disable_service_links
+      template[:spec][:template][:spec][:enableServiceLinks] ||= false
```
grosser: we should set that on pod_templates for consistency, not only Deployments, use `pod_template`

PR #3747, opened 2020-03-13
Use api check releases

patrobinson:
This is the alternative as suggested in #3739

Instead of trying to do a `git clone` while processing a web request, we instead use the Github API to fetch releases and validate the release we've created.

This unmasks a different bug though, in that the code to checkout a release before a deployment is run doesn't update the mirror first (it calls `ensure_mirror_updated` but that doesn't do as it says on the tin).

We have to call `update_mirror` explicitly, otherwise deployments fail because the mirror is not fresh and therefore doesn't contain the release we're about to deploy.

### DIFF
--- a/app/models/job_execution.rb
+++ b/app/models/job_execution.rb
@@ -168,6 +168,7 @@ class JobExecution
   end
 
   def resolve_ref_to_commit
+    @repository.update_mirror
     commit = @repository.commit_from_ref(@reference)
     tag = @repository.fuzzy_tag_from_ref(@reference)
     if commit
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -97,6 +97,7 @@ class Release < ActiveRecord::Base
     tag = GITHUB.compare(project.repository_path, commit, commit)
     self.commit = tag[:base_commit][:sha]
   rescue Octokit::NotFound
+    ::Rails.logger.info("Couldn't convert #{commit} to tag")
   end
 end
 Samson::Hooks.load_decorators(Release)
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -94,7 +94,9 @@ class Release < ActiveRecord::Base
 
   def convert_ref_to_sha
     return if commit.blank? || commit =~ Build::SHA1_REGEX
-    self.commit = project.repository.commit_from_ref(commit)
+    tag = GITHUB.compare(project.repository_path, commit, commit)
+    self.commit = tag[:base_commit][:sha]
+  rescue Octokit::NotFound
   end
 end
 Samson::Hooks.load_decorators(Release)
--- a/app/models/release.rb
+++ b/app/models/release.rb
@@ -7,7 +7,7 @@ class Release < ActiveRecord::Base
   belongs_to :author, class_name: "User", inverse_of: false
 
   before_validation :assign_release_number
-  before_validation :covert_ref_to_sha
+  before_validation :convert_ref_to_sha
 
   validates :number, format: {with: NUMBER_REGEX, message: "may only contain numbers and decimals."}
   validates :commit, format: {with: Build::SHA1_REGEX, message: "can only be a full sha"}, on: :create
@@ -92,7 +92,7 @@ class Release < ActiveRecord::Base
     current_version.to_s.dup.sub!(/\d+$/) { |d| d.to_i + 1 }
   end
 
-  def covert_ref_to_sha
+  def convert_ref_to_sha
     return if commit.blank? || commit =~ Build::SHA1_REGEX
     self.commit = project.repository.commit_from_ref(commit)
   end
--- a/app/models/release_service.rb
+++ b/app/models/release_service.rb
@@ -27,11 +27,9 @@ class ReleaseService
   end
 
   def ensure_tag_in_git_repository(tag)
-    tries = Integer(ENV["RELEASE_TAG_IN_REPO_RETRIES"] || 4)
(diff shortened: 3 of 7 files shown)


app/models/release.rb
```diff
@@ -92,9 +92,12 @@ def next_number(current_version)
     current_version.to_s.dup.sub!(/\d+$/) { |d| d.to_i + 1 }
   end
 
-  def covert_ref_to_sha
+  def convert_ref_to_sha
     return if commit.blank? || commit =~ Build::SHA1_REGEX
-    self.commit = project.repository.commit_from_ref(commit)
+    tag = GITHUB.compare(project.repository_path, commit, commit)
```
grosser: please use `Project#repo_commit_from_ref`

app/models/job_execution.rb
```diff
@@ -168,6 +168,7 @@ def setup(dir)
   end
 
   def resolve_ref_to_commit
+    @repository.update_mirror
```
grosser: hmm this should happen somewhere in the setup ? ... kinda surprising if it happens here ?

app/models/release.rb
patrobinson: The comment from that function
>   # quick commit from reference without pulling the repository ... unless we have no remote

The issue we're trying to resolve is, when there's no cached repository then the webhook request process will take a long time (and timeout).
AFAICT this is still part of that request, as we create the Release as part of that.

So I don't think using that function will solve this.

app/models/job_execution.rb
patrobinson: As explained in PR comments, this calls `commit_from_ref` which calls `ensure_mirror_current` which doesn't do what it says it does.

grosser:
does https://github.com/zendesk/samson/pull/3748 help ?
(even if it does not completely solve the issue)

patrobinson:
@grosser I don't even know if that causes a failure, I'm guessing it does because a Release must be created as part of the webhook process.

The issue we have observed is in `ReleaseService#ensure_tag_in_git_repository`

grosser:
amended the PR to cover that case too looks good ?

patrobinson:
@grosser I don't think your PR will work, because it only fetches a commit _if the remote already exists_.

The problem is we have 60 or so projects, maybe 2/3 are active. Anytime we deploy a new Samson instance it starts with an empty cache (no remotes), which will result in an attempted clone the first time a deployment webhook arrives. A good portion of these will fail, some will never succeed and have to be manually deployed, which populates the cache.

This is why I've moved to using the API, as you suggested in #3739, because it avoids performing the first clone while processing a webhook request. Instead we perform an initial clone the first time we deploy to a fresh instance, which is fine to take ~80 seconds.

grosser:
repo_commit_from_ref goes directly to github api ... why would that trigger a clone ?

grosser:
... also if you have a chance of persisting the checkout cache, that would help a lot :)

patrobinson:
@grosser From reading it's code that wasn't clear, I'll give this a whirl in our environment.

Persisting the cache for all 60 repos isn't an option at this stage.

grosser:
any results yet / is this a good step forward ?

patrobinson:
@grosser I deployed this change and have seen one deploy complete successfully. I will let it soak for another 24 hours though.

patrobinson:
@grosser I deployed #3748 and we're seeing the issue i fixed in 63d258a5d1d257896bb93eb3e8832c5e308ce993

When a webhook arrives the release is created correctly, but when the deployment starts it fails because the mirror is not up to date.

```
[21:09:36] Could not find commit for v11799
```

grosser:
so that means we need an extra pull when the deploy starts, the rest seems
to be working ?

grosser:
... commit_from_ref calls `ensure_mirror_current` which calls `update_mirror` so idk how this would be any better :(

patrobinson:
@grosser `ensure_mirror_current` only calls `update_mirror` *if* `mirror_current?` returns false.
This checks if `@mirror_current` is nil. `@mirror_current` is only updated by `ensure_mirror_current` so it will set this variable the first time it runs, subsequent runs will always return without updating the mirror, despite the fact it's likely not up to date.

grosser:
`@mirror_current` is cached per request ... why would it be needed twice in the same "deploy" request ?

patrobinson:
@grosser GitRepository is memoized [here](https://github.com/zendesk/samson/blob/09af0724d9a33a0bfecdae91683b38616bf9e8a7/app/models/project.rb#L176) which means it's persistent across requests. On the first deployment it'll work fine, because `@mirror_current` will be nil, but subsequent deployments will get a stale mirror.

grosser:
Project instances are not shared across requests, they are fetched fresh from the db on each request

patrobinson:
Ok I have no idea what's going on then. WIthout 63d258a we see the behaviour described above

grosser:
could be a race condition from the release controller creating the tag and
it not yet being ready when the deploy starts ...
we could make the release controller wait a bit until the tag is "findable"

patrobinson:
If it was a race condition it should still be present even with the fix, which forces an update to the mirror?

patrobinson:
So in short:
- I'm happy with #3748
- Unfortunately it unmasks another bug though

I'd love to get _some_ progress into getting these changes upstream, I fear otherwise we'l but stuck on a forked version for some time.

grosser:
can you give https://github.com/zendesk/samson/pull/3758 a spin ?

patrobinson:
@grosser It'll be next week

PR #3749, opened 2020-03-20
PAAS-6412 allow 1 deployment for each version by appending tag to nam…

grosser:
…e and making tag+name available for replacement

hacky, so keeping it all in 1 place (instead of removing keep_name requirement with more logic)

@zendesk/compute 
@fsat

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -119,6 +119,13 @@ module Kubernetes
         data.merge!(YAML.safe_load(yaml))
       end
 
+      env = static_env
+      if template.dig(:metadata, :annotations, :"samson/append_tag_to_name") == "true"
+        tag = env.fetch(:TAG)
+        env["TAG_AS_JSON"] = tag.to_json
+        env["NAME_AS_JSON"] = "#{template.dig_fetch(:metadata, :name)}-#{tag}".to_json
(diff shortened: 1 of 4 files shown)


uthark (approved):
👍 on 🍏

fsat (approved):
@grosser - thank you for doing this for us 👍

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -119,6 +119,13 @@ def set_via_env_json
         data.merge!(YAML.safe_load(yaml))
       end
 
+      env = static_env
+      if template.dig(:metadata, :annotations, :"samson/append_tag_to_name") == "true"
+        tag = env.fetch(:TAG)
+        env["TAG_AS_JSON"] = tag.to_json
+        env["NAME_AS_JSON"] = "#{template.dig_fetch(:metadata, :name)}-#{tag}".to_json
```
adammw: couldn't they configure this in their samson config since it has some interpolation, rather than having to build a whole feature for it?
e.g. set `NAME_AS_JSON=my-project-$TAG` in the config, along with the set-from-env annotations?

grosser:
yeah good idea :D
... I tried that first, but TAG is not part of the interpolation of env
vars, I'll try again and see how much hackery that is in comparison, I'm
slightly scared that will break other logic

PR #3751, opened 2020-03-21
stream output even when command is cancelled so we see what it does i…

grosser:
…f it traps/prints anything as it dies

previously we interrupted streaming with the `cancel`, now we let it continue until the process is dead

@zendesk/compute 

```
[02:01:07] » ruby -e 'begin; sleep 100; rescue Exception; puts "GOT THAT #{$!.class}";end'
[02:01:08] GOT THAT Interrupt
```

### DIFF
--- a/app/models/terminal_executor.rb
+++ b/app/models/terminal_executor.rb
@@ -41,23 +41,24 @@ class TerminalExecutor
       script_as_executable(script(commands)) do |command|
         output, _, pid = PTY.spawn(whitelisted_env, command, in: '/dev/null', unsetenv_others: true)
         record_pid(pid) do
-          begin
-            Timeout.timeout(timeout) do
-              stream from: output, to: @output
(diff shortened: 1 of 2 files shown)


app/models/terminal_executor.rb
```diff
@@ -41,23 +41,24 @@ def execute(*commands, timeout: @timeout)
       script_as_executable(script(commands)) do |command|
         output, _, pid = PTY.spawn(whitelisted_env, command, in: '/dev/null', unsetenv_others: true)
         record_pid(pid) do
-          begin
-            Timeout.timeout(timeout) do
-              stream from: output, to: @output
```
adammw: How did this work, ie. why was cancel affecting this? wouldn't this keep going until the stream was closed?

grosser: `cancel` raises and the `raise` interrupted the "read from the stream" code ... so output still came in, but was not read

PR #3752, opened 2020-03-23
PAAS-6412 make TAG and other dynamic env vars availble for interpolation

grosser:
replaces https://github.com/zendesk/samson/pull/3749

@zendesk/compute @fsat

### DIFF
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -24,8 +24,8 @@ class EnvironmentVariable < ActiveRecord::Base
     # preview parameter can be used to not raise an error,
     # but return a value with a helpful message
     # also used by an external plugin
-    def env(deploy, deploy_group, resolve_secrets:, project_specific: nil)
-      env = {}
+    def env(deploy, deploy_group, resolve_secrets:, project_specific: nil, base: {})
+      env = base.dup
 
       if deploy_group
         env.merge! env_vars_from_external_groups(deploy.project, deploy_group)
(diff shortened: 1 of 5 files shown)


plugins/env/app/models/environment_variable.rb
```diff
@@ -24,9 +24,7 @@ class << self
     # preview parameter can be used to not raise an error,
     # but return a value with a helpful message
     # also used by an external plugin
-    def env(deploy, deploy_group, resolve_secrets:, project_specific: nil)
-      env = {}
-
+    def env(deploy, deploy_group, resolve_secrets:, project_specific: nil, env: {})
```
adammw: maybe `env` should be `defaults` or something? sounds a little odd (also the fact that it's passed in, rather than just being merged in the caller)

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -1050,6 +1050,21 @@ def secret_annotations(hash)
...
+        raw_template[:metadata][:annotations] = {
+          "samson/keep_name": "true",
+          "samson/set_via_env_json-metadata.name": "NAME_AS_JSON",
+          "samson/set_via_env_json-spec.matches_service": "NAME_AS_JSON",
+          "samson/set_via_env_json-spec.template.metadata.labels.tag": "TAG_AS_JSON"
+        }
+        t = template.to_hash
+        t.dig(:metadata, :name).must_equal "foo-master"
+        t.dig(:spec, :matches_service).must_equal "foo-master"
+        t.dig(:spec, :template, :metadata, :labels, :tag).must_equal "master"
```
adammw: also test environment variable set?

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -1050,6 +1050,23 @@ def secret_annotations(hash)
...
+        t.dig(:spec, :template, :spec, :containers, 0, :env).
+          detect { |h| h[:name] == "TAG" }[:value].must_equal "master"
```
adammw: should be testing `TAG_AS_JSON`, TAG will always be there.

adammw (approved):
👍  apart from comment in test

PR #3753, opened 2020-03-24
Tune requests/limits

uthark:

Tune samson secret puller container.

Screen Shot 2020-03-24 at 10 15 41 AM

### References
- Jira link: 

### Risks
- Low: init container might be OOMKilled

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -323,8 +323,8 @@ module Kubernetes
           {name: "VAULT_PREFIX", value: Samson::Secrets::VaultClientManager::PREFIX}
         ],
         resources: {
-          requests: {cpu: "100m", memory: "100Mi"},
-          limits: {cpu: "500m", memory: "256Mi"}
+          requests: {cpu: "100m", memory: "64Mi"},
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -323,8 +323,8 @@ def set_secret_puller
           {name: "VAULT_PREFIX", value: Samson::Secrets::VaultClientManager::PREFIX}
         ],
         resources: {
-          requests: {cpu: "100m", memory: "100Mi"},
-          limits: {cpu: "500m", memory: "256Mi"}
+          requests: {cpu: "10m", memory: "64Mi"},
+          limits: {cpu: "10m", memory: "64Mi"}
```
grosser: I'd rather request more cpu so it finishes fast

grosser (approved):

plugins/kubernetes/app/models/kubernetes/template_filler.rb
grosser: ... an inline comment about why we set requests = limits would be good too

lennardseah (approved):

grosser (approved):

PR #3755, opened 2020-03-25
give us more kill timeout to handle cancel interrupt

grosser:
1s is a little short ...
if we do more fancy things then maybe this needs to be configurable

@zendesk/compute

### DIFF
--- a/app/models/terminal_executor.rb
+++ b/app/models/terminal_executor.rb
@@ -17,7 +17,7 @@ class TerminalExecutor
   SECRET_PREFIX = "secret://"
   HIDDEN_PREFIX = "hidden://"
   HIDDEN_TXT = "HIDDEN"
-  KILL_TIMEOUT = 1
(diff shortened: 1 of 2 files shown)


app/models/terminal_executor.rb
```diff
@@ -17,7 +17,7 @@ class TerminalExecutor
   SECRET_PREFIX = "secret://"
   HIDDEN_PREFIX = "hidden://"
   HIDDEN_TXT = "HIDDEN"
-  KILL_TIMEOUT = 1
+  KILL_TIMEOUT = (Rails.env.test? ? 1 : 5)
```
adammw: hmm, still seems short. also not a fan of just random hardcoded config

grosser:
good enough for our usecases ... and nobody wanted that anyway in so long
...
if we need more flexibility we can turn it into a column

app/models/terminal_executor.rb
grosser: turned it into an env var 🤷‍♂

PR #3756, opened 2020-03-25
Ignore multiple RoleBindings

uthark:

Ignore multiple RoleBindings

### References
- Jira link: 

### Risks
- Low


uthark:
cc @adammw

grosser (commented):
we need to confirm no existing RoleBinding would get it's name changed by this
... maybe simpler to change a single project to use `samson/keep_name: "true"` to resolve it
once the project is migrated to namespaces these restrictions no longer apply

adammw (approved):

grosser (changes requested):
see comment above

uthark:
@grosser but `Role`, `ConfigMap` and `PodSecurityPolicy` are namespaced..

grosser:
the logic is "you using namespaces ? ... k, no name change foobar for you" :)

uthark:
Got it. So, @adammw then needs to use `samson/keep_name` annotation?

grosser:
Either that or we need to check if it's safe to enable.

PR #3758, opened 2020-03-27
wait for commit to be in git repository

grosser:
trying to fix issue found in https://github.com/zendesk/samson/pull/3747 after https://github.com/zendesk/samson/pull/3748 did not fully fix the issue

/cc @zendesk/compute @patrobinson

### DIFF
--- a/test/models/job_execution_test.rb
+++ b/test/models/job_execution_test.rb
@@ -250,11 +250,17 @@ describe JobExecution do
     job.output.must_include "Could not find commit for nope"
   end
 
-  it 'errors if job commit resultion fails, but checkout works' do
-    GitRepository.any_instance.expects(:commit_from_ref).returns nil
+  it 'errors if job commit resolution fails' do
+    GitRepository.any_instance.expects(:commit_from_ref).times(4).returns nil
     execute_job
     assert_equal 'errored', job.status
     job.output.must_include "Could not find commit for master"
+  end
+
+  it 'succeeds if job commit resolution fails for a bit' do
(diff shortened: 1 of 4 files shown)


test/models/job_execution_test.rb
```diff
@@ -251,7 +251,7 @@ def last_line_of_output
   end
 
   it 'errors if job commit resultion fails, but checkout works' do
```
adammw:
```suggestion
  it 'errors if job commit resolution fails, but checkout works' do
```

test/models/job_execution_test.rb
```diff
@@ -251,7 +251,7 @@ def last_line_of_output
...
-    GitRepository.any_instance.expects(:commit_from_ref).returns nil
+    GitRepository.any_instance.expects(:commit_from_ref).times(4).returns nil
```
adammw: separate test for fails 3 times then works?

patrobinson:
@grosser Unfortunately this doesn't fix the issue. I deployed this commit to our Production environment and it's still happening.

Screen Shot 2020-03-30 at 12 29 56 pm

grosser:
ahhh ... it needs to update the mirror :D

grosser:
hope this does it ... 1 more test plz 🤞

patrobinson:
@grosser Confirmed this is working now, thanks for your help

grosser:
awesome! :D

grosser:
I guess the root issue was that when making a new tag the api will know about it, but it takes a few seconds to be available in git ...

PR #3761, opened 2020-04-02
Allow duplicate resource names in different namespaces

adammw:

Allow duplicate resource names in different namespaces. Currently when doing so you get an error:
> Only use a maximum of 1 of each kind in a role (except Service, APIService, CustomResourceDefinition, ConfigMap, Role, ClusterRole, Namespace, PodSecurityPolicy, and ClusterRoleBinding)

### References
- Jira link: 

### Risks
- Low: could have duplicates if one has the default namespace specified and the other doesn't but gets filled in later by template_filler.


grosser (approved):

grosser:
that should work :)

PR #3763, opened 2020-04-02
support loading env variable groups from external urls

grosser:
... this should make migration easier because we can just set the url on the existing env var groups

Screen Shot 2020-04-02 at 12 10 26 PM

@zendesk/compute @zendesk/config @zendesk/eng-productivity

### DIFF
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -67,7 +67,13 @@ class EnvironmentVariable < ActiveRecord::Base
     end
 
     def env_vars_from_external_groups(project, deploy_group)
-      project.external_environment_variable_groups.each_with_object({}) do |group, envs|
+      external_groups = project.external_environment_variable_groups
+
+      external_groups += (project.environment_variable_groups.select(&:external_url?).map do |env_group|
+        ExternalEnvironmentVariableGroup.new(url: env_group.external_url)
+      end)
+
+      external_groups.each_with_object({}) do |group, envs|
         group_env = group.read[deploy_group.permalink]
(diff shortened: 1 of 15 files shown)


aglover-zendesk:
nice! yeah this would make for a much nicer migration and make it easier to manage the ref to the external data going forward

grosser:
added validations and preview

Screen Shot 2020-04-02 at 2 05 40 PM
Screen Shot 2020-04-02 at 2 14 04 PM
Screen Shot 2020-04-02 at 2 15 06 PM

plugins/env/app/models/environment_variable.rb
```diff
@@ -67,6 +67,12 @@ def env_vars_from_db(deploy, deploy_group, **args)
     end
 
     def env_vars_from_external_groups(project, deploy_group)
+      external_groups = project.external_environment_variable_groups
+
+      external_groups.concat(project.environment_variable_groups.select(&:external_url?).map do |env_group|
+        ExternalEnvironmentVariableGroup.new(url: env_group.external_url)
+      end)
+
       project.external_environment_variable_groups.each_with_object({}) do |group, envs|
```
sathishavm, line 76: it should be `external_groups` here?

grosser: yes, thx!

grosser: ... means I'm missing a test too

grosser: hehe, it actually worked because it modified the external_environment_variable_groups ... 2 bugs make 1 feature ;)

PR #3772, opened 2020-04-15
ignore outofcpu and outofmemory events

uthark:
Ignore `OutOfcpu` and `OutOfmemory` events.
### References
- Jira link: 

### Risks

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource_status.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource_status.rb
@@ -7,7 +7,9 @@ module Kubernetes
       Pod: [
         # These errors may happens when Deployment which uses PVC is updated. Ignore them.
         "FailedAttachVolume",
-        "FailedMount"
+        "FailedMount",
+        "OutOfcpu",
+        "OutOfmemory"
       ],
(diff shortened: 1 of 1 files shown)


eatwithforks (approved):

grosser:
I don't think this helps ... we filter them out by checking `@pod.waiting_for_resources?`

uthark:
If we fitler them out, then why in error logs we see `OutOfCpu`?That was the only error event.

grosser:
not sure ... but sounds like there is a different bug then 🤷 <-> last PR did not work

grosser:
ah... I think what happens is that samson is no longer in the 'getting ready' but 'verify it is stable' phase ... so something being 'pending' makes it fail

grosser:
hmm no, example deploy was still getting ready ... digging ...

grosser:
... yeah it stopped on `phase == 'Failed'`

grosser:
... so would need to say "if pod is failed < 5 min ago && out-of-cpu" then treat it like "waiting for resources"

grosser:
solved by https://github.com/zendesk/samson/pull/3779

PR #3776, opened 2020-04-23
Added external_environment_variable_groups as allowed includes

davidreuss:
For tooling that depends on samson export data there is currently no
good way of pulling what external groups are configured for a project.

Add endpoints to list all external env groups, and allow them as an
include on a projects json export.

### References
- Jira link: https://zendesk.atlassian.net/browse/EPEMEA-372

### Risks
- Low: This only adds export of metadata for external env groups

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -74,9 +74,8 @@ class ProjectsController < ResourceController
   def allowed_includes
     [
       :environment_variable_groups,
-      :environment_variables_with_scope,
-      (:external_environment_variable_groups if ExternalEnvironmentVariableGroup.configured?)
-    ]
+      :environment_variables_with_scope
+    ] + Samson::Hooks.fire(:project_allowed_includes).flatten(1)
   end
 
   def resource_params
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -75,6 +75,7 @@ class ProjectsController < ResourceController
     [
       :environment_variable_groups,
       :environment_variables_with_scope,
+      (:external_environment_variable_groups if ExternalEnvironmentVariableGroup.configured?)
     ]
   end
 
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -215,6 +215,10 @@ class Project < ActiveRecord::Base
     )
   end
 
+  def external_environment_variable_groups
+    ExternalEnvironmentVariableGroup.find_by_project_id(id)
+  end
+
   def as_json(methods: [], **options)
     super(
       {
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -215,10 +215,6 @@ class Project < ActiveRecord::Base
     )
   end
 
-  def external_environment_variable_groups
-    ExternalEnvironmentVariableGroup.find_by_project_id(id)
-  end
-
   def as_json(methods: [], **options)
     super(
       {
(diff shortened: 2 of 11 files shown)


app/controllers/projects_controller.rb
```diff
@@ -75,6 +75,7 @@ def allowed_includes
     [
       :environment_variable_groups,
       :environment_variables_with_scope,
+      (:external_environment_variable_groups if ExternalEnvironmentVariableGroup.configured?)
     ]
```
grosser, line 79: having a nil in here could lead to trouble, better `.compact`
```suggestion
    ].compact
```

grosser: this should be done in the `env` plugin via a decorator or a hook

app/models/project.rb
```diff
@@ -215,6 +215,10 @@ def environment_variables_with_scope
     )
   end
 
+  def external_environment_variable_groups
```
grosser, line 218: - same here, should be done in the plugin
 - name is plural, but returns a single one ... seems fishy

plugins/env/config/routes.rb
```diff
@@ -6,6 +6,9 @@
     end
   end
 
+  get '/external_environment_variable_groups(:format)',
```
grosser, line 9: this already exists from the `resources` block above

plugins/env/app/controllers/external_environment_variable_groups_controller.rb
```diff
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 class ExternalEnvironmentVariableGroupsController < ApplicationController
+  def index
+    @groups = ExternalEnvironmentVariableGroup.all
+    respond_to do |format|
+      format.html
```
grosser, line 6: view does not exist ... so delete this ?

davidreuss:
yeah, this is a bit of wip .. 

could you give me some pointers on how to move the logic into the plugin itself? .. any examples to go on?

app/models/project.rb
grosser: already exists as external_environment_variable_groups

grosser:
pushed a commit to use hooks, rest should be straight-forward

plugins/env/config/routes.rb
davidreuss: isn't that only for `environment_variable_groups` ? .. not the _external_ ones?

grosser: ah yes ... I pushed a commit to clean up the routes :)

test/controllers/projects_controller_test.rb
```diff
@@ -4,6 +4,8 @@
 SingleCov.covered!
 
 describe ProjectsController do
+  with_env EXTERNAL_ENV_GROUP_S3_REGION: "us-east-1", EXTERNAL_ENV_GROUP_S3_BUCKET: "a-bucket"
```
grosser, line 7: this should only be in the test in 180 as
```
  with_env EXTERNAL_ENV_GROUP_S3_REGION: "us-east-1", EXTERNAL_ENV_GROUP_S3_BUCKET: "a-bucket" do
    ... test ...
  end
```

grosser: pushed a fix

davidreuss:
thanks

when i removed the route locally it broke — so awaiting test runs :)

davidreuss:
yay, let’s try it

PR #3781, opened 2020-05-06
allow overwriting static env vars if user requests it

grosser:
alternative to https://github.com/zendesk/samson/pull/3780 without "might break lots of things" concern ... new logic only for those that request it

- also allows overriding plugin/db env vars
- when container does not set the var it was asked to keep it come out as unset

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -511,10 +511,18 @@ module Kubernetes
       end
 
       env_containers.each do |c|
+        extra = all
         env = (c[:env] ||= [])
-        env.concat all
 
-        # unique, but keep last elements
+        # keep container env var if requested, so static+plugin env can be overwritten
+        if keep = samson_container_config(c, :"samson/keep_env_var").to_s.split(/, ?| /).presence
+          extra = all.dup
+          keep.each { |var| extra.delete_if { |e| e[:name] == var } }
+        end
+
+        env.concat extra
+
+        # unique, but keep user configured overrides
         env.reverse!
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -511,10 +511,18 @@ def set_env
       end
 
       env_containers.each do |c|
+        extra = all
         env = (c[:env] ||= [])
-        env.concat all
 
-        # unique, but keep last elements
+        # keep container env var if requested, so static+plugin env can be overwritten
+        if keep = samson_container_config(c, :"samson/keep_env_var").to_s.split(",").presence
```
craig-day, line 518: can we split this on whitespace? Then we can use the nice `>` or `|` YAML operators?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -511,10 +511,18 @@ def set_env
...
+          extra = all.dup
```
craig-day, line 519: any reason we need to `dup` here? can we `dup` above and avoid reassigning extra to the same value here?

craig-day (approved):
Seems fairly simple.

plugins/kubernetes/app/models/kubernetes/template_filler.rb
grosser: yeah, can do both

plugins/kubernetes/app/models/kubernetes/template_filler.rb
grosser: I'm avoiding the extra work for the 99% route

PR #3786, opened 2020-05-20
Allow skipped status checks

ellimist:
* Add description
* Add unit tests

### References
- Jira link: 

### Risks

### DIFF
--- a/test/models/commit_status_test.rb
+++ b/test/models/commit_status_test.rb
@@ -291,6 +291,12 @@ describe CommitStatus do
         status.state.must_equal 'error'
       end
 
+      it 'maps skipped status to success' do
+        stub_github_api(check_suite_url, check_suites: [{conclusion: 'skipped', id: 1}])
+
+        status.state.must_equal 'success'
+      end
+
       it 'picks highest priority check conclusion/status equivalent' do
         stub_github_api(
           check_suite_url,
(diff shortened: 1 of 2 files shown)


grosser:
how do skipped statuses happen ?

davidreuss:
They happen when you have a github action workflow with a `if` condition on the job, so it's skipped if it evaluates to false.

In this case, we have different users creating `Deployment` objects through the GitHub Deployment API, and we don't want to run our workflow unless it's a specific user who created it.

In any case, i don't think it should crash on this status as it can clearly happen, and likely will be happening much more with the recent adoption of github actions for implementing all sorts of workflows.

test/models/commit_status_test.rb
```diff
@@ -291,6 +291,12 @@ def stub_checks_api(commit_status: status)
         status.state.must_equal 'error'
       end
 
+      it 'maps skipped status to success' do
```
grosser, line 294: FYI test should explain why something happens, not what happens
```
it "does not fail when action was intentionally skipped"
```

grosser:
I can roll this out tomorrow

ronan-mch:
@davidreuss - do you expect that further use of GitHub Actions will lead to more status check types? I'm wondering if we're going to need to keep on making this change as we add more Actions...

davidreuss:
@ronan-mch I don't know. But i'm not expecting it, but there must be some documentation on the types of statuses that the github app sets for github action statuses (whew, that was hard to re-read)

PR #3787, opened 2020-05-25
fix boot check crashing when logger was loaded first

grosser:
logger requires ruby concurrent and that sets up some global thread finalizer, nothing we need to worry about and hard to disable ... so let's ignore it

@zendesk/compute

### DIFF
--- a/lib/samson/boot_check.rb
+++ b/lib/samson/boot_check.rb
@@ -10,12 +10,16 @@ module Samson
             ActiveRecord::Base.connection_pool.stat.fetch(:busy) == 0
           end
         else
+          extra_threads = (Thread.list - [Thread.current]).reject do |t|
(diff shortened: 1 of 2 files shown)


lib/samson/boot_check.rb
```diff
@@ -10,12 +10,16 @@ def check
             ActiveRecord::Base.connection_pool.stat.fetch(:busy) == 0
           end
         else
+          extra_threads = (Thread.list - [Thread.current]).reject do |t|
+            t.backtrace(1).first.match?(/ruby_thread_local_var/) # get started by loading logger
```
adammw, line 14:
```suggestion
            t.backtrace(1).first.match?(/ruby_thread_local_var/) # these threads get started by loading logger - ignore them
```

PR #3793, opened 2020-06-02
prevent duplicate deploys by disabling button after clicking

grosser:
I had a few duplicate deploys because I tried to click through Review->Deploy but the button started on deploy, so it did 2 deploys ... this should stop that

Jun-01-2020 18-26-02

@zendesk/compute

### DIFF
--- a/app/assets/javascripts/deploys.js
+++ b/app/assets/javascripts/deploys.js
@@ -84,6 +84,28 @@ $(function () {
     }
   }
 
+  function showChangeset($form) {
+    toggleConfirmed();
+    $("#deploy-confirmation").show();
+
+    storeActiveTab($form);
+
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/deploys.js
```diff
@@ -92,30 +114,15 @@ $(function () {
...
-        success: function(data) {
-          $placeholderPanes.detach();
-          $container.append(data);
-          showActiveTab($this);
-        }
-      });
+    var $form = $(this);
 
+    // user pressed `Review`: load in changeset
+    // user pressed `Deploy!`: start the deploy
```
adammw, line 120: isn't this the second branch? put this comment on the else statement?

PR #3804, opened 2020-06-12
allow explicit recreate to get rid of current state (service NodePort…

grosser:
…->ClusterIP for example)

atm moving a service from NodePort to ClusterIP needs a full delete/deploy cycle which takes ~1min and needs to be done manually
error in UI is this atm:
```
[19:57:31] JobExecution failed: Kubernetes error Service foo-service sandbox Sandbox: Service "foo-service" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'
```

this option should be useful in general and make it more like ~1s

@zendesk/compute

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -316,6 +316,15 @@ Delete old resource and create new when an update fails because it `cannot chang
 metadata.annotation.samson/force_update: "true"
 ```
 
+### Forcing a delete-create to get back to a clean state
+
+Delete old resource and create new.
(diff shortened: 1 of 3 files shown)


adammw:
didn't we already have a flag for this? force_update or something?

uthark (approved):

grosser:
yes, force_update, but this error is not a conflict but a "invalid" and there is no way of telling if the invalid came from a bad deploy or the mix-and-match ... so I'm thinking a separate flag might be better

PR #3806, opened 2020-06-16
[WIP] Add support for check_suit event type for Github Action

sfdye:
## Add a new webhook event for GitHub - check_suite.

We want GitHub to be the controller and determine when to ping Samson (rather than the CI piece of the pipeline). When a GitHub Action completes, GitHub becomes aware of the results of the unit test, and when the result is a success, GitHub can post out to notify the world of the success, using the check_suite api payload. Samson, Spinnaker, and another CD tooling should be aware of this use-case and respond correctly.

### DIFF
--- a/app/models/changeset/check_suite.rb
+++ b/app/models/changeset/check_suite.rb
@@ -12,7 +12,7 @@ class Changeset::CheckSuite
   end
 
   def self.valid_webhook?(payload)
-    payload['action'] == 'completed' && payload['check_suite']['status'] == 'completed' && payload['check_suite']['conclusion'] == 'success'
+    payload['check_suite']['status'] == 'completed' && payload['check_suite']['conclusion'] == 'success'
   end
 
   def sha
--- a/app/models/changeset/check_suite.rb
+++ b/app/models/changeset/check_suite.rb
@@ -12,7 +12,7 @@ class Changeset::CheckSuite
   end
 
   def self.valid_webhook?(payload)
-    payload['check_suite']['status'] == 'completed' && payload['check_suite']['conclusion'] == 'success'
+    payload['check_suite']['conclusion'] == 'success'
   end
 
   def sha
--- a/app/models/changeset/check_suite.rb
+++ b/app/models/changeset/check_suite.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+class Changeset::CheckSuite
+  attr_reader :repo, :data
+
+  def initialize(repo, data)
+    @repo = repo
+    @data = data
+  end
+
+  def self.changeset_from_webhook(project, payload)
+    new(project.repository_path, payload)
+  end
+
+  def self.valid_webhook?(payload)
+    payload['action'] == 'completed' && payload['check_suite']['status'] == 'completed' && payload['check_suite']['conclusion'] == 'success'
+  end
+
+  def sha
+    data['check_suite']['head_sha']
+  end
+
+  def branch
+    data['check_suite']['head_branch']
+  end
+
+  def message
+    nil
+  end
+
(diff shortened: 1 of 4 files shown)


grosser:
looks about right ... what usecase are you trying to solve ?

app/models/changeset/check_suite.rb
```diff
@@ -0,0 +1,33 @@
...
+    @repo = repo
+    @data = data
+  end
+
+  def self.changeset_from_webhook(project, payload)
+    new(project.repository_path, payload)
+  end
+
+  def self.valid_webhook?(payload)
+    payload['action'] == 'completed' && payload['check_suite']['status'] == 'completed' && payload['check_suite']['conclusion'] == 'success'
```
grosser, line 15: does it need to check all 3 ?
I'd imagine a `completed` hook has status 'completed' too

test/controllers/integrations/github_controller_test.rb
```diff
@@ -138,4 +138,31 @@ def payload
...
+    end
+
+    let(:payload) do
+      {
+          action: 'completed',
+          check_suite: {
+              head_branch: 'master',
+              head_sha: '31b148e19a8ef2a033cb4ff485949c3f3d689140',
+              status: 'completed',
+              conclusion: 'success'
```
grosser, line 156: 2-space indent plz

app/models/changeset/check_suite.rb
sfdye: The use case is we want to send a webhook to inform Samson once all the jobs within a Github workflow have completed. This is per [Github official recommendation](https://github.community/t/official-webhook-action/116972), but currently no such action exists.

sfdye:
> looks about right ... what usecase are you trying to solve ?

The use case is we want to send a webhook to inform Samson once all the jobs within a Github workflow have completed. This is per Github official recommendation, but currently no such action exists.

grosser:
but why do you want to send that webhook ... to trigger a deploy ? to refresh the cache ?

sfdye:
> but why do you want to send that webhook ... to trigger a deploy ? to refresh the cache ?

Yes, to trigger a deploy, or just simply tag creation.

everops-alex:
Thanks @sfdye for making this PR!

@grosser essentially, we are looking to avoid triggering webhooks through github actions (which is how travis communicates w/ samson).

We want GitHub to be the controller and determine when to ping samson (rather than the CI piece of the pipeline). When a GitHub Action completes, GitHub becomes aware of the results of the unit test, and when the result is a success, GitHub can post out to notify the world of the success, using the `check_suite` api payload. Samson, spinnaker, and other CD tooling should be aware of this use-case and respond correctly.

This removes "triggering CD" from the unit test logic, and ensures GitHub Actions compute cycles are used purely for CI initiatives with no ability to trigger downstream events from the CI logic.

app/models/changeset/check_suite.rb
```diff
@@ -0,0 +1,33 @@
...
+    @repo = repo
+    @data = data
+  end
+
+  def self.changeset_from_webhook(project, payload)
+    new(project.repository_path, payload)
+  end
+
+  def self.valid_webhook?(payload)
+    payload['check_suite']['status'] == 'completed' && payload['check_suite']['conclusion'] == 'success'
```
grosser, line 15: if it has conclusion success, then it's also completed, right ?
```suggestion
    payload['check_suite']['conclusion'] == 'success'
```

grosser:
let me know if you need help merging or deploying

shoukoo:
@grosser , thanks for approving this PR!
Could you please help me deploying this to Samson staging?

PR #3821, opened 2020-06-26
remove unused ingress code

grosser:
we stopped using this a while ago, time to clean it up


uthark (approved):
replacement for this feature: use OPA gatekeeper.

PR #3823, opened 2020-06-26
patch request do not have template as first arguments / crash when co…

grosser:
…nflic happens

noticed that when playing with server-side-apply dryRun


eatwithforks (approved):

adammw:
how can a patch conflict? so the intention is just to hard crash instead?

grosser:
that code was never reached or hard-crashed ... if it ever turns out to be an issue we can bring it back

PR #3826, opened 2020-07-06
replace server-side-apply hack with bumped gem

grosser:
https://github.com/abonas/kubeclient/pull/448
@zendesk/compute


eatwithforks (approved):

cben:
any more remaining uses of `with_header` that are not covered by kubeclient?

grosser:
good catch, did not check for that :D

https://github.com/zendesk/samson/pull/3829

PR #3827, opened 2020-07-07
Add external build metric

hdt91:
### Descriptions
In an setup where Samson integrates with an external build service, builds are very likely created continuously. Adding a metric to track this event can help detect some issues from the build service, broken integration or maybe Samson if the count disappears.  

### References
- Related to one of internal remediation actions.

### Risks
- Low: Only add a new metric.

@zendesk/eng-productivity

### DIFF
--- a/app/controllers/builds_controller.rb
+++ b/app/controllers/builds_controller.rb
@@ -63,6 +63,10 @@ class BuildsController < ApplicationController
 
     start_docker_build if saved && !registering_external_build?
 
+    # Emit a metric to help users assess condition of integration between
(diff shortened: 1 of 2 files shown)


hdt91:
> would listening to the builds controller metric + status-code not be enough ?
> samson.app.web.request.status.200 + controller:buildscontroller
> 
> ... ideally that should be cleaned up to `samson.app.web.request` with `status:200` and `controller:builds`

That sounds like it should definitely work. I'll try it and will close the PR if it sufficient. Thanks for pointing that out :)

PR #3830, opened 2020-07-08
Handle stale state for Checks

sathishavm:
Added code to handle `stale` state for GitHub checks suites.

https://developer.github.com/v3/checks/runs/
```
If a check run is in a incomplete state for more than 14 days, then the check run's conclusion becomes stale and appears on GitHub as stale with . Only GitHub can mark check runs as stale
```


grosser:
nice catch :D

PR #3832, opened 2020-07-11
simplify github action names / setup

grosser:
@zendesk/compute 
/cc @everops-john

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,4 +1,4 @@
-name: repo-checks
+name: CI
 
 on:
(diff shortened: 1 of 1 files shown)


adammw:
idk.. i assume this is generated by some script, now this would be different to all other repos. 🤷

grosser:
I think nobody is going to go over these repos and re-apply the script ... so now it's on us to make it as simple as possible

grosser:
only thing we could keep is `name: repo-checks` ... but I still think CI is nicer

PR #3833, opened 2020-07-14
allow requesting current logs while still deploying

grosser:
Screen Shot 2020-07-14 at 1 58 24 PM
 
@zendesk/compute 

ideally users would stream the deploy, but that's more complicated to support / to consume

### DIFF
--- a/app/controllers/deploys_controller.rb
+++ b/app/controllers/deploys_controller.rb
@@ -97,6 +97,7 @@ class DeploysController < ApplicationController
           type: 'text/plain'
       end
       format.json do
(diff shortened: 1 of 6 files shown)


app/controllers/deploys_controller.rb
```diff
@@ -97,6 +97,7 @@ def show
           type: 'text/plain'
       end
       format.json do
+        @deploy.job.serialize_execution_output if params[:serialize_execution_output] # show live job output
```
adammw, line 100: little obtuse parameter naming, but at least it's 1:1 with the method name

PR #3834, opened 2020-07-15
Revert "Add support for check_suit event type for Github Action"

sfdye:
Reverts zendesk/samson#3806

Reverting as the implementation is not entirely correct and we found an alternative by using https://github.com/zendesk/ga/tree/master/samson-create-release.


grosser:
thx for cleaning it up!
can you elaborate on what is not correct in case someone wants to redo this PR ?

sfdye:
IIRC the issue is `check_suite` does not imply it's coming from Github Action, it could be from other checks such as dependabot or Zendesk language check.

PR #3840, opened 2020-08-11
do not duplicate volumes/mounts so users can add their own

grosser:
users might have their own vaultauth usage or trying to transition out of samson
@zendesk/compute

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -340,11 +340,14 @@ module Kubernetes
 
       # share secrets volume between all pod containers
       pod_containers.each do |container|
-        (container[:volumeMounts] ||= []).push secret_vol
+        mounts = (container[:volumeMounts] ||= [])
+        mounts.push secret_vol
+        mounts.uniq!
       end
 
       # define the shared volumes in the pod
-      (pod_template[:spec][:volumes] ||= []).concat [
+      volumes = (pod_template[:spec][:volumes] ||= [])
+      volumes.concat [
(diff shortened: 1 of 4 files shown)


eatwithforks (approved):

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -340,11 +340,14 @@ def set_secret_puller
 
       # share secrets volume between all pod containers
       pod_containers.each do |container|
-        (container[:volumeMounts] ||= []).push secret_vol
+        mounts = (container[:volumeMounts] ||= [])
+        mounts.push secret_vol
+        mounts.uniq!
```
adammw, line 345: shouldn't it be unique per-path?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -354,6 +357,7 @@ def set_secret_puller
           }
         }
       ]
+      volumes.uniq!
```
adammw, line 360: unique per name?

grosser: for now just dealing with exact copies, if 2 things have the same name and a different config, the user needs to see what is going on and figure out a compromise

adammw: still pretty jank.. maybe just raise if they're the same?

grosser: that's what we had before, having the same volume made samson blow up because the manifest became invalid
so users had to have a locally invalid manifest that uses a vaultauth mount but no volume

PR #3845, opened 2020-08-18
use apply to sync namespaces to avoid get+post

grosser:
we can't apply a List since that's not an api
but this should cut down on requests
also moved the parallelization further down so work gets dealt out equally and no threads are done early

@zendesk/compute


eatwithforks (approved):
⭐

grosser:
also looked into if we can do lazy updates, but with fieldManager stuff this all gets complicated and it would not behave like an apply anymore, meaning you could never remove something you added :/

grosser:
... ok reworked it to do a diff and then only apply as necessary, this should basically make it instant unless there are clusters that time out

eatwithforks (approved):

PR #3850, opened 2020-08-27
Improve logging for `create_release?` logic

disposedtrolley:
This PR changes the log message generated after `create_release?` is called to determine if a release should be created for the current request. The previous log message didn't mention additional checks for `service_type` and `service_name` which factored into the outcome, and caused some confusion as to why releases were being skipped.

Since this log is returned in the response body from inbound webhooks, it'd be good to include this extra context.

### References
- Jira link: 

### Risks
- Low

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -16,7 +16,7 @@ class Integrations::BaseController < ApplicationController
 
     if branch
       create_release = project.create_release?(branch, service_type, service_name)
-      record_log :info, "Branch #{branch} is release branch: #{create_release}"
+      record_log :info, "Create release for branch [#{branch}], service_type [#{service_type}], service_name [#{service_name}]: #{create_release}"
       release = find_or_create_release if create_release
     else
       record_log :info, "No branch found, assuming this is a tag and not creating a release"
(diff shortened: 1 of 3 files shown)


app/controllers/integrations/base_controller.rb
```diff
@@ -16,7 +16,8 @@ def create
 
     if branch
       create_release = project.create_release?(branch, service_type, service_name)
-      record_log :info, "Branch #{branch} is release branch: #{create_release}"
+      record_log :info, "Create release for "\
+        "branch [#{branch}], service_type [#{service_type}], service_name [#{service_name}]: #{create_release}"
```
grosser, line 20:
```suggestion
      record_log(
        :info, 
        "Created release for branch: #{branch}, service_type: #{service_type}, service_name: #{service_name}: #{create_release}"
      )
```

disposedtrolley: It's still a _little_ too long for Rubocop's liking :(

grosser: dammit :D

disposedtrolley:
Do you mind merging this one in @grosser? I don't seem to have permission!

PR #3855, opened 2020-09-15
add inline docs for $deploy_group path replacement

grosser:
docs for https://github.com/zendesk/samson/pull/3714
Screen Shot 2020-09-15 at 4 20 54 PM


eatwithforks (approved):

adammw:
Would prefer something along the lines of "Can use variable substitution to have a dynamic file path. Supported variables: $DEPLOY_GROUP, $ENVIRONMENT, (actually list what it supports, or link to a list)"

grosser:
updated
only deploy_group is supported

PR #3857, opened 2020-09-21
Allow users to generate deploy csv report for multiple projects

anthonywoo:
There is a use case where users need to get all the deploys from a certain time period for a select list of projects. 

Samson already has a similar feature in the [deploys search controller](https://github.com/zendesk/samson/blob/master/app/controllers/deploys_controller.rb#L147-L150), however, this is limited to 1000 deploys. I've copied similar logic to the csv export controller which does not have a limit on the number of deploys

Note: When using `project_permalinks`, I'm expecting another system to generate a link to the csv exports page with the intended projects. An example url: `https://samson.com/csv_exports/new?project_permalinks=projectone%2Cprojecttwo%2Cprojectthree`

### Risks
- Low

### DIFF
--- a/app/controllers/csv_exports_controller.rb
+++ b/app/controllers/csv_exports_controller.rb
@@ -122,7 +122,10 @@ class CsvExportsController < ApplicationController
       end
     end
 
-    if project = params[:project]&.to_i
+    if project_permalinks = params[:project_permalinks].to_s.split(",").presence
+      project_ids_from_permalinks = Project.where(permalink: project_permalinks).pluck(:id)
+      filter['stages.project_id'] = project_ids_from_permalinks
+    elsif project = params[:project]&.to_i
       if project > 0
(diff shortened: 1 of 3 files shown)


app/controllers/csv_exports_controller.rb
```diff
@@ -122,7 +122,10 @@ def deploy_filter
       end
     end
 
-    if project = params[:project]&.to_i
+    if project_permalinks = params[:project_permalinks].to_s.split(",").presence
+      project_ids_from_permalinks = Project.where(permalink: project_permalinks).pluck(:id)
+      filter['stages.project_id'] = project_ids_from_permalinks
```
grosser, line 127: 🤷 
```suggestion
      filter['stages.project_id'] = Project.where(permalink: project_permalinks).pluck(:id)
```

grosser (approved):
seems straight forward 👍

PR #3861, opened 2020-09-23
change CsvExport filters column from string to text

anthonywoo:
Now that multiple project ids can be filtered, it is possible for the csvexport filters to be more than 255 characters.

Change the column type to text

I still need to test this on staging before merging to see if there are any additional edge cases

One weird thing when investigating was the CsvExports were able to be saved even if the filters were more than 255 characters, but it would drop the additional characters (but this would lead to an error later when trying to load the report as the filter couldn't be deserialized)
From the db:
```
"{\"deploys.created_at\":\"2020-04-01..2020-09-23\",\"environments.production\":true,\"stages.project_id\":[5,6,7,12,13,14,21,30,31,55,56,57,61,69,79,80,83,87,89,91,92,103,109,112,122,129,161,166,168,173,187,188,191,200,203,206,208,216,217,218,219,221,223,225,226,"
```

### Risks
- Low

### DIFF
--- a/app/models/csv_export.rb
+++ b/app/models/csv_export.rb
@@ -2,6 +2,8 @@
 class CsvExport < ActiveRecord::Base
   belongs_to :user, inverse_of: :csv_exports
   serialize :filters, JSON
+  attribute :filters, :text, default: {}
+
   STATUS_VALUES = ['pending', 'started', 'finished', 'downloaded', 'failed', 'deleted'].freeze
 
   before_destroy :delete_file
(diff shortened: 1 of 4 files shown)


grosser:
need to check that a new export with nil in the db does not crash ... it used to be `{}` before

anthonywoo:
It doesn't look possible for an end user to create an export with filters to be nil
https://github.com/zendesk/samson/blob/master/app/controllers/csv_exports_controller.rb#L87-L89

but I added a default value to filters on the model as an assurance

grosser (approved):

anthonywoo:
Tested on staging. Able to successfully create exports when specifying no projects and when specifying all projects.

PR #3862, opened 2020-09-28
do not consider None as a missing risks

grosser:
PRs like
Screen Shot 2020-09-28 at 4 24 07 PM

should not lead to deploy warnings about missing risks
also removed the permanent info heading in risks tab since it shows nicely when risks are missing

@zendesk/compute 
/cc @bquorning who found it 🎉

### DIFF
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -131,8 +131,12 @@ class Changeset::PullRequest
   def risks
     return @risks if defined?(@risks)
     @risks = parse_risks(@data.body.to_s)
-    @missing_risks = @risks.nil?
-    @risks = nil if @risks&.match?(/\A\s*\-?\s*None\Z/i)
+    if @risks&.match?(/\A\s*\-?\s*None\Z/i)
+      @risks = nil
+      @missing_risks = false
+    else
+      @missing_risks = @risks.nil?
+    end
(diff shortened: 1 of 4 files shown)


app/models/changeset/pull_request.rb
```diff
@@ -131,8 +131,12 @@ def risky?
   def risks
     return @risks if defined?(@risks)
     @risks = parse_risks(@data.body.to_s)
-    @missing_risks = @risks.nil?
-    @risks = nil if @risks&.match?(/\A\s*\-?\s*None\Z/i)
+    if @risks&.match?(/\A\s*\-?\s*None\Z/i)
+      @risks = nil
```
bquorning, line 135: It looks like the only job of `@risks` is to be asked the question `risks.present?` by `#risky?`. Would it be better to encapsulate this logic in a small object?

```rb
class Risk
  def initialize(string)
    # ...
  end

  def risky?
    # ...
  end
end
```

bquorning (approved):

eatwithforks (approved):

app/models/changeset/pull_request.rb
grosser: possible ... there are lots of ways to refactoring this, but that can be it's own PR, trying to keep this simple :)

PR #3866, opened 2020-10-08
say which kind is missing labels

grosser:


eatwithforks (approved):

uthark (approved):
Add also resource  name in addition to kind.

adammw:
would expect kind + namespace/name

PR #3870, opened 2020-10-26
Add support for PodTemplate in template filler

santouras:
Samson already modifies aspects of a Deployment/StatefulSet/DaemonSet etc. Fun fact, all of these types embed a `PodTemplateSpec`, generally at the key `.spec.template`.

Add in support for modifying a `PodTemplate` resource type directly. The reason this needs to be handled separately is its key is `.template`, rather than the `.spec.template` of these other resources.

### References
- Jira link: 

### Risks
- Low: a `PodTemplate` will not be modified correctly, which is low risk as it isn't being modified at the moment anyway

### DIFF
--- a/plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
@@ -786,6 +786,30 @@ describe Kubernetes::TemplateFiller do
       end
     end
 
+    describe "podtemplate" do
+      let(:result) { template.to_hash }
+      let(:containers) { result.dig_fetch(:template, :spec, :containers) }
+      let(:container) { containers.first }
+      let(:build) { builds(:docker_build) }
+      let(:image) { build.docker_repo_digest }
+
+      before do
+        raw_template.replace(
+          YAML.safe_load(
+            read_kubernetes_sample_file('kubernetes_podtemplate.yml')
+          ).deep_symbolize_keys
+        )
+      end
+
+      it "does not populate spec attribute" do
+        result.fetch(:spec, nil).must_be_nil
+      end
+
+      it "overrides image" do
+        container.fetch(:image).must_equal image
+      end
+    end
+
     describe "pod" do
       before do
         original_metadata = raw_template.fetch(:metadata)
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -786,6 +786,16 @@ def secret_annotations(hash)
       end
     end
 
+    describe "podtemplate" do
+      before do
+        raw_template.replace(YAML.safe_load(read_kubernetes_sample_file('kubernetes_podtemplate.yml')).deep_symbolize_keys)
+      end
+
+      it "works" do
+        template.to_hash
```
grosser, line 795: test that the container is modified in some way, this test would have passed before adding the new code
also test for replicas not being set

grosser (approved):
looks like it should work 👍

uthark (approved):

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
santouras: Also added in test that `:spec` is not filled, given `:template` is the top level element

grosser:
lmk if you need help merging or deploying

plugins/kubernetes/test/models/kubernetes/template_filler_test.rb
```diff
@@ -786,6 +786,29 @@ def secret_annotations(hash)
...
+      let(:containers) { result.dig_fetch(:template, :spec, :containers) }
+      let(:container) { containers.first }
+      let(:build) { builds(:docker_build) }
+      let(:image) { build.docker_repo_digest }
+
+      before do
+        raw_template.replace \
+          YAML.safe_load(
+            read_kubernetes_sample_file('kubernetes_podtemplate.yml')
+          ).deep_symbolize_keys
```
grosser, line 800: FYI same but less funky :D
```suggestion
        raw_template.replace(
          YAML.safe_load(
            read_kubernetes_sample_file('kubernetes_podtemplate.yml')
          ).deep_symbolize_keys
        )
```

santouras: thats only a matter of opinion ;)

no worries, have amended. Thanks!

PR #3871, opened 2020-10-28
Set desired_pod_count for PodTemplate to be 0

santouras:
PodTemplate does not actually create any pods, so zero out this value,
ensuring our deploys don't hang waiting for pods to create that never
will.

### Risks
- Low: PodTemplate is not a resource that is widely used, meaning any fallout is going to be minimal

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -446,6 +446,12 @@ module Kubernetes
       end
     end
 
+    class PodTemplate < VersionedUpdate
+      def desired_pod_count
+        0 # PodTemplates don't actually create pods
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


adammw (approved):

plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -115,6 +115,8 @@ def kind
       def desired_pod_count
         if @delete_resource
           0
+        elsif kind == "PodTemplate"
```
grosser, line 118: not a fan ... all the other overrides work with inheritance

santouras: right you are sir, fixed 👍

grosser (approved):
🎉

PR #3872, opened 2020-10-29
Support token scopes with underscores

russell:
Some of the controllers like `deploy_groups` have underscores in their names and we should be able to add them to our token scope.

### References
- Jira link: 

### Risks


grosser (approved):

grosser:
lmk if you need help merging/deploying

russell:
Yeah, that would be helpful if you could 👍

PR #3873, opened 2020-11-02
Remove deprecated docker login --email flag

anthonywoo:
The `--email` flag was removed on `06/2017` (https://docs.docker.com/engine/release-notes/17.06/#17060-ce)
Lets not support that flag anymore in Samson

### Risks
- Low. Docker login won't work if anyone is using a very old version of docker


grosser (approved):
❤️

anthonywoo:
I verified docker can login successfully on staging

PR #3875, opened 2020-11-04
allow deploying with a branch and an exact commit to avoid race condi…

grosser:
…tions

some branch might be constantly changing, but we want the deploy to still have a nice name

@zendesk/compute

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -92,10 +92,15 @@ class Stage < ActiveRecord::Base
 
   def create_deploy(user, attributes = {})
     before_command = attributes.delete(:before_command)
+    commit = attributes.delete(:commit)
     deploys.create(attributes.merge(release: !no_code_deployed, project: project)) do |deploy|
       commands = before_command.to_s.dup << script
       deploy.build_job(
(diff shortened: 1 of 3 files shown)


uthark (approved):

app/models/stage.rb
```diff
@@ -92,10 +92,15 @@ def deployed_or_running_deploy
...
     before_command = attributes.delete(:before_command)
+    commit = attributes.delete(:commit)
     deploys.create(attributes.merge(release: !no_code_deployed, project: project)) do |deploy|
       commands = before_command.to_s.dup << script
       deploy.build_job(
-        project: project, user: user, command: commands, commit: deploy.reference, bypass_global_lock_check: true
+        project: project,
+        user: user,
+        command: commands,
+        commit: commit || deploy.reference,
```
adammw, line 102: think you should assert that the reference resolves to the commit, otherwise we could get some weird situations where the commit is completely different...

grosser: commits can be non-shas 😞 
old bug/feature

PR #3876, opened 2020-11-12
By default preview all objects for a role

davidreuss:
When faced with migration to spinnaker we actually want to see all roles
to do a faithful comparison of manifests to ensure we got everything
right.

Currently we can only verify the "primary" document which is a good
indicator, but -- we might as well go full monty.

### References
- Jira link: GUIDEOPS-1884

### Risk
- Low: might generate more load on manifest preview, but that's a manual action, so we should be fine.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -72,10 +72,9 @@ module Kubernetes
     end
 
     # Temporary template we run validations on ... so can be cheap / not fully fleshed out
-    # and only be the primary since services/configmaps are not very interesting anyway
     def verification_template
       primary_config = raw_template.detect { |e| Kubernetes::RoleConfigFile.primary?(e) } || raw_template.first
-      Kubernetes::TemplateFiller.new(self, primary_config, index: 0)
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -72,10 +72,9 @@ def custom_resource_definitions
     end
 
     # Temporary template we run validations on ... so can be cheap / not fully fleshed out
-    # and only be the primary since services/configmaps are not very interesting anyway
     def verification_template
       primary_config = raw_template.detect { |e| Kubernetes::RoleConfigFile.primary?(e) } || raw_template.first
-      Kubernetes::TemplateFiller.new(self, primary_config, index: 0)
+      Kubernetes::TemplateFiller.new(self, primary_config, index: index)
```
grosser, line 77: where is index coming from ? ... need to pass that in ?
... also please add a new test (unless is breaks the old tests)
can pair if this is getting tricky

davidreuss:
oh, duh -- i got the context all wrong here, yeah -- might need some guidance on how to do this, not sure if there's a better API or not? it would help if you could decide if you want all roles, or a single role from the preview endpoint.

PR #3880, opened 2020-11-16
Fix paste bug in project environment variables

aleksandre:
There is a bug in the "Paste" feature that prevents us from creating an environment variable from a string.

The Javascript selector that is supposed to find a previous environment variable to create a copy ends up creating a copy of the wrong HTML element. 

Adding one more call to previous fixes the issue.

As a future improvement, we should consider creating a proper id or class for environment variables so we can find them easily instead of walking the DOM like this.

Input:
image

Result:
image

### References
- Jira link: https://zendesk.atlassian.net/browse/SCALE-1566

### Risks
- Low


grosser (approved):
good fun :D

PR #3885, opened 2020-11-22
[k8s] create a configmap of env vars to keep container def clean

craig-day:
The env list in containers can get very large for large projects. This creates large kubernetes objects that contain lots of the same data and those objects can put lots of load on etcd (or whatever you're using). 

This change leverages a Kubernetes feature to create [all key-value pairs in a configmap as container environment variables](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables). By doing this, we can keep the env vars around in one spot and every copy of a container can reference that same spot rather than duplicate the content. 

**TODO:**

- [x] ~try to remove the change of `to_hash` to `to_a` and eliminate the flatmap. This is a big sweeping change that would be nice to avoid.~ This was removed and the `to_hash` was left as-is.
- [x] figure out if there is an order dependency/race here. What happens when the configmap updates before the container spec. Is there a window where a workload could get rescheduled elsewhere running the old image but read in the new configmap? 
- [x] look into versioning these configmaps to alleviate the potential race mentioned.

To address the race/collision problems above, we use the blue/green naming scheme for the configmaps here. This involved some changes to the core release logic, but should be low impact. The most notable change is we will always assign a phase to a release, even if we don't use the phase name in the resource names. This allows us to ensure that the configmaps always flip-flop and the current deploy should never compete with a previous deploy. It will mean we keep exactly two copies of the configmap around which seems reasonable.

### Risks
- Medium. Containers could reference stale/incorrect env if the configmap isn't written correctly and at the right time.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -30,7 +30,7 @@ module Kubernetes
         release.blue_green_color =
           release.previous_succeeded_release&.blue_green_color == "blue" ? "green" : "blue"
       end
-      release.send :build_release_docs, roles if release.valid?
+      release.send(:build_release_docs, roles) if release.valid?
       release
     end
 
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -67,14 +67,6 @@ module Kubernetes
       deploy.previous_succeeded_deploy&.kubernetes_release
     end
 
-    def blue_green?
-      roles.flatten(1).any? { |dgr| dgr.kubernetes_role.blue_green? }
-    end
-
-    def blue_green_color
-      super || "blue"
-    end
-
     private
 
     # Creates a ReleaseDoc per DeployGroupRole
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -67,6 +67,14 @@ module Kubernetes
       deploy.previous_succeeded_deploy&.kubernetes_release
     end
 
+    def blue_green?
+      roles.flatten(1).any? { |dgr| dgr.kubernetes_role.blue_green? }
+    end
+
+    def blue_green_color
+      super || "blue"
+    end
+
     private
 
     # Creates a ReleaseDoc per DeployGroupRole
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -134,6 +134,45 @@ module Kubernetes
       end
     end
 
+    def metadata
+      @metadata ||=
+        Kubernetes::Release.pod_selector(kubernetes_release.id, deploy_group.id, query: false).merge(
+          deploy_id: kubernetes_release.deploy_id,
+          project_id: kubernetes_release.project_id,
+          role_id: kubernetes_role.id,
+          deploy_group: deploy_group.env_value,
+          revision: kubernetes_release.git_sha,
+          tag: kubernetes_release.git_ref
+        )
+    end
+
+    def static_env
+      @static_env ||= begin
+        env = {}
+
+        [:REVISION, :TAG, :DEPLOY_ID, :DEPLOY_GROUP].each do |k|
+          env[k.to_s] = metadata.fetch(k.downcase).to_s
+        end
+
+        if reference_resource
+          [:PROJECT, :ROLE].each do |k|
+            env[k.to_s] = reference_resource.dig_fetch(:metadata, :labels, k.downcase).dup
+          end
+        end
+
+        # name of the cluster
+        env["KUBERNETES_CLUSTER_NAME"] = deploy_group.kubernetes_cluster.name.to_s
+
+        # blue-green phase
+        env["BLUE_GREEN"] = blue_green_color.dup if blue_green_color
+
+        # env from plugins
+        deploy = kubernetes_release.deploy || Deploy.new(project: kubernetes_release.project)
+        plugin_envs = Samson::Hooks.fire(:deploy_env, deploy, deploy_group, resolve_secrets: false, base: env)
+        plugin_envs.compact.inject({}, :merge!)
+      end
+    end
+
     private
 
     def resource_template=(value)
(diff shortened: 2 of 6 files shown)


grosser:
need to version the configmaps or things will get weird but then we also need cleanup ...
could use blue-green naming scheme to avoid endless versions ... for POC just use the hash as version or so and we'll clean it up later if it proves worthwhile

any easy hack could be to keep to_hash and then delete `:extras` from it ... similar to have we have a special `:delete` key

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -516,30 +515,52 @@ def set_env
...
           valueFrom: {fieldRef: {fieldPath: v}}
         }
       end
 
-      env_containers.each do |c|
-        extra = all
-        env = (c[:env] ||= [])
+      all = static_vars + dynamic_vars
+
+      if env_containers.any?(&method(:env_from_config_map?))
```
grosser, line 527: store this as local var and reuse in the `if` below ?

grosser:
overall looks like it could work ... ideally break less tests in the process :D
... don't have a good idea how to handle the configmap cleanup, blue-green where it switches on it being a successful deploy could work, can reuse the existing blue-green logic maybe

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -165,6 +231,49 @@ def disruption_budget_target(deployment)
       end
     end
 
+    def add_env_config_map
+      return unless env_from_config_map?
+      return if env_config_map_exists?
```
thebritican, line 236: maybe i'm just not a rubyist enough but this reads better to me as a generalist...

`return if env_config_map_exists? || !env_from_config_map?`

gregziegan (approved):

plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -24,12 +24,11 @@ def user
...
-        if roles.flatten(1).any? { |dgr| dgr.kubernetes_role.blue_green? }
-          release.blue_green_color = begin
-            release.previous_succeeded_release&.blue_green_color == "blue" ? "green" : "blue"
-          end
-        end
+      release = new(params) do |r|
+        # We always set the blue green color but the template filler might not always use it.
+        # This allows us to support deploy-specific features like env vars from config maps.
+        r.blue_green_color =
+          r.previous_succeeded_release&.blue_green_color == "blue" ? "green" : "blue"
```
grosser, line 31:
```suggestion
        release.blue_green_color =
          release.previous_succeeded_release&.blue_green_color == "blue" ? "green" : "blue"
```

grosser, line 31:
```suggestion
      release = new(params) do |release|
        # We always set the blue green color but the template filler might not always use it.
        # This allows us to support deploy-specific features like env vars from config maps.
        release.blue_green_color =
          release.previous_succeeded_release&.blue_green_color == "blue" ? "green" : "blue"
```

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -94,6 +94,45 @@ def build_selectors
       verification_templates(main_only: false).first.build_selectors
     end
 
+    def metadata
```
grosser, line 97: prefer calling this deploy_metadata or something like that since `metadata` is already used for kubernetes metadata

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -104,19 +143,46 @@ def resource_template=(value)
     # dynamically fill out the templates and store the result
     def store_resource_template
       add_pod_disruption_budget
+      add_env_config_map
```
grosser, line 146: prefer something like `env_config_map = add_env_config_map` and then use `if env_config_map` below to cut down on extra methods/interactions

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -104,19 +143,46 @@ def resource_template=(value)
...
+        opts[:env_config_map] = env_config_map_name if env_from_config_map?
+        TemplateFiller.new(self, resource, **opts).to_hash
       end
     end
 
+    # The resource we want to copy things from, like name, labels, and annotations
+    def reference_resource
+      return @reference_resource if defined?(@reference_resource)
+      @reference_resource = raw_template.find { |r| %w[Deployment StatefulSet].include?(r[:kind]) }
+    end
```
grosser, line 162: this is usually the `primary` resource and there are already methods in here to find it

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -104,19 +143,46 @@ def resource_template=(value)
...
+      reference_resource&.dig(:metadata, :namespace)
+    end
+
+    def reference_labels
+      reference_resource&.dig(:metadata, :labels) || {}
+    end
+
+    def reference_annotations
+      reference_resource&.dig(:metadata, :annotations) || {}
+    end
```
grosser, line 178: prefer to inline all these to reduce noise

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -165,6 +231,49 @@ def disruption_budget_target(deployment)
       end
     end
 
+    def add_env_config_map
+      return unless env_from_config_map?
+      return if env_config_map_exists?
+
+      annotations = reference_annotations.
```
grosser, line 238: should be fine to use `primary` here and leave add_pod_disruption_budget as it was

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -165,6 +231,49 @@ def disruption_budget_target(deployment)
...
+      }
+
+      if ns = reference_namespace
+        cm[:metadata][:namespace] = ns
+      end
+
+      raw_template << cm
+    end
+
+    def env_config_map_exists?
```
grosser, line 264: `exists?` would mean "is created in the cluster", so prefer "_given?"

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -165,6 +231,49 @@ def disruption_budget_target(deployment)
...
+      raw_template.any? { |r| r[:kind] == "ConfigMap" && r.dig(:metadata, :annotations, :"samson/envConfigMap") }
+    end
+
+    def env_from_config_map?
+      !!reference_resource&.dig(:metadata, :annotations, :"samson/env_from_config_map")
+    end
+
+    def env_config_map_name
+      version = kubernetes_release.blue_green_color || "blue"
+      "#{reference_name}-#{version}-env"
```
grosser, line 274:
```suggestion
      color = kubernetes_release.blue_green_color || "blue"
      "#{reference_name}-#{color}-env"
```
also a little sus that we `||` here since it should always be passed in, so maybe don't pass in the default "blue" to not repeat it here

grosser:
extracting the "static_env+metadata is on release doc" part into it's own PR could make this here a lot less noisy

plugins/kubernetes/app/models/kubernetes/release.rb
craig-day: I don't really like shadowing variables here, which is why I left it as a single letter, but I'll work something out.

grosser: ah yeah, did not see that ... `r` is fine then, shadowing in this way seems alright to me too, so do whatever

plugins/kubernetes/app/models/kubernetes/release_doc.rb
craig-day: Most would get inlined twice. I just extracted the logic the PDB was using and reused it for the configmap here.

craig-day:
I extracted part of this out in #3892

plugins/kubernetes/app/models/kubernetes/release_doc.rb
grosser: the configmap logic should use primary and not "Deployment or Statefulset" logic

craig-day:
Extracted another small piece that isn't explicitly the configmap in #3893

craig-day:
Closing in favor of #3896

PR #3891, opened 2020-12-01
allow setting the default docker build method

grosser:
we want to set external builds as the default since local builds don't work most of the time / are not supposed to be used

Screen Shot 2020-12-01 at 11 52 06 AM

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -229,7 +229,7 @@ class Project < ActiveRecord::Base
   end
 
   def docker_image_building_disabled?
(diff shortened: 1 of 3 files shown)


app/models/project.rb
```diff
@@ -229,7 +229,7 @@ def as_json(methods: [], **options)
   end
 
   def docker_image_building_disabled?
-    force_external_build? ? true : docker_image_building_disabled
+    force_external_build? || docker_image_building_disabled
```
grosser, line 232: random cleanup

eatwithforks (approved):

PR #3892, opened 2020-12-02
[k8s] move deploy meta and static env to release doc for reusability

craig-day:
In prep for #3885, this lifts the metadata and static env computation up from the template filler to the release doc. In the future, this will allow us to re-use these objects more easily to generate deployment level constructs, such as configmaps, in the proper location rather than all the way down in the template filler.

### Risks
- Low. K8s templates could be filled incorrectly if the release doc does not contain the same information the template filler has.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -94,7 +94,52 @@ module Kubernetes
       verification_templates(main_only: true).first.build_selectors
     end
 
+    def deploy_metadata
+      @deploy_metadata ||= Kubernetes::Release.
+        pod_selector(kubernetes_release.id, deploy_group.id, query: false).
+        merge(
+          deploy_id: kubernetes_release.deploy_id,
+          project_id: kubernetes_release.project_id,
+          role_id: kubernetes_role.id,
+          deploy_group: deploy_group.env_value,
+          revision: kubernetes_release.git_sha,
+          tag: kubernetes_release.git_ref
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -94,8 +94,53 @@ def build_selectors
...
+        plugin_envs.compact.inject({}, :merge!)
+      end
+    end
+
     private
 
+    def primary_resource
+      return @primary_resource if defined?(@primary_resource)
+      @primary_resource = raw_template.detect { |r| ["Deployment", "StatefulSet"].include? r[:kind] }
+    end
```
grosser, line 142: primary logic is different let's not mix it with "things that need a PDB" logic

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -94,8 +94,53 @@ def build_selectors
...
+
+    def static_env
+      @static_env ||= begin
+        env = {}
+
+        [:REVISION, :TAG, :DEPLOY_ID, :DEPLOY_GROUP].each do |k|
+          env[k.to_s] = deploy_metadata.fetch(k.downcase).to_s
+        end
+
+        if primary_resource
```
grosser, line 118: all resources have these labels, so no need to check if they are primary

craig-day:
@grosser all feedback addressed. Thoughts?

grosser (approved):

PR #3893, opened 2020-12-02
[k8s] always set blue/green on release to give unique deploy refs if …

craig-day:
…needed

In prep for #3885 this always assigns a color to a release, even if we don't use it. This ensures that top level resources like a configmap can always be given a unique name for every deploy. 

### Risks
- Low. Blue/Green naming could sneak into a k8s resource name unintentionally.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release.rb
@@ -25,11 +25,10 @@ module Kubernetes
     def self.build_release_with_docs(params)
       roles = params.delete(:grouped_deploy_group_roles).to_a
       release = new(params) do |release|
-        if roles.flatten(1).any? { |dgr| dgr.kubernetes_role.blue_green? }
-          release.blue_green_color = begin
-            release.previous_succeeded_release&.blue_green_color == "blue" ? "green" : "blue"
-          end
-        end
+        # We always set the blue green color but the template filler might not always use it.
+        # This allows us to support deploy-specific features like env vars from config maps.
+        release.blue_green_color =
+          release.previous_succeeded_release&.blue_green_color == "blue" ? "green" : "blue"
       end
       release.send :build_release_docs, roles if release.valid?
       release
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -19,6 +19,8 @@ module Kubernetes
     attr_reader :previous_resources
     attr_writer :deploy_group_role
 
+    delegate :blue_green?, to: :kubernetes_role
+
     def deploy
       @previous_resources = resources.map(&:resource)
       resources.each(&:deploy)
@@ -79,7 +81,7 @@ module Kubernetes
     end
 
     def blue_green_color
-      kubernetes_release.blue_green_color if kubernetes_role.blue_green?
+      kubernetes_release.blue_green_color if blue_green?
     end
 
     def prerequisite?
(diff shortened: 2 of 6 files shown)


plugins/kubernetes/app/models/kubernetes/release.rb
```diff
@@ -25,11 +25,10 @@ def user
...
       roles = params.delete(:grouped_deploy_group_roles).to_a
       release = new(params) do |release|
-        if roles.flatten(1).any? { |dgr| dgr.kubernetes_role.blue_green? }
-          release.blue_green_color = begin
-            release.previous_succeeded_release&.blue_green_color == "blue" ? "green" : "blue"
-          end
-        end
+        # We always set the blue green color but the template filler might not always use it.
+        # This allows us to support deploy-specific features like env vars from config maps.
+        release.blue_green_color =
```
grosser, line 30: this means we now always set the BLUE_GREEN env var ?
... did you check all other places that use blue_green_color to make sure it does not do anything silly ?

craig-day: Yeah, I need to do some digging, we should probably check the `blue_green?` flag rather than operate on the presence of a color.

craig-day: So, this isn't actually a problem. Because the template filler get's its `blue_green_color` from the release doc

https://github.com/zendesk/samson/blob/2951152daf3f001206f348f6267d7e6aaa9ef1ce/plugins/kubernetes/app/models/kubernetes/template_filler.rb#L570-L573

and the release doc guards the blue green color with the blue_green setting from the role. 

https://github.com/zendesk/samson/blob/2951152daf3f001206f348f6267d7e6aaa9ef1ce/plugins/kubernetes/app/models/kubernetes/release_doc.rb#L81-L83

similarly, the blue green env var pulls its value from the doc, which again is guarded by the role setting. 

https://github.com/zendesk/samson/blob/2951152daf3f001206f348f6267d7e6aaa9ef1ce/plugins/kubernetes/app/models/kubernetes/release_doc.rb#L125-L126

with that said, I would still be happy to define `blue_green?` on the release doc and guard the usage of the color with that instead of presence if we want. Not sure if we want that many changes though.

grosser: guarding the use with `blue_green?` sounds good ... and then we use `blue_green` anyway for env configmaps

grosser (approved):
I guess as long as this does not break env vars or deploy names this is good enough 👍

craig-day:
@grosser I pushed a commit to explicitly check `blue_green?` instead of color presence. Thoughts?

plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -78,8 +78,12 @@ def verification_templates(main_only: false)
       templates.each_with_index.map { |c, i| Kubernetes::TemplateFiller.new(self, c, index: i) }
     end
 
+    def blue_green?
+      kubernetes_role.blue_green?
+    end
+
     def blue_green_color
-      kubernetes_release.blue_green_color if kubernetes_role.blue_green?
+      kubernetes_release.blue_green_color if blue_green?
```
grosser, line 86: would remove the check here so we can use it in the configmap hackery, but good either way

grosser (approved):

plugins/kubernetes/app/models/kubernetes/release_doc.rb
craig-day: I think the configmap will be created within the release_doc similar to the PDB, so we can reference the release color directly. Or I'll remove the check in the configmap PR if necessary?

grosser:
sounds good

PR #3896, opened 2020-12-04
[k8s] add option to source env from a config map rather than container spec

craig-day:
This is the rework of #3885 after merging in some incremental changes.

The env list in containers can get very large for large projects. This creates large kubernetes objects that contain lots of the same data and those objects can put lots of load on etcd (or whatever you're using). 

This change leverages a Kubernetes feature to create [all key-value pairs in a configmap as container environment variables](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables). By doing this, we can keep the env vars around in one spot and every copy of a container can reference that same spot rather than duplicate the content. 

To address race/collision problems, we use the blue/green naming scheme for the configmaps here. This allows us to ensure that the configmaps always flip-flop and the current deploy should never compete with a previous deploy. It will mean we keep exactly two copies of the configmap around which seems reasonable.

### Risks
- Medium. Containers could reference stale/incorrect env if the configmap isn't written correctly and at the right time.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -144,10 +144,11 @@ module Kubernetes
     # dynamically fill out the templates and store the result
     def store_resource_template
       add_pod_disruption_budget
+      env_config_map = create_env_config_map
       counter = Hash.new(-1)
       self.resource_template = raw_template.map do |resource|
         index = (counter[resource.fetch(:kind)] += 1)
-        TemplateFiller.new(self, resource, index: index).to_hash
+        TemplateFiller.new(self, resource, index: index, env_config_map: env_config_map&.dig(:metadata, :name)).to_hash
       end
     end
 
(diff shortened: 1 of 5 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -205,6 +206,48 @@ def disruption_budget_target(deployment)
...
+        },
+        immutable: true,
+        data: static_env
+      }
+
+      if deployment[:metadata].key?(:namespace)
+        cm[:metadata][:namespace] = deployment.dig(:metadata, :namespace)
+      end
+
+      raw_template << cm
```
grosser, line 247: prefer that to happen outside the method for visibility, and ideally not change raw_template but just the returned templates if possible

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -484,7 +485,7 @@ def verify_env
     def set_env
       all = []
 
-      @doc.static_env.each { |k, v| all << {name: k.to_s, value: v.to_s} }
+      @doc.static_env.each { |k, v| all << {name: k.to_s, value: v.to_s} } unless @env_config_map
```
grosser, line 488: call it the `static_env_map` instead ?

grosser (commented):
that looks pretty clean ... only thing suspicious to me is the raw_template modification

craig-day:
I'm dropping this back to draft because I need to think more. This will create the same configmap many times, once for each "role" I think. We really just want one per project.

grosser:
once per role is fine though ... not perfect, but a big step forward

grosser:
alternative would be to do 1 per release ... and the deploy it with the prerequisites ... maybe as a fake separate role

PR #3897, opened 2020-12-05
use kubernetes propagationPolicy to do deletion logic for us

grosser:
previously we manually cleaned up pods to prevent bugs when deleting and creating right after
with propagationPolicy we no longer need to do that since we know that all children are cleaned up when the parent resource is deleted. This is similar to how kubectl waits with the deletion of the parent until all children are deleted too.

### Risks
- Low: risk is leaving pods behind when deleting (verified locally)
- Low: strange things happening because deletion order is not the default order kubectl uses

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -142,8 +142,10 @@ module Kubernetes
         raise "Unable to #{reason} (#{error_location})"
       end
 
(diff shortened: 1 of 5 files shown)


eatwithforks (approved):

uthark (approved):

yizhang-zen (approved):
Any risk?

grosser:
added risks ... already verified locally so doubt they'll hit us 🤞

PR #3904, opened 2020-12-27
fix reloading breaking plugins decorators

grosser:
previously the require_dependency did not get reloaded since it's not in the autoload-path
this lead to errors like `(undefined method `environment_variables' for #<Deploy:0x00007fb519ffb2b8>` when modifying anything in a model


grosser:
@zendesk/compute

yizhang-zen (approved):

PR #3910, opened 2020-12-31
stop people from deleting all tokens

grosser:
Screen Shot 2020-12-31 at 3 25 17 PM


uthark (approved):

adammw:
oof.. would make more sense to have a red button with a js confirm that tells them what it will do/mean, rather than expand that column unnecessarily

adammw:
also doesn't "stop" them, just warns them so pr title is wrong

grosser:
nobody reads confirms, we already learned that the hard way :D

adammw:
bitmoji

PR #3912, opened 2021-01-11
do not fail when comitter had an invalid email

grosser:
`[04:31:21] Finish hook failed: Mail::AddressList can not parse |49699333+dependabot[bot]@users.noreply.github.com|: Only able to parse up to "49699333+dependabot"
`

@zendesk/compute 

### Risks
- Low: jenkins hook failing or not sending to all emails

### DIFF
--- a/plugins/jenkins/app/models/samson/jenkins.rb
+++ b/plugins/jenkins/app/models/samson/jenkins.rb
@@ -250,8 +250,15 @@ module Samson
       end
 
       emails.compact!
-      emails.select! { |e| e.include?('@') }
-      emails.map! { |x| Mail::Address.new(x) }
+      emails.map! do |a|
+        begin
+          Mail::Address.new(a)
+        rescue
+          nil
+        end
+      end
+      emails.compact!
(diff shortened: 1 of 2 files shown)


plugins/jenkins/app/models/samson/jenkins.rb
```diff
@@ -250,8 +250,15 @@ def notify_emails
       end
 
       emails.compact!
-      emails.select! { |e| e.include?('@') }
-      emails.map! { |x| Mail::Address.new(x) }
+      emails.map! do |a|
+        begin
+          Mail::Address.new(a)
```
uthark, line 255: do we need to log this situation?

uthark (approved):

plugins/jenkins/app/models/samson/jenkins.rb
grosser: did not log it before 🤷 
... also nobody checks the logs anyway and it's easy to reproduce if an email does not get sent since we have all the data

PR #3915, opened 2021-01-16
add a deploy status endpoint

craig-day:
Adds a publicly accessible endpoint to get the status of a deploy. 

This is to allow external systems to integrate with Samson and track the status of a deploy. It simply returns the status string in a JSON response, so I felt it was safe to skip all forms of auth so external systems can do this without needing to provide credentials.

### Risks
- Low: This could open the door for a rogue integration to spam the status endpoint.

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -232,7 +232,7 @@ class Deploy < ActiveRecord::Base
   end
 
   def as_json(methods: [])
-    hash = super(methods: [:status, :status_url, :url, :production, :commit] + methods)
+    hash = super(methods: [:status, :url, :production, :commit] + methods)
     hash["summary"] = summary_for_timeline
     hash
   end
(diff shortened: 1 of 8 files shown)


grosser:
how do you get the deploy id if you don't have auth ?

craig-day:
The webhook response. For example, Spinnaker can fire a webhook, then extract a URL from the response and poll it to check status.

app/models/deploy.rb
```diff
@@ -211,6 +211,10 @@ def url
     Rails.application.routes.url_helpers.project_deploy_url(project, self)
   end
 
+  def status_url
```
grosser, line 214: can we not send this back or make it opt-in via `includes`, nobody else needs this

grosser (approved):

craig-day:
@grosser I removed `status_url` from the deploy controller responses and made it opt-in via the `includes` parameter for the webhooks controller. Thoughts?

grosser (approved):

grosser:
it's a little wonky, but so is only returning deploy_ids :)

craig-day:
@grosser the integrations controller tests were overriding the routes so I had to do something a little funky. Thoughts?

grosser (approved):

PR #3917, opened 2021-01-22
increase deletion wait times since we now do foreground deletion

grosser:
takes some time to get rid of all pods from a deployment, so let's wait a bit longer 🤞

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/resource.rb
+++ b/plugins/kubernetes/app/models/kubernetes/resource.rb
@@ -8,6 +8,8 @@ module Kubernetes
   # run an example file through `kubectl create/replace/delete -f test.yml -v8`
   # and see what it does internally ... simple create/update/delete requests or special magic ?
   module Resource
+    DELETE_BACKOFF = [0.0, 0.1, 0.2, 0.5, 1, 2, 4, 8, 16, 32].freeze # seconds
+
     module PatchReplace
       def patch_replace?
         !@delete_resource && !server_side_apply? && exist?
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/resource.rb
```diff
@@ -89,7 +89,7 @@ def revert(previous)
       def delete
         return true unless exist?
         request_delete
-        backoff_wait([0.0, 0.1, 0.2, 0.5, 1, 2, 4, 8, 16], "delete resource") do
+        backoff_wait([0.0, 0.1, 0.2, 0.5, 1, 2, 4, 8, 16, 32], "delete resource") do
```
LiMina, line 92: is this in seconds? (or another unit?)
curious if we have metrics around how long it takes to take down all pods from a deployment (although I'm sure it various a lot from deployment to deployment)

also `sum of waits should be ~30s which is the default delete timeout` comment is longer now, yes?

uthark (approved):

plugins/kubernetes/test/models/kubernetes/resource_test.rb
```diff
@@ -268,7 +268,7 @@ def assert_create_and_delete_requests(**args, &block)
 
       it "fails when deletion fails" do
         assert_request(:delete, url, to_return: {body: "{}"}) do
-          tries = 9
+          tries = 10
```
uthark, line 271: extract array in backoff_wait to a constant and use it's length here?

plugins/kubernetes/app/models/kubernetes/resource.rb
LiMina: Oh, I guess the code comment answered my question about time unit, haha!

PR #3919, opened 2021-01-25
Move from github-markdown to github-markup

russell:
They are very similar gems, but the former isn't supported on newer
versions of OSX.  Basically you'll see an error like.

```
gh-markdown.c:56:29: error: implicitly declaring library function 'isspace' with type 'int (int)' [-Werror,-Wimplicit-function-declaration]
                while (i < lang->size && !isspace(lang->data[i]))
                                          ^
gh-markdown.c:56:29: note: include the header <ctype.h> or explicitly provide a declaration for 'isspace'
```

Since the github-markdown gem is deprecated, we need to change gems.

### Risks
- Low: anything that could go wrong will be covered in tests


grosser (dismissed):

russell:
@grosser i think we could actually trim this down a bit and only use the https://github.com/gjtorikian/commonmarker instead of the github wrapper, they turn on some options but we could do the same instead of using the gem?

https://github.com/github/markup/blob/cd01f9ec87c86ce5a7c70188a74ef40fc4669c5b/lib/github/markup/markdown.rb#L7-L10

Let me know what you'd prefer?

grosser:
it's only used to display the risks, so if it renders a bit wrong that should not be an issue, just make sure it cannot be used for xss (like <script> being escaped etc)

PR #3920, opened 2021-01-25
Return the release reference to the calling client

russell:
With some of our migration work in the deploy process, we want to be
able to trigger some calls to Samson to create releases, but then have
the version name returned so we can perform further tasks on that tag.

I'm using clone in this code because the underlying deploy stage uses hash modifying operations, namely `.delete`.

### Risks
- Low: could cause errors, but will still result in deploys/releases

### DIFF
--- a/app/controllers/integrations/base_controller.rb
+++ b/app/controllers/integrations/base_controller.rb
@@ -53,6 +53,7 @@ class Integrations::BaseController < ApplicationController
       deploy_ids: deploys.map(&:id).compact,
       messages: @recorded_log.to_s
     }
+    json[:release] = release if release
 
     if params[:includes].to_s.split(',').include?('status_urls')
       json[:status_urls] = deploys.map(&:status_url).compact
--- a/test/controllers/integrations/base_controller_test.rb
+++ b/test/controllers/integrations/base_controller_test.rb
@@ -111,10 +111,12 @@ describe Integrations::BaseController do
       result = WebhookRecorder.read(project)
       log = "INFO: Create release for branch [master], service_type [ci], service_name [base_test]: true\n"
 
+      project.reload
       result.fetch(:log).must_equal log
       result.fetch(:response_code).must_equal 200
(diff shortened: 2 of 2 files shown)


davidreuss (approved):

app/controllers/integrations/base_controller.rb
```diff
@@ -32,14 +32,14 @@ def create
...
+      end
     deploys = stages.map do |stage|
-      options =
-        if params[:deploy_branch_with_commit]
-          {reference: branch, commit: commit} # nice display and exact commit
-        else
-          {reference: release&.version || commit} # exact display and commit
-        end
-      deploy_service.deploy(stage, options)
+      deploy_service.deploy(stage, options.clone)
```
grosser, line 42:
```suggestion
      deploy_service.deploy(stage, options.dup)
```

app/controllers/integrations/base_controller.rb
```diff
@@ -51,6 +51,7 @@ def create
 
     json = {
       deploy_ids: deploys.map(&:id).compact,
+      deploy_options: options,
```
grosser, line 54: I was kinda expecting `release: release` ?
... so return something we have a serializer for instead of some internal options hash
... but either way is fine, the part you are interested in is unlikely to change much

grosser (approved):

app/controllers/integrations/base_controller.rb
russell: Thanks!  I'll have a read

app/controllers/integrations/base_controller.rb
russell: Yeah, i wasn't sure what to do.  Since the `release` isn't always there.   But to be honest I would rather do it the way you suggest, so i'll update the PR

davidreuss (approved):
nice 👍

russell:
@grosser I'm merging this now, but i'm open to doing other changes at your request :D

PR #3921, opened 2021-01-28
Remove annoying gray line

dudeuter:
Overrides conflicting style from bootstrap's badge class that was breaking the layout

I don't know how frowned upon inline style is, but it guarantees highest specificity, and that the style won't be picked up by any where else.

### Risks:
* Low: Could potentially shift something else by one pixel

Screen Shot 2021-01-27 at 10 47 46 PM
Screen Shot 2021-01-27 at 10 47 29 PM

### DIFF
--- a/app/views/deploys/show.html.erb
+++ b/app/views/deploys/show.html.erb
@@ -15,7 +15,7 @@
 
   <%= render 'changeset/tab_list' %>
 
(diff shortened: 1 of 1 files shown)


grosser:
replaced by https://github.com/zendesk/samson/pull/3922

FYI make a named branch instead of working from master

PR #3924, opened 2021-01-31
cache external env group reads since they repeat a lot

grosser:
saw lots of slow down in samson once we switched to external env groups, this should help


grosser:
@zendesk/compute

eatwithforks (approved):

PR #3926, opened 2021-02-02
Update lock UI to deal with multiple locks/warnings

adammw:

Update lock UI and validations to deal with multiple locks/warnings nicer. No more single resolve button that only can resolve the first lock, but instead inline the button in the lock on the relevant page (ie you can only unlock on the resource's page and with permission). Also change the text on some of the warnings/locks so their scope is clearer.

Screen Shot 2021-02-03 at 10 34 29 am

Screen Shot 2021-02-03 at 10 15 07 am

### Risks
- Low: could break locks in other weird ways, especially where multiple global locks are involved

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -76,15 +76,16 @@ module ApplicationHelper
   end
 
   # tested via link_to_resource
-  def link_parts_for_resource(resource)
+  def link_parts_for_resource(resource, with_locks: true)
     case resource
-    when Project, DeployGroup, Environment, User, Samson::Secrets::VaultServer then [resource.name, resource]
+    when DeployGroup, User, Samson::Secrets::VaultServer then [resource.name, resource]
+    when Environment then ["Environment #{resource.name}", resource]
     when Command then ["Command ##{resource.id}", resource]
     when UserProjectRole then ["Role for ##{resource.user.name}", resource.user]
-    when Stage
+    when Project, Stage
       name = resource.name
-      name = (resource.lock.warning? ? warning_icon : lock_icon) + " " + name if resource.lock
-      [name, [resource.project, resource]]
+      name = (resource.lock.warning? ? warning_icon : lock_icon) + " " + name if with_locks && resource.lock
+      [name, resource.is_a?(Project) ? resource : [resource.project, resource]]
     when Deploy then ["Deploy ##{resource.id}", [resource.project, resource]]
     when SecretSharingGrant then [resource.key, resource]
     when OutboundWebhook then [resource.url, resource]
@@ -96,8 +97,8 @@ module ApplicationHelper
     end
   end
 
-  def link_to_resource(resource)
-    name, path = link_parts_for_resource(resource)
+  def link_to_resource(resource, with_locks: true)
+    name, path = link_parts_for_resource(resource, with_locks: with_locks)
     if Array(path).any?(&:nil?)
       name
     else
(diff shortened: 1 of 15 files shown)


app/helpers/locks_helper.rb
```diff
@@ -37,7 +37,7 @@ def resource_lock_icon(resource)
 
   def lock_affected(lock)
     if lock.resource_type == "Stage"
-      "stage"
+      "Stage #{lock.resource.name}"
```
grosser, line 40: can we always do type + name to not have to change link_to_resource ?

app/models/lock.rb
```diff
@@ -53,6 +52,7 @@ def remove_expired_locks
     private
 
     def all_cached
+      Rails.cache.clear
```
grosser, line 55: wut ?

grosser:
can you make the buttons smaller so they don't bloat the warning message ?

app/views/locks/_lock.html.erb
```diff
@@ -1,8 +1,20 @@
-<div class="alert alert-warning">
+<div class="alert alert-<%= lock.warning? ? "warning" : "danger" %> clearfix">
+  <% if ((lock.global? && show_resolve == :global) || lock.resource_equal?(show_resolve)) && can?(:write, :locks, show_resolve) %>
+    <% if lock.warning? %>
```
grosser, line 3: do the conditional around the icon/text and reuse the link_to

app/models/lock.rb
adammw: 🙈

grosser:
banners should appear above the breadcrumb consistently ?

app/helpers/application_helper.rb
```diff
@@ -106,7 +107,9 @@ def link_to_resource(resource)
   end
 
   def manual_breadcrumb(items)
-    items.unshift ["Home", root_path]
+    home_text = "Home"
+    home_text = (global_locks.all?(&:warning?) ? warning_icon : lock_icon) + " " + home_text if global_locks.present?
```
grosser, line 111: seems kinda overkill since we have the warning on every page anyway ?

app/helpers/locks_helper.rb
```diff
@@ -26,20 +26,22 @@ def global_locks
...
+    render partial: '/locks/lock', collection: locks, as: :lock, locals: {show_resolve: resource} if locks.any?
   end
 
   def resource_lock_icon(resource)
-    return unless lock = Lock.for_resource(resource).first
-    text = (lock.warning? ? "#{warning_icon} Warning" : "#{lock_icon} Locked")
-    content_tag :span, text.html_safe, class: "label label-warning", title: strip_tags(lock.summary)
+    return unless locks = Lock.for_resource(resource).presence
+    text = (locks.all?(&:warning?) ? "#{warning_icon} Warning" : "#{lock_icon} Locked")
+    text += " (#{locks.count})" if locks.count > 1
```
grosser, line 35: do we really care about the count ?

grosser (approved):

app/helpers/application_helper.rb
adammw: not _every_ page, only those with `render_locks` (which is usually the index pages)

app/helpers/locks_helper.rb
adammw: thought it was worthwhile to distinguish when there's multiple reasons for a stage to be locked

PR #3927, opened 2021-02-02
[ESS-1116] Support pod templates when editing k8s deploy roles

grim-fendango:

This furthers our endeavours on adding native support for PodTemplate resource types. Currently when editing existing pod template resources in the k8s resources tab, it blows up with the following:

```
KeyError: key not found: :spec
1 File "/data/samson/releases/20210126-124035-19961a5/plugins/kubernetes/app/models/kubernetes/role.rb" line 131 in fetch
2 File "/data/samson/releases/20210126-124035-19961a5/plugins/kubernetes/app/models/kubernetes/role.rb" line 131 in defaults
```

This is caused when a `resource.fetch(:spec)` call is made on a pod template that does not have a top level spec property. This PR addresses this problem.

This PR also updates the "does not fail without spec" test case that uses a ConfigMap. Upon closer inspection, this test was returning a false positive as it was not actually testing against a ConfigMap but rather the primary Deployment resource. When switching it over to actually use only the ConfigMap resource properly, it no longer worked as expected as defined in the spec, i.e., it could not load up the resource properly via `resource = role_config_file(...)` and as such was unable return the defaults. This has been changed to assert for nil instead.

### References
- Jira link: https://zendesk.atlassian.net/browse/ESS-1116

### Risks
- Low/Med/High: Worst case scenario, editing k8s deploy group roles in the UI breaks for all resource types. Apparently there are workarounds for this.

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role.rb
@@ -128,12 +128,16 @@ module Kubernetes
     # TODO: support dynamic folders
     def defaults
       return unless resource = role_config_file('HEAD', deploy_group: nil, ignore_errors: true)&.primary
-      spec = resource.fetch(:spec)
-      if resource[:kind] == "Pod"
+      case resource[:kind]
+      when "Pod"
         replicas = 0 # these are one-off tasks most of the time, so we should not count them in totals
+        spec = resource.fetch(:spec)
+      when "PodTemplate"
+        replicas = 1
+        spec = resource.dig(:template, :spec) || {}
       else
-        replicas = spec[:replicas] || 1
-        spec = spec.dig(:template, :spec) || {}
+        replicas = resource.dig(:spec, :replicas) || 1
+        spec = resource.dig(:spec, :template, :spec) || {}
       end
 
       return unless limits = spec.dig(:containers, 0, :resources, :limits)
--- a/plugins/kubernetes/test/models/kubernetes/role_test.rb
+++ b/plugins/kubernetes/test/models/kubernetes/role_test.rb
@@ -32,6 +32,16 @@ describe Kubernetes::Role do
       spec: {containers: [{name: 'foo', resources: {limits: {cpu: '0.5', memory: '300Mi'}}}]}
     }
   end
+  let(:pod_template) do
+    {
+      kind: 'PodTemplate',
+      apiVersion: 'v1',
+      metadata: {name: 'foo', labels: {role: 'migrate', project: 'bar'}},
+      template: {
+        spec: {containers: [{name: 'foo', resources: {limits: {cpu: '0.5', memory: '300Mi'}}}]}
+      }
+    }
+  end
   let(:config_content) do
     YAML.load_stream(read_kubernetes_sample_file('kubernetes_deployment.yml'))
   end
@@ -332,7 +342,7 @@ describe Kubernetes::Role do
       role.defaults[:replicas].must_equal 1
     end
 
-    it "does not fail without spec" do
+    it "does not fail without primary kind" do
       labels = {project: 'some-project', role: 'some-role'}
       map = {
         kind: 'ConfigMap',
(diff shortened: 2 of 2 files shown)


plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -128,9 +128,12 @@ def seed_service_name(config_file)
     # TODO: support dynamic folders
     def defaults
       return unless resource = role_config_file('HEAD', deploy_group: nil, ignore_errors: true)&.primary
-      spec = resource.fetch(:spec)
+      spec = resource.fetch(:spec, {})
       if resource[:kind] == "Pod"
         replicas = 0 # these are one-off tasks most of the time, so we should not count them in totals
+      elsif resource[:kind] == "PodTemplate"
```
grosser, line 134:
```suggestion
      if resource[:kind] == "Pod"
        replicas = 0 # these are one-off tasks most of the time, so we should not count them in totals
        spec = resource.fetch(:spec)
      elsif resource[:kind] == "PodTemplate"
```

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -128,9 +128,12 @@ def seed_service_name(config_file)
...
+        replicas = 1
+        spec = resource.dig(:template, :spec) || {}
       else
         replicas = spec[:replicas] || 1
         spec = spec.dig(:template, :spec) || {}
```
grosser, line 139:
```suggestion
      if resource[:kind] == "Pod"
        replicas = 0 # these are one-off tasks most of the time, so we should not count them in totals
        spec = resource.fetch(:spec)
      elsif resource[:kind] == "PodTemplate"
        replicas = 1
        spec = resource.dig(:template, :spec) || {}
      else
        replicas = spec[:replicas] || 1
        spec = resource.dig(:spec, :template, :spec) || {}
```

plugins/kubernetes/test/models/kubernetes/role_test.rb
```diff
@@ -332,22 +342,14 @@ def commit
       role.defaults[:replicas].must_equal 1
     end
 
-    it "does not fail without spec" do
```
grosser, line 335: I think this is still a valid case, but maybe rewrite as `it "does not fail without primary kind" do`

plugins/kubernetes/app/models/kubernetes/role.rb
grosser: also this duplicate the logic used to find out if it's `primary`, so best reuse that here ?

grosser (commented):
makes sense, just some minor cleanup

plugins/kubernetes/app/models/kubernetes/role.rb
grim-fendango: Logically yes. Semantically? Not quite sure. This isn't my area of expertise and my knowledge is very limited. I'm happy for you suggest a variant that utilises primary instead, as I don't see how it can be used.

grim-fendango: I'm sorry but what you have suggested does not work. In the case of deployments and other resources with embedded templates, we need to retrieve the the replicas from the first instance of spec, before diving deeper into the template spec. The implementation outlined in this PR accounts for this.

I'll keep digging into the primary thing. Sorry, kind of new to this space.

grosser: ah yes, they are different specs ... might need some better variable nameing there too
don't worry about the primary, I can clean that up after this PR is done

grim-fendango: Updated your logic to work as intended. Thanks for the suggestion ❤️

grim-fendango: I'll let you handle the primary refactor later on if needed.

grosser (approved):

PR #3933, opened 2021-02-23
allow background deletion when there are no pods to avoid getting stu…

grosser:
…ck in mystery error

```
RuntimeError: Unable to delete resource (Deployment provisioning-app-server pod998 Pod998)
File "plugins/kubernetes/app/models/kubernetes/resource.rb" line 144 in backoff_wait
File "plugins/kubernetes/app/models/kubernetes/resource.rb" line 94 in delete
```


yizhang-zen (approved):
🙇‍♀️

eatwithforks (approved):

PR #3936, opened 2021-02-25
show parsing errors in seeding stage so they can be fixed early

grosser:
I often get asked why seeding did not work in xyz case and mostly the answer is some validation failing
and instead of reproducing that locally to debug let's just show it in the UI so debugging is easier

@zendesk/compute


eatwithforks (approved):
🎉

PR #3939, opened 2021-03-03
nicely show github action statuses that do not have a description

grosser:
Screen Shot 2021-03-03 at 12 03 33 PM

### DIFF
--- a/app/assets/javascripts/ref_status_typeahead.js
+++ b/app/assets/javascripts/ref_status_typeahead.js
@@ -73,7 +73,8 @@ function refStatusTypeahead(options){
       var item = $("<li>");
       $ref_problem_list.append(item);
       // State and status comes from GitHub or Samson
-      item.html(status.state + ": " + status.description);
+      var description = (status.description.toString() === "" ? status.context : status.description);
+      item.html(status.state + ": " + description);
       if(status.target_url) {
         item.append(' <a href="' + status.target_url + '">details</a>');
       }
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/ref_status_typeahead.js
```diff
@@ -73,7 +73,7 @@ function refStatusTypeahead(options){
       var item = $("<li>");
       $ref_problem_list.append(item);
       // State and status comes from GitHub or Samson
-      item.html(status.state + ": " + status.description);
+      item.html(status.state + ": " + (status.description || status.context));
```
adammw, line 76:
```suggestion
      item.html(status.state + ": " + (status.description.presence || status.context));
```

adammw (approved):

app/assets/javascripts/ref_status_typeahead.js
adammw: also should probably be using `item.text()` to avoid XSS

grosser: we have lots of users that need html:
```
Deploy.first(100).uniq { |d| d.project_id }.map { |d| CommitStatus.new(d.project, d.reference, stage: d.stage).statuses }
```
so added sanitization in the controller

grosser: Screen Shot 2021-03-04 at 10 41 44 AM

PR #3944, opened 2021-04-13
Improve risks parsing to allow for emojis to the left of header

PierceAndy:
### Description
When attempting to add simple emojis to the left of section headers to improve readability at a glance:
image

We noticed that the risks section failed to be parsed in samson:
Screenshot 2021-04-14 at 1 45 57 PM

This PR allows emojis to be added to the left of section headers without otherwise affecting current behaviour too much

### Risks
- Low: May break risks section parsing

### DIFF
--- a/app/models/changeset/pull_request.rb
+++ b/app/models/changeset/pull_request.rb
@@ -161,21 +161,7 @@ class Changeset::PullRequest
 
   def section_content(section_title, text)
     # ### Risks or Risks followed by === / ---
-    emoji_regex =
-      '(?:'\
-        '\p{RI}\p{RI}'\
-        '|\p{Emoji}'\
-          '(?:\p{EmojiModifier}'\
-          '|\uFE0F\u20E3?'\
-          '|[\uE0020-\uE007E]+\uE007F)?'\
-          '(?:\u200D\p{Emoji}'\
-            '(?:\p{EmojiModifier}'\
-            '|\uFE0F\u20E3?'\
-            '|[\uE0020-\uE007E]+\uE007F)?'\
-          ')*'\
-      ')' # translated from unicode standard: https://unicode.org/reports/tr51/#EBNF_and_Regex
-    desired_header_regexp = "^(?:\\s*#+\\s*#{emoji_regex}*\\s*#{section_title}.*"\
-      "|\\s*#{emoji_regex}*\\s*#{section_title}.*\\n\\s*(?:-{2,}|={2,}))\\n"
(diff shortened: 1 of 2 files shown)


test/models/changeset/pull_request_test.rb
```diff
@@ -6,7 +6,10 @@
 describe Changeset::PullRequest do
   def add_risks
     body.replace(+<<~BODY)
-      # Risks
+      # ⚠️ Risks
```
grosser, line 9: please add a new test instead of modifying an old one

app/models/changeset/pull_request.rb
```diff
@@ -161,7 +161,21 @@ def message
...
+          '|\uFE0F\u20E3?'\
+          '|[\uE0020-\uE007E]+\uE007F)?'\
+          '(?:\u200D\p{Emoji}'\
+            '(?:\p{EmojiModifier}'\
+            '|\uFE0F\u20E3?'\
+            '|[\uE0020-\uE007E]+\uE007F)?'\
+          ')*'\
+      ')' # translated from unicode standard: https://unicode.org/reports/tr51/#EBNF_and_Regex
+    desired_header_regexp = "^(?:\\s*#+\\s*#{emoji_regex}*\\s*#{section_title}.*"\
+      "|\\s*#{emoji_regex}*\\s*#{section_title}.*\\n\\s*(?:-{2,}|={2,}))\\n"
```
grosser, line 178: this is a bit complex ... can we simplify by removing all the non-word chars left of the heading and just keep the old code ?
```
text.gsub!(/(^#+\s*)\W+/, "\\1")
```

joemeister:
How is this tested?

app/models/changeset/pull_request.rb
PierceAndy: This bit of code fails this test case: https://github.com/zendesk/samson/pull/3944/files#diff-a1fa917b3ccda77ed24658b564d3e06572c84cf95d080491faa7f39a839aaba2R477-R484

Let me think of an alternative solution 🤔

PierceAndy: What about updating 
`desired_header_regexp = "^(?:\\s*#+\\s*#{section_title}.*|\\s*#{section_title}.*\\n\\s*(?:-{2,}|={2,}))\\n"`
to
`desired_header_regexp = "^(?:\\s*#+\\s*\\W*\\s*#{section_title}.*|\\s*\\W*\\s*#{section_title}.*\\n\\s*(?:-{2,}|={2,}))\\n"`

PierceAndy:
> How is this tested?

there is https://samsontest.zende.sk/, but I'm still figuring out how to test deploys in that

PierceAndy:
> How is this tested?

there's an internal staging samson that can be used. I'm currently looking for an existing project that I can piggyback off to test this branch on

hyperair (approved):

PR #3945, opened 2021-04-13
Increase MAX_LIMITS_TO_REQUESTS_RATIO

adammw:

Increase MAX_LIMITS_TO_REQUESTS_RATIO as we occasionally need for pods with small cpu request (e.g. 0.1) to have larger limit (e.g. 3 cpu).

### Risks
- Low: increased chance of misconfiguring limits too high


grosser (commented):
let's just remove the feature then ... 1-to-100 is not really catching anything anymore

PR #3947, opened 2021-04-16
allow users to request associated project locks

anthonywoo:
Add `locks` as an `?includes` option when requesting project info via the API

### Risks
- Low


grosser (approved):
FYI there can be multiple locks, but the association does not exist ... so this should be good but there might be edge-cases where you delete the lock and then the project is still locked

Issue #3948, opened 2021-04-18
assets:precompile not generate env plugin assets


Willis0826:
# Issue
I ran `RAILS_ENV=production PRECOMPILE=1 PLUGINS=all bundle exec rake assets:precompile --trace` with `v3290` can not generate `env` plugin assets under `public/assets` folder. Also, there are no error while running the command.

However, I can verify the `v3260` work as expected.

# Expected
I should be able to see `env` plugin assets after ran the precompile command.

# How to reproduce
1. `git checkout v3290`
2. `bundle`
3. `RAILS_ENV=production PRECOMPILE=1 PLUGINS=all bundle exec rake assets:precompile --trace`
4. `ls public/assets`
5. There is no `env` plugin

grosser:
yeah, the add_assets_to_precompile hook is still getting called, but the fils are not rendered :/

grosser:
found it ... it's an ordering problem :/

grosser:
https://github.com/zendesk/samson/pull/3949

Willis0826:
Thanks for the quick response and fix : )

PR #3949, opened 2021-04-18
fix assets precompile that broke during rails 6.1 upgrade

grosser:
fixes https://github.com/zendesk/samson/issues/3948#
nobody noticed because old assets stayed around

/cc @zendesk/compute

### DIFF
--- a/lib/samson/hooks.rb
+++ b/lib/samson/hooks.rb
@@ -110,10 +110,6 @@ module Samson
         end
       end
 
-      def add_assets_to_precompile
-        engine.config.assets.precompile += ["#{name}/application.css", "#{name}/application.js"]
-      end
-
       def engine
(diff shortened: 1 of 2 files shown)


adammw:
worth adding a CI test to ensure precompile works and includes expected plugin files?

lib/samson/hooks.rb
```diff
@@ -110,10 +110,6 @@ def add_decorators
         end
       end
 
-      def add_assets_to_precompile
```
adammw, line 113: why is this no longer required? rails upgrade does it automatically?

adammw (approved):

lib/samson/hooks.rb
grosser: yeah, it's automatic, either because of the new version or because it was in the wrong position before

grosser:
idk how to write that test, so not going to invest more effort

PR #3950, opened 2021-04-20
kubernetes increased min memory

grosser:
@zendesk/compute 
```
  Warning  Failed     4s (x3 over 20s)  kubelet            Error: Error response from daemon: Minimum memory limit allowed is 6MB
```

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role.rb
@@ -20,7 +20,7 @@ module Kubernetes
       "" => 1,
       'm' => 0.001
     }.freeze
-    MIN_MEMORY = 4
+    MIN_MEMORY = 6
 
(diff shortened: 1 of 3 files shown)


eatwithforks (approved):

plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -20,7 +20,7 @@ class Role < ActiveRecord::Base
       "" => 1,
       'm' => 0.001
     }.freeze
-    MIN_MEMORY = 4
+    MIN_MEMORY = 6
```
uthark, line 23: Wish we would ignore it for the roles that have only CRDs.

uthark:
Limit comes from here: https://github.com/moby/moby/blob/master/daemon/daemon_unix.go#L71-L72

uthark (approved):

PR #3956, opened 2021-05-07
Allow Kubernetes connectivity validation to be public

lara:
@zendesk/compute

## Description
- Change `connection_valid?` to be a public method since it will be used externally.
- Update setup docs.

### Risks
- Low: Method made public to be used externally, should not cause any issues.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -23,7 +23,7 @@ Complicated to start, but easy to develop from.
 ```bash
 # install the database you want to use, samson needs mysql, postgresql, or sqlite
 sudo apt-get install mysql-dev pg-dev nodejs
-brew install postgresql sqlite mysql
+brew install postgresql sqlite mysql nodejs
 
 bin/setup # Run the setup script to use the test credentials.
 ./bin/rails s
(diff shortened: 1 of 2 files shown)


grosser (approved):
merge when plugin is merged

plugins/kubernetes/app/models/kubernetes/cluster.rb
```diff
@@ -11,10 +11,10 @@ class Cluster < ActiveRecord::Base
     attr_encrypted :client_key
 
     has_many :cluster_deploy_groups,
-      class_name: 'Kubernetes::ClusterDeployGroup',
-      foreign_key: :kubernetes_cluster_id,
-      dependent: nil,
-      inverse_of: :cluster
+    class_name: 'Kubernetes::ClusterDeployGroup',
```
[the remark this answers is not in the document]
lara: not quiet sure what happened there..

docs/setup.md
```diff
@@ -23,7 +23,7 @@ Complicated to start, but easy to develop from.
 ```bash
 # install the database you want to use, samson needs mysql, postgresql, or sqlite
 sudo apt-get install mysql-dev pg-dev nodejs
-brew install postgresql sqlite mysql
+brew install postgresql sqlite mysql nodejs
```
uthark, line 26: why does it need nodejs?

uthark (approved):

yizhang-zen (approved):

docs/setup.md
uthark: So, it is a dependency for new rails?

lara: I know Rails has a JS runtime dependency but I'm assuming since we don't have `execjs`, `nodejs`, or `therubyracer` in the Gemfile, Rails tries to default to system `nodejs` (which probably many has it installed for other reasons).

PR #3958, opened 2021-05-12
Report if jobs are executed with Kubernetes

lara:
@zendesk/compute 

Add tags for metrics to indicate whether jobs are executed with Kubernetes, which will help create separation of concern when building SLOs.

### DIFF
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -65,6 +65,10 @@ ActiveSupport::Notifications.subscribe("job_status.samson") do |*, payload|
     "stage:#{payload.fetch(:stage)}",
   ]
 
+  # report if things were run with kubernetes
+  kubernetes = payload.fetch(:kubernetes)
+  tags << "kubernetes:#{kubernetes}" unless kubernetes.nil?
+
   payload.fetch(:cycle_time).each do |key, value|
     Samson.statsd.timing "jobs.deploy.cycle_time.#{key}", value * 1000, tags: tags
   end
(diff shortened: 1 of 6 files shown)


test/models/job_execution_test.rb
```diff
@@ -22,7 +22,7 @@ def last_line_of_output
   let(:pod2) { deploy_groups(:pod2) }
   let(:deploy_group_hack) { DeployGroup.create!(name: ';|sudo make-sandwich /', environment: pod1.environment) }
   let(:project) { Project.create!(name: 'duck', repository_url: repo_temp_dir) }
-  let(:stage) { Stage.create!(name: 'stage4', project: project, deploy_groups: [pod1, pod2, deploy_group_hack]) }
+  let(:stage) { Stage.create!(name: 'stage4', project: project, deploy_groups: [pod1, pod2, deploy_group_hack], kubernetes: false) }
```
grosser, line 25: false is the default

config/initializers/statsd.rb
```diff
@@ -61,6 +65,10 @@ class << Samson
     "stage:#{payload.fetch(:stage)}",
   ]
 
+  # report if things were run with kubernetes
+  kubernetes = payload.fetch(:kubernetes)
+  tags << "kubernetes:#{kubernetes}" unless kubernetes.nil?
```
grosser, line 70:
```suggestion
  tags << "kubernetes:#{!!payload.fetch(:kubernetes)}"
```

grosser:
we might not need this if the number of regular deploys is very small in comparison to kubernetes deploys, so let's check that first

config/initializers/statsd.rb
lara: want to keep it consistent with the above tag, but also not send any kubernetes-related tag if it's for a build (which is when `kubernetes` would be `nil`)

lara:
> we might not need this if the number of regular deploys is very small in comparison to kubernetes deploys, so let's check that first

Right now, the SLOs would be very off. This should help with building better ones and also can be used to construct any new SLI.

grosser (approved):

PR #3962, opened 2021-05-21
bump nokogiri for vuln

grosser:
@zendesk/compute 

replaces https://github.com/zendesk/samson/pull/3961


adammw (approved):

Gemfile.lock
```diff
@@ -409,10 +409,10 @@ GEM
     netrc (0.11.0)
     newrelic_rpm (6.7.0.359)
     nio4r (2.5.7)
-    nokogiri (1.11.3)
+    nokogiri (1.11.5)
       mini_portile2 (~> 2.5.0)
       racc (~> 1.4)
-    nokogiri (1.11.3-x86_64-darwin)
+    nokogiri (1.11.5-x86_64-darwin)
```
adammw, line 415: why do we lock the darwin version of this? shouldn't it depend on the OS / probably be the linux version for docker?

grosser: yeah, can lock both, but it seems to work this way 🤷

PR #3964, opened 2021-06-23
Changes lockable to have many locks

wa-ty:
Something that is lockable can have more than one lock. When getting `?includes=lock` currently one lock is returned, this allows for all locks associated with a lockable to be returned.

### DIFF
--- a/app/models/concerns/lockable.rb
+++ b/app/models/concerns/lockable.rb
@@ -4,7 +4,8 @@ module Lockable
   extend ActiveSupport::Concern
 
   included do
-    has_many :lock, as: :resource, dependent: :destroy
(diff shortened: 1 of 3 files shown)


app/models/concerns/lockable.rb
```diff
@@ -4,7 +4,7 @@ module Lockable
   extend ActiveSupport::Concern
 
   included do
-    has_one :lock, as: :resource, dependent: :destroy
+    has_many :lock, as: :resource, dependent: :destroy
```
grosser, line 7: this should be locks and lock should be marked deprecated but kept around for at least the api

grosser (approved):

PR #3967, opened 2021-07-15
also force-update when switching from nodeport to clusterip

grosser:
force update did not trigger on `spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'`

### Risks
- None


eatwithforks (approved):

uthark:
This would force service to be recreated, which can lead to 30 seconds blips during upgrades.

The proper solution should be generating PATCH that would switch Type to NodePort and remove generated `nodePorts` from the Service spec.

grosser:
yes, but that's a lot more work
this is a opt-in to make rollout to kubernetes <1.20 smoother

PR #3970, opened 2021-07-28
Add support for wildcard key for external env var groups

adammw:

Add support for wildcard key for external env var groups.

Rationale is that it is cumbersome to keep having to add each new deploy group to these external env var files when the value is constant across all of them (an example is where the STATSD env var host is always "localhost" or similar).

cc @zendesk/config @zendesk/samson 

### References
- Jira link: 

### Risks
- Low: could make env var group lookup more complex

### DIFF
--- a/plugins/env/app/models/environment_variable.rb
+++ b/plugins/env/app/models/environment_variable.rb
@@ -78,7 +78,7 @@ class EnvironmentVariable < ActiveRecord::Base
         all_groups = Rails.cache.fetch("env-#{group.url}-read", expires_in: EXTERNAL_VARIABLE_CACHE_DURATION) do
           group.read
         end
-        all_groups[deploy_group.permalink]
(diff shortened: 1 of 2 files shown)


plugins/env/app/models/environment_variable.rb
```diff
@@ -78,7 +78,7 @@ def env_vars_from_external_groups(project, deploy_group)
         all_groups = Rails.cache.fetch("env-#{group.url}-read", expires_in: EXTERNAL_VARIABLE_CACHE_DURATION) do
           group.read
         end
-        all_groups[deploy_group.permalink]
+        all_groups[deploy_group.permalink] || all_groups["*"]
```
grosser, line 81: app/models/environment.rb uses `[["All", nil]]` so would be nice if we can use any of those

grosser (approved):
great QOL change

PR #3975, opened 2021-09-13
Create new task to remove deploy groups from stages given an owner

lara:
@zendesk/compute 

- Add confirmation step that allows user to proceed with proposed actions
- Optionally delete emptied stages

### DIFF
--- a/lib/tasks/maintenance.rake
+++ b/lib/tasks/maintenance.rake
@@ -29,7 +29,8 @@ namespace :maintenance do
         puts "#{project.name} FROM and TO are the same"
       else
         empty = (from_stage.deploy_groups == [move_group])
-        puts "#{project.name} to be moved from:#{from_stage&.name}#{" (+ delete stage)" if empty} to:#{to_stage&.name}"
+        puts "#{project.name} to be moved from:#{from_stage&.name}" \
+          "#{" (+ delete stage)" if empty && delete_empty} to:#{to_stage&.name}"
         true
       end
     end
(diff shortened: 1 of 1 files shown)


lib/tasks/maintenance.rake
```diff
@@ -43,4 +43,38 @@ namespace :maintenance do
       from_stage.destroy! if delete_empty && from_stage.deploy_groups.empty?
     end
   end
+
+  # FROM and TO handles multiple deploy groups to not mess with projects that have each deploy group in a single stage.
```
grosser, line 47: wut ?
... we want to mess with these too

lib/tasks/maintenance.rake
```diff
@@ -43,4 +43,38 @@ namespace :maintenance do
...
+  desc "delete deploy groups across stages for all projects. OWNER= DELETE=dg1,dg2 [DELETE_EMPTY=true]"
```
grosser, line 48: we usually remove just 1 ... and even if we remove 2 I'd rather run it twice
```suggestion
  desc "delete deploy groups across stages for all projects. OWNER= DELETE=dg1 [DELETE_EMPTY=true]"
```

lib/tasks/maintenance.rake
```diff
@@ -43,4 +43,38 @@ namespace :maintenance do
...
+  task delete_deploy_group: :environment do
+    projects = Project.where(owner: ENV.fetch("OWNER"))
+    delete = ENV.fetch("DELETE").split(",")
+    delete_group = DeployGroup.find_by_permalink!(delete)
```
grosser, line 52: find by array does not work
```suggestion
    delete = ENV.fetch("DELETE")
    delete_group = DeployGroup.find_by_permalink!(delete)
```

lib/tasks/maintenance.rake
```diff
@@ -43,4 +43,38 @@ namespace :maintenance do
...
+    delete_empty = (ENV["DELETE_EMPTY"] == "true")
+
+    puts "found #{projects.size} projects"
+
+    actions = projects.map do |project|
+      from_stage = project.stages.detect { |stage| delete & stage.deploy_groups.map(&:permalink) == delete }
```
grosser, line 58:
```suggestion
      from_stage = project.stages.detect { |stage| stage.deploy_groups.map(&:permalink).include? delete }
```

lib/tasks/maintenance.rake
```diff
@@ -43,4 +43,38 @@ namespace :maintenance do
...
+      [project, from_stage]
+    end
+
+    actions.select! do |project, from_stage|
+      if !from_stage
+        puts "#{project.name} Unable to find DELETE"
+      else
+        empty = (from_stage.deploy_groups == [delete_group])
+        puts "#{project.name} to be removed from:#{from_stage&.name}#{" (+ delete stage)" if empty}}"
```
grosser, line 67:
```suggestion
        puts "#{project.name} to be removed from:#{from_stage&.name}#{" (+ delete stage)" if empty && delete_empty}}"
```

lib/tasks/maintenance.rake
```diff
@@ -43,4 +43,38 @@ namespace :maintenance do
...
+    puts "found #{projects.size} projects"
+
+    actions = projects.map do |project|
+      from_stage = project.stages.detect { |stage| delete & stage.deploy_groups.map(&:permalink) == delete }
+      [project, from_stage]
+    end
+
+    actions.select! do |project, from_stage|
+      if !from_stage
+        puts "#{project.name} Unable to find DELETE"
```
uthark, line 64:
```suggestion
        puts "#{project.name} Unable to find DELETE"
        false 
```

make it explicit?

lib/tasks/maintenance.rake
grosser: ... doing the logic twice to decide if we should delete is kinda meh anyway, better to add a true/false flag to the `actions` ?

lib/tasks/maintenance.rake
uthark, line 52: if we pass multiple with comma, how will it fork? IS delete_group would be an array then?

lib/tasks/maintenance.rake
lara: copy pasta from other task

lib/tasks/maintenance.rake
lara: yeah I can see how that's not a good idea `alias watch_the_world_burn="rake maintenance:delete_deploy_groups DELETE=pod17,pod18,pod19,pod20"` :D

lib/tasks/maintenance.rake
lara: it wouldn't have but we'll do one group at a time for safety anyways

grosser (approved):

PR #3976, opened 2021-09-23
Add wait time on retries

zendesk-mattlefevre:

## Description
When accessing a vault through the vault client, 3 retries without any wait time between them can sometimes cause rate-limiting problems for larger deployments. By upping the retries and adding a short wait time for those retries, this should allow larger deployments to successfully access secrets without adding an unnecessary burden on smaller deploys. 

### Risks
- Low/Med/High: Low: While accessing the vault may take slightly longer in some cases, this time delay is intentional. My only real concern is that we may actually want the wait time to be longer.

### DIFF
--- a/lib/samson/secrets/vault_client_manager.rb
+++ b/lib/samson/secrets/vault_client_manager.rb
@@ -103,7 +103,7 @@ module Samson
       private
 
       def with_retries(&block)
-        Vault.with_retries(Vault::HTTPConnectionError, attempts: 3, &block)
+        Vault.with_retries(Vault::HTTPConnectionError, attempts: 5, wait_time: 0.1, &block)
       end
 
       # - local server for deploy-group specific id
--- a/lib/samson/secrets/vault_client_manager.rb
+++ b/lib/samson/secrets/vault_client_manager.rb
@@ -103,7 +103,7 @@ module Samson
       private
 
       def with_retries(&block)
-        Vault.with_retries(Vault::HTTPConnectionError, attempts: 5, wait_time: 0.1, &block)
+        Vault.with_retries(Vault::HTTPConnectionError, attempts: 5, &block)
       end
 
       # - local server for deploy-group specific id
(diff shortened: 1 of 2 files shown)


grosser (approved):

grosser:
tests have the 3 hardcoded ... so either revert that part and add more time or fix the tests

lib/samson/secrets/vault_client_manager.rb
```diff
@@ -103,7 +103,7 @@ def client(deploy_group_permalink)
       private
 
       def with_retries(&block)
-        Vault.with_retries(Vault::HTTPConnectionError, attempts: 3, &block)
+        Vault.with_retries(Vault::HTTPConnectionError, attempts: 5, wait_time: 0.1, &block)
```
anthonywoo, line 106: this is using this gem? https://github.com/zendesk/vault-ruby#making-requests
`wait_time` isn't mentioned in the docs

grosser: code is here https://github.com/hashicorp/vault-ruby/blob/v0.12.0/lib/vault/client.rb#L412

grosser: and it comes with a builtin backoff ... so yeah this needs to either just bump retries or change `base` option

test/lib/samson/secrets/vault_client_manager_test.rb
```diff
@@ -122,7 +122,7 @@
         :put,
         "http://vault-land.com/v1/auth/token/renew-self",
         to_timeout: [],
-        times: 8 # 2 servers with 1 initial try and 3 re-tries
+        times: 12 # 2 servers with 1 initial try and 3 re-tries
```
anthonywoo, line 125: nit: update the comment

anthonywoo (approved):
minor nit on a test, but lets try this out

test/lib/samson/secrets/vault_client_manager_test.rb
zendesk-mattlefevre: Good catch. Updated.

grosser (approved):

grosser:
lmk if you need help deploying

PR #3984, opened 2021-11-15
support opa crds so we can deploy policies and rules together

grosser:
@zendesk/compute 

testing:
- installed opa via `kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/release-3.5/deploy/gatekeeper.yaml
`
- deployed https://github.com/samson-test-org/example-kubernetes branch `grosser/opa` which includes a new policy and rule

### Risks
- Low

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/release_doc.rb
+++ b/plugins/kubernetes/app/models/kubernetes/release_doc.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 module Kubernetes
   class ReleaseDoc < ActiveRecord::Base
+    CRD_CREATING = {
+      "ConstraintTemplate" => [:spec, :crd] # OPA
+    }.freeze
+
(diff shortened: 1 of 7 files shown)


plugins/kubernetes/app/models/kubernetes/release_doc.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 module Kubernetes
   class ReleaseDoc < ActiveRecord::Base
+    CRD_CREATING = {
+      "ConstraintTemplate" => [:spec, :crd] # OPA
```
uthark, line 5: can we do it generic?

grosser:
not really, we don't know that all "crd" keys are crds, also nobody else
needed it yet

eatwithforks (approved):

PR #3989, opened 2022-01-25
Show error when no deploy groups set

adammw:

Show error when no deploy groups set.
Previously this was triggering a nil exception.

### Risks
- Low: could break Kubernetes deployments


grosser (approved):
✨

uthark (approved):

lara (approved):

PR #3994, opened 2022-02-10
feat(changeset): change how recover Pull Request ID

benjamingarcia:
As we could add the Pull Request ID in any place of the commit message,
we would like to change how we parse the commit message. The only
mandatory thing is to have parenthesis around it ex: (#12347).
We also remove the commit message truncation to avoid ignoring PR number
if located more than 80 characters away (at the end of the first line or
in the next lines)

* [x] Add description
* [ ] Add screenshots when changing the UI
* [ ] Add unit tests

### References
- Jira link: 

### Risks
- Med

### DIFF
--- a/app/models/changeset/commit.rb
+++ b/app/models/changeset/commit.rb
@@ -23,6 +23,10 @@ class Changeset::Commit
     @author ||= Changeset::GithubUser.new(@data.author) if @data.author
   end
 
+  def summary
+    summary = @data.commit.message.split("\n").first
+  end
+
   def sha
     @data.sha
   end
@@ -33,8 +37,7 @@ class Changeset::Commit
 
   # @return [Integer, NilClass]
   def pull_request_number
-    commit_message = @data.commit.message
-    if number = commit_message[PULL_REQUEST_MERGE_MESSAGE, 1] || commit_message[PULL_REQUEST_SQUASH_MESSAGE, 1]
+    if number = summary[PULL_REQUEST_MERGE_MESSAGE, 1] || summary[PULL_REQUEST_SQUASH_MESSAGE, 1]
       Integer(number)
     end
   end
--- a/app/models/changeset/commit.rb
+++ b/app/models/changeset/commit.rb
@@ -24,7 +24,11 @@ class Changeset::Commit
   end
 
   def summary
-    summary = @data.commit.message.split("\n").first
+    summary_long.truncate(80)
+  end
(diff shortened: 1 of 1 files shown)


app/models/changeset/commit.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 class Changeset::Commit
-  PULL_REQUEST_MERGE_MESSAGE = /\AMerge pull request #(\d+)/.freeze
-  PULL_REQUEST_SQUASH_MESSAGE = /\A.*\(#(\d+)\)$/.freeze
+  PULL_REQUEST_MESSAGE = /\A.*\(#(\d+)\)$/.freeze
```
grosser, line 3: no longer matches `pull request #123`

app/models/changeset/commit.rb
```diff
@@ -38,7 +32,7 @@ def short_sha
 
   # @return [Integer, NilClass]
   def pull_request_number
-    if number = summary[PULL_REQUEST_MERGE_MESSAGE, 1] || summary[PULL_REQUEST_SQUASH_MESSAGE, 1]
+    if number = @data.commit.message[PULL_REQUEST_MESSAGE, 1]
```
grosser, line 35: truncate was weird, but we should match in the first line

app/models/changeset/commit.rb
pfiorentino-zendesk: True 😊
Should be changed 👍

pfiorentino-zendesk: Fixed 😉

Thanks a lot for your vigilance 🙏

app/models/changeset/commit.rb
pfiorentino-zendesk: Yup, you're right 😊
Sorted now 👍

app/models/changeset/commit.rb
```diff
@@ -25,7 +25,6 @@ def author
 
   def summary
     summary = @data.commit.message.split("\n").first
-    summary.truncate(80)
   end
```
grosser, line 28: need to check where that summary is used ... could either change everything to `summary.truncate(80)` or extract a `summary_long` to use internally

benjamingarcia: ok, `summary_long` created 🙏🏼

grosser (approved):

PR #3995, opened 2022-02-15
fix nils blowing up the new markdown

grosser:
TypeError: text must be a String; got a NilClass!

### Risks
- Low


test/helpers/application_helper_test.rb
```diff
@@ -80,6 +80,11 @@
       result = markdown("Hey ~~FOO~~ bar")
       result.must_equal "<p>Hey <del>FOO</del> bar</p>\n"
     end
+
+    it "supports nil" do
+      result = markdown("")
```
adammw, line 85: shouldn't that be passing nil for a test?

adammw (approved):

test/helpers/application_helper_test.rb
grosser: yes 🤦

PR #3998, opened 2022-02-24
Add deploy groups along the stage info.

uthark:

* Add description
* Add unit tests

### References

- Jira link:

### Risks

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -50,7 +50,8 @@ class Deploy < ActiveRecord::Base
   # job has almost identical code, keep it in sync
   def summary(show_project: false)
     project_name = " #{project&.name}" if show_project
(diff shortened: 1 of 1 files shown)


app/models/deploy.rb
```diff
@@ -50,7 +50,8 @@ def job_execution_queue_name
   # job has almost identical code, keep it in sync
   def summary(show_project: false)
     project_name = " #{project&.name}" if show_project
-    deploy_details = "#{short_reference} to#{project_name} #{stage&.name}"
+    groups = stage.deploy_groups.to_a.presence.map(&:name).join(" ")
```
grosser, line 53:
```suggestion
    groups = stage.deploy_groups.map(&:name).join(" ")
```

grosser: `[].presence -> nil` and `nil.map` -> 💥

grosser (approved):

PR #3999, opened 2022-03-04
allow filtering stages by failed

grosser:
for projects with tons of stages we need to filter by if they are failed to hunt down problems

Screen Shot 2022-03-04 at 2 04 52 PM

### Risks
- Low: project show page breaking

@zendesk/compute


yizhang-zen (approved):
👍

PR #4006, opened 2022-04-12
Improve json support for secrets controller

adammw:

* Improve json support for secrets controller.
Previously requesting JSON either gave 500s due to Missing Template or caused redirects.

### References
- Jira link: 

### Risks
- Low: could expose secrets if filtering logic is incorrect
- Low: JSON support for secrets could still be broken


grosser (approved):
✨

uthark (approved):

PR #4015, opened 2022-06-09
Capability to add a default Pod TopologySpreadConstraints

a7i:

* Add description
* Add unit tests

### References
- Jira link: 

### Risks
- Low/Med/High: Low, it's guarded by an env var

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -221,6 +221,20 @@ and increase the `spec.terminationGracePeriodSeconds` if necessary.
 - Set `KUBERNETES_ADD_PRESTOP=true` to enable
 - Set `KUBERNETES_PRESTOP_SLEEP_DURATION=30` in seconds to override default sleep duration (3 seconds)
 
+### Adding a default Pod TopologySpreadConstraint
+
+In order to control how Pods are spread across the cluster among failure domains, you can set a default 
+[`TopologySpreadConstraint`](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/)
+to help achieve high availability as well as efficient resource utilization.
+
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/README.md
```diff
@@ -221,6 +221,20 @@ and increase the `spec.terminationGracePeriodSeconds` if necessary.
...
 
+### Adding a default Pod TopologySpreadConstraint
+
+In order to control how Pods are spread across the cluster among failure domains, you can set a default 
+[`TopologySpreadConstraint`](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/)
+to help achieve high availability as well as efficient resource utilization.
+
+(will only add if pod `metadata.annotations.samson/podTopologySpread` is not set to `disabled` and a `topologySpreadConstraints` is not defined)
+
+- Set `KUBERNETES_ADD_POD_TOPOLOGY_SPREAD` to enable
```
grosser, line 232: can we do this instead ?
```suggestion
- Set `KUBERNETES_POD_TOPOLOGY_SPREAD={config: here}` to enable
```

grosser (approved):
maybe the initial version could do `if annotation == enable` so we can test it out with a few deployments ?
... idk if you got local kubernetes setup running but it can be hard

uthark (approved):
I'd rather do it in kube-admission.

jklaw90:
@uthark @grosser https://github.com/zendesk/cicd-toolkit/pull/1232 i did it in cicd too, should i remove it?

grosser:
maybe best to have an rfc for this, can steal pro/cons from the slack thread and then we also have a nice "why" for whatever way we decide on

a7i:
Going to close this PR since it can be achieved via admission webhook in a generic fashion rather than adding logic/complexity to Samson.

PR #4016, opened 2022-06-15
Add Sentry plugin

zendesk-mattlefevre:
## Description
Add a plugin to enable use of [Sentry](sentry.io) with the use of the `SENTRY_DSN` environmental variable (along with the existing `PLUGINS` environmental variable)

### References
- Jira link: DPLY-1778

### Risks
- Low: All executable code is behind a Sentry-specific Env var. The only effect of enabling it is logging.


docs/plugins.md
```diff
@@ -26,6 +26,7 @@ Available plugins:
  - [Release Number From CI](https://github.com/redbubble/samson-release-number-from-ci)
  - [NewRelic](https://github.com/zendesk/samson/tree/master/plugins/new_relic)
  - [Pipelined deploys](https://github.com/zendesk/samson/tree/master/plugins/pipelines)
+ - [Rollbar](https://github.com/zendesk/samson/tree/master/plugins/rollbar)
```
grosser, line 29: add sentry ?

plugins/sentry/config/initializers/sentry.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+if dsn = ENV['SENTRY_DSN']
+  Sentry.init do |config|
+    config.dsn = dsn
+    config.breadcrumbs_logger = [:active_support_logger, :http_logger]
+    config.environment = Rails.env
+
+    # We recommend adjusting this value in production:
+    config.traces_sample_rate = 0.5
```
grosser, line 10: prefer 0 (at least for the initial rollout), we had issues with this in other apps

plugins/sentry/config/initializers/sentry.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+end
```
grosser, line 12: can add this too:
```ruby
# make events clickable see https://github.com/getsentry/sentry-ruby/issues/1786
if project = ENV["SENTRY_PROJECT"]
  Sentry::LoggingHelper.prepend(Module.new do
    def log_info(message)
      super(message.sub(
        /(\S+) to Sentry/,
        "https://sentry.io/organizations/#{project}/issues/?query=\\1 to Sentry"
      ))
    end
  end)
end
```

plugins/sentry/samson_sentry.gemspec
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+Gem::Specification.new 'samson_sentry', '0.0.0' do |s|
+  s.summary = 'Samson Sentry plugin'
+  s.authors = ['zendesk-mattlefevre']
+  s.email = ['matthew.lefevre@zendesk.com']
+  s.files = Dir['{app,config,db,lib}/**/*']
+
+  s.add_runtime_dependency 'sentry-rails'
+  s.add_runtime_dependency 'sentry-ruby'
```
grosser, line 9: don't need this line afaik

grosser (approved):

docs/plugins.md
zendesk-mattlefevre: Thank you. 🤦

plugins/sentry/samson_sentry.gemspec
zendesk-mattlefevre: You may be right; just following [what's in the docs](https://docs.sentry.io/platforms/ruby/guides/rails/).

Gemfile.lock
```diff
@@ -164,28 +171,9 @@ PATH
...
-    rails-assets-jstimezonedetect (1.0.5)
-    rails-assets-typeahead.js (0.11.1)
-      rails-assets-jquery (>= 1.7)
-    rails-assets-underscore (1.8.3)
-    rails-assets-x-editable (1.5.1)
-      rails-assets-jquery (>= 1.6)
-
 GEM
   remote: https://rubygems.org/
+  remote: https://rails-assets.org/
```
anthonywoo, line 176: can you regenerate the `Gemfile.lock` when using bundler `2.3.12`?

plugins/sentry/config/initializers/sentry.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+unless ENV['SENTRY_DSN'].nil?
+  Sentry.init do |config|
+    config.dsn = ENV['SENTRY_DSN']
```
grosser, line 5: FYI dry-er
```suggestion
if sentry_dns = ENV['SENTRY_DSN']
  Sentry.init do |config|
    config.dsn = sentry_dns
```

plugins/sentry/config/initializers/sentry.rb
```diff
@@ -0,0 +1,22 @@
...
+    # We recommend adjusting this value in production:
+    config.traces_sample_rate = 0
+  end
+
+  # make events clickable clickable
+  Sentry::LoggingHelper.prepend(Module.new do
+    def log_info(message)
+      super(message.sub(
+        /(\S+) to Sentry/,
+        "https://sentry.io/organizations/#{project}/issues/?query=\\1 to Sentry"
```
grosser, line 18: where is this `project` coming from ?
... need to read from an env var afaik

zendesk-mattlefevre: Missed a line. Added it in.

plugins/sentry/config/initializers/sentry.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    # We recommend adjusting this value in production:
+    config.traces_sample_rate = 0
+  end
+
+  # make events clickable see https://github.com/getsentry/sentry-ruby/issues/1786
+  if project = ENV["SENTRY_PROJECT"]
+    Sentry::LoggingHelper.prepend(Module.new do
+      def log_info(message)
+        super(message.sub( /(\S+) to Sentry/, "https://sentry.io/organizations/#{project}/issues/?query=\\1 to Sentry"
```
grosser, line 17: this won't work because `project` won't be in scope, so need to repeat the `ENV["SENTRY_PROJECT"]` or use a constant

test/integration/tasks_test.rb
```diff
@@ -25,12 +19,13 @@
   end
 
   it "produces the current schema from checked in migrations" do
-    tasks["db:schema:dump"].execute
+    # Loading all tasks here results in a circular import due to Sentry::Tasks. Only loading the necessary task.
```
grosser, line 22: this seems like a bandaid for a deeper problem, loading all tasks should be possible ...

PR #4021, opened 2022-07-14
replace with json: allow use of \\. to avoid splitting

grosser:
### Risks
- Low: deploys failing because we split incorrectly


rajeeshckr (approved):
Looks good! Thanks @grosser

PR #4024, opened 2022-07-21
allow inlining commands for stages

grosser:
to do an automate migration we want to use the api to check what commands exist and then update them
updating them is already supported via command_ids, but getting the, was not

SCR-20220721-ky5

### Risks
- Low


shoukoo (approved):
Thanks

PR #4027, opened 2022-08-27
feat: allow setting kubernetes well-known labels

a7i:
Signed-off-by: Amir Alavi <amir.alavi@zendesk.com>

For PR description, please view README addition.
For comparison, other deployments tools such as Spinnaker add these two labels as well. See Doc [here](https://spinnaker.io/docs/reference/providers/kubernetes-v2/#reserved-labels)

* Add description ✅ 
* ~Add screenshots when changing the UI~
* Add unit tests ✅ 

### References
N/A

### Risks
- Low: these well-known labels are not immutable nor selector labels so they should not affect functionality.

### DIFF
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -337,3 +337,12 @@ If a warning event fails deploys, but application owners deem them safe to ignor
 
 When using the namespaces UI to create new namespaces, set `KUBERNETES_COPY_SECRETS_TO_NEW_NAMESPACE=my-docker-auth,other-stuff`,
 it will then copy that secret from the `default` namespace to any newly created namespace.
+
+### Adding Well-Known Labels
+
+In accordance with [Kubernetes Well-Known Labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-managed-by),
+Samson can set the labels:
+- `app.kubernetes.io/managed-by` to `samson`
+- `app.kubernetes.io/name` to the resource name
+
+This feature can be enabled by setting `KUBERNETES_ADD_WELL_KNOWN_LABELS=true`.
--- a/plugins/kubernetes/README.md
+++ b/plugins/kubernetes/README.md
@@ -343,6 +343,6 @@ it will then copy that secret from the `default` namespace to any newly created 
 In accordance with [Kubernetes Well-Known Labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-managed-by),
 Samson can set the labels:
 - `app.kubernetes.io/managed-by` to `samson`
-- `app.kubernetes.io/name` to the resource name
+- `app.kubernetes.io/name` to the project permalink
 
 This feature can be enabled by setting `KUBERNETES_ADD_WELL_KNOWN_LABELS=true`.
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -6,6 +6,7 @@ module Kubernetes
 
     SECRET_PULLER_IMAGE = ENV['SECRET_PULLER_IMAGE'].presence
     KUBERNETES_ADD_PRESTOP = Samson::EnvCheck.set?('KUBERNETES_ADD_PRESTOP')
+    KUBERNETES_ADD_WELL_KNOWN_LABELS = Samson::EnvCheck.set?('KUBERNETES_ADD_WELL_KNOWN_LABELS')
     SECRET_PREFIX = "secret/"
     DOCKERFILE_NONE = 'none'
     DEFAULT_TERMINATION_GRACE_PERIOD = 30
@@ -58,6 +59,8 @@ module Kubernetes
         else
           set_name
         end
+
+        set_well_known_labels # last in order to ensure app name has been set
 
         template
       end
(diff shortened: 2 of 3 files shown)


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -630,6 +631,20 @@ def set_update_timestamp
...
+      return unless KUBERNETES_ADD_WELL_KNOWN_LABELS
+
+      metadata_labels_paths.each do |path|
+        next unless labels = template.dig(*path)
+
+        # always overwrite managed-by label since it is managed by samson
+        labels[:"app.kubernetes.io/managed-by"] = "samson"
+
+        # do not overwrite existing name label, if already set
+        labels[:"app.kubernetes.io/name"] = template[:metadata][:name] unless labels.key?(:"app.kubernetes.io/name")
```
grosser, line 644:
```suggestion
        labels[:"app.kubernetes.io/name"] ||= template[:metadata][:name]
```

grosser (commented):
won't this give us lots of broken names ? (since we just set whatever the name is to the name field and that's not consistent)

a7i:
> won't this give us lots of broken names ? (since we just set whatever the name is to the name field and that's not consistent)

I'm not too familiar with the code base, would you elaborate? What should it be set to?

grosser:
I don't think what we want is to put every metadata.name into the canonical name field.
(If it were that easy k8s would already do it)
We want the apps name, which afaik is the project label.
... and for those where it's not the project label the naming would be random if we just guessed it.
So for example a project might have a name, but the deployment is named "foo-worker", but that should not go into the name field.

So:
if we want to set it for consistency: use the project label.
if we want to set it because the project label is not good enough: better just not set it automatically.

a7i:
> I don't think what we want is to put every metadata.name into the canonical name field. (If it were that easy k8s would already do it) We want the apps name, which afaik is the project label. ... and for those where it's not the project label the naming would be random if we just guessed it. So for example a project might have a name, but the deployment is named "foo-worker", but that should not go into the name field.
> 
> So: if we want to set it for consistency: use the project label. if we want to set it because the project label is not good enough: better just not set it automatically.

project label makes sense but in that case, we could also set `app.kubernetes.io/instance` (to `role`). `project` and `label` are "proprietary" labels only understood by Samson or in-house custom tooling. `app.kubernetes.io/name` is a canonical label well understood by other software in the k8s space (e.g. [Istio](https://github.com/istio/istio/blob/4cb2fcf9195138ce331d0120b7e37b8fa5add1b2/pkg/kube/labels/labels.go#L26)). This is why Helm, Spinnaker, Tekton (and others) automatically add this label so then it raises the question as to why shouldn't Samson?

plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -630,6 +630,20 @@ def set_update_timestamp
...
+      return unless KUBERNETES_ADD_WELL_KNOWN_LABELS
+
+      metadata_labels_paths.each do |path|
+        next unless labels = template.dig(*path)
+
+        # always overwrite managed-by label since it is managed by samson
+        labels[:"app.kubernetes.io/managed-by"] = "samson"
+
+        # do not overwrite existing name label, if already set
+        labels[:"app.kubernetes.io/name"] ||= project.name
```
a7i, line 643: I set this to `project.name` as opposed to `project.permalink`. The rationale is that these are not immutable labels and if the project name is changed, so should the label.

However, this will be different from how the `set_project_label` works, where `permalink` is used. Thoughts? Concerns?

grosser: we use the name for stats, so permalink might be safer ... but 🤷

plugins/kubernetes/README.md
```diff
@@ -337,3 +337,12 @@ If a warning event fails deploys, but application owners deem them safe to ignor
 
 When using the namespaces UI to create new namespaces, set `KUBERNETES_COPY_SECRETS_TO_NEW_NAMESPACE=my-docker-auth,other-stuff`,
 it will then copy that secret from the `default` namespace to any newly created namespace.
+
+### Adding Well-Known Labels
+
+In accordance with [Kubernetes Well-Known Labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-managed-by),
+Samson can set the labels:
+- `app.kubernetes.io/managed-by` to `samson`
+- `app.kubernetes.io/name` to the resource name
```
grosser, line 346:
```suggestion
- `app.kubernetes.io/name` to the project permalink
```

grosser (approved):

plugins/kubernetes/app/models/kubernetes/template_filler.rb
adammw: slight preference for permalink just since its usually a nicer tag value and doesn't add more gotchas/duplication

adammw (approved):

aglover-zendesk (approved):

jklaw90 (approved):
👏

rubysocks (approved):
👍 from Deploy team.

grosser:
Does helm set different labels when a project has multiple components ?
I just don't want us to set random values in bless there is a clear benefit
for setting something.

Issue #4029, opened 2022-09-21
Latest releases not published to Docker Hub


jamesmbourne:
It seems that the most recent version available on Docker Hub is [`v3346`](https://hub.docker.com/layers/zendesk/samson/v3346/images/sha256-1e1ae2cdf364dd7f665cc8632045d02bd0801adec9d17c6324a9a7cec3c00522?context=explore) but the most recent GitHub release is [`v3362`](https://github.com/zendesk/samson/releases/tag/v3362)

I cannot see any builds more recent than `v3346` which ran 5 months ago on https://hub.docker.com/r/zendesk/samson/builds/?page=1

Thanks!

jamesmbourne:
Thanks @grosser - appreciate your efforts in continuing to maintain this. I will take a look at building from source and see if I can get the changes back into the upstream repo.

grosser:
this was easy ... https://github.com/zendesk/samson/pull/4031
problem was that docker builds were [failing](https://hub.docker.com/repository/registry-1.docker.io/zendesk/samson/builds/7cccad22-bcb8-4be1-91e0-4d062944dfae)
... I had feared it's another auth or build setup issue :)

PR #4031, opened 2022-10-09
fix docker build

grosser:
[failure](https://hub.docker.com/repository/registry-1.docker.io/zendesk/samson/builds/7cccad22-bcb8-4be1-91e0-4d062944dfae)

fixes https://github.com/zendesk/samson/issues/4029

### Risks
- Low

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 FROM ruby:2.7.6-slim
 
 # Install dependencies
-RUN apt-get update \
-  && apt-get install -y --no-install-recommends \
+RUN \
+  apt-get update && \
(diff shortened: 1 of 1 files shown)


adammw:
Will probably end up with old versions of node/npm in the image... might be fine considering the minimal use at runtime?

adammw (approved):

grosser:
there is no runtime use since packages are pre-generated

grosser:
```
npm -v
7.5.2
```
and before we had 6, so it's a step up :D

grosser:
worked :)
SCR-20221012-d82

PR #4056, opened 2023-05-02
Update to latest ruby and Nokogiri point releases

zendesk-mattlefevre:
## Description
Update to latest Ruby and Nokogiri point-release.

### Risks
- Low/Med/High: Low


Gemfile.lock
```diff
@@ -171,28 +171,9 @@ PATH
     samson_zendesk (0.0.0)
       zendesk_api
 
-GEM
```
grosser, line 174: try with `bundle update --bundler` and see if that goes away ... because when I bundle it, it comes right back

zendesk-mattlefevre: That worked perfectly. Thanks, @grosser .

grosser (approved):

grosser:
yeah, just merge with CI failing to keep this one small

PR #4057, opened 2023-05-03
allow not setting replicas for deployments

grosser:
ideally that would be automatic, but that could break lots of things ... so using a flag to opt-in instead

also: cleaning up support for ShardedDeployment since that is EOL

### Risks
- Low: deploys failing on some kinds being configured with multiple replicas

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/template_filler.rb
+++ b/plugins/kubernetes/app/models/kubernetes/template_filler.rb
@@ -44,7 +44,11 @@ module Kubernetes
           make_stateful_set_match_service if kind == 'StatefulSet'
           set_pre_stop if kind == 'Deployment'
           set_name
-          (set_replica_target || validate_replica_target_is_supported) if kind != 'PodTemplate'
+          if ['Deployment', 'StatefulSet'].include?(kind)
+            set_replica_target
+          else
+            validate_replica_target_is_supported
+          end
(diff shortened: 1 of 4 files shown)


plugins/kubernetes/app/models/kubernetes/template_filler.rb
```diff
@@ -44,7 +44,11 @@ def to_hash(verification: false)
           make_stateful_set_match_service if kind == 'StatefulSet'
           set_pre_stop if kind == 'Deployment'
           set_name
-          (set_replica_target || validate_replica_target_is_supported) if kind != 'PodTemplate'
+          if ['Deployment', 'StatefulSet'].include?(kind)
```
a7i, line 47: this is a change in behavior, correct? Previously it would attempt to set replicas on anything (comment "custom resource that has replicas set on itself or it's template")

grosser: yeah, that was 1-off logic for shardeddeploments that we no longer need 🤞

a7i: ah I see!

a7i (approved):
🏆

grosser:
had to also add that server-side-apply is enabled or this ends up setting 1, since that's the replica default when using "replace" workflow

PR #4060, opened 2023-05-18
podtemplate: allow setting 0 replicas

grosser:
last PR https://github.com/zendesk/samson/pull/4057 remove that feature by accident
but it's useful since podtemplates should not count into resource usage math

### Risks


ragravi (approved):
Thanks!! LGTM

jklaw90 (approved):

PR #4061, opened 2023-05-18
unblock deploys + fix audit

grosser:
### Risks
- Low: we are on EOL version either way and should only be for a short time


zendesk-mattlefevre (approved):

zendesk-mattlefevre (changes requested):
Gotta update the dockerfile to match the ruby-version file.

zendesk-mattlefevre (approved):

PR #4069, opened 2023-07-14
remove oauth applications from manage dropdown

anthonywoo:
### Description
Currently a single user could delete all access tokens if they delete the associated OAuth Application. I want to make it more difficult for a single user to make a mistake. Remove the OAuth Application option from the Manage dropdown menu. The OAuth Application UI can still be accessed via `/oauth/applications`

An alternative approach may also be to duplicate the view from the doorkeeper gem into Samson and remove the Delete link

Also happy to discuss other alternatives

### Risks
- Low. Remove option from dropdown, but the view still exists


grosser (approved):
yeah that's fair :D

anthonywoo:
Forgot about this PR, merging with bypass as change is minimal

PR #4071, opened 2023-08-04
safely deserialize k8 yaml

anthonywoo:

Using [YAML.load_stream](https://github.com/zendesk/samson/blob/107efb4a252425966aac5e77d0c3670f9b5d7229/plugins/kubernetes/app/models/kubernetes/util.rb#L9) is unsafe and allows instantiating arbitrary classes which may lead to remote code execution (RCE)

I had to use multiple streams because Samson needs to support multiple documents in a single yaml file

### References
- Jira link: 

### Risks
- Medium? Some k8 deploys may fail if it is using a disallowed class

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/util.rb
+++ b/plugins/kubernetes/app/models/kubernetes/util.rb
@@ -20,14 +20,11 @@ module Kubernetes
     end
 
     def self.yaml_safe_load_stream(contents, filename)
-      documents = []
-
-      YAML.parse_stream(contents, filename: filename).children.each do |child|
+      YAML.parse_stream(contents, filename: filename).children.map do |child|
         temp_stream = Psych::Nodes::Stream.new
         temp_stream.children << child
-        documents << YAML.safe_load(temp_stream.to_yaml, [Symbol], aliases: true)
+        YAML.safe_load(temp_stream.to_yaml, [Symbol], aliases: true)
       end
-      documents
     end
   end
 end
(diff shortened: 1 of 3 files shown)


grosser (commented):
there are no other usages of this method ?

anthonywoo:
> there are no other usages of this method ?

I couldn't find any other usages where the stream content can come from an untrusted source

plugins/kubernetes/app/models/kubernetes/util.rb
```diff
@@ -18,5 +18,16 @@ def self.log(message, extra_info = {})
...
+
+    def self.yaml_safe_load_stream(contents, filename)
+      documents = []
+
+      YAML.parse_stream(contents, filename: filename).children.each do |child|
+        temp_stream = Psych::Nodes::Stream.new
+        temp_stream.children << child
+        documents << YAML.safe_load(temp_stream.to_yaml, [Symbol], aliases: true)
+      end
+      documents
```
grosser, line 30: nit:
```suggestion
      YAML.parse_stream(contents, filename: filename).children.map do |child|
        temp_stream = Psych::Nodes::Stream.new
        temp_stream.children << child
        YAML.safe_load(temp_stream.to_yaml, [Symbol], aliases: true)
      end
```

grosser (approved):
nice test case 🎉

grosser:
don't worry about bundle_audit that can be another PR

PR #4072, opened 2023-08-24
bump ruby patch version, oauth2, and faraday

zendesk-mattlefevre:
### Description
Vulnerability bumps
- Bump the ruby patch version to 2.7.8
- Bump the Faraday gem's patch version
- Upgrade to oauth2 version 2.0.*

### Risks
- Med: I believe I caught all the changes that the major bump for the Oauth2 gem requires, but that's something to keep an eye on.


grosser (approved):

config/logging.rb
```diff
@@ -44,7 +44,7 @@
   config.lograge.custom_options = ->(event) do
     # show params for every request
     unwanted_keys = ['format', 'action', 'controller']
-    params = event.payload[:params].reject { |key, _| unwanted_keys.include? key }
+    params = event.payload[:params]&.reject { |key, _| unwanted_keys.include? key }
```
grosser, line 47: looks a bit sus ... make sure it still logs other params fine

zendesk-mattlefevre: I originally added this in because we're getting a bunch of errors that event.payload[params] is Nil. I'll dig in and see if this is going to cause more problems than it solves.

zendesk-mattlefevre: Tested this in staging and it didn't fix it anyway 😆 . Reverting.

PR #4074, opened 2023-09-15
Update Ruby and Rails

sarahjmiller:

* Add description
* Add unit tests

### References
- Jira link: 

### Risks


grosser (approved):

sarahjmiller:
We reverted the PR for v3386 because we noticed a secret resolution issue after deploying. But we were unable to identify how any of the security bumps in that PR could have caused it. We planned to apply the bumps iteratively (hence this PR) but we tested v3386 again and did not see the same secret resolution issue. We now believe it was a transitory connection issue with one or more of our secret backends that happened to coincide with the previous release, negating the need for this PR.

PR #4076, opened 2023-10-31
Remove reporting of unqueried metrics.

sarahjmiller:

The metrics cost optimization project requested that we remove the process stats being reported by Samson, which are not queried by any existing monitors. I've commented it out so that it will be easy to reinstate if we do need to monitor the process stats.

### References
- Jira link: https://zendesk.atlassian.net/browse/DPLY-3582

### Risks
- Low: only removing stats reporting that aren't being monitored.

### DIFF
--- a/config/initializers/periodical.rb
+++ b/config/initializers/periodical.rb
@@ -46,7 +46,7 @@ end
 # These are unqueried and were removed at the request of the metrics cost optimization project
 # Uncomment if creating a monitor for process stats.
 # Samson::Periodical.register :report_process_stats, "Report process stats" do
-  # Samson::ProcessUtils.report_to_statsd
+#   Samson::ProcessUtils.report_to_statsd
 # end
 
 Samson::Periodical.register :repo_provider_status, "Refresh repo provider status" do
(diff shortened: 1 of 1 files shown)


grosser (commented):
- why is this even an issue when these have ~10 total series ?
- we should monitor periodicals taking too long because that means they are doing something funky (either consuming too many resources or stuck in a loop)

sarahjmiller:
Thanks for the feedback, @grosser!

> why is this even an issue when these have ~10 total series ?

In the [metrics cost optimization request](https://zendesk.slack.com/archives/C0206JAJFPU/p1697159279732069) they mention both an improvement to ingest cost and cardinality cost, so even a small number of series might make a difference?

> we should monitor periodicals taking too long because that means they are doing something funky (either consuming too many resources or stuck in a loop)

The request also says that these metrics aren't queried by any existing dashboards, monitors, or users. If Compute is monitoring these metrics, would you be willing to respond in the [Slack thread](https://zendesk.slack.com/archives/C0206JAJFPU/p1697159279732069) with the rationale for keeping them?

grosser:
ah nvm I read the code wrong ... this is just process list
that would have big cardinality and be not generally useful 🎉

grosser (approved):

PR #4078, opened 2023-11-02
Bump omniauth gems

zendesk-mattlefevre:
## Description
Bump `omniauth` and related gems to patch known vulns.

### Risks
- Med: This does touch authentication, so this could cause problems


grosser (approved):
looks good
make sure to verify locally+in staging

Carrie-Crow (approved):
Can confirm that I was able to log into and out of staging reliably!

Gemfile
```diff
@@ -25,11 +25,13 @@ gem 'attr_encrypted'
...
 gem 'oauth2', '~>2.0.9'
-gem 'omniauth'
+gem 'omniauth', '~>2.0'
 gem 'omniauth-oauth2'
 gem 'omniauth-github', git: "https://github.com/omniauth/omniauth-github.git" # needs >1.3.0
 gem 'omniauth-google-oauth2'
-gem 'omniauth-ldap'
+gem 'omniauth-ldap',
+  git: 'https://github.com/omniauth/omniauth-ldap.git',
+  ref: 'fb485bb4613074b224b8d6b28ad1c7f366a6b6d8'
```
bquorning, line 34: Why not use v2.0.1 (`~> 2.0.0`) instead of a specific git sha?

grosser: latest is 2.0.0 https://rubygems.org/gems/omniauth-ldap

bquorning: Why not use v2.0.0 (`~> 2.0.0`) instead of a specific git sha?

grosser: that [release](https://github.com/omniauth/omniauth-ldap/commit/3242f85968f26938d9a6ce028e511e788cc54722) is 6y old, so assuming we need some recently merged fix

grosser: ... we might want to just rip this out, doubt there are many uses

PR #4079, opened 2023-11-08
Bump Doorkeeper Gem

zendesk-mattlefevre:

## Description
To keep up on known vulnerabilities, we are updating to the latest version of the Doorkeeper gem. 

### Risks
- Med: While only a minor version bump, the change does seem to change some logic we touch. We'll want to test this thoroughly.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ group :preload do
   gem 'pagy'
   gem 'audited'
   gem 'soft_deletion'
-  gem 'doorkeeper', '~> 5.4.0' # TODO: upgrade breaks `rails c`
+  gem 'doorkeeper', '~> 5.6.0' # TODO: upgrade breaks `rails c`
   gem 'mail', '~> 2.7.0' # TODO: upgrade breaks `rails c`
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ group :preload do
   gem 'pagy'
   gem 'audited'
   gem 'soft_deletion'
-  gem 'doorkeeper'
+  gem 'doorkeeper', '~>5.6.0'
   gem 'mail', '~> 2.7.0' # TODO: upgrade breaks `rails c`
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ group :preload do
   gem 'pagy'
   gem 'audited'
   gem 'soft_deletion'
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -15,7 +15,7 @@ group :preload do
   gem 'pagy'
   gem 'audited'
   gem 'soft_deletion'
-  gem 'doorkeeper', '~> 5.4.0' # TODO: upgrade breaks `rails c`
+  gem 'doorkeeper', '~> 5.6.0' # TODO: upgrade breaks `rails c`
```
grosser, line 18:
```suggestion
  gem 'doorkeeper', '~> 5.6.0'
```

config/initializers/doorkeeper.rb
```diff
@@ -11,15 +11,15 @@ module Doorkeeper
...
-  module Orm
-    module ActiveRecord
-      class << self
-        undef initialize_models!
-        def initialize_models!
-        end
-      end
-    end
-  end
+  # module Orm
```
grosser, line 14: should delete or explain why it's commented out

config/initializers/doorkeeper.rb
```diff
@@ -11,15 +11,15 @@ module Doorkeeper
...
+  #   module ActiveRecord
+  #     class << self
+  #       # undef initialize_models!
+  #       def initialize_models!
+  #       end
+  #     end
+  #   end
+  # end
```
grosser, line 22: redo test explained above to make sure this is safe

config/initializers/doorkeeper.rb
```diff
@@ -11,15 +11,6 @@ module Doorkeeper
...
   autoload :RedirectUriValidator, 'doorkeeper/orm/active_record/redirect_uri_validator'
-  module Orm
-    module ActiveRecord
-      class << self
-        undef initialize_models!
-        def initialize_models!
-        end
-      end
-    end
-  end
```
sarahjmiller, line 22: Just curious why this needed to be deleted.

sarahjmiller (approved):
Looks good to me! Approving for Deploy Experience team, although we may also want to be sure to get sign off from @grosser.

config/initializers/doorkeeper.rb
zendesk-mattlefevre: This function no longer exists in the Doorkeeper gem, so we don't need to replace it with an empty method.

grosser (approved):

PR #4085, opened 2023-12-18
kubernetes: dynamic file name based on deploy_group permalink

a7i:

* Add description: Provide the ability to perform dynamic file name substitution based on deploy_group permalink
* Add screenshots when changing the UI ❌ 
* Add unit tests ✔️ 

### References
- Jira link: 

### Risks
- Low: only impacts new apps with this new pattern in the file name

### DIFF
--- a/plugins/kubernetes/app/models/kubernetes/role.rb
+++ b/plugins/kubernetes/app/models/kubernetes/role.rb
@@ -164,6 +164,7 @@ module Kubernetes
       file = config_file
       if deploy_group && dynamic_folders?
         file = file.
+          sub('$deploy_group_permalink', deploy_group.permalink).
           sub('$deploy_group', deploy_group.env_value).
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/app/models/kubernetes/role.rb
```diff
@@ -164,6 +164,7 @@ def role_config_file(reference, deploy_group:, project: project(), **args) # rub
       file = config_file
       if deploy_group && dynamic_folders?
         file = file.
+          sub('$deploy_group_permalink', deploy_group.permalink).
```
a7i, line 167: it has to come before `$deploy_group` otherwise it will replace it to `pod998_permalink`

grosser (approved):
kinda hate that we need it ... but not too bad ...

a7i:
wasn't needed so closing

PR #4086, opened 2024-01-10
add nil check if there is no commit message

anthonywoo:

Fix an issue where the deploys page breaks if a commit has no message

### References
- Jira link: 

### Risks
- Low

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,2 +1,0 @@
----
-BUNDLE_WITHOUT: "nothing:postgres"
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,0 +1,2 @@
+---
+BUNDLE_WITHOUT: "nothing:postgres"
--- a/app/models/changeset/commit.rb
+++ b/app/models/changeset/commit.rb
@@ -28,13 +28,7 @@ class Changeset::Commit
   end
 
   def summary_long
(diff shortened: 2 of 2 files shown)


rsb (approved):

app/models/changeset/commit.rb
```diff
@@ -28,7 +28,13 @@ def summary
...
 
   def summary_long
-    @data.commit.message.split("\n").first
+    summary = @data.commit.message.split("\n").first
+
+    if summary.nil?
+      return ""
+    end
+
+    summary
```
grosser, line 37:
```suggestion
    @data.commit.message.split("\n").first || ""
```

grosser (approved):

grosser (commented):
test case would also be nice

app/models/changeset/commit.rb
rsb: This has been updated to reflect your one liner.

.bundle/config
```diff
@@ -0,0 +1,2 @@
+---
+BUNDLE_WITHOUT: "nothing:postgres"
```
grosser, line 2: should be in your global .gitignore

rsb: removed. Thanks

grosser (approved):

PR #4089, opened 2024-02-09
Align `as_json` method (use args instead of kwargs)

orien:
### Context

There's a change in Ruby 3.0 that [separates positional and keyword arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/).

### Change

Some Samson classes have the `as_json` method overridden. The [method definition in the Rails superclass](https://api.rubyonrails.org/classes/ActiveModel/Serializers/JSON.html) accepts an options hash argument. However, the method definition in the subclasses is changed to accept keyword arguments instead. This inconsistency won't work on Ruby 3.

In preparation for the Ruby 3.0 upgrade, align the subclass method definitions with the superclass by accepting an options hash argument instead of keyword arguments.

### Considerations

This change is far from resolving all the args/kwargs issues when running on Ruby 3.0, but it's a safe start.


grosser (approved):
looks right 👍

orien:
Thanks @grosser. Could please get your help to merge this PR as well?

PR #4090, opened 2024-02-09
Tests: use keyword arguments for Ruby 3 compatibility

orien:
### Context

There's a change in Ruby 3.0 that [separates positional and keyword arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/).

### Change

In preparation for the Ruby 3.0 upgrade, update tests to correctly pass keyword arguments in method calls.


test/lib/samson/form_builder_test.rb
```diff
@@ -172,9 +172,9 @@ def fake_erb_rendering
 
   # NOTE: ideally don't use a plugin model, but we need something with accepts_nested_attributes_for
   describe '#fields_for_many' do
-    def render(*args)
+    def render(*args, **kwargs)
       project.rollbar_dashboards_settings.build # TODO: this does not get rendered
-      builder.fields_for_many(*args) do |p|
+      builder.fields_for_many(*args, **kwargs) do |p|
```
grosser, line 177: FYI could use `...` here too

grosser (approved):

test/lib/samson/form_builder_test.rb
orien: > FYI could use `...` here too

Yes. However, I prefer this way as it makes it clear that block arguments are not passed along with args and kwargs.

orien:
Thanks @grosser. Could please get your help to merge this PR as well?

PR #4097, opened 2024-02-12
use ruby 3.2

grosser:
can't install 3.0 locally so let's go further

### Risks
- Low: all kinds of errors


a7i (approved):

grosser:
confirmed working 🎉

PR #4108, opened 2024-02-29
pass kwargs in history

zendesk-mattlefevre:
We're seeing an error in lib/samson/secrets/manager.rb where the wrong number of values are being passed. Aiming to fix this error in this PR.

### Risks


lib/samson/secrets/manager.rb
```diff
@@ -38,7 +38,7 @@ def read(id, *args, include_value: false)
         end
 
         def history(id, include_value: false, **options)
-          history = backend.history(id, options) || raise(ActiveRecord::RecordNotFound)
+          history = backend.history(id, **options) || raise(ActiveRecord::RecordNotFound)
```
grosser, line 41: can you make a test that fails when this is not added (I assume by using the vault backend in the test)
... or somehow changing the current testing backend to behave the same as the backend that failed

grosser (approved):

PR #4113, opened 2024-08-02
Apply security patches for dependencies

orien:
Apply patches for software dependencies:

- Bump Ruby from 3.2.4 to 3.2.5
- Bump Bundler from 2.5.11 to 2.5.17
- Bump rexml from 3.2.8 to 3.3.4

Resolves:

- CVE-2024-39908
- CVE-2024-41123
- CVE-2024-41946

PR #4114, opened 2024-08-05
Dont error if secret visible key does not exist 

anthonywoo:

Secrets can be created outside of Samson, and then managed through Samson. If a secret was created outside of Samson, it is possible for the secret not to have the `visible` key and users won't be able to access that secret in Samson

### References
- Jira link: 

### Risks
- Low. There may be other issues with secrets created outside of Samson

### DIFF
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -228,7 +228,7 @@ class SecretsController < ApplicationController
 
   def find_secret
     @secret = Samson::Secrets::Manager.read(id, include_value: true)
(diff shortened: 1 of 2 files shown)


grosser (approved):
best also add a test that fails without this fix

anthonywoo:
@grosser added a test. `bundle_audit` is failing, but that should be addressed in this [PR](https://github.com/zendesk/samson/pull/4113)

unless there are other concerns, my plan is to merge and deploy this tomorrow morning

PR #4116, opened 2024-09-21
We don't use Fossa anymore

bquorning:
### Risks
- None


grosser:
can you also remove [this](https://github.com/zendesk/samson/blob/27d9522c398b66cc6ca959d9d610cbae09f08bf3/.github/workflows/actions.yml#L91)

bquorning:
This PR should be ready to merge now @grosser.

PR #4124, opened 2025-01-13
Return error output when git commands fail

orien:
Following on from #3738. This time minimising changes by using a triplet array.

Context: Envato has been running Samson with these changes for a few years now.

### DIFF
--- a/app/models/git_repository.rb
+++ b/app/models/git_repository.rb
@@ -184,13 +184,13 @@ class GitRepository
   # success: stdout as string
   # error: nil
   def capture_stdout(*command, dir: repo_cache_dir)
-    success, output = Samson::CommandExecutor.execute(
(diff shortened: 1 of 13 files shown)


grosser (commented):
I like the idea/implementation,
I'm worried that this will hide stderr when executing a command via the UI though
(as in the err output will no longer show unless the command fails at the end, and it will only show at the end)
so for that case we should still interleave the stderr ?

orien:
> I like the idea/implementation, I'm worried that this will hide stderr when executing a command via the UI though (as in the err output will no longer show unless the command fails at the end, and it will only show at the end) so for that case we should still interleave the stderr ?

That's very true. I will simplify this.

grosser (approved):

grosser:
super clean solution 🎉

PR #4129, opened 2025-03-07
Patch Ruby dependencies to resolve security issues

orien:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:3.3.6-slim
+FROM ruby:3.3.7-slim
```
grosser, line 1: we might actually want to do 3.3 here to get automatic bumps, but that can be it's own PR

Relsecuren: I wonder if the new puch notifications I'm receiving are part of the update, they are reading ruby is telling me "to do the exploit?" Ruby has never prompted me to do an exploit.

